PR #7, opened 2019-03-05
Attach RDS read replicas to each other

achantavy:
New relationship:

```
(replica:RDSInstance)-[r:IS_READ_REPLICA_OF]->(source:RDSInstance)
```


cartography/intel/aws/rds.py
```diff
@@ -98,6 +104,33 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+    MATCH (replica:RDSInstance{id:{ReplicaArn}}),
+    (source:RDSInstance{db_instance_identifier:{SourceInstanceIdentifier}})
+    MERGE (replica)-[r:IS_READ_REPLICA_OF]->(source)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
+    """
+    for replica in read_replicas:
+        if not replica.get('ReadReplicaSourceDBInstanceIdentifier'):
+            logger.debug(f"Expected RDSInstance to be a read replica but its ReadReplicaSourceDBInstanceIdentifier "
+                         f"field is None.  Here is the object: {replica}")
```
ecdavis: Python's logging library handles string formatting in the "old" way and we use that through the rest of the codebase. For consistency's sake please use old-style formatting when logging. See https://github.com/lyft/cartography/blob/master/cartography/sync.py#L65

cartography/intel/aws/rds.py
```diff
@@ -98,6 +104,33 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+        elif not replica.get('DBInstanceArn'):
+            logger.debug(f"Expected RDSInstance to have a DBInstanceArn but it doesn't.  Here is the object: {replica}")
```
ecdavis: As above.

PR #9, opened 2019-03-06
Attach EC2 Security Groups to RDS Instances

achantavy:


cartography/intel/aws/rds.py
```diff
@@ -105,9 +105,33 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+        _attach_ec2_security_groups(neo4j_session, rds, aws_update_tag)
     _attach_read_replicas(neo4j_session, read_replicas, aws_update_tag)
 
 
+def _attach_ec2_security_groups(neo4j_session, instance, aws_update_tag):
+    """
+    Attach an RDS instance to its EC2SecurityGroups
+    """
+    attach_rds_to_group = """
+    MATCH (rds:RDSInstance{id:{RdsArn}}),
```
sachafaust: I know we have ARN as ID but we could benefit from having this simpler for our OSS users and use the ARN attribute (assuming it's indexed)

cartography/intel/aws/rds.py
```diff
@@ -105,9 +105,33 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+    (sg:EC2SecurityGroup{groupid:{GroupId}})
+    MERGE (rds)-[m:MEMBER_OF_EC2_SECURITY_GROUP]->(sg)
```
sachafaust: BLOCK - missing cleanup job?

sachafaust:
added blocking comment

cartography/intel/aws/rds.py
achantavy: See https://github.com/lyft/cartography/blob/master/cartography/data/indexes.cypher#L40.

We also index on RDS' db_instance_identifier which is more of a friendly name.  I can make it clearer in the schema docs that this is the case.

cartography/intel/aws/rds.py
achantavy: We clean up RDS instances [here](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/rds.py#L98) (actual job is [here](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json) and EC2 Security Groups [here](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/ec2.py#L583) (actual job is [here](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/cleanup/aws_import_ec2_security_groupinfo_cleanup.json).

sachafaust: What about MEMBER_OF_EC2_SECURITY_GROUP?

achantavy: [`detach delete`](https://neo4j.com/docs/cypher-manual/current/clauses/delete/#query-delete-introduction) removes the relationships as well.

sachafaust: What if both rds and security group are still there but they aren't associated anymore?

achantavy: Ok, we can add a statement to do this.

achantavy: Added more to the cleanup job, how's this look now?

sachafaust: LGTM

PR #10, opened 2019-03-07
Attach EC2Subnets to RDSInstances

achantavy:
Introduced:
```
(RDSInstance)-[:MEMBER_OF_DB_SUBNET_GROUP]->(DBSubnetGroup)

It is a many-to-1 relationship.
```
and
```
(DBSubnetGroup)-[:RESOURCE]->(EC2Subnet)

This is a 1-to-many relationship.
```


cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json
```diff
@@ -1,8 +1,15 @@
 {
-  "statements": [{
-    "query": "MATCH (n:RDSInstance)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
-    "iterative": true,
-    "iterationsize": 100
-  }],
+  "statements": [
+    {
+      "query": "MATCH (sng:DBSubnetGroup)<-[:PART_OF_DB_SUBNET_GROUP]-(:RDSInstance)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE sng.lastupdated <> {UPDATE_TAG} WITH sng LIMIT {LIMIT_SIZE} DETACH DELETE (sng) return COUNT(*) as TotalCompleted",
```
sachafaust: Are the SubnetGroup an RDS specific concept? If so, should we drive toward consistency and rename as RDSSubnectGroup ?

cartography/intel/aws/rds.py
```diff
@@ -98,6 +105,132 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+        _attach_ec2_subnet_groups(neo4j_session, rds, aws_update_tag)
+    _attach_read_replicas(neo4j_session, read_replicas, aws_update_tag)
+
+
+def _attach_ec2_subnet_groups(neo4j_session, instance, aws_update_tag):
+    """
+    Attach RDS instance to its EC2 subnets
+    """
+    attach_rds_to_subnet_group = """
+    MERGE(sng:DBSubnetGroup{id:{DBSubnetGroupName}})
```
sachafaust: Are the SubnetGroup an RDS specific concept? If so, should we drive toward consistency and rename as RDSSubnectGroup ?

cartography/intel/aws/rds.py
```diff
@@ -98,6 +105,132 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+    ON CREATE SET sng.firstseen = timestamp()
+    SET sng.vpc_id = {VpcId},
+    sng.description = {DBSubnetGroupDescription},
+    sng.status = {DBSubnetGroupStatus},
+    sng.lastupdated = {aws_update_tag}
+    WITH sng
+    MATCH(rds:RDSInstance{id:{DBInstanceArn}})
+    MERGE(rds)-[r:PART_OF_DB_SUBNET_GROUP]->(sng)
```
sachafaust: Seems to deviate from our naming convention of group membership/part. Example MEMBER_OF_RDS_SUBNET_GROUP ?

cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json
```diff
@@ -1,8 +1,15 @@
...
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:RDSInstance)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
+      "iterative": true,
+      "iterationsize": 100
+    }
+  ],
```
sachafaust: missing MERGE(sng)-[r:RESOURCE]->(subnet) cleanup ?

cartography/intel/aws/rds.py
```diff
@@ -98,6 +105,132 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+
+    From https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_VPC.WorkingWithRDSInstanceinaVPC.html:
+    `Each DB subnet group should have subnets in at least two Availability Zones in a given region. When creating a DB
+    instance in a VPC, you must select a DB subnet group. Amazon RDS uses that DB subnet group and your preferred
+    Availability Zone to select a subnet and an IP address within that subnet to associate with your DB instance.`
+    """
+    attach_subnets_to_sng = """
+    MATCH(subnet:EC2Subnet{subnetid:{SubnetIdentifier}}),
+    (sng:DBSubnetGroup{id:{DBSubnetGroupName}})
+    MERGE(sng)-[r:RESOURCE]->(subnet)
```
sachafaust: BLOCK - missing cleanup

cartography/intel/aws/rds.py
```diff
@@ -98,6 +105,132 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+
+
+def _attach_ec2_security_groups(neo4j_session, instance, aws_update_tag):
+    """
+    Attach an RDS instance to its EC2SecurityGroups
+    """
+    attach_rds_to_group = """
+    MATCH (rds:RDSInstance{id:{RdsArn}}),
+    (sg:EC2SecurityGroup{groupid:{GroupId}})
+    MERGE (rds)-[m:MEMBER_OF_EC2_SECURITY_GROUP]->(sg)
```
sachafaust: BLOCK - missing cleanup

cartography/intel/aws/rds.py
```diff
@@ -98,6 +105,132 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+
+
+def _attach_read_replicas(neo4j_session, read_replicas, aws_update_tag):
+    """
+    Attach read replicas to their source instances
+    """
+    attach_replica_to_source = """
+    MATCH (replica:RDSInstance{id:{ReplicaArn}}),
+    (source:RDSInstance{db_instance_identifier:{SourceInstanceIdentifier}})
+    MERGE (replica)-[r:IS_READ_REPLICA_OF]->(source)
```
sachafaust: BLOCK - missing cleanup

cartography/intel/aws/rds.py
achantavy: I believe so.  I don't want to name it an RDSSubnetGroup though because the [aws datatype](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBSubnetGroup.html) is actually called DBSubnetGroup.

cartography/intel/aws/rds.py
achantavy: I was originally following the PART_OF_SUBNET pattern, but I can change it to the MEMBER_OF_* format. @ecdavis: thoughts?

cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json
achantavy: EC2Subnets are already cleaned up [here](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/cleanup/aws_import_ec2_instances_cleanup.json#L13).

cartography/intel/aws/rds.py
achantavy: The DBSubnetGroups are cleaned up [here](https://github.com/lyft/cartography/pull/10/files#diff-db18c81378ce7f90e09bd4f343906bedR4) and the EC2Subnets are cleaned up [here](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/cleanup/aws_import_ec2_instances_cleanup.json#L13).

cartography/intel/aws/rds.py
achantavy: The DBSubnetGroups are cleaned up [here](https://github.com/lyft/cartography/pull/10/files#diff-db18c81378ce7f90e09bd4f343906bedR4) and the RDSInstances are cleaned up [here](https://github.com/lyft/cartography/pull/10/files#diff-db18c81378ce7f90e09bd4f343906bedR9).

cartography/intel/aws/rds.py
achantavy: RDS Instances are all cleaned up [here](https://github.com/lyft/cartography/pull/10/files#diff-db18c81378ce7f90e09bd4f343906bedR9).

cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json
sachafaust: The MERGE statement implies there are possibilities for creating a new else we could be explicit with MATCH.

It seems reasonable to keep MERGE as subnet may have been created since we did EC2 sync for example (creation while sync in running)... A general rule I try to keep for myself is if we do MERGE, we should cleanup that specific node or relationship.

cartography/intel/aws/rds.py
sachafaust: My thinking there is to be more explicit about what nodes are and coming from. We don't know what the future will bring and also what users will add. I do admit that we may not have considered that earlier on but something to revisit as we move forward. Looping @evandavis so we can reach consensus

cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json
achantavy: > The MERGE statement implies there are possibilities for creating a new else we could be explicit with MATCH.

Ohh thanks good point, I missed that!

> my 0.02$, sounds good to me to keep MERGE as subnet may have been created since we did EC2 sync for example (creation while sync in running)

In the specific case you mentioned, the ECSubnet wouldn't have the other relationships that it normally has.  I think I would rather not include the incomplete EC2Subnet in this case, so I like the idea of being strict with match.

cartography/intel/aws/rds.py
sachafaust: bit of consistency here would help. There is some consistency but not 100% there yet. Would prefer to avoid another and backtrack the previous ones once we agree on 1.

Current group nodes relationships:

(AWSGroup)<- MEMBER_AWS_GROUP
EC2SecurityGroup <- MEMBER_OF_EC2_SECURITY_GROUP
AutoScalingGroup <- MEMBER_AUTO_SCALE_GROUP
JamfComputerGroup <- MEMBER_OF_JAMF_GROUP
JamfMobileDeviceGroup <- MEMBER_OF_JAMF_GROUP

cartography/intel/aws/rds.py
```diff
@@ -98,6 +105,132 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+    """
+    attach_rds_to_subnet_group = """
+    MERGE(sng:DBSubnetGroup{id:{DBSubnetGroupName}})
+    ON CREATE SET sng.firstseen = timestamp()
+    SET sng.vpc_id = {VpcId},
+    sng.description = {DBSubnetGroupDescription},
+    sng.status = {DBSubnetGroupStatus},
+    sng.lastupdated = {aws_update_tag}
+    WITH sng
+    MATCH(rds:RDSInstance{id:{DBInstanceArn}})
```
sachafaust: uniqueness/block - Are we sure that DBSubnetGroupName are unique across all AWS Accounts? my gut feeling tells me we should create unique identifier ourselves

cartography/intel/aws/rds.py
sachafaust: referring to the RESOURCE relationship.

cartography/intel/aws/rds.py
sachafaust: What about the MEMBER_OF_EC2_SECURITY_GROUP relationship? modules doing MERGE should cleanup on their own and not assume other modules does it (ordering is not guaranteed)

cartography/intel/aws/rds.py
sachafaust: refering to IS_READ_REPLICA_OF

cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json
sachafaust: if our goal is to represent the state as accurately as possible, MERGE makes sense to me. It also avoids trying to figure out what other modules are doing and ordering. Imagine RDS runs before EC2 (lamba, queue ...) in the future...

cartography/intel/aws/rds.py
achantavy: Got it, fixed.

cartography/intel/aws/rds.py
achantavy: Good point.  Got the ARN now and am using that now.

cartography/data/jobs/cleanup/aws_import_rds_instances_cleanup.json
achantavy: I'm deciding to not add new EC2Subnet nodes here.  We will just rely on the data provided by the previously run EC2 module since in this RDS module the only field we have is the subnet ID and nothing else.

achantavy: doing merge as per convo

cartography/intel/aws/rds.py
```diff
@@ -106,16 +106,96 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+    MERGE(rds)-[r:MEMBER_OF_DB_SUBNET_GROUP]->(sng)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
+    """
+    if not instance.get('DBInstanceArn'):
+        logger.debug("Expected RDSInstance to have a DBInstanceArn but it doesn't.  Here is the object: %r", instance)
+    elif not instance.get('DBSubnetGroup'):
+        logger.debug("Expected RDSInstance to have a DBSubnetGroup but it doesn't.  Here is the object: %r", instance)
```
ecdavis: Did you see cases where one of these values was missing?

cartography/intel/aws/rds.py
```diff
@@ -106,16 +106,96 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+    else:
+        db_sng = instance.get('DBSubnetGroup')
```
ecdavis: Use direct indexing instead of get here.

cartography/intel/aws/rds.py
```diff
@@ -106,16 +106,96 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+        arn = _get_db_subnet_group_arn(region, current_aws_account_id, db_sng.get('DBSubnetGroupName'))
+        neo4j_session.run(
+            attach_rds_to_subnet_group,
+            sng_arn=arn,
+            DBSubnetGroupName=db_sng.get('DBSubnetGroupName'),
+            VpcId=db_sng.get("VpcId", None),
```
ecdavis: dict.get's default return value is always None so there's no need to specify it here (and in the calls below)

cartography/intel/aws/rds.py
```diff
@@ -106,16 +106,96 @@ def load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_
...
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag},
+    subnet.availability_zone = {SubnetAvailabilityZone},
+    subnet.lastupdated = {aws_update_tag}
+    """
+    # TODO: Note: RDS data has subnet availability zone data, but I don't think that this is the right place to
+    # TODO: update it.  We should update the availability zone data in the EC2 ingestion part.
+    if not db_subnet_group.get('DBSubnetGroupName'):
+        logger.debug("Expected DBSubnetGroup to have a DBSubnetGroupName but it doesn't.  Here is the object: %r",
+                     db_subnet_group)
```
ecdavis: As above, did you see a case where this was missing?

cartography/intel/aws/rds.py
```diff
@@ -169,9 +249,18 @@ def _validate_rds_endpoint(rds):
...
 
 
-def cleanup_rds_instances(neo4j_session, common_job_parameters):
+def _get_db_subnet_group_arn(region, current_aws_account_id, db_subnet_group_name):
+    """
+    Return an ARN for the DB subnet group name by concatenating the account name and region.
+    This is done to avoid another AWS API call since the describe_db_instances boto call does not return the DB subnet
+    group ARN.
+    """
+    return f"arn:aws:{region}:{current_aws_account_id}:{db_subnet_group_name}"
```
ecdavis: Format for subnet ARNs is `arn:aws:ec2:region:account-id:subnet/subnet-id`.

cartography/intel/aws/rds.py
```diff
@@ -185,4 +274,4 @@ def sync_rds_instances(neo4j_session, boto3_session, regions, current_aws_accoun
         logger.info("Syncing RDS for region '%s' in account '%s'.", region, current_aws_account_id)
         data = get_rds_instance_data(boto3_session, region)
         load_rds_instances(neo4j_session, data, region, current_aws_account_id, aws_update_tag)
-    cleanup_rds_instances(neo4j_session, common_job_parameters)
+        cleanup_rds_instances_and_db_subnet_groups(neo4j_session, common_job_parameters)
```
ecdavis: Was the indentation change intentional? Why?

achantavy: good catch thanks

cartography/intel/aws/rds.py
achantavy: Didn't see this particular case, but I also didn't want to make too strong an assumption on the data that would be returned and have this crash a sync.  

There was another case with a different module where we made an assumption and it crashed the sync but I can't recall it right now.

cartography/intel/aws/rds.py
achantavy: Great catch and good doc, thanks

PR #14, opened 2019-03-12
Add docs on extending with Analysis Jobs

achantavy:


docs/writing-analysis-jobs.md
```diff
@@ -0,0 +1,126 @@
+# How to extend Cartography with Analysis Jobs
+
+## Intro
+In a nutshell, Analysis Jobs let you add your own customizations to Cartography by writing Neo4j queries.  This helps you add powerful enhancements to your data without the need to write Python code.
+
+### The stages
+There are 3 stages to a cartography sync.  First we create database indexes, next we ingest AWS assets, and finally we can run Analysis Jobs on the database (see [cartography.sync.build\_default\_sync()](https://github.com/lyft/cartography/blob/master/cartography/sync.py)).  This tutorial focuses on Analysis Jobs.
+
+### How to run
+Each Analysis Job is a JSON file with a list of Neo4j statements which get run in order.  To run Analysis Jobs, in your call to `cartography`, set the `--analysis-job-directory` parameter to the folder path of your jobs.  Although the order of statements within a single job is preserved, different jobs will run in a different order each time.
```
ecdavis: "different jobs will run in a different order each time" it's more accurate to say we don't guarantee the order in which jobs are executed. They probably will run in the same order each time on most platforms.

docs/writing-analysis-jobs.md
```diff
@@ -0,0 +1,126 @@
...
+    ```
+    MATCH 
+    (:IpRange{id: '0.0.0.0/0'})-[:MEMBER_OF_IP_RULE]->(:IpPermissionInbound)
+    -[:MEMBER_OF_EC2_SECURITY_GROUP]->(group:EC2SecurityGroup)
+    <-[:MEMBER_OF_EC2_SECURITY_GROUP]-(instance:EC2Instance)
+    
+    SET instance.exposed_internet = true, 
+        instance.exposed_internet_type = coalesce(instance.exposed_internet_type , []) + 'direct';
+    ```
+    In the `SET` clause we add `exposed_internet = True` to the instance.  We also add a field for `exposed_internet_type` to denote what type of internet exposure has occurred here.  You can read the [documentation for `coalesce`](https://neo4j.com/docs/cypher-manual/current/functions/scalar/#functions-coalesce), but in English this last part says "mark `exposed_internet_type` as `direct` unless this field was already previously defined on this instance.
```
ecdavis: > "mark `exposed_internet_type` as `direct` unless this field was already previously defined on this instance.

It's more like "add 'direct' to the list of ways this instance is exposed to the Internet".

PR #21, opened 2019-03-14
Add tags: true to Travis CI deploy section to ensure releases only happen on tagged commits.

ecdavis:


ecdavis:
https://docs.travis-ci.com/user/deployment#conditional-releases-with-on

PR #32, opened 2019-03-29
Docker compose configuration

dggc:
Hello, congrats for the cool project.

I thought spinning up cartography together with a Neo4j database in docker would be nice for people curious to use cartography without having to install Neo4j to their computers.

This is a simple bare-bones configuration for docker-compose.

Cheers!

PR #36, opened 2019-04-02
Dockerized

nishils:
I went ahead and dockerized the app. 

- We ran into some issues with reconnecting with the database based on out of order container startups. Those got fixed with some retries

- We also ran into issues when using SAML-based auth to our AWS environment. This helps to take care of that issue by passing the AWS user secrets at runtime and then saves it into a file.

- We have to experiment with this on an EC2 machine that has the security audit role and see if we can query based on instance metadata.

- The docker version also builds the master branch from scratch so works well for dev use cases.

Looking for feedback based on this.


nishils:
@achantavy I think I resolved most of the issues that you had with the other docker PR. Just want to put this one on your radar as it doesn't involve other dependencies.

achantavy:
@nishils Sorry for the delay on this, need a bit more time.

nishils:
1. thanks

2. Plan on deploying to our prod infrastructure so my goal was to cover both. I haven't gotten as far as deploying to an EC2 instance and have it query the credentials using the instance metadata, but this injection allows us to deploy to prod and use it locally. I'd be curious to know how this break your use case. 

3. I've been spinning these containers up and down pretty regularly and it works well. Anyways, retry logic is good to have for database connections even outside of docker use cases. Obviously, infinite retry is bad but some amount of retries can be helpful in the instance that the database has started but not yet accepting connections for whatever reason.

4. I can copy bits of the readme if you'd like. I'm just not a fan of wait-for-it.sh as this is not a needed dependency in most docker use cases. #32 also does not meet our needs because our aws credentials are only stored in memory as environment variables and never in an aws config file.

ecdavis:
Closing this as we are not accepting deployment/containerization configuration changes. See #159 for further discussion on this topic.

PR #38, opened 2019-04-02
Fix #34 - handle RDS instances without subnet groups

### LINKED ISSUE #34: Handle RDS instances without subnet groups
I'm getting an exception during data collection when a RDS instance doesn't have an attached `DBSubnetGroup`. While this is unlikely, it's possible and the error condition should probably be handled by catching a `KeyError` [here](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/rds.py#L131).

achantavy:
Also added requirement note in readme that we need python 3

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ You can learn more about the story behind Cartography in our [presentation at BS
 
 ## Installation
 
-Time to set up the server that will run Cartography.  Cartography _should_ work on both Linux and Windows servers, but bear in mind we've only tested it in Linux so far.  Cartography requires Python 3.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ You can learn more about the story behind Cartography in our [presentation at BS
 
 ## Installation
 
-Time to set up the server that will run Cartography.  Cartography _should_ work on both Linux and Windows servers, but bear in mind we've only tested it in Linux so far.
+Time to set up the server that will run Cartography.  Cartography _should_ work on both Linux and Windows servers, but bear in mind we've only tested it in Linux so far.  Cartography requires Python 3.
```
tayasteere: Specify which rev of Python 3? The Neo4J driver requires at least 3.4, might want to check the other requirements.

PR #48, opened 2019-04-08
Cross AWS VPC Peering linking

sachafaust:
Add VPC information along with associated blocks and cross-account peering


sachafaust:
@achantavy @ecdavis Please review.

cartography/intel/aws/ec2.py
```diff
@@ -1,6 +1,7 @@
 import botocore.config
 import logging
 import time
+import json
```
achantavy: Still used?  I might have missed it but I can't find any mention of it in the file.

cartography/intel/aws/ec2.py
```diff
@@ -249,6 +254,7 @@ def load_ec2_security_groupinfo(session, data, region, current_aws_account_id, a
             GroupId=group_id,
             GroupName=group.get("GroupName", ""),
             Description=group.get("Description", ""),
+            VpcId=group.get("VpcId", ""),
```
achantavy: Recommend replacing `""` with `None` if the member does not exist.  In my opinion it makes more sense if the property does not exist for it to not be on the node at all than as an empty string.

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +612,259 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
         data = get_loadbalancer_data(boto3_session, region)
         load_load_balancers(session, data, region, current_aws_account_id, aws_update_tag)
     cleanup_load_balancers(session, common_job_parameters)
+
+
+def get_ec2_vpcs(session):
+    client = session.client('ec2', config=_get_botocore_config())
+    # paginator not supported by boto
```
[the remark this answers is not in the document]
sachafaust: tried it and it throws error saying pagination is not supported

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +612,259 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
...
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}"""
+
+    for vpc in data['Vpcs']:
+        vpc_id = vpc["VpcId"]  # fail if not present
+
+        session.run(
+            ingest_vpc,
+            VpcId=vpc_id,
+            InstanceTenancy=vpc.get("InstanceTenancy", ""),
```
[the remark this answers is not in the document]
sachafaust: This may add complexity for queries as the attribute will not be present on the node/edge and requires the use of EXISTS(attrib) to avoid query to break. "" was used across AWS modules elsewhere

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +612,259 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
...
+    logger.debug("Syncing EC2 Vpc in account '%s'.", current_aws_account_id)
+    data = get_ec2_vpcs(boto3_session)
+    load_ec2_vpcs(session, data, current_aws_account_id, aws_update_tag)
+
+    cleanup_ec2_vpcs(session, common_job_parameters)
+
+
+def get_ec2_vpc_peering(session):
+    client = session.client('ec2', config=_get_botocore_config())
+    # paginator not supported by boto
```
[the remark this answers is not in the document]
sachafaust: Same as above. tried it and throws error saying pagination not supported

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +612,259 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
...
+    # }
+
+    # We assume the accept data is already in the graph since we run after all AWS account in scope
+    # We don't assume the receiver data is in the graph as it can be a foreign AWS account
+    # IPV6 peering is not supported, we default to AWSIpv4CidrBlock
+    ingest_peering = """
+    MATCH (accepter_block:AWSIpv4CidrBlock{id: {AccepterVpcId} + '|' + {AccepterCidrBlock}})
+    WITH accepter_block
+    MERGE (requestor_account:AWSAccount{id: {RequesterOwnerId}})
+    ON CREATE SET requestor_account.firstseen = timestamp(), requestor_account.foreign = true
```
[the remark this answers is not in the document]
sachafaust: We don't get the name from the API, only the ID

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +612,259 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
...
+    SET requestor_account.lastupdated = {aws_update_tag}
+    WITH accepter_block, requestor_account
+    MERGE (requestor_vpc:AWSVpc{id: {RequestorVpcId}})
+    ON CREATE SET requestor_vpc.firstseen = timestamp()
```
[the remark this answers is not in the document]
sachafaust: fixed

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +612,259 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
...
+                                    aws_update_tag=aws_update_tag)
+
+        ingest_cidr_association_set(session,
+                                    vpc_id=vpc_id,
+                                    type="ipv6",
+                                    data=vpc.get("Ipv6CidrBlockAssociationSet", []),
+                                    aws_update_tag=aws_update_tag)
+
+
+def ingest_cidr_association_set(session, vpc_id, type, data, aws_update_tag):
```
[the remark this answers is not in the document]
sachafaust: fine either way. This trick was used elsewhere in EC2 ingestion (load_ec2_security_group_rule). Also avoids having 2 code block that needs to be maintained over time.

cartography/intel/aws/ec2.py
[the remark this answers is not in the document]
sachafaust: Agreed on None

cartography/intel/aws/ec2.py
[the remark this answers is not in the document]
sachafaust: for UNWIND, I use to have that concern but seems neo4j address the past issue. We use UNWIND extensively in our internal Ubuntu CVE mapper without problems.

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +611,259 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
         data = get_loadbalancer_data(boto3_session, region)
         load_load_balancers(session, data, region, current_aws_account_id, aws_update_tag)
     cleanup_load_balancers(session, common_job_parameters)
+
+
+def get_ec2_vpcs(session):
```
ecdavis: Please reorganize these methods to be consistent with the rest of the file -- all get_ methods in one block, then all load_ methods, then all cleanup_ methods, then all sync_ methods.

cartography/intel/aws/ec2.py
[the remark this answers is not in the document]
ecdavis: I like Alex's feedback to avoid string replacement like this. IMO it makes the code much harder to reason about. That being said, I also agree that a single query is preferable as it will help us avoid being bitten in future when maintaining this code. I think we can achieve this as follows:

1. Factor out data transformation into it's own function. Make the T in ETL explicit.
2. Pass the output of the transformation to this ingest function (should be renamed load_ to be consistent).
3. To solve the label problem, run a subsequent query that labels nodes based on an attribute they have set.

Can discuss this in person if the above is not sufficiently clear.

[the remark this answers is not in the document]
sachafaust: Took feedback and made some changes. 
From evan, I did 1,2 but #3 I felt was adding complexity

Why not #3 
* added complexity where we needed to store a flag on the node and later on scan all new nodes for that flag and set the right type. This would have required 2 pass (ipv6 and ipv4)
* used graph to store a flag that is not needed over time. I felt we shouldn't use the graph in that way

cartography/intel/aws/__init__.py
```diff
@@ -29,10 +29,12 @@ def _sync_one_account(session, boto3_session, account_id, regions, sync_tag, com
     # EC2
     # TODO move this to EC2 module
     logger.info("Syncing EC2 for account '%s'.", account_id)
+    ec2.sync_vpc(session, boto3_session, account_id, sync_tag, common_job_parameters)
```
achantavy: I might not be searching correctly but I can't find this function defined in the EC2 module (https://github.com/lyft/cartography/blob/20f64d18f139d10f281e88016deb2c76438d547a/cartography/intel/aws/ec2.py), can you link it to me?

cartography/intel/aws/__init__.py
```diff
@@ -29,10 +29,12 @@ def _sync_one_account(session, boto3_session, account_id, regions, sync_tag, com
...
     ec2.sync_ec2_security_groupinfo(session, boto3_session, regions, account_id, sync_tag, common_job_parameters)
     ec2.sync_ec2_instances(session, boto3_session, regions, account_id, sync_tag, common_job_parameters)
     ec2.sync_ec2_auto_scaling_groups(session, boto3_session, regions, account_id, sync_tag, common_job_parameters)
     ec2.sync_load_balancers(session, boto3_session, regions, account_id, sync_tag, common_job_parameters)
+    ec2.sync_vpc_peering(session, boto3_session, sync_tag, common_job_parameters)
```
achantavy: Same comment as ec2.sync_vpc; I don't see where this function is defined

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +737,129 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
...
+        SET r.lastupdated = {aws_update_tag}"""
+
+    BLOCK_CIDR = "CidrBlock"
+    STATE_NAME = "CidrBlockState"
+
+    # base label type. We add the AWS ipv4 or 6 depending on block type
+    BLOCK_TYPE = "AWSCidrBlock"
+    
+    if block_type == "ipv6":
+        BLOCK_CIDR = "Ipv6" + BLOCK_CIDR
```
achantavy: If in the future we need to string search for `Ipv6CidrBlock`, we wouldn't be able to easily find it in the code.  To address this we can add a comment like `# :Ipv6Cidrblock defined here` or something similar.

docs/schema.md
```diff
@@ -270,7 +299,32 @@ Representation of an AWS [IAM Role](https://docs.aws.amazon.com/IAM/latest/APIRe
     (AWSAccount)-[AWS_ROLE]->(AWSRole)
     ```
 
+## AWSVpc
+Representation of an AWS VPC configuration
```
achantavy: Adding the link to the datatype: 
`Representation of an AWS [VPC configuration](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Vpc.html)`

docs/schema.md
```diff
@@ -99,6 +104,30 @@ Representation of an AWS Account.
 	(AWSAccount)-[AWS_ROLE]->(AWSRole)
 	```
 
+## AWSCidrBlock
+### AWSIpv4CidrBlock
+### AWSIpv6CidrBlock
+Representation of an AWS CidrBlock used in VPC configuration. The `AWSCidrBlock` defines the base label
```
achantavy: Adding link: 

`Representation of an [AWS CidrBlock used in VPC configuration](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_VpcCidrBlockAssociation.html).`

cartography/intel/aws/ec2.py
```diff
@@ -550,6 +567,112 @@ def load_load_balancer_listeners(session, load_balancer_id, listener_data, aws_u
...
+    #             "AccepterVpcInfo": {
+    #                 "OwnerId": "444455556666",
+    #                 "VpcId": "vpc-33cc44dd"
+    #             }
+    #         }
+    #     ]
+    # }
+
+    # We assume the accept data is already in the graph since we run after all AWS account in scope
+    # We don't assume the receiver data is in the graph as it can be a foreign AWS account
```
achantavy: Nit: do you mean 'requestor' instead of 'receiver'?

cartography/intel/aws/ec2.py
sachafaust: The new nodes are discoverable with str search. The difficulty with the STS_ASSUME role was that the full name of the relationship was a string concat of STS_ASSUME +#Access# 

BLOCK_TYPE = "AWSCidrBlock"
BLOCK_TYPE = BLOCK_TYPE + ":AWSIpv6CidrBlock"
BLOCK_TYPE = BLOCK_TYPE + ":AWSIpv4CidrBlock"

cartography/intel/aws/ec2.py
achantavy: Please remove, not used and it fails flake8.

```
(venv) ➜  cartography git:(vpc_peering) ✗ make test
flake8
./cartography/intel/aws/ec2.py:4: [F401] 'json' imported but unused
./cartography/intel/aws/ec2.py:907: [W391] blank line at end of file
make: *** [test_lint] Error 1
```

cartography/intel/aws/ec2.py
```diff
@@ -606,3 +890,18 @@ def sync_load_balancers(session, boto3_session, regions, current_aws_account_id,
...
+    load_ec2_vpcs(session, data, current_aws_account_id, aws_update_tag)
+    cleanup_ec2_vpcs(session, common_job_parameters)
+
+
+def sync_vpc_peering(session, boto3_session, current_aws_account_id, aws_update_tag, common_job_parameters):
+    logger.debug("Syncing EC2 VPC peering in account '%s'.", current_aws_account_id)
+    data = get_ec2_vpc_peering(boto3_session)
+    load_ec2_vpc_peering(session, data, aws_update_tag)
+    cleanup_ec2_vpc_peering(session, common_job_parameters)
+
```
achantavy: Please remove blank line; it fails flake8:

```
(venv) ➜  cartography git:(vpc_peering) ✗ make test
flake8
./cartography/intel/aws/ec2.py:4: [F401] 'json' imported but unused
./cartography/intel/aws/ec2.py:907: [W391] blank line at end of file
make: *** [test_lint] Error 1
```

PR #53, opened 2019-04-12
Add IAM role policy retrieval, parsing, and relationship creation to AWS intel module.

ecdavis:
Currently cartography will parse IAM group policies and create `(:AWSGroup)-[:STS_ASSUMEROLE_ALLOW]->(:AWSRole)` relationships, this change adds retrieval and parsing of IAM role policies and the creation of `(:AWSRole)-[:STS_ASSUMEROLE_ALLOW]->(:AWSRole)` relationships to enable exploration of transitive role assumption.

This adds a significant amount of time to the IAM sync depending on the number of IAM roles in the target accounts.


cartography/intel/aws/iam.py
```diff
@@ -259,6 +273,27 @@ def load_group_memberships(session, group_memberships, aws_update_tag):
...
+                if isinstance(role_arns, str):
+                    role_arns = [role_arns]
+
+                ret.extend(role_arns)
+    return ret
+
+
 def load_group_policies(session, group_policies, aws_update_tag):
     ingest_policies_assume_role = """
     MATCH (group:AWSGroup{name: {GroupName}})
```
sachafaust: is that safe in multi AWS setup?

cartography/intel/aws/iam.py
```diff
@@ -274,30 +309,41 @@ def load_group_policies(session, group_policies, aws_update_tag):
...
+                    RoleArn=role_arn,
+                    aws_update_tag=aws_update_tag
+                )
+
+
+def load_role_policies(session, role_policies, aws_update_tag):
+    ingest_policies_assume_role = """
+    MATCH (assumer:AWSRole{name: {RoleName}})
+    WITH assumer
+    MERGE (role:AWSRole{arn: {RoleArn}})
```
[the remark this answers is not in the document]
ecdavis: I'll address that in a future PR. Currently this is just duplicating what we do for groups.

cartography/intel/aws/iam.py
ecdavis: As above, I'll address this in a future PR. It's almost certainly not safe in a multi-account situation, we should be using ARNs.

sachafaust: since we are touching this now. Would it be better to fix that multi aws account bug that slipped before?

cartography/intel/aws/iam.py
```diff
@@ -259,6 +273,27 @@ def load_group_memberships(session, group_memberships, aws_update_tag):
             )
 
 
+def _find_roles_assumable_in_policy(policy_data):
+    ret = []
+    for statement in policy_data["PolicyDocument"]["Statement"]:
+        action = statement.get('Action')
+        if not action:
+            continue
```
ecdavis: Bug here. Action may be a list containing multiple actions.

ecdavis: Added a comment reflecting this.

Issue #55, opened 2019-04-15
cartographer only syncs VPCs in one region


kadrach:
AWS VPCs are regional. cartography should scan all accounts across all regions, at the moment it's missing most of them, and with VPC peering support, it's missing cross-region peers :)

sachafaust:
Tks for bringing this up. It's being fixed before we move this into official release.

PR #57, opened 2019-04-20
Fix bug #55 - VPC multi region support

sachafaust:
Address issue #55 and add support for multi-region VPC and peering connectivity


cartography/intel/aws/ec2.py
```diff
@@ -737,6 +737,7 @@ def load_ec2_vpcs(session, data, current_aws_account_id, aws_update_tag):
     new_vpc.is_default = {IsDefault},
     new_vpc.primary_cidr_block = {PrimaryCIDRBlock},
     new_vpc.dhcp_options_id = {DhcpOptionsId},
+    new_vpc.region = {Region},
```
achantavy: Non-blocking question: `requestor_vpc` doesn't have a `region` (probably because these can be "foreign".  Are we able to infer the region of these somehow? 

Just wanted to point out that some VPC nodes may have the `region` field defined and others might not, so it might be helpful to point this out in our docs.

sachafaust: correct. only foreign will have no region

PR #62, opened 2019-04-30
Fix bug in IAM role policy sync caused by AWSRole nodes created without name attributes.

ecdavis:
Roles without names can be created during both group and role policy syncs right now.

Fixes bugs introduced in #53.


ecdavis:
Tested internally and this fixed the issue we were seeing.

PR #68, opened 2019-05-07
Add ability to auth to Google Compute Platform and ingest GCP Organization objects

achantavy:
#worksonmybox 🙂

### DIFF
--- a/cartography/intel/gcp/__init__.py
+++ b/cartography/intel/gcp/__init__.py
@@ -1,4 +1,4 @@
-from oauth2client.client import GoogleCredentials
+from oauth2client.client import GoogleCredentials, ApplicationDefaultCredentialsError
 import logging
 
 from cartography.intel.gcp import crm
@@ -17,7 +17,7 @@ def start_gcp_ingestion(session, config):
         # See https://oauth2client.readthedocs.io/en/latest/source/
         #             oauth2client.client.html#oauth2client.client.OAuth2Credentials
         credentials = GoogleCredentials.get_application_default()
-    except Exception as e:
+    except ApplicationDefaultCredentialsError as e:
         logger.debug("Error occurred calling GoogleCredentials.get_application_default().", exc_info=True)
         logger.error(
             (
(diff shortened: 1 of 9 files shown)


cartography/intel/gcp/crm.py
```diff
@@ -0,0 +1,47 @@
...
+
+from cartography.util import run_cleanup_job
+
+logger = logging.getLogger(__name__)
+
+
+def get_gcp_organizations(credentials):
+    # See https://github.com/googleapis/google-api-python-client/issues/299#issuecomment-268915510 and related issues
+    # on why we set cache_discovery=False to suppress scary-looking warnings.
+    crm = googleapiclient.discovery.build('cloudresourcemanager', 'v1', credentials=credentials, cache_discovery=False)
```
ecdavis: Separate service creation from use.

cartography/intel/gcp/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    crm.sync_organizations(session, credentials, config.update_tag, common_job_parameters)
+
+    """
+    # TODO - add internet exposure
+    run_analysis_job(
+        'gcp_compute_asset_exposure.json',
+        session,
+        common_job_parameters
+    )
+    """
```
ecdavis: Either implement or remove from this PR (applies to the TODO in the imports block as well).

cartography/intel/gcp/crm.py
achantavy: Think I should move the `discovery.build()` call to `sync_organizations()` or to `start_gcp_ingestion()`?

Additional background: we will need to make additional calls to `discovery.build()` to create other `Resource` objects to pull other objects that don't fall under the Compute Resource Manager (e.g. Compute).

achantavy: I'm kinda leaning toward `sync_organizations` since `start_*_ingestion()` functions seem to only handle cred initialization

ecdavis: I would suggest factoring it into its own method, then calling it in `sync_organizations` and passing the service object to `get_gcp_organizations`.

ecdavis:
@achantavy please update PR title to a more complete, long-term relevant commit message.

achantavy:
Will do. 

> @achantavy please update PR title to a more complete, long-term relevant commit message.

Will do.  To avoid needing a `git commit amend --force` is it ok if I do this on Github form dialog before I merge?

PR #71, opened 2019-05-15
Ingest GCP folders, projects, and instances

achantavy:


cartography/intel/gcp/__init__.py
```diff
@@ -1,11 +1,35 @@
...
+import googleapiclient.discovery
 import logging
 
-from cartography.intel.gcp import crm
+from cartography.intel.gcp import crm, compute
 
 logger = logging.getLogger(__name__)
 
 
+def _get_crm_resource_v1(credentials):
```
sachafaust: We may want to be consistent on documenting parameters

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+    :param compute: The compute resource object created by googleapiclient.discovery.build()
+    :param max_results: Optional cap on number of results returned by this function. Default = None, which means no cap.
+    :return: List of a project's zone objects if Compute API is turned on, else None.
+    """
+    try:
+        req = compute.zones().list(project=project_id, maxResults=max_results)
+        res = req.execute()
+        return res['items']
+    except HttpError as e:
+        reason = json.loads(e.content)['error']['errors'][0]['reason']
```
sachafaust: are we sure the structure will always be present? We might spin up exception inside the handler

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+        if reason == 'accessNotConfigured':
+            logger.debug(
+                (
+                    "Google Compute Engine API access is not configured for project %s. "
+                    "Full details: %s"
+                ),
+                project_id,
+                e
+            )
+            return None
```
sachafaust: How does the user know the user doesn't have API access? should we bubble this up to the CLI vs returning None?

cartography/intel/gcp/__init__.py
```diff
@@ -1,11 +1,35 @@
...
+
+
+def _get_compute_resource(credentials):
+    return googleapiclient.discovery.build('compute', 'v1', credentials=credentials, cache_discovery=False)
+
+
+def _initialize_resources(credentials):
+    return {'crm_v1': _get_crm_resource_v1(credentials),
+            'crm_v2': _get_crm_resource_v2(credentials),
+            'compute': _get_compute_resource(credentials)}
```
ecdavis: Not a fan of this pattern -- strings and mutable data structures are two places where bugs tend to sneak in. If you think it's important to pass around a group of clients like this then I would suggest creating a class (maybe a namedtuple or something similar). That way you get to leverage linting and are more likely to catch issues with test values.

cartography/intel/gcp/compute.py
ecdavis: +1 -- does the library not provide a better interface for getting this information?

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+
+def sync_gcp_instances(session, resources, gcp_update_tag, common_job_parameters):
+    crm_v1 = resources['crm_v1']
+    compute = resources['compute']
+
+    # Note: `crm.sync_gcp_projects()` calls `crm.get_gcp_projects()` before `compute.sync_gcp_instances()` does.
+    # We could run into an issue where the return values are different here.
+    # Still, we do need project_ids in order to pull instance data.
+    projects = get_gcp_projects(crm_v1)
+    for project in projects:
```
ecdavis: How is this going to evolve when we add other GCP sync modules? I would suggest matching the AWS multi-account pattern unless there's a compelling reason otherwise. Ideally we call get_projects once and we avoid:
1. The issue you identified in the comment above.
2. The cross-dependency between compute and crm.

cartography/intel/gcp/compute.py
achantavy: This `logger.debug` call and the Google SDK's built-in telemetry logs a debug message to the CLI.  Google's message looks like this:
`WARNING:googleapiclient.http:Encountered 403 Forbidden with reason "accessNotConfigured"
INFO:googleapiclient.discovery:URL being requested: GET https://www.googleapis.com/compute/v1/projects/<project_name>/zones?alt=json`

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+            instance['zone_name'] = zone['name']
+            instances.append(instance)
+    return instances
+
+
+def load_gcp_instances(neo4j_session, data, gcp_update_tag):
+    query = """
+    MATCH (p:GCPProject{id:{ProjectId}})
+    MERGE (i:GCPInstance{id:{InstanceId}})
+    ON CREATE SET i.firstseen = timestamp()
```
sachafaust: should we add the instanceid attribute? It's a dupe of the id but it's clearer and avoid users to know what we base our index/id on.

cartography/intel/gcp/__init__.py
achantavy: Fair point, I can add these.

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+        for instance in zone_instances:
+            # Include project_id and zone_name on the `instance` dict so we can load it more easily
+            instance['project_id'] = project_id
+            instance['zone_name'] = zone['name']
+            instances.append(instance)
+    return instances
+
+
+def load_gcp_instances(neo4j_session, data, gcp_update_tag):
+    query = """
```
sachafaust: missing function doc

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+            InstanceId=instance['id'],
+            DisplayName=instance.get('name', None),
+            Hostname=instance.get('hostname', None),
+            ZoneName=instance['zone_name'],
+            gcp_update_tag=gcp_update_tag
+        )
+
+
+def cleanup_gcp_instances(session, common_job_parameters):
+    run_cleanup_job('gcp_compute_instance_cleanup.json', session, common_job_parameters)
```
sachafaust: missing function doc

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def sync_gcp_instances(session, resources, gcp_update_tag, common_job_parameters):
+    crm_v1 = resources['crm_v1']
```
sachafaust: missing function doc

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,118 @@
...
+    for instance in data:
+        neo4j_session.run(
+            query,
+            ProjectId=instance['project_id'],
+            InstanceId=instance['id'],
+            DisplayName=instance.get('name', None),
+            Hostname=instance.get('hostname', None),
+            ZoneName=instance['zone_name'],
+            gcp_update_tag=gcp_update_tag
+        )
```
ecdavis: Non-blocking: are there any pieces of data that apply to compute instances across cloud providers? Would be cool if we could define an abstract `:Instance` label with a limited but cross-cloud data set.

cartography/intel/gcp/compute.py
achantavy: True.  I can change to `res.get('items', [])`

cartography/intel/gcp/compute.py
sachafaust: probably do the same with GCPProject

cartography/intel/gcp/crm.py
```diff
@@ -45,14 +49,119 @@ def load_gcp_organizations(neo4j_session, data, gcp_update_tag):
         )
 
 
+def load_gcp_folders(neo4j_session, data, gcp_update_tag):
```
ecdavis: I'm going to be annoying and request that you separate data transformation from loading in this and the below function. That is:
```
load_gcp_folders(session, folders, gcp_update_tag)
```
should become:
```
transformed_data = transform_gcp_folders(folders)
load_gcp_folders(session, transformed, gcp_update_tag)
```
and the same for the function below.

cartography/intel/gcp/crm.py
```diff
@@ -45,14 +49,119 @@ def load_gcp_organizations(neo4j_session, data, gcp_update_tag):
...
+            ON CREATE SET parent.firstseen = timestamp()
+            """
+        query += """
+        MERGE (folder:GCPFolder{id:{FolderName}})
+        ON CREATE SET folder.firstseen = timestamp()
+        SET folder.displayname = {DisplayName},
+        folder.lifecyclestate = {LifecycleState},
+        folder.lastupdated = {gcp_update_tag}
+        WITH parent, folder
+        MERGE (folder)-[r:PARENT]->(parent)
```
sachafaust: should we do child->parent (reverse) or parent -> child (forward)? Most time people use forward for hierarchy

cartography/intel/gcp/crm.py
```diff
@@ -45,14 +49,119 @@ def load_gcp_organizations(neo4j_session, data, gcp_update_tag):
...
+            """
+            parentid = f"folders/{project['parent']['id']}"
+        query += """
+        MERGE (project:GCPProject{id:{ProjectId}})
+        ON CREATE SET project.firstseen = timestamp()
+        SET project.displayname = {DisplayName},
+        project.lifecyclestate = {LifecycleState},
+        project.lastupdated = {gcp_update_tag}
+        WITH parent, project
+        MERGE (project)-[r:PARENT]->(parent)
```
sachafaust: same above regarding backward/forward hierarchy

docs/schema.md
```diff
@@ -1104,4 +1104,98 @@ Representation of a GCP [Organization](https://cloud.google.com/resource-manager
 | lastupdated |  Timestamp of the last time the node was updated | 
 | id | The name of the GCP Organization, e.g. "organizations/1234" |
 | displayname | The "friendly name", e.g. "My Company"
-| lifecyclestate | The organization's current lifecycle state. Assigned by the server.  See the [official docs](https://cloud.google.com/resource-manager/reference/rest/v1/organizations#LifecycleState). 
\ No newline at end of file
+| lifecyclestate | The organization's current lifecycle state. Assigned by the server.  See the [official docs](https://cloud.google.com/resource-manager/reference/rest/v1/organizations#LifecycleState).
```
sachafaust: it's starting to get pretty big. Should we consider breaking the doc into core pieces?

cartography/intel/gcp/compute.py
ecdavis: From google SDK:

```python
    try:
      data = json.loads(self.content.decode('utf-8'))
      if isinstance(data, dict):
        reason = data['error']['reason']
      elif isinstance(data, list) and len(data) > 0:
        first_error = data[0]
        reason = first_error['error']['reason']
    except (ValueError, KeyError, TypeError):
        pass
```

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,170 @@
...
+from cartography.util import run_cleanup_job
+
+logger = logging.getLogger(__name__)
+
+
+def _get_error_reason(http_error):
+    """
+    Helper function to get an error reason out of the googeapiclient's HttpError object
+    This function copies the structure of
+    https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/errors.py#L46.
```
ecdavis: To prevent link rot, link to a SHA rather than to master.

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    try:
+        data = json.loads(http_error.content.decode('utf-8'))
+        if isinstance(data, dict):
+            first_error = data
+        elif isinstance(data, list) and len(data) > 0:
+            first_error = data[0]
+        reason = first_error['error']['errors'][0]['reason']
+    except (ValueError, KeyError, TypeError):
+        logger.error("Could not parse HttpError object, returning `cartography.ExceptionNotParsable` as reason."
```
ecdavis: As the code is currently written this error will be logged three (four if you count stdout) times with different messages each time. First here, second when the return value of this function is checked by `get_zones_in_project` (line 79 in this file as of this comment), third when the exception raised in `get_zones_in_project` is caught by `Sync.run` and then fourth when the program crashes due to the uncaught exception.

We need to think harder about how to handle this situation. Log-spam is the immediate issue but there's a larger question about exceptions and how they effect the execution of the sync in general which we have discussed before.

Leaving this comment to spark a broader discussion, but the immediate ask is to address the log-spam somehow. We can figure out a good way of approaching that together.

cartography/intel/gcp/crm.py
```diff
@@ -45,14 +69,158 @@ def load_gcp_organizations(neo4j_session, data, gcp_update_tag):
...
+            parentid = f"folders/{project['parent']['id']}"
+        query += """
+        MERGE (project:GCPProject{id:{ProjectId}})
+        ON CREATE SET project.firstseen = timestamp()
+        SET project.projectid = {ProjectId},
+        project.displayname = {DisplayName},
+        project.lifecyclestate = {LifecycleState},
+        project.lastupdated = {gcp_update_tag}
+        WITH parent, project
+        MERGE (parent)-[r:RESOURCE]->(project)
```
ecdavis: I don't know if this is the right relationship to represent the folder/organizations contains project concept. That said, I don't have any good alternative to offer. Thoughts?

cartography/intel/gcp/crm.py
```diff
@@ -45,14 +69,158 @@ def load_gcp_organizations(neo4j_session, data, gcp_update_tag):
...
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    :return: Nothing
+    """
+    for project in data:
+        if project['parent']['type'] == "organization":
+            query = "MATCH (parent:GCPOrganization{id:{ParentId}})"
+            parentid = f"organizations/{project['parent']['id']}"
+        elif project['parent']['type'] == "folder":
+            query = """
+            MERGE (parent:GCPFolder{id:{ParentId}})
```
ecdavis: Do you think it's worth walking up the folder tree and creating a relationship from the organization to _all_ projects it contains, regardless of which folder they're in? I don't think this is something we should add in this PR, but we should think about it for the future IMO.

cartography/intel/gcp/__init__.py
```diff
@@ -1,12 +1,100 @@
...
+
+def _initialize_resources(credentials):
+    """
+    Create namedtuple of all resource objects necessary for GCP data gathering.
+    :param credentials: The GoogleCredentials object
+    :return: namedtuple of all resource objects
+    """
+    return Resources(crm_v1=_get_crm_resource_v1(credentials),
+                     crm_v2=_get_crm_resource_v2(credentials),
+                     compute=_get_compute_resource(credentials))
```
ecdavis: The style we use elsewhere is:
```python
    return Resources(
        crm_v1=_get_crm_resource_v1(credentials),
        crm_v2=_get_crm_resource_v2(credentials),
        compute=_get_compute_resource(credentials)
    )
```

cartography/intel/gcp/__init__.py
```diff
@@ -27,4 +115,12 @@ def start_gcp_ingestion(session, config):
             e
         )
         return
-    crm.sync_gcp_organizations(session, credentials, config.update_tag, common_job_parameters)
+    resources = _initialize_resources(credentials)
+
+    # If we don't have perms to pull Orgs or Folders from GCP, we will skip safely
```
ecdavis: Is this comment accurate? I can't see any code explicitly handling this case.

cartography/intel/gcp/__init__.py
```diff
@@ -27,4 +115,12 @@ def start_gcp_ingestion(session, config):
...
+    crm.sync_gcp_organizations(session, resources.crm_v1, config.update_tag, common_job_parameters)
+    crm.sync_gcp_folders(session, resources.crm_v2, config.update_tag, common_job_parameters)
+
+    projects = crm.get_gcp_projects(resources.crm_v1)
+
```
ecdavis: Please add a comment reflecting that we currently don't handle individual projects outside of an organization. We could also add an issue for this -- it may be a good first issue for someone to work on.

cartography/intel/gcp/crm.py
```diff
@@ -45,14 +69,158 @@ def load_gcp_organizations(neo4j_session, data, gcp_update_tag):
...
+            """
+        query += """
+        MERGE (folder:GCPFolder{id:{FolderName}})
+        ON CREATE SET folder.firstseen = timestamp()
+        SET folder.foldername = {FolderName},
+        folder.displayname = {DisplayName},
+        folder.lifecyclestate = {LifecycleState},
+        folder.lastupdated = {gcp_update_tag}
+        WITH parent, folder
+        MERGE (parent)-[r:RESOURCE]->(folder)
```
ecdavis: See comment below -- not sure if this is the right relationship for this concept.

cartography/intel/gcp/crm.py
achantavy: Had it originally as :PARENT since that's what GCP seems to call this link, but it seems that we have :RESOURCE elsewhere to describe parent-child relationships.  I'm open for changes, though I think one benefit of setting all of these parent-child relationships as [:RESOURCE] is that we could query for multiple hops like this:

```cypher
match p=(o:GCPOrganization)-[r:RESOURCE*..]->(g:GCPInstance) return p
```

which would return all instances under an organization regardless of recursion depth.  This only works if we have all these rels as :RESOURCE.

cartography/intel/gcp/crm.py
achantavy: I see your point, since if you're using the GCP API you don't necessarily need to care about a project being in a folder (well, for that point you don't really need to care about a project being under an organization tbh, you just need to know the projectId).  We can think about creating these shortcut relationships, though this is partially mitigated by having everything set as :RESOURCE as in my comment earlier.

cartography/intel/gcp/__init__.py
achantavy: Added code handling projects without a parent.

cartography/intel/gcp/__init__.py
achantavy: Added code handling this

cartography/intel/gcp/compute.py
achantavy: Got it, I like the general principle where we re-raise all exceptions unless we know how to actually handle them but I missed that here.  My latest push does this I think

cartography/intel/gcp/__init__.py
```diff
@@ -1,12 +1,102 @@
...
+    Handles graph sync for a single GCP project.
+    :param session: The Neo4j session
+    :param resources: namedtuple of the GCP resource objects
+    :param project_id: The project ID number to sync.  See  the `projectId` field in
+    https://cloud.google.com/resource-manager/reference/rest/v1/projects
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    :param common_job_parameters: Other parameters sent to Neo4j
+    :return: Nothing
+    """
+    compute.sync_gcp_instances(session, resources.compute, project_id, gcp_update_tag, common_job_parameters)
```
ecdavis: See AWS module: log an INFO-level message that compute is being synced for this GCP project.

https://github.com/lyft/cartography/blob/master/cartography/intel/aws/__init__.py#L32

cartography/intel/gcp/__init__.py
```diff
@@ -1,12 +1,102 @@
...
+    This is the returned data from `crm.get_gcp_projects()`.
+    See https://cloud.google.com/resource-manager/reference/rest/v1/projects.
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    :param common_job_parameters: Other parameters sent to Neo4j
+    :return: Nothing
+    """
+    crm.sync_gcp_projects(session, projects, gcp_update_tag, common_job_parameters)
+
+    for project in projects:
+        project_id = project['projectId']
```
ecdavis: See AWS module: log an INFO-level message that this GCP project is being synced.

https://github.com/lyft/cartography/blob/master/cartography/intel/aws/__init__.py#L59

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    try:
+        data = json.loads(http_error.content.decode('utf-8'))
+        if isinstance(data, dict):
+            reason = data['error']['errors'][0]['reason']
+        else:
+            reason = data[0]['error']['errors']['reason']
+    except Exception:
+        # Failed to parse the error. Let's fail the whole sync.
+        raise
```
ecdavis: Rather than re-raising, I suggest you return the empty string and allow `get_zones_in_project` to decide what to do (if the empty string is returned it will re-raise the HTTPError.

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,175 @@
...
+    :param http_error: The googleapi HttpError object
+    :return: The error reason as a string
+    """
+    try:
+        data = json.loads(http_error.content.decode('utf-8'))
+        if isinstance(data, dict):
+            reason = data['error']['errors'][0]['reason']
+        else:
+            reason = data[0]['error']['errors']['reason']
+    except Exception:
```
ecdavis: Stick to catching a specific set of exceptions. As written, this will catch OutOfMemory, KeyboardException, SystemExit, etc.

cartography/intel/gcp/compute.py
```diff
@@ -0,0 +1,175 @@
...
+            # Write some additional advice and then re-raise
+            logger.error(
+                (
+                    "Encountered a 403-Forbidden error trying to list zones on project %s - Please ensure that the "
+                    "identity that your GOOGLE_APPLICATION_CREDENTIALS point to has the securityReviewer role defined "
+                    "on this project."
+                ),
+                project_id
+            )
+            raise
```
ecdavis: Because we log an ERROR-level message for all uncaught exceptions in `Sync.run` this block will result in multiple log messages for the same error. I would suggest removing this branch entirely and letting the exception be bubble up from the else branch or, if you feel the extra context is necessary, create a new exception object, wrap the HTTPError, and raise the new exception.

PR #72, opened 2019-05-16
Add syntax tests for job data files.

ecdavis:


tests/integration/cartography/data/jobs/test_syntax.py
```diff
@@ -0,0 +1,27 @@
+import cartography.util
+import pytest
+import sys
+if sys.version_info >= (3, 7):
+    from importlib.resources import contents
+else:
+    from importlib_resources import contents
+
+
+def test_analysis_jobs_actually_execute(neo4j_session):
```
achantavy: Non-blocking: I read the function name as "test if job actually starts", when something like "test if job's cypher queries have correct syntax" is probably more complete.  I'd consider renaming it to indicate a syntax check, but this is non-blocking

PR #74, opened 2019-05-29
Use partial_uri as unique identifier on GCP instances

achantavy:
GCP uses the `name` field to uniquely identify resources, not the `id` field (which in the case of GCP instances returns a number assigned by the server that doesn't communicate much meaning to the user).  

As such, this PR makes it so that GCP instances are now uniquely identified by their [partial resource URIs](https://cloud.google.com/apis/design/resource_names#relative_resource_name).  Compute instance partial URIs have the form `projects/{project}/zones/{zone name}/instances/{instance name}`.

### DIFF
--- a/cartography/intel/gcp/compute.py
+++ b/cartography/intel/gcp/compute.py
@@ -91,11 +91,25 @@ def get_gcp_instances_in_project(project_id, compute):
         res = req.execute()
         zone_instances = res.get('items', [])
         for instance in zone_instances:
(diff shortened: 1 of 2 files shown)


ecdavis:
Looks good. This is a breaking change so we'll need to remember to bump the minor version for the next release. Also, nitpicking, please remove "Fix:" prefix from PR title -- this is more of a behavior change than a bug fix.

PR #76, opened 2019-05-30
Add GCP VPCs, subnets, network interfaces, NIC access configs

achantavy:


cartography/data/indexes.cypher
```diff
@@ -48,4 +48,5 @@ CREATE INDEX ON :AWSIpv6CidrBlock(id);
 CREATE INDEX ON :GCPOrganization(id);
 CREATE INDEX ON :GCPFolder(id);
 CREATE INDEX ON :GCPProject(id);
-CREATE INDEX ON :GCPInstance(id);
\ No newline at end of file
+CREATE INDEX ON :GCPInstance(id);
+CREATE INDEX ON :GCPVpc(id);
```
sachafaust: [block] missing GCPSubnet

cartography/intel/gcp/compute.py
```diff
@@ -112,6 +136,62 @@ def transform_gcp_instance(instance, project_id, zone_name):
...
 
+def transform_gcp_vpcs(vpc_res):
+    """
+    Transform the VPC response object for Neo4j ingestion
+    :param vpc_res: The return data
+    :return: List of VPCs ready for ingestion to Neo4j
+    """
+    vpc_list = []
+    prefix = vpc_res['id']
+    projectid = prefix.split('/')[1]
```
sachafaust: adding examples would help understand/maintain the code

sachafaust: [refactor] is it similar to AWS ARN? if so, we could benefit from common method to parse them and get dictionary back. We redo the same logic in other functions also

cartography/intel/gcp/compute.py
```diff
@@ -147,6 +227,171 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
...
+    """
+    Ingest VPCs to Neo4j
+    :param neo4j_session: The Neo4j session object
+    :param vpcs: List of VPCs to ingest
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    :return: Nothing
+    """
+    query = """
+    MATCH(p:GCPProject{id:{ProjectId}})
+    MERGE(vpc:GCPVpc{id:{PartialUri}})
```
sachafaust: [question] what guaranty this to be unique across GCP?

cartography/intel/gcp/compute.py
```diff
@@ -147,6 +227,171 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
...
+    ON CREATE SET vpc.firstseen = timestamp()
+    SET vpc.partial_uri = {PartialUri},
```
sachafaust: If we base the ID on PartialUri, we should move it to ON CREATE SET as it will never change and should be added once.

cartography/intel/gcp/compute.py
```diff
@@ -147,6 +227,171 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
...
+    vpc.auto_create_subnetworks = {AutoCreateSubnetworks},
+    vpc.routing_config_routing_mode = {RoutingMode},
+    vpc.description = {Description},
+    vpc.lastupdated = {gcp_update_tag}
+
+    MERGE (p)-[r:RESOURCE]->(vpc)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {gcp_update_tag}
+    """
+    for vpc in vpcs:
```
sachafaust: Could we use UNWIND here instead and optimize network calls?

cartography/intel/gcp/compute.py
```diff
@@ -147,6 +227,171 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
...
+def load_gcp_subnets(neo4j_session, subnets, gcp_update_tag):
+    """
+    Ingest GCP subnet data to Neo4j
+    :param neo4j_session: The Neo4j session
+    :param subnets: List of the subnets
+    :param gcp_update_tag: The timestamp to set these Neo4j nodes with
+    :return: Nothing
+    """
+    query = """
+    MERGE(vpc:GCPVpc{self_link:{VpcSelfLink}})
```
sachafaust: [non standard] we should always MERGE on the id attribute

cartography/intel/gcp/compute.py
```diff
@@ -188,4 +493,10 @@ def sync(session, compute, project_id, gcp_update_tag, common_job_parameters):
     :return: Nothing
     """
     logger.info("Syncing Compute objects for project %s.", project_id)
-    sync_gcp_instances(session, compute, project_id, gcp_update_tag, common_job_parameters)
+    zones = get_zones_in_project(project_id, compute)
+    # Only pull additional assets for this project if the Compute API is enabled
+    if zones:
+        regions = _zones_to_regions(zones)
+        sync_gcp_vpcs(session, compute, project_id, gcp_update_tag, common_job_parameters)
```
ecdavis: Is it possible to have VPCs without zones?

cartography/intel/gcp/compute.py
achantavy: ProjectIDs are unique across GCP : 

"The project number and project ID are unique across Google Cloud Platform. If another user owns a project ID for their project, you won't be able to use the same project ID." -- https://cloud.google.com/resource-manager/docs/creating-managing-projects

cartography/intel/gcp/compute.py
achantavy: Yes.  A VPC is a [global resource](https://cloud.google.com/compute/docs/regions-zones/global-regional-zonal-resources#globalresources) that doesn't depend on a zone, and the URI for a VPC doesn't have a zone in it.

This check is more to ensure that the Compute API is enabled for the given project.  This code currently makes the assumption that if there are no zones returned, then the Compute API is disabled.

That said, good catch - I actually don't know if it is possible for the Compute API to be enabled **and** 0 zones are returned.  I'll make an adjustment to account for this.

cartography/intel/gcp/compute.py
achantavy: Yeah, a URI/partial URI in GCP is the same concept of an AWS ARN. I see your point on creating a generic function to handle this - and I'm open to more feedback -  but I would rather wait until there's another place in the code where we need to parse VPC URIs before refactoring.  I will add a comment to make this more readable though.

achantavy: I'll add a comment on the form of the prefix and what we're parsing out.

cartography/intel/gcp/compute.py
ecdavis: From chatting in person, this seems more explicit to me:
```
if zones is None:
    return  # compute is not enabled
else:
    regions = _zones_to_regions(zones)
    ...
```

tests/integration/cartography/data/jobs/test_syntax.py
```diff
@@ -7,6 +7,7 @@
     from importlib_resources import contents
 
 
+@pytest.mark.skip(reason="I have Neo4j 3.3.9")
```
ecdavis: Remove :)

cartography/intel/gcp/compute.py
```diff
@@ -147,6 +227,171 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
...
+    ON CREATE SET nic.firstseen = timestamp()
+    SET nic.private_ip = {NetworkIP},
+    nic.name = {NicName},
+    nic.lastupdated = {gcp_update_tag}
+
+    MERGE (i)-[r:NETWORK_INTERFACE]->(nic)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {gcp_update_tag}
+
+    MERGE (subnet:GCPSubnet{self_link:{SubnetSelfLink}})
```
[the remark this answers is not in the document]
sachafaust: seemed to have been missed

cartography/intel/gcp/compute.py
```diff
@@ -122,14 +228,15 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
     :return: Nothing
     """
     query = """
-    MATCH (p:GCPProject{id:{ProjectId}})
+    MERGE (p:GCPProject{id:{ProjectId}})
     MERGE (i:Instance:GCPInstance{id:{PartialUri}})
     ON CREATE SET i.firstseen = timestamp()
```
sachafaust: partial_uri should be in the ON CREATE SET since it's the id map and will never change.

cartography/intel/gcp/compute.py
```diff
@@ -147,6 +254,178 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
...
+def load_gcp_vpcs(neo4j_session, vpcs, gcp_update_tag):
+    """
+    Ingest VPCs to Neo4j
+    :param neo4j_session: The Neo4j session object
+    :param vpcs: List of VPCs to ingest
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    :return: Nothing
+    """
+    query = """
+    MERGE(p:GCPProject{id:{ProjectId}})
```
sachafaust: [block] missing ON CREATE SET and firstseen + lastupdated

cartography/intel/gcp/compute.py
```diff
@@ -147,6 +254,178 @@ def load_gcp_instances(neo4j_session, data, gcp_update_tag):
...
+    MERGE(vpc:GCPVpc{id:{PartialUri}})
+    ON CREATE SET vpc.firstseen = timestamp(),
+    vpc.partial_uri = {PartialUri}
```
sachafaust: move partial_uri to ON CREATE SET

achantavy: It currently already is, it's just on the second line.

cartography/intel/gcp/compute.py
achantavy: Fixed

cartography/intel/gcp/compute.py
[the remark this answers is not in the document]
achantavy: Joining on the id field now and I got the standard attributes added.

cartography/intel/gcp/compute.py
achantavy: Fixed

PR #77, opened 2019-05-31
Added extra properties to dynamodb nodes

thomashli:


cartography/intel/aws/dynamodb.py
```diff
@@ -10,32 +10,68 @@ def get_dynamodb_tables(session, region):
...
         session.run(
             ingest_table,
-            Arn=arn,
-            TableName=table_name,
+            Arn=table['Table']['TableArn'],
+            TableName=table['Table']['TableName'],
             Region=region,
             AWS_ACCOUNT_ID=current_aws_account_id,
-            aws_update_tag=aws_update_tag
+            aws_update_tag=aws_update_tag,
```
ecdavis: Our convention is to have the `AWS_ACCOUNT_ID` and `aws_update_tag` parameters at the end of the function invocation.

cartography/intel/aws/dynamodb.py
```diff
@@ -10,32 +10,68 @@ def get_dynamodb_tables(session, region):
...
+    """
+
+    for gsi in table['Table']['GlobalSecondaryIndexes']:
+        session.run(
+            ingest_gsi,
+            Arn=gsi['IndexArn'],
+            GSIName=gsi['IndexName'],
+            Region=region,
+            AWS_ACCOUNT_ID=current_aws_account_id,
+            aws_update_tag=aws_update_tag,
```
ecdavis: See comment above.

cartography/intel/aws/dynamodb.py
```diff
@@ -10,32 +10,68 @@ def get_dynamodb_tables(session, region):
...
+            Rows=table['Table']['ItemCount'],
+            Size=table['Table']['TableSizeBytes'],
+            ProvisionedThroughputReadCapacityUnits=table['Table']['ProvisionedThroughput']['ReadCapacityUnits'],
+            ProvisionedThroughputWriteCapacityUnits=table['Table']['ProvisionedThroughput']['WriteCapacityUnits']
+        )
+        load_gsi(session, region, current_aws_account_id, aws_update_tag, table)
+
+
+def load_gsi(session, region, current_aws_account_id, aws_update_tag, table):
```
ecdavis: Move `table` to the second parameter, I think.

tests/data/aws/dynamodb.py
```diff
@@ -0,0 +1,254 @@
+
+
+LIST_DYNAMODB_TABLES = {
+    "Tables": [
+        {
+            "Table": {
+                "TableArn": "arn:aws:dynamodb::000000000000:table/example-table",
```
ecdavis: The ARNs in this file should include a region for the sake of completeness.

tests/data/aws/dynamodb.py
```diff
@@ -0,0 +1,254 @@
...
+                },
+                "TableSizeBytes": 100000000,
+                "TableName": "example-table",
+                "TableStatus": "ACTIVE",
+                "StreamSpecification": {
+                    "StreamViewType": "SAMPLE_STREAM_VIEW_TYPE",
+                    "StreamEnabled": True
+                },
+                "TableId": "00000000-0000-0000-0000-000000000000",
+                "LatestStreamLabel": "0000-00-00000:00:00.000",
```
ecdavis: Does the API return this as a string or a datetime object?

cartography/intel/aws/dynamodb.py
```diff
@@ -10,32 +10,68 @@ def get_dynamodb_tables(session, region):
...
+    ingest_gsi = """
+    MERGE (gsi:DynamoDBTableGSI{id: {Arn}})
```
ecdavis: I think `DynamoDBIndex` is a clearer name for this node type. What do you think?

Since we're adding a new node type we should also add an index here: https://github.com/lyft/cartography/blob/master/cartography/data/indexes.cypher

achantavy: And the schema docs! https://github.com/lyft/cartography/blob/master/docs/schema/aws.md#dynamodbtable

thomashli: I found out that tables can contain two types of indexes, Local Secondary Indexes and Global Secondary Indexes. I think distinguishing the types might be useful if some team wanted to obtain information about a table's Local Secondary Indexes in the future. I'll make it `DynamoDBGlobalSecondaryIndex` for readability.

achantavy: Need to add a cleanup job to [this](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/cleanup/aws_import_dynamodb_tables_cleanup.json) to delete nodes and relationships that are out of date.

cartography/intel/aws/dynamodb.py
```diff
@@ -10,32 +10,68 @@ def get_dynamodb_tables(session, region):
...
+    ON CREATE SET gsi.firstseen = timestamp(), gsi.arn = {Arn}, gsi.name = {GSIName},
+    gsi.region = {Region}
+    SET gsi.lastupdated = {aws_update_tag},
+    gsi.provisioned_throughput_read_capacity_units = {ProvisionedThroughputReadCapacityUnits},
+    gsi.provisioned_throughput_write_capacity_units = {ProvisionedThroughputWriteCapacityUnits}
+    WITH gsi
+    MATCH (table:DynamoDBTable{arn: {TableArn}})
+    MERGE (table)-[r:GSI]->(gsi)
```
achantavy: I think we should rename this to `:GLOBAL_SECONDARY_INDEX` so it's easier to understand. Open to go either way.  Just so I understand, the relationship is that this table __has__ an index, right?

cartography/intel/aws/dynamodb.py
```diff
@@ -10,32 +10,68 @@ def get_dynamodb_tables(session, region):
...
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
+    """
+
+    for gsi in table['Table']['GlobalSecondaryIndexes']:
```
achantavy: Are all tables guaranteed to have GSIs?

As it is right now the code will crash with a KeyError if 'GlobalSecondaryIndexes' is not present.  If we are expecting all the tables to have GSIs then this behavior is ok but if not then we should do something like 

`for gsi in table['table'].get('GlobalSecondaryIndexes', [])`

cartography/intel/aws/dynamodb.py
thomashli: I like `GLOBAL_SECONDARY_INDEX` as a name so I'll use that. The relationship is that the index belongs to that table. I could also change the relationship to `CONTAINS`.

cartography/intel/aws/dynamodb.py
thomashli: Missed that one. Fixed.

tests/data/aws/dynamodb.py
thomashli: This should be a string from what I read in the boto3 documentation, but I changed some other parameters to datetime objects which I incorrectly made strings.

achantavy: I think Evan's pointing out that our versions of Neo4j (3.2 and 3.3) don't support Python datetime objects so you need to run `str()` any datetimes before loading them in to Neo4j.  It doesn't look like your `load_*` functions ingest any datetime objects as far as I can tell though.

tests/integration/cartography/intel/aws/test_dynamodb.py
```diff
@@ -0,0 +1,18 @@
+import cartography.intel.aws.dynamodb
+import tests.data.aws.dynamodb
+
+TEST_ACCOUNT_ID = '000000000000'
+TEST_REGION = 'us-east-1'
+TEST_UPDATE_TAG = 123456789
+
+
+def test_load_dynamodb(neo4j_session):
```
achantavy: Non-blocking: For extra-extra confidence that your nodes are getting loaded with the members that you expect, you _could_ add an integration test like this: https://github.com/lyft/cartography/pull/76/files#diff-a58e210adeb9fd885c695de8a51f486eR32

I'll let @ecdavis decide if this is necessary; I'm good either way, but it could be a good thing to have.

thomashli: Wrote some basic tests in the format. Should be good!

PR #79, opened 2019-06-04
Add AWSPrincipal label to AWSRole nodes.

ecdavis:
An IAM Role is a type of Principal (see https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html ), we add the AWSPrincipal label to AWSUser nodes, so we should be consistent and add it to AWSRole nodes as well.

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -83,7 +83,7 @@ def get_account_access_key_data(session, username):
 def load_users(session, users, current_aws_account_id, aws_update_tag):
     ingest_user = """
     MERGE (unode:AWSUser{arn: {ARN}})
-    ON CREATE SET unode :AWSPrincipal, unode.userid = {USERID}, unode.firstseen = timestamp(),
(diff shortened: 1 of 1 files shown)


cartography/intel/aws/iam.py
```diff
@@ -168,7 +168,8 @@ def load_policies(session, policies, current_aws_account_id, aws_update_tag):
 def load_roles(session, roles, current_aws_account_id, aws_update_tag):
     ingest_role = """
     MERGE (rnode:AWSRole{arn: {Arn}})
```
achantavy: I would add the label with `MERGE rnode:AWSRole:AWSPrincipal{arn: {Arn}})` versus at the `on create set` so that all the labels for a node are defined in the same place.

ecdavis: Wouldn't that result in a MATCH being done on all AWSPrincipal nodes rather than just AWSRole nodes though?

achantavy: Oh, good point, oops

PR #80, opened 2019-06-04
Add AWSPrincipal nodes to represent the root identity of each AWS account being synced.

ecdavis:
Each AWS account has a root identity which is a principal (ref: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html ). This PR does a couple of things:

1. Explicitly adds a principal node for the root identity in each AWS account.
2. Refactors AWSPrincipal cleanup to make sure it does not delete principals which will be re-added during multi-account sync.

### DIFF
--- a/cartography/data/jobs/cleanup/aws_import_principals_cleanup.json
+++ b/cartography/data/jobs/cleanup/aws_import_principals_cleanup.json
@@ -1,0 +1,8 @@
+{
+  "statements": [{
+    "query": "MATCH (:AWSAccount{id: {AWS_ID}})-[:RESOURCE]->(n:AWSPrincipal) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
+    "iterative": true,
+    "iterationsize": 100
+  }],
+  "name": "cleanup AWSPrincipal"
+}
--- a/cartography/intel/aws/organizations.py
+++ b/cartography/intel/aws/organizations.py
@@ -89,9 +89,13 @@ def load_aws_accounts(neo4j_session, aws_accounts, aws_update_tag, common_job_pa
     ON CREATE SET aa.firstseen = timestamp()
     SET aa.lastupdated = {aws_update_tag}, aa.name = {ACCOUNT_NAME}
     WITH aa
-    MERGE (aa)-[:RESOURCE]->(root:AWSPrincipal{arn: {RootArn})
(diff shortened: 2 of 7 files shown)


cartography/intel/aws/organizations.py
```diff
@@ -88,12 +88,18 @@ def load_aws_accounts(neo4j_session, aws_accounts, aws_update_tag, common_job_pa
     MERGE (aa:AWSAccount{id: {ACCOUNT_ID}})
     ON CREATE SET aa.firstseen = timestamp()
     SET aa.lastupdated = {aws_update_tag}, aa.name = {ACCOUNT_NAME}
+    WITH aa
+    MERGE (aa)-[:RESOURCE]->(root:AWSPrincipal{arn: {RootArn})
```
achantavy: I think we need to set the `lastupdated` and `firstseen` on the `:RESOURCE` relationship here.

cartography/data/jobs/cleanup/aws_import_principals_cleanup.json
```diff
@@ -0,0 +1,8 @@
+{
+  "statements": [{
+    "query": "MATCH (:AWSAccount{id: {AWS_ID}})-[:RESOURCE]->(n:AWSPrincipal) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
achantavy: In other places we have a second cleanup statement for the relationship like:
```
"MATCH (:AWSAccount{id: {AWS_ID}})-[r:RESOURCE]->(:AWSPrincipal) WHERE r.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted
```

This would handle the case where the nodes are up-to-date but the relationship between them no longer exists (although I'm not sure if this happens in this specific scenario, like a root Principal can't exactly change what account it's associated with right?).

Still, it would be good to explicitly delete stale relationships since that's a pattern we use in a bunch of other places.

ecdavis: We don't follow that pattern for RESOURCE relationships due to the reason you described. Instead we rely on DETACH DELETE to do the right thing and remove the RESOURCE relationship when cleaning up stale nodes.

achantavy: Fair point 👍

PR #81, opened 2019-06-05
Change AWS IAM sync to evaluate role trust policies and expand wildcards.

ecdavis:
This change has two effects:
1. STS_ASSUMEROLE_ALLOW relationships will be deleted if a trust relationship does not exist between the role being assumed and the origin account.
2. STS_ASSUMEROLE_ALLOW relationships will be expanded if the target role is a wildcard (e.g. '*') to all roles matching the wildcard.


cartography/intel/aws/iam.py
```diff
@@ -336,6 +337,64 @@ def load_user_access_keys(session, user_access_keys, aws_update_tag):
...
+    WHERE NOT role.arn CONTAINS '*'
+    RETURN role.arn;
+    """
+
+    query = """
+    MATCH (origin:AWSAccount)-[:AWS_ROLE]->(assumer:AWSRole)
+        -[:STS_ASSUMEROLE_ALLOW]->(wildcard:AWSRole{arn: {WildcardArn}})
+    WITH origin, assumer, wildcard
+    MATCH (absolute:AWSRole{arn: {AbsoluteArn}})
+    WHERE exists( (absolute)-[:TRUSTS_AWS_PRINCIPAL]->(:AWSPrincipal{arn: 'arn:aws:iam::' + origin.id + ':root"}) )
```
achantavy: Using the reasoning in my earlier comment about partial trusts, I think we can remove this `where exists` clause so that we wildcard expand to all computed absolute roles (not just the ones that have trusts back to the origin account's root principal).

achantavy:
#283 performs this wildcard expansion, closing.

PR #84, opened 2019-06-06
Fix IAM statement parsing bug which would cause single-statement policies to crash sync

ecdavis:


cartography/intel/aws/iam.py
```diff
@@ -247,6 +247,8 @@ def load_group_memberships(session, group_memberships, aws_update_tag):
 def _find_roles_assumable_in_policy(policy_data):
     ret = []
     for statement in policy_data["PolicyDocument"]["Statement"]:
+        if isinstance(statement, str):
```
achantavy: What situation would make it only a string?

ecdavis: Serves me right for copying previous code. Looks like `policy_data["PolicyDocument"]["Statement"]` can either be a dictionary or a list of dictionaries. In the case where it's a dictionary the for loop would iterate over its keys, so the loop variable `statement` would be a string (the dictionary key).

achantavy:
LGTM

PR #87, opened 2019-06-12
Connect firewalls directly to the instances they apply to, and mark firewalls that have target service accounts

achantavy:
This PR directly connects instances to the firewalls that apply to them and takes into consideration the following cases:

- Target tags defined on a firewall cause the firewall to apply only to instances with those tags in the firewall's VPC
- Target service accounts defined on a firewall cause the firewall to apply only to instances with those service accounts in the firewall's VPC (this is not fully implemented as we need IAM objects for full functionality; for now this PR sets a `has_target_service_accounts` placeholder field set on the firewall object so we know to not apply the firewall to all instances in the VPC)
- Neither target tag nor target service account are defined so the firewall rule applies to all instances in the VPC

### DIFF
--- a/cartography/data/jobs/analysis/gcp_compute_asset_inet_exposure.json
+++ b/cartography/data/jobs/analysis/gcp_compute_asset_inet_exposure.json
@@ -12,9 +12,9 @@
     "__comment__": "Connect GCP ingress firewall rules to the instances that they apply to via target tags"
   },
   {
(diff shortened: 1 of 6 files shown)


cartography/data/jobs/analysis/gcp_compute_asset_inet_exposure.json
```diff
@@ -0,0 +1,26 @@
...
+    "__comment__": "Connect GCP ingress firewall rules that don't specify target tags and don't specify target service accounts to the instances that they apply to via sharing the same VPC"
+  },
+  {
+    "query": "MATCH (fw:GCPFirewall)-[a:FIREWALL_INGRESS]->(inst:GCPInstance)\nWHERE a.lastupdated <> {UPDATE_TAG}\nDELETE (a)\nRETURN count(*) as TotalCompleted",
+    "iterative": false,
+    "__comment__": "Delete stale firewall ingress relationships"
+  }
+],
+  "name": "GCP asset internet exposure"
+}
```
ecdavis: This job won't execute automatically, it needs to be invoked somewhere.

Issue #35, opened 2019-04-02
Unable to get running on MacOS


heydonovan:
I'm using pyenv if that helps at all:

```
$ cartography
Traceback (most recent call last):
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/bin/cartography", line 11, in <module>
    load_entry_point('cartography==0.2.2rc1', 'console_scripts', 'cartography')()
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/pkg_resources/__init__.py", line 480, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2693, in load_entry_point
    return ep.load()
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2324, in load
    return self.resolve()
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2330, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/cartography/cli.py", line 7, in <module>
    import cartography.sync
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/cartography/sync.py", line 8, in <module>
    import cartography.intel.aws
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/cartography/intel/aws/__init__.py", line 5, in <module>
    from cartography.intel.aws import dynamodb, ec2, elasticsearch, iam, organizations, route53, s3, rds
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/cartography/intel/aws/elasticsearch.py", line 2, in <module>
    from cartography.intel.dns import ingest_dns_record_by_fqdn
  File "/Users/donovanhernandez/.pyenv/versions/2.7.15/lib/python2.7/site-packages/cartography/intel/dns.py", line 2, in <module>
    import dns.resolver
ImportError: No module named resolver

$ /Users/donovanhernandez/.pyenv/versions/2.7.15/bin/pip list | grep dns
[...]

PR #93, opened 2019-06-21
Correct requirements to specify python3

ramonpetgrave64:
Fixes issue #35


achantavy:
Some user configs have pip point to pip3, and pip3 may not be present on all systems.  Closing this one, but on a broader note, I'm open to feedback on how to make the docs clearer.

PR #98, opened 2019-07-03
Support federated principal

mtanda:
Follow up https://github.com/lyft/cartography/pull/96 .

I check my environment, and notice that we have `Federated` principal type.

According to this document, AWS define 4 principal types.
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html
- `AWS`
- `Service`
- `Federated `
- `CanonicalUser`

Actually, we use SAML, it is the reason why `Federated` principal is included.
Currently `cartography` works well in my env.
So this is not required for me, but ignoring `Federated` principal type is not corrected behavior.
I try to fix it.
(I can't test for `CanonicalUser`, I ignore it now...)

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -212,6 +212,8 @@ def load_roles(session, roles, current_aws_account_id, aws_update_tag):
                 principal_type, principal_values = 'AWS', principal['AWS']
             elif 'Service' in principal:
                 principal_type, principal_values = 'Service', principal['Service']
+            elif 'Federated' in principal:
+                principal_type, principal_values = 'Federated', principal['Federated']
             if not isinstance(principal_values, list):
                 principal_values = [principal_values]
             for principal_value in principal_values:
--- a/tests/data/aws/iam.py
+++ b/tests/data/aws/iam.py
@@ -135,5 +135,30 @@ LIST_ROLES = {
             "Path": "/",
             "Arn": "arn:aws:iam::000000000000:role/example-role-2"
         },
+        {
+            "AssumeRolePolicyDocument": {
+                "Version": "2012-10-17",
+                "Statement": [
+                    {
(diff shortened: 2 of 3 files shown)


achantavy:
Overall it looks good to me.

The one note I will make is that we use `arn` as a unique identifier on `AWSPrincipal`, but in the [AWS web federation case](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html)(scroll to 'Federated users (using web identity federation)'), it looks like the unique identifier is a URL and not an ARN.

In general we tried to follow the convention where each node has an `id` field but in the case of `AWSPrincipals` it looks like we only use `arn` - so yeah, this is our fault.

Anyway, I'm personally ok with your code the way it is (allowing `AWSPrincipal.arn` be a URL in the web-federated case), but I'll run it over with the rest of the team and we'll reply back!

mtanda:
Thank you for taking your time so much.

cartography/intel/aws/iam.py
```diff
@@ -208,10 +208,9 @@ def load_roles(session, roles, current_aws_account_id, aws_update_tag):
         for statement in role["AssumeRolePolicyDocument"]["Statement"]:
             principal = statement["Principal"]
             principal_values = []
-            if 'AWS' in principal:
-                principal_type, principal_values = 'AWS', principal['AWS']
-            elif 'Service' in principal:
-                principal_type, principal_values = 'Service', principal['Service']
+            for t in ['AWS', 'Service', 'Federated']:
+                if t in principal:
```
achantavy: I understand what this code does but I feel like there's a neater way to do this without the `if` clause.  Are you able to revise?

mtanda: Sure! I did it!

mtanda:
Thanks lot!

PR #102, opened 2019-07-10
Add CRXcavator intel module to add visibility of Chrome extensions and associated risk scores.

tayasteere:
Added support for ingestion of CRXcavator (https://crxcavator.io) Chrome extension mapping and risk scoring. 
Updated documentation


cartography/intel/crxcavator.py
```diff
@@ -0,0 +1,210 @@
...
+        headers={'Accept': 'application/json',
+                 'API-Key': CRXcavator_API_KEY})
+    if data.ok:
+        return data.json()
+    else:
+        logger.error('Error during CRXcavator API call {}'.format(data.status_code))
+        return None
+
+
+def ingest_extensions(session, update_tag):
```
achantavy: Please refactor this to the following pattern:

`get_extension()` - to get the data (seems you already have this)
`transform_extension()` - to perform data adjustments (if needed) on the returned json data from `get_extension()`, for example it looks like handling permission warnings below performs some data massaging
`load_extension()` - to actually load the transformed data into neo4j

Then, define a function called `sync_extension` to call `get`, `transform`, and `load`.

cartography/intel/crxcavator.py
```diff
@@ -0,0 +1,210 @@
...
+            'users': data['webstore'].get('users'),
+            'website': data['webstore'].get('website'),
+            'type': data['webstore'].get('type'),
+            'price': data['webstore'].get('price')
+        })
+    logger.info('Ingesting {} extensions'.format(len(extensions)))
+    session.run(ingestion_cypher, ExtensionsData=extensions, UpdateTag=update_tag)
+
+
+def ingest_users_extensions(session, update_tag):
```
achantavy: Same comment about Get+Transform+Load.

cartography/intel/crxcavator.py
```diff
@@ -0,0 +1,210 @@
+import logging
+import json
+import requests
+import requests.auth
+import os
+
+from cartography.util import run_cleanup_job
+
+logger = logging.getLogger(__name__)
+
```
achantavy: Are there other assets under crxcavator that we might want to ingest in the future?  If so, I recommend making a cartography/intel/crxcavator folder and moving this logic to `crxcavator/__init__.py`.

cartography/intel/crxcavator.py
```diff
@@ -0,0 +1,210 @@
...
+def start_extension_ingestion(session, config):
+    """
+    If this module is configured, perform ingestion of CRXcavator data. Otherwise warn and exit
+    :param session: Neo4J session for database interface
+    :param config: Neo4J server URI and Update tag for data freshness
+    :return: None
+    """
+    if not CRXcavator_API_BASE_URL or not CRXcavator_API_KEY:
+        logger.warning('CRXcavator import is not configured - skipping this module')
+        return None
```
achantavy: No need to `return None`.  Just `return` to break out.

cartography/intel/crxcavator.py
```diff
@@ -0,0 +1,210 @@
...
+    Ingests the extension details into Neo4J
+    :param session: Neo4J session object for server communication
+    :param update_tag: Timestamp used to determine data freshness
+    :return: None
+    """
+    # the JSON returned from the CRXcavator API does not return well formatted objects
+    # instead, each object is named after it's key, making enumeration more difficult
+    # will build a cleaner object for import into graph
+
+    _data_index = 1
```
achantavy: Is this needed?  `ingest_users_extensions()` below doesn't use variables for row/col indexing.

cartography/intel/crxcavator.py
```diff
@@ -0,0 +1,210 @@
...
+        if not details:
+            continue
+        extension_id = details['extension_id']
+        version = details['version']
+        data = details['data']
+        # permissions warnings can be a list or null, if list concat to a single string
+        permission_warnings = data['webstore'].get('permission_warnings')
+        if permission_warnings:
+            permission_warnings = ', '.join(permission_warnings)
+        extensions.append({
```
achantavy: Please add some example return data to `cartography/tests/data/` so that we can some tests on ingesting crxcavator data.  Something similar to https://github.com/lyft/cartography/blob/master/tests/data/gcp/compute.py

tayasteere: done

cartography/intel/crxcavator.py
tayasteere: refactored to follow get,transform,load methodology for extensions and users

cartography/intel/crxcavator.py
tayasteere: Currently crxcavator is limited to chrome extensions, but refactored to this model in case of future changes.

cartography/intel/crxcavator.py
tayasteere: There is a big difference in the way the data is returned between the two API calls. See the data samples added to tests. Open to suggestions for refactoring if there is an better method to clean this data up.

cartography/intel/crxcavator/__init__.py
```diff
@@ -0,0 +1,35 @@
+import logging
+import os
+
+from cartography.intel.crxcavator.crxcavator import sync_extensions
+from cartography.util import run_cleanup_job
+
+logger = logging.getLogger(__name__)
+
+# API key for CRXcavator generaated from crxcavator.io portal
+CRXcavator_API_KEY = os.environ.get('CREDENTIALS_CRXCAVATOR_API_KEY')
```
achantavy: Use all-caps for the variable name here since it's a constant.

cartography/intel/crxcavator/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+# API for the CRXcavator API - https://api.crxcavator.io/v1 as of 07/09/19
+CRXcavator_API_BASE_URL = os.environ.get('CRXCAVATOR_URL')
```
achantavy: Ditto

cartography/intel/crxcavator.py
achantavy: Gotcha; was just wondering why you chose to have a var name for the indexes in this function but not the other.  Yeah, having the sample data helps.

cartography/intel/crxcavator.py
achantavy: Now that there's sample data, can you please add an integration test for your `load_*` functions? :)  Basically copy the format of [this](https://github.com/lyft/cartography/blob/master/tests/integration/cartography/intel/aws/test_dynamodb.py)

cartography/intel/crxcavator.py
tayasteere: done

cartography/intel/crxcavator.py
tayasteere: done

cartography/intel/crxcavator.py
tayasteere: integration test added

cartography/sync.py
```diff
@@ -141,9 +141,9 @@ def build_default_sync():
     sync = Sync()
     sync.add_stages([
         ('create-indexes', cartography.intel.create_indexes.run),
-        ('aws', cartography.intel.aws.start_aws_ingestion),
-        ('gcp', cartography.intel.gcp.start_gcp_ingestion),
+        # ('aws', cartography.intel.aws.start_aws_ingestion),
```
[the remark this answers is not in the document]
tayasteere: fixed

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    Performs the sequential tasks to collect, transform, and sync extension data
+    :param session: Neo4J session for database interface
+    :param common_job_parameters: Common job parameters containing UPDATE_TAG
+    :param crxcavator_api_key: The API key to access the CRXcavator service
+    :param crxcavator_base_url: The URL for the CRXcavator API
+    :return: None
+    """
+    extension_json = get_extensions(crxcavator_api_key, crxcavator_base_url)
+    if not extension_json:
```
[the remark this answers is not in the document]
tayasteere: change exception handling based on discussion with you and edavis

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    uri = crxcavator_base_url + api_and_parameters
+    data = requests.get(
+        uri,
+        headers={'Accept': 'application/json',
+                 'API-Key': crxcavator_api_key})
+    if data.ok:
+        return data.json()
+    else:
+        logger.error('Error during CRXcavator API call {}'.format(data.status_code))
```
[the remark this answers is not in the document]
tayasteere: added Requests exception raiser

cartography/intel/crxcavator/__init__.py
tayasteere: fixed

cartography/intel/crxcavator/__init__.py
tayasteere: fixed

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    # will build a cleaner object for import into graph
+
+    _data_index = 1
+    _row_index = 0
+
+    extensions = []
+    for extension in extension_json.items():
+        details = extension[_data_index][_row_index]
+        if not details:
+            continue
```
sachafaust: Seems expected. Should we log here so we know when this is happening?

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+            'short_description': data['webstore'].get('short_description'),
+            'size': data['webstore'].get('size'),
+            'support_site': data['webstore'].get('support_site'),
+            'users': data['webstore'].get('users'),
+            'website': data['webstore'].get('website'),
+            'type': data['webstore'].get('type'),
+            'price': data['webstore'].get('price')
+        })
+    if len(extensions) == 0:
+        raise ValueError('No extensions could be parsed from JSON data')
```
sachafaust: Is there cases where we could have no extensions? For example, if no extensions are installed by users, would we hit this or never get here? I'm worried about throwing exception here and crashing sync when we might just log.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    :param session: Neo4J session object for server communication
+    :param update_tag: Timestamp used to determine data freshness
+    :return: None
+    """
+    ingestion_cypher = """
+    UNWIND {ExtensionsData} as extension
+    MERGE (e:ChromeExtension{id: extension.id})
+    ON CREATE SET
+    e.extension_id = extension.extension_id,
+    e.version = extension.version,
```
sachafaust: I assume the extension id is unique per version?

README.md
```diff
@@ -226,6 +231,13 @@ WHERE rds.storage_encrypted = false
 return a.name as AWSAccount, count(rds) as UnencryptedInstances
 ```
 
+### What users have the Grammarly for Chrome extension installed?
```
sachafaust: I think it's beneficial to provide some queries around analyzing risks. This would also help in the PR review to see if the schema makes it easy.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+        details = extension[_data_index][_row_index]
+        if not details:
+            continue
+        extension_id = details['extension_id']
+        version = details['version']
+        data = details['data']
+        # permissions warnings can be a list or null, if list concat to a single string
+        permission_warnings = data['webstore'].get('permission_warnings')
+        if permission_warnings:
+            permission_warnings = ', '.join(permission_warnings)
```
sachafaust: [schema] We could use array here which are supported by Neo4j. Would make managing and querying them easier

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+        extensions.append({
+            'id': "{0}|{1}".format(extension_id, version),
+            'extension_id': extension_id,
+            'version': version,
+            'risk_total': data['risk'].get('total'),
```
sachafaust: [schema] is there a way to represent risk as Risk node like we do for other scenarios (cve, vulndb, safedb..). Some example of question and query to identify risk would help validate the schema (made same comment for the documentation)

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+            'risk_metadata': json.dumps(data['risk'].get('metadata')),
+            'address': data['webstore'].get('address'),
+            'email': data['webstore'].get('email'),
+            'icon': data['webstore'].get('icon'),
+            'last_updated': data['webstore'].get('last_updated'),
```
sachafaust: [schema-block] let's avoid issues with our internal lastupdated and rename to crxcavator_last_update for example

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    ingestion_cypher = """
+    UNWIND {ExtensionsData} as extension
+    MERGE (e:ChromeExtension{id: extension.id})
+    ON CREATE SET
+    e.extension_id = extension.extension_id,
+    e.version = extension.version,
+    e.firstseen = timestamp()
+    SET
+    e.extcalls = extension.extcalls,
+    e.risk_total = extension.risk_total,
```
sachafaust: [schema] is there a way to represent risk as Risk node like we do for other scenarios (cve, vulndb, safedb..). Some example of question and query to identify risk would help validate the schema (made same comment for the documentation)

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    e.risk_metadata = extension.risk_metadata,
+    e.address = extension.address,
+    e.email = extension.email,
+    e.icon = extension.icon,
+    e.last_updated = extension.last_updated,
```
sachafaust: [schema-block] let's avoid issues with our internal lastupdated and rename to crxcavator_last_update for example

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    users_set = set()
+    extensions_by_user = []
+    for extension in user_extensions:
+        for details in extension[1].items():
+            for user in details[1]['users']:
+                users_set.add(user)
+                extensions_by_user.append({
+                    'id': "{0}|{1}".format(extension[0], details[0]),
+                    'user': user})
+    if len(users_set) == 0:
```
sachafaust: Both case would make the sync crash. Is that what we want or should theses exceptions be handled up the stack and/or logged as warning?

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    Ingests the extension to user mapping details into Neo4J
+    :param users: List of user objects to create for mapping
+    :param extensions_by_user: List of user to extension id mappings
+    :param session: Neo4J session object for server communication
+    :param update_tag: Timestamp used to determine data freshness
+    :return: None
+    """
+
+    # todo: once G Suite ingestion is created, move first MERGE to that process
```
sachafaust: I think MERGE is fine now and in the future. I think it's fair to assume that each intel modules aren't aware of others

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    user_ingestion_cypher = """
+    UNWIND {Users} as user_email
+    MERGE (user:GSuiteUser{id: user_email, email: user_email})
+    ON CREATE SET
+    user.firstseen = timestamp()
+    SET user.lastupdated = {UpdateTag}
+    """
+    # todo: add relationship to computer node when JAMF data is moved to Cartography
```
sachafaust: Connecting modules together should be done through enrichment jobs. The expectation is that our internal version does have such connectivity and we can push it back to open source once JAMF module gets transfered

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    extension_ingestion_cypher = """
+    UNWIND {ExtensionsUsers} as extension_user
+    MATCH (user:GSuiteUser{email: extension_user.user}),(ext:ChromeExtension{id:extension_user.id})
+    MERGE (user)-[r:INSTALLS]->(ext)
```
sachafaust: [schema] other areas of the graph use DEPLOYED vs INSTALLS, should we have consistency there?

cartography/intel/crxcavator/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+def start_extension_ingestion(session, config):
+    """
+    If this module is configured, perform ingestion of CRXcavator data. Otherwise warn and exit
+    :param session: Neo4J session for database interface
+    :param config: Neo4J server URI and Update tag for data freshness
+    :return: None
+    """
+    if not CRXCAVATOR_API_BASE_URL or not CRXCAVATOR_API_KEY:
+        logger.debug('CRXcavator import is not configured - skipping this module. See docs to configure.')
```
sachafaust: should we print something to console and/or warning?

cartography/intel/crxcavator/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+# API for the CRXcavator API - https://api.crxcavator.io/v1 as of 07/09/19
+CRXCAVATOR_API_BASE_URL = os.environ.get('CRXCAVATOR_URL')
+
+
+def start_extension_ingestion(session, config):
+    """
+    If this module is configured, perform ingestion of CRXcavator data. Otherwise warn and exit
+    :param session: Neo4J session for database interface
+    :param config: Neo4J server URI and Update tag for data freshness
```
sachafaust: for consistency - 

:param config: A `cartography.config` object

cartography/intel/crxcavator/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    :return: None
+    """
+    if not CRXCAVATOR_API_BASE_URL or not CRXCAVATOR_API_KEY:
+        logger.debug('CRXcavator import is not configured - skipping this module. See docs to configure.')
+        return
+
+    common_job_parameters = {
+        "UPDATE_TAG": config.update_tag,
+    }
+    sync_extensions(session, common_job_parameters, CRXCAVATOR_API_KEY, CRXCAVATOR_API_BASE_URL)
```
sachafaust: should we test the creds first and indicate that they are wrong to the user before moving to ingestion?

sachafaust: @ecdavis is this something we want to do or let it move ingest and error?

cartography/intel/crxcavator/crxcavator.py
tayasteere: Prior version had log and return, but discussion with ecdavis was that preferred method was to fail sync on unrecoverable error. While it is possible no extensions would be returned if none are installed, if they have configured this module this case should not occur and an error would be warranted.

cartography/intel/crxcavator/crxcavator.py
tayasteere: This would only occur when an extension has bad data returned from the API. Added logging statement.

cartography/intel/crxcavator/crxcavator.py
tayasteere: changed to array

cartography/intel/crxcavator/crxcavator.py
tayasteere: Discussed in person earlier. In this context, risk score is part of the already derived data from CRXcavator and does not make sense as a separate node.

cartography/intel/crxcavator/crxcavator.py
tayasteere: updated

cartography/intel/crxcavator/crxcavator.py
tayasteere: Same as prior comment, current approach is to crash sync per discussion with ecdavis.

cartography/intel/crxcavator/crxcavator.py
tayasteere: Based on discussion during design for this module. We're only creating this information here until full G Suite ingest is created. At that time, this would change to a relationship creation only.

cartography/intel/crxcavator/crxcavator.py
tayasteere: This is here as a reminder to do that work when JAMF is connected.

cartography/intel/crxcavator/crxcavator.py
tayasteere: DEPLOYED would be the appropriate tag when linking to a computer object (after JAMF ingestion). INSTALLS is used here because the linked user has the extension in their profile which would be installed on any Chrome instance they log into.

cartography/intel/crxcavator/__init__.py
tayasteere: Was originally info, downgraded to debug to match similar pattern in GCP, but noticed it is logged as both debug and error. Changed to warning.

cartography/intel/crxcavator/__init__.py
tayasteere: changed

README.md
tayasteere: gave example of querying by risk_total

sollecrab:
The report part that Taya shared with me looks great, thanks!

cartography/intel/crxcavator/crxcavator.py
sachafaust: What about day 1 setup where they got the service going but waiting for the extension to deploy? Might be edge case but still valid to avoid hard crash. Not PR blocker.

cartography/intel/crxcavator/crxcavator.py
sachafaust: [not pr blocker] it make sense but what about the scenario of calculating risk across asset? For example

(Risk)-[..*]->(Human) . count(risk). In the intent to measure risk, we may want to normalize. Can be done now vs later but I find it's important to think about schema and future in open source as we would introduce breaking changes others may not appreciate. CC: @ecdavis

README.md
sachafaust: [blocker] Let's not make a judgment on extension but focus on the risk score provided by the data.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -0,0 +1,220 @@
...
+    :param extensions_by_user: List of user to extension id mappings
+    :param session: Neo4J session object for server communication
+    :param update_tag: Timestamp used to determine data freshness
+    :return: None
+    """
+
+    # todo: once G Suite ingestion is created, move first MERGE to that process
+    user_ingestion_cypher = """
+    UNWIND {Users} as user_email
+    MERGE (user:GSuiteUser{id: user_email, email: user_email})
```
sachafaust: using id and email to merge concerns me. We stick to id everywhere else. Email for Gsuite User is also mutable.

cartography/intel/crxcavator/crxcavator.py
sachafaust: we need that analysis job internally.

cartography/intel/crxcavator/crxcavator.py
sachafaust: [not pr blocker] . @ecdavis to comment related to schema consistency.

cartography/intel/crxcavator/crxcavator.py
vatsaldesai93: If there are no users, there are no extensions. If there are users (at least one), they will at least have the 'CRXCavator Gatherer' extension installed. 'CRXCavator Gatherer' is mandatory for user/extension data to get added/updated to the CRX platform.

README.md
```diff
@@ -32,16 +31,22 @@ Time to set up the server that will run Cartography.  Cartography _should_ work
...
     3. Ensure that the machine you are running Cartography on can authenticate to this identity.  
         - **Method 1**: You can do this by setting your `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to a json file containing your credentials.  As per SecurityCommonSense™️, please ensure that only the user account that runs Cartography has read-access to this sensitive file.
         - **Method 2**: If you are running Cartography on a GCE instance or other GCP service, you can make use of the credential management provided by the default service accounts on these services.  See the [official docs](https://cloud.google.com/docs/authentication/production) on Application Default Credentials for more details.
 
+4. If you're a CRXcavator user, **prepare your CRXcavator API key**
+
+    1. Generate an API key from your CRXcavator [user page](https://crxcavator.io/user/settings#)
+    2. Populate the following environment variables in the shell running the sync service
+        1. CRXcavator_API_KEY - the full URL to the CRXcavator API. https://api.crxcavator.io/v1 as of 07/09/19
```
achantavy: I think this is a typo.  Did you mean `CRXCAVATOR_URL`?

README.md
achantavy: Personally I wouldn't block on adding risk analysis documentation.  I think a better place for that would be a blog post.

README.md
```diff
@@ -226,6 +231,21 @@ WHERE rds.storage_encrypted = false
...
 ```
 
+### What users have the Grammarly for Chrome extension installed?
+```
+MATCH (u:GSuiteUser)-[r:INSTALLS]->(ext:ChromeExtension)
+WHERE ext.name CONTAINS 'Grammarly'
+return ext.name, ext.version, u.email
+```
+
+### What users have installed extensions that are risky based on [CRXcavator scoring](https://crxcavator.io/docs#/risk_breakdown)?
```
achantavy: Use `####` not `###` to match indent level with the other example questions in this section.

README.md
```diff
@@ -226,6 +231,21 @@ WHERE rds.storage_encrypted = false
 return a.name as AWSAccount, count(rds) as UnencryptedInstances
 ```
```
achantavy: Use `####` not `###` to match indent level with the other example questions in this section.

README.md
```diff
@@ -32,16 +31,22 @@ Time to set up the server that will run Cartography.  Cartography _should_ work
...
     2. Ensure that this identity has the [securityReviewer](https://cloud.google.com/iam/docs/understanding-roles) role attached to it.
     3. Ensure that the machine you are running Cartography on can authenticate to this identity.  
         - **Method 1**: You can do this by setting your `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to a json file containing your credentials.  As per SecurityCommonSense™️, please ensure that only the user account that runs Cartography has read-access to this sensitive file.
         - **Method 2**: If you are running Cartography on a GCE instance or other GCP service, you can make use of the credential management provided by the default service accounts on these services.  See the [official docs](https://cloud.google.com/docs/authentication/production) on Application Default Credentials for more details.
 
+4. If you're a CRXcavator user, **prepare your CRXcavator API key**
+
+    1. Generate an API key from your CRXcavator [user page](https://crxcavator.io/user/settings#)
+    2. Populate the following environment variables in the shell running the sync service
```
achantavy: Replace 'the sync service' with 'cartography'

cartography/intel/crxcavator/__init__.py
achantavy: I don't think there's a need to check for cred validity twice.

This module is different from AWS and GCP in that it uses the `requests` library to make raw HTTP calls rather than using a client library.  `crxcavator.call_crxcavator_api()` does this and `data.raise_for_status()` [will raise an exception](https://2.python-requests.org//en/master/_modules/requests/models/#Response.raise_for_status) if the creds don't work.

cartography/intel/crxcavator/crxcavator.py
achantavy: I'm with @sachafaust here.  If there is an error, then `data.raise_for_status()` in `crxcavator.call_crxcavator_api()` will raise an exception and the sync would have failed much earlier on.  I don't see a case for specifically raising an exception if the length of data here is 0.

cartography/intel/crxcavator/crxcavator.py
achantavy: I would be in favor of using email because the [GSuite APIs use it as a primary key](https://developers.google.com/admin-sdk/directory/v1/guides/manage-users#json-response).  To manage @sachafaust's concern, there is also a "unique user id" mentioned in the doc too that we could use, but I don't know if you have this visibility from CRXcavator.  Additional doc: https://developers.google.com/admin-sdk/directory/v1/reference/users/get

cartography/intel/crxcavator/crxcavator.py
achantavy: I'm in favor of removing the TODO from code and logging an issue in Github to track it.

tests/integration/cartography/intel/crxcavator/test_crxcavator.py
```diff
@@ -0,0 +1,181 @@
...
+                                                                 user_extensions_list,
+                                                                 neo4j_session,
+                                                                 TEST_UPDATE_TAG)
+
+    query = """
+    MATCH(user:GSuiteUser{id:{UserId}})
+    RETURN user.id, user.email
+    """
+    expected_user_id = 'user@example.com'
+    nodes = neo4j_session.run(query,
```
achantavy: Nit: Use this style if the line is too long for pep8
```
object.function_call(
    arg1,
    arg2,
    ...,
    arg3)
```

tests/integration/cartography/intel/crxcavator/test_crxcavator.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_transform_and_load_user_extensions(neo4j_session):
+    """
+    Ensure we can transform and load users and extension mapping.
+    """
+    users_res = tests.data.crxcavator.crxcavator.USER_RESPONSE
+    type(users_res)
+    users_list, user_extensions_list = cartography.intel.crxcavator.crxcavator.transform_user_extensions(users_res)
+    cartography.intel.crxcavator.crxcavator.load_user_extensions(users_list,
```
achantavy: Nit: Use this style if the line is too long for pep8
```
object.function_call(
    arg1,
    arg2,
    ...,
    arg3)
```

cartography/intel/crxcavator/crxcavator.py
tayasteere: Absolutely agree. However, this whole bit is temporary until G Suite user data is ingested directly from Google. We're only creating a place holder user and email is the only attribute known by CRXcavator. We could use a temporary made up id for now or email, either way it'll get replaced by the real value later.

README.md
tayasteere: No judgement was made on risk of a specific extension, the first example only gives an example of querying by name. The second example is querying all extensions by risk score using some sample criteria from the risk scoring page. Changed name in sample to a fake name to clear up any potential concern though.

cartography/intel/crxcavator/crxcavator.py
tayasteere: Agree, we have to connect to Computer and People nodes internally, but that is outside the scope of this PR.

README.md
tayasteere: fixed

README.md
tayasteere: fixed

README.md
tayasteere: fixed

README.md
tayasteere: fixed

tests/integration/cartography/intel/crxcavator/test_crxcavator.py
tayasteere: fixed

tests/integration/cartography/intel/crxcavator/test_crxcavator.py
tayasteere: fixed

cartography/intel/crxcavator/crxcavator.py
tayasteere: Added issue to update after G Suite data is ingested. We should use the unique user id as the node id at that time.

cartography/intel/crxcavator/crxcavator.py
ecdavis: Agreed we should normalize at some point however at this stage I think we're fairly unclear what a good normalization would look like. I think that will become clearer as we gather data from more sources and better understand how we use that data. The Risk node as it currently exists represents a discrete issue vs. a general risk assessment. We probably want to address this discrepancy in future, but I think that's a problem for a subsequent PR.

cartography/intel/crxcavator/crxcavator.py
ecdavis: Agree with @tayasteere -- DEPLOYED is a relationship from a piece of software to a piece of hardware (or virtual hardware). INSTALLS in this case goes from user to software. With JAMF we may be able to populate DEPLOYED relationships based on the presence of INSTALLS relationships.

PR #108, opened 2019-07-18
discard unneeded query results.

gleentea:
Hi maintainers.

When a large number of queries were executed, the socket buffer was stuck unless the execution results were not read.
I fixed it by discard unneeded query result.

### DIFF
--- a/cartography/intel/aws/dynamodb.py
+++ b/cartography/intel/aws/dynamodb.py
@@ -42,7 +42,7 @@ def load_dynamodb_tables(session, data, region, current_aws_account_id, aws_upda
             Rows=table['Table']['ItemCount'],
             AWS_ACCOUNT_ID=current_aws_account_id,
             aws_update_tag=aws_update_tag
-        )
+        ).detach()
         load_gsi(session, table, region, current_aws_account_id, aws_update_tag)
 
 
@@ -72,7 +72,7 @@ def load_gsi(session, table, region, current_aws_account_id, aws_update_tag):
             GSIName=gsi['IndexName'],
             AWS_ACCOUNT_ID=current_aws_account_id,
             aws_update_tag=aws_update_tag
-        )
+        ).detach()
 
 
 def cleanup_dynamodb_tables(session, common_job_parameters):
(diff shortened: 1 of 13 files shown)


achantavy:
Hi there, thanks for your PR!  Can you describe the problem a bit more, maybe post your console log?

achantavy:
Can you also post back with your steps to reproduce the error?  Which session.run() did this fail on?

gleentea:
## Reproduce

Run following PoC.
```
import neo4j

uri = 'bolt://localhost'
auth=("neo4j","password")
driver = neo4j.GraphDatabase.driver(uri, auth=auth)
with driver.session() as session:
    count = 0
    while True:
        session.run('MATCH (n:NotAvailableNode) return n') # run without .detach()
        count += 1
        if not count % 10000 :
            print(count)
```

## Result

Running PoC without detach() will fill the TCP socket buffer and it stop when 'count' reaches about 20000.

Example state of tcp socket.
```
Active Internet connections (including servers)
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)
tcp4  408044 294444  127.0.0.1.7687         127.0.0.1.51477        ESTABLISHED
tcp4  608756 146965  127.0.0.1.51477        127.0.0.1.7687         ESTABLISHED
```

Running PoC with detach() will the TCP socket buffer does not fill up and PoC continues to run.

This bug is caused by a problem with the neo4j driver usage.

sachafaust:
@achantavy @ecdavis should we look at implementing a generic run_query(session, query, params ...) type method that incorporates the detach()? This would also open up a place where we can make improvements centrally like (logging, retry logic, ....)

achantavy:
Interesting finding, thank you very much for this!

Ran your script and without detach I get 

```
Failed at count = 540606
  File "/Users/achantavy/Desktop/neo4j-detach-stress-test.py", line 16, in <module>
  [...]    
BrokenPipeError: [Errno 32] Broken pipe
```

With detach my count is at about 1.5 million and still going with no problems.  We'll take a closer look soon.

ecdavis:
Created #170 for long-term tracking of this issue. Not going to merge this PR as-is.

PR #110, opened 2019-07-30
Allow add shortcuts to have shortcuts as arguments.

thomashli:


tests/integration/driftdetect/test_shortcuts.py
```diff
@@ -30,6 +30,40 @@ def test_basic_add_shortcuts():
...
 
+def test_shortcut_parsing():
+    """
+    Tests add_shortcut can parse shortcuts.
+    """
+    cli = CLI(prog="cartography-detectdrift")
+    directory = "tests/data/test_cli_detectors/detector"
+    alias = "test_shortcut"
+    alias_2 = "test_shortcut_2"
+    file = "1.json"
```
achantavy: `file` is a reserved word; rename to something else.

cartography/driftdetect/storage.py
```diff
@@ -39,3 +39,13 @@ def walk(cls, drift_detection_directory):
         for root, directories, _ in os.walk(drift_detection_directory):
             for directory in directories:
                 yield os.path.join(root, directory)
+
+    @classmethod
+    def has_file(cls, file):
```
achantavy: Rename `file`

cartography/driftdetect/add_shortcut.py
```diff
@@ -49,7 +45,11 @@ def add_shortcut(storage, shortcut_serializer, query_directory, alias, filename)
     shortcut_path = os.path.join(query_directory, "shortcut.json")
     shortcut_data = storage.load(shortcut_path)
     shortcut = shortcut_serializer.load(shortcut_data)
-    shortcut.shortcuts[alias] = filename
+    fp = shortcut.shortcuts.get(filename, filename)
```
achantavy: How about this case?
Suppose:
```
filename = 'a'
alias = 'a'
```

and `shortcut.shortcuts` is already defined:
```
shortcut.shortcuts = {
 'a': 'somethingelse.json'
}
```

Then,
```
fp = shortcut.shortcuts.get('a', 'a') ==> 'somethingelse.json'
```

so `somethingelse.json` is still the filename and not `a`, so the code did not update the shortcut.

thomashli: Yeah I just noticed an issue where if the shortcut is the name of another file that produces that error. I added a check for it, but I'm worried that there might be too many IO calls between the storage unit and the program. It's a pretty pathological counter example which I don't think would occur in practice, and I'm debating between the check and just adding a warning to the help documentation for the subparser. What do you think?

PR #111, opened 2019-07-30
Drift Detection: refactored reporting to use lists instead of dicts to maintain result order. 

thomashli:


cartography/driftdetect/detect_deviations.py
```diff
@@ -97,12 +95,12 @@ def compare_states(start_state, end_state):
...
         if result in start_state.results:
             continue
-        drift_info = {}
-        for prop, field in zip(end_state.properties, result):
+        drift = []
+        for field in result:
             value = field.split("|")
             if len(value) > 1:
-                drift_info[prop] = value
+                drift.append(value)
```
achantavy: This line appends a list to `drift`, and the `else` case appends a `string` to `drift`.  Have you run into problems with mixed types here?

thomashli: Nope. I unit tested the serialization and deserialization and they seem to work fine.

PR #115, opened 2019-07-30
[noop] pre-commit linting

skiptomyliu:
Standardizing our coding with pre-commit.


achantavy:
Is there a way to add this linter to the `make test`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,42 @@
+repos:
```
skiptomyliu: Add pre-commit config

Makefile
```diff
@@ -1,7 +1,7 @@
 test: test_lint test_unit test_integration
 
 test_lint:
-	flake8
+	pre-commit run --all-files --show-diff-on-failure
```
skiptomyliu: Replace `flake8` with `pre-commit` which runs `flake8`

test-requirements.txt
```diff
@@ -1,3 +1,3 @@
-flake8
+pre-commit
```
skiptomyliu: add `pre-commit` as a test dependency.

skiptomyliu:
Thanks -- just pushed another commit to add it as a new dependency.

ecdavis:
Is the double-quote string fixer necessary? I'm not sure what the benefit is and it's actually introduced inconsistencies because we _have_ to double-quote strings which contain single-quote characters (or escape the single-quotes... or change them to double-quotes I guess).

skiptomyliu:
@ecdavis - not necessary, I've eliminated the `double-quote-string-fixer`, and have re-run it.

achantavy:
Woot looks like this passed our internal tests.  Please update https://github.com/lyft/cartography/blob/master/docs/testing.md with the new mention of pre-commit and any other changes you think make sense

PR #120, opened 2019-08-05
add ec2 keypairs ingestion

ramonpetgrave64:
Ingest EC2 Key Pairs and associate them with EC2 Instances.
Fixes issue #121.

### DIFF
--- a/cartography/data/indexes.cypher
+++ b/cartography/data/indexes.cypher
@@ -17,7 +17,7 @@ CREATE INDEX ON :DynamoDBGlobalSecondaryIndex(id);
 CREATE INDEX ON :EC2Instance(instanceid);
 CREATE INDEX ON :EC2Instance(publicdnsname);
 CREATE INDEX ON :EC2Reservation(reservationid);
-CREATE INDEX ON :EC2KeyPair(keyname);
+CREATE INDEX ON :EC2KeyPair(arn);
 CREATE INDEX ON :EC2SecurityGroup(id);
 CREATE INDEX ON :EC2SecurityGroup(groupid);
 CREATE INDEX ON :EC2Subnet(id);
--- a/cartography/data/indexes.cypher
+++ b/cartography/data/indexes.cypher
@@ -17,6 +17,7 @@ CREATE INDEX ON :DynamoDBGlobalSecondaryIndex(id);
 CREATE INDEX ON :EC2Instance(instanceid);
 CREATE INDEX ON :EC2Instance(publicdnsname);
 CREATE INDEX ON :EC2Reservation(reservationid);
+CREATE INDEX ON :EC2KeyPair(keyname);
 CREATE INDEX ON :EC2SecurityGroup(id);
 CREATE INDEX ON :EC2SecurityGroup(groupid);
 CREATE INDEX ON :EC2Subnet(id);
--- a/cartography/intel/aws/ec2.py
+++ b/cartography/intel/aws/ec2.py
@@ -33,6 +33,18 @@ def get_ec2_security_group_data(session, region):
     return {'SecurityGroups': security_groups}
 
 
+def get_ec2_key_pairs(session, region):
+    client = session.client('ec2', region_name=region, config=_get_botocore_config())
+    # No paginator for describe_key_pairs.
+    # paginator = client.get_paginator('describe_key_pairs')
+    # key_pairs = []
+    # for page in paginator.paginate():
+    #     key_pairs.extend(page['KeyPairs'])
+    # return {'KeyPairs': key_pairs}
+    result = client.describe_key_pairs()
+    return result
+
+
 def get_ec2_instances(session, region):
     client = session.client('ec2', region_name=region, config=_get_botocore_config())
     paginator = client.get_paginator('describe_instances')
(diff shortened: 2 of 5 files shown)


cartography/intel/aws/ec2.py
```diff
@@ -72,6 +84,32 @@ def get_ec2_vpcs(session, region):
     return client.describe_vpcs()
 
 
+def load_ec2_key_pairs(session, data, region, current_aws_account_id, aws_update_tag):
+    ingest_key_pair = """
+    MERGE (keypair:KeyPair:EC2KeyPair{keyname: {KeyName}, region: {Region}})
```
sachafaust: [schema standard] we recommend the use of a unique identifying attribute 'id' to help standardize on what to MERGE on and index.

cartography/data/indexes.cypher
```diff
@@ -17,6 +17,7 @@ CREATE INDEX ON :DynamoDBGlobalSecondaryIndex(id);
 CREATE INDEX ON :EC2Instance(instanceid);
 CREATE INDEX ON :EC2Instance(publicdnsname);
 CREATE INDEX ON :EC2Reservation(reservationid);
+CREATE INDEX ON :EC2KeyPair(keyname);
```
sachafaust: [schema] see comment below. having ID attribute is recommended by our standard.

cartography/intel/aws/ec2.py
```diff
@@ -111,6 +149,22 @@ def load_ec2_instances(session, data, region, current_aws_account_id, aws_update
     SET r.lastupdated = {aws_update_tag}
     """
 
+    ingest_key_pair = """
+    MERGE (keypair:KeyPair:EC2KeyPair{keyname: {KeyName}, region: {Region}})
```
sachafaust: recommend to merge on ID only

cartography/intel/aws/ec2.py
sachafaust: The id could be inferred by keyname-region for example.

cartography/intel/aws/ec2.py
```diff
@@ -111,6 +149,22 @@ def load_ec2_instances(session, data, region, current_aws_account_id, aws_update
...
+    ON CREATE SET keypair.firstseen = timestamp()
+    SET keypair.lastupdated = {aws_update_tag}
+    WITH keypair
+    MATCH (aa:AWSAccount{id: {AWS_ACCOUNT_ID}})
+    MERGE (aa)-[r:RESOURCE]->(keypair)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
+    with keypair
+    MATCH (instance:EC2Instance{instanceid: {InstanceId}})
```
sachafaust: Is it possible to have a Keypair that is not associated with EC2Instance? If so, I recommend to decouple the KeyPair node creation and the link to EC2Instance into separate statement

cartography/intel/aws/ec2.py
achantavy: To elaborate on Sacha's comment, we recommend doing something like
```
MERGE (keypair:KeyPair:EC2KeyPair{id: {KeyName}, keyname: {KeyName}, region: {Region}})
# assuming of course that {KeyName} uniquely identifies a KeyPair in all of AWS
# For this reason we like ARNs if possible
```

cartography/intel/aws/ec2.py
```diff
@@ -72,6 +84,32 @@ def get_ec2_vpcs(session, region):
...
+    SET keypair.keyfingerprint = {KeyFingerprint}, keypair.lastupdated = {aws_update_tag}
+    WITH keypair
+    MATCH (aa:AWSAccount{id: {AWS_ACCOUNT_ID}})
+    MERGE (aa)-[r:RESOURCE]->(keypair)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
+    """
+
+    for key_pair in data['KeyPairs']:
+        key_name = key_pair.get("KeyName", "")
```
achantavy: Since this code uses `KeyName` as an `id`, I don't think it would make sense to ingest a KeyPair that doesn't have a `KeyName`.  

For this reason, I think we should use `key_name = key_pair['KeyName']` so that the sync will crash (indicated bad data) if 'KeyName' is not present.

cartography/intel/aws/ec2.py
```diff
@@ -72,6 +84,32 @@ def get_ec2_vpcs(session, region):
...
+        key_fingerprint = key_pair.get("KeyFingerprint", "")
```
achantavy: I recommend changing this to

```
key_fingerprint = key_pair.get('KeyFingerprint')
```

This way the field will not show up on the Neo4j node if it is undefined.  I'd rather have the field not be there rather than as an empty string.  (I realize there's some inconsistencies in other parts of our code but I think going forward we should make missing fields `None` rather than the empty string.)
```

cartography/intel/aws/ec2.py
ramonpetgrave64: It is possible, and `load_ec2_key_pairs()` handles that. But I think the cleanest way to be aware of the association between and EC2Instance and an EC2KeyPair is from within `load_ec2_instances()`

ramonpetgrave64:
Resubmitted with #121.

PR #126, opened 2019-08-09
Ingest gsuite users and groups

skiptomyliu:
Pull gsuite users and groups from google's api.


cartography/intel/gsuite/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+import googleapiclient.discovery
+from oauth2client.client import ApplicationDefaultCredentialsError
+from oauth2client.client import GoogleCredentials
+
+from cartography.intel.gsuite import api
+
+
+OAUTH_SCOPE = [
+    'https://www.googleapis.com/auth/admin.directory.user',
```
skiptomyliu: oauth scope needs to be changed to `readonly` format before merge.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,108 @@
...
+from googleapiclient.discovery import HttpError
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_all_groups(admin):
+    """
+    Return list of Google Groups in your organization
+    Returns empty list if we are unable to enumerate the users for any reasons
```
skiptomyliu: `users` -> `groups` copy pasta

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,152 @@
...
+    groups = []
+    while request is not None:
+        try:
+            resp = request.execute()
+        except HttpError as e:
+            logger.warning('HttpError occurred in api.get_all_groups(), returning empty list. Details: %r', e)
+            groups = []
+            break
+        groups = groups + resp.get('groups', [])
+        break
```
skiptomyliu: remove when merge.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,152 @@
...
+    users = []
+    while request is not None:
+        try:
+            resp = request.execute()
+        except HttpError as e:
+            logger.warning('HttpError occurred in api.get_all_users(), returning empty list. Details: %r', e)
+            users = []
+            break
+        users = users + resp.get('users', [])
+        break
```
skiptomyliu: remove when merge

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+        ON CREATE SET r.lastupdated = {UpdateTag}
+    """
+    return """
+        UNWIND {MemberData} as member
+        MATCH (user:GSuiteUser {id: member.id}),(group:GSuiteGroup {id: {GroupID} })
+        MERGE (user)-[r:MEMBER_GSUITE_GROUP]->(group)
+    """
+
+
+def load_gsuite_members(session, group, members, gsuite_update_tag):
```
skiptomyliu: Need an extra pair of eyes on this function.  It's doing two queries to link:
1.  users to groups
2.  groups to groups (since a group can have a group)

Not sure if there's a more performant way of doing this.

achantavy: Two separate queries is perfectly fine.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
+import logging
+
+from googleapiclient.discovery import HttpError
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_all_groups(admin):
```
achantavy: Consider having some of the `get_*` functions return the raw response objects.  If I recall correctly, `response[id]` in the Google APIs is a helpful prefix that I've used in the past to uniquely identify nodes.  See https://github.com/lyft/cartography/blob/master/cartography/intel/gcp/compute.py#L141

On the other hand, since you use a group email later on to enumerate users, keeping the return type here the same can make sense.

In general though I'd recommend having a `get_*` in the Google API family return response objects and then have the transform functions massage them to usable objects for ingestion.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def get_members_for_group(admin, group_email):
+    """ Get all members for a google group
+
+    :param group_email: A string representing the email address for the group
+
+    :return: List of dictionaries representing Users or Groups.
+    """
+    request = admin.members().list(
```
achantavy: Is it possible to pass in the group as a parameter to this function so we don't need to do a `list` twice and possibly get conflicting data?  (maybe a new group was added in the second call)

In https://github.com/lyft/cartography/blob/master/cartography/intel/gcp/compute.py#L939, we get the `zones` in the `sync` function and pass that object as input to `sync_gcp_instances()`.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+            logger.warning('HttpError occurred in api.get_members_for_group(), returning empty list. Details: %r', e)
+            members = []
+            break
+        members = members + resp.get('members', [])
+        request = admin.members().list_next(request, resp)
+
+    return members
+
+
+def get_all_users(admin):
```
achantavy: Same comments about looping and returning the raw response objects.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+            logger.warning('HttpError occurred in api.get_all_users(), returning empty list. Details: %r', e)
+            users = []
+            break
+        users = users + resp.get('users', [])
+        # break
+        request = admin.users().list_next(request, resp)
+    return users
+
+
+def get_ingestion_groups_qry():
```
achantavy: No need to factor this out to its own function.  I would keep this in the `load` for readability.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+        g.lastupdated = {UpdateTag}
+    """
+
+
+def load_gsuite_groups(session, groups, gsuite_update_tag):
+    logger.info('Ingesting {} gsuite groups'.format(len(groups)))
+    session.run(get_ingestion_groups_qry(), GroupData=groups, UpdateTag=gsuite_update_tag)
+
+
+def get_ingestion_users_qry():
```
achantavy: Moving this within the `load` is fine

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+        u.lastupdated = {UpdateTag}
+    """
+
+
+def load_gsuite_users(session, users, gsuite_update_tag):
+    logger.info('Ingesting {} gsuite users'.format(len(users)))
+    session.run(get_ingestion_users_qry(), UserData=users, UpdateTag=gsuite_update_tag)
+
+
+def get_ingestion_membership_qry():
```
achantavy: Move this within the `load`

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+def cleanup_gsuite_groups(session, common_job_parameters):
+    # XXX: TODO
+    pass
+
+
+def sync_gsuite_users(session, admin, gsuite_update_tag, common_job_parameters):
+    """
+    Get GCP organization data using the CRM v1 resource object, load the data to Neo4j, and clean up stale nodes.
+    :param session: The Neo4j session
+    :param crm_v1: The Compute Resource Manager v1 resource object created by `googleapiclient.discovery.build()`.
```
achantavy: Nit picky reminder: clean this up at some point :-)

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+    """
+    Return list of Google Groups in your organization
+    Returns empty list if we are unable to enumerate the groups for any reasons
+
+    :param admin: apiclient discovery resource object
+    See
+    :return: List of Google groups in domain
+
+    """
+    request = admin.groups().list(customer='my_customer', maxResults=20, orderBy='email')
```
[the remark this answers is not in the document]
skiptomyliu: `my_customer` means the current GSuite account that the key is tied to.  
https://developers.google.com/admin-sdk/directory/v1/guides/manage-customers

Otherwise you need to populate it with the `uid` of your gsuite.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+    groups = []
+    while request is not None:
```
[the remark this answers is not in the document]
skiptomyliu: Believe it's needed because it's a paginated API.  `request` is `None` if there are no results to be returned, which is used by https://github.com/lyft/cartography/blob/d636154174794d9851924c4a8664f62403539cbc/cartography/intel/gsuite/api.py#L29

cartography/intel/gsuite/api.py
skiptomyliu: the `list` in this function is actually different than the `list` in the get groups:

`admin.members()` . 
https://developers.google.com/admin-sdk/directory/v1/reference/members/list

vs

`admin.groups()`
https://developers.google.com/admin-sdk/directory/v1/reference/groups/list

Unfortunately you don't get members from the latter and they're unique API calls.

cartography/intel/gsuite/api.py
skiptomyliu: Yeah I went back and forth on this -- I put this in its own function so it'd be easier to unittest:
https://github.com/lyft/cartography/pull/126/files/d636154174794d9851924c4a8664f62403539cbc#diff-c8a41175b54e8c80d326ce5f8a5e4da4R123

Otherwise, if you modified the query, then you'd need to modify two files (the unit test as well).

I'm ambivalent towards both and I can revert it back into it's own function.

cartography/intel/gsuite/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+        "UPDATE_TAG": config.update_tag,
+    }
+
+    try:
+        # Explicitly use Application Default Credentials.
+        # See https://oauth2client.readthedocs.io/en/latest/source/
+        #             oauth2client.client.html#oauth2client.client.OAuth2Credentials
+        credentials = GoogleCredentials.get_application_default()
+        credentials = credentials.create_scoped(OAUTH_SCOPE)
+        credentials = credentials.create_delegated('cartography-admin@lyft.com')
```
skiptomyliu: `cartography-admin@lyft.com` needs to be extrapolated into a config file.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+
+def load_gsuite_groups(session, groups, gsuite_update_tag):
+    logger.info('Ingesting {} gsuite groups'.format(len(groups)))
+    session.run(get_ingestion_groups_qry(), GroupData=groups, UpdateTag=gsuite_update_tag)
+
+
+def get_ingestion_users_qry():
+    return """
+        UNWIND {UserData} as user
+        MERGE (u:GSuiteUser{id: user.id})
```
achantavy: Is this ID guaranteed to be unique?  What does it look like?

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+        users = users + resp.get('users', [])
+        # break
+        request = admin.users().list_next(request, resp)
+    return users
+
+
+def get_ingestion_groups_qry():
+    return """
+        UNWIND {GroupData} as group
+        MERGE (g:GSuiteGroup{id: group.id})
```
achantavy: What does this ID look like?  Is it guaranteed to be unique across all GSuite?

cartography/intel/gsuite/__init__.py
achantavy: Yeah, or moved to an env var.  Oh and readme needs to be updated :)

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+        ON CREATE SET
```
achantavy: Also, if we decide to use `group.id` as the unique identifier for a GSuiteGroup node, set another attribute on this node to `group_id: group.id`.  This way we know what field uniquely identifies the node, if that makes sense.  I'm kinda rambling here but ping me if this doesn't make sense.

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+def load_gsuite_groups(session, groups, gsuite_update_tag):
+    logger.info('Ingesting {} gsuite groups'.format(len(groups)))
+    session.run(get_ingestion_groups_qry(), GroupData=groups, UpdateTag=gsuite_update_tag)
+
+
+def get_ingestion_users_qry():
+    return """
+        UNWIND {UserData} as user
+        MERGE (u:GSuiteUser{id: user.id})
+        ON CREATE SET
```
achantavy: Same thing with `user_id: user.id`

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def load_gsuite_users(session, users, gsuite_update_tag):
+    logger.info('Ingesting {} gsuite users'.format(len(users)))
+    session.run(get_ingestion_users_qry(), UserData=users, UpdateTag=gsuite_update_tag)
+
+
+def get_ingestion_membership_qry():
+    """
+    MATCH (user:GSuiteUser),(group:GSuiteGroup)
```
achantavy: This match is (probably) not necessary because the MERGE will ensure that the pattern exists.  We can talk more on this later

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,243 @@
...
+    session.run(
+        get_ingestion_membership_qry(),
+        MemberData=members,
+        GroupID=group.get("id"),
+        UpdateTag=gsuite_update_tag,
+    )
+
+    qry = """
+    UNWIND {MemberData} as member
+    MATCH(user: GSuiteGroup{id: member.id}), (group:GSuiteGroup {id: {GroupID}})
```
achantavy: Same here - the match is probably not necessary

cartography/sync.py
```diff
@@ -142,10 +143,11 @@ def build_default_sync():
...
     sync.add_stages([
-        ('create-indexes', cartography.intel.create_indexes.run),
-        ('aws', cartography.intel.aws.start_aws_ingestion),
-        ('gcp', cartography.intel.gcp.start_gcp_ingestion),
-        ('crxcavator', cartography.intel.crxcavator.start_extension_ingestion),
-        ('analysis', cartography.intel.analysis.run),
+        # ('create-indexes', cartography.intel.create_indexes.run),
+        # ('aws', cartography.intel.aws.start_aws_ingestion),
+        # ('gcp', cartography.intel.gcp.start_gcp_ingestion),
+        ('gsuite', cartography.intel.gsuite.start_gsuite_ingestion),
```
skiptomyliu: todo: re-uncomment

cartography/intel/gsuite/api.py
skiptomyliu: the ID generated by Google is a unique id.  It's length 15, lower case alphanumeric.  ex: `o36w3mzh00fkzbq`

achantavy: Cool, that works.

cartography/intel/gsuite/api.py
skiptomyliu: Done (if I understood correctly):
https://github.com/lyft/cartography/blob/c2f475f9ffa9fa91b355da016bf38c8dff2e34d5/cartography/intel/gsuite/api.py#L112

and for users
https://github.com/lyft/cartography/blob/c2f475f9ffa9fa91b355da016bf38c8dff2e34d5/cartography/intel/gsuite/api.py#L131

cartography/intel/gsuite/api.py
skiptomyliu: I can do that, but as a consideration it'll add an additional for loop to transform the objects.

skiptomyliu: Before I apply this to users workflow as well, here's how it looks like for `groups`
https://github.com/lyft/cartography/pull/126/commits/9afebd1a2f28bf02e69067b0e35f3250ff1427b4#diff-732718514a18eeb9183d728497f14713R34-R39

I'm not entirely convinced this would be the best course of action because we're essentially passing around large lists for transformation and adding additional loops.  However, if this is the workflow we want, i'll go ahead and apply.

cartography/intel/gsuite/__init__.py
skiptomyliu: now set to `env` var

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,277 @@
...
+            break
+        response_objects.append(resp)
+        request = admin.users().list_next(request, resp)
+    return response_objects
+
+
+def load_gsuite_groups(session, groups, gsuite_update_tag):
+    ingestion_qry = """
+        UNWIND {GroupData} as group
+        MERGE (g:GSuiteGroup{id: group.id})
```
achantavy: ON CREATE SET --> firstseen

SET ... --> everything else

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,277 @@
...
+        u.given_name = user.name.givenName,
+        u.org_unit_path = user.orgUnitPath,
+        u.primary_email = user.primaryEmail,
+        u.email = user.primaryEmail,
+        u.suspended = user.suspended,
+        u.thumbnail_photo_etag = user.thumbnailPhotoEtag,
+        u.thumbnail_photo_url = user.thumbnailPhotoUrl,
+        u.firstseen = {UpdateTag},
+        u.lastupdated = {UpdateTag}
+        ON MATCH SET
```
achantavy: Ditto

cartography/intel/gsuite/api.py
```diff
@@ -0,0 +1,277 @@
...
+        ingestion_qry,
+        MemberData=members,
+        GroupID=group.get("id"),
+        UpdateTag=gsuite_update_tag,
+    )
+    membership_qry = """
+        UNWIND {MemberData} as member
+        MATCH(group_1: GSuiteGroup{id: member.id}), (group_2:GSuiteGroup {id: {GroupID}})
+        MERGE (group_1)-[r:MEMBER_GSUITE_GROUP]->(group_2)
+        ON CREATE SET
```
achantavy: Ditto

tests/unit/cartography/intel/gsuite/test_api.py
```diff
@@ -0,0 +1,203 @@
...
+        g.admin_created = group.adminCreated,
+        g.description = group.description,
+        g.direct_members_count = group.directMembersCount,
+        g.email = group.email,
+        g.etag = group.etag,
+        g.kind = group.kind,
+        g.name = group.name,
+        g.firstseen = {UpdateTag},
+        g.lastupdated = {UpdateTag}
+        ON MATCH SET
```
achantavy: Same

tests/unit/cartography/intel/gsuite/test_api.py
```diff
@@ -0,0 +1,203 @@
...
+        u.given_name = user.name.givenName,
+        u.org_unit_path = user.orgUnitPath,
+        u.primary_email = user.primaryEmail,
+        u.email = user.primaryEmail,
+        u.suspended = user.suspended,
+        u.thumbnail_photo_etag = user.thumbnailPhotoEtag,
+        u.thumbnail_photo_url = user.thumbnailPhotoUrl,
+        u.firstseen = {UpdateTag},
+        u.lastupdated = {UpdateTag}
+        ON MATCH SET
```
achantavy: Same

PR #128, opened 2019-08-09
Adding new lines to end of certain test files so that certain Drift Detection tests are no longer inconsistent. 

thomashli:


ecdavis:
Rather than doing this in the tests, please add `json_file.write("\n")` to the FileSystem class in the storage module.

PR #130, opened 2019-08-13
Adding ELBv2 Support

scriptsrc:
**Description:**
Adds ELBv2 to Lyft Cartography

This loads ELBv2 into the GraphDB along with associated subnets, security groups, listeners, and ec2 instances.

- ALB’s that are internet accessible are tagged accordingly. (Public IP, 0.0.0.0/0 in SG on listener port).  
- Attached EC2 instances are tagged accordingly.
- Documentation has been updated.
- Passes all pre-commit formatting and linting rules.

**Example Queries:**
```
MATCH (cidr:IpRange)—->(perm:IpPermissionInbound)—->(sg:EC2SecurityGroup)<-[:MEMBER_OF_EC2_SECURITY_GROUP]-(elbv2:LoadBalancerV2)—->(listener:ELBV2Listener)
WHERE elbv2.scheme = "internet-facing" AND cidr.range = "0.0.0.0/0" AND listener.port>=perm.fromport AND listener.port<=perm.toport
RETURN elbv2.dnsname, listener.port
ORDER by elbv2.dnsname, listener.port
```

```
MATCH (elbv2:LoadBalancerV2{exposed_internet: true})-->(listener:ELBV2Listener)
RETURN elbv2.dnsname, listener.port
ORDER BY elbv2.dnsname, listener.port
```

**Potential Next Steps:**
Lambdas can also be attached to an ALB, so we may want to add those into the graphdb and tag them with `exposed_internet`.

I'm not doing anything smart with Network Load Balancers.


scriptsrc:
(Just signed the CLA)

scriptsrc:
I'm not sure why CLA check is failing. I've signed the CLA. Please let me know if I am doing something wrong.

cartography/intel/aws/ec2.py
```diff
@@ -432,6 +472,129 @@ def load_ec2_auto_scaling_groups(session, data, region, current_aws_account_id,
...
+                    'TargetGroupArn': 'string',
+                    ...
+                    'TargetType': 'instance'|'ip'|'lambda',
+                    'Targets': ["i-0f76fade"]
+                ]
+            },
+        ],
+    """
+    ingest_load_balancer_v2 = """
+    MERGE (elbv2:LoadBalancerV2{id: {ID}})
```
sachafaust: Could we leverage the LoadBalancer node and add a version attribute? This would help build queries to look at all load balancer in a simpler query. Another trick could be to use label/node "inheritence" and use

MERGE (elbv2:LoadBalancer:LoadBalancerV2{id: {ID}})

scriptsrc: Hey @sachafaust,

I've been thinking about it and here are the potential rough spots:

1) Classic ELBs have both SOURCE SECURITY GROUPS and SECURITY GROUPS.  The distinction is important for identifying access through a classic ELB.  ALB's have attached security groups, but there's no concept of a source security group.  So any graphql query which relies on `(lb:LoadBalancer)-[SOURCE_SECURITY_GROUP]->(sg:EC2SecurityGroup)` would fail to return the appropriate results on an ALB.

2) This one is probably bigger than #1.  The next PR that would build on this one would improve our visibility around NLBs.  NLB's are also a type of `elbv2`, but with type `network` instead of type `application`.  NLB's don't have any security groups directly attached.  They may have security groups attached to ENI's.  NLBs can be included in AWS PrivateLink, which itself has a kind of resource policy which can allow endpoint connections from different IAM Users, Roles, or AWS accounts. So the queries that will be required to identify NLB's that might be internet accessible, or connected to an internet-connectable privatelink, will be very different from both classic ELBs and ALBs. This makes me think we should treat them as different technologies in GraphQL so we can avoid overly complex edge cases on our queries. (There may be more to learn about NLBs; I'm learning now.)

3) If we treat classic ELB, ALB, and NLB the same, then would we also treat an APIGateway the same?  At some point our GraphQL representation of our AWS environment may start to be difficult to understand because it's oversimplifying the environment. Looking at how the tool has been built out until now, it looks like it's trying to built an accurate model of AWS, leaving simplifications to things like linking `EC2Instances` directly to `LoadBalancers` instead of more accurately connecting them through `ELBListeners`.  I followed this kind of simplification for ALBs. For ALBs the accurate representation would be `(LoadBalancerV2)-->(ELBV2Listener)-->(TargetGroup)-->(EC2Instance)`.

So with all of that information, could I ask you to think about how we should be building this out, not just for ALBs, but also for NLBs and APIGateways, and let me know which direction I should take? Thanks a bunch for this tool.

Unrelated: As the author of `policyuniverse`, I'll try and jump on those issues at some point :).

scriptsrc:
Is `indexes.cypher` only there to improve our query performance?  (Still learning GraphQL)

@sachafaust ^

cartography/intel/aws/ec2.py
scriptsrc: @sachafaust ^ Waiting for your input before we proceed.

scriptsrc: @sachafaust - Should I continue to wait for your thoughts your just proceed with your original guidance?

sachafaust:
> Is `indexes.cypher` only there to improve our query performance? (Still learning GraphQL)
> 
> @sachafaust ^

Yes

cartography/intel/aws/ec2.py
sachafaust: I think it make sense to proceed but this goes beyond my understanding of AWS ELB.  Looping the eng folks on our end. @ecdavis @achantavy I'll defer to you.

ecdavis: I think it's worth keeping them separate for the reasons listed above. Thanks for the in-depth explanation @scriptsrc -- aiming to get this merged soon.

scriptsrc:
`indexes.cypher` has now been updated. I think this PR is ready.

cartography/intel/aws/ec2.py
```diff
@@ -607,6 +786,134 @@ def load_load_balancer_subnets(neo4j_session, load_balancer_id, subnets_data, aw
...
+        },
+    ],
+    """
+    ingest_listener = """
+    MATCH (elbv2:LoadBalancerV2{id: {LoadBalancerId}})
+    WITH elbv2
+    UNWIND {Listeners} as data
+        MERGE (l:Endpoint:ELBV2Listener{id: elbv2.id + toString(data.Port) +
+                toString(data.Protocol)})
+        ON CREATE SET l.port = data.Port, l.protocol = data.Protocol,
```
achantavy: Is `data.protocol` a list or string separated by a character?  I think neo4j supports lists on nodes with https://neo4j.com/docs/cypher-manual/current/functions/list/.

In general, to represent lists I think we should create new nodes and attach to them, but since this is a list of protocols it doesn't make much sense to create new nodes,   so it'd be most intuitive to represent these protocols with neo4j lists.

docs/schema/aws.md
```diff
@@ -946,6 +947,52 @@ Represents an AWS Elastic Load Balancer.  See [spec for details](https://docs.aw
 	(AWSDNSRecord, DNSRecord)-[DNS_POINTS_TO]->(LoadBalancer)
 	```
 
+## LoadBalancerV2
```
achantavy: Please also add entry for `:Endpoint:ELBV2Listener`.

cartography/intel/aws/ec2.py
```diff
@@ -496,6 +536,129 @@ def load_ec2_auto_scaling_groups(neo4j_session, data, region, current_aws_accoun
                 )
 
 
+def load_load_balancer_v2s(neo4j_session, data, region, current_aws_account_id, aws_update_tag):
+    """
+    For each elbv2
+        'LoadBalancers': [
```
achantavy: This comment showing the expected data type is helpful, but can you please add  integration tests for `load_load_balancer_v2s()`, `load_load_balancer_v2_subnets()`,  `load_load_balancer_v2_target_groups()`, and `load_load_balancer_v2_listeners()`?

In general we just need tests for the `load_*` functions and you can use the mock data you've specified here already.  A good place to copy from is here: https://github.com/lyft/cartography/blob/master/tests/integration/cartography/intel/aws/test_dynamodb.py

cartography/intel/aws/ec2.py
scriptsrc: `data.protocol` is a string. It must be **one of** `HTTP`, `HTTPS`, `TCP`, `TLS`, `UDP`, `TCP_UDP`

docs/schema/aws.md
scriptsrc: `:Endpoint:ELBV2Listener` has been added and links to the documentation.

cartography/intel/aws/ec2.py
scriptsrc: Integration tests have been written for all four `load_load_balancer_v2_...` methods.

cartography/intel/aws/ec2.py
```diff
@@ -607,6 +786,134 @@ def load_load_balancer_subnets(neo4j_session, load_balancer_id, subnets_data, aw
...
+                },
+            ]
+        },
+    ],
+    """
+    ingest_listener = """
+    MATCH (elbv2:LoadBalancerV2{id: {LoadBalancerId}})
+    WITH elbv2
+    UNWIND {Listeners} as data
+        MERGE (l:Endpoint:ELBV2Listener{id: elbv2.id + toString(data.Port) +
```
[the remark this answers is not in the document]
scriptsrc: I've replaced the this part with the listener ARN.
Listeners don't have a name.

scriptsrc:
FYI - Travis is failing because of an apt-get issue:

```
WARNING: The following packages cannot be authenticated!
  cypher-shell neo4j
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
apt-get.diagnostics
apt-get install failed
...
W: GPG error: https://debian.neo4j.org/repo stable/ Release: The following signatures were invalid: BADSIG 26C95CF201182252 Neo Technology Admins <admins@neotechnology.com>  The following signatures were invalid: KEYEXPIRED 1572135620
W: The repository 'https://debian.neo4j.org/repo stable/ Release' is not signed.
```

I'm thoroughly embarrassed that I don't know why the commit history on this PR is so crazy. It's 2:30a so maybe I just need to sleep on it. I'll take a look tomorrow.

I wrote the integration tests and made the other fixes requested by @achantavy.  Thank you for your review.

achantavy:
It looks like Neo4j's signing key has expired: https://github.com/neo4j/neo4j/issues/12326, https://stackoverflow.com/questions/58591012/is-neo4j-gpg-key-expired, https://superuser.com/questions/1496370/neo4j-apt-key-refused

scriptsrc:
Hey @achantavy - Thanks to some awesome `git` help from @mikegrima at Netflix, I was able to cleanup the commit history on this PR. Build is still failing on the expired Neo4j signing key.

scriptsrc:
Hey @achantavy and @sachafaust - this should be totally good to go now.

achantavy:
@scriptsrc looking at this today, sorry for delays!

cartography/intel/aws/ec2.py
```diff
@@ -496,6 +536,74 @@ def load_ec2_auto_scaling_groups(neo4j_session, data, region, current_aws_accoun
...
+    for lb in data['LoadBalancers']:
+        load_balancer_id = lb["DNSName"]
+
+        neo4j_session.run(
+            ingest_load_balancer_v2,
+            ID=load_balancer_id,
+            CREATED_TIME=str(lb["CreatedTime"]),
+            NAME=lb["LoadBalancerName"],
+            DNS_NAME=load_balancer_id,
+            HOSTED_ZONE_NAME_ID=lb.get("CanonicalHostedZoneNameID", ""),
```
achantavy: I'm in favor of changing this to 
`HOSTED_ZONE_NAME_ID=lb.get('CanonicalHostedZoneNameID')` so that it returns None instead of empty string if the field does not exist.  The neo4j python driver knows how to deal with None values: it simply does not add the field to the node.

Same comment applies to all the other empty string assignments below this one.

cartography/intel/aws/ec2.py
```diff
@@ -590,6 +698,27 @@ def load_load_balancers(neo4j_session, data, region, current_aws_account_id, aws
             load_load_balancer_listeners(neo4j_session, load_balancer_id, lb["ListenerDescriptions"], aws_update_tag)
 
 
+def load_load_balancer_v2_subnets(neo4j_session, load_balancer_id, az_data, region, aws_update_tag):
+    ingest_load_balancer_subnet = """
+    MATCH (elbv2:LoadBalancerV2{id: {ID}})
+    MERGE (subnet:EC2Subnet{subnetid: {SubnetId}})
```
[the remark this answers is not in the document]
scriptsrc: I modeled the elbv2 subnet method after the classic elb implementation right below it:

```
def load_load_balancer_v2_subnets(neo4j_session, load_balancer_id, az_data, region, aws_update_tag):
    ingest_load_balancer_subnet = """
    MATCH (elbv2:LoadBalancerV2{id: {ID}})
    MERGE (subnet:EC2Subnet{subnetid: {SubnetId}})
```
vs
```
def load_load_balancer_subnets(neo4j_session, load_balancer_id, subnets_data, aws_update_tag):
    ingest_load_balancer_subnet = """
    MATCH (elb:LoadBalancer{id: {ID}}), (subnet:EC2Subnet{subnetid: {SUBNET_ID}})
```

Renaming subnetid to `id` would require updates to a number of places. (I did not create `EC2Subnet`; I'm simply using it consistent with other uses.)

If the `subnetid` vs `id` bothers you, I suggest we attack that in a separate PR so this one can stay focused on `elbv2`. BTW - does graphql have migration scripts the way flask and Django apps have them when using a relational database? (I haven't seen any.)

cartography/intel/aws/ec2.py
scriptsrc: Done :).

achantavy:
Thank you @scriptsrc!

cartography/intel/aws/ec2.py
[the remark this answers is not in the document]
scriptsrc: Renaming fields in flask apps that rely on say MySQL requires that you create a migration script to rename the table column and make any adjustments. Flask-Migrate is a tool that helps auto-create these scripts - https://flask-migrate.readthedocs.io/en/latest/

As the task here is to rename a database field from `subnetid` to `id`, I was wondering if we would have to write a script somewhere to find all the existing entries in graphql and update them.

I think the answer is no. I think the cleanup scripts in cartography ensure that the graphql db only has current items so there are no existing items that need to be updated. (The DB is re-built every time. )

PR #137, opened 2019-08-20
Add Okta intel module.

sachafaust:


cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+from okta.framework.OktaError import OktaError
+from okta.framework.PagedResults import PagedResults
+from okta.models.usergroup import UserGroup
+
+from cartography.config import Config
+from cartography.util import run_cleanup_job
+
+logger = logging.getLogger(__name__)
+
+OKTA_API_KEY = os.environ.get('CREDENTIALS_OKTA_API_KEY')
```
SecPrez: Where is the documentation about the need for this env?

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+    user_props["second_email"] = okta_user.profile.secondEmail
+    user_props["mobile_phone"] = okta_user.profile.mobilePhone
+
+    # https://github.com/okta/okta-sdk-python/blob/master/okta/models/user/User.py
+    user_props["id"] = okta_user.id
+    user_props["created"] = okta_user.created.strftime("%m/%d/%Y, %H:%M:%S")
+    if okta_user.activated:
+        user_props["activated"] = okta_user.activated.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        user_props["activated"] = None
```
achantavy: I don't think explicitly setting this to None is necessary

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_user.statusChanged:
+        user_props["status_changed"] = okta_user.statusChanged.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        user_props["status_changed"] = None
```
achantavy: Ditto

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_user.lastLogin:
+        user_props["last_login"] = okta_user.lastLogin.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        user_props["last_login"] = None
```
achantavy: Ditto

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_user.lastUpdated:
+        user_props["okta_last_updated"] = okta_user.lastUpdated.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        user_props["okta_last_updated"] = None
```
achantavy: Ditto

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_user.passwordChanged:
+        user_props["password_changed"] = okta_user.passwordChanged.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        user_props["password_changed"] = None
```
achantavy: Ditto

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_user.transitioningToStatus:
+        user_props["transition_to_status"] = okta_user.transitioningToStatus
+    else:
+        user_props["transition_to_status"] = None
```
achantavy: Ditto

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+    new_user.okta_last_updated = user_data.okta_last_updated,
+    new_user.password_changed = user_data.password_changed,
+    new_user.transition_to_status = user_data.transition_to_status,
+    new_user.lastupdated = {okta_update_tag}
+    WITH new_user, org
+    MERGE (org)-[org_r:RESOURCE]->(new_user)
+    ON CREATE SET org_r.firstseen = timestamp()
+    SET org_r.lastupdated = {okta_update_tag}
+    WITH new_user
+    MERGE (h:Human{email: new_user.email})
```
achantavy: Human is not an open-sourced node.  We should refactor this out and use internally as an analysis job.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+    """
+    # https://github.com/okta/okta-sdk-python/blob/master/okta/models/usergroup/UserGroup.py
+    group_props = {}
+    group_props["id"] = okta_group.id
+    group_props["name"] = okta_group.profile.name
+    group_props["description"] = okta_group.profile.description
+    if okta_group.profile.samAccountName:
+        group_props["sam_account_name"] = okta_group.profile.samAccountName
+    else:
+        group_props["sam_account_name"] = None
```
achantavy: Not necessary

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_group.profile.dn:
+        group_props["dn"] = okta_group.profile.dn
+    else:
+        group_props["dn"] = None
```
achantavy: Same

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_group.profile.windowsDomainQualifiedName:
+        group_props["windows_domain_qualified_name"] = okta_group.profile.windowsDomainQualifiedName
+    else:
+        group_props["windows_domain_qualified_name"] = None
```
achantavy: Same

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_group.profile.externalId:
+        group_props["external_id"] = okta_group.profile.externalId
+    else:
+        group_props["external_id"] = None
```
achantavy: Same

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    neo4j_session.run(
+        ingest_statement,
+        ORG_ID=okta_org_id,
+        GROUP_LIST=group_list,
+        okta_update_tag=okta_update_tag,
+    )
+
+
+def _get_okta_groups_id_from_graph(neo4j_session, okta_org_id):
```
achantavy: Rather than query the graph for data we had just loaded into it, I think we should just use the groups object returned from a previous call.  This will save IO and network calls.  

Instead, I recommend using the `attach` pattern used here: https://github.com/lyft/cartography/blob/master/cartography/intel/aws/rds.py#L108

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+    group_query = "MATCH (:OktaOrganization{id: {ORG_ID}})-[:RESOURCE]->(group:OktaGroup) return group.id as id"
+
+    result = neo4j_session.run(group_query, ORG_ID=okta_org_id)
+
+    groups = [r['id'] for r in result]
+
+    return groups
+
+
+def _get_okta_application_id_from_graph(neo4j_session, okta_org_id):
```
achantavy: Same comment about using the `attach` pattern from https://github.com/lyft/cartography/blob/master/cartography/intel/aws/rds.py#L108

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+def transform_okta_application(okta_application):
+    # https://github.com/okta/okta-sdk-python/blob/master/okta/models/app/AppInstance.py
+    app_props = {}
+    app_props["id"] = okta_application.id
+    app_props["name"] = okta_application.name
+    app_props["label"] = okta_application.label
+    if okta_application.created:
+        app_props["created"] = okta_application.created.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        app_props["created"] = None
```
achantavy: Not needed

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_application.lastUpdated:
+        app_props["okta_last_updated"] = okta_application.lastUpdated.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        app_props["okta_last_updated"] = None
```
achantavy: Same

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    app_props["status"] = okta_application.status
+
+    if okta_application.activated:
+        app_props["activated"] = okta_application.activated.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        app_props["activated"] = None
```
achantavy: Same

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    neo4j_session.run(
+        ingest,
+        APP_ID=app_id,
+        GROUP_LIST=group_list,
+        okta_update_tag=okta_update_tag,
+    )
+
+
+def _get_user_id_from_graph(neo4j_session, okta_org_id):
```
achantavy: Same comment about using the `attach` pattern: https://github.com/lyft/cartography/blob/master/cartography/intel/aws/rds.py#L108

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+    # https://github.com/okta/okta-sdk-python/blob/master/okta/models/factor/Factor.py
+    factor_props = {}
+    factor_props["id"] = okta_factor_info.id
+    factor_props["factor_type"] = okta_factor_info.factorType
+    factor_props["provider"] = okta_factor_info.provider
+    factor_props["status"] = okta_factor_info.status
+    if okta_factor_info.created:
+        factor_props["created"] = okta_factor_info.created.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        factor_props["created"] = None
```
achantavy: Not needed

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+
+    if okta_factor_info.lastUpdated:
+        factor_props["okta_last_updated"] = okta_factor_info.lastUpdated.strftime("%m/%d/%Y, %H:%M:%S")
+    else:
+        factor_props["okta_last_updated"] = None
```
achantavy: Same

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+        role_props["label"] = role["label"]
+        role_props["type"] = role["type"]
+        role_props["id"] = "{}-{}".format(okta_org_id, role["type"])
+
+        user_roles.append(role_props)
+
+    return user_roles
+
+
+# TODO - Make this one fail gracefuly with warning as it requires super admin token
```
achantavy: Can you elaborate a bit on this?

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1225 @@
...
+    _sync_okta_groups(neo4j_session, okta_organization, config.update_tag)
+    _sync_okta_applications(session, okta_organization, last_update)
+    _sync_users_factors(session, okta_organization, last_update)
+    _sync_trusted_origins(session, okta_organization, last_update)
+
+    # need creds with permission
+    # soft fail as some won't be able to get such high priv token
+    try:
+        _sync_roles(session, okta_organization, last_update)
+    except Exception as exception:
```
achantavy: Recommend against catching the base exception - keep this scoped to only exceptions that we know will happen.  What is the name of exception if the token has insufficient privs?

docs/schema/okta.md
```diff
@@ -0,0 +1,238 @@
...
+ - OktaUser is a resource of an OktaOrganization
+    ```
+    (OktaUser)<-[RESOURCE]->(OkOrganization)
+    ```
+ - OktaUser is assigned OktaApplication
+
+    ```
+    (OktaUser)-[APPLICATION]->(OktaApplication)
+    ```
+ - OktaUser is an identity for a Human
```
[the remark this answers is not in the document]
sachafaust: It does now with this module

cartography/cli.py
```diff
@@ -124,6 +124,14 @@ def _build_parser(self):
                 'jobs are executed.'
             ),
         )
+        parser.add_argument(
```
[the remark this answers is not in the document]
sachafaust: It's needed in case of Okta, there's no way I could find to get that information using the API Token.

cartography/intel/okta/oktaintel.py
sachafaust: This is to make sure the property is set and avoids checking in the future and have it in the graph object

cartography/intel/okta/oktaintel.py
sachafaust: we are introducing it as part of this module. The idea is to connect Human node to other pieces of data in the future ( other PR coming).

cartography/intel/okta/oktaintel.py
sachafaust: Addressed the todo already and removed it.

Basically, to get admin role you need full admin read/write token which most won't be able to get. That section of the intel module will fail gracefully with a warning vs error.

cartography/cli.py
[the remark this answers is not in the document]
sachafaust: @ecdavis @achantavy what is the recommended way to pass data to intel modules beyond credentials? Should we all use env variable, cli param or case by case?

cartography/intel/okta/oktaintel.py
sachafaust: Removed the call when we already have the group list. The function is needed when we map roles and applications.

cartography/intel/okta/oktaintel.py
sachafaust: removed

cartography/intel/okta/oktaintel.py
sachafaust: Needed to fetch user factors and roles.

cartography/intel/okta/oktaintel.py
sachafaust: fixed

cartography/cli.py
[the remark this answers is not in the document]
achantavy: I think that CLI parameters should control settings that are in common for _all_ syncs.  Granted, `--aws-sync-all-profiles` is a bit of an exception, but I think that it would be unwieldy to expose params for each intel module through the CLI.  For this reason I think that it would be best to use environment variables.  

What do you think @ecdavis?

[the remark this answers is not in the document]
sachafaust: @ecdavis what's your take?

README.md
```diff
@@ -60,8 +60,13 @@ Time to set up the server that will run Cartography.  Cartography _should_ work
         1. `GSUITE_GOOGLE_APPLICATION_CREDENTIALS` - location of the credentials file.
         2. `GSUITE_DELEGATED_ADMIN` - email address that you created in step 2
 
+6. If you're using Okta intel module, **prepare your Okta API token**
+    1. Generate your API token by following the steps from [Okta Create An API Token documentation](https://developer.okta.com/docs/guides/create-an-api-token/overview/)
+    2. Populate the OKTA_API_KEY environment variable with the API token
+    3. Populate the OKTA_ORG_ID environment variable with the organization id you want to query. The organization id is the first part of the Okta url for your organization.
+       Example org id is **lyft** for https://lyft.okta.com
```
ecdavis: I don't think this line should be included in the general documentation.

cartography/config.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         neo4j_password=None,
         update_tag=None,
         aws_sync_all_profiles=False,
-        analysis_job_directory=None,
+        analysis_job_directory=None
```
ecdavis: Please don't remove this trailing comma.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+from okta.models.usergroup import UserGroup
+
+from cartography.config import Config
+from cartography.util import run_cleanup_job
+
+logger = logging.getLogger(__name__)
+
+OKTA_API_KEY = os.environ.get('CREDENTIALS_OKTA_API_KEY')
+
+OKTA_ORG_ID = os.environ.get("OKTA_ORG_ID")
```
ecdavis: These pieces of configuration should not be read directly from the environment like this. They should be added to the `Config` class, which is initialized by cartography and passed to all intel modules. There are currently no docs covering this -- we can sync on it in person.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
+# Okta intel module
```
ecdavis: This module is far too large. Please decompose it into multiple modules.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+    """
+    Get Okta users from Okta server
+    :param user_client: user client
+    :return: Array of dictionary containing user properties
+    """
+    user_list = []
+    paged_users = user_client.get_paged_users()
+    while True:
+        for current_user in paged_users.result:
+            user_props = transform_okta_user(current_user)
```
ecdavis: Please refactor this function such that it does not call transform directly, rather it should return the raw results from the API which can then be passed to the transform function by the caller. This is the pattern that is used through the rest of the codebase and it was selected to make the code easier to test and debug.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+
+def _get_okta_users(user_client):
+    """
+    Get Okta users from Okta server
+    :param user_client: user client
+    :return: Array of dictionary containing user properties
+    """
+    user_list = []
+    paged_users = user_client.get_paged_users()
+    while True:
```
ecdavis: Avoid `while True:` if possible.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+    Sync okta users
+    :param neo4j_session: Session with Neo4j server
+    :param okta_org_id: Okta organization id to sync
+    :param okta_update_tag: The timestamp value to set our new Neo4j resources with
+    :return: Nothing
+    """
+
+    logger.debug("Syncing Okta users")
+    user_client = _create_user_client(okta_org_id)
+    data = _get_okta_users(user_client)
```
ecdavis: As mentioned above, please follow the pattern established in other intel modules where the raw data is retrieved from the API, then passed to transform by the sync function, then passed to the load function.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+            # https://developer.okta.com/docs/reference/api/groups/#list-groups
+            if next_url:
+                paged_response = api_client.get(next_url)
+            else:
+                params = {
+                    'limit': 10000,
+                }
+                paged_response = api_client.get_path('/', params)
+        except OktaError as okta_error:
+            logger.debug(f"Got error while going through list group {okta_error}")
```
ecdavis: Why is this a debug level log message? Doesn't this indicate a complete failure to sync Okta?

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+            break
```
ecdavis: Why break here rather than raise?

cartography/intel/okta/oktaintel.py
ecdavis: This applies to all the _get functions below as well.

cartography/intel/okta/oktaintel.py
ecdavis: This applies to all the _get functions below as well.

cartography/intel/okta/oktaintel.py
ecdavis: This applies to all the _sync functions below as well.

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+    :return: Nothing
+    """
+
+    for group_info in group_list_info:
+        group_id = group_info["id"]
+        members = _get_okta_group_members(api_client, group_id)
+        _ingest_okta_group_members(neo4j_session, group_id, members, okta_update_tag)
+
+
+def _ingest_okta_group_members(neo4j_session, group_id, member_list, okta_update_tag):
```
ecdavis: Not clear why this is broken into two functions?

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+def transform_application_users(json_app_data):
+    """
+    Transform application users data for graph consumption
+    :param json_app_data: raw json application data
+    :return: individual user id as yield
+    """
+
+    app_data = json.loads(json_app_data)
+    for user in app_data:
+        yield user["id"]
```
ecdavis: Why is yield being used here? Is it necessary for this to be a generator function?

cartography/intel/okta/oktaintel.py
```diff
@@ -0,0 +1,1193 @@
...
+
+        if not _is_last_page(paged_response):
+            next_url = paged_response.links.get("next").get("url")
+        else:
+            break
+
+    return app_groups
+
+
+def transform_applicationg_assigned_groups(json_app_data):
```
ecdavis: Typo in this function name.

cartography/intel/okta/oktaintel.py
ecdavis: This applies to other uses of yield in this module as well.

cartography/intel/okta/oktaintel.py
ecdavis: Also please stick to the organizational pattern used by the other intel modules (generally - API interaction at the top, transform below it, load below that, cleanup below that, and sync at the bottom) as it makes it much easier to navigate and understand the full module.

As is, this is incredibly difficult to review in full.

tests/data/okta/userfactors.py
```diff
@@ -0,0 +1,16 @@
+from datetime import datetime
+
+from okta.models.factor.Factor import Factor
+
+
+def create_test_factor():
+    factor = Factor()
+
+    factor.id = "jkljsdf"
```
ecdavis: Please use real values for these fields. When tests fail it should be easy to understand what the issue is. Failures like `None != "jkljsdf"` will not be helpful.

tests/data/okta/userfactors.py
```diff
@@ -0,0 +1,16 @@
...
+    factor.factorType = "kljdslkfj"
+    factor.provider = "lkjsdflkjsdf"
+    factor.status = "sdf"
+    factor.created = datetime.now()
```
ecdavis: Hardcode these datetime objects so they stay the same across test runs. We've used midnight on January 1 2019 in other places.

tests/data/okta/users.py
```diff
@@ -0,0 +1,29 @@
...
+    user = User()
+
+    user.id = "userid"
+    user.activated = datetime.now()
+    user.created = datetime.now()
+    user.activated = datetime.now()
+    user.statusChanged = datetime.now()
+    user.lastLogin = datetime.now()
+    user.lastUpdated = datetime.now()
+    user.passwordChanged = datetime.now()
```
ecdavis: See comment above on datetimes. Please hardcode these.

tests/unit/cartography/intel/okta/test_admin_roles.py
```diff
@@ -0,0 +1,32 @@
+from cartography.intel.okta.oktaintel import transform_group_roles_data
+from cartography.intel.okta.oktaintel import transform_user_roles_data
+from tests.data.okta.adminroles import LIST_ASSIGNED_GROUP_ROLE_RESPONSE
+from tests.data.okta.adminroles import LIST_ASSIGNED_USER_ROLE_RESPONSE
+
+
+def test_transform_user_roles():
+    org_id = "lyft"
```
ecdavis: Please use something generic like "example-org".

tests/unit/cartography/intel/okta/test_admin_roles.py
```diff
@@ -0,0 +1,32 @@
...
+    props_list = transform_user_roles_data(LIST_ASSIGNED_USER_ROLE_RESPONSE, org_id)
+
+    assert len(props_list) == 2
+    assert props_list[0]["label"] == "Application Administrator"
+    assert props_list[0]["type"] == "APP_ADMIN"
+    assert props_list[0]["id"] == "{}-{}".format(org_id, props_list[0]["type"])
```
ecdavis: Hardcode what we expect the result to be. Duplicating the application logic increases the likelihood of including a bug in the test.

tests/unit/cartography/intel/okta/test_admin_roles.py
```diff
@@ -0,0 +1,32 @@
...
+
+    assert props_list[1]["label"] == "Help Desk Administrator"
+    assert props_list[1]["type"] == "HELP_DESK_ADMIN"
+    assert props_list[1]["id"] == "{}-{}".format(org_id, props_list[1]["type"])
```
ecdavis: As above, hardcode.

tests/unit/cartography/intel/okta/test_admin_roles.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_transform_group_roles():
+    org_id = "lyft"
+    props_list = transform_group_roles_data(LIST_ASSIGNED_GROUP_ROLE_RESPONSE, org_id)
+
+    assert len(props_list) == 2
+    assert props_list[0]["label"] == "Application Administrator"
+    assert props_list[0]["type"] == "APP_ADMIN"
+    assert props_list[0]["id"] == "{}-{}".format(org_id, props_list[0]["type"])
```
ecdavis: As above, hardcode.

tests/unit/cartography/intel/okta/test_admin_roles.py
```diff
@@ -0,0 +1,32 @@
...
+
+    assert props_list[1]["label"] == "Help Desk Administrator"
+    assert props_list[1]["type"] == "HELP_DESK_ADMIN"
+    assert props_list[1]["id"] == "{}-{}".format(org_id, props_list[1]["type"])
+
+
+def test_transform_group_roles():
+    org_id = "lyft"
```
ecdavis: As above, org-id.

tests/unit/cartography/intel/okta/test_admin_roles.py
```diff
@@ -0,0 +1,32 @@
...
+    props_list = transform_group_roles_data(LIST_ASSIGNED_GROUP_ROLE_RESPONSE, org_id)
+
+    assert len(props_list) == 2
+    assert props_list[0]["label"] == "Application Administrator"
+    assert props_list[0]["type"] == "APP_ADMIN"
+    assert props_list[0]["id"] == "{}-{}".format(org_id, props_list[0]["type"])
+
+    assert props_list[1]["label"] == "Help Desk Administrator"
+    assert props_list[1]["type"] == "HELP_DESK_ADMIN"
+    assert props_list[1]["id"] == "{}-{}".format(org_id, props_list[1]["type"])
```
ecdavis: As above, hardcode.

tests/unit/cartography/intel/okta/test_application.py
```diff
@@ -0,0 +1,93 @@
...
+
+def test_application_with_all_values():
+    app = create_test_application()
+
+    transformed_data = transform_okta_application(app)
+
+    assert transformed_data["id"] == app.id
+    assert transformed_data["name"] == app.name
+    assert transformed_data["label"] == app.label
+    assert transformed_data["created"] == app.created.strftime("%m/%d/%Y, %H:%M:%S")
```
ecdavis: As above, hardcode. This applies to the other dates in this module as well.

tests/unit/cartography/intel/okta/test_factor.py
```diff
@@ -0,0 +1,43 @@
...
+def test_factor_transform_with_all_values():
+    factor = create_test_factor()
+
+    transform_props = transform_okta_user_factor(factor)
+
+    assert transform_props["id"] == factor.id
+    assert transform_props["factor_type"] == factor.factorType
+    assert transform_props["provider"] == factor.provider
+    assert transform_props["status"] == factor.status
+    assert transform_props["created"] == factor.created.strftime("%m/%d/%Y, %H:%M:%S")
```
ecdavis: As above, hardcode. This applies to the other dates in this module as well.

tests/unit/cartography/intel/okta/test_user.py
```diff
@@ -0,0 +1,155 @@
...
+from tests.data.okta.users import create_test_user
+
+
+def test_user_transform_with_all_values():
+    user = create_test_user()
+
+    transform_props = transform_okta_user(user)
+
+    assert transform_props["id"] == user.id
+    assert transform_props["activated"] == user.activated.strftime("%m/%d/%Y, %H:%M:%S")
```
ecdavis: As above, hardcode. This applies to the other dates in this file as well.

tests/unit/cartography/intel/okta/test_admin_roles.py
```diff
@@ -0,0 +1,32 @@
...
+from cartography.intel.okta.oktaintel import transform_user_roles_data
+from tests.data.okta.adminroles import LIST_ASSIGNED_GROUP_ROLE_RESPONSE
+from tests.data.okta.adminroles import LIST_ASSIGNED_USER_ROLE_RESPONSE
+
+
+def test_transform_user_roles():
+    org_id = "lyft"
+    props_list = transform_user_roles_data(LIST_ASSIGNED_USER_ROLE_RESPONSE, org_id)
+
+    assert len(props_list) == 2
```
ecdavis: Rather than using 6 separate assertions, you can use a single assert to compare complex data types. E.g. `assert actual == [{'foo': 'bar'}, 1, 2, 3]` and pytest will figure out where mismatches exist and fail the assertion accordingly. This works reliably with hashable data types, so all the basic Python data types will work great. I recommend replacing the lengthy lists of asserts through these test modules with structured comparisons like this.

ecdavis: Not going to comment on every assert, but the above is feedback for all test files.

cartography/intel/okta/factors.py
```diff
@@ -0,0 +1,140 @@
...
+    :param okta_org_id: okta organization id
+    :param okta_update_tag: The timestamp value to set our new Neo4j resources with
+    :param okta_api_key: Okta API key
+    :return: Nothing
+    """
+
+    logger.debug("Syncing Okta User Factors")
+
+    factor_client = _create_factor_client(okta_org_id, okta_api_key)
+    users = get_user_id_from_graph(neo4j_session, okta_org_id)
```
achantavy: As I mentioned in a previous pass of this PR, we already have the user IDs in memory here: https://github.com/lyft/cartography/blob/952620fdf3ff15c6d8a3872227ff6c32253f9826/cartography/intel/okta/users.py#L105 and I think it is a waste of a round trip to re-query the graph.  

This comment also applies to querying users [here](https://github.com/lyft/cartography/blob/952620fdf3ff15c6d8a3872227ff6c32253f9826/cartography/intel/okta/roles.py#L153) and groups [here](https://github.com/lyft/cartography/blob/952620fdf3ff15c6d8a3872227ff6c32253f9826/cartography/intel/okta/roles.py#L162).

The solution I proposed before was to use our [`_attach()` pattern](https://github.com/lyft/cartography/blob/9da67c69389e1a6d4555a5efcfe3e3120fa26460/docs/writing-intel-modules.md#connecting-different-node-types-with-the-_attach-pattern).  However, this would be a significant refactor as we would need to move FactorClient around, and we would also need to do similar work for roles.  I recognize though that Okta's API around Factors does not list them, requiring us to provide a user ID (which is somewhat cumbersome for us).

I think we should remove all of the `get_*_from_graph()` functions because we already have this data in memory at various points in the code.  We should refactor to pass pointers to this data where it is needed and avoid incurring extraneous RTT hits.

I'll defer to others on the team if they decide that this isn't that big of an issue though.

cartography/intel/okta/roles.py
```diff
@@ -0,0 +1,168 @@
...
+
+    neo4j_session.run(
+        ingest,
+        GROUP_ID=group_id,
+        ROLES_DATA=roles_data,
+        okta_update_tag=okta_update_tag,
+    )
+
+
+def sync_roles(neo4j_session, okta_org_id, okta_update_tag, okta_api_key):
```
achantavy: See above comments on `get_*_from_graph_functions()`

sachafaust:
@achantavy addressed feedback. Please approve

cartography/intel/okta/groups.py
```diff
@@ -82,11 +82,13 @@ def _get_okta_group_members(api_client, group_id):
 
 def transform_okta_group_list(okta_group_list):
     groups = []
+    groups_id = []
 
     for current in okta_group_list:
         groups.append(transform_okta_group(current))
+        groups_id.append(current.id)
```
achantavy: Correct me if I'm wrong, but it seems that the group ID already exists here: https://github.com/lyft/cartography/blob/6954bc838dc5c2a3d20e0d0c4fcb02d21395b6aa/cartography/intel/okta/groups.py#L102

Is this additional `groups_id` list still necessary?

cartography/intel/okta/groups.py
```diff
@@ -228,20 +230,25 @@ def _sync_okta_group_membership(neo4j_session, api_client, group_list_info, okta
...
     :param okta_update_tag: The timestamp value to set our new Neo4j resources with
     :param okta_api_key: Okta API key
+    :param sync_state: Okta sync state
     :return: Nothing
     """
     logger.debug("Syncing Okta groups")
     api_client = create_api_client(okta_org_id, "/api/v1/groups", okta_api_key)
 
     okta_group_data = _get_okta_groups(api_client)
     group_list_info = transform_okta_group_list(okta_group_data)
```
achantavy: You changed `transform_okta_group_list()` above to return two arrays but this call only uses one of them.  Is this intentional?

I would keep this line the way it is right now.  As I mentioned above, I think `groups_id` isn't necessary because the group ID is already in the other `groups` list.  Please correct me if I got that wrong though.

cartography/intel/okta/sync_state.py
```diff
@@ -0,0 +1,20 @@
...
+    Object used to store sync data to allow each stages to share data. For example, saving user and group list returned
+    by Okta instead of asking the Graph for factor and role stages.
+    :type users: array of string
+    :param users: Array of user id as string. Optional.
+    :type groups: array of string
+    :param groups: Array of group id as string. Optional
+    """
+    def __init__(
+        self,
+        user=None,
```
achantavy: For consistency, let's rename this to `users`.

cartography/intel/okta/users.py
```diff
@@ -43,11 +43,13 @@ def _get_okta_users(user_client):
 
 def transform_okta_user_list(okta_user_list):
     users = []
+    user_ids = []
 
     for current in okta_user_list:
         users.append(transform_okta_user(current))
+        user_ids.append(current.id)
```
achantavy: Each object of the `users` list already has an ID: https://github.com/lyft/cartography/blob/6954bc838dc5c2a3d20e0d0c4fcb02d21395b6aa/cartography/intel/okta/users.py#L72

Is this additional `user_ids` array needed?

cartography/intel/okta/users.py
```diff
@@ -153,19 +155,23 @@ def _load_okta_users(neo4j_session, okta_org_id, user_list, okta_update_tag):
...
     :param okta_api_key: Okta API key
+    :param sync_state: Okta sync state
     :return: Nothing
     """
 
     logger.debug("Syncing Okta users")
     user_client = _create_user_client(okta_org_id, okta_api_key)
     data = _get_okta_users(user_client)
-    users = transform_okta_user_list(data)
+    users_data, user_ids = transform_okta_user_list(data)
```
achantavy: As above, I might have missed something but the `user_ids` array might not be needed because objects in `users_data` already have IDs in them.

achantavy:
Looks like the CI linter is complaining.  To fix, rerun `make test_lint` (or just run `make` to rerun the whole suite if that doesn't work).

cartography/intel/okta/users.py
```diff
@@ -153,19 +155,23 @@ def _load_okta_users(neo4j_session, okta_org_id, user_list, okta_update_tag):
...
 
-    _load_okta_users(neo4j_session, okta_org_id, users, okta_update_tag)
+    _load_okta_users(neo4j_session, okta_org_id, users_data, okta_update_tag)
+
+    # store result for later use
+    sync_state.users = user_ids
```
achantavy: I think this could be replaced with `sync_state.users = users_data`.

cartography/intel/okta/factors.py
```diff
@@ -119,22 +118,22 @@ def _load_user_factors(neo4j_session, user_id, factors, okta_update_tag):
...
     :return: Nothing
     """
 
     logger.debug("Syncing Okta User Factors")
 
     factor_client = _create_factor_client(okta_org_id, okta_api_key)
-    users = get_user_id_from_graph(neo4j_session, okta_org_id)
 
-    for user_id in users:
+    for user_id in sync_state.users:
```
achantavy: Actually, I'm not sure if this code is going to run because `OktaSyncState` doesn't have a `users` field - it has a `_users` field.

cartography/intel/okta/groups.py
```diff
@@ -228,20 +230,25 @@ def _sync_okta_group_membership(neo4j_session, api_client, group_list_info, okta
...
+
+    # store result for later use
+    sync_state.groups = group_list_info
```
achantavy: OktaSyncState doesn't have a groups field - it has a _groups field.

cartography/intel/okta/roles.py
```diff
@@ -135,33 +133,29 @@ def _load_group_role(neo4j_session, group_id, roles_data, okta_update_tag):
...
         user_roles_data = _get_user_roles(api_client, user_id, okta_org_id)
         user_roles = transform_user_roles_data(user_roles_data, okta_org_id)
         if len(user_roles) > 0:
             _load_user_role(neo4j_session, user_id, user_roles, okta_update_tag)
 
-    # groups
-    groups = get_okta_groups_id_from_graph(neo4j_session, okta_org_id)
-
-    for group_id in groups:
+    for group_id in sync_state.groups:
```
achantavy: Ditto on the field name.

cartography/intel/okta/users.py
achantavy: Also `sync_state.users` is undefined; it's currently `_users`.

cartography/intel/okta/groups.py
sachafaust: good catch.

cartography/intel/okta/users.py
sachafaust: it avoid keeping the entire data of each users in memory for no reason. This array can be very large.

cartography/intel/okta/users.py
sachafaust: see comment above. big waste of memory to keep all that data around. We keep it short to id which is all we need,

achantavy: I won't block on this, (and correct me if I'm wrong or missing something big) but I think it's actually more memory to keep track of an additional array.

`users_data` is a list of dicts which each contain an `id` field.  Lists and dict parameters are passed in Python as object reference, so there are no additional copies of the ID being made - we just need to access the reference to the ID, which can save memory instead of creating an additional `user_ids` list.

If perf ends up being a problem in this module then we will revisit and optimize this call in the future.

achantavy:
@ecdavis Mind doing another pass when free?

cartography/data/jobs/cleanup/okta_import_cleanup.json
```diff
@@ -3,61 +3,67 @@
     {
       "query": "MATCH (:OktaOrganization{id: {OKTA_ORG_ID}})-[:RESOURCE]->(n:OktaUser)-[:FACTOR]->(n:OktaUserFactor) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
       "iterative": true,
-      "iterationsize": 100
+      "iterationsize": 100,
+      "__comment__": "Deletate stale OktaUserFactor"
```
achantavy: Please fix this typo.

PR #144, opened 2019-09-03
Add AWS ECR intel module.

ecdavis:
TODO: tests ;) -- please wait until I add them before reviewing


cartography/intel/aws/ecr.py
```diff
@@ -0,0 +1,100 @@
+import logging
+
+from cartography.util import run_cleanup_job
+
+logger = logging.getLogger(__name__)
+
+
+def get_ecr_repositories(boto3_session, region):
```
marco-lancini, line 8: would it be possible to add the `@aws_handle_regions` decorator?
Sample: https://github.com/lyft/cartography/blob/master/cartography/intel/aws/eks.py

cartography/intel/aws/ecr.py
```diff
@@ -0,0 +1,100 @@
...
+    client = boto3_session.client('ecr', region_name=region)
+    paginator = client.get_paginator('describe_repositories')
+    ecr_repositories = []
+    for page in paginator.paginate():
+        ecr_repositories.extend(page['repositories'])
+    return {'repositories': ecr_repositories}
+
+
+def get_ecr_repository_images(boto3_session, region, repository_name):
```
marco-lancini, line 17: would it be possible to add the `@aws_handle_regions` decorator?
Sample: https://github.com/lyft/cartography/blob/master/cartography/intel/aws/eks.py

cartography/intel/aws/ecr.py
achantavy: Oh yeah, good idea.

achantavy:
Replaced by #429

PR #145, opened 2019-09-05
Fix crxcavator api details - issue 142

tayasteere:
Had to change API used to pull extension details since combined report only provided details on last two
Updated to include risk score sub components
Updated tests and docs


cartography/intel/crxcavator/crxcavator.py
```diff
@@ -49,55 +52,142 @@ def call_crxcavator_api(api_and_parameters, crxcavator_api_key, crxcavator_base_
...
+        data = extension.get('data')
+        if not data:
+            logger.warning(f'Could not retrieve details for extension {extension}')
+            continue
+        risk_data = data.get('risk')
+        if not risk_data:
+            risk = {}
+        else:
+            risk = parse_risk(risk_data)
+        webstore = data.get('webstore')
```
ecdavis: Since `webstore` is used below, this needs to be `data.get('webstore', {})`

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -49,55 +52,142 @@ def call_crxcavator_api(api_and_parameters, crxcavator_api_key, crxcavator_base_
...
+            'rating': webstore.get('rating'),
+            'rating_users': webstore.get('rating_users'),
+            'short_description': webstore.get('short_description'),
+            'size': webstore.get('size'),
+            'support_site': webstore.get('support_site'),
+            'users': webstore.get('users'),
+            'website': webstore.get('website'),
+            'type': webstore.get('type'),
+            'price': webstore.get('price'),
+            'report_link': "https://crxcavator.io/report/" + extension_id + "/" + version,
```
ecdavis: Please use string interpolation.

cartography/intel/crxcavator/crxcavator.py
tayasteere: I checked for null immediately after this, but went ahead and changed anyways.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -49,55 +52,139 @@ def call_crxcavator_api(api_and_parameters, crxcavator_api_key, crxcavator_base_
...
+    extensions_details = []
+    for extension in extensions_list:
+        extension_id = extension.get('extension_id')
+        version = extension.get('version')
+        name = extension.get('name')
+        try:
+            details = get_extension_details(crxcavator_api_key, crxcavator_base_url, extension_id, version)
+            if not details:
+                # we only have the name and version from group API, create minimal version
+                logger.info(f"No results returned from report API for extension {extension_id} {version}")
```
ecdavis: This log message is very high volume, can we change it to debug-level, or is it useful to have at info?

tayasteere:
Will change to debug before merge now that I have confirmed from Vatsal
this is expected behavior.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -49,55 +52,139 @@ def call_crxcavator_api(api_and_parameters, crxcavator_api_key, crxcavator_base_
...
+    """
+    Retrieves the detailed information for all the extension_id and version pairs
+    :param crxcavator_api_key: The API key to access the CRXcavator service
+    :param crxcavator_base_url: The URL for the CRXcavator API
+    :param extensions_list: list of dictonary items containing the extension_id and version pairs
+    :return: list containing all metadata for extensions
+    """
+    extensions_details = []
+    for extension in extensions_list:
+        extension_id = extension.get('extension_id')
```
skiptomyliu: should we be changing from `.get` to index referencing?  ex:  `extensions['extension_id']`.   

I'm under the assumption that extensions may not have an an `extension_id` if we're using `.get`

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -49,55 +52,139 @@ def call_crxcavator_api(api_and_parameters, crxcavator_api_key, crxcavator_base_
...
+            'users': webstore.get('users'),
+            'website': webstore.get('website'),
+            'type': webstore.get('type'),
+            'price': webstore.get('price'),
+            'report_link': f"https://crxcavator.io/report/{extension_id}/{version}",
+        })
     return extensions
 
 
+def parse_risk(risk_data):
```
skiptomyliu: This function can probably be refactored to be a bit more general.

```
def get_data(data_dict, key):
    data = data_dict.get(key)
    data_score = 0
    if data:
        data_score = permissions.get('total')
    return data, data_score
```

Thus you'd be able to just call a bunch of lines with:

```
permissions, permissions_score = get_data(risk_data, 'permissions')
...
```

PR #146, opened 2019-09-07
AWS - use account specific region list.

tobli:
The list of enabled regions might differ between AWS accounts. The describe_regions API call will only list regions available for the current account. Therefor don’t call describe_regions only once, for the default account, but rather once per account.


tobli:
An additional reason for this change is that it allows cartography to work with a centralised AWS account only for IAM, that has region listing disabled.
I'm using it for a setup with a default account that only has users and policies that allows assuming roles in other accounts. All non-default AWS accounts use a role_arn and "source_profile = default".

cartography/intel/aws/__init__.py
```diff
@@ -17,9 +17,12 @@
 logger = logging.getLogger(__name__)
 
 
-def _sync_one_account(neo4j_session, boto3_session, account_id, regions, sync_tag, common_job_parameters):
+def _sync_one_account(neo4j_session, boto3_session, account_id, sync_tag, common_job_parameters):
     iam.sync(neo4j_session, boto3_session, account_id, sync_tag, common_job_parameters)
     s3.sync(neo4j_session, boto3_session, account_id, sync_tag, common_job_parameters)
+
+    regions = ec2.get_ec2_regions(boto3_session)
```
achantavy: Let's surround this with the `try` `except` from the original line 93-106 with something like this:

```python
    try:	    
        regions = ec2.get_ec2_regions(boto3_session)	
        logger.debug("Error occurred getting EC2 regions.", exc_info=True)	
        logger.error(	
            (	
                "Failed to retrieve AWS region list, an error occurred: %s. Could not get regions for "	
                "account %s."	
            ),	
            e,
            account_id,
        )	
        return
```

PR #148, opened 2019-09-10
Fix incorrect linking of policies to groups/roles.

tobli:
If a group or role name was duplicated between different AWS account, cartographer could incorrectly setup STS_ASSUMEROLE_ALLOW relationsships.


achantavy:
Finally getting around to testing this interally; sorry about this.

Issue #141, opened 2019-08-26
Feature request: ingest Google Cloud Storage buckets


achantavy:
Would like to ingest [Google Cloud Storage](https://cloud.google.com/storage/) [buckets](https://googleapis.github.io/google-cloud-python/latest/storage/client.html#google.cloud.storage.client.Client.list_buckets) with Cartography.

If anyone has cycles to work on this, contributions are definitely super welcome :-)

manesioz:
@achantavy  I'd like to look into this if no one else is. I'm assuming we would want similar functionality that we currently have for AWS [s3](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/s3.py)? Anything I should be aware of?

achantavy:
@manesioz Thanks for offering to help!  I think a first PR should focus on just retrieving the buckets and connecting them to GCP projects.  Here's some pointers on getting started:

Getting the objects appears to be fairly straightforward.  You first instantiate a storage client like this:
```
storage = googleapiclient.discovery.build('storage', 'v1', credentials=credentials)
```

and then you can call list buckets on a project like this:
```
req = storage.buckets().list(project='some-project-name')
res = req.execute()
```

Then `pprint(res)` ends up looking like this:
```
{'items': [{'etag': 'etag',
            'iamConfiguration': {'bucketPolicyOnly': {'enabled': False},
                                 'uniformBucketLevelAccess': {'enabled': False}},
            'id': 'bucket-id',
            'kind': 'storage#bucket',
            'location': 'US-EAST4',
            'locationType': 'region',
            'metageneration': '1',
            'name': 'bucket-id',
            'projectNumber': '1234',
            'selfLink': 'https://www.googleapis.com/storage/v1/b/bucket-id',
            'storageClass': 'REGIONAL',
            'timeCreated': '2019-02-20T22:17:17.899Z',
            'updated': '2019-02-20T22:17:17.899Z'}],
 'kind': 'storage#buckets'}
```

The easiest way to attack this task would be to copy the patterns from our existing GCP code.  For example:
- Using resource providers: https://github.com/lyft/cartography/blob/master/cartography/intel/gcp/__init__.py#L48
- Using the get + transform + load + cleanup pattern: I would recommend copying VPC ingestion here because it attaches VPCs to projects: https://github.com/lyft/cartography/blob/master/cartography/intel/gcp/compute.py#L873

We'll definitely welcome a rough work-in-progress and be willing to work with you on improving it.

[...]

PR #151, opened 2019-09-12
Ingest Google Cloud Buckets

manesioz:
A [WIP] PR regarding the following issue: [#141](https://github.com/lyft/cartography/issues/141)  
So far I have added methods in cartography/intel/gcp/storage.py for getting bucket meta-data, listing all buckets in a project, and getting IAM Policies for buckets. 

Additionally, the cartography/intel/gcp/__init__.py file instantiates storage resource objects that enable the API calls. 

Questions: 

- Other than the single-to-many relationship  `project_id -> bucket_id` what relationships should we be looking to visualize?  How should the Neo4j integration be approached?  
- In the issue [#141](https://github.com/lyft/cartography/issues/141) I was told that we should use the get + transform + load + cleanup pattern for VPC resources; I am unclear of the use of VPC resources as well as how we use them for `storage` since `compute.networks()` is only a method for Google Compute objects.  

I apologize for the rough PR, this is my first real contribution to any open-source project. 

Edit: I will also add unit/integration tests


achantavy:
> Other than the single-to-many relationship project_id -> bucket_id what relationships should we be looking to visualize? How should the Neo4j integration be approached?

I think we should focus on just this for now.  We can research other relationships and how else to connect these nodes in a follow-up PR.

> In the issue #141 I was told that we should use the get + transform + load + cleanup pattern for VPC resources; I am unclear of the use of VPC resources as well as how we use them for storage since compute.networks() is only a method for Google Compute objects.

Ah, I didn't mean to use VPC nodes themselves - I meant to copy the patterns used in that example to write your new code.

> Edit: I will also add unit/integration tests

This is a good starting point for getting local tests set up if you haven't done this already: https://github.com/lyft/cartography/blob/master/docs/testing.md

>I apologize for the rough PR, this is my first real contribution to any open-source project.

Thank you very much for working on this!  Definitely reach out if you need help and give us feedback on your experience; we're trying to make it easier to contribute to this project.  We have those testing docs I linked above but we haven't publicly documented a lot of patterns and principles yet; so far these take the form of just PR comments haha but we'll get there.

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,114 @@
...
+        elif reason == 'forbidden':
+            logger.debug(
+                ("You do not have storage.bucket.get access to the bucket %s. "
+                 "Full details: %s"), bucket, e, )
+            return None
+        else:
+            raise
+
+
+def list_buckets(storage, project):
```
achantavy: I would rename to get_buckets just so for consistency with the other get* functions.

manesioz: Done! Nice catch, sorry for the delay.

manesioz: I need to review neo4j syntax for the load* function, but I should be fine, thanks! If I have any specific questions I'll reach out within the next few days.

manesioz:
Hey @achantavy , so I was wondering how you thought we should map `project_id` to `bucket_id`? 
Seems like a pretty fundamental relationship but the [api](https://cloud.google.com/storage/docs/json_api/v1/buckets) docs seem to indicate that it only maps `bucket_id` to `project_number`. My guess is that we could parse the results of the `get_gcp_projects()` method, but is there a preferred way?

achantavy:
@manesioz See https://github.com/lyft/cartography/blob/cb8c5ca38e569840245f7487c4593eeacaab3083/cartography/intel/gcp/compute.py#L193 for an example on `transform_*`.  Response objects returned by GCP APIs usually contain an `id` field which you can parse out to get the `project_id`.

I'm glad you brought this up, since your `get_*` function probably needs to return the actual `response` object and not `response.items`.  It's helpful to pass a reference to the full object because we can get the project ID easily from parsing `id`.

Once that is done, you can link the project node to the storage bucket node using `MERGE` in the `load_*` function.  The process would look like [this](https://github.com/lyft/cartography/blob/cb8c5ca38e569840245f7487c4593eeacaab3083/cartography/intel/gcp/compute.py#L451) and [this](https://github.com/lyft/cartography/blob/cb8c5ca38e569840245f7487c4593eeacaab3083/cartography/intel/gcp/compute.py#L466).

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,223 @@
...
+    :param project: The Google Project name that you are retrieving buckets from
+    :rtype: list
+    :return: List of storage objects
+    """
+    try:
+        req = storage.buckets().list(project=project)
+        res = req.execute()
+        all_objects = []
+        all_objects.extend(res.get('items', []))
+        return all_objects
```
achantavy: As I mentioned in my comment, let's try return `res` itself for now so that we have access to `res.id` in the `transform_*` function to get the `project_id`.

manesioz: Hmm.. Forgive me if I'm missing it but there doesn't seem to be any `project_id` that is able to be parsed from the storage resource object. I got the following object:
```json 
{
 "kind": "storage#buckets",
 "items": [
  {
   "kind": "storage#bucket",
   "id": "bucket_name",
   "selfLink": "https://www.googleapis.com/storage/v1/b/bucket_name",
   "projectNumber": "some_num",
   "name": "bucket_name",
   "timeCreated": "some_time",
   "updated": "some_time",
   "metageneration": "1",
   "iamConfiguration": {
    "bucketPolicyOnly": {
     "enabled": false
    },
    "uniformBucketLevelAccess": {
     "enabled": false
    }
   },
   "location": "US",
   "locationType": "multi-region",
   "defaultEventBasedHold": false,
   "storageClass": "STANDARD",
   "etag": "CAE="
  }
 ]
}
```
Am I missing something? Maybe I'm not looking at the documentation right. 
I know the python client library doesn't have this capability (see [this](https://github.com/googleapis/google-cloud-python/issues/9358) issue) but we are authenticating the API differently anyway (using `storage = googleapiclient.discovery.build('storage', 'v1', credentials=credentials)`) . Even in the [issue](https://github.com/lyft/cartography/issues/141) you raised, `project_id` doesn't seem to be present.

achantavy: Good catch!  We didn't include projectnumber in a GCP project at first, but I just submitted https://github.com/lyft/cartography/pull/172 to include it.

After that is merged, you should be able to do a `MERGE` query to connect the project to the storage bucket.

manesioz: Thanks. In the `MERGE` statement, should I create another schema `GCPProjectName`? Since the [current](https://github.com/lyft/cartography/blob/master/docs/schema/gcp.md) one has `id` as `project_id` not `project_number`.

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,223 @@
...
+    :param bucket_res: The return data
+    
+    :rtype: list
+    :return: List of buckets ready for ingestion to Neo4j
+    '''
+    
+    bucket_list = []
+    for b in bucket_res.get('items', []):
+        bucket = {}
+        bucket['etag'] = b.get('etag', '')
```
achantavy: I think we should replace all of the `dict.get('value', '')` calls with `dict.get('value')`.  I'd prefer not to have a field as an empty string ingested to the graph.

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,223 @@
+import logging
+from googleapiclient.discovery import HttpError
+
+from cartography.intel.gcp import compute
+logger = logging.getLogger(__name__)
+
+
+def get_gcp_bucket_metadata(storage, bucket):
```
achantavy: I think this is super helpful but not a must for a first PR.  Let's just get the buckets first and then drill down in a second PR.

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,223 @@
...
+        elif reason == 'forbidden':
+            logger.debug(
+                ("You do not have storage.bucket.list access to the project %s. "
+                 "Full details: %s"), project, e, )
+            return None
+        else:
+            raise
+
+
+def get_gcp_bucket_iam_policy(storage, bucket):
```
achantavy: Same comment here about super nice to have, but not necessary in a first PR.

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,223 @@
...
+    :param buckets: List of GCP Buckets to injest 
+    
+    :type gcp_update_tag: timestamp
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    
+    :rtype: NoneType
+    :return: Nothing 
+    '''
+    
+    query = """
```
achantavy: Might be tricky to handle `labels` here :-).  I _think_ we can do a list type on the node: https://neo4j.com/docs/cypher-manual/current/functions/list/

cartography/intel/gcp/storage.py
manesioz: Will do!

cartography/intel/gcp/storage.py
manesioz: [Seems like](https://stackoverflow.com/a/58229056/10981911) you can pass them as any other parameter. Also, should probably start migrating our neo4j parameter syntax from `{param}` to `$param` since the former is [depreciated](https://neo4j.com/docs/cypher-manual/current/syntax/parameters/), but that should probably be a topic for another PR.

cartography/intel/gcp/storage.py
manesioz: Good call, done.

cartography/intel/gcp/storage.py
achantavy: > You can pass them as any other parameter

Sounds good.

> Also, we should probably start migrating our neo4j parameter syntax from {param} to $param since the former is depreciated, but that should probably be a topic for another PR.

Boo :|  Thanks for pointing that out.

> Also, there seems to be multiple instances of nested SET clauses, but I thought that wasn't allowed?

`ON CREATE SET` and `SET` are two separate clauses.  We call `ON CREATE SET` when making the `firstseen` attribute, and we call `SET` when updating the other fields of a node.

achantavy:
It looks like you've figured out most if not all of this by now, but thought I'd let you know that I added a guide on how to make intel modules anyway :P 

https://github.com/lyft/cartography/blob/3e1dcb36679d69119e8bf2bff5197b2d0937d6ca/docs/writing-intel-modules.md

manesioz:
I'll look it over, that's really helpful. Thanks for your patience.

achantavy:
I'm going to get this into our internal testing today.  In the meantime, can you please run `make test_lint` and push up the results from that to make the Travis CI linter happy?

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,158 @@
...
+            RetentionPeriod=bucket['retention_period'],
+            IamConfigBucketPolicyOnly=bucket['iam_config_bucket_policy_only'],
+            OwnerEntity=bucket['owner_entity'],
+            OwnerEntityId=bucket['owner_entity_id'],
+            VersioningEnabled=bucket['versioning_enabled'],
+            LogBucket=bucket['log_bucket'],
+            RequesterPays=bucket['requester_pays'],
+            DefaultKmsKeyName=bucket['default_kms_key_name'],
+            gcp_update_tag=gcp_update_tag,
+        )
```
achantavy: Ah, we're super close but I totally missed this: we need a `sync_buckets()` function which will call your `get` + `transform` + `load`.

Then we need to call `sync` here: https://github.com/lyft/cartography/blob/6e060389fbeb14f4ccc3e58005230129f1c6962f/cartography/intel/gcp/__init__.py#L61

manesioz: My bad, I should have noticed, thanks!

manesioz: Hm, for some reason it's not passing Travis somehow despite running `make test_lint`

cartography/intel/gcp/__init__.py
```diff
@@ -55,9 +68,24 @@ def _initialize_resources(credentials):
         crm_v1=_get_crm_resource_v1(credentials),
         crm_v2=_get_crm_resource_v2(credentials),
         compute=_get_compute_resource(credentials),
+        storage=_get_storage_resource(credentials),
     )
 
 
+def _sync_gcp_bucket(neo4j_session, resources, project_id, gcp_update_tag, common_job_parameters):
```
achantavy: I think we can remove this function and replace it by putting 

```
storage.sync_gcp_buckets(neo4j_session, resources.storage, project_id, gcp_update_tag, common_job_parameters)
```
on line 101 and then it should work (after making the linter happy).

This way, `_sync_single_project()` will pass a `project_id` to `storage.sync_gcp_buckets()`.

manesioz: Like this?
```python
def _sync_single_project(neo4j_session, resources, project_id, gcp_update_tag, common_job_parameters):
    compute.sync(neo4j_session, resources.compute, project_id, gcp_update_tag, common_job_parameters)
    storage.sync(neo4j_session, resources.storage, project_id, gcp_update_tag, common_job_parameters)
````

What if someone just wants to `sync` projects? Wouldn't they inadvertently also `sync` buckets because it was abstracted away? 

If the reason for this is to ensure that `projects` are up-to-date before syncing `buckets`, maybe we could add the following to `_sync_gcp_bucket()`: 

```python
def _sync_gcp_bucket(neo4j_session, resources, project_id, gcp_update_tag, common_job_parameters):
    _sync_single_project(neo4j_session, resources, project_id, gcp_update_tag, common_job_parameters)
    storage.sync(neo4j_session, resources.storage, project_id, gcp_update_tag, common_job_parameters)
````
Maybe I'm not thinking about this right, but I'd appreciate your insight

File: cartography/intel/gcp/__init__.py [hunk 9]
```diff
@@ -70,6 +84,7 @@ def _sync_single_project(neo4j_session, resources, project_id, gcp_update_tag, c
     :return: Nothing
     """
     compute.sync(neo4j_session, resources.compute, project_id, gcp_update_tag, common_job_parameters)
+    storage.sync(neo4j_session, resources.storage, project_id, gcp_update_tag, common_job_parameters)
```
achantavy: Isn't the function name actually `storage.sync_gcp_buckets()`?

manesioz: Oops, apologies

manesioz:
Hey, any update since internal testing?

achantavy:
Sorry for the delay!  We've been tied up but I'll try to get back to you today or tomorrow (probably tomorrow).

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,204 @@
...
+        if reason == 'invalid':
+            logger.debug(
+                (
+                    "The project %s is invalid - returned a 400 invalid error."
+                    "Full details: %s"
+                ),
+                project_id,
+                e,
+            )
+            return None
```
achantavy: We found a case where a project ID exists in GCP, but still hits this code path.  Can you please change the `logger.debug` to `logger.warn`, and have this `return {}` instead?  

Returning `None` causes transform to fail later on with 

```
File \"/cartography/intel/gcp/storage.py\", line 202, in sync_gcp_buckets
    bucket_list = transform_gcp_buckets(storage_res)
  File \"/cartography/intel/gcp/storage.py\", line 62, in transform_gcp_buckets
    for b in bucket_res.get('items', []):
AttributeError: 'NoneType' object has no attribute 'get'"
```

This way when we hit line 62, `bucket_res.get()` will be the same as `{}.get()` and not `None.get()`, so we won't get the above exception.

cartography/intel/gcp/storage.py
```diff
@@ -0,0 +1,204 @@
...
+        elif reason == 'forbidden':
+            logger.debug(
+                (
+                    "You do not have storage.bucket.list access to the project %s. "
+                    "Full details: %s"
+                ), project_id, e, )
+            return None
```
achantavy: As above, can you please change the `logger.debug` to `logger.warning`?

Same with `return {}`.

cartography/intel/gcp/storage.py
manesioz: Done! Sorry for that over-look on my part. Also I used `logger.warning` instead of `logger.warn` because the latter is [depreciated](https://stackoverflow.com/a/15655674/10981911)

cartography/intel/gcp/storage.py
manesioz: Done!

cartography/intel/gcp/storage.py
manesioz: Any idea why it hit that code path despite having a valid `project_id`?

achantavy: The project exists, but running get_gcp_buckets() on it resulted in a 400 with reason="invalid".  I'm not sure why this is the case (we might have some odd permissions thing going on? idk), but we did get other buckets.

manesioz:
Awesome, thank you so much for your patience, I learned a lot! I hope to collab more in the future too, thanks again :)

PR #152, opened 2019-09-12
Resolve conflicts between GSuite and CRXcavator ingest

tayasteere:
Removed GSuiteUser generation from crxcavator module
Removed cleanup of GSsuiteUser nodes from crxcavator module

### DIFF
--- a/tests/integration/cartography/intel/crxcavator/test_crxcavator.py
+++ b/tests/integration/cartography/intel/crxcavator/test_crxcavator.py
@@ -148,10 +148,9 @@ def test_transform_and_load_user_extensions(neo4j_session):
     """
     users_res = tests.data.crxcavator.crxcavator.USER_RESPONSE
     type(users_res)
-    users_list, extensions_list, user_extensions_list = \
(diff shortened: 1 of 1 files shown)


cartography/data/jobs/cleanup/crxcavator_import_cleanup.json
```diff
@@ -4,11 +4,6 @@
     "iterative": true,
     "iterationsize": 100
   },
-  {
-    "query": "MATCH (n:GSuiteUser) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
sachafaust: might want to keep this since you do a MERGE - If Gsuite module runs first, there's small chance this would cause issues with extra writes. Also, assume GSuite intel may not be running also for various reasons

tayasteere:
Please hold off on reviewing while additional troubleshooting is done with crx - gsuite integration.

PR #155, opened 2019-09-17
Change AWS sync function names to be consistent throughout the intel module.

tobli:
Push details of ec2 syncing into separate file, to align with all other aspects of AWS syncing. Also align function names, to use ”sync” everywhere.


tobli:
More consistency means code base is easier to navigate (especially for new contributors).

PR #165, opened 2019-09-26
Remove usage of old neo4j driver api.

tobli:
As noted on https://neo4j.com/docs/api/python-driver/current/#installation, the neo4j driver should be installed via the neo4j package, and use of neo4j.v1 import should be removed, due to future removal. Update accordingly.


achantavy:
Thanks for looking into this, we'll test internally and get back to you.

setup.py
```diff
@@ -29,7 +29,7 @@
         "boto3>=1.7.0",
         "botocore>=1.12.0",
         "dnspython>=1.15.0",
-        "neo4j-driver>=1.7.0",
+        "neo4j>=1.7.0",
```
ecdavis: After digging into this it looks like the `neo4j-driver` package will cease being published, but the next version of the `neo4j` package is 4.0. While we're at it, can you please add a `<4.0` constraint here?

PR #169, opened 2019-10-01
Add repeat requests helper method for Google APIs (GSuite)

skiptomyliu:
- We make a lot of requests to GSuite. This makes it more resilient in case of a Google server blip, which we've seen.
- Add retry for GCP compute fetching.
- Refactor the transform objects method to be more general to prevent copy-pasta.


cartography/intel/helper/google_request.py
```diff
@@ -0,0 +1,33 @@
+import logging
+import time
+
+from googleapiclient.discovery import HttpError
+logger = logging.getLogger(__name__)
+
+
+def repeat_request(req, req_args, req_next=None, retries=5, retry_delay_ms=500):
```
sachafaust: Could we implement this as generic retry operation we can use across all modules and tasks?

For example of method signature:
invoke_with_retry(operation, retries, retry_delay, retriable_exception_types[])

achantavy:
As @sachafaust, @ecdavis, and I discussed earlier, let's make sure that the cleanup stage does not get executed if all of the retries fail.

achantavy:
#184 seems to handle the GSuite work.  I wonder if we could use a similar approach for GCP in this PR.

skiptomyliu:
> #184 seems to handle the GSuite work. I wonder if we could use a similar approach for GCP in this PR.

Just updated GCP with the retries.

PR #174, opened 2019-10-05
Add AWS tags for EC2 resources.

tobli:
Add tags for EC2Instance, NetworkInterface, EC2SecurityGroup, EC2Subnet, and AWSVPC.

The new Tag label should be reusable for other taggable resources.


tobli:
image
Here's an instance tagged with two tags: Name and Environment.

cartography/intel/aws/ec2.py
```diff
@@ -847,6 +888,50 @@ def load_ec2_vpcs(neo4j_session, data, region, current_aws_account_id, aws_updat
...
+
+TAG_INGEST_TEMPLATE = Template("""
+MATCH (n:$label{$property: {id}})
+WITH n
+MERGE (tag:Tag{key: {Key}})
+ON CREATE SET tag.firstseen = timestamp()
+WITH tag, n
+MERGE (n)<-[r:TAGGED]-(tag)
+ON CREATE SET r.firstseen = timestamp()
+SET r.value = {Value}, r.lastupdated = {aws_update_tag}
```
ramonpetgrave64: Can you instead make `value` a property of the `Tag` node, itself?
```
MERGE (tag:{key: {Key}, value:{Value}})
```
That way, it might be simpler to do something like this
```
MATCH (tag:Tag)--[:TAGGED]->(n)
WHERE tag.key = 'Environment' and tag.value= 'Dev'
RETURN tag, n
```
, or even this
```
MATCH (tag:TAG)
WHERE tag.key = 'Environment'
RETURN tag
```

achantavy: Agree on making `value` a property of the node instead of the relationship.  @SecPrez knows more on this but there's something about properties on relationships being slow.

cartography/intel/aws/ec2.py
```diff
@@ -847,6 +888,50 @@ def load_ec2_vpcs(neo4j_session, data, region, current_aws_account_id, aws_updat
...
+    'network-interface': {'label': 'NetworkInterface', 'property': 'id'},
+    'security-group': {'label': 'EC2SecurityGroup', 'property': 'id'},
+    'subnet': {'label': 'EC2Subnet', 'property': 'subnetid'},
+    'vpc': {'label': 'AWSVPC', 'property': 'id'},
+}
+
+TAG_INGEST_TEMPLATE = Template("""
+MATCH (n:$label{$property: {id}})
+WITH n
+MERGE (tag:Tag{key: {Key}})
```
achantavy: Let's add another label here: `:Tag` can be a general label, and for AWS assets we can call this an `AWSTag` or something similar.

We would also have to add the generic `:Tag` label [to GCPNetworkTags](https://github.com/lyft/cartography/blob/1148c98e58238a83833a4a20f620fa49df1bf009/cartography/intel/gcp/compute.py#L542).

cartography/intel/aws/ec2.py
achantavy: I might be missing it, but where is `{Key}` set?

What is a way to uniquely identify tags?  If an ARN or similar doesn't exist in AWS then we might need to create a fake one like in [GCP](https://github.com/lyft/cartography/blob/1148c98e58238a83833a4a20f620fa49df1bf009/cartography/intel/gcp/compute.py#L184).

cartography/intel/aws/ec2.py
```diff
@@ -1031,6 +1071,50 @@ def load_ec2_vpcs(neo4j_session, data, region, current_aws_account_id, aws_updat
...
+ON CREATE SET tag.firstseen = timestamp()
+WITH tag, n
+MERGE (n)<-[r:TAGGED]-(tag)
```
achantavy: For consistency, would it make sense to reverse the directionality here?

At least in the case of GCP, this way makes more sense to me: 
```
(GCPInstance)-[:TAGGED]->(GCPNetworkTag)
```
https://github.com/lyft/cartography/blob/master/docs/schema/gcp.md#relationships-5

tests/data/aws/ec2.py
```diff
@@ -15,6 +15,41 @@
     ],
 }
 
+DESCRIBE_TAGS = {
```
achantavy: Thanks very much for including mock data and integration tests.  Can you also please add an entry to docs/schema.md?

cartography/intel/aws/ec2.py
achantavy: One more related note: we should create an index for `AWSTag` and decide on what field to use for it.

achantavy:
Hey @tobli and everyone else following this PR,
Sorry again for the late responses, but I'm considering closing this PR and replacing it with one that uses the AWS [resourcegroupstagging API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html) instead, because the resourcegroupstagging API seems to return tags for _all_ taggable resources, not just EC2 assets.

As seen here, it also returns tags for S3 buckets:
image (In the CLI run something like `aws resourcegroupstaggingapi get-resources --tags-per-page 100`)

I'd probably start work on it this week if time permits but I definitely welcome feedback.  Do you (or anyone else reading this) have any thoughts?

ramonpetgrave64:
@achantavy this is nicer, but could be more difficult to fully implement, unless tags are getting ingested last.

achantavy:
>unless tags are getting ingested last.

@ramonpetgrave64 Yup!  That's how I would do it.

Issue #11, opened 2019-03-07
Need to create index on AWSPrincipal and not Principal


achantavy:
See https://github.com/lyft/cartography/blob/master/cartography/data/indexes.cypher#L37:

We create an index on a `Principal` node, but we don't ingest that anywhere.  Perhaps this should be an [AWSPrincipal](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/iam.py#L72) instead?

achantavy:
Resolved in #175!

PR #175, opened 2019-10-05
Update incorrect Neo4j index: remove unused :Principal index and add missing :AWSPrincipal index.

tobli:
There’s a AWSPrincipal label but no Principal label. Update index accordingly.


tobli:
see #11

PR #180, opened 2019-10-11
Change CRXcavator intel module to pull configuration from Config object rather than directly from environment.

ecdavis:


cartography/cli.py
```diff
@@ -112,6 +112,23 @@ def _build_parser(self):
...
+                'Base URI for the CRXcavator API. Required if you are using the CRXcavator intel module. Ignored '
+                'otherwise.'
+            ),
+        )
+        parser.add_argument(
+            '--crxcavator-api-key-env-var',
+            type=str,
+            default=None,
+            help=(
+                'Key for the CRXcavator API. Required if you are using the CRXcavator intel module. Ignored otherwise.'
```
achantavy: How about: "Name of environment variable that contains your CRXcavator API key. Required if you are using the CRXcavator intel module. Ignored otherwise."?

cartography/cli.py
```diff
@@ -112,6 +112,24 @@ def _build_parser(self):
                 'respects the AWS CLI/SDK environment variables and does not override them.'
             ),
         )
+        parser.add_argument(
+            '--crxcavator-api-base-uri',
```
sachafaust: update documentation to reflect that cli parameters are available

achantavy: @tayasteere Can you please update the docs to reflect the new CLI params?

achantavy:
Obsoleted by #235

Issue #179, opened 2019-10-10
Change neo4j query templating to use string.Template


ecdavis:
See https://github.com/lyft/cartography/pull/174/files

tobli:
I'm assuming this refers to when #VARIABLE_NAME# is used with String.replace to create dynamic Cypher statements, not for variables in "regular" Cypher statements.

i.e.
https://github.com/lyft/cartography/blob/1a440cf42ba6f88fc079608d83b84921222ca220/cartography/intel/aws/ec2.py#L391
https://github.com/lyft/cartography/blob/1a440cf42ba6f88fc079608d83b84921222ca220/cartography/intel/aws/ec2.py#L882-L884

achantavy:
Closing this issue; the codebase is thankfully clean now from this bad pattern :P  Will reopen if it ends up that I have missed something.

PR #181, opened 2019-10-13
Replace custom templating with python Template in AWS EC2.

tobli:
Instead of chaining string.replace, use Template strings from the Python standard library.


tobli:
See #179

achantavy:
I really like this.  We're currently backed up on testing at the moment so it'll be a little while for us to merge unfortunately. 

@ramonpetgrave64  I think that this approach will work well to fix #183, as we can parameterize `r`  [here](https://github.com/lyft/cartography/blob/master/cartography/intel/gcp/compute.py#L757) and [here](https://github.com/lyft/cartography/blob/master/cartography/intel/gcp/compute.py#L763) to either a `[DENIED_BY]` or an `[ALLOWED_BY]` relationship.

achantavy:
I'll test this one later today or tomorrow.

PR #182, opened 2019-10-13
Fix syntax error in Travis file.

tobli:
Apt packages is an array.


achantavy:
Thanks for doing this!

PR #185, opened 2019-10-15
Crxcavator: add permission nodes

tayasteere:
Adds ChromeExtensionPermission nodes to allow enumeration by permission. This capability was requested internally. 
Added new test
Updated docs


cartography/intel/crxcavator/crxcavator.py
```diff
@@ -195,8 +246,31 @@ def load_extensions(extensions, session, update_tag):
     e.lastupdated = {UpdateTag}
     """
 
+    permissions_ingestion_cypher = """
+    UNWIND {Permissions} as permission
+    MERGE (e:ChromeExtensionPermission{name: permission})
```
achantavy: Also, I'm not familiar with this new-style syntax.  `permission` looks like it's an object, but you're making it the `name` member?  Does it work?

I'm out of the loop on this lol.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -195,8 +246,31 @@ def load_extensions(extensions, session, update_tag):
...
+    ON CREATE SET
+    e.firstseen = timestamp()
+    """
```
achantavy: Also do a `SET e.lastupdated`

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -103,6 +107,23 @@ def transform_extensions(extension_details):
...
+            if type(permission) is dict:
+                parsed_permissions = parse_permissions_dict(permission)
+                permissions_set.update(parsed_permissions)
+                for sub_permission in parsed_permissions:
+                    extensions_permissions.append({
+                        'id': f"{extension_id}|{version}",
+                        'permission': sub_permission,
+                    })
+                continue
+            permissions_set.add(permission)
```
achantavy: If `permission` somehow happens to _not_ be a dict, then we still add it to `permissions_set`.

Later on in `load_extensions()` when this is `permissions_list`, if there is an object in `permissions_list` isn't a dict, will the sync crash?

I guess what I'm asking is to clarify the exact contract of the output objects from this function.  The doc of this function says that we "return a unique list of permissions", but can you please clarify what each permission object is expected to contain?

cartography/intel/crxcavator/crxcavator.py
tayasteere: Permission is just a string after parsing it - see details on next comment

cartography/intel/crxcavator/crxcavator.py
achantavy: @ecdavis can weigh in here, but I think each entry in the list should have the same datatype so that it will be easier to debug in the future.

tayasteere: not sure what the issue is here? Each entry in the list will be a string after it is parsed out.

cartography/data/indexes.cypher
```diff
@@ -64,3 +64,4 @@ CREATE INDEX ON :RDSInstance(id);
 CREATE INDEX ON :S3Acl(id);
 CREATE INDEX ON :S3Bucket(name);
 CREATE INDEX ON :User(arn);
+CREATE INDEX ON :ChromeExtensionPermission(name);
```
achantavy: Please index on `id` at minimum.  You can add an additional index for `name` if you expect other callers to query based on `name` and not `id`.

Also, please move this in alphabetical order.

cartography/intel/crxcavator/crxcavator.py
achantavy: I misread your comment earlier; got it.

docs/schema/crxcavator.md
```diff
@@ -74,10 +71,31 @@ Placeholder representation of a single G Suite [user object](https://developers.
...
+| lastupdated |  Timestamp of the last time the node was updated |
+| id | The permission name" |
+| name | The specific permission declared in the manifest
+
+### Relationships
+
+- ChromeExtension declares ChromeExtensionPermission.
+
+    ```
+    (ChromeExtension)-[DECLARES]->(ChromeExtensionPermission)
```
achantavy: Would like at least another set of eyes on this PR.

@SecPrez what do you think about representing permissions (https://developer.chrome.com/apps/declare_permissions#manifest) as nodes?  Would it be better if this was a field (like a list) on the ChromeExtension itself?

My thought is that ChromeExtension-to-ChromeExtensionPermission is many-to-1 relationship, so it could be unwieldy if someone were to expand all nodes from a ChromeExtensionPermission node.

Hm... but it **would** be very helpful to ask the question "what extensions ask for this super powerful permission?".

Sorry for the rambling reply; I'm fine with the way it is right now but I wanted to make sure @SecPrez saw it too.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -195,8 +246,34 @@ def load_extensions(extensions, session, update_tag):
...
+    ON CREATE SET
+    e.firstseen = timestamp()
+    SET
+    e.name = permission,
+    e.lastupdated = {UpdateTag}
+    """
+
+    extensions_permissions_cypher = """
+    UNWIND {ExtensionPermissions} as extension_permission
+    MATCH (perm:ChromeExtensionPermission{name: extension_permission.permission}),
```
achantavy: If you change the index [here](https://github.com/lyft/cartography/blob/4667376439ed96d695b633a78532560daae3eccb/cartography/data/indexes.cypher#L67) to ID, make sure that the `match` uses an indexed field.

cartography/intel/crxcavator/crxcavator.py
```diff
@@ -134,7 +155,35 @@ def transform_extensions(extension_details):
...
+    for socket_permission in socket:
+        permissions.append(f"socket-{socket_permission}")
+    media_galleries = permissions_dict.get('mediaGalleries', [])
+    for media in media_galleries:
+        permissions.append(f"mediagalleries-{media}")
+    if len(permissions) == 0:
+        # this is a case not currently handled, so return the raw json object for ingestion
+        permission = json.dumps(permissions_dict)
+        logger.warning(f"Unknown permissions dict type {permission}")
+        permissions.append(permission)
```
achantavy: Do we still want to add this if we don't know what the permission is?

`json.dumps()` yields a string representation of the dict and I think it'll include brackets which could make merging it into a neo4j node messy.

docs/schema/crxcavator.md
SecPrez: If it is a many-to-1 permission then it will need a node per permission. We should see how many nodes we are going to import. If the number is too high then only import the security interesting permissions vs all of them.

tayasteere: 2789 permission nodes during the test run, not a huge total. We could trim out the specific URI permissions, which would cut out most of those if this is a problem.

SecPrez: That sounds reasonable. No objections here.

cartography/intel/crxcavator/crxcavator.py
tayasteere: removed this, logging only.

PR #187, opened 2019-10-19
Split 1k+ line ec2 module to multiple files.

tobli:
The ec2 module is super large, making it hard to navigate.
This commit splits it into multiple files, one per sync method (related resource types). This allows for updated method names that are consistent between the different files. Again making in easier to navigate and recognise familiar patterns.
The naming changes are:
- get_XYZ -> get_from_aws
- load_XYZ -> update_graph
- cleanup_XYZ -> cleanup_graph


tobli:
Here's a proposal for a pattern for splitting large intel modules into smaller, more manageable, chunks. If this makes sense I can apply the same pattern to other intel modules.
Note that the method names are mere suggestions, based on my personal preference. E.g. `update_graph` felt more natural to me than `load_…` for methods that writes to the graph database.
Also, this is not intended to be a proposal for an ideal design, there's a number of opportunities
for tweaks that could make code reuse, and additional configurability (e.g. disabling certain modules, or parts of modules) easier. I wanted to post this early, to get some feedback if this could be a helpful step on the way.

cartography/intel/aws/ec2/_init_.py
```diff
@@ -0,0 +1,26 @@
+import logging
+
+from intel.aws.ec2 import auto_scaling_groups
+from intel.aws.ec2 import instances
+from intel.aws.ec2 import key_pairs
+from intel.aws.ec2 import load_balancers
+from intel.aws.ec2 import security_groups
+from intel.aws.ec2 import vpc
+from intel.aws.ec2 import vpc_peering
```
tobli: I'd love to make these `from . import …` but the mypy pre-commit check fails with `error: No parent module -- cannot perform relative import`.

achantavy:
Hi @tobli, thank you for putting this together.  We like the change and agree that the EC2 module could benefit from being split up in the future, but right now our EC2 work is very difficult to validate with our current lack of automated tests, so we currently don't have the bandwidth to merge this PR.

Sorry for the late response, but let's revisit this once we have adequate unit/integration tests for EC2.

PR #188, opened 2019-10-19
Update pre-commit hooks.

tobli:
Results of running `pre-commit autoupdate`, and fixing resulting lint errors. Also make sure all hooks execute with python 3.


tobli:
This reduces, but does not fix, my issues of getting `make test_lint` to run locally. I still get the following errors:
```
Reorder python imports...................................................Failed
hookid: reorder-python-imports

Files were modified by this hook. Additional output:

Reordering imports in cartography/sync.py
Reordering imports in tests/integration/conftest.py
Reordering imports in tests/integration/test_basic.py
Reordering imports in cartography/driftdetect/get_states.py
```

Fixing the import orders locally gets my local lint (python 3.7.4) working, but it fails in Travis (that wants to revert my changes). I don't know enough python to understand what's going on with that. At least this PR makes Travis and my local setup disagree a bit less…

achantavy:
Can you provide more information your setup?  In particular what are the imports that are failing?

achantavy:
You probably found this out already, but if `make test_lint` gives an error the first time, pre-commit will typically make a change to fix the lint problem:
Screen Shot 2019-12-05 at 2 30 55 PM

After that, you can add and commit the change and CI should work.

Issue #183, opened 2019-10-14
Error while cleaning GCP Instances


marco-lancini:
Currently Cartography raises the following error while cleaning up GCP Instances:

```
...
INFO:cartography.sync:Finishing sync stage 'aws'
INFO:cartography.sync:Starting sync stage 'gcp'
INFO:oauth2client.transport:Attempting refresh to obtain initial access_token
INFO:oauth2client.client:Refreshing access_token
INFO:oauth2client.transport:Attempting refresh to obtain initial access_token
INFO:oauth2client.client:Refreshing access_token
INFO:cartography.intel.gcp:Syncing GCP project example-project.
INFO:cartography.intel.gcp.compute:Syncing Compute objects for project example-project.
INFO:oauth2client.transport:Attempting refresh to obtain initial access_token
INFO:oauth2client.client:Refreshing access_token

ERROR:cartography.graph.job:Unhandled error while executing statement in job 'cleanup GCP Instances': Variable `r` already declared (line 25, column 26 (offset: 799))
" MERGE (fw)<-[r:DENIED_BY]-(rule)"
^
ERROR:cartography.sync:Unhandled exception during sync stage 'gcp'
Traceback (most recent call last):
File "/usr/lib/python3.7/site-packages/cartography/sync.py", line 68, in run
stage_func(neo4j_session, config)
File "/usr/lib/python3.7/site-packages/cartography/intel/gcp/__init__.py", line 134, in start_gcp_ingestion
_sync_multiple_projects(neo4j_session, resources, projects, config.update_tag, common_job_parameters)
File "/usr/lib/python3.7/site-packages/cartography/intel/gcp/__init__.py", line 94, in _sync_multiple_projects
_sync_single_project(neo4j_session, resources, project_id, gcp_update_tag, common_job_parameters)
File "/usr/lib/python3.7/site-packages/cartography/intel/gcp/__init__.py", line 72, in _sync_single_project
compute.sync(neo4j_session, resources.compute, project_id, gcp_update_tag, common_job_parameters)
File "/usr/lib/python3.7/site-packages/cartography/intel/gcp/compute.py", line 948, in sync
sync_gcp_firewall_rules(neo4j_session, compute, project_id, gcp_update_tag, common_job_parameters)
File "/usr/lib/python3.7/site-packages/cartography/intel/gcp/compute.py", line 911, in sync_gcp_firewall_rules
cleanup_gcp_firewall_rules(neo4j_session, common_job_parameters)
[...]

PR #189, opened 2019-10-21
Fix query for gcp firewall ingress deny rules.

ramonpetgrave64:
The bug happens whenever there are existing rules in both `fw['transformed_allow_list']` and in `fw['transformed_deny_list']`. In that case, both of the sub-queries creating the realtionships `[r:ALLOWED_BY]` and `[r:DENIED_BY]` would have been added to the main `query` string, causing and error. This could be related to the error mentioned in #183 .

I haven't tested this yet, but I believe this should fix it.


achantavy:
Thanks very much for getting started on this.  I think that using an approach like #181 might be a very clean way to parameterize the relationship type of ALLOWED_BY and DENIED_BY, would you consider doing it that way?

marco-lancini:
Hi! I hope this is not going to be closed, as at the moment Cartography crashes when this condition is verified (as mentioned in #183)

achantavy:
I'll do some internal testing on this later on and report back.

ramonpetgrave64:
Okay. This fix also seems to deal with issue #183 . Still not sure why. 

I had trouble testing, since I don't have GCP Organization set up with my personal account, so I did a hacky workaround for that.

Also did the subquery replacement with `Template`, as @achantavy suggested.

marco-lancini:
Hello! :wave: 
I can confirm I just tested this diff and it does solve issue #183! :tada:

achantavy:
Sorry again for how long this has taken.  My test run of this PR has started right now; will report back.

ramonpetgrave64:
We should probably also add test cases for this, too.

achantavy:
Ah right, since the [existing test](https://github.com/lyft/cartography/blob/e2d7a626ba29df30ff308a9d3c311b297a7c8d30/tests/integration/cartography/intel/gcp/test_compute.py#L336) only checks for `ALLOWED_BY`.

marco-lancini:
@achantavy: would it be possible to push a minor release of Cartography, so to include this fix into the PyPi package (https://pypi.org/project/cartography/)?

achantavy:
@marco-lancini On it, sorry for delays.

achantavy:
Alright done.

PR #191, opened 2019-10-22
Map Okta groups to AWS roles and add Okta Application ReplyUris

SecPrez:
Added the ReplyUri nodes and the REPLYURI relationships to okta applications to see if any oauth phishing is possible.
The okta application client does not return the reply urls from okta, so I switched the list applications function to use the api_client so that all data is available.


README.md
```diff
@@ -59,55 +59,57 @@ Time to set up the server that will run Cartography.  Cartography _should_ work
 
 			⚠️ At this time we run our automated tests on Neo4j version 3.5.\*.  Other versions may work but are not explicitly supported. ⚠️
 
-	2. [Install](https://neo4j.com/docs/operations-manual/current/installation/) Neo4j on the server you will run Cartography on.
+	1. [Install](https://neo4j.com/docs/operations-manual/current/installation/) Neo4j on the server you will run Cartography on.
```
tayasteere: Numbering on readme is broken now

cartography/data/indexes.cypher
```diff
@@ -68,6 +68,7 @@ CREATE INDEX ON :OktaUserFactor(id);
 CREATE INDEX ON :OktaTrustedOrigin(id);
 CREATE INDEX ON: OktaAdministrationRole(id);
 CREATE INDEX ON :PublicIpAddress(ip);
+CREATE INDEX ON :ReplyUri(uri)
```
tayasteere: Put in alphabetical order

cartography/data/jobs/cleanup/okta_import_cleanup.json
```diff
@@ -1,70 +1,95 @@
...
+                  "__comment__": "Delete Stale OktaGroup to AWSRole ALLOWED_BY relationship"
+            },
+            {
+                  "query": "MATCH (:AWSRole)<-[r:CAN_ASSUME_ROLE]-(:Human) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE r return COUNT(*) as TotalCompleted",
+                  "iterative": true,
+                  "iterationsize": 100,
+                  "__comment__": "Delete Stale Human to AWSRole CAN_ASSUME_ROLE relationship"
+            },
+            {
+                  "query": "MATCH (n:ReplyUri) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
tayasteere: Move up with node statements

cartography/intel/okta/applications.py
```diff
@@ -1,47 +1,46 @@
 # Okta intel module - Applications
 import json
 import logging
+import socket
```
tayasteere: What socket functionality is used? Is this something we can do with requests like the other modules?

SecPrez: socket is used to check the dns for the domain of each reply url.

cartography/intel/okta/applications.py
```diff
@@ -181,34 +180,59 @@ def transform_okta_application_list(okta_applications):
...
+    Extracts the reply uri information from an okta app
+    and determines if the dns of the reply url is valid
+    """
+    uris = []
+    if "oauthClient" in okta_application["settings"]:
+        if "redirect_uris" in okta_application["settings"]["oauthClient"]:
+            for uri in okta_application["settings"]["oauthClient"]["redirect_uris"]:
+                netloc = urlparse(uri).netloc
+                try:
+                    socket.gethostbyname(netloc)
```
achantavy: Ooh interesting...

We talked about this a bit in person.  I really like that we're starting to explore more "active" forms of data validation.  At bare minimum for safety and speed reasons, I think this step needs a toggle switch with its default set to "do not perform gethostbyname()".

I almost think this kind of functionality might fit in better with a separate tool that would integrate with Cartography: kind of an "active reconnaissance data validator".

CC: @ecdavis

README.md
SecPrez: This is the standard way to doing numbering in markdown. It will render the page the same as it was, with incrementing numbers, but it will be easier to modify in the future.

cartography/data/indexes.cypher
SecPrez: fixed

cartography/intel/okta/applications.py
SecPrez: I would really like to get this in, so if I need to pull this functionality I can. Personally i think it fits as its passive data gathering. We could add in a cli parameter and opt-in or opt-out of these queries.

cartography/data/jobs/cleanup/okta_import_cleanup.json
SecPrez: Done

cartography/cli.py
```diff
@@ -141,6 +141,17 @@ def _build_parser(self):
...
         )
+        parser.add_argument(
+            '--okta-saml-role-regex',
+            type=str,
+            default=r"^aws\#\S+\#(?{{role}}[\w\-]+)\#(?{{accountid}}\d+)$",
+            help=(
+                'The regex used to map okta groups to aws roles when using okta as a SAML provider.'
+                'The regex is the one entered in Step 5: Enabling Group Based Role Mapping in Okta'
+                'https://saml-doc.okta.com/SAML_Docs/How-to-Configure-SAML-2.0-for-Amazon-Web-Service#c-step5'
+                'It the regex must contain the {{role}} and {{accountid}} tags'
```
SecPrez: Fix wording

cartography/intel/okta/awssaml.py
```diff
@@ -0,0 +1,112 @@
...
+        role_arn = f"arn:aws:iam::{accountid}:role/{role}"
+        return {"groupid": group_id, "role": role_arn}
+
+
+def query_for_okta_to_aws_role_mapping(neo4j_session, mapping_regex):
+    """
+    Query the graph for all groups associated with the amazon_aws application and map them to AWSRoles
+    :param neo4j_session: session from the Neo4j server
+    :param mapping_regex: the regex used by the organization to map groups to aws roles
+    """
```
achantavy: Is there a way to replace this with a `merge` statement?  I have some resistance to using `match` to return results from neo4j to use somewhere else; I think there should be a way to avoid the extra back and forth of transferring data.

cartography/cli.py
```diff
@@ -141,6 +141,17 @@ def _build_parser(self):
                 'Required if you are using the Okta intel module. Ignored otherwise.'
             ),
         )
+        parser.add_argument(
+            '--okta-saml-role-regex',
+            type=str,
+            default=r"^aws\#\S+\#(?{{role}}[\w\-]+)\#(?{{accountid}}\d+)$",
+            help=(
+                'The regex used to map okta groups to aws roles when using okta as a SAML provider.'
+                'The regex is the one entered in Step 5: Enabling Group Based Role Mapping in Okta'
```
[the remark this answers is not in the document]
SecPrez: There are no bookmarks on the page, so they need the regex from that section. Step 5

cartography/intel/okta/awssaml.py
```diff
@@ -0,0 +1,112 @@
...
+    query = "MATCH (app:OktaApplication{name:'amazon_aws'})--(group:OktaGroup) return group.id, group.name"
```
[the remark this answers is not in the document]
SecPrez: I am adding an index to the name field

cartography/data/indexes.cypher
```diff
@@ -70,6 +70,7 @@ CREATE INDEX ON: OktaAdministrationRole(id);
 CREATE INDEX ON :PublicIpAddress(ip);
 CREATE INDEX ON :RDSInstance(db_instance_identifier);
 CREATE INDEX ON :RDSInstance(id);
+CREATE INDEX ON :ReplyUri(uri)
```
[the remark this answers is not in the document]
SecPrez: done

docs/schema/okta.md
```diff
@@ -259,3 +265,22 @@ Representation of Okta administration roles. For more information, visit Okta do
     ```
     (OktaOrganization)-[RESOURCE]->(OktaAdministrationRole)
     ```
+
+## Reply Uri
+
+Representation of Okta application ReplyUri. For more information, visit Okta documentation https://developer.okta.com/docs/reference/api/apps/
+
+| Field | Description |
```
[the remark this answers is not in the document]
SecPrez: done

cartography/intel/okta/awssaml.py
SecPrez: We cannot due that because we need to evaluate a regex on each of the groups and extract values from regex. If the regex was static that might be possible to come up with a query that works, but as people can change the regex we have to run some parsing.

cartography/data/jobs/cleanup/okta_import_cleanup.json
```diff
@@ -1,70 +1,95 @@
...
+                  "__comment__": "Delete stale OktaOrganization"
+            },
+            {
+                  "query": "MATCH (n:ReplyUri) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
+                  "iterative": true,
+                  "iterationsize": 100,
+                  "__comment__": "Delete Stale ReplyUri nodes"
+            },
+            {
+                  "query": "MATCH (:ReplyUri)<-[r:REPLYURI]-(:OktaApplication) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE r return COUNT(*) as TotalCompleted",
```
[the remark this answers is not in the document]
SecPrez: Two apps can have the same replyuri. The first deletion steps deletes replyuris that are no longer used by any application. If app a and app b had the same reply uri and then app a removed it. The replyuri node will still be there with a relationship to app b. The second step will delete the relationship between the replyuri and app a as only that relationship is no longer valid.

cartography/data/jobs/cleanup/okta_import_cleanup.json
```diff
@@ -1,70 +1,95 @@
...
+                  "iterative": true,
+                  "iterationsize": 100,
+                  "__comment__": "Delete Stale ReplyUri relationships"
+            },
+            {
+                  "query": "MATCH (:AWSRole)<-[r:ALLOWED_BY]-(:OktaGroup) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE r return COUNT(*) as TotalCompleted",
```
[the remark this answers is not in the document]
SecPrez: I have moved it, but this is only used to clean up an invalid relationship between two nodes that are still valid.

PR #194, opened 2019-10-30
Update Python requirement to 3.6

tobli:
Cartography requires Python 3.6 (e.g. in its' use of Literal String Interpolation), so update the readme to reflect that.


achantavy:
Thanks for this!  Will merge after CI is done.

PR #199, opened 2019-11-07
Fixing positional argument bug in SAML mapping

SecPrez:


SecPrez:
> I think we need a new release as 0.12.0 has this bug. Probably will do this later today.

If you point me to the docs I can do it

Issue #202, opened 2019-11-19
AttributeError: 'Namespace' object has no attribute 'okta_api_key'


pizza0rodeo:
I've installed Cartography 0.13.0 in a Python 3.6.8 venv and am running it like so:

```
cartography --neo4j-uri bolt://localhost:7687
```

...but I get an uncaught exception when it gets to the Okta stage:

```
INFO:cartography.sync:Starting sync stage 'okta'
ERROR:cartography.sync:Unhandled exception during sync stage 'okta'
Traceback (most recent call last):
  File "/home/ec2-user/venv/lib64/python3.6/site-packages/cartography/sync.py", line 69, in run
    stage_func(neo4j_session, config)
  File "/home/ec2-user/venv/lib64/python3.6/site-packages/cartography/intel/okta/__init__.py", line 48, in start_okta_ingestion
    users.sync_okta_users(neo4j_session, config.okta_org_id, config.update_tag, config.okta_api_key, state)
AttributeError: 'Namespace' object has no attribute 'okta_api_key'
Traceback (most recent call last):
  File "/home/ec2-user/venv/bin/cartography", line 11, in <module>
    load_entry_point('cartography==0.13.0', 'console_scripts', 'cartography')()
  File "/home/ec2-user/venv/lib64/python3.6/site-packages/cartography/cli.py", line 216, in main
    return CLI(default_sync, prog='cartography').main(argv)
  File "/home/ec2-user/venv/lib64/python3.6/site-packages/cartography/cli.py", line 196, in main
    return cartography.sync.run_with_config(self.sync, config)
  File "/home/ec2-user/venv/lib64/python3.6/site-packages/cartography/sync.py", line 135, in run_with_config
    return sync.run(neo4j_driver, config)
  File "/home/ec2-user/venv/lib64/python3.6/site-packages/cartography/sync.py", line 69, in run
    stage_func(neo4j_session, config)
  File "/home/ec2-user/venv/lib64/python3.6/site-packages/cartography/intel/okta/__init__.py", line 48, in start_okta_ingestion
    users.sync_okta_users(neo4j_session, config.okta_org_id, config.update_tag, config.okta_api_key, state)
AttributeError: 'Namespace' object has no attribute 'okta_api_key'
```

I'm not using Okta, and have not provided it any of the CLI options or environment variables for Okta, but it looks as if Cartography is assuming that the `okta_api_key` attribute will be there despite that. Am I doing something wrong here?

Thanks for your help!

[...]

PR #204, opened 2019-11-23
Skip Okta sync if API key is undefined

achantavy:
Fix for #202.


achantavy:
@pizza0rodeo Can you give this a try and let us know how it goes?

Issue #205, opened 2019-11-23
ContextualVersionConflict when running cartography CLI


achantavy:
Anyone else see this error when running `cartography`?  I don't remember this happening before and it doesn't look like we changed the associated package versions.

# Repro steps
1. Make a brand new, fresh virtualenv and activate it.

1. `cd ~/cartography`

1. `pip install -e .`

    You may or may not get this error:

        ERROR: botocore 1.13.26 has requirement python-dateutil<2.8.1,>=2.1; python_version >= "2.7", but you'll have python-dateutil 2.8.1 which is incompatible.

1. Attempt to run cartography with

        `cartography --help`

Result:
```
Traceback (most recent call last):
  File "/Users/me/.virtualenvs/env5/lib/python3.7/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "/Users/me/.virtualenvs/env5/lib/python3.7/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/Users/me/.virtualenvs/env5/lib/python3.7/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (python-dateutil 2.8.1 (/Users/me/.virtualenvs/env5/lib/python3.7/site-packages), Requirement.parse('python-dateutil<2.8.1,>=2.1; python_version >= "2.7"'), {'botocore'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/me/.virtualenvs/env5/bin/cartography", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/Users/me/.virtualenvs/env5/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3251, in <module>
    @_call_aside
  File "/Users/me/.virtualenvs/env5/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3235, in _call_aside
    f(*args, **kwargs)
  File "/Users/me/.virtualenvs/env5/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3264, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/Users/me/.virtualenvs/env5/lib/python3.7/site-packages/pkg_resources/__init__.py", line 585, in _build_master
    return cls._build_from_requirements(__requires__)
[...]

PR #206, opened 2019-11-25
#205: Delete python-dateutil from setup.py since we don't use it directly

achantavy:
See #205 and https://github.com/boto/botocore/issues/1872, in particular [this comment](https://github.com/boto/botocore/issues/1872#issuecomment-549612354).

> This would only be an issue if you're installing something else that is pulling in the latest version of dateutils. If I just use pip install botocore, it pulls in the right version of dateutils.

Cartography pulls in the latest version of dateutils which conflicts with botocore, so we need to cap it for now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
         "neo4j>=1.7.0,<4.0.0",
         "neobolt>=1.7.0,<4.0.0",
         "policyuniverse>=1.1.0.0",
-        "python-dateutil>=2.7.0",
+        "python-dateutil<2.8.1",
         "google-api-python-client>=1.7.8",
         "oauth2client>=4.1.3",
         "marshmallow>=3.0.0rc7",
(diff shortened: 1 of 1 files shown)


achantavy:
For additional context see https://github.com/RasaHQ/rasa/issues/4825 and https://github.com/boto/botocore/issues/1872.

setup.py
```diff
@@ -32,7 +32,7 @@
         "neo4j>=1.7.0,<4.0.0",
         "neobolt>=1.7.0,<4.0.0",
         "policyuniverse>=1.1.0.0",
-        "python-dateutil>=2.7.0",
```
ecdavis: Please add a comment linking to the botocore issue as the reason for the upper constraint.

ecdavis:
Looks like we don't actually use dateutil anywhere, so we can delete it?

ecdavis:
Please update PR title.

Issue #208, opened 2019-11-25
AWS DNS zone are not unique


SecPrez:
AWS zones are merged by name, however, many different AWS accounts can have an AWS zone with the same name. Need to merge on ZoneId instead.

SecPrez:
Because the other queries match on zoneid the graph will be missing a lot of data.

achantavy:
https://docs.aws.amazon.com/Route53/latest/APIReference/API_HostedZone.html

It doesn't look like there is an ARN defined on hosted zones.  Are IDs globally unique?  If not, we'll need to concatenate together an ID (and I dislike doing this even though we've been forced to do this in many other places ☹️)

SecPrez:
There is a unique zoneid in AWS. However cartography is merging on name

achantavy:
Are we sure that is AWS-globally unique?

SecPrez:
Yes

PR #209, opened 2019-11-25
Updates to Route53 Parsing

SecPrez:
Fix issue #208 
Add in nameservers for each domain
Map subzones


cartography/intel/aws/route53.py
```diff
@@ -94,8 +95,8 @@ def load_cname_records(neo4j_session, records, update_tag):
 
 def load_zone(neo4j_session, zone, current_aws_id, update_tag):
     ingest_z = """
-    MERGE (zone:DNSZone:AWSDNSZone{name: {ZoneName}})
-    ON CREATE SET zone.firstseen = timestamp(), zone.zoneid = {ZoneId}
+    MERGE (zone:DNSZone:AWSDNSZone{zoneid:{ZoneId}})
```
SecPrez: Fix for issue #208

achantavy: The change makes sense to me, but I would get a lot more confidence in this significant schema change (changing a primary identifier that is used in merging) if we added an integration test.

This would involve including a tiny redacted sample of AWS DNS data and showing that it gets loaded to the graph as expected and connected with other nodes as expected.

I'm not saying create tests for everything in this module - just ensure that the nodes that you're changing are still connected to the other nodes we expect them to be connected to.

cartography/intel/aws/route53.py
```diff
@@ -9,6 +9,7 @@ def link_aws_resources(neo4j_session, update_tag):
     # find records that point to other records
     link_records = """
     MATCH (n:AWSDNSRecord) WITH n MATCH (v:AWSDNSRecord{value: n.name})
+    WHERE NOT n = v
```
achantavy: I would be more specific here.
e.g. `WHERE n.value <> v.value` or whatever the correct way of identifying them is.

cartography/intel/aws/route53.py
```diff
@@ -114,6 +115,70 @@ def load_zone(neo4j_session, zone, current_aws_id, update_tag):
...
+def link_sub_zones(neo4j_session, update_tag):
+    query = """
+    match (z:AWSDNSZone)
+    <-[:MEMBER_OF_DNS_ZONE]-
+    (record:DNSRecord{type:"NS"})
+    -[:DNS_POINTS_TO]->
+    (ns:NameServer)
+    <-[:NAMESERVER]-
+    (z2)
+    WHERE record.name=z2.name AND NOT z=z2
```
achantavy: For not equal, what do you think about`<>` instead of `NOT`?  I won't block on this comment but I think `<>` is easier to read but that could be just me.

cartography/intel/aws/route53.py
SecPrez: I will look into it because I generally agree on adding integration tests. In this case you can see that every other statement uses the zoneid field which is causing missing relationships. I see this more as a bug fix than a significant schema change.

cartography/data/jobs/cleanup/aws_dns_cleanup.json
```diff
@@ -5,10 +5,26 @@
...
+      "iterationsize": 100
+    },
     {
       "query": "MATCH ()<-[r:DNS_POINTS_TO]-(:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
       "iterative": true,
       "iterationsize": 100
-  }],
+    },
+    {
+      "query": "MATCH (:AWSDNSZone)<-[r:NAMESERVER]-(:NameServer) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
```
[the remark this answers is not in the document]
SecPrez: Going the other way the relationship would be NAMESERVER_FOR. If not NAMESERVER it would could be HOSTED_BY or HOSTS

cartography/intel/aws/route53.py
SecPrez: I want to make sure the node is not equal to the other node, not just making sure a property is not the same. This was causing self-relationships

SecPrez:
There are number discrepancies between staging and prod. AWSDNSZone numbers don't align and AWSDNSRecords are way down. Need to investigate before merging

PR #211, opened 2019-11-26
Adjust log levels on GSuite init

achantavy:
Not all Cartography users will have GSuite so we should log a warning and skip the sync if env vars are undefined.  If the env vars are defined and we fail to initialize the ApplicationDefaultCredentials, then erroring is fine.

### DIFF
--- a/cartography/intel/gsuite/__init__.py
+++ b/cartography/intel/gsuite/__init__.py
@@ -57,6 +57,11 @@ def start_gsuite_ingestion(session, config):
         "UPDATE_TAG": config.update_tag,
     }
 
+    # If the env vars are undefined, the user probably doesn't want to run the GSuite sync.
+    if not GSUITE_CREDS or not GSUITE_DELEGATED_ADMIN:
+        logger.warning("Unable to initialize GSuite creds; skipping GSuite sync.")
+        return
+
     try:
         credentials = GoogleCredentials.from_stream(GSUITE_CREDS)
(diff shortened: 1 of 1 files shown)


ramonpetgrave64:
Also, right now Cartography doesn't ingest  GCP resources that aren't attached to an `organization`, which I think is an extra paid resource. 

It might be good to create a Null, or NoOrg organization node in Cartography to attach those resources to.

achantavy:
@ramonpetgrave64 Thanks for bringing this up, I'll file an issue.

achantavy:
@ramonpetgrave64 Actually, I think we _do_ ingest GCP projects that don't tie back to an organization: https://github.com/lyft/cartography/blob/811990606c22a42791d213c7ca845b15f87e47f1/cartography/intel/gcp/crm.py#L55

As far as I can tell (correct me if I'm wrong), this doesn't rely on having an organization node.  Can you let me know what we're missing?

ramonpetgrave64:
> @ramonpetgrave64 Actually, I think we _do_ ingest GCP projects that don't tie back to an organization:
> 
> https://github.com/lyft/cartography/blob/811990606c22a42791d213c7ca845b15f87e47f1/cartography/intel/gcp/crm.py#L55
> 
> As far as I can tell (correct me if I'm wrong), this doesn't rely on having an organization node. Can you let me know what we're missing?

Projects still don't get saved because of an issue with the query. I filled an issue here. https://github.com/lyft/cartography/issues/217

Issue #212, opened 2019-11-27
Required permissions missing from documentation?


marco-lancini:
Hi,
when deploying Cartography to scan both AWS and GCP, I noticed that the following permissions, although required, are not mentioned in the README:

| CSP  | Permission | Needed for |
| ------  | --------------- | --------------- |
| AWS | `ec2:DescribeRegions` | Cartography will fail without this permission, as it won't be able to fetch all the regions used by the AWS accounts |
| GCP | `roles/resourcemanager.organizationViewer` | Needed to list/get the GCP Organization. Cartography returns an empty list if this permission is not granted |
| GCP | `roles/resourcemanager.folderViewer` | Needed to list/get Folders. Cartography returns an empty list if this permission is not granted, which means no Projects will be analysed |

If you agree on this (unless I'm missing something) I will be happy to create a PR to extend the README.

ramonpetgrave64:
At my organization, it was tricky getting it to fully work across our entire GCP account: I had to grant the service account permissions at the `organization` level. We used a specific set of required permissions, since we felt that  the `SecurityReviwer` role might have been overly permissive. 

But I agree that the docs could specify which predefined roles are needed to `list` orgs, folders, projects, instances, etc.

achantavy:
> If you agree on this (unless I'm missing something) I will be happy to create a PR to extend the README.

@marco-lancini Please do so! :-)

> But I agree that the docs could specify which predefined roles are needed to list orgs, folders, projects, instances, etc.

@ramonpetgrave64 For sure.

marco-lancini:
A first draft is now on PR #213

PR #213, opened 2019-12-02
Added required permissions missing from documentation

marco-lancini:
Address Issue #212, by extending the README with additional permissions required by Cartography and not previously mentioned

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ Cartography is a Python tool that consolidates infrastructure assets and the rel
     - [What users have the TotallyFake extension installed?](#what-users-have-the-totallyfake-extension-installed)
     - [What users have installed extensions that are risky based on CRXcavator scoring?](#what-users-have-installed-extensions-that-are-risky-based-on-crxcavator-scoring)
   - [Data Enrichment](#data-enrichment)
-- [Multiple AWS Account Setup](#multiple-aws-account-setup)
+- [Cross-Account Auditing](#cross-account-auditing)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -416,3 +431,8 @@ There are many ways to allow Cartography to pull from more than one AWS account.
 
 		... etc ...
 		```
+
+
+#### Multiple GCP Project Setup
+
+In order for Cartography to be able to pull all assets from all GCP Projects within an Organization, the User/Service Account assigned to Cartography needs to be created at the **Organization** level.
```
achantavy: Do you have a reference on this/are there other official Google links that we provide here?

README.md
```diff
@@ -381,12 +390,18 @@ There are many ways to allow Cartography to pull from more than one AWS account.
...
+				  "Effect": "Allow",
+				  "Action": "ec2:DescribeRegions",
+				  "Resource": "*"
+				}
 			  ]
 			}
 			```
 
-			This allows the Hub role to assume the `cartography-read-only` role on your Spoke accounts.
+			This allows the Hub role to assume the `cartography-read-only` role on your Spoke accounts and to fetch all the different regions used by the Spoke accounts.
```
achantavy: Ahh, good callout.

achantavy:
Also it looks like CI is angry: I think you need to do `make test_lint` and then commit and push.

README.md
marco-lancini: Good point, I've just added a reference

achantavy:
Thanks for updating the docs!

PR #219, opened 2020-01-07
Map DNS zones to their DNS servers

SecPrez:


cartography/intel/aws/route53.py
```diff
@@ -114,6 +115,70 @@ def load_zone(neo4j_session, zone, current_aws_id, update_tag):
...
+    query = """
+    match (z:AWSDNSZone)
+    <-[:MEMBER_OF_DNS_ZONE]-
+    (record:DNSRecord{type:"NS"})
+    -[:DNS_POINTS_TO]->
+    (ns:NameServer)
+    <-[:NAMESERVER]-
+    (z2)
+    WHERE record.name=z2.name AND NOT z=z2
+    MERGE (z2)<-[r:SUBZONE]-(z)
```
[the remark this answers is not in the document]
achantavy: Note to self: we should add in our dev guidelines notes about directionality in relationships.

cartography/intel/aws/route53.py
```diff
@@ -171,6 +236,18 @@ def parse_record_set(record_set, zone_id):
             }
 
 
+def parse_ns_record_set(record_set, zone_id):
+    if "ResourceRecords" in record_set:
+        servers = [record["Value"][:-1] for record in record_set["ResourceRecords"]]
+        return {
+            "zoneid": zone_id,
+            "type": "NS",
+            "name": record_set["Name"][:-1],
```
achantavy: Just add a small comment here about what `record_set["Name"][:-1]` looks like.
E.g. maybe something like:
```
# `record_set["Name"][:-1]` looks like "name.some.fqdn.net.", so this removes the trailing comma.
```

tests/integration/cartography/intel/aws/test_route53.py
```diff
@@ -0,0 +1,33 @@
...
+import tests.data.aws.route53
+
+TEST_UPDATE_TAG = 123456789
+TEST_ZONE_ID = "TESTZONEID"
+TEST_ZONE_NAME = "TESTZONENAME"
+TEST_AWS_ACCOUNTID = "AWSID"
+
+
+def test_cname(neo4j_session):
+    # test that cnames can be parsed and loaded
```
achantavy: Nit: use proper caps

tests/integration/cartography/intel/aws/test_route53.py
```diff
@@ -0,0 +1,33 @@
...
+    data = tests.data.aws.route53.CNAME_RECORD
+    parsed_data = cartography.intel.aws.route53.parse_record_set(data, TEST_ZONE_ID)
+    cartography.intel.aws.route53.load_cname_records(neo4j_session, parsed_data, TEST_UPDATE_TAG)
+
+
+def test_ns(neo4j_session):
+    # test that ns records can be parsed and loaded
```
achantavy: Nit: use proper caps

tests/integration/cartography/intel/aws/test_route53.py
```diff
@@ -0,0 +1,33 @@
...
+    data = tests.data.aws.route53.NS_RECORD
+    parsed_data = [cartography.intel.aws.route53.parse_ns_record_set(data, TEST_ZONE_ID)]
+    cartography.intel.aws.route53.load_ns_records(neo4j_session, parsed_data, TEST_ZONE_NAME, TEST_UPDATE_TAG)
+
+
+def test_zone(neo4j_session):
+    # testtest that zone are being added by zone id
```
achantavy: Nit: "testtest"? 
Also use proper caps.
Sorry for being annoying 😛

cartography/intel/aws/route53.py
```diff
@@ -9,6 +9,7 @@ def link_aws_resources(neo4j_session, update_tag):
     # find records that point to other records
     link_records = """
     MATCH (n:AWSDNSRecord) WITH n MATCH (v:AWSDNSRecord{value: n.name})
+    WHERE NOT n = v
```
achantavy: Is this possible, or do you need to do something like `WHERE NOT n.id = v.id`?

cartography/intel/aws/route53.py
SecPrez: added

tests/integration/cartography/intel/aws/test_route53.py
SecPrez: fixed

tests/integration/cartography/intel/aws/test_route53.py
SecPrez: fixed

cartography/intel/aws/route53.py
SecPrez: Yes, it is just checking that the node is not equal instead of just a property.

PR #221, opened 2020-01-08
Provide IamInstanceProfile as a member in EC2Instances - Fixes #220

### LINKED ISSUE #220: IamInstanceProfile not pulled as part of ec2
IamInstanceProfile is not available as part of the data when browsing Ec2 instances.

One of my team members figured out the right code to get it working. I'm submitting a PR on our behalf; opening this issue since there should always be an issue to go along with PRs imho.

kmcquade:
Fixes #220

We are trying to solve the question: "How many EC2 instances are publicly facing and have an Instance Profile attached?" This helps us do that.

@achantavy FYI


cartography/intel/aws/ec2.py
```diff
@@ -255,6 +255,7 @@ def load_ec2_instances(neo4j_session, data, region, current_aws_account_id, aws_
                 ImageId=instance.get("ImageId", ""),
                 SubnetId=instance.get("SubnetId", ""),
                 InstanceType=instance.get("InstanceType", ""),
+                IamInstanceProfile=instance.get("IamInstanceProfile", {}).get("Arn", ""),
```
achantavy: Would prefer to leave a field blank than add an empty string.

Change this to 
```
IamInstanceProfile=instance.get("IamInstanceProfile", {}).get("Arn")
```

and I think it should be good.  Once you do that, we'll test internally and then get back to you :)

kmcquade: Done. I appreciate the quick response!

achantavy:
Looks good, thanks for your work!

PR #222, opened 2020-01-09
#217 - Correctly ingest GCP projects without parents

### LINKED ISSUE #217: GCP: Incorrect Ingestion for projects without organizations
https://cloud.google.com/resource-manager/reference/rest/v1/projects#Project
From the gcp project items in `projects.list` method response, the `project.parent` attribute is optional, and isn't set for projects that are not under folders, projects, or organizations.

I can't verify more details now, but I think the query on line 172 fails if the `parentid` variable is unset.
https://github.com/lyft/cartography/blob/95e8e11913e2a44a4d4682506d8364a638ceac69/cartography/intel/gcp/crm.py#L138-L173

I think it might even be putting parentless projects into incorrect organizations or folders, because of variable scope. 

I'd been using this little snippet here to deal with that. 
```
for project in data:
        if project.get('parent', None):
            if project['parent']['type'] == "organization":
                query = """
                MERGE (parent:GCPOrganization{id:{ParentId}})
                ON CREATE SET parent.firstseen = timestamp()
                """
                parentid = f"organizations/{project['parent']['id']}"
            elif project['parent']['type'] == "folder":
                query = """
                MERGE (parent:GCPFolder{id:{ParentId}})
                ON CREATE SET parent.firstseen = timestamp()
                """
                parentid = f"folders/{project['parent']['id']}"
        # attempt to put into an empty org
        else:
            query = """
            MERGE (parent:GCPOrganization{id:{ParentId}})
            ON CREATE SET parent.firstseen = timestamp()
            """
            parentid = "organizations/EMPTYORG"
        query += """
        MERGE (project:GCPProject{id:{ProjectId}})
        ...
```

achantavy:
Fixes #217 - allows ingestion of GCPProjects without parent nodes.

### DIFF
--- a/cartography/intel/gcp/crm.py
+++ b/cartography/intel/gcp/crm.py
@@ -1,6 +1,7 @@
 # Google Compute Resource Manager
 # https://cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy
 import logging
+from string import Template
 
 from googleapiclient.discovery import HttpError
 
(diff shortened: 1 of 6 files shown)


achantavy:
As expected, we do fail the test: 
image

Thanks again @ramonpetgrave64 for bringing this to our attention.  Stay tuned.

achantavy:
Interested in your thoughts too @ramonpetgrave64 and @marco-lancini.

marco-lancini:
Hi! Fortunately  for us (unfortunately for this test case) we haven't faced this problem (yet).
Is there any way I could help with this?

tests/integration/cartography/intel/gcp/test_crm.py
```diff
@@ -6,20 +6,81 @@
...
         MATCH (d:GCPProject) return d.id, d.projectnumber
         """
     )
     actual_nodes = {(n['d.id'], n['d.projectnumber']) for n in nodes}
     assert actual_nodes == expected_nodes
+
+    # Expect (GCPProject{project-232323})<-[:RESOURCE]-(GCPFolder{1414})
+    #             <-[:RESOURCE]-(GCPOrganization{1337}) to be connected
+    query = """
+    MATCH (p:GCPProject{id:{ProjectId}})<-[:RESOURCE]-(f:GCPFolder)<-[:RESOURCE]-(o:GCPOrganization)
```
ramonpetgrave64: I think we can test the parentless projects here, too, with `OPTIONAL MATCH`.

cartography/intel/gcp/crm.py
```diff
@@ -136,41 +137,67 @@ def load_gcp_projects(neo4j_session, data, gcp_update_tag):
...
+            """)
+            neo4j_session.run(
+                INGEST_PARENT_TEMPLATE.safe_substitute(parent_label=parent_label),
+                ParentId=parentid,
+                ProjectId=project['projectId'],
+                ProjectNumber=project['projectNumber'],
+                DisplayName=project.get('name', None),
+                LifecycleState=project.get('lifecycleState', None),
+                gcp_update_tag=gcp_update_tag,
+            )
```
ramonpetgrave64: We could put this in a separate function `_attach_gpc_project(neo4j_session, project, gcp_update_tag)`.

Either that, or make two templates: `project_template` and `parent_template`. Then `neo4j_session.run()` can either run
```
project_template.substitute(
    sub_query=parent_template.substitute(parent_label=parent_label)
)
```
or `project_template.substitute(sub_query='')`

cartography/intel/gcp/crm.py
```diff
@@ -136,41 +137,67 @@ def load_gcp_projects(neo4j_session, data, gcp_update_tag):
...
+            project.displayname = {DisplayName},
+            project.lifecyclestate = {LifecycleState},
+            project.lastupdated = {gcp_update_tag}
+
+            MERGE (parent)-[r:RESOURCE]->(project)
+            ON CREATE SET r.firstseen = timestamp()
+            SET r.lastupdated = {gcp_update_tag}
+            """)
+            neo4j_session.run(
+                INGEST_PARENT_TEMPLATE.safe_substitute(parent_label=parent_label),
```
ramonpetgrave64: By the way, how was the neo4j driver handling the error from the query failing before? I feel like it might be safer to use `substitute()`, rather than `safe_substitute()`, since `substitute()` throws a `KeyError`, but I can't investigate right now how the driver deals with floating $s.

achantavy: Before, the error didn't even get to Neo4j because it failed at the python level:
Screen Shot 2020-01-09 at 2 37 44 PM

tests/integration/cartography/intel/gcp/test_crm.py
achantavy: Definitely can do that, but I'd rather keep that as a separate test so that it's easier to see if that case ever does fail.

cartography/intel/gcp/crm.py
achantavy: I tried out the first option of separating it out into functions - how's this look?  There's a bit of repeated code, but I'm personally ok with repetition if it's easy to understand.

ramonpetgrave64: I was thinking more like this edit here. We would ingest each project from the loop within `load_gcp_projects()`, but then the loop would also ingest a project's parent by calling `_attach_gcp_project()`.

```
def load_gcp_projects(neo4j_session, data, gcp_update_tag):
    """
    Ingest the GCP projects to Neo4j
    :param neo4j_session: The Neo4j session
    :param data: List of GCP projects; output from crm.get_gcp_projects()
    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
    :return: Nothing
    """
    query = Template("""
    MERGE (project:GCPProject{id:{ProjectId}})
    ON CREATE SET project.firstseen = timestamp()
    SET project.projectid = {ProjectId},
    project.projectnumber = {ProjectNumber},
    project.displayname = {DisplayName},
    project.lifecyclestate = {LifecycleState},
    project.lastupdated = {gcp_update_tag}
    """)
    for project in data:
        neo4j_session.run(
            query,
            ProjectId=project['projectId'],
            ProjectNumber=project['projectNumber'],
            DisplayName=project.get('name', None),
            LifecycleState=project.get('lifecycleState', None),
            gcp_update_tag=gcp_update_tag,
        )
        _attach_gcp_project(neo4j_session, project, gcp_update_tag)


def _attach_gcp_project(neo4j_session, project, gcp_update_tag):
    """
    Attach a Projects to their respective parent, as in the Resource Hierarchy.
    """
    if project['parent']['type'] == 'organization':
        parent_label = 'GCPOrganization'
    elif project['parent']['type'] == 'folder':
        parent_label = 'GCPFolder'
    else:
        raise NotImplementedError(
            "Ingestion of GCP {}s as parent nodes is currently not supported. "
            "Please file an issue at https://github.com/lyft/cartography/issues.".format(
                project['parent']['type'],
            ),
        )
    parentid = f"{project['parent']['type']}s/{project['parent']['id']}"
    template = Template("""
    MATCH (project:GCPProject{id:{ProjectId}})

    MERGE (parent:$parent_label{id:{ParentId}})
    ON CREATE SET parent.firstseen = timestamp()

    MERGE (parent)-[r:RESOURCE]->(project)
    SET r.lastupdated = {gcp_update_tag}
    """)
    neo4j_session.run(
        template.substitute(parent_label=parent_label),
        ProjectId=project['projectId'],
        ParentId=parentid,
        gcp_update_tag=gcp_update_tag,
    )
```

cartography/intel/gcp/crm.py
ramonpetgrave64: I don't recall seeing any errors in my console before, though. Maybe I just didn't set to `--verbose`?

But @marco-lancini, you likely woudn't encounter this bug if you are already using the GCP [organization resource](https://cloud.google.com/resource-manager/docs/creating-managing-organization), as all new projects would automatically go under the `organization`. But even if you've transitioned to using an `organization`, it's still possible for older projects to not belong, and be parentless.

achantavy: Yeah that's odd - if your project data didn't have a `parent` field, then the old code would have crashed with that UnboundLocalError because `query` wouldn't have been defined.

ramonpetgrave64: This was back when I was working on #189, and I did my `EMPTYORG` workaround. But I'm glad you're doing this now.

cartography/intel/gcp/crm.py
achantavy: Great suggestion.  Just revised it, how's it now?

tests/integration/cartography/intel/gcp/test_crm.py
```diff
@@ -6,20 +6,81 @@
...
+        TEST_UPDATE_TAG,
+    )
+
+    expected_nodes = {
+        ("my-parentless-project-987654"),
+    }
     nodes = neo4j_session.run(
         """
-        MATCH (d:GCPProject) return d.id, d.projectnumber
+        MATCH (d:GCPProject) WHERE NOT (d)<-[:RESOURCE]-() RETURN d.id
```
achantavy: I changed this to just return one field to improve readability of the test.

cartography/intel/gcp/crm.py
```diff
@@ -135,42 +136,62 @@ def load_gcp_projects(neo4j_session, data, gcp_update_tag):
...
             query,
-            ParentId=parentid,
             ProjectId=project['projectId'],
             ProjectNumber=project['projectNumber'],
             DisplayName=project.get('name', None),
             LifecycleState=project.get('lifecycleState', None),
             gcp_update_tag=gcp_update_tag,
         )
+        if project.get('parent'):
+            _attach_gcp_project_parent(neo4j_session, project, gcp_update_tag)
```
achantavy: @ramonpetgrave64 I did the attach thing, whatchu think?

ramonpetgrave64: Looks good, and it conforms better with the other` _attach` functions.

ramonpetgrave64:
@achantavy Thanks for doing this!

Issue #224, opened 2020-01-14
AWS EC2 instances are identified by `instanceid` and not `id`


achantavy:
**Bug Template**

*Title*: *AWS EC2 instances are identified by `instanceid` and not `id`*

*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

The standard is to represent each node with an `id` field.  AWS EC2 instances are identified by `instanceid`.  We should fix this inconsistency.

PR #225, opened 2020-01-14
#223 and #224: Use `id` as primary identifier in AWS S3Buckets and EC2Instances

### LINKED ISSUE #223: AWS buckets are identified by `name` and not `id`
**Bug Template**

*Title*: *AWS buckets are identified by `name` and not `id`*

*Description*:

The standard is to represent each node with an `id` field.  AWS buckets are identified by `name`.  We should fix this inconsistency.

achantavy:
Fixes #223 and #224.

This change standardizes S3Bucket and EC2Instance nodes so that they have `id` as a unique identifier when they are created.  Previously, these nodes used `name` and `instanceid` respectively and did not have an `id` field.

Integration tests are provided as well.


SecPrez:
+1

PR #226, opened 2020-01-16
Fix AWSDNSRecord IDs to include zoneid

SecPrez:
A CNAME DNS record for lyft.com can exist in multiple zones. Currently the id of an AWSDNSRecord is a the record name + record type. That means only one lyft.com+CNAME record is created and matched to multiple zones. This change adds the zoneid to the AWSDNSRecord id field so each record is unique to their zone.


SecPrez:
ptal @achantavy

achantavy:
Please give this PR a descriptive title and give a description what it fixes (what was the expected behavior, what was the actual behavior, how does this change fix that).

SecPrez:
> Please give this PR a descriptive title and give a description what it fixes (what was the expected behavior, what was the actual behavior, how does this change fix that).

Done

SecPrez:
ptal @achantavy

PR #232, opened 2020-01-22
fix: use regional clients for s3 buckets

### LINKED ISSUE #231: S3 buckets in `me-south-1` region fails with IllegalLocationConstraintException
*Title*: S3 buckets in `me-south-1` region fails with IllegalLocationConstraintException

*Description*:
Initial sync fails when scanning a bucket in `me-south-1` region

*To Reproduce*:
Create S3 bucket in `me-south-1` I suspect, but I don't have another account to test with at the moment.

*Logs*:
```
ERROR:cartography.sync:Unhandled exception during sync stage 'aws'
  File "/usr/local/bin/cartography", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/cartography/cli.py", line 216, in main
    return CLI(default_sync, prog='cartography').main(argv)
  File "/usr/local/lib/python3.6/site-packages/cartography/cli.py", line 196, in main
    return cartography.sync.run_with_config(self.sync, config)
  File "/usr/local/lib/python3.6/site-packages/cartography/sync.py", line 135, in run_with_config
    return sync.run(neo4j_driver, config)
  File "/usr/local/lib/python3.6/site-packages/cartography/sync.py", line 69, in run
    stage_func(neo4j_session, config)
  File "/usr/local/lib/python3.6/site-packages/cartography/intel/aws/__init__.py", line 106, in start_aws_ingestion
    _sync_multiple_accounts(neo4j_session, aws_accounts, regions, config.update_tag, common_job_parameters)
  File "/usr/local/lib/python3.6/site-packages/cartography/intel/aws/__init__.py", line 44, in _sync_multiple_accounts
    _sync_one_account(neo4j_session, boto3_session, account_id, regions, sync_tag, common_job_parameters)
  File "/usr/local/lib/python3.6/site-packages/cartography/intel/aws/__init__.py", line 22, in _sync_one_account
  File "/usr/local/lib/python3.6/site-packages/cartography/intel/aws/s3.py", line 326, in sync
    load_s3_details(neo4j_session, acl_and_policy_data_iter, current_aws_account_id, aws_update_tag)
  File "/usr/local/lib/python3.6/site-packages/cartography/intel/aws/s3.py", line 136, in load_s3_details
    for bucket, acl, policy in s3_details_iter:
  File "/usr/local/lib/python3.6/site-packages/cartography/intel/aws/s3.py", line 26, in get_s3_bucket_details
    acl = get_acl(bucket, client)
  File "/usr/local/lib/python3.6/site-packages/cartography/intel/aws/s3.py", line 54, in get_acl
    acl = client.get_bucket_acl(Bucket=bucket['Name'])
  File "/Users/drosenbloom/Library/Python/3.6/lib/python/site-packages/botocore/client.py", line 272, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/drosenbloom/Library/Python/3.6/lib/python/site-packages/botocore/client.py", line 576, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (IllegalLocationConstraintException) when calling the GetBucketAcl operation: The me-south-1 location constraint is incompatible for the region specific endpoint this request was sent to.
```
>  - Cartography release version: 0.15.0
>  - Python version: 3.6.6
>  - OS: OSX 10.4.6 (probably unrelated)

*Additional context*:

flyinbutrs:
Closes #231

### DIFF
--- a/cartography/intel/aws/s3.py
+++ b/cartography/intel/aws/s3.py
@@ -17,14 +17,40 @@ def get_s3_bucket_list(boto3_session):
     return client.list_buckets()
 
 
+def get_s3_clients(boto3_session):
+    """
+    Generates S3 clients for all active regions
+
+    Arguments:
+        boto3_session {boto3.session} -- Current AWS Session
+
+    Returns:
+        dict -- dictionary of regional s3 clients
+    """
+    ec2 = boto3_session.client('ec2')
+    resp = ec2.describe_regions()['Regions']
+    active_regions = [
+        x['RegionName'] for x in resp
+        if x['OptInStatus'] in ['opt-in-not-required', 'opted-in']
+    ]
+
+    clients = {
+        region: boto3_session.client('s3', region_name=region)
+        for region in active_regions
+    }
+    return clients
+
+
 def get_s3_bucket_details(boto3_session, bucket_data):
     """
     Iterates over all S3 buckets. Yields bucket name (string) and pairs of S3 bucket policies (JSON) and ACLs (JSON)
     """
+    clients = get_s3_clients(boto3_session)
     client = boto3_session.client('s3')
     for bucket in bucket_data['Buckets']:
-        acl = get_acl(bucket, client)
-        policy = get_policy(bucket, client)
+        region_name = client.get_bucket_location(Bucket=bucket['Name'])['LocationConstraint'] or 'us-east-1'
+        acl = get_acl(bucket, clients[region_name])
+        policy = get_policy(bucket, clients[region_name])
         yield bucket['Name'], acl, policy
 
 


achantavy:
Hi there!  I haven't had a chance to dig in deep yet, but is there any chance that this is related to this PR? https://github.com/lyft/cartography/pull/146

flyinbutrs:
Hi!

I suspect this is a similar issue. It seems (though undocumented), that s3 servers and clients in the opt-in regions must match to properly function for many API calls, including `create_bucket`, as mentioned here: https://github.com/boto/boto3/issues/781

I wanted to be thorough in testing this out, and wrote a script to test all permutations of source/target region to confirm that the s3 API will not function properly opt-in regions unless the client/server regions match. Might have gone a bit overboard, but here's my script:

```python
import boto3
import csv
import logging
import botocore.config
from uuid import uuid4
from botocore.exceptions import ClientError
from itertools import permutations

logging.basicConfig()
LOG = logging.getLogger('test_s3')
LOG.setLevel(logging.INFO)

BOTOCORECONFIG = botocore.config.Config(
    read_timeout=360,
    retries={
        'max_attempts': 10,
    },
)

def all_enabled_regions():
    ec2 = boto3.client('ec2')
    resp = ec2.describe_regions()['Regions']
    active_regions = [
        x['RegionName'] for x in resp
        if x['OptInStatus'] in ['opt-in-not-required', 'opted-in']
    ]
    return active_regions

def create_bucket_configuration(target):
    if target == 'us-east-1':
        return {}
    return {'CreateBucketConfiguration': {'LocationConstraint': target}}

def test_cross_region_capabilities(source, target):
    target_s3 = boto3.client('s3', region_name=target, config=BOTOCORECONFIG)
    source_s3 = boto3.client('s3', region_name=source, config=BOTOCORECONFIG)
    bucket_name = f"test-s3-bucket-{str(uuid4())}"
    try:
        target_s3.create_bucket(Bucket=bucket_name, **create_bucket_configuration(target))
        location = target_s3.get_bucket_location(Bucket=bucket_name)['LocationConstraint'] or 'us-east-1'
        LOG.debug(f"Created {bucket_name} in {location}")
    except ClientError as error:
        LOG.error(f"Failed creating bucket {bucket_name} in {target} for {source}: {error}")
        return "ERROR"

    try:
        source_s3.get_bucket_acl(Bucket=bucket_name)
        LOG.info(f"Success getting ACL from {source} in {target}")
        result = 'SUCCESS'
    except ClientError as error:
        LOG.error(f"Error getting ACL from {source} in {target}: {error.args[0]}")
        result = 'FAIL'

    try:
        target_s3.delete_bucket(Bucket=bucket_name)
        LOG.debug(f"Deleted {bucket_name}")
    except ClientError as error:
        LOG.error(f"Error deleting bucket {bucket_name}")

    return result

output = [['Source', 'Target', 'Result']]

for source, target in permutations(all_enabled_regions(), 2):
    result = test_cross_region_capabilities(source, target)
    output.append([source,target,result])

with open('test_s3_results.csv', 'w') as output_file:
    writer = csv.writer(output_file)
    writer.writerows(output)
```

The output confirms, if you're using an opt-in region, you must have a matching client/server region for s3 APIs to work properly. If you'd like, I can modify the code to only use specific regional s3 clients if the regions are opt-in. 

image

cartography/intel/aws/s3.py
```diff
@@ -17,14 +17,40 @@ def get_s3_bucket_list(boto3_session):
...
+        boto3_session {boto3.session} -- Current AWS Session
+
+    Returns:
+        dict -- dictionary of regional s3 clients
+    """
+    ec2 = boto3_session.client('ec2')
+    resp = ec2.describe_regions()['Regions']
+    active_regions = [
+        x['RegionName'] for x in resp
+        if x['OptInStatus'] in ['opt-in-not-required', 'opted-in']
```
achantavy: Ah I see, so this would only return regions that don't require being opted in?  I didn't know AWS had a concept like this 😅

achantavy:
@flyinbutrs Thanks for your detailed writeup!
We're first going to (omg finally) put #146 through internal testing early next week since it also deals with regions.  Would you be able to test if that change fixes the `IllegalConstraint` error?  I suspect it might not since it doesn't check for opt-in, but I'm still curious.  

After merging #146 I think we should do something to incorporate your "opt-in logic" in, as I don't think s3.py should call `describe-regions`.  

Rather, once #146 is merged and if you experience the same exception, I think it would make more sense to put that opt-in logic in `cartography.intel.aws.sync_one_account()`.

What do you think?

flyinbutrs:
Yeah, I can checkout that branch and retry, no prob. The `opted-in` region concept is new. I suspect they're trying to minimize the number of resources that need to be replicated to new regions as they keep adding more every year.

flyinbutrs:
Just ran a test, same bug is present with #146 merged in. I can rebase mine on that branch and then repush. I think it will require a slightly different tweak to get the `ec2` calls out of `s3.py`, but should still work fine. 

Do you want me to create an s3 client per region, or just for the `opted-in` regions? I can do either way.

achantavy:
@flyinbutrs Going to do some investigation and digging

achantavy:
@flyinbutrs I'm going to try to summarize things.  Admittedly I am confused, can you check if my understanding is correct?  Here goes:

Some newer AWS regions require your AWS account to be [opted in](https://aws.amazon.com/blogs/security/setting-permissions-to-enable-accounts-for-upcoming-aws-regions/) to perform actions on them.

With boto3, you can point your HTTP calls to a specific region in the `client` object, the `session` object, `credentials` file, and `config` file.  As seen in boto/boto3#781, you can define a `{CreateBucket/GetBucket/Whatever}Configuration` object to specify which region your s3 call will hit.

Cartography's s3 sync [does not specify region in the boto3 `client`, and it does not specify region in the boto3 `session`](https://github.com/lyft/cartography/blob/6803357bb8f693e8706bb329e52ffc9a3ee052e1/cartography/intel/aws/__init__.py#L22).  Therefore, Cartography's s3 sync is probably relying on the `config` or the `credentials` files, which results in confusing behavior including the problem you found.

I re-read your writeup and your code, but I'm not sure what you mean about a "client" versus "server".  

In any case, after reading that boto3 issue, I think that the right solution would be to supply a `regions` object to the s3 sync (like in [these lines](https://github.com/lyft/cartography/blob/6803357bb8f693e8706bb329e52ffc9a3ee052e1/cartography/intel/aws/__init__.py#L37-L39)).  If need be, we can limit the regions supplied to the s3 sync to only those that have been opted-in.

flyinbutrs:
client = `boto3.client`
server = the AWS managed S3 endpoint

While boto/boto3#781 is referring to the same issue, they're not actually giving the complete response. Based on my testing above, the opt-in regions require _any_ API calls made to a bucket in those regions to have the region encoded in the `client` or `session`, `config` or `credentials` files.

I think the correct situation is to do these two changes (and I'm happy to update the PR, though it may take a few days):

1. pass in the `regions` object to the s3 sync.
2. use region specific s3 clients for opted-in regions.

achantavy:
> 1. pass in the regions object to the s3 sync.
Right, and that would look like 
```
def _sync_one_account(neo4j_session, boto3_session, account_id, sync_tag, common_job_parameters):
    iam.sync(neo4j_session, boto3_session, account_id, sync_tag, common_job_parameters)

    try:
        regions = ec2.get_ec2_regions(boto3_session)
        logger.debug("Error occurred getting EC2 regions.", exc_info=True)
        logger.error(
            (
                "Failed to retrieve AWS region list, an error occurred: %s. Could not get regions for account %s."
            ),
            e,
            account_id,
        )
        return

    s3.sync(neo4j_session, boto3_session, regions, account_id, sync_tag, common_job_parameters)
    dynamodb.sync(neo4j_session, boto3_session, regions, account_id, sync_tag, common_job_parameters)
    # ... etc ...
```
(side note: we might want to dig in to way IAM doesn't need regions right now but one problem at a time lol)

Also I'm just sketching this out and it's untested and all that.

> 2. use region specific s3 clients for opted-in regions.
Yeah I agree with this change too.  We can change [ec2.get_ec2_regions()](https://github.com/lyft/cartography/blob/4ac824471cb13dd4d9dab5e23cf7a9ae9807d3d9/cartography/intel/aws/ec2.py#L22) to return only opted-in regions; maybe something like this (incorporating your code):

```
def get_ec2_regions(boto3_session):
    client = boto3_session.client('ec2')
    resp = client.describe_regions()
    active_regions = [
        x['RegionName'] for x in resp if x['OptInStatus'] in ['opt-in-not-required', 'opted-in']
    return [r['RegionName'] for r in result['Regions']]
```
Then we would need to create s3 clients with the proper regions, which makes us need to change `[s3.sync()](https://github.com/lyft/cartography/blob/4ac824471cb13dd4d9dab5e23cf7a9ae9807d3d9/cartography/intel/aws/s3.py#L318)` like this:
```
def sync(neo4j_session, boto3_session, regions, account_id, sync_tag, common_job_parameters):
	for region in regions: 
	    logger.info("Syncing S3 for account '%s' in region '%s'.", current_aws_account_id, region)
	    bucket_data = get_s3_bucket_list(boto3_session, region)

	    load_s3_buckets(neo4j_session, bucket_data, current_aws_account_id, aws_update_tag)
	    cleanup_s3_buckets(neo4j_session, common_job_parameters)

	    acl_and_policy_data_iter = get_s3_bucket_details(boto3_session, bucket_data, region)
	    load_s3_details(neo4j_session, acl_and_policy_data_iter, current_aws_account_id, aws_update_tag)
	    cleanup_s3_bucket_acl_and_policy(neo4j_session, common_job_parameters)
```

Then `[get_s3_bucket_list()](https://github.com/lyft/cartography/blob/4ac824471cb13dd4d9dab5e23cf7a9ae9807d3d9/cartography/intel/aws/s3.py#L14)` would look like 

```
def get_s3_bucket_list(boto3_session, region):
    client = boto3_session.client('s3', region_name=region)
    # NOTE no paginator available for this operation
    return client.list_buckets()
```
And then `[get_s3_bucket_details()](https://github.com/lyft/cartography/blob/4ac824471cb13dd4d9dab5e23cf7a9ae9807d3d9/cartography/intel/aws/s3.py#L20)` will do basically the same thing:
```
def get_s3_bucket_details(boto3_session, bucket_data, region):
    """
    Iterates over all S3 buckets. Yields bucket name (string) and pairs of S3 bucket policies (JSON) and ACLs (JSON)
    """
    client = boto3_session.client('s3', region_name=region)
```

What do you think?

Sorry again on my slowness, wanted to make sure I gave this a thorough look since this is a pretty important change.

PR #234, opened 2020-01-23
Change CRXcavator intel module to pull configuration from Config object rather than directly from environment.

tayasteere:
Supersedes PR 180 due to Evan's departure. 

Adds config support for CRXcavator
Documentation updates


tayasteere:
still has CLA check fail, will create as new fix

PR #260, opened 2020-03-14
Fix #259 - Make DNS node types explicit on creation and removal

### LINKED ISSUE #259: AWS DNS cleanup should only delete relationships between nodes that the AWS DNS sync created
> Expected behavior

AWS DNS cleanup should only delete relationships between DNSRecords and nodes created by the AWS DNS sync.

> Actual behavior

As seen here,
https://github.com/lyft/cartography/blob/37ede738bd2e6e87684ccc419e3f7dd314e00111/cartography/data/jobs/cleanup/aws_post_ingestion_dns_cleanup.json#L9

the cleanup job deletes `DNS_POINTS_TO` relationships from all (:DNSRecord) nodes to **any other node**, including those that were not created during this sync.  This deletes relationships that other syncs expect to exist, and can cause data confusion.

achantavy:
Previously our DNS cleanup jobs would delete all nodes connected to a `DNS_POINTS_TO` relationship.  This causes other difficult to trace bugs in syncs that depend on `DNS_POINTS_TO` relationships.  See #259 for full details.

This PR addresses that concern by making the node labels explicit on creation and explicit on clean-up.


achantavy:
cc: @lyftbryan

lyftbryan:
👍

cartography/data/jobs/cleanup/aws_account_dns_cleanup.json
```diff
@@ -6,7 +6,7 @@
     "iterationsize": 100
   },
   {
-    "query": "MATCH ()<-[r:DNS_POINTS_TO]-(:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DETACH DELETE (r) return COUNT(*) as TotalCompleted",
+    "query": "MATCH (:ESDomain:Ip:NameServer:EC2Instance:AWSDNSRecord:DNSRecord)<-[r:DNS_POINTS_TO]-(:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DETACH DELETE (r) return COUNT(*) as TotalCompleted",
```
sachafaust: Doesn't look scalable. For example, do we know how many labels can be expressed that way in Cypher Query? Another way could be to use WHERE labels(n) in []. Either way, I question the need to be explicitly define labels here and possibly  move the logic on when this cleanup occurs.

cartography/intel/dns.py
```diff
@@ -79,23 +81,24 @@ def ingest_dns_record(neo4j_session, name, value, type, update_tag, points_to_re
...
     :return: the intel graph node id for the new/merged record
     """
-    ingest = """
+    template = Template("""
     MERGE (record:DNSRecord{id: {Id}})
     ON CREATE SET record.firstseen = timestamp(), record.name = {Name}, record.type = {Type}
     SET record.lastupdated = {update_tag}, record.value = {Value}
     WITH record
-    MATCH (n{id: {PointsToId}})
+    MATCH (n:$node_label{id: {PointsToId}})
```
sachafaust: can you add notes on why the node_label is needed since the node id is specified and expected to be unique?

sachafaust:
👎 I think we can design it differently

achantavy:
@sachafaust This is not ready for review yet.

achantavy:
@sachafaust Here's some more context.  Our internal EnvoyRoute sync links DNSRecords with EnvoyRoutes like this:
```
(:AWSDNSRecord)-[:DNS_POINTS_TO]->(:EnvoyRoute)
```  

This EnvoyRoute sync happens about 3 or 4 hours after the AWS sync.

The AWS DNS sync currently cleans up stale nodes with
```
MATCH ()<-[r:DNS_POINTS_TO]-(:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) 
WHERE r.lastupdated <> {UPDATE_TAG}
DELETE r
```

The problem with this code is that it deletes `:DNS_POINTS_TO` relationships between `(:AWSDNSRecords)` and **all** other node types, including EnvoyRoutes.  This results in 3-4 hours out of our standard work hours where we do not have DNSRecords pointing to EnvoyRoutes.

We could fix this in an operational way by deploying a Neo4j read replica so that our consumers only read data from a DB instance that is not currently undergoing a sync, however I think that the right solution is to fix this in the software itself.

I have identified that the :DNS_POINTS_TO relationships are created in these places:

### cartography.intel.aws.[route53](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/route53.py)
- link_aws_resources(): `AWSDNSRecord-[:DNS_POINTS_TO]->AWSDNSRecord`
- link_aws_resources(): `AWSDNSRecord-[:DNS_POINTS_TO]->LoadBalancer`
- link_aws_resources(): `AWSDNSRecord-[:DNS_POINTS_TO]->EC2Instance`

and

### cartography.intel.[dns](https://github.com/lyft/cartography/blob/master/cartography/intel/dns.py)
- _link_ip_to_A_record(): `DNSRecord-[:DNS_POINTS_TO]->Ip`
- ingest_dns_record(): `:DNSRecord-->Anything`   (this function is currently only used by [:ESDomains](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/elasticsearch.py) but could be used by other modules as well)

My proposed fix is to be explicit in these DNS cleanup jobs.  That is, if a sync creates a relationship between node labels A and B, then it should only remove stale relationships between node labels A and B and nothing more.  An explicit approach also makes the code considerably easier to debug and helps in avoiding difficult to diagnose bugs like these.

cartography/data/jobs/cleanup/aws_post_ingestion_dns_cleanup.json
```diff
@@ -1,14 +1,40 @@
 {
   "statements": [
     {
-      "query": "MATCH (n:DNSRecord) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
+      "query": "MATCH (n:AWSDNSRecord) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
achantavy, line 4: This cleanup job runs at the [end of a single AWS account's sync](https://github.com/lyft/cartography/blob/c39a3bbfc1e8a7f8a6a04658467508b644db1c89/cartography/intel/aws/__init__.py#L48).  Other sync jobs in the future such as GCP might use the generic `:DNSRecord` label, so we should not delete all stale `:DNSRecords`.  Rather, we should only delete `:AWSDNSRecords`.

cartography/data/jobs/cleanup/aws_dns_cleanup.json
```diff
@@ -11,9 +11,28 @@
       "iterationsize": 100
     },
     {
-      "query": "MATCH ()<-[r:DNS_POINTS_TO]-(:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
+      "query": "MATCH (:AWSDNSRecord)<-[r:DNS_POINTS_TO]-(:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
```
achantavy, line 14: This cleanup job runs at the [end of the route53 module](https://github.com/lyft/cartography/blob/c39a3bbfc1e8a7f8a6a04658467508b644db1c89/cartography/intel/aws/route53.py#L338) for a single AWS account.  The route53 sync creates the following `DNS_POINTS_TO` relationships:

- `(:AWSDNSRecord)-[:DNS_POINTS_TO]->(:AWSDNSRecord)`
- `(:AWSDNSRecord)-[:DNS_POINTS_TO]->(:LoadBalancer)`
- `(:AWSDNSRecord)-[:DNS_POINTS_TO]->(:EC2Instance)`
- `(:AWSDNSRecord)-[:DNS_POINTS_TO]->(:NameServer)`

As such, this cleanup job now explicitly deletes nodes with these labels instead of deleting all node types connected to AWSDNSRecords.

cartography/data/jobs/cleanup/aws_account_dns_cleanup.json
```diff
@@ -4,11 +4,6 @@
     "query": "MATCH (n:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
     "iterative": true,
     "iterationsize": 100
```
achantavy, line 6: This cleanup job runs at the [end of a single AWS account's sync](https://github.com/lyft/cartography/blob/c39a3bbfc1e8a7f8a6a04658467508b644db1c89/cartography/intel/aws/__init__.py#L48).  This statement is no longer necessary because the nodes are now explicitly deleted in aws_dns_cleanup.json below.

sachafaust:
> @sachafaust Here's some more context. Our internal EnvoyRoute sync links DNSRecords with EnvoyRoutes like this:
> 
> ```
> (:DNSRecord)-[:DNS_POINTS_TO]->(:EnvoyRoute)
> ```
> 
> This EnvoyRoute sync happens about 3 or 4 hours after the AWS sync.
> 
> The AWS DNS sync currently cleans up stale nodes with
> 
> ```
> MATCH ()<-[r:DNS_POINTS_TO]-(:AWSDNSRecord)-[:MEMBER_OF_DNS_ZONE]->(:AWSDNSZone)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) 
> WHERE r.lastupdated <> {UPDATE_TAG}
> DELETE r
> ```
> 
> The problem with this code is that it deletes `:DNS_POINTS_TO` relationships between `(:AWSDNSRecords)` and **all** other node types, including EnvoyRoutes. This results in 3-4 hours out of our standard work hours where we do not have DNSRecords pointing to EnvoyRoutes.
> 
> We could fix this in an operational way by deploying a Neo4j read replica so that our consumers only read data from a DB instance that is not currently undergoing a sync, however I think that the right solution is to fix this in the software itself. Further, if other users of Cartography also take our approach of creating their own custom modules that rely on Route53 sync, they will also run into this bug.
> 
> I have identified that the :DNS_POINTS_TO relationships are created in these places:
> 
> ### cartography.intel.aws.[route53](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/route53.py)
> * link_aws_resources(): `AWSDNSRecord-[:DNS_POINTS_TO]->AWSDNSRecord`
> * link_aws_resources(): `AWSDNSRecord-[:DNS_POINTS_TO]->LoadBalancer`
> * link_aws_resources(): `AWSDNSRecord-[:DNS_POINTS_TO]->EC2Instance`
> 
> and
> 
> ### cartography.intel.[dns](https://github.com/lyft/cartography/blob/master/cartography/intel/dns.py)
> * _link_ip_to_A_record(): `DNSRecord-[:DNS_POINTS_TO]->Ip`
> * ingest_dns_record(): `:DNSRecord-->Anything`   (this function is currently only used by [:ESDomains](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/elasticsearch.py) but could be used by other modules as well)
> 
> My proposed fix is to be explicit in these DNS cleanup jobs. That is, if a sync creates a relationship between node labels A and B, then it should only remove stale relationships between node labels A and B and nothing more. An explicit approach also makes the code considerably easier to debug and helps in avoiding difficult to diagnose bugs like these.

Instead of using labeling which concerns me scale wise, could adding node attribute like "producer=module" help here? This would let cleanup generic and not have to manually curate labels and stick to producer attribute? Might also be useful for other cases.

achantavy:
> Instead of using labeling which concerns me scale wise, could adding node attribute like "producer=module" help here?

@sachafaust Marking the `producer` could help.  Can you describe the scale problem wrt labeling?  I'm not sure I understand.

achantavy:
Completed a big end-to-end run, looks like it works as expected as the relationship update times are all the same:
Screen Shot 2020-04-08 at 6 10 27 PM

Node update times were good too but I didn't get a screenshot.

achantavy:
Completed a big end-to-end run - there is just one relationship lastupdated time so the cleanups behave as expected.  Node update times are good too but I didn't get a screenshot.
image

sachafaust:
> > Instead of using labeling which concerns me scale wise, could adding node attribute like "producer=module" help here?
> 
> @sachafaust Marking the `producer` could help. Can you describe the scale problem wrt labeling? I'm not sure I understand.

Tks for the context with EnvoyRoute. The scale issue was how the cleanup was define using MATCH with n # of label. I see that the rewrite broke them out into individual cleanup. I think this is a case where "What you create, you clean" guideline applies. I'm good with the change.

cartography/intel/dns.py
sachafaust: ok so perf reasons. Just curiousity, would love to see perf stats on that if you have on if the indexing is applied with or without labeling.

cartography/data/jobs/cleanup/aws_account_dns_cleanup.json
sachafaust: fix in updated code and cleanup is broken out per label

cartography/intel/dns.py
achantavy: @sachafaust I just learned this the other day, it's pretty interesting!  If you `match` on an indexed ID, Neo4j performs a NodeIndexSeek and processes just 1 estimated row so performance should be fast:
image

However, without node labels Neo4j has to do an AllNodesScan, which in my case here involves over 2 million rows:
image

As Neo4j docs say, we should [avoid AllNodesScan as much as possible](https://neo4j.com/docs/cypher-manual/current/execution-plans/operators/#query-plan-all-nodes-scan).

Issue #214, opened 2019-12-10
[Feature Request] Integration with GKE/EKS


marco-lancini:
Hello! Do you have any plan regarding the ingestion of data from GKE/EKS?

achantavy:
Thanks for the suggestion!  We want visibility on k8s too but can't make any commitments at the moment.  As always, we welcome PRs and discussion.

marco-lancini:
* EKS has been integrated in PR #266 and #269.
* GKE will be next

marco-lancini:
* GKE has been integrated in PR #271

PR #266, opened 2020-03-29
Add EKS Integration

marco-lancini:
First iteration of the EKS Integration (as requested in #214), which pulls clusters and their main attributes.
Next iterations will include relations with VPCs and analysis on the data (i.e., define `internet_exposed`)


achantavy:
Yay!

cartography/intel/aws/eks.py
```diff
@@ -0,0 +1,85 @@
...
+        for page in paginator.paginate():
+            clusters.extend(page['clusters'])
+    except client.exceptions.ClientError as e:
+        # The account is not authorized to use this service in this region
+        # so we can continue without raising an exception
+        if e.response['Error']['Code'] == 'AccessDeniedException' \
+                and 'not authorized to use this service' in e.response['Error']['Message']:
+            logger.warn("{} in this region. Skipping...".format(e.response['Error']['Message']))
+        else:
+            raise(e)
```
achantavy: Can we change this to `raise`?  If I recall correctly we lose exception state if we do `raise(e)` but I could be wrong.

tests/integration/cartography/intel/aws/test_eks.py
```diff
@@ -0,0 +1,30 @@
...
+        TEST_UPDATE_TAG,
+    )
+    expected_nodes = {
+        "arn:aws:eks:eu-west-1:111111111111:cluster/cluster_1",
+        "arn:aws:eks:eu-west-2:222222222222:cluster/cluster_2",
+    }
+
+    nodes = neo4j_session.run(
+        """
+        MATCH (r:EKSCluster) RETURN r.arn;
```
achantavy: I think we should add test to show connection between AWSAccount and the EKSCluster - verify that `(:AWSAccount)--(:EKSCluster)` exists.

cartography/intel/aws/eks.py
marco-lancini: Done!

tests/integration/cartography/intel/aws/test_eks.py
marco-lancini: I've added the function `test_load_eks_clusters_relationships` below. Let me know if it's ok for you!

PR #269, opened 2020-03-31
Improve EKS Integration

marco-lancini:
This diff builds upon #266 :
* Improves cleanup job: removes stale relationships with AWSAccounts
* Add new fields: audit logging, role,  exposed_internet


cartography/intel/aws/eks.py
```diff
@@ -56,16 +59,42 @@ def load_eks_clusters(neo4j_session, cluster_data, region, current_aws_account_i
...
+    at least one entry has audit logging set to Enabled.
+    """
+    logging = False
+    cluster_logging = cluster.get('logging', {}).get('clusterLogging')
+    if cluster_logging:
+        logging = any(filter(lambda x: 'audit' in x['types'] and x['enabled'], cluster_logging))
+    return logging
+
+
+def _process_access_policy(cluster):
```
achantavy: The pattern used for marking compute instances as internet-exposed is to have it as an [analysis job](https://github.com/lyft/cartography/blob/master/docs/dev/writing-analysis-jobs.md) here: https://github.com/lyft/cartography/tree/master/cartography/data/jobs/analysis

We might want to move it there for consistency.  Any thoughts @sachafaust?

achantavy: @marco-lancini I thought about it a bit more, and I think we should move this to an analysis job.  Let's put all derived fields in analysis jobs and have the python intel modules focus on the getting and setting.

sachafaust: Seems reasonable. I would also add the endpointPublicAccess attribute on the cluster node also.

marco-lancini: Sounds good to me.

I've put it here because (after reviewing other intel modules like EC2), I had the impression that analysis jobs were used when there is the need to compute a value (e.g., for EC2 by checking security groups, etc.), whereas in this case the `internet_exposed` attribute is directly provided by the API (even if under another name)

achantavy: As a general rule of thumb, I think we should include API-provided attributes to the node in the python code (e.g. `endpointPublicAccess` or `s3.anonymous_actions`), and then set the `internet_exposed` flag only in analysis jobs.

marco-lancini: Gotcha! Will amend tomorrow :)

marco-lancini: Added an analysis job, let me know if it does sound right!

docs/schema/aws.md
```diff
@@ -790,19 +790,23 @@ Representation of an AWS EC2 [Subnet](https://docs.aws.amazon.com/AWSEC2/latest/
...
+| created_at       | The date and time the cluster was created                                                                   |
+| region           | The AWS region                                                                                              |
+| **arn**          | AWS-unique identifier for this object                                                                       |
+| id               | same as `arn`                                                                                               |
+| name             | Name of the EKS Cluster                                                                                     |
+| endpoint         | The endpoint for the Kubernetes API server.                                                                 |
+| endpoint_public_access | Indicates whether the Amazon EKS public API server endpoint is enabled                                |
+| exposed_internet | Set to True if the EKS Cluster public API server endpoint is enabled                                        |
+| rolearn          | The ARN of the IAM role that provides permissions for the Kubernetes control plane to make calls to AWS API |
```
achantavy, line 805: [comment] In future this would be neat to draw a relationship to

PR #271, opened 2020-04-02
Add GKE Integration

marco-lancini:
First iteration of the GKE Integration (as requested in #214), which pulls clusters and their main attributes.


cartography/intel/gcp/gke.py
```diff
@@ -0,0 +1,195 @@
...
+    :param container: The Container resource object created by googleapiclient.discovery.build()
+
+    :type project_id: str
+    :param project_id: The Google Project Id that you are retrieving clusters from
+
+    :rtype: Cluster Object
+    :return: Cluster response object
+    """
+    try:
+        req = container.projects().zones().clusters().list(projectId=project_id, zone='-')
```
achantavy, line 24: What does `zone='-'` do?

marco-lancini: From https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters/list#path-parameters:

> The name of the Google Compute Engine zone in which the cluster resides, or "-" for all zones.

PR #275, opened 2020-04-06
adding Docker support to cartography

nishils:
@dylancruise  and @achantavy  - This is the docker and has been appropriately rebased on master.

Let me know if you have any questions.


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.6-alpine
```
marco-lancini, line 1: Any reason why we need `3.6` rather than `3.7`+?

cartography/sync.py
```diff
@@ -93,43 +93,51 @@ def run_with_config(sync, config):
...
-        logger.debug("Error occurred during Neo4j auth.", exc_info=True)
-        if not neo4j_auth:
-            logger.error(
-                (
-                    "Unable to auth to Neo4j, an error occurred: '%s'. cartography attempted to connect to Neo4j "
-                    "without any auth. Check your Neo4j server settings to see if auth is required and, if it is, "
-                    "provide cartography with a valid username and password."
-                ),
-                e,
+
```
marco-lancini, line 96: I don't think these changes are related to Docker support.
Would it be possible to extract them into their own PR?

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+ENV HOME /home
+RUN mkdir ~/.aws && mkdir ~/app
+
+## CONFIGURE AWS
+COPY config /home/.aws/
```
marco-lancini, line 7: where is `config` defined? Can't seem to find it in this PR

run_cartography.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+echo "Starting up..."
+
+# Boto3 has a bug where passing in environment variables does not work as documented in
+# https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variables
+echo "Settings AWS Credentials File"
```
marco-lancini, line 7: As in a previous comment, what about `.aws/config`?

run_cartography.sh
```diff
@@ -0,0 +1,12 @@
...
+echo -e "[default]\naws_access_key_id=${AWS_ACCESS_KEY_ID}\naws_secret_access_key=${AWS_SECRET_ACCESS_KEY}\naws_session_token=${AWS_SESSION_TOKEN}" >> ~/.aws/credentials
+echo "Done with settings aws credentials"
+
+echo "Starting Cartography..."
+cartography --neo4j-uri bolt://neo4j:7687 --neo4j-user $NEO4J_USERNAME --neo4j-password-env-var NEO4J_PASSWORD
```
marco-lancini, line 12: shall we add `--aws-sync-all-profiles`?

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      # Then, you need to relaunch cartography process so it has the updated password
+      - NEO4J_AUTH=${NEO4J_USERNAME}/${NEO4J_PASSWORD}
+    image: neo4j:3.5
+    healthcheck:
+      test: curl -i http://127.0.0.1:7474 2>&1 | grep -c -e '200 OK'
+    ports:
+      - "7687:7687" # host:container
+      - "7474:7474"
+    volumes:
+      - $HOME/neo4j/data:/data
```
marco-lancini, line 17: This assumes users will want to store this test data in their own directory.
What do you think of using a volume instead?

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  cartography:
+    environment:
```
marco-lancini, line 20: We might want to think how to add support for other providers currently supported by Cartography (GCP, GSuite, Okta...)

Dockerfile
nishils: Nope, I didn't realize 3.7 was out. I will update.

cartography/sync.py
nishils: It is related.

Docker doesn't support waiting for containers to become healthy before spinning up the next container.

This retry mechanism allows docker to spin up cartography when neo4j is still booting up. Without this cartography will immediately fail and exit because neo4j won't have spun up yet.

Dockerfile
nishils: I thought it was needed by boto3 but on further testing, it looks like it is not needed.

run_cartography.sh
nishils: Sure, I have no idea what it does, but I can add it

run_cartography.sh
nishils: This is used by boto3 which is aws sdk that cartography uses under the hood. This file is used by that SDK

docker-compose.yml
nishils: We can update the shell file that checks for the existence of those env vars and then runs cartography with the appropriate shell commands.

I figured doing this as the MVP and then getting support for everything else next will be much easier by me or anyone else needing those features.

docker-compose.yml
nishils: Every time we spin up the docker container, cartography would have to resync against the aws account. I think this is pretty expensive to do.

I can make this another docker-compose variable possibly like `${NEO4J_VOLUME}` Thoughts?

cartography/sync.py
marco-lancini: Can we do something like this instead?
https://docs.docker.com/compose/startup-order/

nishils: Startup order does not wait for the container to be healthy. It only ensures order of initialization.

In the link you provided, it specifically says
```
The problem of waiting for a database (for example) to be ready is really just a subset of a much larger problem of distributed systems. In production, your database could become unavailable or move hosts at any time. Your application needs to be resilient to these types of failures.

To handle this, design your application to attempt to re-establish a connection to the database after a failure. If the application retries the connection, it can eventually connect to the database.
```

nishils:
Sorry for the delay. I'm taking a look at the feedback and will try to address them shortly.

achantavy:
@nishils If it's helpful, feel free to also start a shared document to spec out the feature so we can share ideas before any code is written. We can also put together a hangout and chat there as well.

docker-compose.yml
nishils: Yes, that is the goal. There may need to be some updates to the shell script to test for those env vars and apply the correct flags when running cartography

run_cartography.sh
nishils: Yea, I don't want to do this but when I only pass in the environment variables, I get the following warning.

`cartography    | WARNING:cartography.intel.aws:No valid AWS credentials could be found. No AWS accounts can be synced. Exiting AWS sync stage.`

It is really quite strange. My goal is to delete this if I can figure out what is going on here. I know the env vars are being passed properly because I can echo them from this shell script. I am trying to debug why boto3 isn't picking up aws env vars

run_cartography.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
nishils: done

run_cartography.sh
```diff
@@ -0,0 +1,12 @@
...
+
+echo "Starting up..."
+
+# Boto3 has a bug where passing in environment variables does not work as documented in
```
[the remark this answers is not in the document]
nishils: oh let me try this. This might be issue I was running into!

docs/setup/install.md
```diff
@@ -43,3 +43,14 @@ Time to set up the server that will run Cartography.  Cartography _should_ work
 	    You can view a full list of Cartography's CLI arguments by running `cartography --help`
 
 		The sync will pull data from your configured accounts and ingest data to Neo4j!  This process might take a long time if your account has a lot of assets.
+
+## Docker Deploys
```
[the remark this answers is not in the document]
nishils: done

run_cartography.sh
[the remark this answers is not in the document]
nishils: This didn't work unfortunately.

The bug is that the credentials file is respected by  boto3, but when simply passing the environment variables to the docker container, boto3 doesn't use them when it should.

achantavy:
I posted this in our public Slack but I'll repost here for visibility in case anyone else is interested.

I started on a skeleton for a Cartography Docker spec [here](https://docs.google.com/document/d/1Z26YeG6jkWjtgkgsN3Vdz9N5i5Jvxh6uBrgGa-OmQUA/edit#)

I encourage everyone to please provide input on their needs for a Docker solution in that document. In most cases we don’t use specs for PRs but since this change has the potential to help lots of users I think it’s helpful and worth capturing everyone’s ideas in a document.

To be clear, the first version of a solution does **not** need to hit all of these requirements in order to be merged, but we should plan a couple steps ahead and break down what we need into milestones to minimize needing to rearchitect a solution.

nishils:
I'm looking at the hardcoded creds issue still for aws. I got aws support to take a look and they had some helpful suggestions

achantavy:
@nishils When you get a chance, can you please fill in your ideas for the approaches in our [spec](https://docs.google.com/document/d/1Z26YeG6jkWjtgkgsN3Vdz9N5i5Jvxh6uBrgGa-OmQUA/edit#)?

PR #281, opened 2020-04-09
Allow boto3 to determine the current profile name

rsclarke-vgw:
A small fix to revert the changes that picks up the `AWS_PROFILE` environment variable and now lets boto3 decide, as highlighted in https://github.com/lyft/cartography/pull/272#discussion_r406471526 

For a single account, `boto3_session.profile_name` will contain the correct value at https://github.com/lyft/cartography/blob/713cc5b3f9bc410e61ae5f7e595b15b39ab588bf/cartography/intel/aws/organizations.py#L27

### DIFF
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -1,5 +1,4 @@
 import logging
-import os
 
 import boto3
 import botocore.exceptions
@@ -79,8 +78,7 @@ def start_aws_ingestion(neo4j_session, config):
         "UPDATE_TAG": config.update_tag,
     }
     try:
-        profile_name = os.getenv("AWS_PROFILE", default="default") or "default"
-        boto3_session = boto3.Session(profile_name=profile_name)
+        boto3_session = boto3.Session()
     except (botocore.exceptions.BotoCoreError, botocore.exceptions.ClientError) as e:
         logger.debug("Error occurred calling boto3.Session().", exc_info=True)
         logger.error(


achantavy:
@SecPrez Thoughts?

I'm think good with this PR because (1) it defers to boto3 for determining the credential source, and (2) we still have [@rsclarke-vgw's change](https://github.com/rsclarke-vgw/cartography/blob/537d3236236962b15c2f9db59c607a620fe42fa7/cartography/intel/aws/organizations.py#L27) where we don't erroneously insist on using a 'default' profile that doesn't exist in the user's config.

There is a broader question though about how we want to manage params going forward (env vars versus CLI switches versus config files) and how should we be consistent.  We can all talk about that here in this thread or in Slack or some other venue.

SecPrez:
> @SecPrez Thoughts?
> 
> I think I'm good with this PR because (1) it defers to boto3 for determining the profile/credential source, and (2) we still have [@rsclarke-vgw's other change](https://github.com/rsclarke-vgw/cartography/blob/537d3236236962b15c2f9db59c607a620fe42fa7/cartography/intel/aws/organizations.py#L27) where we don't erroneously insist on using a 'default' profile that doesn't exist in the user's config.
> 
> There is a broader question though about how we want to manage params going forward (env vars versus CLI switches versus config files) and how should we be consistent. We can all talk about that here in this thread or in Slack or some other venue.

Yes, I agree. This is what I do on my local dev environment

PR #283, opened 2020-04-13
IAM improvements and permission mapping

SecPrez:
Spec - https://docs.google.com/document/d/12N5Jf3DbCVo7d84fV7iJshLK6bGRoQl1nOe4Dynd2JE/edit#heading=h.yo6yqt7nefe2

This PR makes several improvements to the IAM sync as well as adds in permission relationship mapping between IAM and AWS resources.


cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,200 @@
...
+    # return cases
+    # (True, False) - The policy allows the action
+    # (False, False) - The poliy does not allow the action, but also doesn't explicitly deny it
+    # (False, True) - The policy specifically denies the action. There is no need to evaluate other policies
+
+    allow_statements = [s for s in statements if s["effect"] == "Allow"]
+    deny_statements = [s for s in statements if s["effect"] == "Deny"]
+    for permission in permissions:
+        if evaluate_statements_for_permission(deny_statements, permission, resource_arn):
+            return False, True
```
achantavy, line 55: From reading this function by itself, It isn't clear why returning `False, True` here represents "The policy does not allow the action, but also doesn't explicitly deny it" as explained above.

We should reconsider this set of function interfaces to make them clearer. The code should be self-documenting in that it should be clear why `False, True` is being returned from here as opposed to any other truth pair.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,200 @@
...
+def compile_statement(statements):
+    properties = ['action', 'resource', 'notresource', 'notaction']
+    for statement in statements:
+        for statement_property in properties:
+            if statement_property in statement:
+                statement[statement_property] = [compile_regex(item) for item in statement[statement_property]]
+    return statements
+
+
+def get_principals_for_account(neo4j_session, account_id):
```
achantavy, line 104: If possible we should avoid reading from the graph and then processing it through python because of the IO overhead from transferring stuff between Neo4j and python. Instead, we should try to keep all of this logic in Neo4j itself by merging on principal IDs.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,200 @@
...
+        # If the action is explicitly denied then no other policy can override it
+        if explicit_deny:
+            return False
+        if not granted and allowed:
+            granted = True
+
+    return granted
+
+
+def evaluate_relationships(principals, resource_arns, permission):
```
achantavy, line 75: Add docstring.

We might also want to reconsider the name of the function so that it describes the return output.

cartography/intel/aws/iam.py
```diff
@@ -364,11 +384,125 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
+            load_policy_statements(neo4j_session, policy_id, name, statements, aws_update_tag)
+
+
 @timeit
 def sync_users(neo4j_session, boto3_session, current_aws_account_id, aws_update_tag, common_job_parameters):
     logger.debug("Syncing IAM users for account '%s'.", current_aws_account_id)
     data = get_user_list_data(boto3_session)
     load_users(neo4j_session, data['Users'], current_aws_account_id, aws_update_tag)
+
+    policy_data = get_user_policy_data(boto3_session, data['Users'])
```
achantavy, line 500: To keep things in smaller testable chunks and for consistency, I think this should be separated out to its own sync() function.

cartography/intel/aws/iam.py
```diff
@@ -364,11 +384,125 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
+    load_policy_data(neo4j_session, policy_data, aws_update_tag)
+
+    managed_policy_data = get_user_managed_policy_data(boto3_session, data['Users'])
```
achantavy, line 503: Same here on separating out to its own separate sync() function.

cartography/intel/aws/iam.py
```diff
@@ -377,22 +511,29 @@ def sync_groups(neo4j_session, boto3_session, current_aws_account_id, aws_update
     logger.debug("Syncing IAM groups for account '%s'.", current_aws_account_id)
     data = get_group_list_data(boto3_session)
     load_groups(neo4j_session, data['Groups'], current_aws_account_id, aws_update_tag)
-    run_cleanup_job('aws_import_groups_cleanup.json', neo4j_session, common_job_parameters)
 
+    policy_data = get_group_policy_data(boto3_session, data["Groups"])
```
achantavy, line 515: Ditto

cartography/intel/aws/iam.py
```diff
@@ -377,22 +511,29 @@ def sync_groups(neo4j_session, boto3_session, current_aws_account_id, aws_update
...
+    load_policy_data(neo4j_session, policy_data, aws_update_tag)
 
-@timeit
-def sync_policies(neo4j_session, boto3_session, current_aws_account_id, aws_update_tag, common_job_parameters):
-    logger.debug("Syncing IAM policies for account '%s'.", current_aws_account_id)
-    data = get_policy_list_data(boto3_session)
-    load_policies(neo4j_session, data['Policies'], current_aws_account_id, aws_update_tag)
-    run_cleanup_job('aws_import_policies_cleanup.json', neo4j_session, common_job_parameters)
+    managed_policy_data = get_group_managed_policy_data(boto3_session, data["Groups"])
```
achantavy, line 518: Ditto

cartography/intel/aws/iam.py
```diff
@@ -377,22 +511,29 @@ def sync_groups(neo4j_session, boto3_session, current_aws_account_id, aws_update
...
 
 
 @timeit
 def sync_roles(neo4j_session, boto3_session, current_aws_account_id, aws_update_tag, common_job_parameters):
     logger.debug("Syncing IAM roles for account '%s'.", current_aws_account_id)
     data = get_role_list_data(boto3_session)
     load_roles(neo4j_session, data['Roles'], current_aws_account_id, aws_update_tag)
+
+    logger.debug("Syncing IAM role inline policies for account '%s'.", current_aws_account_id)
+    inline_policy_data = get_role_policy_data(boto3_session, data["Roles"])
```
achantavy, line 531: Same

cartography/intel/aws/iam.py
```diff
@@ -377,22 +511,29 @@ def sync_groups(neo4j_session, boto3_session, current_aws_account_id, aws_update
...
+    load_policy_data(neo4j_session, inline_policy_data, aws_update_tag)
+
+    logger.debug("Syncing IAM role managed policies for account '%s'.", current_aws_account_id)
+    managed_policy_data = get_role_managed_policy_data(boto3_session, data["Roles"])
```
achantavy, line 535: Same

cartography/intel/aws/iam.py
```diff
@@ -270,70 +294,66 @@ def load_group_memberships(neo4j_session, group_memberships, aws_update_tag):
...
-    MATCH (group:AWSGroup{arn: {GroupArn}})
-    WITH group
-    MERGE (role:AWSRole{arn: {RoleArn}})
-    ON CREATE SET role.firstseen = timestamp()
-    SET role.lastupdated = {aws_update_tag}
-    WITH role, group
-    MERGE (group)-[r:STS_ASSUMEROLE_ALLOW]->(role)
-    ON CREATE SET r.firstseen = timestamp()
-    SET r.lastupdated = {aws_update_tag}
+def get_policies_for_principal(neo4j_session, principal_arn):
```
achantavy, line 298: If possible we should avoid reading from the graph and then processing it through python because of the IO overhead from transferring stuff between Neo4j and python. Instead, we should try to keep all of this logic in Neo4j itself by merging on policy IDs.

cartography/data/permission_relationships.yaml
```diff
@@ -0,0 +1,18 @@
+# map principals who can read items in a S3Bucket
```
achantavy, line 1: I know we talked about different config formats, but we could also consider writing this simply as a python file as described in this [blog](https://beepb00p.xyz/configs-suck.html) and shown in this [popular example project](https://github.com/getpelican/pelican/blob/master/pelican/settings.py).

Anyone else have any other thoughts?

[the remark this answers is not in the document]
SecPrez: working on it now

cartography/intel/aws/iam.py
```diff
@@ -32,8 +35,77 @@ def get_group_membership_data(boto3_session, group_name):
...
+        name = group["GroupName"]
+        arn = group["Arn"]
+        resource_group = resource_client.Group(name)
+        policies[arn] = list(resource_group.policies.all())
+    return policies
+
+
+def get_group_managed_policy_data(boto3_session, group_list):
+    # Needs to be fixed to return {arn: policy list}
+    resource_client = boto3_session.resource('iam')
```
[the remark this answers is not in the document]
SecPrez: the resource clients handle pagination.

cartography/intel/aws/iam.py
```diff
@@ -270,70 +294,66 @@ def load_group_memberships(neo4j_session, group_memberships, aws_update_tag):
...
+        get_policy_query,
+        Arn=principal_arn,
+    )
+    policies = {r["policy_id"]: parse_statement_node_group(r["statements"]) for r in results}
+    return policies
 
 
 @timeit
-def load_role_policies(neo4j_session, role_policies, aws_update_tag):
+def sync_assume_role(neo4j_session, current_aws_account_id, aws_update_tag, common_job_parameters):
```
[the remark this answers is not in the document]
SecPrez: I would rather keep the name what it is modifying in the graph then how it got the data. Analyze doesn't let me know its making sync changes.

cartography/intel/aws/iam.py
SecPrez: When I profile this, the overhead is bearly measurable even in the full graph. This is a clone of exiting behavior that existed pre refactor.

cartography/intel/aws/iam.py
```diff
@@ -345,31 +362,149 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
                 )
 
 
+def ensure_list(obj):
+    if not isinstance(obj, list):
+        obj = [obj]
+    return obj
+
+
+def _generate_policy_statements(statements, policy_id):
```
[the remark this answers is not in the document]
SecPrez: fixed

cartography/intel/aws/iam.py
SecPrez: They all require the same user list. I can break out the sub parts into three functions, but the testability would be the same.

cartography/intel/aws/iam.py
```diff
@@ -446,12 +538,12 @@ def sync_user_access_keys(neo4j_session, boto3_session, current_aws_account_id,
 
 def sync(neo4j_session, boto3_session, account_id, update_tag, common_job_parameters):
     logger.info("Syncing IAM for account '%s'.", account_id)
+    # This module only syncs IAM information that is in use.
```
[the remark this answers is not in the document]
SecPrez: read the next line

cartography/intel/aws/iam.py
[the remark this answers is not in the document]
achantavy: How about `sync_assumerole_relationships()` since it creates `STS_ASSUMEROLE_ALLOW` relationships?

cartography/intel/aws/iam.py
achantavy: Fair point, I won't block on this. In future let's avoid doing this unless there's no way around it; there's lots of messy practices in the code :\

cartography/intel/aws/iam.py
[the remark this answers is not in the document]
achantavy: Oh, also can we add a comment saying that this function *must* be run after users, groups, roles, and group_memberships have been synced?

I initially didn't want to use the `sync` verb because this relies on other IAM nodes already being in the graph and the others don't. I won't block on the verb name, but I think there's a better way to communicate this.

cartography/intel/aws/iam.py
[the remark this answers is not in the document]
achantavy: Gotcha. In your end-to-end runs do the numbers of nodes match up with what you expect?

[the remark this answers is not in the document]
SecPrez: yes

cartography/intel/aws/iam.py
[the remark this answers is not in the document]
SecPrez: Done

cartography/intel/aws/iam.py
```diff
@@ -364,38 +385,198 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
+                load_policy(neo4j_session, policy_id, name, "inline", principal_arn, aws_update_tag)
+                load_policy_statements(neo4j_session, policy_id, name, statements, aws_update_tag)
+            except ClientError:
+                # Avoid crashing the sync
+                logger.warning("inline policy {name} failed with ClientError; skipping.")
+                continue
+
+
+@timeit
+def load_managed_policy_data(neo4j_session, policy_map, aws_update_tag):
```
achantavy, line 492: Same comment from `load_policy_data()`.

cartography/intel/aws/iam.py
```diff
@@ -364,38 +385,198 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
+        for policy in policy_list:
+            name = policy.policy_name
+            policy_id = policy.arn
+            try:
+                statements = _transform_policy_statements(
+                    policy.default_version.document["Statement"], policy_id,
+                )
+                load_policy(neo4j_session, policy_id, name, "managed", principal_arn, aws_update_tag)
+                load_policy_statements(neo4j_session, policy_id, name, statements, aws_update_tag)
+            except ClientError:
```
achantavy, line 503: Same comment here: will this ever be thrown? There are no boto3 calls in this function.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+            return False, True
+        else:
+            if evaluate_statements_for_permission(allow_statements, permission, resource_arn):
+                # The action is allowed by this policy
+                return True, False
+    # The action is not allowed by this policy, but not specifically denied either
+    return False, False
+
+
+def evaluate_policies_against_resource(policies, resource_arn, permissions):
```
achantavy, line 108: Reading 
```
if evaluate_policies_against_resource(policies, resource_arn, permission):
```

could probably be a bit smoother. How about something like `principal_allowed_on_resource()`? (it's pretty late and I'm running out of ideas)

cartography/intel/aws/iam.py
```diff
@@ -364,38 +385,198 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
+        for policy in policy_list:
+            name = policy.name
+            policy_id = f"{principal_arn}/inline_policy/{name}"
+            try:
+                statements = _transform_policy_statements(
+                    policy.policy_document["Statement"], policy_id,
+                )
+                load_policy(neo4j_session, policy_id, name, "inline", principal_arn, aws_update_tag)
+                load_policy_statements(neo4j_session, policy_id, name, statements, aws_update_tag)
+            except ClientError:
```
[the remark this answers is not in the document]
SecPrez: I want to get away from thinking of roles, groups and users seperatly and instead think of them as AWSPrincipals. So while the upstream requires separate methods when it get to loading it they are all just policies attached to AWSPrincipals. You are going to end up with identical functions. 
I think your concerns could be mitigated with some data checking in the load_policy function so it will fail fast on developer machines. However splitting them up means developers would make changes to role inline policies and not group inline policies which is not what I want.

cartography/intel/aws/iam.py
```diff
@@ -364,38 +385,198 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
+        injest_policy_statement,
+        PolicyId=policy_id,
+        PolicyName=policy_name,
+        Statements=statements,
+        aws_update_tag=aws_update_tag,
+    ).consume()
+
+
+@timeit
+def load_policy_data(neo4j_session, policy_map, aws_update_tag):
```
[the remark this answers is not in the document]
SecPrez: I will clean it up

cartography/intel/aws/iam.py
SecPrez: The resource clients can still throw client errors when accessing related items.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+    """ Evaluates an entire policy for specific permissions to a resource.
+    AWS Policy evaluation reference
+    https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic.html
+
+    Arguments:
+        statements {[dict]} -- The list of statements for the policy
+        permissions {[str]} -- The permissions to evaluate
+        resource_arn {[type]} -- The resource to test the permission against
+
+    Returns:
```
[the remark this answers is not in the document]
SecPrez: Doesn't exist.

cartography/intel/aws/iam.py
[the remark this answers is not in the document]
SecPrez: I want to get away from thinking of roles, groups and users seperatly and instead think of them as AWSPrincipals. So while the upstream requires separate methods when it get to loading it they are all just policies attached to AWSPrincipals. You are going to end up with identical functions.
I think your concerns could be mitigated with some data checking in the load_policy function so it will fail fast on developer machines. However splitting them up means developers would make changes to role inline policies and not group inline policies which is not what I want.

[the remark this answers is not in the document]
achantavy: My main concern is to make our call graph as simple as possible. I think each object should get its own sync function and each sync function should have its own get, transform, and loads.

> However splitting them up means developers would make changes to role inline policies and not group inline policies which is not what I want.

If we don't want developers to think of this as separate things, then it sounds like we might want to do some consolidation and combine these functions together as one.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+        [bool] -- True if the clause matched, False otherwise
+    """
+    #
+
+    result = compile_regex(clause).fullmatch(match)
+    return result is not None
+
+
+def evaluate_statement_clause_for_permission(statement, clause_name, match, missing_clause_return=False):
```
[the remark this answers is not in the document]
SecPrez: I don't see when you would ever want to use those individually. The word applies doesn't make sense. The policy can only be evaluated if each of the pieces work. Does applies mean it passes the permission test, or that it is used in evaluating the policy. The base method still has to exist to test all of the pieces. 
Furthermore I think this is very confusing when it comes to missing clause returns.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+        statements {[dict]} -- The list of statements to be evaluated
+        permission {str} -- The permission to evaluate
+        resource_arn {[type]} -- The resource to test the permission against
+
+    Returns:
+        [bool] -- If the statement grants the specific permission to the resource
+    """
+    allowed = False
+    for statement in statements:
+        if not evaluate_statement_clause_for_permission(statement, "notaction", permission):
```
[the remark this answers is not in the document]
SecPrez: It has to pass all of them. So the order is determined by how many of each type of statements we had and optimizing for speed.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+    Arguments:
+        policies {[dict]} -- The policys to evaluate
+        resource_arn {str} -- The resource to test the permission against
+        permissions {[str]} -- The permissions to evaluate
+
+    Returns:
+        [type] -- [description]
+    """
+    granted = False
+    for policy_id, statements in policies.items():
```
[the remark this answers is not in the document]
SecPrez: fixing

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+        logger.warn(f"Permission relationshp mapping file {file} not found, skipping injestion ")
+        return []
+
+
+def sync(neo4j_session, account_id, update_tag, common_job_parameters):
+    logger.info("Syncing Permission Relationships for account '%s'.", account_id)
+    principals = get_principals_for_account(neo4j_session, account_id)
+    relationship_mapping = parse_permission_relationship_file(common_job_parameters["permission_relationship_file"])
+    for rpr in relationship_mapping:
+        permissions = rpr["permissions"]
```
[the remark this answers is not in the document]
SecPrez: added in a check

cartography/intel/aws/permission_relationships.py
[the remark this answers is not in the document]
SecPrez: fixed

cartography/intel/aws/permission_relationships.py
SecPrez: done

cartography/intel/aws/permission_relationships.py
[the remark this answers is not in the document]
SecPrez: and yes, there is a doc link on the main method. None of these sub methods make any sense to call individually.

cartography/intel/aws/permission_relationships.py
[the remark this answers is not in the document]
achantavy: How about `evaluate_{notaction/action/notresource/resource}_for_permission()`? I think we should avoid that `missing_clause_return`.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+            return True
+    return False
+
+
+def evaluate_statements_for_permission(statements, permission, resource_arn):
+    """ Evaluate an entire statement for a specific permission against a resource
+
+    Arguments:
+        statements {[dict]} -- The list of statements to be evaluated
+        permission {str} -- The permission to evaluate
```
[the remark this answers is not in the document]
SecPrez: added

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,300 @@
...
+        clause_name {str} -- The Clause in the statement to evaluate.
+            Can be one of action, notaction, resource, notresource
+        match {str, re.Pattern} -- The item to match against
+
+    Keyword Arguments:
+        missing_clause_return {bool} -- If the statement doesn't contain the clause_name this is the return type
+        (default: {False})
+
+    Returns:
+        [type] -- If the specific clause_name grants access to the item
```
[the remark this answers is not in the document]
SecPrez: rewritten

cartography/intel/aws/permission_relationships.py
achantavy: Ignore  this.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,314 @@
...
+        else:
+            if evaluate_statements_for_permission(allow_statements, permission, resource_arn):
+                # The action is allowed by this policy
+                return True, False
+    # The action is not allowed by this policy, but not specifically denied either
+    return False, False
+
+
+def principal_allowed_on_resource(policies, resource_arn, permissions):
+    """ Evaluates an enture set of policies for a specific resource for a specific permission
```
ramonpetgrave64, line 109: Make these descriptions more clear that we're not actually looking at resource-based policies?

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+def parse_statement_node(node_group):
+    """ Parse a dict from group of Neo4J node
+
+    Arguments:
+        node_group {[Neo4j.Node]} -- the node to parse
+
+    Returns:
+        [dict] -- A dictionary of statements from the node
```
ramonpetgrave64, line 158: A *list* of statements?

cartography/data/jobs/cleanup/aws_import_principals_cleanup.json
```diff
@@ -1,8 +1,31 @@
...
+    },
+    {
     "query": "MATCH (:AWSAccount{id: {AWS_ID}})-[:RESOURCE]->(n:AWSPrincipal) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
     "iterative": true,
     "iterationsize": 100
-  }],
+  }
+
+],
   "name": "cleanup AWSPrincipal"
```
ramonpetgrave64, line 30: Better description here, please, since now we are doing a bit more than just cleaning up AWSPrincipal.

docs/schema/aws.md
```diff
@@ -205,22 +207,58 @@ Representation of an [AWS Policy](https://docs.aws.amazon.com/IAM/latest/APIRefe
...
 	```
-	(AWSAccount)-[AWS_POLICY]->(AWSPolicy)
+	(AWSAccount)-[RESOURCE]->(AWSPolicy)
+	```
+- `AWSPrincipals` contain `AWSPolicies`
+
+	```
+	(AWSPrincipal)-[POLICY]->(AWSPolicy)
+	```
+
```
ramonpetgrave64, line 235: Add AWSPolicyStatement

docs/schema/aws.md
```diff
@@ -205,22 +207,58 @@ Representation of an [AWS Policy](https://docs.aws.amazon.com/IAM/latest/APIRefe
...
+- An `AWSPolicy` node is defined in an `AWSAccount`.
+
+	```
+	(AWSAccount)-[RESOURCE]->(AWSPolicy)
+	```
+- `AWSPrincipals` contain `AWSPolicies`
+
+	```
+	(AWSPrincipal)-[POLICY]->(AWSPolicy)
```
ramonpetgrave64, line 261: Correct this, please.

cartography/intel/aws/iam.py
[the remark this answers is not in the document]
SecPrez: Fixed

cartography/intel/aws/permission_relationships.py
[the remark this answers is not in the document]
SecPrez: Just for you @achantavy. Just for you

docs/schema/aws.md
SecPrez: fixing

cartography/data/jobs/cleanup/aws_import_principals_cleanup.json
SecPrez: fixing

cartography/intel/aws/permission_relationships.py
SecPrez: I have updated the main method calculate_permission_relationships to be more descriptive.

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,338 @@
...
+
+    Arguments:
+        item {str} -- the item to create the regex for
+
+    Returns:
+        [re.Pattern] -- The precompiled regex pattern.
+    """
+    if isinstance(item, str):
+        item = item.replace(".", "\\.").replace("*", ".*")
+        item = re.compile(item, flags=re.IGNORECASE)
```
achantavy, line 198: As we talked about, we should surround this with `try-except` for `sre_constants.error`. 

Something like

```python
try:
    re.compile( ... )
except sreconstants.error as e:
    logger.warning("Skipping clause '%s'; could not parse regex due to error '%s'", item, e)
```

cartography/intel/aws/permission_relationships.py
```diff
@@ -0,0 +1,338 @@
...
+
+def parse_permission_relationship_file(file):
+    try:
+        if not os.path.isabs(file):
+            file = os.path.join(os.getcwd(), file)
+        with open(file) as f:
+            relationship_mapping = yaml.load(f, Loader=yaml.FullLoader)
+        return relationship_mapping
+    except FileNotFoundError:
+        logger.warn(f"Permission relationshp mapping file {file} not found, skipping injestion ")
```
achantavy, line 305: Pycharm is complaining that we should use logger.warning instead because .warn is deprecated.

cartography/intel/aws/iam.py
```diff
@@ -364,38 +392,198 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
```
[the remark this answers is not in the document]
SecPrez: Do you want combined functions or not? It makes no sense to combine this when I was forced to split others.

cartography/intel/aws/permission_relationships.py
SecPrez: fixed

cartography/intel/aws/permission_relationships.py
SecPrez: fixed

cartography/intel/aws/iam.py
```diff
@@ -364,38 +392,198 @@ def load_user_access_keys(neo4j_session, user_access_keys, aws_update_tag):
...
+            policy_id = f"{principal_arn}/inline_policy/{policy_name}"
+            load_policy(neo4j_session, policy_id, policy_name, "inline", principal_arn, aws_update_tag)
+            load_policy_statements(neo4j_session, policy_id, policy_name, statements, aws_update_tag)
+
+
+@timeit
+def load_managed_policy_data(neo4j_session, policy_map, aws_update_tag):
+    for principal_arn, policy_list in policy_map.items():
+        for policy_name, statements in policy_list.items():
+            policy_id = f"{principal_arn}/managed_policy/{policy_name}"
```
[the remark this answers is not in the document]
SecPrez: This is following the standard ARN style in which there is no plurals 
arn:aws:iam::id:"policy"/policyname
arn:aws:iam::accountid:"instance-profile"/rolename

PR #287, opened 2020-04-15
Move statsd initialization to sync.run_with_config() and add doc strings

achantavy:
Previously the statsd client would not get initialized unless cartography was run through the CLI. I've moved this initialization to `sync.run_with_config()`.  I think this fits better and this is better for consistency because this is also where we initialize the neo4j_driver.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -4,8 +4,6 @@ import logging
 import os
 import sys
 
-from statsd import StatsClient
-
 import cartography.sync
 import cartography.util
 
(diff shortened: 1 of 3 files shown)


achantavy:
With this code change it is now possible to run custom syncs and have a statsd client initialized like this:

```
import cartography.config

if __name__ == '__main__':
    sync = build_sync()   
    config = cartography.config.Config(
        neo4j_uri='bolt://localhost:7687',
        statsd_enabled=True,
        statsd_host='127.0.0.1',
        statsd_port=8125,
        statsd_prefix='my_prefix',
    )
    return cartography.sync.run_with_config(sync, config)
```

SecPrez:
+1

PR #288, opened 2020-04-17
Correctly reference config.okta_api_key

achantavy:
Fixes this crash:
```
  File \"{PATH}/cartography/intel/okta/__init__.py\", line 40, in start_okta_ingestion
    if 'okta_api_key' not in config:
    TypeError: argument of type 'Config' is not iterable"
```
I am really not sure how this didn't fail earlier. The correct field is [here](https://github.com/lyft/cartography/blob/7ed380ede47c621a10e9a5afd8192c8e92519ab0/cartography/config.py#L70).


dansec-lyft:
👍

Issue #265, opened 2020-03-28
Automatically identify new accounts/projects


marco-lancini:
*Title*: I'd like to extend Cartography so to provide automatic detection of new AWS accounts/GCP projects

*Description*:
Cartography is usually setup to scan all AWS accounts and all GCP projects within an organization. Hence, it can be extended so to leverage its drift detection capabilities to notify whenever a new AWS account or GCP project appears.

* AWS: we can leverage the [list_accounts](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/organizations.html#Organizations.Client.list_accounts) method to obtain a list of accounts within the organization
* GCP: we can leverage [gcloud projects list](https://cloud.google.com/sdk/gcloud/reference/projects/list) to do the same

The main question would be where is the best place to run this collection: as a new analysis job/intel module/or?

It is worth noting this feature can be useful as it will extend the use cases for Cartography: this will allow to use Cartography as an independent, complementary, solution to Terraform drift detection (for cases of accounts not managed by TF, and/or as an additional verification).

Any comments/ideas/objections?

achantavy:
We experimented with this and it is possible but tricky. I will share the code that we used - I think this is a cool project if we can do it right but there are a few nuances.

This process takes the following steps:

1. Use the organizations API to get all accounts accessible by the current set of AWS creds
1. Create the boto3_* objects dynamically so that we are able to access assets in these accounts
1. Access the accounts

To get step 2 working, we created a subclass of `botocore.session.Session` that allowed us to set the desired AWS accounts in memory by overriding [`Session.full_config`](https://github.com/boto/botocore/blob/68a20a9153af0a15ef67db41222dc8c4ca51274b/botocore/session.py#L342). This was necessary because the vanilla boto3 interfaces don't allow you to specify this in memory; as far as we could tell, you are forced to read from the config file on disk. Although this worked, our implementation made assumptions about the `credential_sources` and other fields. An alternate way to do this would be to write to ~/.aws/config on the filesystem, but this isn't a good option as it thrashes a user's config.

Also, many workplaces don't have all accounts available in an organization object (as much as they should!), so we would need to support both automated account discovery as well as the current flow of deferring to boto3 as much as possible.

marco-lancini:
Hi @achantavy, thank you for providing these details!
[...]

PR #298, opened 2020-05-13
Aws autodiscovery

marco-lancini:
As discussed in #265, first draft for AWS Autodiscovery


marco-lancini:
@achantavy: Something that is left to define is how to differentiate the autodiscovered accounts from the one we were already aware of.
I can think of 2 options:

1) Add a field (`autodiscovered: True`) to the autodiscovered accounts. This implies we cannot reuse the code from `organisations.py`, we will have to apply/unapply the `autodiscovered` field, etc. Overall more overhead
2) We can distinguish them implicitly cause they will not have any resource attached to them:
```
MATCH (a:AWSAccount) WHERE NOT (a)-[:RESOURCES]-() RETURN *
```

achantavy:
I'm a fan of option 2. Code looks good so far!

Issue #258, opened 2020-03-13
DynamoDB Sync fails


marco-lancini:
Hi,
I'm observing a weird behaviour with the DynamoDB sync, which fails only on one account.

As you now, I'm running Cartography against a fleet of AWS accounts, all setup in the same way using the Hub and Spoke model (as outlined in [this blogpost](https://www.marcolancini.it/2019/blog-cross-account-auditing/)).

Halfway through the ingestion, Cartography raises the following error (notice that it successfully managed to ingest DynamoDB tables for previous accounts):

```
INFO:cartography.intel.aws.dynamodb:Syncing DynamoDB for region in 'XXX' in account 'XXXX'.
ERROR:cartography.sync:Unhandled exception during sync stage 'aws'
botocore.exceptions.ClientError: An error occurred (UnrecognizedClientException) when calling the ListTables operation: The security token included in the request is invalid
Traceback (most recent call last):
  File "/usr/bin/cartography", line 11, in <module>
    load_entry_point('cartography==0.18.0', 'console_scripts', 'cartography')()
  File "/usr/lib/python3.7/site-packages/cartography/cli.py", line 259, in main
    return CLI(default_sync, prog='cartography').main(argv)
  File "/usr/lib/python3.7/site-packages/cartography/cli.py", line 239, in main
    return cartography.sync.run_with_config(self.sync, config)
  File "/usr/lib/python3.7/site-packages/cartography/sync.py", line 136, in run_with_config
    return sync.run(neo4j_driver, config)
  File "/usr/lib/python3.7/site-packages/cartography/sync.py", line 70, in run
    stage_func(neo4j_session, config)
  File "/usr/lib/python3.7/site-packages/cartography/intel/aws/__init__.py", line 107, in start_aws_ingestion
    _sync_multiple_accounts(neo4j_session, aws_accounts, config.update_tag, common_job_parameters)
  File "/usr/lib/python3.7/site-packages/cartography/intel/aws/__init__.py", line 58, in _sync_multiple_accounts
    _sync_one_account(neo4j_session, boto3_session, account_id, sync_tag, common_job_parameters)
  File "/usr/lib/python3.7/site-packages/cartography/intel/aws/__init__.py", line 37, in _sync_one_account
    dynamodb.sync(neo4j_session, boto3_session, regions, account_id, sync_tag, common_job_parameters)
  File "/usr/lib/python3.7/site-packages/cartography/intel/aws/dynamodb.py", line 99, in sync
    common_job_parameters,
[...]

PR #299, opened 2020-05-19
Handle DynamoDB collection in ap-east-1

marco-lancini:
As raised in #258, DynamoDB ingestion fails in the `ap-east-1` region.
This seems to be a known issue within botocore, and, while we wait for a fix upstream, this diff introduces an exception handling for this case.


achantavy:
Thanks for taking this on, much appreciated

PR #308, opened 2020-06-01
AWS transit gateway

via-jordan-sokolic:
Add support for AWS Transit Gateways to cartography.

Addresses https://github.com/lyft/cartography/issues/303


via-jordan-sokolic:
Will reopen after reviewing the CLA with legal 👍

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
+import logging
+
+from .util import get_botocore_config
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
```
achantavy, line 10: For the `get_*` functions, can you please add the new `@aws_handle_regions` decorator? See https://github.com/lyft/cartography/blob/cd55965dbe24fcdc10d146328b6e460d197e4670/cartography/util.py#L68.

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
...
+def get_tgw_attachments(boto3_session, region):
+    client = boto3_session.client('ec2', region_name=region, config=get_botocore_config())
+    paginator = client.get_paginator('describe_transit_gateway_attachments')
+    tgw_attachments = []
+    for page in paginator.paginate():
+        tgw_attachments.extend(page['TransitGatewayAttachments'])
+    return {"TransitGatewayAttachments": tgw_attachments}
```
achantavy, line 23: With the decorator added, the function is expected to return just a list, so this line would become:
`return tgw_attachments`.

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
...
+    for page in paginator.paginate():
+        tgw_vpc_attachments.extend(page['TransitGatewayVpcAttachments'])
+    return {"TransitGatewayVpcAttachments": tgw_vpc_attachments}
+
+
+@timeit
+def load_transit_gateways(neo4j_session, data, region, current_aws_account_id, aws_update_tag):
+    ingest_transit_gateway = """
+    MERGE (ownerAccount:AWSAccount {id: {OwnerId}})
+    ON CREATE SET ownerAccount.firstseen = timestamp(), ownerAccount.foreign = true
```
achantavy, line 40: [comment] Neat, it seems like you really understand the code 👍 ; I am interested if you've seen any foreign accounts here

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@timeit
+def get_tgw_vpc_attachments(boto3_session, region):
+    client = boto3_session.client('ec2', region_name=region, config=get_botocore_config())
+    paginator = client.get_paginator('describe_transit_gateway_vpc_attachments')
+    tgw_vpc_attachments = []
+    for page in paginator.paginate():
+        tgw_vpc_attachments.extend(page['TransitGatewayVpcAttachments'])
+    return {"TransitGatewayVpcAttachments": tgw_vpc_attachments}
```
achantavy, line 33: Same here about just doing `return tgw_vpc_attachments`

cartography/data/jobs/cleanup/aws_import_tgw_cleanup.json
```diff
@@ -0,0 +1,24 @@
+{
+  "statements": [
+  {
+    "query": "MATCH (n:AWSTransitGateway)-[r:RESOURCE|:SHARED_WITH]-() WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DETACH DELETE (r) return COUNT(*) as TotalCompleted",
```
achantavy, line 4: [consider] For readability I think it's better to be explicit and match on specific node labels rather than `()`.

I won't block on this though because there's no perf impact and the code isn't doing something that'll create bugs down the road like what I call an "unlabeled delete" where you do something like `match (n)-[:SOME_REL]->(:SomeOtherLabel) detach delete n`. We've gotten really hard to trace bugs due to this in the past.

Anyway, just something to consider.

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@timeit
+def load_transit_gateways(neo4j_session, data, region, current_aws_account_id, aws_update_tag):
```
achantavy, line 37: Please add integration tests with dummy data so that we know the new nodes are created with the expected relationships. You can copy https://github.com/lyft/cartography/blob/master/tests/integration/cartography/intel/aws/test_eks.py.

via-jordan-sokolic:
Thanks for reviewing @achantavy, I'll incorporate your notes and update soon.

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
...
+    ON CREATE SET ownerAccount.firstseen = timestamp(), ownerAccount.foreign = true
+    SET ownerAccount.lastupdated = {aws_update_tag}
+
+    MERGE (tgw:AWSTransitGateway {id: {TgwId}, arn: {ARN}})
+    ON CREATE SET tgw.firstseen = timestamp()
+    SET tgw.ownerid = {OwnerId}, tgw.state = {State},
+    tgw.description = {Description},
+    tgw.region = {Region},
+    tgw.lastupdated = {aws_update_tag}
+
```
sachafaust, line 49: add WITH tgw

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
...
+            TransitGatewayId=tgw["TransitGatewayId"],
+            AWS_ACCOUNT_ID=current_aws_account_id,
+            aws_update_tag=aws_update_tag,
+        )
+
+
+@timeit
+def load_tgw_attachments(neo4j_session, data, region, current_aws_account_id, aws_update_tag):
+    ingest_transit_gateway = """
+    MERGE (tgwa:AWSTransitGatewayAttachment{id: {TgwAttachmentId}})
```
sachafaust, line 100: q: are we sure TgwAttachementId will be unique across AWS accounts? unless specified in the api doc, I would assume not.

docs/schema/aws.md
```diff
@@ -24,66 +24,70 @@
   - [Relationships](#relationships-7)
 - [AWSPrincipal::AWSRole](#awsprincipalawsrole)
   - [Relationships](#relationships-8)
-- [AWSVpc](#awsvpc)
+- [AWSTransitGateway](#awstransitgateway)
```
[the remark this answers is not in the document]
via-jordan-sokolic: The doc seemed to be in alphabetical order, so I tried to keep within that order, but if you think it would be better to move it to the bottom I'll do that.

cartography/data/jobs/cleanup/aws_import_tgw_cleanup.json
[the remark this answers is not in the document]
via-jordan-sokolic: Makes sense. 100% due to my lack of familiarity with Cypher. Thanks for pointing that out @sachafaust

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@timeit
+def load_transit_gateways(neo4j_session, data, region, current_aws_account_id, aws_update_tag):
+    ingest_transit_gateway = """
+    MERGE (ownerAccount:AWSAccount {id: {OwnerId}})
+    ON CREATE SET ownerAccount.firstseen = timestamp(), ownerAccount.foreign = true
+    SET ownerAccount.lastupdated = {aws_update_tag}
+
+    MERGE (tgw:AWSTransitGateway {id: {TgwId}, arn: {ARN}})
```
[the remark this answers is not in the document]
via-jordan-sokolic: re: global uniqueness, not sure about Transit Gateway IDs but ARNs are globally unique, so I guess merging on the ARN alone is fine.

cartography/intel/aws/ec2/tgw.py
via-jordan-sokolic: I'm going to assume that it is, and even if it isn't, it would be astronomically unlikely that we'd have an overlap in our domain of inter-connected AWS accounts. We've built all of our infrastructure so far with this assumption and haven't had any problems, so I hope I'm right!

cartography/intel/aws/ec2/tgw.py
via-jordan-sokolic: will do

cartography/intel/aws/ec2/tgw.py
via-jordan-sokolic: Ha, I tried 😆 
re: foreign accounts, not in our case since we only use TGW attachments across our accounts, but I added it to support the case of cross-organization transit gateway attachments.

cartography/data/jobs/cleanup/aws_import_tgw_cleanup.json
via-jordan-sokolic: Good point. Those single-line queries were just getting really long. I'd even consider switching to YAML for the cleanup queries, so you could take advantage of line folding:
```yaml
statements:
- query: |
    MATCH (n:AWSTransitGateway)-[r:RESOURCE|:SHARED_WITH]-()
    WHERE r.lastupdated <> {UPDATE_TAG} 
    WITH r 
    LIMIT {LIMIT_SIZE} DETACH DELETE (r) return COUNT(*) as TotalCompleted
  iterative: true
  iteration_size: 100
```

cartography/intel/aws/ec2/tgw.py
via-jordan-sokolic: Added some basic tests just checking that TGWs and TGW attachments are added.

cartography/intel/aws/ec2/tgw.py
sachafaust: I like to see the foreign attribute being used and happy that it's used if understood. The question I have on this is are we potentially adding foreign = true to an account we just haven't sync yet but scheduled to be sync later in the flow?

Should we use the foreign flag during post sync analysis vs during ingestion?

cartography/intel/aws/ec2/tgw.py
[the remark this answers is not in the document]
sachafaust: I agree. we have a general to merge on what will be come the internal id in the graph (arn) in this case. using that approach, you would have to created specific attribute for the transit gateway id from aws (ie twgid) and update documentation. It's a cautious to avoid collisions which are hard to find in large graph.

cartography/intel/aws/ec2/tgw.py
sachafaust: could be, image some that are merging 20-30K accounts like my team aims for now. Collisions of that kind are very hard to spot and often found during incident/urgency.

cartography/intel/aws/ec2/tgw.py
sachafaust: I actually introduced the same bug initially in :) https://github.com/lyft/cartography/blob/ae6c790c9a799f381b19945a9d2902b6ac5c3ecd/cartography/intel/aws/ec2/vpc_peering.py#L84 . I think this deserve a post sync analysis vs doing it at different stages.

sachafaust: I add new bug https://github.com/lyft/cartography/issues/350 to track the foreign account. Feel free to push your code with it.I'll address it in a generic way and cleanup instances.

via-jordan-sokolic: Thanks @sachafaust. While we don't have any foreign accounts attached through TGW we do have a couple of foreign accounts peered through PCX. Since we are running cartography with `--aws-sync-all-profiles` with all org accounts in the boto config, foreign accounts really are non-org accounts for us, so this code works as expected.

cartography/intel/aws/ec2/tgw.py
via-jordan-sokolic: 20-30K 😮 how long does cartography take to run on such a huge deployment?

I really think it's exceedingly unlikely to find a collision among roughly 10e+4 objects with a 17-character hex ID providing ~ 10e+20 unique IDs.
The TGW attachment API does not return an ARN but I suppose if we really wanted to we can construct one from all given info, e.g., `f'arn:aws:ec2:{aws_region}:{ResourceOwnerId}:transit-gateway-attachment/tgw-attach-0f0f0f0f0f0f0f0f0`
(side note, I didn't find any reference to the ARN structure for a TGW attachment in AWS API docs, so I made one up)

sachafaust: Tks for clarifying so we can rely on it for the id field. For scanning large # of account, it requires a different pipeline, you can sync them in parallel but Cartography doesn't support it. Might be something we contribute back.

sachafaust:
@achantavy I'm good with the PR on my end. I'll let you approve since you have change request pending.

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@timeit
+def load_transit_gateways(neo4j_session, data, region, current_aws_account_id, aws_update_tag):
+    ingest_transit_gateway = """
+    MERGE (ownerAccount:AWSAccount {id: {OwnerId}})
+    ON CREATE SET ownerAccount.firstseen = timestamp(), ownerAccount.foreign = true
+    SET ownerAccount.lastupdated = {aws_update_tag}
+
+    MERGE (tgw:AWSTransitGateway {arn: {ARN}})
```
achantavy, line 48: Can we change this to
```
MERGE (tgw:AWSTransitGateway {id: {ARN}})  // see note 1
ON CREATE SET tgw.arn = {ARN},             // see note 2
...
SET tgw.tgw_id = {TgwId}                   // see note 3
```
?

Note 1: I realize cartography is inconsistent in this handling of IDs, but we're trying to standardize now on using `id` as the main identifier.

Note 2: tgw.arn will be equivalent to tgw.id. This is so that the user will use `id` as the unique identifier and know that `arn` was used to derive that identifier.

Note 3: It is also helpful to include an asset identifier like TGW ID separately

cartography/intel/aws/ec2/tgw.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+        _attach_shared_transit_gateway(
+            neo4j_session, tgw, region, current_aws_account_id, aws_update_tag,
+        )
+
+
+@timeit
+def _attach_shared_transit_gateway(neo4j_session, tgw, region, current_aws_account_id, aws_update_tag):
+    attach_tgw = """
+    MERGE (tgw:AWSTransitGateway {arn: {ARN}})
```
achantavy, line 85: See my other note about IDs.

cartography/intel/aws/resourcegroupstaggingapi.py
```diff
@@ -41,6 +41,11 @@ def get_bucket_name_from_arn(bucket_arn):
     'ec2:security-group': {'label': 'EC2SecurityGroup', 'property': 'id', 'id_func': get_short_id_from_ec2_arn},
     'ec2:subnet': {'label': 'EC2Subnet', 'property': 'subnetid', 'id_func': get_short_id_from_ec2_arn},
     'ec2:vpc': {'label': 'AWSVpc', 'property': 'id', 'id_func': get_short_id_from_ec2_arn},
+    'ec2:transit-gateway': {'label': 'AWSTransitGateway', 'property': 'id', 'id_func': get_short_id_from_ec2_arn},
```
achantavy, line 44: First of all I am very very impressed you were able to use this :-)

With the suggested change regarding IDs above, we can change this to 

```
'ec2:transit-gateway': {'label': 'AWSTransitGateway', 'property': 'id',},
```

cartography/intel/aws/resourcegroupstaggingapi.py
```diff
@@ -41,6 +41,11 @@ def get_bucket_name_from_arn(bucket_arn):
...
+    'ec2:transit-gateway-attachment': {
+        'label': 'AWSTransitGatewayAttachment', 'property': 'id',
+        'id_func': get_short_id_from_ec2_arn,
```
achantavy, line 47: Similar idea here: we should remove `id_func`.

In fact, I don't think `get_short_id_from_ec2_arn()` was meant for strings like the TGW IDs here. There's no observed bug because `get_short_id_from_ec2_arn("tgw-0123456789abcdef0")` will still return `"tgw-0123456789abcdef0"`, but if we're not careful in the future another PR could introduce a side effect and produce a hard to track down bug.

cartography/intel/aws/resourcegroupstaggingapi.py
via-jordan-sokolic: Ha, TBH I hadn't used this API before. Took a bit of reading but it's super convenient!

cartography/intel/aws/resourcegroupstaggingapi.py
via-jordan-sokolic: Will do

cartography/intel/aws/ec2/tgw.py
via-jordan-sokolic: Implemented in bebf6a0

PR #316, opened 2020-06-17
Exploration of AWS ingestion within opt-in regions

marco-lancini:
I went down the rabbit hole and patched (in-place) every resource affected by permission issues in regions like `ap-east-1`.

This diff (as it is) is just to show the extent of the issue, and shouldn't get merged (as I'd prefer for these errors to be handled in a single, centralised place)

### DIFF
--- a/cartography/intel/aws/ec2/auto_scaling_groups.py
+++ b/cartography/intel/aws/ec2/auto_scaling_groups.py
@@ -12,8 +12,18 @@ def get_ec2_auto_scaling_groups(boto3_session, region):
     client = boto3_session.client('autoscaling', region_name=region, config=get_botocore_config())
     paginator = client.get_paginator('describe_auto_scaling_groups')
     asgs = []
-    for page in paginator.paginate():
-        asgs.extend(page['AutoScalingGroups'])
+    try:
+        for page in paginator.paginate():
+            asgs.extend(page['AutoScalingGroups'])
+    except client.exceptions.ClientError as e:
+        # The account is not authorized to use this service in this region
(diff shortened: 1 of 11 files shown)


cartography/util.py
```diff
@@ -58,3 +63,25 @@ def timed(*args, **kwargs):
...
+    ]
+
+    def inner_function(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except botocore.exceptions.ClientError as e:
+            # The account is not authorized to use this service in this region
+            # so we can continue without raising an exception
+            if e.response['Error']['Code'] in ERROR_CODES:
+                logger.warn("{} in this region. Skipping...".format(e.response['Error']['Message']))
```
achantavy, line 83: Let's use `logger.warning()` now; I'm getting a message in Pycharm that `logger.warn()` is going to be deprecated

cartography/intel/aws/ec2/auto_scaling_groups.py
```diff
@@ -1,20 +1,22 @@
...
 @timeit
+@aws_handle_regions
 def get_ec2_auto_scaling_groups(boto3_session, region):
     client = boto3_session.client('autoscaling', region_name=region, config=get_botocore_config())
     paginator = client.get_paginator('describe_auto_scaling_groups')
     asgs = []
     for page in paginator.paginate():
         asgs.extend(page['AutoScalingGroups'])
-    return {'AutoScalingGroups': asgs}
+    return asgs
```
achantavy, line 19: Thank you for make this cleaner, haha. This has been bugging me for a while.

PR #323, opened 2020-06-24
Add initial support for AWS Redshift clusters

achantavy:
Add AWS Redshift clusters and link them to their VPC security groups.

### DIFF
--- a/cartography/intel/aws/redshift.py
+++ b/cartography/intel/aws/redshift.py
@@ -1,0 +1,118 @@
+import logging
+
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_redshift_cluster_data(boto3_session, region):
+    client = boto3_session.client('redshift', region_name=region)
+    paginator = client.get_paginator('describe_clusters')
+    clusters = []
+    for page in paginator.paginate():
+        clusters.extend(page['Clusters'])
+    return clusters
+
+
+def _make_redshift_cluster_arn(region, aws_account_id, cluster_identifier):
+    return f'arn:aws:redshift:{region}:{aws_account_id}:cluster:{cluster_identifier}'
+
+
+@timeit
+def transform_redshift_cluster_data(clusters, region, current_aws_account_id):
+    for cluster in clusters:
+        # Cluster ARN format https://docs.aws.amazon.com/redshift/latest/mgmt/redshift-iam-access-control-overview.html
+        cluster['arn'] = _make_redshift_cluster_arn(region, current_aws_account_id, cluster["ClusterIdentifier"])
+
+
+@timeit
+def load_redshift_cluster_data(neo4j_session, clusters, region, current_aws_account_id, aws_update_tag):
+    ingest_cluster = """
+    MERGE (cluster:RedshiftCluster{id: {Arn}})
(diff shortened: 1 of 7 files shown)


cartography/intel/aws/redshift.py
```diff
@@ -0,0 +1,118 @@
...
+
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_redshift_cluster_data(boto3_session, region):
+    client = boto3_session.client('redshift', region_name=region)
```
irhkang, line 11: client can be a global var so you don't need to initialize upon each call.

If you don't want to initialize on load, make this a singleton

```
_client = None

def _client():
    global _client
    if _client is None:
        _client = boto3_session.client('redshift', region_name=region)
    return _client
```

cartography/intel/aws/redshift.py
```diff
@@ -0,0 +1,118 @@
...
+        clusters.extend(page['Clusters'])
+    return clusters
+
+
+def _make_redshift_cluster_arn(region, aws_account_id, cluster_identifier):
+    return f'arn:aws:redshift:{region}:{aws_account_id}:cluster:{cluster_identifier}'
+
+
+@timeit
+def transform_redshift_cluster_data(clusters, region, current_aws_account_id):
```
irhkang, line 24: will you ever call `get_redshift_cluster_data` without this?  Aka: is the arn mandatory? If it is, it might be worth moving this into `get_redshift_cluster_data` so the functionality is more tightly coupled.  Otherwise people might not realize that they need to do this.

This method also really doesn't need a timeit as you're just doing string formatting.  There is no real way to make this faster and it also shouldn't take up much time.  If you really want to optimize, you can explore parallelizing calls using gevent, asyncio or whatever framework pleases you in `load_redshift_cluster_data`

cartography/intel/aws/redshift.py
```diff
@@ -0,0 +1,118 @@
...
+    cluster.lastupdated = {aws_update_tag},
+    cluster.region = {Region}
+    WITH cluster
+    MATCH (aa:AWSAccount{id: {AWS_ACCOUNT_ID}})
+    MERGE (aa)-[r:RESOURCE]->(cluster)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
+    """
+    for cluster in clusters:
+        cluster_create_time = str(cluster['ClusterCreateTime']) if 'ClusterCreateTime' in cluster else None
```
skiptomyliu, line 58: Can be condensed to:
cluster_create_time = str(cluster.get('ClusterCreateTime'))

skiptomyliu: Oh, ignore.  I see it now.  We want to store `None`

cartography/intel/aws/redshift.py
achantavy: Good point, I'll keep this in consideration if we decide to make these sync scripts more generalized since there's a lot of boilerplate that I'm copying over and over. Right now though I'm going to keep this part as-is since it's the same as the other sync scripts.

cartography/intel/aws/redshift.py
achantavy: Yeah good call, I'll remove. I got too eager with copypasta.

PR #324, opened 2020-06-24
Attach tags, IAM roles, and VPCs to Redshift clusters

achantavy:
Also actually set indexes for Redshift nodes.

### DIFF
--- a/cartography/intel/aws/redshift.py
+++ b/cartography/intel/aws/redshift.py
@@ -31,7 +31,8 @@ def transform_redshift_cluster_data(clusters, region, current_aws_account_id):
 def load_redshift_cluster_data(neo4j_session, clusters, region, current_aws_account_id, aws_update_tag):
     ingest_cluster = """
     MERGE (cluster:RedshiftCluster{id: {Arn}})
-    ON CREATE SET cluster.firstseen = timestamp()
+    ON CREATE SET cluster.firstseen = timestamp(),
+    cluster.arn = {Arn}
     SET cluster.availability_zone = {AZ},
     cluster.cluster_create_time = {ClusterCreateTime},
     cluster.cluster_identifier = {ClusterIdentifier},
--- a/cartography/intel/aws/redshift.py
+++ b/cartography/intel/aws/redshift.py
@@ -78,9 +78,9 @@ def load_redshift_cluster_data(neo4j_session, clusters, region, current_aws_acco
             aws_update_tag=aws_update_tag,
         )
         _attach_ec2_security_groups(neo4j_session, cluster, aws_update_tag)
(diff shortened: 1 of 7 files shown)


achantavy:
The resulting graph looks like this:
image

cartography/intel/aws/redshift.py
```diff
@@ -97,6 +99,40 @@ def _attach_ec2_security_groups(neo4j_session, cluster, aws_update_tag):
...
+
+def _attach_aws_vpc(neo4j_session, cluster, aws_update_tag):
+    attach_cluster_to_vpc = """
+    MATCH (c:RedshiftCluster{id:{ClusterArn}})
+    MERGE (v:AWSVpc{id:{VpcId}})
+    MERGE (c)-[m:MEMBER_OF_AWS_VPC]->(v)
+    ON CREATE SET m.firstseen = timestamp()
+    SET m.lastupdated = {aws_update_tag}
+    """
+    if 'VpcId' in cluster and cluster.get('VpcId'):
```
irhkang, line 127: this can be written as `if cluster.get('VpcId')`.  If `cluster['VpcId']` can't be None when it exists, `if 'VpcId' in cluster` is good enough

cartography/intel/aws/redshift.py
```diff
@@ -77,9 +78,10 @@ def load_redshift_cluster_data(neo4j_session, clusters, region, current_aws_acco
             aws_update_tag=aws_update_tag,
         )
         _attach_ec2_security_groups(neo4j_session, cluster, aws_update_tag)
+        _attach_iam_roles(neo4j_session, cluster, aws_update_tag)
+        _attach_aws_vpc(neo4j_session, cluster, aws_update_tag)
 
 
-@timeit
```
irhkang, line 82: I actually like having a timer on each individual method so you know which ones are slow.  That being said if the overall latency is really low then it might not be necessary

cartography/intel/aws/redshift.py
```diff
@@ -77,9 +78,10 @@ def load_redshift_cluster_data(neo4j_session, clusters, region, current_aws_acco
             aws_update_tag=aws_update_tag,
         )
         _attach_ec2_security_groups(neo4j_session, cluster, aws_update_tag)
```
irhkang, line 80: do these need to be run in order? If not, using gevent would make these significantly faster.  If latency isn't a big deal then no biggie but fast is nice

achantavy: For these `_attach*()` functions, no. I'll look into gevent, thanks!

achantavy: Actually, I'll experiment with gevent later on a bigger bottleneck. In this particular case the `_attach()` functions don't take much time at all.

PR #327, opened 2020-06-26
Map RedshiftClusters to their admin AWSPrincipals

achantavy:
Creates the `(:AWSPrincipal)-[:CAN_ADMINISTER]->(:RedshiftCluster)` permission relationship.


achantavy:
See this image for the result: image
This shows a statement with the `redshift:*`action. This statement is attached to a policy, which is attached to the yellow `example` AWSPrincipal in the middle. resource_permissions.py evaluated these connections and added `CAN_ADMINISTER` links from the principal to all of the RedshiftClusters.

cartography/data/permission_relationships.yaml
```diff
@@ -1,18 +1,29 @@
...
-# map principals retrieve data fron dynamodb
+# Map principals that can retrieve data from DynamoDB.
 - target_label: DynamoDBTable
   permissions:
   - dynamodb:BatchGetItem
   - dynamodb:GetItem
   - dynamodb:GetRecords
   - dynamodb:Query
   relationship_name: CAN_QUERY
+# Map principals that can administer RedshiftClusters.
```
achantavy, line 22: I realize that not every user may have a use for Redshift, but I think the perf impact to having this in here is almost negligible and it's also helpful to show a wildcard action.

cartography/data/permission_relationships.yaml
```diff
@@ -1,18 +1,29 @@
...
+- target_label: RedshiftCluster
+  permissions:
+  - redshift:*
+  - redshift:CreateClusterUser
+  - redshift:GetClusterCredentials
+  - redshift:JoinGroup
+  relationship_name: CAN_ADMINISTER
```
achantavy, line 29: I'm not sure I like this name but I am open to ideas.

PR #329, opened 2020-06-29
exception handling on NoSuchEntityException when getting user's policies

mdrdannyr:
added in exception handler. Had to add client object in order to get exceptions. Unfortunately resource objects don't support exceptions yet.


cartography/intel/aws/iam.py
```diff
@@ -74,13 +74,18 @@ def get_group_managed_policy_data(boto3_session, group_list):
...
     resource_client = boto3_session.resource('iam')
     policies = {}
     for user in user_list:
         name = user["UserName"]
         arn = user["Arn"]
         resource_user = resource_client.User(name)
-        policies[arn] = {p.name: p.policy_document["Statement"] for p in resource_user.policies.all()}
+        try:
+            policies[arn] = {p.name: p.policy_document["Statement"] for p in resource_user.policies.all()}
+        except client.exceptions.NoSuchEntityException:
```
achantavy, line 86: I think we can catch `botocore.errorfactory.NoSuchEntityException` instead. This way we don't need to instantiate a boto3 client that isn't actually used.

mdrdannyr: Yup this makes sense to me. Just seen that you've caught aws exceptions in different ways in this project - wasn't too sure which way you wanted to go with here :) I'll change to botocore :)

achantavy: @mdrdannyr Oh actually, use [this technique](https://stackoverflow.com/a/44526994) to catch the exception from a resource. My bad; it seems you can't actually import `botocore.errorfactory.*`.

cartography/intel/aws/iam.py
```diff
@@ -83,7 +82,7 @@ def get_user_policy_data(boto3_session, user_list):
         resource_user = resource_client.User(name)
         try:
             policies[arn] = {p.name: p.policy_document["Statement"] for p in resource_user.policies.all()}
-        except client.exceptions.NoSuchEntityException:
+        except resource_client.meta.client.exceptions.NoSuchEntityException:
```
mdrdannyr, line 85: Turns out you can access the exceptions from resource by using .meta. 

Removed unnecessary instantiation of client. Much nicer this way :) Good comment before 👍

cartography/intel/aws/iam.py
mdrdannyr: Yup already committed with that method. Thought it was better as well :)

cartography/intel/aws/iam.py
```diff
@@ -80,7 +80,11 @@ def get_user_policy_data(boto3_session, user_list):
         name = user["UserName"]
         arn = user["Arn"]
         resource_user = resource_client.User(name)
-        policies[arn] = {p.name: p.policy_document["Statement"] for p in resource_user.policies.all()}
+        try:
+            policies[arn] = {p.name: p.policy_document["Statement"] for p in resource_user.policies.all()}
+        except resource_client.meta.client.exceptions.NoSuchEntityException:
+            logger.warning("Could not get user's policies: %s.policies.all() failed with NoSuchEntityException; skipping.", resource_user)
```
achantavy: The CI is complaining about the line being too long lol: `cartography/intel/aws/iam.py:86: [E501] line too long (138 > 120 characters)`.

Maybe do something like this:
```
logger.warning(
    f"Could not get policies for user {resource_user} due to NoSuchEntityException; skipping."
)
```

Btw to test the linter before pushing it up to the CI, you can do `pip install -r test-requirements.txt` and then `make test_lint`.

Let me know if you have trouble with the CLA, and should be good to go after that!

mdrdannyr: Thanks for the tip. i'll fix up the linting issues - Had an extra issue with commas i didn't see. Will Push the update once linter has run successfully locally. 

Regarding CLA - I've already signed it but it still says it's failing. Not too sure what to do about this?

cartography/intel/aws/iam.py
```diff
@@ -80,7 +80,14 @@ def get_user_policy_data(boto3_session, user_list):
         name = user["UserName"]
         arn = user["Arn"]
         resource_user = resource_client.User(name)
-        policies[arn] = {p.name: p.policy_document["Statement"] for p in resource_user.policies.all()}
+        try:
+            policies[arn] = {p.name: p.policy_document["Statement"] for p in resource_user.policies.all()}
+        except resource_client.meta.client.exceptions.NoSuchEntityException:
+            logger.warning(
+                "Could not get user's policies: %s.policies.all() "
```
achantavy, line 87: Minor point: this log message for my username would look like `Could not get user's policies: achantavy.policies.all() failed with NoSuchEntityException; skipping.`. Is this intended? Something like `Could not get policies for user achantavy due to NoSuchEntityException; skipping.` might make more sense.

mdrdannyr: The full error message is as follows: 

```
WARNING:cartography.intel.aws.iam:Could not get user's policies: iam.User(name='USERNAME').policies.all() failed with NoSuchEntityException; skipping.
```

I'll change it to produce the suggest: 

```
Could not get policies for user achantavy due to NoSuchEntityException; skipping.
```

PR #331, opened 2020-06-30
Fix #330 - IAM Principal parsing

### LINKED ISSUE #330: AWS IAM: Principals are not parsed correctly for multiple principal map entries
**AWS IAM: Principals are not parsed correctly for multiple principal map entries**

*Initial Fix + Unit Test*
roshinis78@0459cb317ee23f5b99b9668c19b283544db4594e

*Description*
In AWS policy statements, principals can have more than one entry (including "AWS", "Service", "Federated", or "CanonicalUser"). If more than one entry exists in the principal object, all entries should be ingested into the database as principals. For example, for a policy with entries for "AWS" and "Service", both entries should be ingested into the database. Currently, only the first entry (in this case, with the key "AWS") is ingested.

*To Reproduce*
Try ingesting a policy statement with the following principal. Only "test-role-1" is ingested into the database.
```python
SINGLE_PRINCIPAL = {
    "AWS": "test-role-1",
    "Service": ["test-service-1", "test-service-2"],
    "Federated": "test-provider-1"
}
```

*Notes*
When I ran the pre-commit lint with `make test_lint`, trailing commas were added to files that I didn't modify, so I included these in a separate commit: roshinis78@473b185d85f9086b9c24fb0bb586b45ec0e1d6dd

*Version Info*
>  - Cartography commit hash: fb68aac03f785a67990b5c31139a539c0798b255
>  - Python version: 3.8.1

*Additional Context*
> [AWS Policy Grammar](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html)

roshinis78:
Fixes #330

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -219,8 +219,11 @@ def load_groups(neo4j_session, groups, current_aws_account_id, aws_update_tag):
         )
 
 
-@timeit
 def _parse_principal_entries(principal):
+    """ 
+    Returns a list of tuples of the form (principal_type, principal_value)
+    e.g. [('AWS', 'example-role-name'), ('Service', 'example-service')]
+    """
     principal_entries = []
     for principal_type in principal:
         principal_values = principal[principal_type]
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -225,7 +225,7 @@ def load_groups(neo4j_session, groups, current_aws_account_id, aws_update_tag):
 
 
 def _parse_principal_entries(principal):
-    """ 
+    """
     Returns a list of tuples of the form (principal_type, principal_value)
     e.g. [('AWS', 'example-role-name'), ('Service', 'example-service')]
     """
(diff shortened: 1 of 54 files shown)


cartography/intel/aws/iam.py
```diff
@@ -224,6 +224,18 @@ def load_groups(neo4j_session, groups, current_aws_account_id, aws_update_tag):
         )
 
 
+@timeit
```
achantavy, line 227: You can remove the `@timeit` here, I don't think this function is slow enough to have its own timer

cartography/intel/aws/iam.py
```diff
@@ -224,6 +224,18 @@ def load_groups(neo4j_session, groups, current_aws_account_id, aws_update_tag):
...
+def _parse_principal_entries(principal):
```
achantavy, line 228: Please add a comment to this function describing the shape of the data returned e.g.
`List of tuples of the form (principal_type, principal_value) e.g. ('AWS', 'role-name')`

tests/unit/cartography/intel/aws/test_iam.py
```diff
@@ -5,10 +5,26 @@
     "Action": "*",
 }
 
+SINGLE_PRINCIPAL = {
```
achantavy, line 8: Please add a comment with a link to the AWS datatype that this is meant to copy (the `Principal` field in a policy statement).

PR #334, opened 2020-07-02
Add lambda module

eRaMvn:
Added the AWS lambda module to cartography


cartography/intel/aws/lambda_function.py
```diff
@@ -0,0 +1,85 @@
+import logging
+
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
```
marco-lancini, line 9: Could you please add the `@aws_handle_regions` decorator?

cartography/intel/aws/lambda_function.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    Create an Lambda boto3 client and grab all the lambda functions.
+    """
+    client = boto3_session.client('lambda', region_name=region)
+    paginator = client.get_paginator('list_functions')
+    lambda_functions = []
+    for page in paginator.paginate():
+        for each_function in page['Functions']:
+            lambda_functions.append(each_function)
+    return ({'AWSLambda': lambda_functions})
```
marco-lancini, line 20: With the decorator added, the function is expected to return just a list, so this line would become:
`return lambda_functions`

cartography/intel/aws/lambda_function.py
```diff
@@ -0,0 +1,85 @@
...
+    lambda.memory = {MemorySize},
+    lambda.lastupdated = {aws_update_tag}
+    WITH lambda
+    MATCH (role:AWSPrincipal{arn: {Role}})
+    MERGE (lambda)-[r:STS_ASSUME_ROLE_ALLOW]->(role)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
+    """
+
+    for lambda_function in data["AWSLambda"]:
```
marco-lancini, line 43: following the changes in line `20`, this should become:
`    for lambda_function in data:`

tests/integration/cartography/intel/aws/test_lambda.py
```diff
@@ -0,0 +1,18 @@
...
+def test_load_lambda_functions(neo4j_session):
+    data = tests.data.aws.lambda_function.LIST_LAMBDA_FUNCTIONS
+
+    cartography.intel.aws.lambda_function.load_lambda_functions(
+        neo4j_session,
+        data,
+        TEST_REGION,
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+    )
```
achantavy, line 18: I won't block on this, but it'd be helpful to add in an assertion statement here just to check on lambda IDs.

eRaMvn: Thank you, I fixed this in the latest PR

Issue #347, opened 2020-07-10
Map Python libraries to GitHub repos


achantavy:
**Feature request template**

*Title*: *Load Python requirements.txt files from GitHub repos*

*Description*:
> Describe your idea.  Please be detailed.  If a feature request, please
describe the desired behavior, what scenario it enables, and how it
would be used.

We can add a GraphQL query like this
```
"""
query ($repo: String!, $owner: String!){
    repository(name: $repo, owner: $owner) {
        requirements3:object(expression: "$branch:requirements3.txt") {
            ... on Blob {
                text
            }
        }
        requirements:object(expression: "$branch:requirements.txt") {
            ... on Blob {
                text
            }
        }
    }
}
```

to the our [existing GraphQL query for GitHub repos](https://github.com/lyft/cartography/blob/ae6c790c9a799f381b19945a9d2902b6ac5c3ecd/cartography/intel/github/repos.py#L16), and get visibility on which Python libraries are installed on which GitHub repos.

The relationships could look like this:

```
(:GitHubRepository)-[:BRANCH]->(:GitHubBranch)-[:REQUIRES]->(:PythonLibrary)

(:GitHubRepository)-[:REQUIRES]->(:PythonLibrary)
```

achantavy:
Done in #347.

PR #358, opened 2020-07-17
#347 - Add support for linking Python libraries in GitHub repos

achantavy:
Adds support for Python libraries in GitHub repos: parses requirements.txt files from repos and creates the following new path:

```
(:GitHubRepository)-[:REQUIRES{specifier}]->(:PythonLibrary)
```

Credit to @tayasteere for writing the first version of this.


cartography/intel/github/repos.py
```diff
@@ -178,6 +196,53 @@ def _transform_repo_languages(repo_url, repo, repo_languages):
...
+
+                spec = 'Unknown'
+                if len(req.specs) > 0:
+                    # We only want the version number not the specifier
+                    spec = req.specs[0][1]
+                elif req.revision:
+                    # For dynamic imports get the version from the uri if possible
+                    spec = req.revision
+
+                out_requirements_files.append({
```
aneeshusa, line 237: Not necessary in the first version, but it would be nice to include [extras](https://www.python.org/dev/peps/pep-0508/#extras) later on, which depending on the parser you may get for free.

cartography/intel/github/repos.py
```diff
@@ -1,6 +1,8 @@
 import logging
 from string import Template
 
+import requirements
```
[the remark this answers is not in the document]
achantavy: Yeah, noticed that 2018 thing as well. I can do a quick pass on evaling the other libraries. Ideally we'd get something super lightweight (I mean, how involved could parsing be rite? :p )

@tayasteere do you remember why we went with requirements-parser?

cartography/intel/github/repos.py
aneeshusa: To be clear: I don't have a use case for extras right now. This will also likely complicate how we name graph nodes/where we put edges so probably best to save for a follow-up.

cartography/data/indexes.cypher
```diff
@@ -87,6 +87,7 @@ CREATE INDEX ON :OktaTrustedOrigin(id);
 CREATE INDEX ON :OktaAdministrationRole(id);
 CREATE INDEX ON :ProgrammingLanguage(id);
 CREATE INDEX ON :PublicIpAddress(ip);
+CREATE INDEX ON :PythonLibrary(id);
```
sachafaust, line 90: Missing Dependency node

cartography/data/jobs/cleanup/github_import_cleanup.json
```diff
@@ -53,6 +53,11 @@
     "query": "MATCH (:GitHubUser)-[r:MEMBER_OF]->(:GitHubOrganization) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
     "iterative": true,
     "iterationsize": 100
+  },
+  {
+    "query": "MATCH (:GitHubRepository)-[r:REQUIRES]->(:PythonLibrary) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
```
sachafaust, line 58: Nitpick - prefer DEPENDS ->

cartography/intel/github/repos.py
```diff
@@ -292,6 +357,30 @@ def load_github_owners(session, update_tag, repo_owners):
...
+    UNWIND {Requirements} AS req
+        MERGE (lib:PythonLibrary:Dependency{id: req.id})
+        ON CREATE SET lib.firstseen = timestamp(),
+        lib.name = req.name
+        SET lib.version = req.version,
+        lib.uri = req.uri,
+        lib.lastupdated = {UpdateTag}
+
+        WITH lib, req
+        MATCH (repo:GitHubRepository{id: req.repo_url})
```
sachafaust, line 372: Please give context on why MATCH is used here vs MERGE. It's unclear to me where the node would be created in the sync.

sachafaust (dismissed):

tests/data/github/repos.py
```diff
@@ -30,6 +30,7 @@
             'login': 'example_org',
             '__typename': 'Organization',
         },
+        'requirements': {'text': 'cartography\nhttplib2>=0.7.0\njinja2\nlxml\n'},
```
aneeshusa, line 33: I would recommend having some `==` in here and less unpinned deps as requirements.txt files should have fully pinned deps (as the output from piptools-compile). Still reasonable to have test cases for unpinned/loosely bound since (in OSS/industry) not everyone pins everything in requirements.txt and we should degrade gracefully.

achantavy: Good idea, I'll add these test cases.

cartography/intel/github/repos.py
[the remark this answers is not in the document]
achantavy: I dug into using pip's internal `parse_requirements`, and maybe I need a bit more time with it but it seemed like a lot of work to [make a pip session](https://github.com/pypa/pip/blob/89a51a6fefec826256fb334ea6244dfb0b3455a0/src/pip/_internal/req/req_file.py#L129) in order to [actually call the line parsing logic](https://github.com/pypa/pip/blob/89a51a6fefec826256fb334ea6244dfb0b3455a0/src/pip/_internal/req/req_file.py#L395).

`packaging` could work:
image

I'll play with this and hopefully it'll go smoothly.

>  but doesn't handle e.g. editable installs

What does editable installs mean in this context? Is that a syntax thing in a requirements.txt?

Oh one more question, on a `PythonLibrary` node, should we set the `version` of a library (as currently implemented, e.g. `SET lib.version = "0.1.0"`) or should we set the `specifier` instead (e.g. `SET lib.specifier = "==0.1.0")? I think the latter is probably closer to being correct but I don't feel strongly one way or the other.

cartography/intel/github/repos.py
achantavy: > It's unclear to me where the node would be created in the sync.

I moved out all the `load*` functions into a function called `load()` below to hopefully make this more clear. We call `load_github_repos()` first to load repos so by the time `load_python_requirements()` is called, the repo data is already in the graph and we can just match on them.

tests/data/github/repos.py
achantavy: Added, see the test case below for SampleRepo2.

tests/data/github/repos.py
```diff
@@ -60,69 +61,6 @@
             'url': 'https://github.com/example_org',
             'login': 'example_org', '__typename': 'Organization',
         },
+        'requirements': {'text': 'cartography==0.1.0\nhttplib2>=0.7.0\njinja2\nlxml\n'},
     },
 ]
-
-TRANSFORMED_REPOS_DATA = {
```
achantavy, line 66: I removed this because it's unfeasible to maintain and doesn't add much value in testing; we want to test the whole set of ETL steps anyway.

cartography/intel/github/repos.py
```diff
@@ -49,11 +51,23 @@
                         login
                         __typename
                     }
+                    requirements:object(expression: "HEAD:requirements.txt") {
+                        ... on Blob {
+                            text
+                        }
+                    }
+                    requirements3:object(expression: "HEAD:requirements3.txt") {
```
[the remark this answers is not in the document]
achantavy: My plan is to keep requirements.txt parsing standard in open source, and since requirements3.txt is just a Lyft thing we'll roll our own function to do that internally; we already have patterns in place for this sort of customization. I do want to keep requirements.txt parsing in OSS even if not all users use Python because there is negligible perf difference if they don't have requirements.txts and it shows them an example on how they can parse blobs in GitHub repos.

cartography/intel/github/repos.py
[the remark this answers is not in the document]
achantavy: I went ahead and changed this from `version` to `specifier` since I think this makes more sense; see the new schema.md.

aneeshusa:
Per our discussion today, one thing to think about will be having a concept/node for a particular library in general (not attached to a version) so it can be queried - e.g. "show all versions for a library" or "list all libraries in the graph".

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -186,3 +186,27 @@ def test_repository_to_collaborators(neo4j_session):
...
+    """
+    # Create the path (:Repo1)-[:REQUIRES]->(:PythonLibrary{'Cartography'})<-[:REQUIRES]-(:Repo1),
+    # and test that exactly 1 repo is connected to the PythonLibrary.
+    _ensure_local_neo4j_has_test_data(neo4j_session)
+
+    nodes = neo4j_session.run("""
+    MATCH (repo:GitHubRepository)-[:REQUIRES]->(lib:PythonLibrary{name:'cartography'})
+    RETURN lib.id
+    """)
+    print(list(n['lib.id'] for n in nodes))
```
aneeshusa, line 203: Looks like a debug print?

cartography/intel/github/repos.py
```diff
@@ -214,6 +226,58 @@ def _transform_collaborators(collaborators, repo_url, transformed_collaborators)
...
+    Performs data transformations for the requirements.txt files in a GitHub repo, if available.
+    :param repo_object: The repo object.
+    :param out_requirements_files: Output array to append transformed results to.
+    :return: Nothing.
+    """
+    req_file_contents = repo_object['requirements']
+    if req_file_contents and req_file_contents.get('text'):
+        # Remove --no-binary tags if present
+        text_contents = req_file_contents['text']
+        text_contents.replace('--no-binary ', '')
```
aneeshusa, line 240:
```suggestion
        text_contents = req_file_contents['text']
```
Non-requirements like a `--no-binary` line should already get skipped by the way we ignore lines that fail to parse as a Requirement.

cartography/intel/github/repos.py
```diff
@@ -214,6 +226,58 @@ def _transform_collaborators(collaborators, repo_url, transformed_collaborators)
...
+
+            if len(req.specifier._specs) == 1:
+                # req.specifier._specs is a frozenset so to manipulate it we turn it into a list and get the 1st item.
+                spec_str = str(list(req.specifier._specs)[0])
+                # If the spec is pinned to 1 version with ==, we only want the version number, not the full specifier.
+                spec = spec_str.strip('==') if spec_str.startswith('==') else spec
+
+            out_requirements_files.append({
+                "id": f"{req.name}|{spec}",
+                "name": req.name,
```
aneeshusa, line 274: We will want to canonicalize the package name for the node; packaging has a function we can leverage here: https://packaging.pypa.io/en/latest/utils/#packaging.utils.canonicalize_name.
This ensures multiple repos depending on the same version of a library map to the same node even if they spell (e.g. capitalize) it differently.

Per our chat we discussed potentially putting the specifier as an attribute on the edge; could be reasonable to leave the original naming (and e.g. markers) there.

aneeshusa (commented):
Per our chat today, we decided to have the :PythonLibrary nodes have concrete/pinned `version`s and not `specifier`s, which we decided we could have as (non-indexable) attributes on the edges (maybe add in V2). For requirements where we are unable to parse a pin, we decided:
- We could skip them for now
- Later on, we could record an edge to a :PythonLibrary node with a null/empty/missing version, and save the specifier as an attribute on the edge.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -186,3 +186,27 @@ def test_repository_to_collaborators(neo4j_session):
...
+
+    query = """
+    MATCH (repo:GitHubRepository)-[:REQUIRES]->(lib:PythonLibrary{id:'cartography|0.1.0'})
```
aneeshusa, line 206: It would be good to have some tests that cover non-pinned dependencies and how our code handles them (e.g. adding an edge to a node with no version, and potentially recording the specifier on the edge).

cartography/intel/github/repos.py
```diff
@@ -214,6 +227,62 @@ def _transform_collaborators(collaborators, repo_url, transformed_collaborators)
...
+                # req.specifier._specs is a frozenset so to manipulate it we turn it into a list and get the 1st item.
+                spec_str = str(list(req.specifier._specs)[0])
+                # If the spec is pinned to 1 version with ==, we only want the version number, not the full specifier.
+                spec = spec_str.strip('==') if spec_str.startswith('==') else spec
+
+            # Ingest `None` to the graph instead of empty string.
+            if spec == '':
+                spec = None
+
+            canon_name = canonicalize_name(req.name)
```
achantavy, line 275: @aneeshusa Used `canonicalize_name()`.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -186,3 +186,44 @@ def test_repository_to_collaborators(neo4j_session):
     actual_nodes = {n['collab_count'] for n in nodes}
     expected_nodes = {5}
     assert actual_nodes == expected_nodes
+
+
+def test_pinned_python_library_to_repo(neo4j_session):
```
achantavy, line 191: @aneeshusa Added tests for pinned and unpinned libs below

cartography/intel/github/repos.py
```diff
@@ -363,9 +432,33 @@ def load(neo4j_session, common_job_parameters, repo_data):
...
+        lib.name = req.name
+        SET lib.url = req.url,
+        lib.lastupdated = {UpdateTag}
+
+        WITH lib, req
+        MATCH (repo:GitHubRepository{id: req.repo_url})
+        MERGE (repo)-[r:REQUIRES]->(lib)
+        ON CREATE SET r.firstseen = timestamp()
+        SET r.lastupdated = {UpdateTag},
+        r.specifier = req.specifier
```
achantavy, line 453: @aneeshusa I moved the specifier from the PythonLibrary node to the `:REQUIRES` edge. Let me know if you think if there's a better name for it. 

Right now if the specifier is a pinned version, we trim the leading `==`. If the specifier is something more complicated, we leave the whole string in there. If the specifier does not exist, we leave it as None.

cartography/intel/github/repos.py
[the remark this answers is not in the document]
aneeshusa: Sounds good.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -186,3 +186,44 @@ def test_repository_to_collaborators(neo4j_session):
...
+    WHERE r.specifier = "0.1.0"
+    RETURN count(repo) as repo_count
+    """
+    nodes = neo4j_session.run(query)
+    actual_nodes = {n['repo_count'] for n in nodes}
+    expected_nodes = {1}
+    assert actual_nodes == expected_nodes
+
+
+def test_upinned_python_library_to_repo(neo4j_session):
```
aneeshusa, line 211: Would be good to have a test case for a URL-based dep too, to make sure we ignore it OK (doesn't make it into the graph) without erroring out. (packaging will throw an InvalidRequirement and we'll skip it).

cartography/intel/github/repos.py
```diff
@@ -214,6 +227,62 @@ def _transform_collaborators(collaborators, repo_url, transformed_collaborators)
...
+            if spec == '':
+                spec = None
+
+            canon_name = canonicalize_name(req.name)
+
+            out_requirements_files.append({
+                "id": canon_name,
+                "name": canon_name,
+                "specifier": spec,
+                "url": req.url,
```
aneeshusa, line 281: IMO we shouldn't support this (at least in V1) - we don't use this internally. (We do have some URLs in our requirements files but it's actually a slightly different format to this.)

aneeshusa (commented):
Didn't look at the tests yet.

cartography/intel/github/repos.py
```diff
@@ -214,6 +227,62 @@ def _transform_collaborators(collaborators, repo_url, transformed_collaborators)
...
+                    parsed_list.append(req)
+            except ValueError as e:
+                logger.info(
+                    f"Failed to parse line \"{line}\" in repo {repo_object['url']}'s requirements.txt; skipping. "
+                    f"Details: {e}.",
+                )
+                continue
+
+        for req in parsed_list:
+            if not req.name:
```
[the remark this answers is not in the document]
achantavy: Yeah good catch; I re-tested it and it wasn't needed now that we use `packaging.Requirement`.

aneeshusa (approved):

PR #359, opened 2020-07-17
Fix #348 - Connect AWS Lambda functions to their associated AWSAccount nodes

### LINKED ISSUE #348: AWSLambda nodes aren't connected to other nodes
**Bug Template**

*Title*: *AWSLambda nodes aren't connected to other nodes*

*Description*:

```
MATCH (a:AWSLambda)-[r:STS_ASSUME_ROLE_ALLOW]->(n) return * limit 30
```
Returns nothing, although we would expect to see the IAM roles they are able to assume.

*To Reproduce*:
Run the code.

ae6c790c9a799f381b19945a9d2902b6ac5c3ecd

We should ensure that they are connected at minimum to their accounts and verify that they are connected to their IAM roles.

eRaMvn:
Fixed the current missing link

### DIFF
--- a/cartography/intel/aws/lambda_function.py
+++ b/cartography/intel/aws/lambda_function.py
@@ -36,6 +36,9 @@ def load_lambda_functions(neo4j_session, data, region, current_aws_account_id, a
     lambda.memory = {MemorySize},
     lambda.lastupdated = {aws_update_tag}
     WITH lambda
+    MATCH (owner:AWSAccount{id: {AWS_ACCOUNT_ID}})
+    MERGE (owner)-[r:RESOURCE]->(lambda)
+    WITH lambda
     MATCH (role:AWSPrincipal{arn: {Role}})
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/lambda_function.py
```diff
@@ -36,6 +36,9 @@ def load_lambda_functions(neo4j_session, data, region, current_aws_account_id, a
     lambda.memory = {MemorySize},
     lambda.lastupdated = {aws_update_tag}
     WITH lambda
+    MATCH (owner:AWSAccount{id: {AWS_ACCOUNT_ID}})
+    MERGE (owner)-[r:RESOURCE]->(lambda)
```
achantavy, line 40: Need to add

```
```

Also need to update the Lambda cleanup job to delete stale `:RESOURCE` relationships between Lambdas and their AWS accounts.

PR #380, opened 2020-07-30
Fix #379: run GitHub cleanup only on successful sync

### LINKED ISSUE #379: All GitHub assets are cleaned up even if the sync fails
*Title*: *All GitHub assets are cleaned up even if the sync fails*

*Description*:

We observed a silent failure somewhere in the GitHub repos sync. I can't find it, but my guess is that when it happens, an exception is raised, which causes no GitHubRepository objects to be loaded, and then [the cleanup job triggers anyway](https://github.com/lyft/cartography/blob/f57a9bdddb89244d392e09591fcc475b89419a46/cartography/intel/github/__init__.py#L50) and deletes all previously stored GitHub data.

Instead, we would want cleanup to occur only if a sync has been successful.

*To Reproduce*:

Have GitHub sync fail after user sync but before repo sync finishes.

f57a9bdddb89244d392e09591fcc475b89419a46

achantavy:
- Splits up the original GitHub cleanup job into two parts: one for users and one for repos.
- Only runs a cleanup job when a sync completes.
- Ensure object sent to logger is a string -  hopefully this fixes the silent failures

### DIFF
--- a/cartography/data/jobs/cleanup/github_repos_cleanup.json
+++ b/cartography/data/jobs/cleanup/github_repos_cleanup.json
@@ -9,7 +9,6 @@
     "iterative": true,
     "iterationsize": 100
   },
-
   {
     "query": "MATCH (n:ProgrammingLanguage) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
     "iterative": true,
--- a/cartography/data/jobs/cleanup/github_repos_cleanup.json
+++ b/cartography/data/jobs/cleanup/github_repos_cleanup.json
@@ -9,33 +9,14 @@
     "iterative": true,
     "iterationsize": 100
   },
-  {
-    "query": "MATCH (n:GitHubUser) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
-    "iterative": true,
-    "iterationsize": 100
-  },
-  {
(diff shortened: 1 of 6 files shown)


cartography/intel/github/__init__.py
```diff
@@ -47,8 +46,3 @@ def start_github_ingestion(neo4j_session, config):
             )
         except exceptions.RequestException as e:
             logger.error("Could not complete request to the GitHub API: {}", e)
-    run_cleanup_job(
```
achantavy, line 50: This is the main change: moving this cleanup function out of the `__init__` and splitting it up into `github.users` and `github.repos`.

cartography/data/jobs/cleanup/github_repos_cleanup.json
```diff
@@ -9,36 +9,16 @@
...
-    "query": "MATCH (n:GitHubOrganization) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
-    "iterative": true,
-    "iterationsize": 100
-  },
   {
     "query": "MATCH (n:ProgrammingLanguage) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
     "iterative": true,
     "iterationsize": 100
   },
-  {
```
achantavy, line 27: This is a typo that's been around for a long time.

cartography/intel/github/repos.py
```diff
@@ -179,10 +180,10 @@ def _transform_repo_languages(repo_url, repo, repo_languages):
 
 
 @timeit
-def load_github_repos(session, update_tag, repo_data):
+def load_github_repos(neo4j_session, update_tag, repo_data):
```
achantavy, line 183: While I'm here I figured I'd standardize on clarifying what type of session object this is, like we do in the other syncs.

PR #382, opened 2020-07-31
Fix #377: correctly parse GKE errors

### LINKED ISSUE #377: GCP GKE HttpError 403; requires container.clusters.list
*Title*: *GCP GKE HttpError 403; requires container.clusters.list*

*Description*:

Even if GKE is enabled for the project from a `serviceusage` perspective, our GCP identity may not necessarily have permissions to list GKE assets. This will cause the sync to crash.

In the short term, I think the expected failure behavior for our syncs is to skip and continue with the rest of the sync.

*To Reproduce*:

Have a project where you have GKE assets and try to run the sync using an identity that lacks the container.clusters.list permission.

*Logs*:

```
  File "{path}/site-packages/cartography/sync.py", line 71, in run
    stage_func(neo4j_session, config)
  File "{path}/site-packages/cartography/util.py", line 58, in timed
  File "{path}/site-packages/cartography/intel/gcp/__init__.py", line 218, in start_gcp_ingestion
    _sync_multiple_projects(neo4j_session, resources, projects, config.update_tag, common_job_parameters)
  File "{path}/site-packages/cartography/intel/gcp/__init__.py", line 177, in _sync_multiple_projects
    _sync_single_project(neo4j_session, resources, project_id, gcp_update_tag, common_job_parameters)
  File "{path}/site-packages/cartography/intel/gcp/__init__.py", line 155, in _sync_single_project
    gke.sync_gke_clusters(neo4j_session, resources.container, project_id, gcp_update_tag, common_job_parameters)
  File "{path}/site-packages/cartography/util.py", line 58, in timed
  File "{path}/site-packages/cartography/intel/gcp/gke.py", line 197, in sync_gke_clusters
    gke_res = get_gke_clusters(container, project_id)
  File "{path}/site-packages/cartography/util.py", line 58, in timed
  File "{path}/site-packages/cartography/intel/gcp/gke.py", line 28, in get_gke_clusters
    res = req.execute()
  File "{path}/site-packages/googleapiclient/_helpers.py", line 134, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "{path}/site-packages/googleapiclient/http.py", line 907, in execute
    raise HttpError(resp, content, uri=self.uri)
googleapiclient.errors.HttpError: <HttpError 403 when requesting https://container.googleapis.com/v1/projects/{ProjectName}}/zones/-/clusters?alt=json returned "Required "container.clusters.list" permission(s) for "projects/{ProjectName}".">
```

f57a9bdddb89244d392e09591fcc475b89419a46

*Additional context*:
Somewhat related to #371

It is a little surprising to me that this error was not caught [here](https://github.com/lyft/cartography/blob/f57a9bdddb89244d392e09591fcc475b89419a46/cartography/intel/gcp/gke.py#L42); I'll do some testing later on to verify the returned value of `reason`.

## Proposed fix:

My idea is to raise a custom exception if there is a permissions failure in gcp.compute.get_zones_in_project() and handle it [here](https://github.com/lyft/cartography/blob/f57a9bdddb89244d392e09591fcc475b89419a46/cartography/intel/gcp/__init__.py#L151) by skipping over the subsequent storage and GKE syncs.

cc: @marco-lancini for viz.

achantavy:
Implements error parsing specific to the GKE sync; for some reason the GCP APIs have different error objects returned depending on the resource type.

Tested this end-to-end and it survived my sync.

Proof that this logs a warning but does not crash:
image

### DIFF
--- a/cartography/intel/gcp/gke.py
+++ b/cartography/intel/gcp/gke.py
@@ -28,13 +28,12 @@ def get_gke_clusters(container, project_id):
         res = req.execute()
         return res
     except HttpError as e:
-        err_as_dict = json.loads(e.content.decode('utf-8'))
-        status = err_as_dict['error']['status']
-        if status == 'PERMISSION_DENIED':
+        err = json.loads(e.content.decode('utf-8'))['error']
+        if err['status'] == 'PERMISSION_DENIED':
             logger.warning(
                 (
                     "Could not retrieve GKE clusters on project %s due to permissions issue. Code: %s, Message: %s"
-                ), project_id, err_as_dict['error']['code'], err_as_dict['error']['message'],
+                ), project_id, err['code'], err['message'],
             )
             return {}
         else:
--- a/cartography/intel/gcp/gke.py
+++ b/cartography/intel/gcp/gke.py
@@ -1,8 +1,8 @@
+import json
 import logging
 
 from googleapiclient.discovery import HttpError
 
-from cartography.intel.gcp import compute
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
 
@@ -12,7 +12,7 @@ logger = logging.getLogger(__name__)
 @timeit
 def get_gke_clusters(container, project_id):
     """
-    Returns a list of GKE clusters within some given project.
+    Returns a GCP response object containing a list of GKE clusters within the given project.
 
     :type container: The GCP Container resource object
     :param container: The Container resource object created by googleapiclient.discovery.build()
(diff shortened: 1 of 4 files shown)


cartography/intel/gcp/gke.py
```diff
@@ -28,38 +28,29 @@ def get_gke_clusters(container, project_id):
         res = req.execute()
         return res
     except HttpError as e:
-        reason = compute._get_error_reason(e)
-        if reason == 'invalid':
+        err_as_dict = json.loads(e.content.decode('utf-8'))
```
irhkang, line 31: nit: I would do `err = json.loads(e.content.decode('utf-8'))['error'])`
and remove the `status` variable

PR #394, opened 2020-08-22
Add support for GitHub outside collaborators

achantavy:
Adds the `(:GitHubRepository)-[:OUTSIDE_COLLAB_{permission}]->(:GitHubUser)` relationship. Allows an organization to list all external users that have permissions on its repos.


tests/data/github/repos.py
```diff
@@ -60,69 +61,70 @@
...
-            'url': 'https://github.com/example_org/SampleRepo2',
-            'sshurl': 'git@github.com:example_org/SampleRepo2.git',
-            'updatedat': '2020-07-03T00:25:25Z',
+    {
+        'name': 'cartography',
+        'nameWithOwner': 'lyft/cartography',
+        'primaryLanguage': {'name': 'Python'},
+        'url': 'https://github.com/lyft/cartography',
+        'sshUrl': 'git@github.com:lyft/cartography.git',
+        'createdAt': '2019-02-27T00:16:29Z',
```
achantavy, line 72: Cartography's birthday!

cartography/intel/github/repos.py
```diff
@@ -50,6 +50,16 @@
                         login
                         __typename
                     }
+                    collaborators(affiliation: OUTSIDE, first: 100) {
```
achantavy, line 53: I am making an assumption that a single repo will have 100 or fewer outside contributors. I also don't want to re-write this query to do dual pagination :(

cartography/intel/github/util.py
```diff
@@ -81,6 +81,12 @@ def fetch_all(token, api_url, organization, query, resource_type, field_name):
                 resource_type,
             )
             break
+        except requests.exceptions.HTTPError as e:
```
irhkang, line 84: what are all the call patterns for these?  

Breaking instead of outright failing has caused problems in the past for ACL.  I in general prefer to fail outright but it depends on the call patterns

cartography/intel/github/repos.py
```diff
@@ -77,19 +87,24 @@ def transform(repos_json):
...
     transformed_repo_list = []
     transformed_repo_languages = []
     transformed_repo_owners = []
+    # See https://docs.github.com/en/graphql/reference/enums#repositorypermission
+    transformed_collaborators = {'ADMIN': [], 'MAINTAIN': [], 'READ': [], 'TRIAGE': [], 'WRITE': []}
     for repo_object in repos_json:
         _transform_repo_languages(repo_object['url'], repo_object, transformed_repo_languages)
         _transform_repo_objects(repo_object, transformed_repo_list)
         _transform_repo_owners(repo_object['owner']['url'], repo_object, transformed_repo_owners)
+        _transform_collaborators(repo_object['collaborators'], repo_object['url'], transformed_collaborators)
```
irhkang, line 102: I get that this is following the conventions above..

This is unnecessary mutation IMO.  You can just have `_transform_*` take in 1 less parameter and return the dict or the list.  Mutation just makes code harder to follow and it's easier to write insidious bugs once you expand on this.

irhkang (approved):

cartography/intel/github/util.py
achantavy: Our sync for GitHub objects is best effort and until we get DAGs working or a more robust retry mechanism, I'd rather not let a GitHub failure cause the rest of the sync to end.

I agree with the whole 'fail-loudly' idea, but for now until we build configurable error handling, I'd prefer to 'warn-loudly' that there was missing data, load what we have, and then the caller can manually do a re-run.

cartography/intel/github/repos.py
achantavy: Good feedback, thanks, I'll fix the out-param behavior in a follow-up.

PR #405, opened 2020-09-18
Upgrade to Neo4j v4

mingfang:


achantavy:
Thanks for the PR! We'll have to change our CI config to get Neo4j v4 on it, but do you know if the integration tests still pass?

We will need to review other Cypher syntax changes as documented [https://neo4j.com/docs/cypher-manual/current/deprecations-additions-removals-compatibility/#cypher-deprecations-additions-removals-4.0](here); I'm pretty sure we use removed syntax in addition to `{}`.

I can't commit to a timeline on merging this PR though because upgrading to v4 may break some users (I know _for sure_ it breaks my own setup). I'd like to talk about this at our next community meeting.

PR #417, opened 2020-10-13
#414 Add AWS EC2 network interface details

### LINKED ISSUE #414: ec2 network-interfaces support
*Title*: Import ec2 network interfaces

*Description*:
Import all EC2 Network Interfaces, so that ELB IPs can be identified from the database

Import the output of ec2 describe-network-interfaces call.

I assume everything is there the only new relation which will come in will be 

(LoadBalancer|LoadBalancerV2)-[NETWORK_INTERFACE]->(NetworkInterface)

krisek:
This pull request fixes #414 
Finally the IP connection mentioned in the thread was not added.

### DIFF
--- a/cartography/intel/aws/ec2/network_interfaces.py
+++ b/cartography/intel/aws/ec2/network_interfaces.py
@@ -1,0 +1,221 @@
+import logging
+
+from .util import get_botocore_config
+from cartography.util import aws_handle_regions
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+import re
+
+logger = logging.getLogger(__name__)
+
+@timeit
+@aws_handle_regions
+def get_network_interface_data(boto3_session, region):
+    client = boto3_session.client('ec2', region_name=region, config=get_botocore_config())
+    paginator = client.get_paginator('describe_network_interfaces')
+    subnets = []
+    for page in paginator.paginate():
+        subnets.extend(page['NetworkInterfaces'])
+    return subnets
+
+
+def load_network_interfaces(neo4j_session, data, region, aws_account_id, aws_update_tag):
+
+
+    elb_associations = []
+    instance_associations = []
+    primary_public_ips = []
+    private_ip_addresses = []
+    for network_interface in data:
+
+        matchObj = re.match(r'^ELB (.*)', network_interface.get('Description',''))
+        if matchObj:
+            elb_associations.append({
+                'netinf_id': network_interface['NetworkInterfaceId'],
+                'elb_name': matchObj[1]
+            })
+
+        try:
+            instance_associations.append({
+                'netinf_id': network_interface['NetworkInterfaceId'],
+                'instance_id': network_interface['Attachment']['InstanceId']
+            })
+        except:
+            pass
+
+        try:
+            primary_public_ips.append({
+                'netinf_id': network_interface['NetworkInterfaceId'],
+                'public_ip': network_interface['Association']['PublicIp']
+            })
+        except:
+            pass
+
+        for private_ip_address in network_interface.get('PrivateIpAddresses', []):
+            if 'Association' in private_ip_address:
+                private_ip_addresses.append({
+                    'network_interface_id': network_interface['NetworkInterfaceId'],
+                    'private_ip_id': '{}:{}'.format(network_interface['NetworkInterfaceId'], private_ip_address['PrivateIpAddress']),
+                    'primary': private_ip_address['Primary'],
+                    'private_ip_address': private_ip_address['PrivateIpAddress'],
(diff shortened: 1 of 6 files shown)


achantavy:
Thanks for putting this together!

I agree with leaving out the generic `:Ip` label for now so we can talk about it (I don't have a particularly strong opinion about finding the perfect abstraction for an IP). I like the change, but would like more eyes on this.

I'm curious, can you send a sample screenshot showing off what your graph looks like with this new set of paths?

@marco-lancini what do you think?

cartography/intel/aws/ec2/network_interfaces.py
```diff
@@ -0,0 +1,182 @@
...
+def get_network_interface_data(boto3_session, region):
+    client = boto3_session.client('ec2', region_name=region, config=get_botocore_config())
+    paginator = client.get_paginator('describe_network_interfaces')
+    subnets = []
+    for page in paginator.paginate():
+        subnets.extend(page['NetworkInterfaces'])
+    return subnets
+
+
+def load_network_interfaces(neo4j_session, data, region, aws_account_id, aws_update_tag):
```
achantavy, line 23: Add the `@timeit` decorator

cartography/intel/aws/ec2/network_interfaces.py
```diff
@@ -0,0 +1,182 @@
...
+
+    elb_associations = []
+    instance_associations = []
+
+    for n_idx, network_interface in enumerate(data):
```
achantavy, line 28: Nit: if `n_idx` is never used, do we need `enumerate()`?

cartography/intel/aws/ec2/network_interfaces.py
```diff
@@ -0,0 +1,182 @@
...
+
+        matchObj = re.match(r'^ELB (.*)', network_interface.get('Description', ''))
```
achantavy, line 30: Can you explain the regex match?

achantavy (commented):
First quick pass

krisek:
> I'm curious, can you send a sample screenshot showing off what your graph looks like with this new set of paths?

I can't remember what was my trigger to implement this exactly 😄 I think someone showed me an IP address and based on this I couldn't find the AWS object. So this was rather a reporting excersise.

Though yesterday I found (#418) a relation missing [:PART_OF_SUBNET] which I will add to this patch, and that was used to create a nice graph on our k8s environments. I'll check if it has any sensitive information / I can share it or not.

cartography/intel/aws/ec2/network_interfaces.py
krisek: yes, finally this is not required, there was a version where I modified data in the loop, I'll remove

cartography/intel/aws/ec2/network_interfaces.py
krisek: I was very suprised seeing this, but this is how AWS notates if and IP belongs to an ELB. There is a similar pattern for EFS as well... for a future enchancement.

More info: https://aws.amazon.com/premiumsupport/knowledge-center/elb-find-load-balancer-IP/ 

I realize now that this works only for classic loadbalancers, I think I will need update this pattern for gen2 ELBs a bit.

krisek:
I implemented PART_OF_SUBNET relations consistently -- for instances, load balancers and for their all interfaces.

IMHO 

```
(LoadBalancer)-[SUBNET]->(EC2Subnet) 
```

and

```
(LoadBalancerV2)-[SUBNET]->(EC2Subnet) 
```

relations can be deprecated.

achantavy:
> > I'm curious, can you send a sample screenshot showing off what your graph looks like with this new set of paths?
> 
> I can't remember what was my trigger to implement this exactly 😄 I think someone showed me an IP address and based on this I couldn't find the AWS object. So this was rather a reporting excersise.
> 
> Though yesterday I found (#418) a relation missing [:PART_OF_SUBNET] which I will add to this patch, and that was used to create a nice graph on our k8s environments. I'll check if it has any sensitive information / I can share it or not.

Sounds good, it's fine to just draw boxes over sensitive stuff but if that's too hard no need. Just interested in seeing your paths live in action. Let me know when you're ready for another pass.

krisek:
Here is a little data extract, the use case was the following. We migrate some environments, need to figure out what connectivity needs to be implemented for the new environment. All I had was the ip address of the load balancer (don't ask my why only the IP of the loadbalancer was noted, this is an imaginery scenario anyway :smile: ). With this patch I could prepare this graph, which shows me in one shot the related instances, vpc, cidr blocks, etc.

Screenshot from 2020-10-16 17-47-39

krisek:
Is there anything missing to merge this?

achantavy (approved):
Thanks for the contribution!

Issue #416, opened 2020-10-12
AccountAccessKey no longer existing still being shown in Cartography


marco-lancini:
**Description**:
I realised that `AccountAccessKey` might not be cleaned up properly at the end of each ingestion,
as I've found multiple occurrences where Cartography stores a number of old `AccountAccessKey` not longer existing.

**To Reproduce**:
One of the queries defined in the [cartography-queries](https://github.com/marco-lancini/cartography-queries/blob/master/queries/queries.json#L102) repo asks:
 
> What AccountAccessKey have been created (to authenticate to AWSPrincipals)?

Here is a simplified version of that query:
```
MATCH (a:AWSAccount)-[:RESOURCE]->(p:AWSPrincipal)-[:AWS_ACCESS_KEY]->(k:AccountAccessKey) 
RETURN a.name, p.name, p.arn, k.accesskeyid 
ORDER BY a.name, p.name
```
At the moment, I can see the **same set** of `AccountAccessKey` replicated throughout each `AWSPrincipal` for a subset of `AWSAccount`s in my environment.

achantavy:
It's possible that this path isn't matching in your environment:

https://github.com/lyft/cartography/blob/4fdc8518f546d8329669f23d1fe928a2158c52d9/cartography/data/jobs/cleanup/aws_import_account_access_key_cleanup.json#L3

A fix would make this match condition less restrictive.

marco-lancini:
I wonder if in the query you pasted above we should replace `AWSUser` with `AWSPrincipal`.
Or maybe even have both conditions.
What do you think?

achantavy:
I think we should replace with with AWSPrincipal.

achantavy:
This will be solved with #1710

PR #421, opened 2020-10-15
Fix 416: Replace AWSUser with AWSPrincipal

marco-lancini:
As discussed in #416, this diff replaces `AWSUser` with `AWSPrincipal` in the cleanup job for AWS Access Keys


achantavy (changes requested):
On a second glance I'm not sure this fixes the issue.

The label that is getting made is `AWSUser` not `AWSPrincipal`, so this change wouldn't work.

https://github.com/lyft/cartography/blob/4fdc8518f546d8329669f23d1fe928a2158c52d9/cartography/intel/aws/iam.py#L410-L418

It might be a problem somewhere here: https://github.com/lyft/cartography/blob/4fdc8518f546d8329669f23d1fe928a2158c52d9/cartography/intel/aws/iam.py#L637

Might e unrelated but I also see another bug: the `AWSAccount` here should be `:AWSAccount: https://github.com/lyft/cartography/blob/4fdc8518f546d8329669f23d1fe928a2158c52d9/cartography/intel/aws/iam.py#L639

Currently written it will do a full nodescan without labels and it'lll be slow.

PR #435, opened 2020-11-03
Remove stale rels in network interface cleanup job

achantavy:
Previously we only `detach delete`'d stale nodes. This PR explicitly `deletes` stale relationships as well.


marco-lancini (approved):
Sorry for the delay!

PR #443, opened 2020-11-06
Sync GCP Forwarding Rules

lgomezma:
Adds support to sync GCP Forwarding Rules and relates them to the appropriate Network or Subnetwork


achantavy (changes requested):
Hi @lgomezma, thanks for working on this!

I gave it a brief skim and it looks pretty good. Can you please also update https://github.com/lyft/cartography/blob/master/docs/schema/gcp.md with your new data types?

lgomezma:
Hi @achantavy Thanks for taking a look at it. I've just updated the doc schema for gcp as requested :)

cartography/intel/gcp/compute.py
```diff
@@ -553,6 +625,113 @@ def load_gcp_subnets(neo4j_session, subnets, gcp_update_tag):
...
+        ON CREATE SET p.firstseen = timestamp()
+        SET p.lastupdated = {gcp_update_tag}
+    """
+
+    neo4j_session.run(
+        query,
+        PartialUri=fwd['partial_uri'],
+        SubNetworkPartialUri=fwd.get('subnetwork_partial_uri', None),
+        gcp_update_tag=gcp_update_tag,
+    )\
```
achantavy, line 708: Nit: does the training '\' cause problems?

docs/schema/gcp.md
```diff
@@ -487,6 +489,43 @@ Representation of a GCP [Firewall](https://cloud.google.com/compute/docs/referen
     ```
 
 
+## GCPForwardingRule
+
+Representation of GCP [Forwarding Rules](https://cloud.google.com/compute/docs/reference/rest/v1/forwardingRules/list) and [Global Forwarding Rules](https://cloud.google.com/compute/docs/reference/rest/v1/globalForwardingRules/list).
```
achantavy, line 494: [question] I don't know GCP well enough, but would there be a benefit in separating global forwarding rules from regional forwarding rules? If so, should we represent them as separate node types, or should we have a field on the node to indicate that they are different?

achantavy (commented):
Left a couple of questions, but it looks good to me.

I'll let @marco-lancini check it before we merge it in though since he uses GCP heavier than me :).

achantavy (approved):

achantavy:
Will hold off on merging though until @marco-lancini gives it a once over.

tests/integration/cartography/intel/aws/test_ecr.py
```diff
@@ -41,7 +41,7 @@ def test_load_ecr_repositories(neo4j_session):
     nodes = neo4j_session.run(
         """
         MATCH (r:ECRRepository) RETURN r.arn;
-        """
+        """,
```
marco-lancini, line 44: not pertinent to this PR?

cartography/intel/gcp/compute.py
marco-lancini: +1 on this

marco-lancini (dismissed):

marco-lancini (dismissed):

achantavy (approved):

Issue #434, opened 2020-11-03
AWS network interfaces run_cleanup_job() takes 3 days to run


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

The EC2 network interface cleanup job ([Neo4j query](https://github.com/lyft/cartography/blob/6a96408ffdedfd91b10635e9b8b8687320beefef/cartography/data/jobs/cleanup/aws_ingest_network_interfaces_cleanup.json)) gets deadlocked. I've previously observed cartography deadlocks when the Neo4j python driver handles large amounts of data. For this case, I've repro'd this issue 4 times in a row during the ec2 network interface cleanup.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

1. I ran the sync and observed that there were no new metrics being emitted after the EC2 network interfaces intel module.
2. I attached to the python process with pyrasite and got these stack frames:

```
# pyrasite-shell {PID}

>>> import traceback
>>> import sys
>>> for tid, frame in sys._current_frames().items():
...   print(f"Stack for thread {tid}")
...   traceback.print_stack(frame)
...   print('')
...
Stack for thread {THREAD1}
  File "/usr/lib/python3.6/threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "<string>", line 167, in run
  File "/usr/lib/python3.6/code.py", line 233, in interact
    more = self.push(line)
  File "/usr/lib/python3.6/code.py", line 259, in push
    more = self.runsource(source, self.filename)
  File "/usr/lib/python3.6/code.py", line 75, in runsource
    self.runcode(code)
  File "/usr/lib/python3.6/code.py", line 91, in runcode
    exec(code, self.locals)
  File "<console>", line 3, in <module>

Stack for thread {THREAD2}
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "{PATH1}/intelmodules/syncgraph.py", line 220, in <module>
    main(argv)
  File "{PATH1}/intelmodules/syncgraph.py", line 210, in main
    return cartography.sync.run_with_config(sync, config)
[...]

PR #451, opened 2020-11-10
[do not merge] AWS network interface sync disabled

achantavy:
Doing some testing: trying to isolate problems seen in #434.


achantavy:
Removing since #505 is now merged  :)

PR #455, opened 2020-11-11
Incorrect path to ImageNotFound exception fixed

jfoote-lyft:
Quickly fixed this typo on the except line which was trying to access the exception through the boto session. As we already have reference to the client we can instead access more simply.
This matches other cases in cartography as well on how they import Exceptions.


achantavy (approved):
Thanks!

PR #467, opened 2020-11-16
Standardize AWS sync* func signature

achantavy:
If all sync* functions have the same signature, then we will be able to write another script to more easily schedule them.


achantavy:
Removing this in favor of #530.

PR #472, opened 2020-11-18
Speed up AWS ECR repo images sync with UNWIND

achantavy:
This batches up the write operations in one go rather than calling `neo4j_session.run()` potentially thousands of times.

### DIFF
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -72,7 +72,11 @@ def transform_ecr_repository_images(repo_data):
             if 'imageDigest' in img and img['imageDigest']:
                 filtered_imgs.append(img)
             else:
-                logger.warning("Repo %s has an image that has no imageDigest; skipping.", repo_uri)
+                logger.warning(
+                    "Repo %s has an image that has no imageDigest. Its tag is %s. Continuing on.",
+                    repo_uri,
+                    img.get('imageTag'),
+                )
 
         repo_images_list.append({
             'repo_uri': repo_uri,
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -63,44 +63,47 @@ def load_ecr_repositories(neo4j_session, data, region, current_aws_account_id, a
 @timeit
 def load_ecr_repository_images(neo4j_session, repo_data, region, aws_update_tag):
     query = """
-    MERGE (repo_image:ECRRepositoryImage{id: {RepositoryImageUri}})
-    ON CREATE SET repo_image.firstseen = timestamp()
-    SET repo_image.lastupdated = {aws_update_tag}, repo_image.tag = {ImageTag},
-        repo_image.uri = {RepositoryImageUri}
-    WITH repo_image
+    UNWIND {RepoList} as repo_item
+        UNWIND repo_item.repo_images as repo_img
+            MERGE (ri:ECRRepositoryImage{id: repo_item.repo_uri + ":" + repo_img.imageTag})
+            ON CREATE SET ri.firstseen = timestamp()
+            SET ri.lastupdated = {aws_update_tag},
+            ri.tag = repo_img.imageTag,
+            ri.uri = repo_item.repo_uri + ":" + repo_img.imageTag
+            WITH ri, repo_img, repo_item
 
-    MERGE (image:ECRImage{id: {ImageDigest}})
-    ON CREATE SET image.firstseen = timestamp(), image.digest = {ImageDigest}
-    SET image.lastupdated = {aws_update_tag},
-    image.region = {Region}
-    WITH repo_image, image
-    MERGE (repo_image)-[r1:IMAGE]->(image)
-    ON CREATE SET r1.firstseen = timestamp()
-    SET r1.lastupdated = {aws_update_tag}
-    WITH repo_image
+            MERGE (img:ECRImage{id: repo_img.imageDigest})
+            ON CREATE SET img.firstseen = timestamp(),
+            img.digest = repo_img.imageDigest
+            SET img.lastupdated = {aws_update_tag},
+            img.region = {Region}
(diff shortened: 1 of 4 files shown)


kunaals:
+1

kunaals (dismissed):

cartography/intel/aws/ecr.py
```diff
@@ -63,44 +63,47 @@ def load_ecr_repositories(neo4j_session, data, region, current_aws_account_id, a
...
-    MERGE (repo)-[r2:REPO_IMAGE]->(repo_image)
-    ON CREATE SET r2.firstseen = timestamp()
-    SET r2.lastupdated = {aws_update_tag}
+    UNWIND {RepoList} as repo_item
+        UNWIND repo_item.repo_images as repo_img
+            MERGE (ri:ECRRepositoryImage{id: repo_item.repo_uri + ":" + repo_img.imageTag})
+            ON CREATE SET ri.firstseen = timestamp()
+            SET ri.lastupdated = {aws_update_tag},
+            ri.tag = repo_img.imageTag,
+            ri.uri = repo_item.repo_uri + ":" + repo_img.imageTag
```
gkleiman, line 72: Nit: indentation is off by one level

cartography/intel/aws/ecr.py
```diff
@@ -63,44 +63,47 @@ def load_ecr_repositories(neo4j_session, data, region, current_aws_account_id, a
...
+            WITH ri, repo_img, repo_item
+
+            MERGE (img:ECRImage{id: repo_img.imageDigest})
+            ON CREATE SET img.firstseen = timestamp(),
+            img.digest = repo_img.imageDigest
```
gkleiman, line 77: Ditto indentation

cartography/intel/aws/ecr.py
```diff
@@ -63,44 +63,47 @@ def load_ecr_repositories(neo4j_session, data, region, current_aws_account_id, a
...
+            SET img.lastupdated = {aws_update_tag},
+            img.region = {Region}
```
gkleiman, line 79: Ditto indentation

Is an image in a single region or can it be in multiple regions? If it's the latter then we should make this a list instead of overwriting the value.

cartography/intel/aws/ecr.py
```diff
@@ -63,44 +63,47 @@ def load_ecr_repositories(neo4j_session, data, region, current_aws_account_id, a
...
+            WITH ri, img, repo_img, repo_item
```
gkleiman, line 80: Looks like you don't need `repo_img` after this line.

gkleiman (changes requested):

achantavy:
This fails if the IDs happen to be null: Neo4j fails with a merge on null ID failure or something. I'll update the tests and try again later on.

cartography/intel/aws/ecr.py
achantavy: Single region, since an image lives in a repository that is in a single region.

Come to think of it, we might not actually need a region field for the image since the repository has it defined already but ./shrug.

tests/integration/cartography/intel/aws/test_ecr.py
```diff
@@ -17,18 +17,6 @@ def _ensure_local_neo4j_has_test_ecr_repo_data(neo4j_session):
     )
 
 
-def _ensure_local_neo4j_has_test_repo_and_image_data(neo4j_session):
```
achantavy, line 20: This function was no longer used.

kunaals (approved):

cartography/intel/aws/ecr.py
gkleiman: Nit: the indentation here is still off =)

cartography/intel/aws/ecr.py
```diff
@@ -61,46 +61,59 @@ def load_ecr_repositories(neo4j_session, data, region, current_aws_account_id, a
 
 
 @timeit
-def load_ecr_repository_images(neo4j_session, repo_data, region, aws_update_tag):
+def transform_ecr_repository_images(repo_data):
+    """
+    Ensure that we only load ECRImage nodes to the graph if they have a defined imageDigest field.
```
gkleiman, line 66: Do you know why/in which cases images don't have an imageDigest field?

I think it would be useful to look at an example and see what's going on there. This may be happening if the image is a "multiarch" image with a manifest list and we would have to see how to handle it.

gkleiman (approved):

gkleiman:
Approving to unblock things, but we should definitely follow up on images without a digest.

cartography/intel/aws/ecr.py
achantavy: Good point, I'll file a ticket on this. I find it odd that we only encountered this issue recently; the existing cartography.intel.aws.ecr.load_ecr_repository_images() code would have run into this issue as well, see https://github.com/lyft/cartography/blob/6ab67f8895995f59bf2d0638275e69110d4598d4/cartography/intel/aws/ecr.py#L96

cartography/intel/aws/ecr.py
```diff
@@ -61,46 +61,69 @@ def load_ecr_repositories(neo4j_session, data, region, current_aws_account_id, a
...
+    for repo_uri, repo_images in repo_data.items():
+        filtered_imgs = []
+        for img in repo_images:
+            if 'imageDigest' in img and img['imageDigest']:
+                filtered_imgs.append(img)
+            else:
+                logger.warning(
+                    "Repo %s has an image that has no imageDigest. Its tag is %s. Continuing on.",
+                    repo_uri,
+                    img.get('imageTag'),
```
achantavy, line 78: I am allowing imageTag to be None. I really don't think this will ever happen, but then again I also didn't think that we'd ever have a case where imageDigest returned None :-)

gkleiman (approved):

PR #477, opened 2020-11-23
[WIP] Limit scope of GCP cleanups to current gcp project

lgomezma:
Related to [this issue](https://github.com/lyft/cartography/issues/478).


lgomezma:
@marco-lancini @achantavy I think to make this fully workable, we will need to add `project_id` on Nodes that don't contain them at the moment like `GCPNetworkInterface` or `GCPFirewall`. Any thoughts?

achantavy:
Most GCP resources have a connection to a GCPProject: https://github.com/lyft/cartography/blob/master/docs/schema/gcp.md#gcpproject (although that entry in the schema doc has somehow gotten a bit stale), so we can do what we do with AWS cleanup jobs and delete all stale resources of a given type that are attached to the current Project. Like this: https://github.com/lyft/cartography/blob/0f71b3f0246665d5fa065afa2e3dc46c22d6c689/cartography/data/jobs/cleanup/aws_import_ecr_cleanup.json#L4

We would also copy the AWS cleanup jobs by doing something like this: https://github.com/lyft/cartography/blob/0f71b3f0246665d5fa065afa2e3dc46c22d6c689/cartography/intel/aws/__init__.py#L93
where we would add in the current GCPProject ID to the `common_job_parameters` variable that gets passed to every cleanup job.

I can help on the details of this if needed, I'm kinda rushing explaining this right now - let me know if there's something that doesn't make sense.

lgomezma:
thanks @achantavy indeed that's the approach I'm taking. If you see on my PR I do exactly that, add the GPC project_id to that `common_job_parameters` variable that gets passed to all the cleanup jobs. I tried to see if I could directly add a project_id property on all GCP nodes but it proves a bit of a challenge to do that while maintaining compatibility with the current state of affairs. I will explore using the relation to the GCP project like you suggested for those GCP Resources that don't have that project_id in their property list.

PR #487, opened 2020-12-07
ec2 vpc peerings implemented as nodes instead of relations

krisek:
Hi, I think I did something wrong with the previous pull request, so here we go with a clean one. Apologies for the lost context now, but this way we can start (again) with a clean sheet.

### DIFF
--- a/cartography/intel/aws/ec2/vpc_peerings.py
+++ b/cartography/intel/aws/ec2/vpc_peerings.py
@@ -51,6 +51,21 @@ def load_vpc_peerings(neo4j_session, data, region, aws_account_id, aws_update_ta
     ON CREATE SET raccount.firstseen = timestamp(), raccount.foreign = true
     SET raccount.lastupdated = {aws_update_tag}
 
+    MERGE (pcx)-[rav:ACCEPTER_VPC]->(avpc)
+    ON CREATE SET rav.firstseen = timestamp()
+    SET rav.lastupdated = {aws_update_tag}
+
+    MERGE (aaccount)-[ra:RESOURCE]->(avpc)
+    ON CREATE SET ra.firstseen = timestamp()
+    SET ra.lastupdated = {aws_update_tag}
+
+    MERGE (pcx)-[rrv:REQUESTER_VPC]->(rvpc)
+    ON CREATE SET rrv.firstseen = timestamp()
+    SET rrv.lastupdated = {aws_update_tag}
+
+    MERGE (raccount)-[rr:RESOURCE]->(rvpc)
+    ON CREATE SET rr.firstseen = timestamp()
+    SET rr.lastupdated = {aws_update_tag}
 
     """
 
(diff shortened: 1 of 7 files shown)


achantavy (commented):
Thanks for continuing the work on this, do you think it's time to deprecate https://github.com/lyft/cartography/blob/master/cartography/intel/aws/ec2/vpc_peering.py?

krisek:
intel/aws/ec2/vpc_peering.py is definitely obsoleted by this PR, but if I remove it, Travis builds do not pass, hence I left the file in the repo

krisek:
I discovered one foreign account property setting was missing.

cartography/intel/aws/ec2/vpc_peerings.py
```diff
@@ -0,0 +1,173 @@
...
+    MERGE (avpc:AWSVpc{id: vpc_peering.AccepterVpcInfo.VpcId})
+    ON CREATE SET avpc.firstseen = timestamp()
+    SET avpc.lastupdated = {aws_update_tag}, avpc.vpcid = vpc_peering.AccepterVpcInfo.VpcId
+
+    MERGE (rvpc:AWSVpc{id: vpc_peering.RequesterVpcInfo.VpcId})
+    ON CREATE SET rvpc.firstseen = timestamp()
+    SET rvpc.lastupdated = {aws_update_tag}, rvpc.vpcid = vpc_peering.RequesterVpcInfo.VpcId
+
+    MERGE (aaccount:AWSAccount{id: vpc_peering.AccepterVpcInfo.OwnerId})
+    ON CREATE SET aaccount.firstseen = timestamp(), aaccount.foreign = true
```
achantavy, line 47: I don't think this treatment of `foreign` is the same as the older vpc_peering.py. In that file, I believe `foreign` was used to denote AWS accounts that were not included in the set of accounts that we "initially knew about", if that makes sense. We can talk about this in chat too.

If you get stuck on what to do here, I think it's fine to not set `foreign` for now.

cartography/intel/aws/ec2/vpc_peerings.py
```diff
@@ -0,0 +1,173 @@
...
+    return client.describe_vpc_peering_connections()['VpcPeeringConnections']
+
+
+@timeit
+def load_vpc_peerings(neo4j_session, data, region, aws_account_id, aws_update_tag):
+
+    ingest_vpc_peerings = """
+    UNWIND {vpc_peerings} AS vpc_peering
+
+    MERGE (pcx:PeeringConnection:AWSPeeringConnection{id: vpc_peering.VpcPeeringConnectionId})
```
achantavy, line 24: Is it possible to attach `pcx` to `avpc` and then `avpc` to `aaccount` etc etc in this query? I think we should try to make sure AWS resource nodes are connected to their accounts as early as possible.

cartography/intel/aws/ec2/vpc_peerings.py
```diff
@@ -0,0 +1,173 @@
...
+
+@timeit
+@aws_handle_regions
+def get_vpc_peerings(boto3_session, region):
+    client = boto3_session.client('ec2', region_name=region, config=get_botocore_config())
+    return client.describe_vpc_peering_connections()['VpcPeeringConnections']
+
+
+@timeit
+def load_vpc_peerings(neo4j_session, data, region, aws_account_id, aws_update_tag):
```
achantavy, line 19: Can you split this into one function per query to make it easier to test? Like one `load()` function that calls `load_{sub_object}()`?

achantavy (changes requested):
I tested this manually end-to-end and it runs pretty fast :).

I left a comment on splitting the `load()` function to a few sub-parts, let me your thoughts.

cartography/intel/aws/ec2/vpc_peerings.py
krisek: I hope I managed to fix this with the last commit #6dc5950

cartography/intel/aws/ec2/vpc_peerings.py
krisek: I hope I managed to fix this with the last commit #6dc5950 I was not fully sure what you meant by split

krisek:
This code got completely outdated.

Issue #488, opened 2020-12-08
AWSDNSRecords not linked with LoadBalancersV2 


serain:
*Title*: AWSDNSRecords not linked with LoadBalancersV2

*Description*:
AWSDNSRecords are linked with V1 LoadBalancers, but they are not linked with LoadBalancersV2.

i.e. `(AWSDNSRecord)-[DNS_POINTS_TO]->(LoadBalancer)` returns expected relationships but `(AWSDNSRecord)-[DNS_POINTS_TO]->(LoadBalancerV2)` returns nothing.

serain:
PR for this: https://github.com/lyft/cartography/pull/489

PR #489, opened 2020-12-08
link AWSDNSRecord and LoadBalancerV2

serain:
This resolves issue #488 

Links AWSDNSRecord to LoadBalancerV2.

### DIFF
--- a/cartography/intel/aws/route53.py
+++ b/cartography/intel/aws/route53.py
@@ -27,6 +27,15 @@ def link_aws_resources(neo4j_session, update_tag):
     """
     neo4j_session.run(link_elb, aws_update_tag=update_tag)
 
+    # find records that point to AWS LoadBalancersV2
+    link_elbv2 = """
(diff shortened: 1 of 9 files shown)


achantavy (changes requested):
Thank you for working on this!

Can you please also update the tests?

For example, this code tests if ELBv2 nodes are attached to EC2 instances:
https://github.com/lyft/cartography/blob/master/tests/integration/cartography/intel/aws/ec2/test_ec2_load_balancers.py

and the source sample test data is here:
https://github.com/lyft/cartography/blob/master/tests/data/aws/ec2/load_balancers.py

Let us know if you need help on this.

serain:
I gave this a crack, but I'm not sure it's what you had in mind.

I added a test to route53 (felt appropriate as it's the route53 module that creates the links). I added a DNS record to the test data and then load both the EC2 and Route53 test data and check there's the expected link.

Let me know what you think.

achantavy (approved):
This looks great, thanks @serain!

PR #496, opened 2020-12-10
Fix #492: Handle accessdenied on get_bucket_location

### LINKED ISSUE #492: AWS S3 get_bucket_location fails with Access Denied
*Description*:

S3 get_bucket_location() can fail with a permissions error. Given that as much of cartography should run with only SecurityAudit, we should handle this exception gracefully.

*To Reproduce*:

Run the s3 sync on an account that does not have get_bucket_location permissions.

*Logs*:

```
  File "/{PATH/cartography/intel/aws/__init__.py", line 98, in _sync_multiple_accounts
    _sync_one_account(neo4j_session, boto3_session, account_id, sync_tag, common_job_parameters)
  File "/{PATH/cartography/intel/aws/__init__.py", line 29, in _sync_one_account
  File "/{PATH/cartography/util.py", line 58, in timed
  File "/{PATH/cartography/intel/aws/s3.py", line 361, in sync
    bucket_data = get_s3_bucket_list(boto3_session)
  File "/{PATH/cartography/util.py", line 58, in timed
  File "/{PATH/cartography/intel/aws/s3.py", line 21, in get_s3_bucket_list
    bucket['Region'] = client.get_bucket_location(Bucket=bucket['Name'])['LocationConstraint']
  File "/{PATH/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/{PATH/botocore/client.py", line 676, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the GetBucketLocation operation: Access Denied
```

0f71b3f0246665d5fa065afa2e3dc46c22d6c689

*Additional context*:

This is an unintended side effect of https://github.com/lyft/cartography/pull/470.

achantavy:


gkleiman:
+1

gkleiman (approved):
LGTM

A test would be nice, but it doesn't look like there are other unit tests that mock clients/services.

achantavy:
One day! :)

Issue #503, opened 2020-12-14
fix syncing multiple AWS accounts in parallel


krisek:
Bug

*Title*: *Fix syncing multiple AWS accounts in parallel*

*Description*:
At the moment if you are syncing more AWS accounts, you have to sync all of them in a shot as the cleanup job cleans all non-updated accounts. This is not inteded behaviour, it has to be fixed.

achantavy:
> the cleanup job cleans all non-updated accounts

Can you please point out where this happens? This is not intended behavior and is absolutely a bug.

krisek:
I'll submit a pull request with the issues I found.

PR #504, opened 2020-12-19
#503 aws-keep-unseen-accounts

krisek:
This PR implements the `--aws-keep-unseen-accounts` cli parameter and fixes the issues encountered during syncing several accounts independently to the same Neo4j database. (Bug report/feature request #503)

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -114,6 +114,14 @@ class CLI:
             ),
         )
         parser.add_argument(
+            '--aws-keep-unseen-accounts',
+            action='store_true',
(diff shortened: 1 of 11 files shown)


achantavy (changes requested):
Thanks for going through the trouble of implementing this, but I think that the correct behavior is to delete nodes only attached to the current account in all cases.

So I would say
1. Remove the CLI switch
2. Modify the existing cleanup jobs to connect to the current AWS accounts.

krisek:
As discussed closing this.

PR #507, opened 2020-12-28
Add internet_exposed attribute to GCP Compute Instances with 'ALL' firewall rules

jg10:
This adds the internet_exposed attribute to GCP Compute Instances that have a firewall rule associated with them that allows ALL protocols, as opposed to explicitly TCP or UDP. It is a copy of the TCP and UDP checks but with 'ALL' for the protocol. I have tested it and it correctly adds the internet_exposed attribute on hosts that are accessible over the Internet which didn't have the attribute before.


achantavy (approved):
Looks good, thanks for the contribution!

PR #509, opened 2021-01-12
Fix error return codes when connection to Neo4J DB fails

jg10:
Added return code of 1 to the two errors that can occur when initially connecting to the neo4j DB. Not returning an error code has caused issues when running as a K8s cronjob as the job is seen as successfully completing with a zero return code when it has failed to connect to the DB. This fixes that and returns a 1 return code if cartography fails to connect to the DB.


achantavy (approved):
Thanks for the PR!

PR #514, opened 2021-01-19
Test aws.start_aws_ingestion and sync_multiple_accounts

achantavy:
Adding tests for AWS module glue-code.

## Motivation

Many of our intelmodules have decent test coverage where we verify that nodes are created with their expected nodes and relationships. However, the only way to test the code paths leading up to the intel modules is to manually run Cartography  end-to-end. The goal here is to start adding automated sanity checks against code changes that affect our critical path.

As always, open to feedback and discussion; I did my best to avoid writing tests that are overly brittle and I'd like to hear your thoughts.

### DIFF
--- a/tests/integration/cartography/intel/aws/test_init.py
+++ b/tests/integration/cartography/intel/aws/test_init.py
@@ -1,0 +1,52 @@
+from unittest import mock
+
+import cartography.config
+import cartography.intel.aws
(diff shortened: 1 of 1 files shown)


marco-lancini (approved):

kunaals (approved):
Looks good to me

PR #524, opened 2021-02-01
Migrate from travis-ci.org to GitHub actions

achantavy:


.github/pull_request.yml
```diff
@@ -0,0 +1,34 @@
+name: Pull Request
```
danielhochman, line 1: move to `.github/workflows/pull_request.yml`

.github/pull_request.yml
```diff
@@ -0,0 +1,34 @@
...
+on: [pull_request, push]
+
+jobs:
+  make_test:
```
dschaller, line 5: super nit: You might want to change the naming of this workflow/job since they end up in the format `{workflow} / {job} (pull_request)` on PRs

.github/pull_request.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Install Neo4j
+        run: sudo apt-get install neo4j
+
+      - name: Disable auth for neo4j
+        run: |
+          sudo sed -i 's/#dbms.security.auth_enabled=false/dbms.security.auth_enabled=false/g' /etc/neo4j/neo4j.conf
+          sudo service restart neo4j
+
+      - name: Install cartography
```
dschaller, line 28: You can wrap this in a caching action for improved installation speeds https://github.com/actions/cache/blob/main/examples.md#python---pip

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,34 @@
+name: Pull Request
+on: [pull_request, push]
```
dschaller, line 2: You can also specify that this should only run on PRs or pushes to individual branches if you want: https://github.com/lyft/clutch/blob/main/.github/workflows/frontend.yml#L3-L8

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,55 @@
...
```
danielhochman, line 2: you probably want

```suggestion
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master```

Comment (achantavy) [reply .github/workflows/pull_request.yml]: I was wondering about this, thanks!

File: .github/workflows/test_suite.yml [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.6"
+      - uses: pre-commit/action@v2.0.0
```
achantavy, line 18: Not sure if the [way I had it before](https://github.com/lyft/cartography/blob/c153bad7343320fcfb6a0d3233bf0b00d92b1482/.github/workflows/pull_request.yml#L13-L22) is faster (use cache, pip install, run `pre-commit --run --all-files`). I don't have a strong opinion either way. What do you think?

achantavy:
@danielhochman @dschaller Thanks for the help so far, mind giving this another quick skim?

Oh, and to get rid of this: 
image
I think I just need to get rid of the webhook in this repo's config?

skiptomyliu (approved):

achantavy:
> @danielhochman @dschaller Thanks for the help so far, mind giving this another quick skim?
> 
> Oh, and to get rid of this:
> image
> I think I just need to get rid of the webhook in this repo's config?

For future reference, I had to update [branch protection rules](https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule)

PR #526, opened 2021-02-02
Load ecr repository images in batches to resolve timeouts

### LINKED ISSUE #522: neobolt.exceptions.ServiceUnavailable in ecr.load_ecr_repository_images() when connecting to remote Neo4j
*Description*:

When loading 50MB worth of data to a **remote** Neo4j server (i.e. not located on the same machine), ecr.load_ecr_repository_images() crashes with a neobolt.exceptions.ServiceUnavailable error after running for 2 hours.

*To Reproduce*:

Run `ecr.load_ecr_repository_images()` with 50MB of data.

POC code:

```python
from neo4j import GraphDatabase
import cartography.intel.aws.ecr
import time
# You will need to provide your own data here.
# data shape = [{'repo_uri': 'uri', 'repo_images': [{'imageDigest': 'mydigest', 'imageTag': 'mytag'}, ...]},...]
from image_data import image_list

neo4j_driver = GraphDatabase.driver("bolt://your-remote-endpoint:7687")
neo4j_session = neo4j_driver.session()
account_id = '1234'
aws_update_tag = int(time.time())
region = 'us-east-1'

common_job_parameters = {
    "UPDATE_TAG": aws_update_tag,
    "AWS_ID": account_id,
}

cartography.intel.aws.ecr.load_ecr_repository_images(neo4j_session, image_list, region, aws_update_tag)
cartography.intel.aws.ecr.cleanup(neo4j_session, common_job_parameters)
```

*Logs*:

```
  File "/Users/achantavy/.pyenv/versions/3.7.9/lib/python3.7/ssl.py", line 1071, in recv_into
    return self.read(nbytes, buffer)
  File "/Users/achantavy/.pyenv/versions/3.7.9/lib/python3.7/ssl.py", line 929, in read
    return self._sslobj.read(len, buffer)
ConnectionResetError: [Errno 54] Connection reset by peer
Exception ignored in: 'neobolt.bolt._io.ChunkedInputBuffer.receive'
  File "/Users/achantavy/.pyenv/versions/3.7.9/lib/python3.7/ssl.py", line 1071, in recv_into
    return self.read(nbytes, buffer)
  File "/Users/achantavy/.pyenv/versions/3.7.9/lib/python3.7/ssl.py", line 929, in read
    return self._sslobj.read(len, buffer)
ConnectionResetError: [Errno 54] Connection reset by peer
  File "load_ecr_list_images.py", line 17, in <module>
    cartography.intel.aws.ecr.load_ecr_repository_images(neo4j_session, image_list, region, aws_update_tag)
  File "/Users/achantavy/lyftsrc/cartography/cartography/util.py", line 63, in timed
    return method(*args, **kwargs)
  File "/Users/achantavy/lyftsrc/cartography/cartography/intel/aws/ecr.py", line 122, in load_ecr_repository_images
    Region=region,
  File "/Users/achantavy/.virtualenvs/env11/lib/python3.7/site-packages/neo4j/__init__.py", line 503, in run
    self._connection.fetch()
  File "/Users/achantavy/.virtualenvs/env11/lib/python3.7/site-packages/neobolt/direct.py", line 414, in fetch
    return self._fetch()
  File "/Users/achantavy/.virtualenvs/env11/lib/python3.7/site-packages/neobolt/direct.py", line 431, in _fetch
    self._receive()
  File "/Users/achantavy/.virtualenvs/env11/lib/python3.7/site-packages/neobolt/direct.py", line 472, in _receive
    raise self.Error("Failed to read from defunct connection {!r}".format(self.server.address))
neobolt.exceptions.ServiceUnavailable: Failed to read from defunct connection Address(host={IP}, port=7687)
```

0c9a662672cf4925ac8214b44451b1c50374aa97

3.7.9

Have observed this in a Docker container based on Debian as well as my OSX laptop. Neither of them appear to be resource constrained: CPU usage is around 0%, memory usage of the python process is about 200-300MB.

*Additional context*:

This appears related to 
- #440

- #170 

all of these issues involve sending fairly large objects over the Bolt connection.

----

Update:

I've also observed this issue on load_ecr_repositories().

voutilad:
Attempts to fix #522 by batching the data to minimize transaction size. Should resolve network errors due to unresponsive Neo4j server when under too much memory pressure.

I chose the batch size arbitrarily, but given it's creating nodes as well as relationships to highly dense nodes, it shouldn't be set too high without knowing the characteristics of the database host.

In general, this batching approach can be used in other parts of the codebase, so I put it in the common `util` package.

### DIFF
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -3,6 +3,7 @@ from typing import Dict
 from typing import List
 
 from cartography.util import aws_handle_regions
+from cartography.util import into_batches
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
 
@@ -117,12 +118,14 @@ def load_ecr_repository_images(neo4j_session, repo_images_list, region, aws_upda
             SET r2.lastupdated = {aws_update_tag}
     """
     logger.debug("Loading ECR repository images for region '%s' into graph.", region)
-    neo4j_session.run(
-        query,
-        RepoList=repo_images_list,
-        aws_update_tag=aws_update_tag,
-        Region=region,
-    ).consume()  # See issue #440
+
+    for batch in into_batches(repo_images_list, 500):
+        neo4j_session.run(
+            query,
+            RepoList=batch,
+            aws_update_tag=aws_update_tag,
+            Region=region,
+        ).consume()  # See issue #440
 
 
 @timeit
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -95,5 +95,8 @@ def into_batches(items, batch_size):
 
     See also: https://stackoverflow.com/a/22045226
     """
+    if len(items) <= batch_size:
+        return items
+
     items_iter = iter(items)
     return iter(lambda: list(islice(items_iter, batch_size)), ())
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -1,5 +1,6 @@
 import logging
 import sys
+from itertools import islice
 
 import botocore
 
@@ -85,3 +86,14 @@ def aws_handle_regions(func):
             else:
                 raise
     return inner_function
+
+
+def into_batches(items, batch_size):
+    """
+    Take the given items and chunk them into batches of the given batch size.
+    Returns a new iterator of tuples over the provided items.
+
+    See also: https://stackoverflow.com/a/22045226
+    """
+    items_iter = iter(items)
+    return iter(lambda: list(islice(items_iter, batch_size)), ())


achantavy:
@voutilad Thank you so much! This problem has been driving me crazy. Will play with it and report back.

achantavy:
This is odd. When I run the original ECR sync without your patch on regions us-east-1 and us-west2, it would crash on us-west-2.

Now when I run this patch, it starts on us-east-1 and then 9 hours later it crashes with

```
neobolt.exceptions.ServiceUnavailable: Failed to establish connection to ({IP}, 7687) (reason [Errno 111] Connection refused)
```

I'll keep playing and report back.

voutilad:
Does any data end up committed to that database within that timeframe? Might be helpful to check the logs on the server end, too, to see if there's anything about either pause the world gc events or exhausting worker thread pools.

My initial thought is the use of `session.run()` for this long a load might be problematic because, while it will auto-commit (which means you should have some data if the very first batch succeeded), it doesn't provide any retries for transient errors. I don't think this is considered a transient error, though.

I'm guessing that since the same session is being used for quite some time, the connection might have been interrupted for some reason and upon another invocation of `.run()` the Python driver went to go make a new connection because there wasn't an available in its internal pool. That new connection failed.

achantavy:
> I'm guessing that since the same session is being used for quite some time, the connection might have been interrupted for some reason and upon another invocation of .run() the Python driver went to go make a new connection because there wasn't an available in its internal pool. That new connection failed.

Interesting, do you know the best practice here when it comes to sessions? Our code basically reuses the same session for the entire duration of a sync (in some test runs for literally _days_ at times). What is the recommended lifetime of a session?

Appreciate the insight!

voutilad:
Sessions are designed to be ephemeral and act as a means of borrowing a database connection from the driver’s internal connection pool. In clustered deployments, they also help pass some data from query to query to keep them causally consistent (if working with a Neo4j Enterprise cluster).

It’s typical that an application needs multiple concurrent connections to the database, so using 1 driver instance the app will typically create multiple sessions concurrently at any given time. In this case, since the app isn’t performing any concurrency, 1 session is in some ways fine...but with a bit of a caveat.

I have to double check the Python driver behavior, IIRC the connection is made typically when the session object is created. (That is, the TCP connection is made, bolt handshake occurs to negotiate bolt protocol version, and then authentication occurs immediately.) After that, you effectively have a persistent network connection to a Bolt worker thread on the Neo4j server. The driver can reuse it and will track it in a pool.

It’s possible there’s no connection lifecycle handling after the session is created (i.e. if the connection dies for any reason, it’s not reestablished). It’s also possible that the connection to Neo4j sits idle long enough that either the Neo4j worker thread handling it decides to close it OR a network gateway/load balancer in the middle decides it’s worth closing.

The right solution might be:

1. passing a driver object around instead of a session, letting each function that needs to upsert data into the graph create a session to do so, at the expense of potentially having to create a new TCP connection each time if the connections in the pool timeout
2. switching to “transaction work functions” which provide some amount of retry logic in the driver

I’ve got an internal question to the drivers r&d team to confirm or refute my suspicions.

achantavy:
> It’s possible there’s no connection lifecycle handling after the session is created (i.e. if the connection dies for any reason, it’s not reestablished). It’s also possible that the connection to Neo4j sits idle long enough that either the Neo4j worker thread handling it decides to close it OR a network gateway/load balancer in the middle decides it’s worth closing.

I think you're onto something here. I have a repro:

```python
from neo4j import GraphDatabase
import cartography.intel.aws.ecr
import time
from image_data_us_west_2 import image_list  # about 30MB of data
import logging

neo4j_driver = GraphDatabase.driver("bolt://remote-endpoint.example.com:7687")
neo4j_session = neo4j_driver.session()
account_id = '1234'
aws_update_tag = int(time.time())
region = 'us-east-1'


logging.basicConfig(level=logging.DEBUG)

common_job_parameters = {
    "UPDATE_TAG": aws_update_tag,
    "AWS_ID": account_id,
}
size = len(image_list)

print("size = ", size, int(time.time()))
print('Sleeping for 30 minutes before we start loading items to the graph. the neo4j session will remain dormant.', int(time.time()))
time.sleep(60 * 30)

print('loading to graph now!', int(time.time()))
cartography.intel.aws.ecr.load_ecr_repository_images(neo4j_session, image_list, region, aws_update_tag)

print("Running cleanup", int(time.time()))
cartography.intel.aws.ecr.cleanup(neo4j_session, common_job_parameters)
print("Done", int(time.time()))
```

I ran it and got this output:
```
size =  2474 1613945769
sleeping for 30 minutes before we start loading items to the graph. the neo4j session will remain dormant. 1613945769

loading to graph now! 1613947569
  File "load_ecr_list_images.py", line 75, in <module>
    cartography.intel.aws.ecr.load_ecr_repository_images(neo4j_session, image_list, region, aws_update_tag)
  File "/Users/achantavy/lyftsrc/cartography/cartography/util.py", line 63, in timed
    return method(*args, **kwargs)
  File "/Users/achantavy/cartography/cartography/intel/aws/ecr.py", line 124, in load_ecr_repository_images
    Region=region,
  File "/Users/achantavy/.virtualenvs/env11/lib/python3.7/site-packages/neo4j/__init__.py", line 502, in run
    self._connection.send()
  File "/Users/achantavy/.virtualenvs/env11/lib/python3.7/site-packages/neobolt/direct.py", line 394, in send
    self._send()
  File "/Users/achantavy/.virtualenvs/env11/lib/python3.7/site-packages/neobolt/direct.py", line 409, in _send
    self.socket.sendall(data)
  File "/Users/achantavy/.pyenv/versions/3.7.9/lib/python3.7/ssl.py", line 1034, in sendall
    v = self.send(byte_view[count:])
  File "/Users/achantavy/.pyenv/versions/3.7.9/lib/python3.7/ssl.py", line 1003, in send
    return self._sslobj.write(data)
BrokenPipeError: [Errno 32] Broken pipe
```

This is really interesting because the code works without the `time.sleep()`. The idea here is to intentionally leave the session dormant for a long time, such as in the case when the `get` phase of a sync job takes a very long time. This definitely seems to support your theory and it could be worth passing the driver object instead of a single session object for the whole app.

Will keep playing with different options.

voutilad:
The error in #522 is from a read on the socket while this new one is from a write. In #522 it could be a network gateway in between noticing the Neo4j host hasn't been sending any responses due to being overburdened. 

In this new error, it's a write on the socket. It's possible there's still a network gateway terminating the seemingly idle connection as before. I have to double check, but it could also be the Neo4j server timing out the idle Bolt connection as the default for a worker thread (on the server) is `300s`: https://neo4j.com/docs/operations-manual/3.5/reference/configuration-settings/#config_dbms.connector.bolt.thread_pool_keep_alive

PR #527, opened 2021-02-02
Added support for AWS KMS

mpurusottamc:
This is an initial version of pulling KMS key, alias and grant details and populating graph db.


achantavy:
@mpurusottamc Thanks so much for your PR and sorry for the delays. We've all been a bit more tied up than normal but I'll be able to get you a review late this week/early next week if someone else doesn't beat me to it.

docs/schema/aws.md
```diff
@@ -1758,3 +1760,37 @@ Representation of an AWS S3 [Bucket](https://docs.aws.amazon.com/AmazonS3/latest
...
+- AWS KMS Keys are resources in an AWS Account.
+
+        ```
+        (AWSAccount)-[RESOURCE]->(KMSKey)
+        ```
+
+- AWS KMS Key may also be refered as KMSAlias via aliases.
+
+        ```
+        (KMSKey)-[KNOWN_AS]->(KMSAlias)
```
marco-lancini, line 1789: Would it be possible to add an entry in the schema also for `KMSAlias` and `KMSGrant` since they are treated as separate entities?

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,331 @@
...
+
+    load_kms_keys(neo4j_session, kms_keys, region, current_aws_account_id, aws_update_tag)
+
+    policy_alias_grants_data = get_kms_key_details(boto3_session, kms_keys, region)
+    load_kms_key_details(neo4j_session, policy_alias_grants_data, region, current_aws_account_id, aws_update_tag)
+
+
+@timeit
+def sync(neo4j_session, boto3_session, regions, current_aws_account_id, aws_update_tag, common_job_parameters):
+    logger.info("Inside KMS")
```
marco-lancini, line 326: This could become a bit verbose. It should be fine to remove

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,331 @@
...
+from policyuniverse.policy import Policy
+
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_kms_key_list(boto3_session, region):
```
marco-lancini, line 14: Can you add the `@aws_handle_regions` decorator, please?
(also below to all the functions that fetch data)

docs/schema/aws.md
mpurusottamc: @marco-lancini Updated the schema documentation with KMSAlias and KMSGrant in latest commit

cartography/intel/aws/kms.py
mpurusottamc: @marco-lancini removed verbose log statement in latest commit

cartography/intel/aws/kms.py
mpurusottamc: @marco-lancini added `@aws_handle_regions` decorator in latest commit

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+
+    # The owner data returned by the API maps to the aws account nickname and not the IAM user
+    # there doesn't seem to be a way to retreive the mapping but we can get the current context account
+    # so we map to that directly
+
+    for kms_key in data:
+        neo4j_session.run(
+            ingest_keys,
+            KeyId=kms_key["KeyId"],
```
achantavy, line 297: What is the difference between a KeyId and an Arn?

For AWS syncs, we should to do something like 

```
ON CREATE SET kmskey.id = kms_key.arn,
kmskey.firstseen = timestamp()
SET kmskey.arn = kms_key.arn,
[...]
```

It's a bit duplicative, but at least viewers of your data will be able to tell what field(s) the `id` was based off of.

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+    _load_kms_key_policies(neo4j_session, policies, update_tag)
+    _load_kms_key_aliases(neo4j_session, aliases, update_tag)
+    _load_kms_key_grants(neo4j_session, grants, update_tag)
+    _set_default_values(neo4j_session, aws_account_id)
+
+
+@timeit
+def parse_policy(key, policy):
+    """
+    Uses PolicyUniverse to parse S3 policies and returns the internet accessibility results
```
achantavy, line 210: Copypasta; this isn't an s3 policy

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+@timeit
+def _load_kms_key_policies(neo4j_session, policies, update_tag):
+    """
+    Ingest KMS Key policy results into neo4j.
+    """
+    # NOTE we use the coalesce function so appending works when the value is null initially
+    ingest_policies = """
+    UNWIND {policies} AS policy
+    MATCH (k:KMSKey) where k.name = policy.kms_key
+    SET k.anonymous_access = (coalesce(k.anonymous_access, false) OR policy.internet_accessible),
```
achantavy, line 153: Your schema documentation should include what `kmskey.anonymous_access` means.

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+        )
+
+
+@timeit
+def _load_kms_key_grants(neo4j_session, grants_list, update_tag):
+    """
+    Ingest KMS Key Grants into neo4j.
+    """
+    ingest_grants = """
+    MERGE (g:KMSGrant{id: {GrantId}})
```
achantavy, line 122: Same comment about using `unwind`.

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+    MATCH (kmskey:KMSKey{id: {TargetKeyId}})
+    MERGE (g)-[r:APPLIED_ON]->(kmskey)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {UpdateTag}
+    """
+
+    for grant in grants_list:
+        neo4j_session.run(
+            ingest_grants,
+            GrantId=grant["GrantId"],
```
achantavy, line 135: Is this based off an Arn?

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+
+    neo4j_session.run(
+        set_defaults,
+        AWS_ID=aws_account_id,
+    )
+
+
+@timeit
+def load_kms_key_details(neo4j_session, policy_alias_grants_data, region, aws_account_id, update_tag):
```
achantavy, line 178: I realize you used the s3 sync as a model here, and while this works, I wonder if using Resource Permission Relationships would be more powerful. See https://github.com/lyft/cartography/blob/master/docs/usage/permissions-mapping.md and (for a longer read) https://eng.lyft.com/iam-whatever-you-say-iam-febce59d1e3b.

The rough idea is that you edit the permission_relationships.yaml file and cartography draws links from the IAM principals to the resource nodes based on the allowed actions:
image

This RPR change for KMS keys might not get merged here depending on how it impacts performance (which is why we have that yaml config file that lets users decide which permissions they care about syncing) but I wanted to let you know that this feature exists.

achantavy (changes requested):
Thanks for your PR! I left a few comments about performance, schema conventions, and potentially using resource permission relationships. Let me know if you have any questions and hope this helps.

achantavy:
Oh another thing, please write integration tests to ensure your new nodes and relationships are created as expected; see https://github.com/lyft/cartography/blob/master/tests/integration/cartography/intel/aws/test_eks.py.

cartography/intel/aws/kms.py
kedarghule: Removed (oops!)

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+        else:
+            return None
+    else:
+        return None
+
+
+@timeit
+def load_kms_keys(neo4j_session, data, region, current_aws_account_id, aws_update_tag):
+    ingest_keys = """
+    MERGE (kmskey:KMSKey{id:{KeyId}})
```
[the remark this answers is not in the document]
kedarghule: Done

cartography/intel/aws/kms.py
kedarghule: Done

cartography/intel/aws/kms.py
mpurusottamc: KeyId is just the Key Identifier, whereas ARN is fully formed URI for the Key. While populating Neo4j, we are using KeyId as the primary identifier.

As suggested, we have used UNWIND method for populating keys, grants and aliases.

cartography/intel/aws/kms.py
mpurusottamc: Each KMS Key Grant has an Id. but no ARN. That's why we are using the GrantId as the primary field.

cartography/intel/aws/kms.py
mpurusottamc: schema updated

cartography/intel/aws/kms.py
mpurusottamc: @achantavy you are right, we used S3 as the reference while populating KMS entities into Cartography. Our idea is to capture all the policies, statements, etc. which would allow the users for more fine grained queries

mpurusottamc:
@achantavy we have made all the required changes based on your feedback.

cartography/intel/aws/kms.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+
+    # neo4j does not accept datetime objects and values. This loop is used to convert
+    # these values to string.
+    for key in data:
+        key['CreationDate'] = str(key['CreationDate'])
+        key['DeletionDate'] = str(key.get('DeletionDate'))
+        key['ValidTo'] = str(key.get('ValidTo'))
+
+    # The owner data returned by the API maps to the aws account nickname and not the IAM user
```
achantavy, line 299: I see that this comment was copied from the s3 sync, is it still relevant here?

achantavy (dismissed):
Thank you very much for your contribution! :)

cartography/intel/aws/kms.py
kedarghule: No it's not. Removed it in latest commit :)

achantavy: I think you need to update your branch with the latest master and then I can merge it in.

achantavy (approved):

cartography/intel/aws/kms.py
kedarghule: @achantavy  Done :D

PR #528, opened 2021-02-10
Allow AWS tag sync to accept custom mappings

achantavy:
Callers to resourcegrouptaggingapi.sync() can now specify an optional parameter for custom tag-to-resourcetype mappings. This allows them to run the AWS tag sync with a subset of the mappings if desired.

The default behavior of this sync (using the `TAG_RESOURCE_TYPE_MAPPINGS` dict) is unaffected by this PR.

### DIFF
--- a/cartography/intel/aws/resourcegroupstaggingapi.py
+++ b/cartography/intel/aws/resourcegroupstaggingapi.py
@@ -121,7 +121,12 @@ def cleanup(neo4j_session, common_job_parameters):
 
 @timeit
 def sync(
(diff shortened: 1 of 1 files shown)


achantavy:
Note: we do have integration tests on AWS tags: https://github.com/lyft/cartography/blob/cc9d88b07bd307e4c6f7aad22a7556c00c499d3f/tests/integration/cartography/intel/aws/test_resourcegroupstaggingapi.py#L19 and these are passing :woohoo:

kunaals (approved):

PR #533, opened 2021-02-24
initial fix for single account sync #513

krisek:
With the hope this time it doesn't get staled.


achantavy:
Hey @krisek, instead of adding a `skip_on_syncing_single_account` param, I'd rather make sure each of these cleanup queries only affects assets in a single account; I see that you did that already in some of these. 

We should update the dev docs with this practice and update the cleanups accordingly. Let's talk on Slack or at the meeting if you have questions.

krisek:
Ok, I updated our thread in Slack, I think it will be chalegning to manage without this flag, but let's discuss.

PR #534, opened 2021-02-24
Fix  #319 - Add  Support for AWS EMR

### LINKED ISSUE #319: Add support for AWS Elastic Map Reduce (EMR)
*Title*: *Add support for AWS Elastic Map Reduce*

*Description*:
would be used.

We need support for AWS Elastic Map Reduce clusters and instances.

[optional *Relevant Links*:]
> Any extra documentation required to understand the issue.
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/emr.html
- https://docs.aws.amazon.com/emr/latest/APIReference/API_ListClusters.html
-  https://docs.aws.amazon.com/emr/latest/APIReference/API_ListInstances.html

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
Tested locally as well as in staging AWS environment. One issue I encountered is throttling from the EMR API, which appears to be  more aggressive than that of other AWS APIs. I have hardcoded 1 second sleeps between EMR operations to mitigate this, which appears to work.


cartography/intel/aws/emr.py
```diff
@@ -0,0 +1,88 @@
...
+        SET cluster.lastupdated         = {aws_update_tag}
+        WITH cluster
+
+        MATCH (owner:AWSAccount{id: {AWS_ACCOUNT_ID}})
+        MERGE (owner)-[r:RESOURCE]->(cluster)
+        ON CREATE SET r.firstseen       = timestamp()
+        SET r.lastupdated               = {aws_update_tag}
+    """
+
+    logger.debug("Loading EMR clusters for region '%s' into graph.", region)
```
achantavy, line 57: Let's increase this to INFO level and also log the count of clusters being loaded.

cartography/intel/aws/emr.py
```diff
@@ -0,0 +1,88 @@
...
+from typing import List
+
+from cartography.util import aws_handle_regions
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+# EMR API is subject to aggressive throttling, so we need to sleep a second between each call.
+list_sleep = 1
```
achantavy, line 13: Nit: use all caps for constants.

achantavy (changes requested):
Minor minor comments but looks good!

achantavy:
Oh yeah, and prefix the PR title with #319.

achantavy (approved):

PR #535, opened 2021-02-25
--neo4j-max-connection-lifetime config param, improve GraphStatement.run()

achantavy:
## Background
When migrating our single compute instance cartography+neo4j setup to K8s, we observed `ServiceUnavailable` exceptions on many different occasions. In our setup, the Neo4j DB was fronted by an AWS NLB. After much debugging, I found that

> Elastic Load Balancing sets the idle timeout value for TCP flows to 350 seconds. You cannot modify this value."

([AWS ELB+NLB docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#connection-idle-timeout))

As mentioned in https://github.com/neo4j/neo4j-python-driver/issues/205, the Neo4j driver maintains a pool of TCP connections for speed, setting a default max connection lifetime value for 1 hour before it considers a connection dead.

> NLBs have an idle timeout of 350 seconds which cannot be changed. When connections expire through idle timeout, NLBs terminate the connections silently. An application that is not aware of this timeout would attempt to send data to the same socket. At that point, the NLB would notify the application that the connection has been terminated by sending it an RST packet.

([tenable blog](https://medium.com/tenable-techblog/lessons-from-aws-nlb-timeouts-5028a8f65dda#))

In our case, a default neo4j driver config communicating with an AWS NLB would attempt to use a TCP connection from its connection pool that is actually stale, resulting in a RST.

## What this PR does
- Exposes a --neo4j-max-connection-lifetime parameter for cartography clients that need to specify connection lifetimes other than the default 1 hour. See https://neo4j.com/docs/driver-manual/1.7/client-applications/#driver-config-connection-pool-management.
- (Unrelated) Modifies `GraphStatement.run()` to use explicit neo4j transactions. This function is currently used for analysis and cleanup jobs, and is a good fit for explicit transactions.

I have tested this change and it has resolved at least one network connectivity issue that I previously reported. I'm really excited about testing this on the other connectivity issues I've submitted.

### DIFF
--- a/cartography/graph/statement.py
+++ b/cartography/graph/statement.py
@@ -1,6 +1,8 @@
 import json
 import logging
 
+import neo4j
+
 logger = logging.getLogger(__name__)
 
 
@@ -63,9 +65,9 @@ class GraphStatement:
         """
         Non-iterative statement execution.
         """
-        return session.run(self.query, self.parameters).consume()
+        return session.run(self.query, self.parameters)
 
-    def _run_iterative(self, session):
+    def _run_iterative(self, session: neo4j.Session) -> None:
         """
         Iterative statement execution.
 
@@ -73,17 +75,11 @@ class GraphStatement:
         """
         self.parameters["LIMIT_SIZE"] = self.iterationsize
 
-        # TODO improve this
-        done = False
-        while not done:
-            results = self._run(session)
-            for r in results:
-                total_deleted = int(r['TotalCompleted'])
-                logger.debug("Deleted %d items", total_deleted)
-                if total_deleted == 0:
-                    done = True
-
-                break
+        total_deleted = -1
+        while total_deleted != 0:
+            result: neo4j.Record = self._run(session).single()
+            total_deleted = int(result['TotalCompleted'])
+            logger.debug("Deleted %d items", total_deleted)
 
     @classmethod
     def create_from_json(cls, json_obj):
--- a/cartography/graph/statement.py
+++ b/cartography/graph/statement.py
@@ -78,17 +78,18 @@ class GraphStatement:
         """
         self.parameters["LIMIT_SIZE"] = self.iterationsize
 
-        total_completed = -1
-        while total_completed != 0:
+        while True:
             result: neo4j.StatementResult = self._run(tx)
-
             record: neo4j.Record = result.single()
+
             # TODO: use the BoltStatementResultSummary object to determine the number of items processed
             total_completed = int(record['TotalCompleted'])
             logger.debug("Processed %d items", total_completed)
 
             # Ensure network buffers are cleared
             result.consume()
+            if total_completed == 0:
+                break
 
     @classmethod
     def create_from_json(cls, json_obj):
--- a/cartography/graph/statement.py
+++ b/cartography/graph/statement.py
@@ -41,14 +41,17 @@ class GraphStatement:
         tmp.update(parameters)
         self.parameters = tmp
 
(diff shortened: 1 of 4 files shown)


achantavy:
@jg10: setting the max_connection_lifetime param to `$AWS_NLB_TIMEOUT - 10 seconds` solved a ServiceUnavailable problem for me in the GCP sync. Maybe this helps you solve https://lyftoss.slack.com/archives/CTZUQL0KX/p1611954942028100?thread_ts=1610382981.311800&cid=CTZUQL0KX?

@voutilad: I think I'm getting closer to solving the connection problems.

cartography/graph/statement.py
```diff
@@ -59,30 +64,31 @@ def as_dict(self):
...
-            done = False
-            for r in results:
-                if int(r['TotalCompleted']) == 0:
-                    done = True
+        total_completed = -1
+        while total_completed != 0:
+            result: neo4j.StatementResult = self._run(tx)
+
+            record: neo4j.Record = result.single()
+            # TODO: use the BoltStatementResultSummary object to determine the number of items processed
```
achantavy, line 86: Out of scope for this PR, but it would be great to instead use the built-in BoltStatementResultSummary object and get the # nodes/# rels updated instead of relying on the query writer to do a `MATCH/MERGE ... RETURN ...count(*) as TotalCompleted` just for this logic to work.

cartography/graph/statement.py
```diff
@@ -59,30 +64,31 @@ def as_dict(self):
...
 
-        # TODO improve this
-        done = False
-        while not done:
-            results = self._run(session)
-            done = False
-            for r in results:
-                if int(r['TotalCompleted']) == 0:
-                    done = True
+        total_completed = -1
```
kunaals, line 81: might be cleaner here to just break when the total completed is 0, ie:
```
        while True:
            result: neo4j.StatementResult = self._run(tx)
            record: neo4j.Record = result.single()
            # TODO: use the BoltStatementResultSummary object to determine the number of items processed
            total_completed = int(record['TotalCompleted'])
            logger.debug("Processed %d items", total_completed)

            # Ensure network buffers are cleared
            result.consume()
            if total_completed == 0:
                break
```

cartography/sync.py
```diff
@@ -106,6 +106,7 @@ def run_with_config(sync, config):
         neo4j_driver = GraphDatabase.driver(
             config.neo4j_uri,
             auth=neo4j_auth,
+            max_connection_lifetime=config.neo4j_max_connection_lifetime,
```
kunaals, line 109: the default for `config.neo4j_max_connection_lifetime` is none - if `max_connection_lifetime` gets set to none is that alright? would it be better to explicitly set this to 3600 in CLI and config? that would make the docstring update in config.py make more sense to users as well

kunaals (dismissed):
looks good left a suggestion and a question - will approve so you're not blocked

cartography/sync.py
achantavy: True, yeah will do.

cartography/graph/statement.py
achantavy: I guess it's a style thing, but seeing code like `while True` scares me. I don't have a strong opinion either way though.

cartography/sync.py
achantavy: Done, see cli.py.

kunaals (approved):

PR #536, opened 2021-02-25
Type hints, only enforced  for eks initially

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
I've applied this only to EKS:

```
[mypy-cartography.intel.aws.eks]
ignore_errors = false
```

Note: I ignored the hint for `logging = any(filter(lambda x: 'audit' in x['types'] and x['enabled'], cluster_logging))  # type: ignore`. No idea what that is supposed to be!

### DIFF
--- a/cartography/intel/aws/eks.py
+++ b/cartography/intel/aws/eks.py
@@ -1,4 +1,10 @@
 import logging
+from typing import Any
+from typing import Dict
+from typing import List
+
+import boto3
+import neo4j
 
 from cartography.util import aws_handle_regions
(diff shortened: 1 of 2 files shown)


achantavy (approved):
Ha yeah I faced the same challenge with the lambda.

marco-lancini:
Ha! I see it might be tricky to unwrap.
Basically it filters the dictionary `cluster_logging` to see if at least one entry has audit logging set to `Enabled`.
In the end, `logging` will be a bool.

PR #538, opened 2021-02-26
Type Hints for AWs module: KMS, Lambda, Organizations

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
Also fixed a type that was wrong- regions should be list of str!

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -710,7 +710,7 @@ def sync_user_access_keys(
 
 @timeit
 def sync(
-    neo4j_session: neo4j.Session, boto3_session: boto3.session.Session, regions: str, account_id: str,
+    neo4j_session: neo4j.Session, boto3_session: boto3.session.Session, regions: List[str], account_id: str,
     update_tag: int, common_job_parameters: Dict,
 ) -> None:
     logger.info("Syncing IAM for account '%s'.", account_id)
--- a/cartography/intel/aws/kms.py
+++ b/cartography/intel/aws/kms.py
@@ -1,6 +1,15 @@
 import json
 import logging
+from typing import Any
+from typing import Dict
+from typing import Generator
+from typing import List
+from typing import Optional
+from typing import Tuple
 
+import boto3
+import botocore
+import neo4j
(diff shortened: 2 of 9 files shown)


cartography/intel/aws/iam.py
```diff
@@ -710,7 +710,7 @@ def sync_user_access_keys(
 
 @timeit
 def sync(
-    neo4j_session: neo4j.Session, boto3_session: boto3.session.Session, regions: str, account_id: str,
+    neo4j_session: neo4j.Session, boto3_session: boto3.session.Session, regions: List[str], account_id: str,
```
achantavy, line 713: Can we standardize on `current_aws_account_id`?

cartography/intel/aws/kms.py
```diff
@@ -321,7 +341,10 @@ def sync_kms_keys(neo4j_session, boto3_session, region, current_aws_account_id,
 
 
 @timeit
-def sync(neo4j_session, boto3_session, regions, current_aws_account_id, aws_update_tag, common_job_parameters):
+def sync(
+    neo4j_session: neo4j.Session, boto3_session: boto3.session.Session, regions: List[str], current_aws_account_id: str,
+    aws_update_tag: int, common_job_parameters: Dict,
```
achantavy, line 346: Can we standardize on `update_tag`?

achantavy (changes requested):

cartography/intel/aws/iam.py
FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg: Yeah, I was gonna do that in a future PR where I homogenize variable and function names

cartography/intel/aws/kms.py
FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg: Yeah, I was gonna do that in a future PR where I homogenize variable and function names

achantavy: Roger

achantavy (approved):

PR #541, opened 2021-03-01
Type Hints for AWS modules, part 5 - permission_relationships

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
Just `permission_relationships`. There are comments on variable naming, but it is wrong. I think this  file may need greater scrutiny, but we can do type hints for now and deal with that later?


achantavy (approved):
🚢

achantavy (approved):

PR #547, opened 2021-03-03
Fix the default value of requested_syncs to RESOURCE_FUNCTIONS.keys()

ccrims0n:
Currently, if the `--aws-requested-syncs` param is not specified then no AWS resource is synced rather than the intended functionality where all the AWS resources should be synced.

### DIFF
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -189,7 +189,7 @@ def start_aws_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
             ),
         )
 
-    requested_syncs: List[str] = []
+    requested_syncs: List[str] = RESOURCE_FUNCTIONS.keys()
     if config.aws_requested_syncs:
         requested_syncs = parse_and_validate_aws_requested_syncs(config.aws_requested_syncs)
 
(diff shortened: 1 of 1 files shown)


achantavy:
Thanks very much for helping with this. I missed adding an integration test for calling `start_aws_ingestion` and checking how `aws_requested_syncs` gets passed.

This mIght be a silly question, but doesn't the default list of syncs already get set in `_sync_one_account()`, or is that wrong? https://github.com/lyft/cartography/blob/4b62dd78c40c59a27a4f974f9ad4272343c8448d/cartography/intel/aws/__init__.py#L44

Admittedly I haven't had a chance to fully end-to-end test this one yet (good thing we haven't cut a release if it has this bug 😅)

ccrims0n:
The default value in `_sync_one_account()` is only set when the parameter is not passed. However, in the call to `_sync_one_account()` from `_sync_multiple_accounts()`, `aws_requested_syncs` is passed as an argument which overrides the default value. In case the `--aws-requested-syncs ` param is not present the value of `aws_requested_syncs` is set to empty list.
So the flow basically looks like
https://github.com/lyft/cartography/blob/4b62dd78c40c59a27a4f974f9ad4272343c8448d/cartography/intel/aws/__init__.py#L192-L202

`aws_requested_syncs` value is overriden here
https://github.com/lyft/cartography/blob/4b62dd78c40c59a27a4f974f9ad4272343c8448d/cartography/intel/aws/__init__.py#L133-L140

achantavy (approved):
Ah, that's embarassing. Thanks for the fix!!

PR #548, opened 2021-03-03
Fix arguments for elasticache cleanup()

ccrims0n:


achantavy (approved):
D'oh, good catch, thanks for this!

PR #550, opened 2021-03-04
Add DigitalOcean Intel module

lgomezma:
The intention of this PR is to add initial support for Intel from DigitalOcean. For that I'm creating a new Intel module making use of the `python-digitalocean` library. This Intel module supports for now only a limited subset of resources in DigitalOcean, which are Account, Projects and Droplets. The initial intention was to only add Droplets on the first iteration, but I found it better to respect DO object hierarchy and start from the top as Account has Projects and a Project has Droplets.


cartography/data/jobs/cleanup/digitalocean_account_cleanup.json
```diff
@@ -0,0 +1,11 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (n:Account{id: {DO_ACCOUNT_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
achantavy, line 4: Should this be `match(n:DOAccount)`?

achantavy (changes requested):
Quick first pass. Thanks for the PR, this is exciting!

cartography/data/jobs/cleanup/digitalocean_account_cleanup.json
lgomezma: Indeed, good catch!

achantavy (approved):

PR #552, opened 2021-03-04
okta type hints

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
As a future todo, the okta module has at least one bug, and is overly nested. does anybody actually use it?


kimanthony:
👍

kimanthony (approved):

achantavy:
We use it -- heavily

PR #554, opened 2021-03-05
Add support for Microsoft Azure Compute Resources

mpurusottamc:
This PR is to add initial support of Azure. This intel module currently support only a limited set of Compute related resources like VirtualMachines, DataDisks, Disks and Snapshots. In future, we will extend this furture to include other compute related resources

Currently, the hierarchy is AzureTenant -> AzureSubscription -> VirtualMachine

Also, once reviewed, we want to open pull requests for other resources like Azure Storage, SQL and CosmosDB.


cartography/intel/azure/compute.py
```diff
@@ -0,0 +1,227 @@
...
+        client = get_client(credentials, subscription_id)
+        vm_list = list(map(lambda x: x.as_dict(), client.virtual_machines.list_all()))
+
+        for vm in vm_list:
+            x = vm['id'].split('/')
+            vm['resource_group'] = x[x.index('resourceGroups') + 1]
+
+        return vm_list
+
+    except Exception as e:
```
achantavy, line 31: Can we scope this to a specific exception?

cartography/intel/azure/compute.py
```diff
@@ -0,0 +1,227 @@
...
+        client = get_client(credentials, subscription_id)
+        disk_list = list(map(lambda x: x.as_dict(), client.disks.list()))
+
+        for disk in disk_list:
+            x = disk['id'].split('/')
+            disk['resource_group'] = x[x.index('resourceGroups') + 1]
+
+        return disk_list
+
+    except Exception as e:
```
achantavy, line 109: Same on using a specific exception.

cartography/intel/azure/compute.py
```diff
@@ -0,0 +1,227 @@
...
+        logger.warning(f"Error while retrieving virtual machines - {e}")
+        return []
+
+
+def load_vms(neo4j_session: neo4j.Session, subscription_id: str, vm_list: List[Dict], update_tag: int) -> None:
+    ingest_vm = """
+    UNWIND {vms} AS vm
+    MERGE (v:VirtualMachine{id: vm.id})
```
achantavy, line 39: [question] What do you think about making this `AzureVirtualMachine`, or making this have 2 labels: `VirtualMachine` (like a generic label regardless of the clouds we support) _and_ `AzureVirtualMachine`.

cartography/intel/azure/credentials.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+
+            profile = get_cli_profile()
+
+            return Credentials(
+                arm_credentials, aad_graph_credentials, tenant_id=tenant_id,
+                current_user=profile.get_current_account_user(), subscription_id=subscription_id,
+            )
+
+        except Exception as e:
```
achantavy, line 45: Same comment on specific exceptions

cartography/intel/azure/credentials.py
```diff
@@ -0,0 +1,170 @@
+import logging
```
achantavy, line 1: Could we create a folder for `intel/azure/util` and put `credentials.py` in there?

cartography/intel/azure/credentials.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+
+            profile = get_cli_profile()
+
+            return Credentials(
+                arm_credentials, aad_graph_credentials, tenant_id=tenant_id,
+                current_user=profile.get_current_account_user(),
+            )
+
+        except Exception as e:
```
achantavy, line 87: Here I think we should catch the specific exception if possible, log an error, then `raise` the original exception.

cartography/intel/azure/credentials.py
```diff
@@ -0,0 +1,170 @@
...
+        elif 'tenant_id' in self.aad_graph_credentials.token:
+            return self.aad_graph_credentials.token['tenant_id']
+        else:
+            # This is a last resort, e.g. for MSI authentication
+            try:
+                h = {'Authorization': 'Bearer {}'.format(self.arm_credentials.token['access_token'])}
+                r = requests.get('https://management.azure.com/tenants?api-version=2020-01-01', headers=h)
+                r2 = r.json()
+                return r2.get('value')[0].get('tenantId')
+            except Exception as e:
```
achantavy, line 126: Specific exception

cartography/intel/azure/subscription.py
```diff
@@ -0,0 +1,97 @@
...
+def get_all_azure_subscriptions(credentials: Credentials) -> List[Dict]:
+    # Create the client
+    client = SubscriptionClient(credentials.arm_credentials)
+
+    # Get all the accessible subscriptions
+    subs = list(client.subscriptions.list())
+
+    if not subs:
+        logger.warning('failed to fetch subscriptions for the credentials')
+        raise Exception('The provided credentials do not have access to any subscriptions')
```
achantavy, line 24: Specific exception

cartography/intel/azure/subscription.py
```diff
@@ -0,0 +1,97 @@
...
+def get_current_azure_subscription(credentials: Credentials, subscription_id: str) -> List[Dict]:
+    # Create the client
+    client = SubscriptionClient(credentials.arm_credentials)
+
+    # Get all the accessible subscriptions
+    sub = client.subscriptions.get(subscription_id)
+    print(sub)
+
+    if not sub:
+        raise Exception(f'The provided credentials do not have access to this subscription: {subscription_id}')
```
achantavy, line 47: Specific exception

achantavy (changes requested):
Super exciting to see another cloud added to cartography :-)

Mostly left comments on exception handling but this is looking great so far.

cartography/intel/azure/compute.py
mpurusottamc: updated code to use specific exception

cartography/intel/azure/compute.py
mpurusottamc: updated code to use specific exception

cartography/intel/azure/credentials.py
mpurusottamc: updated code to use specific exception

cartography/intel/azure/credentials.py
mpurusottamc: moved file to util folder

cartography/intel/azure/credentials.py
mpurusottamc: fixed

cartography/intel/azure/credentials.py
mpurusottamc: updated code to use specific exception

cartography/intel/azure/subscription.py
mpurusottamc: updated code to use specific exception

cartography/intel/azure/subscription.py
mpurusottamc: updated code to use specific exception

cartography/intel/azure/compute.py
mpurusottamc: I like AzureVirtualMachine more than VirtualMachine :). I have renamed `VirtualMachine` to `AzureVirtualMachine`. I will create an issue for so that we can create a generic Label for VirtualMachine across clouds.

mpurusottamc:
@achantavy Thanks for reviewing the pull request. All the feedback has been incorporated.

cartography/intel/azure/util/credentials.py
```diff
@@ -0,0 +1,173 @@
...
+                    {e}',
+                )
+
+            raise e
+
+
+class Credentials:
+
+    def __init__(
+        self, arm_credentials: Any, aad_graph_credentials: Any, tenant_id: str = None, subscription_id: str = None,
```
achantavy, line 104: I won't block on this now, but we should aim to have these credential objects have typehints of the actual object type and not `Any`.

cartography/intel/azure/util/credentials.py
```diff
@@ -0,0 +1,173 @@
...
+                arm_credentials, aad_graph_credentials, tenant_id=tenant_id,
+                current_user=profile.get_current_account_user(), subscription_id=subscription_id,
+            )
+
+        except HttpResponseError as e:
+            if ', AdalError: Unsupported wstrust endpoint version. ' \
+                    'Current support version is wstrust2005 or wstrust13.' in e.args:
+                logger.error(
+                    f'You are likely authenticating with a Microsoft Account. \
+                    This authentication mode only support Azure Active Directory principal authentication.\
```
achantavy, line 51: Nit: s/only support/only supports/

cartography/intel/azure/util/credentials.py
```diff
@@ -0,0 +1,173 @@
...
+from msrestazure.azure_active_directory import AADTokenCredentials
+
+
+logger = logging.getLogger(__name__)
+AUTHORITY_HOST_URI = 'https://login.microsoftonline.com'
+
+
+class Authenticator:
+
+    def authenticate_cli(self) -> Any:
```
achantavy, line 22: If I read this correctly, the return type here should be `Credential` and not `Any`

cartography/intel/azure/util/credentials.py
```diff
@@ -0,0 +1,173 @@
...
+            profile = get_cli_profile()
+
+            return Credentials(
+                arm_credentials, aad_graph_credentials, tenant_id=tenant_id,
+                current_user=profile.get_current_account_user(), subscription_id=subscription_id,
+            )
+
+        except HttpResponseError as e:
+            if ', AdalError: Unsupported wstrust endpoint version. ' \
+                    'Current support version is wstrust2005 or wstrust13.' in e.args:
```
achantavy, line 48: Nit: s/Current support version/Current supported version/

cartography/intel/azure/util/credentials.py
```diff
@@ -0,0 +1,173 @@
...
+                arm_credentials, aad_graph_credentials, tenant_id=tenant_id,
+                current_user=profile.get_current_account_user(),
+            )
+
+        except HttpResponseError as e:
+            if ', AdalError: Unsupported wstrust endpoint version. ' \
+                    'Current support version is wstrust2005 or wstrust13.' in e.args:
+                logger.error(
+                    f'You are likely authenticating with a Microsoft Account. \
+                    This authentication mode only support Azure Active Directory principal authentication.\
```
achantavy, line 94: Nit: s/only support/only supports/

cartography/intel/azure/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    common_job_parameters: Dict,
+) -> None:
+    compute.sync(neo4j_session, credentials.arm_credentials, subscription_id, update_tag, common_job_parameters)
+
+
+def _sync_tenant(
+    neo4j_session: neo4j.Session, tenant_id: str, current_user: str, update_tag: int,
+    common_job_parameters: Dict,
+) -> None:
+    logger.debug("Syncing Azure Tenant: %s", tenant_id)
```
achantavy, line 29: Nonblocker - could be worth making this logger.info() and not debug

cartography/intel/azure/tenant.py
```diff
@@ -0,0 +1,49 @@
...
+    return credentials.get_tenant_id()
+
+
+def load_azure_tenant(neo4j_session: neo4j.Session, tenant_id: str, current_user: str, update_tag: int) -> None:
+    query = """
+    MERGE (at:AzureTenant{id: {TENANT_ID}})
+    ON CREATE SET at.firstseen = timestamp()
+    SET at.lastupdated = {update_tag}
+    WITH at
+    MERGE (ap:AzurePrincipal{email: {CURRENT_USER}})
```
achantavy, line 23: Do
```
MERGE (ap:AzurePrincipal{id: {CURRENT_USER}})
ON CREATE SET ap.email = {CURRENT_USER},
ap.lastupdated ...
```

Add an index for `id`. You can keep the existing index on `AzurePrincipal(email)`.

cartography/cli.py
```diff
@@ -126,6 +126,45 @@ def _build_parser(self):
                 'respects the AWS CLI/SDK environment variables and does not override them.'
             ),
         )
+        parser.add_argument(
+            '--azure-sync-all-subscriptions',
+            action='store_true',
+            help=(
+                'Enable Azure sync for all discovered profiles. When this parameter is supplied cartography will '
```
achantavy, line 133: Do you mean "all discovered subscriptions" instead of "all discovered profiles"?

cartography/cli.py
```diff
@@ -314,6 +353,16 @@ def main(self, argv):
             # No need to store the returned value; we're using this for input validation.
             parse_and_validate_aws_requested_syncs(config.aws_requested_syncs)
 
+        # Okta config
```
achantavy, line 356: I think you copy pasted this by accident

achantavy (changes requested):
Nearly there! Some nitpicky doc stuff

mpurusottamc:
@achantavy Updated code to resolve all the comments above.

cartography/intel/azure/util/credentials.py
mpurusottamc: fixed in latest version

cartography/intel/azure/util/credentials.py
mpurusottamc: fixed in latest version

cartography/intel/azure/util/credentials.py
mpurusottamc: fixed in latest version

cartography/intel/azure/tenant.py
mpurusottamc: neat. I have updated code to add this logic.

cartography/intel/azure/__init__.py
mpurusottamc: done

cartography/cli.py
mpurusottamc: fixed in latest version

cartography/cli.py
mpurusottamc: :)

i have updated it now.

cartography/intel/azure/util/credentials.py
mpurusottamc: updated in latest commit

cartography/intel/azure/util/credentials.py
mpurusottamc: makes sense, will do this as a follow-up.

setup.py
```diff
@@ -41,7 +41,14 @@
         "requests>=2.22.0",
         "statsd",
         "packaging",
+        "cryptography<3.4,>=3.2",
         "python-digitalocean>=1.16.0",
+        "PyJWT==1.7.1",
```
achantavy, line 46: Does this need to be pinned? Can it be `>=1.7.1` instead?

achantavy (dismissed):
One final minor comment from me. Ship it :)

setup.py
mpurusottamc: azure-cli-core has pinned it to 1.7.1. I tried with `>=1.7.1` and it works. But, there's an error notice in logs while installing the dependencies. I did not want to be the cause of someone's CI/CD pipeline failure :) 

What do you recommend? @achantavy

achantavy: Pin it to 1.7.1 and add a comment on why -- that should be good enough

mpurusottamc: @achantavy I have added a comment and pinned the version. Please review and merge.

achantavy (approved):
🥂

PR #555, opened 2021-03-06
512 add internet gateway support

cpollard0:
Add internet gateway support to cartography


cpollard0:
Closing because I forked up the commits with my GitHub user name. Will submit a cleaner PR

PR #530, opened 2021-02-16
Implement --aws-requested-syncs, standardize AWS sync function param names

achantavy:
Goals
------
1. Users can specify the AWS assets that they want to sync (default = all)
1. Standardize the names of parameters used in AWS sync functions.

This PR is an alternate approach to #467. Rather than rewrite all function signatures to introduce context object,  we use `kwargs` to assign AWS sync args appropriately, and we use Python introspection to enforce standard param names in these functions.

This allows cartography to run user-specified AWS syncs like this:

```
for func_name in aws_requested_syncs:
    if func_name in RESOURCE_FUNCTIONS:
        RESOURCE_FUNCTIONS[func_name](**sync_args)
```

--- 

This code change also enables the following future plans:

1. Users can specify the AWS regions that they want to sync (default = all)
1. Users can specify all of the above on a per-account basis.

PR #556, opened 2021-03-06
Add in AWS internet gateway support

### LINKED ISSUE #512: AWS: Add support for Internet Gateway
*Title*: *Add support for Internet Gateway*

*Description*:
As of now, Cartography doesn't support Internet Gateway.
An internet gateway allows communication between your VPC and the internet.

The relationships that I've identified are

(AWSAccount) - [RESOURCE] -> (EC2InternetGateway)
(EC2InternetGateway) -[ATTACHED_TO]-> (AWSVpc)

*Relevant Links*
[Internet Gateway](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Internet_Gateway.html)

cpollard0:
Add in internet gateway support; closes #512


cartography/intel/aws/ec2/internet_gateways.py
```diff
@@ -0,0 +1,83 @@
...
+
+@timeit
+def load_internet_gateways(
+    neo4j_session: neo4j.Session, data: List[Dict], region: str,
+    current_aws_account_id: str, update_tag: int,
+) -> None:
+    logger.debug("Loading %d Internet Gateways in %s.", len(data), region)
+    query = """
+            MERGE (ig:AWSInternetGateway{id: {InternetGatewayId}})
+            ON CREATE SET
```
achantavy, line 31: Is the `id` the same as the gateway's `arn`? If so, add `ON CREATE  SET ig.arn = {InternetGatewayId}`

cartography/intel/aws/ec2/internet_gateways.py
```diff
@@ -0,0 +1,83 @@
...
+    client = boto3_session.client('ec2', region_name=region, config=get_botocore_config())
+    return client.describe_internet_gateways()['InternetGateways']
+
+
+@timeit
+def load_internet_gateways(
+    neo4j_session: neo4j.Session, data: List[Dict], region: str,
+    current_aws_account_id: str, update_tag: int,
+) -> None:
+    logger.debug("Loading %d Internet Gateways in %s.", len(data), region)
```
achantavy, line 28: Let's do a `logger.info` instead of debug here.

cartography/intel/aws/ec2/vpc.py
```diff
@@ -32,11 +32,11 @@ def _get_cidr_association_statement(block_type: str) -> str:
         new_block.cidr_block = block_data.$block_cidr,
         new_block.block_state = block_data.$state_name.State,
         new_block.block_state_message = block_data.$state_name.StatusMessage,
-        new_block.lastupdated = {aws_update_tag}
+        new_block.lastupdated = {update_tag}
```
achantavy, line 35: Ooh good catch; I wonder if there are other places where this needs to be  updated.

tests/integration/cartography/intel/aws/ec2/test_internet_gateway.py
```diff
@@ -0,0 +1,32 @@
...
+
+    expected_nodes = {
+        "igw-1234XXX",
+        "igw-7e3a7c18",
+        "igw-f1c81494",
+    }
+
+    nodes = neo4j_session.run(
+        """
+        MATCH (n:AWSInternetGateway) RETURN n.id;
```
achantavy, line 27: Can you add another test for 

```
(:AWSInternetGateway)<-[:RESOURCE]-(:AWSAccount)
```
?

cartography/intel/aws/ec2/internet_gateways.py
```diff
@@ -0,0 +1,83 @@
...
+    query = """
+            MERGE (ig:AWSInternetGateway{id: {InternetGatewayId}})
```
achantavy, line 30: It might be worth rewriting this to use `UNWIND` to something like this:  https://github.com/lyft/cartography/blob/047a6bd1820e846e70bda4c6ddc18027da39e20a/cartography/intel/aws/elasticache.py#L53

This way, the assets get loaded to the graph in one large transaction instead of N transactions, which  can be [significantly faster](https://achantavy.github.io/cartography/performance/cypher/neo4j/2020/07/19/loading-7m-items-to-neo4j-with-and-without-unwind.html).

achantavy (changes requested):
Thanks so much for your PR! Left a few minor comments.

achantavy:
Oh! Don't forget to add an index on `id`: https://github.com/lyft/cartography/blob/047a6bd1820e846e70bda4c6ddc18027da39e20a/cartography/data/indexes.cypher

If there is enough data, queries on nodes without indexes may take _days_ to finish and have unpredictable behavior - see https://github.com/lyft/cartography/pull/505.

cpollard0:
Thanks for putting out a really slick tool! Happy to contribute

cartography/intel/aws/ec2/vpc.py
cpollard0: Is the goal to get rid of all instances of `aws_update_tag` for just `update_tag`? Perhaps just a blank grep | sed should do.

cartography/intel/aws/ec2/internet_gateways.py
cpollard0: The ID is something like `igw-c0a643a9` and the Arn is in the format `arn:$partition:ec2:$region:$account:internet-gateway/$igw-id`. 

Do you want to build the Arn, or just leave it off if we're not receiving it verbatim from the API call?

cartography/intel/aws/ec2/internet_gateways.py
cpollard0: Thanks for sharing the informative blog post!

tests/integration/cartography/intel/aws/ec2/test_internet_gateway.py
cpollard0: I noticed that there seemed to be a few instances where we needed to set up an account in as part of our tests, so I created a common function for that. Let me know if that isn't the direction you wanted to head.

cartography/intel/aws/ec2/internet_gateways.py
achantavy: Hm.. I'm leaning toward building the ARN ourselves, but I don't have a super strong opinion - up to you!

cartography/intel/aws/ec2/vpc.py
achantavy: Yeah, that probably would be fine. This is a bit of a special case in that the "`load` string" is separate from the `load()` function so when I did #530 I missed it. Thanks for catching it.

cartography/intel/aws/ec2/internet_gateways.py
achantavy: Actually I take that back - let's build the ARN so that we can load AWS tags for them basically effortlessly like this: https://github.com/lyft/cartography/blob/047a6bd1820e846e70bda4c6ddc18027da39e20a/cartography/intel/aws/resourcegroupstaggingapi.py#L49

cpollard0:
@achantavy - I think I've addressed all of the issues, let me know if you have additional feedback!

tests/integration/cartography/intel/aws/ec2/test_internet_gateway.py
achantavy: Sure, that works!

achantavy (approved):
Thanks so much!

PR #558, opened 2021-03-08
Add coverage hard lower limit

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
Tests already run coverage, let's enforce a lower limit. Currently we're at 39.5% for unit tests, and 55% for integration.


achantavy (commented):
Our test strategy so far has been to use integration tests on the `load` functions to ensure nodes and relationships are created with some degree of confidence, add mocks for code that ties different parts together (e.g. the `start` and `sync` functions), and unit tests for `transforms` (I should add this to the intel module guide, come to think of it).

Personally I don't find test coverage numbers to be super useful and I don't want devs getting to the point of writing trivially-passing, possibly brittle tests for the sake of hitting a number (as seen in many other projects), but I am open to other opinions. I guess setting a lower bound ensures that new code continues to have automated tests. What are your thoughts?

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
I agree with you on high coverage numbers being pointless. I was thinking of setting this as a lower bound; 50% coverage doesn't seem to be bad to aim for.

Alternatively, perhaps it would be more useful to combine coverage numbers for integration and unit? Pytest appears to be able to do this, though I'm not sure how'd we'd expose it in CI.

achantavy:
Yeah, let's establish a lower bound and revisit increasing that over time.

achantavy (approved):

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
:hammer:

Issue #410, opened 2020-09-28
Ec2RouteTable support


krisek:
*Description*:
At the moment Cartography doesn't give detailed information about traditionally routed environments (ie. environments not using Transit Gateway) in AWS. The aim is to develop the required code to import this information as well.

I identified the following relations so far

(*Ec2RouteTable*) <- [ASSOCIATED] - (EC2Subnet)

(*Ec2RouteTable*) - [ROUTES] -> (*EC2Route*)

(AWSVpc) - [RESOURCE] -> (*Ec2RouteTable*)

(AWSAccount) - [RESOURCE] -> (*Ec2RouteTable*)

(*EC2Route*) - [CIDR] -> (AWSCidrBlock)

(*EC2Route*) - [PREFIXLIST] -> (*AWSPrefixList*)

(*AWSPrefixList*) - [CIDR] -> (AWSCidrBlock)

(*EC2Route*) - [GATEWAY] -> (*InternetGateway|VPCPeering (!)|NATGateway*) (Relation doesn't exist for local prefixes.)

There is already one issue that needs to be clarified: VPC_PEERING is already existing as relation between AWSCidrBlocks, this has to be refactored.

*Relevant Links*:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html

achantavy:
Appreciate you taking the time to put this together. I think follow-up PRs should include just one new relationship each to make them easier to review.

> There is already one issue that needs to be clarified: VPC_PEERING is already existing as relation between AWSCidrBlocks, this has to be refactored.

The reasoning for this schema choice is that VPCs are peered through their CIDR blocks. The path looks something like 
```
(i) VpcA-[:BLOCK_ASSOCIATION]-CidrBlockA-[:VPC_PEERING]-CidrBlockB-[:BLOCK_ASSOCIATION]-VpcB
```
In this case, an instance connected to BlockA can directly talk to BlockB, and we wanted to make that as clear as possible.

What is your proposed refactoring? I **think** you're proposing to add a new "shortcut" relationship equivalent to (i):
```
VpcA-[:NAME_TO_BE_DECIDED]-VpcB
```

krisek:
> Appreciate you taking the time to put this together. I think follow-up PRs should include just one new relationship each to make them easier to review.

No problem. Separating PRs is agreed.
> 
[...]

PR #562, opened 2021-03-13
vpc_peerings implemented as nodes

krisek:
This would be the first step to implement Ec2RouteTables (#410).


achantavy (commented):
Hey, this looks good to me, but can you remind me again why we can't have this logic replace the existing vpc_peering.py, or is deleting that for a future PR?

krisek:
Hi, last time I was trying to delete vpc_peering.py it broke testing.

krisek:
Appearantly this time it worked, what should be the next step?

cartography/data/indexes.cypher
```diff
@@ -112,6 +112,7 @@ CREATE INDEX ON :OktaTrustedOrigin(id);
 CREATE INDEX ON :OktaAdministrationRole(id);
 CREATE INDEX ON :Package(id);
 CREATE INDEX ON :Package(name);
+CREATE INDEX ON :PeeringConnection(id);
```
achantavy, line 115: Nit: should we distinguish this somehow as an AWSPeeringConnection? Does this concept exist in other clouds?

achantavy (dismissed):
Thanks!

achantavy (approved):

PR #563, opened 2021-03-15
Added support for Azure SQL Resources

kedarghule:
This PR is to add initial support for Azure SQL resources. 

The resources populated in the graph db will be SQL Server, Server DNS Aliases, Azure Server AD Administrators, Recoverable Databases, Restorable Dropped Databases, Failover Groups, Elastic Pools, SQL Databases, Replication Links, Transparent Data Encryptions, Restore Points and Database Threat Detection Policy.


kedarghule:
@achantavy I have a question. So in this code, the unit-and-integration tests are failing with the message - Required test coverage of 38.0% not reached. Total coverage: 36.30%. I have written tests for every "ingestion script of neo4j" function like I did for the API Gateway code and like @mpurusottamc did for the AWS KMS and Azure VM codes. I noticed the setup.cfg file adds this 38% constraint. Can this be lowered?

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_server_list(credentials: Credentials, subscription_id: str) -> List[Dict]:
+    """
+    Returning the list of Azure SQL servers.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        server_list = list(map(lambda x: x.as_dict(), client.servers.list()))
```
ramonpetgrave64, line 36: Any particular reason to use `list(map(...))`, rather than list comprehension? I wouldn't block but list comprehension reads quicker.

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+@timeit
+def get_dns_aliases(credentials: Credentials, subscription_id: str, server: Dict) -> List[Dict]:
+    """
+    Returns details of the DNS aliases in a server.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        dns_aliases = list(
+            map(
```
ramonpetgrave64, line 119: Non-blocker: list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_ad_admins(credentials: Credentials, subscription_id: str, server: Dict) -> List[Dict]:
+    """
+    Returns details of the Server AD Administrators in a server.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        ad_admins = list(
```
ramonpetgrave64, line 139: Non-blocker: use list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_recoverable_databases(credentials: Credentials, subscription_id: str, server: Dict) -> List[Dict]:
+    """
+    Returns details of the Recoverable databases in a server.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        recoverable_databases = list(
```
ramonpetgrave64, line 163: Non-blocker: use list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_restorable_dropped_databases(credentials: Credentials, subscription_id: str, server: Dict) -> List[Dict]:
+    """
+    Returns details of the Restorable Dropped Databases in a server.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        restorable_dropped_databases = list(
```
ramonpetgrave64, line 190: Non-blocker: use list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_failover_groups(credentials: Credentials, subscription_id: str, server: Dict) -> List[Dict]:
+    """
+    Returns details of Failover groups in a server.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        failover_groups = list(
```
ramonpetgrave64, line 213: Non-blocker: use list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_elastic_pools(credentials: Credentials, subscription_id: str, server: Dict) -> List[Dict]:
+    """
+    Returns details of Elastic Pools in a server.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        elastic_pools = list(
```
ramonpetgrave64, line 231: Non-blocker: use list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_databases(credentials: Credentials, subscription_id: str, server: Dict) -> List[Dict]:
+    """
+    Returns details of Databases in a SQL server.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        databases = list(
```
ramonpetgrave64, line 249: Non-blocker: use list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_replication_links(credentials: Credentials, subscription_id: str, database: Dict) -> List[Dict]:
+    """
+    Returns the details of replication links in a database.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        replication_links = list(
```
ramonpetgrave64, line 586: Non-blocker: use list comprehension

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+
+
+@timeit
+def get_restore_points(credentials: Credentials, subscription_id: str, database: Dict) -> List[Dict]:
+    """
+    Returns the details of restore points in a database.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        restore_points_list = list(
```
ramonpetgrave64, line 630: Non-blocker: use list comprehension

cartography/data/jobs/cleanup/azure_sql_server_cleanup.json
```diff
@@ -0,0 +1,125 @@
...
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (:AzureSQLServer)<-[r:RESOURCE]-(:AzureSubscription{id: {AZURE_SUBSCRIPTION_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:AzureServerDNSAlias)<-[:USED_BY]-(:AzureSQLServer)<-[:RESOURCE]-(:AzureSubscription{id: {AZURE_SUBSCRIPTION_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
ramonpetgrave64, line 14: Any reason to include this in the queries? `<-[:RESOURCE]-(:AzureSubscription{id: {AZURE_SUBSCRIPTION_ID}})`
I think the queries should match just fine without the extra constraint.

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,855 @@
...
+@timeit
+def get_server_list(credentials: Credentials, subscription_id: str) -> List[Dict]:
+    """
+    Returning the list of Azure SQL servers.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        server_list = list(map(lambda x: x.as_dict(), client.servers.list()))
+
+    except HttpResponseError as e:
```
ramonpetgrave64, line 38: Non blocker: Can you try and get specific on some of the response error reasons? example:
https://github.com/lyft/cartography/blob/8d60311a10156cd8aa16de7e1fe3e109cc3eca0f/cartography/intel/gcp/compute.py#L15

ramonpetgrave64 (commented):
Some changes I would like, but nothing to block merging.

cartography/intel/azure/sql.py
kedarghule: @ramonpetgrave64  We wanted to keep the code consistent throughout all Azure modules and since we used `list(map(...))` in the compute.py code. What I'll do is I'll open a PR in the future and change all the `list(map(...))` codes to list comprehension across all Azure modules we'll add. :)

cartography/intel/azure/sql.py
kedarghule: We wanted to keep the code consistent throughout all Azure modules and since we used list(map(...)) in the compute.py code. What I'll do is I'll open a PR in the future and change all the list(map(...)) codes to list comprehension across all Azure modules we'll add. :)

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/intel/azure/sql.py
kedarghule: Same as above

cartography/data/jobs/cleanup/azure_sql_server_cleanup.json
kedarghule: @ramonpetgrave64 I used the cleanup of the AWS codes and other codes as reference over here and they follow the same approach of specifying the entire resource hierarchy. Also, I used the same approach for the AWS KMS and API Gateway modules we added recently. So, for consistency, I used the same approach in the Azure cleanup codes. Didn't wanna be the cause of any weird errors.

cartography/intel/azure/sql.py
kedarghule: Sure! 👍 I'll commit this change and ping you once it is done. I could get specific on ResourceNotFoundError and ClientAuthenticationError under HttpResponseError.

kedarghule: @ramonpetgrave64 I have made the changes in the recent commit :D

ramonpetgrave64: Alright. In a future CR, let's see if we can implement this exception handling into a single function, so we don't repeat code.

kedarghule: Alright! I thought of doing that but making it as a single function would have generalized the exception handling. Like if we get a ResourceNotFoundError, the function would just print "Azure resource not found" which would hardly be helpful as we wouldn't know which resource was not found and for what resource the exception was raised. This approach kinda personalizes the error message that will be displayed.

kedarghule:
Completed all the changes. Updated the branch @ramonpetgrave64 :)

cartography/data/jobs/cleanup/azure_sql_server_cleanup.json
achantavy: Yeah, keep the whole resource hierarchy - we don't want to have a sync delete resources from a subscription that is not the current subscription.

cartography/intel/azure/sql.py
```diff
@@ -0,0 +1,930 @@
+import logging
```
achantavy, line 1: Won't block here, but we should split this file into multiple ones in a `sql/` folder.

Very much appreciate the work but or future PR, let's try to make them smaller if possible :)

kedarghule: Alright @achantavy ! How about I make those changes in a future PR? I'll add it to my to-do list :)

Also, updated the branch!

achantavy (approved):
🎉

PR #564, opened 2021-03-16
Added support for Azure Storage

kedarghule:
This PR provides initial support for Azure Storage resources. It populates resources like Azure Storage Account, Queue Service and Queues, Table Service and Tables, File Service and File Shares, and Blob Service and Blob Containers in the graph DB.


cartography/intel/azure/storage.py
```diff
@@ -0,0 +1,743 @@
...
+    """
+    Gets the list of table services.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        table_service_list = client.table_services.list(
+            storage_account['resourceGroup'], storage_account['name'],
+        ).as_dict()['value']
+
+    except HttpResponseError as e:
```
ramonpetgrave64, line 143: Just like in #563 , lets catch some specific errors.

cartography/data/jobs/cleanup/azure_storage_account_cleanup.json
```diff
@@ -0,0 +1,95 @@
...
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (:AzureStorageAccount)<-[r:RESOURCE]-(:AzureSubscription{id: {AZURE_SUBSCRIPTION_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:AzureStorageQueueService)<-[:USES]-(:AzureStorageAccount)<-[:RESOURCE]-(:AzureSubscription{id: {AZURE_SUBSCRIPTION_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
ramonpetgrave64, line 14: I understand this kind of MATCH was done before, but we should be sure it is even necessary, else it's misleading to future contributors.
`<-[:RESOURCE]-(:AzureSubscription{id: {AZURE_SUBSCRIPTION_ID}}) `

ramonpetgrave64 (commented):
Non blocking changes requested.

cartography/data/jobs/cleanup/azure_storage_account_cleanup.json
kedarghule: @ramonpetgrave64 This condition helps in multi-subscription setup. When a user has multiple subscriptions, this condition would make sure that resources from other subscription of the same user are not picked up while running query. with the subscription id condition, the scope is restricted to resources of one specific subscription vs the entire tenant. helps in a multi-tenant setup

cartography/intel/azure/storage.py
kedarghule: @ramonpetgrave64 Done! :)

kedarghule:
@ramonpetgrave64 Completed the requested changes and updated code in the branch! :)

cartography/data/jobs/cleanup/azure_storage_account_cleanup.json
ramonpetgrave64: Alright. Thanks for the explanation.

cartography/intel/azure/storage.py
```diff
@@ -0,0 +1,800 @@
...
+        storage_account_list = list(map(lambda x: x.as_dict(), client.storage_accounts.list()))
+
+    # ClientAuthenticationError and ResourceNotFoundError are subclasses under HttpResponseError
+    except ClientAuthenticationError as e:
+        logger.warning(f"Client Authentication Error while retrieving storage accounts - {e}")
+        return []
+    except ResourceNotFoundError as e:
+        logger.warning(f"Storage Account not found error - {e}")
+        return []
+    except HttpResponseError as e:
```
ramonpetgrave64, line 46: I feel we should raise here. If you get the `ClientAuthenticationError` and `ResourceNotFoundError` exceptions upon not having proper permissions, then that's fine, we should continue the sync, otherwise, the other failure cases could indicate an incomplete sync.

ramonpetgrave64 (approved):
One comment on the Exceptions, but it looks good.

PR #566, opened 2021-03-16
Fix #561: Add publish-to-PyPI GitHub Actions workflow

### LINKED ISSUE #561: 0.32.0 release not pushed to pypi
The 0.32.0 release is missing from [pypi](https://pypi.org/project/cartography/#history)

Could be related to: #460

achantavy:
Intent: When I hit 'Draft Release' in the GitHub UI and complete the release, a cartography release should be built and sent to PyPI.

As noted in #561, we lost this functionality when we moved from TravisCI to GitHub Actions and now need to implement this.

Relevant links
--------------
- https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
- https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release
- The old TravisCI flow is here: https://github.com/lyft/cartography/blob/0c9a662672cf4925ac8214b44451b1c50374aa97/.travis.yml#L23-L31

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -14,19 +14,19 @@ jobs:
         with:
           python-version: 3.6
       - name: Install pypa/build
-          run: >-
-            python -m
-            pip install
-            build
-            --user
-            - name: Build a binary wheel and a source tarball
-              run: >-
-                python -m
-                build
-                --sdist
-                --wheel
-                --outdir dist/
-                .
+        run: >-
+          python -m
+          pip install
+          build
+          --user
+      - name: Build a binary wheel and a source tarball
+        run: >-
+          python -m
+          build
+          --sdist
+          --wheel
+          --outdir dist/
+          .
       - name: Publish distribution 📦 to PyPI
         if: startsWith(github.ref, 'refs/tags')
         uses: pypa/gh-action-pypi-publish@master
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -8,7 +8,7 @@ jobs:
     name: Build and publish Python 🐍 distributions 📦 to PyPI
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
       - name: Set up Python 3.6
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: >-
+          python -m
+          build
+          --sdist
+          --wheel
+          --outdir dist/
+          .
+      - name: Publish distribution 📦 to PyPI
+        if: startsWith(github.ref, 'refs/tags')
+        uses: pypa/gh-action-pypi-publish@master
```
khanhldt, line 32: Do we want to follow the example 

```
if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
Or this is intentional?

Also, any reason to use `pypa/gh-action-pypi-publish@master` instead of a particular release tag?

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  release:
+    types: [published]
+jobs:
+  build-and-publish-to-pypi:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python 3.6
```
khanhldt, line 12: Similarly here, do you want to use `actions/checkout@v2` instead?

.github/workflows/publish-to-pypi.yml
achantavy: Short answer: I have no idea what I am doing and am copying https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

I don't want anything to be triggered on a push. I only want to publish to PyPI after I finish manually creating a release.

.github/workflows/publish-to-pypi.yml
achantavy: No idea, should I? :-)

khanhldt: It's just a matter of compatibility when it comes to use a fixed major version instead of main branch. I would suggest to use a fixed major version if possible.

.github/workflows/publish-to-pypi.yml
khanhldt: ah yeah, I see. That's good then. Haven't seen use of other events beside `pull_request` and `push`, but it looks right.

khanhldt (approved):

PR #569, opened 2021-03-17
Add increment util method to increment counter metrics in statsd

pchheda-lyft:
Adding this util method will allow us to support counter metrics publishing to statsd.

### DIFF
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -68,17 +68,17 @@ def timeit(method):
     return timed
 
 
-def increment(stat, count=1, rate=1):
+def increment_counter_stats(counter_name: str, increment_by: int = 1, rate: float = 1):
     """
     This method uses statsd to increment a counter.
     This is only active if config.statsd_enabled is True.
-    :param stat: the name of the counter (string) to increment
-    :param count: the amount (integer) to increment by. May be negative
+    :param counter_name: the name of the counter metric (string) to increment
+    :param increment_by: the amount (integer) to increment by. May be negative
     :param rate: a sample rate, a float between 0 and 1. Will only send data this percentage of the time.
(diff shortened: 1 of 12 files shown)


achantavy (dismissed):
Thanks for doing this!

cartography/util.py
```diff
@@ -68,6 +68,19 @@ def timed(*args, **kwargs):
     return timed
 
 
+def increment(stat, count=1, rate=1):
```
achantavy, line 71: Can you add typehints here and maybe rename this to something clearer - what is actually being incremented here?

pchheda-lyft: Sure. Let me do that!

pchheda-lyft: Refactored

achantavy (approved):

gkleiman:
Have we considered wrapping this in a class (e.g., `StatsClient`) that takes a scope/prefix?

You could have something like:

```
class ScopedStatsClient:
    def __init__(self, parent, scope = None):
         self.prefix = scope
         self.parent = parent

    def scopedClient(self, scope):
         return ScopedStatsClient(self, scope)

    def incr(self, name, count=1, rate=1):
         self.parent.incr(f'{self.prefix}.name, count, rate)
```

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
I like Gaston's suggestion. @achantavy any objections?

achantavy:
I don't have a strong opinion either way. The scenario is to have counter metrics, and having it scoped or not scoped isn't really a blocker, and I think it wouldn't be _that_ bad to refactor later on for scoping. Up to you, @pchheda-lyft

PR #571, opened 2021-03-18
Add scoped stats client

pchheda-lyft:
Summary of changes:
1. Added scoped stats client which will allow us to nest metric names. If statsd is not enabled, these methods will do nothing
2. Added incr method which lets us increment counter metrics
3. Changed timeit to use timer method of scoped stats client. The method stays as is.


achantavy (approved):
I like how you left the existing behavior intact and made the `is_enabled()` behavior cleaner. Thanks! Will let others review too.

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg (approved):

Issue #166, opened 2019-09-26
Determine strategy for consistent retries and failure handling


achantavy:
We need to figure out a way to consistently handle retries and failures when getting data.

We should consider having a way to manage this from one place.

PR #573, opened 2021-03-19
Added Support For Azure CosmosDB

kedarghule:
This PR adds initial support for Azure CosmosDB. Following resources are populated - 

- Azure Database Accounts
- Azure CosmosDB Locations
- Azure Database Accounts Cors Policies
- Azure Database Accounts Failover Policies
- Private Endpoint Connections
- Virtual Network Rules
- SQL Databases
- SQL Containers
- Cassandra Keyspaces
- Cassandra Tables
- MongoDB Databases
- MongoDB Collections
- CosmosDB Table Resources


cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1079 @@
...
+    """
+    for database_account in database_account_list:
+        # cleanup existing cors policy properties
+        run_cleanup_job(
+            'azure_cosmosdb_cors_details.json',
+            neo4j_session,
+            {'UPDATE_TAG': azure_update_tag, 'AZURE_SUBSCRIPTION_ID': subscription_id},
+        )
+
+        if 'cors' in database_account and len(database_account['cors']) > 0:
```
achantavy, line 133: Having all the data checks in this function makes it more complex and difficult to read. Can you move this logic to a `transform()` function so that the `load()` can work on processed objects? You might have seen this already but I'll link it here for others too: https://github.com/lyft/cartography/blob/master/docs/dev/writing-intel-modules.md#transform

achantavy (changes requested):
Hey @kedarghule, really appreciate the contribution. I gave it a quick first pass, but would you be able to please submit smaller chunks to make it easier to review? Perhaps with one data type at a time.

If this is not possible, completely understand. My main comment is to try to make the logic of each get + transform + load function as simple as possible. The `get` function should be a thin wrapper around the API function, and the `transform` function should massage the data into an easily consumable format (usually a list of dicts) for the `load` function to simply `unwind` and `merge` it to the graph.

Ping me on Slack if anything doesn't make sense!

tests/data/azure/cosmosdb.py
```diff
@@ -0,0 +1,474 @@
+cors1_id = "0001"
```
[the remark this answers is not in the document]
kedarghule: Done!

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1079 @@
...
+def load_database_account_data_resources(
+        neo4j_session: neo4j.Session, subscription_id: str, database_account_list: List[Dict], azure_update_tag: int,
+) -> None:
+    """
+    Loading all the resources (like cors policy, failover policy, private endpoint connections, virtual
+    network rules and locations) that we get as a part of the database account data itself.
+    """
+    for database_account in database_account_list:
+        # cleanup existing cors policy properties
+        run_cleanup_job(
```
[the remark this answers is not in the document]
kedarghule: I took a look at this and yeah, this should be removed. My bad. I removed it.

cartography/intel/azure/cosmosdb.py
kedarghule: Alright! I have done so wherever applicable in the code :)

kedarghule: This particular piece of code doesn't really transform anything. So I renamed it to `sync_database_account_data_resources()` since it just checks if the resources are present in the database account response (which we already have) and then runs their respective load functions if the resources are present.

That being said, I have added two transform functions in the code where the manipulation before neo4j ingestion was happening.

kedarghule:
@achantavy  I have made the necessary changes. Lemme know if there is anything more!

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+) -> None:
+    """
+    Create dictionaries for SQL Databases, Cassandra Keyspaces, MongoDB Databases and table resources.
+    """
+    sql_databases: List[Dict] = []
+    cassandra_keyspaces: List[Dict] = []
+    mongodb_databases: List[Dict] = []
+    table_resources: List[Dict] = []
+
+    for account_id, name, resourceGroup, sql_database, cassandra_keyspace, mongodb_database, table in details:
```
achantavy, line 585: Same comment here about decoupling the data transforms from the load functions.

I also  think it's more maintainable to split these by resource type: one function for sql dbs, one for cassandra, one for mongo, one for tables.

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+            for loc in database_account['write_locations']:
+                locations.append(loc)
+        if 'read_locations' in database_account and len(database_account['read_locations']) > 0:
+            for loc in database_account['read_locations']:
+                locations.append(loc)
+        if 'locations' in database_account and len(database_account['locations']) > 0:
+            for loc in database_account['locations']:
+                locations.append(loc)
+        # Selecting only the unique location entries
+        loc = [i for n, i in enumerate(locations) if i not in locations[n + 1:]]
```
achantavy, line 158: How does this select only uniques? Could we use a set instead?

cartography/intel/azure/cosmosdb.py
achantavy: This function does too many things at once. It does a transform+load for each type of policies/rules, and it does a transform+load for each type of location.

I'd recommend separating this for each resource.

I'll clarify this in a follow-up comment if needed; hold off on editing if you're  confused. It's getting late here but I wanted to make sure I gave you a somewhat timely second pass. :)

achantavy (changes requested):

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+
+@timeit
+def get_mongodb_databases(credentials: Credentials, subscription_id: str, database_account: Dict) -> List[Dict]:
+    """
+    Return the list of MongoDB Databases in a database account.
+    """
+    try:
+        client = get_client(credentials, subscription_id)
+        mongodb_database_list = list(
+            map(
```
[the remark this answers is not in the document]
kedarghule: Yes @achantavy, you're right. It returns a list of objects and I am converting them to a list of dicts.

docs/schema/azure.md
```diff
@@ -845,3 +871,391 @@ Representation of an [AzureStorageBlobContainer](https://docs.microsoft.com/en-u
         ```
         (AzureStorageBlobService)-[CONTAINS]->(AzureStorageBlobContainer)
         ```
+
+## AzureDatabaseAccount
+
+Representation of an [AzureDatabaseAccount](https://docs.microsoft.com/en-us/rest/api/cosmos-db-resource-provider/2020-04-01/databaseaccounts).
```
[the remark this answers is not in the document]
kedarghule: Changed the name of the resource.

@achantavy  Seems like Microsoft updated the documentation which removed those links. I can add a link but it seems like whenever they update their documentation, the link will break. Any ideas as to what we can do about this?

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+                lambda x: x.as_dict(),
+                client.table_resources.list_tables(
+                    database_account['resourceGroup'],
+                    database_account['name'],
+                ),
+            ),
+        )
+
+    except ClientAuthenticationError as e:
+        logger.warning(f"Client Authentication Error while retrieving table resources - {e}")
```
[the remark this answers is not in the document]
kedarghule: Aye aye Captain. Updated it throughout the code!

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+    """
+    Iterate over the database accounts and return the list of SQL and MongoDB databases, Cassandra keyspaces and
+    table resources associated with each database account.
+    """
+    for database_account in database_account_list:
+        sql_databases = get_sql_databases(credentials, subscription_id, database_account)
+        cassandra_keyspaces = get_cassandra_keyspaces(credentials, subscription_id, database_account)
+        mongodb_databases = get_mongodb_databases(credentials, subscription_id, database_account)
+        table_resources = get_table_resources(credentials, subscription_id, database_account)
+        yield database_account['id'], database_account['name'], database_account[
```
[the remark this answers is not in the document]
kedarghule: @achantavy I saw the codes for AWS and figured the convention was to use yield in a function like this where we get the secondary resources stemming from our primary resource. I followed that for SQL and Storage so I followed the same for this code too.

cartography/intel/azure/cosmosdb.py
kedarghule: @achantavy The response structure of the following resources - Locations, Cors Policies, Failover Policies, Private Endpoint Connections and Virtual Network Rules; comes inside the response structure of Database Accounts in the `get_database_account_list` function. 

The purpose of this function is to check if the Database Account response has the responses for the above mentioned resources and then call their respective load functions (if the resources are present). I'm not really transforming anything in the response structure here. Just a simple - "check if resource present -if yes, call load function; if no, move on".

I'll hold off on editing this one for now as I'm not fully understanding what you're trying to convey 🙈

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+
+    except ClientAuthenticationError as e:
+        logger.warning(f"Client Authentication Error while retrieving Cassandra keyspaces - {e}")
+        return []
+    except ResourceNotFoundError as e:
+        logger.warning(f"Cassandra keyspaces resource not found error - {e}")
+        return []
+    except HttpResponseError as e:
+        logger.warning(f"Error while retrieving Cassandra keyspaces list - {e}")
+        return []
```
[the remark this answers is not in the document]
kedarghule: cc: @mpurusottamc

cartography/intel/azure/cosmosdb.py
kedarghule: @achantavy - So `locations` is a list of dictionaries with each item having the following sample structure - 
{
                "id": "DA1-eastus",
                "location_name": "East US",
                "document_endpoint": "https://DA1-eastus.documents.azure.com:443/",
                "provisioning_state": "Succeeded",
                "failover_priority": 0,
}
And the plan is to return such unique dicts from the list of locations. Since `locations` is a list of dicts, set() can't be used since it is unhashable. I tried it and got a TypeError and saw that explanation on [stackoverflow](https://stackoverflow.com/questions/27435798/unhashable-type-dict-type-error).

So I used the method #2 on this link - https://www.geeksforgeeks.org/python-removing-duplicate-dicts-in-list/ and it worked like a peach! Tested this code and it works :)

cartography/intel/azure/cosmosdb.py
kedarghule: @achantavy - I added a transform function for the manipulations happening here. My bad I forgot to do that.

About the splitting up bit, firstly, it would just repeat a lot of code. I have split it up later for the individual resources under SQL databases, Cassandra keyspaces and Mongo. Second, these are the direct resources under Database Accounts  (our primary resource) so I figured I'll keep them in this function  as similar conventions were followed in the AWS codes. Lastly, the `details` gets populated in `get_database_account_details()` (line 436). If I have to split up the code for each resource and I'll have to do some major overhauls which I feel might repeat the code.

kedarghule:
@achantavy Made necessary changes and left replies for you :) Lemme know of any more changes.

cartography/intel/azure/cosmosdb.py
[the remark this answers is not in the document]
mpurusottamc: @achantavy Yes, this makes a lot of sense to handle transient HTTP errors. I will open an issue to keep track and work on all the get functions for both Azure and AWS.

One idea is to return multiple values from get functions. First value would be error code and second value would be the actual list. If there's an error, we ignore the load and cleanup steps. Let me know what you think about this.
cc: @kedarghule

cartography/intel/azure/cosmosdb.py
ramonpetgrave64: Are all the items unique by their `"id"`? Here's a nicer way you could get the unique items, borrowing from the third example in your link.
```
>>> test_list = [{'id' : i} for i in range(3)] * 2
>>> test_list
[{'id': 0}, {'id': 1}, {'id': 2}, {'id': 0}, {'id': 1}, {'id': 2}]
>>> unique_lists = {x['id'] : x for x in test_list}.values()
>>> unique_lists
dict_values([{'id': 0}, {'id': 1}, {'id': 2}])
>>> test_list = [{'id' : i} for i in range(3)]*2
>>> test_list
[{'id': 0}, {'id': 1}, {'id': 2}, {'id': 0}, {'id': 1}, {'id': 2}]
>>> uniques = {x['id'] : x for x in test_list}.values()
>>> uniques
dict_values([{'id': 0}, {'id': 1}, {'id': 2}])
>>> list(uniques)
[{'id': 0}, {'id': 1}, {'id': 2}]

```

kedarghule: The `id` field is not unique. Hence I stuck to the method I used.

docs/schema/azure.md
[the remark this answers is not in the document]
ramonpetgrave64: We can't have broken links. It's strange how this month April, working link is now says `2021-03-15`.
https://docs.microsoft.com/en-us/rest/api/cosmos-db-resource-provider/2021-03-15/databaseaccounts

We can at least link to the parent page, for now. Please add the links for all the other schema entries.
https://docs.microsoft.com/en-us/rest/api/cosmos-db-resource-provider/

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+                database_account['private_endpoint_connections'],
+        ) > 0:
+            _load_cosmosdb_private_endpoint_connections(neo4j_session, database_account, azure_update_tag)
+        if 'virtual_network_rules' in database_account and len(database_account['virtual_network_rules']) > 0:
+            _load_cosmosdb_virtual_network_rules(neo4j_session, database_account, azure_update_tag)
+
+        locations = []
+        # Extracting every location
+        if 'write_locations' in database_account and len(database_account['write_locations']) > 0:
+            for loc in database_account['write_locations']:
```
ramonpetgrave64, line 150: better to use `locations.extend(...)`

cartography/intel/azure/cosmosdb.py
ramonpetgrave64: Can you provide an example?

cartography/intel/azure/cosmosdb.py
```diff
@@ -0,0 +1,1090 @@
...
+                locations.append(loc)
+        if 'read_locations' in database_account and len(database_account['read_locations']) > 0:
+            for loc in database_account['read_locations']:
+                locations.append(loc)
+        if 'locations' in database_account and len(database_account['locations']) > 0:
+            for loc in database_account['locations']:
+                locations.append(loc)
+        # Selecting only the unique location entries
+        loc = [i for n, i in enumerate(locations) if i not in locations[n + 1:]]
+        if len(loc) > 0:
```
ramonpetgrave64, line 160: is this `if` necessary? Can you remove it and only have the loop?

cartography/intel/azure/cosmosdb.py
ramonpetgrave64: I agree with @achantavy. You should move the `if` blocks to their respective `_load_...` functions: If the list of things to load is empty, then don't run the query. Also if, for example, `database_account['failover_policies'])` exists, is it ever non-empty? If not, then you should do something like `if database_account.get('failover_policies'):`. 

And while you aren't really transforming the policies/rules, you certainly are doing a transformation against the `locations`.

ramonpetgrave64 (changes requested):
Please address the comments about the `_load_...` discussion.

cartography/intel/azure/cosmosdb.py
kedarghule: Something like this - You may have this in 'read locations'
{
"id": "DA1-eastus",
"location_name": "East US",
"document_endpoint": "https://DA1-eastus.documents.azure.com:443/",
"provisioning_state": "Succeeded",
**"failover_priority": 0,**
}

And you may get this in 'write locations' -
{
"id": "DA1-eastus",
"location_name": "East US",
"document_endpoint": "https://DA1-eastus.documents.azure.com:443/",
"provisioning_state": "Succeeded",
**"failover_priority": 10,**
}

In this case, they need to be treated as two different "locations" because of the different values in `failover_priority`.

ramonpetgrave64: I think I understand now. And later on in line 161-170 you do another filtering of this list. I recommend creating separate lists for each of the types "read", "write", and "associated".

For example
```
write_locations = database_account.get('write_locations', [])
```

And then I think for your `_load_..._locations()` you might also use `UNWIND`.

cartography/intel/azure/cosmosdb.py
kedarghule: I moved the the `if` blocks to the respective `_load_` functions. And made some other major updates in this function by breaking it down into parts. Let me know if it is okay now.

docs/schema/azure.md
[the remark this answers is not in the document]
kedarghule: Done!

cartography/intel/azure/cosmosdb.py
kedarghule: Done!

cartography/intel/azure/cosmosdb.py
kedarghule: Oh yeah. Removed it.

kedarghule:
> Please address the comments about the `_load_...` discussion.

Made the necessary changes. Lemme know if there are any more changes.

docs/schema/azure.md
```diff
@@ -845,3 +871,391 @@ Representation of an [AzureStorageBlobContainer](https://docs.microsoft.com/en-u
...
+        ```
+        (AzureCosmosDBAccount)-[CONTAINS]->(AzureCosmosDBMongoDBDatabase)
+        ```
+- Azure Database Account contains one or more table resource.
+
+        ```
+        (AzureCosmosDBAccount)-[CONTAINS]->(AzureCosmosDBTableResource)
+        ```
+
+## AzureCosmosDBLocation
```
achantavy, line 965: Can you please add a link to the Azure API resource that this is based off of if possible? It's helpful to trace back.

docs/schema/azure.md
```diff
@@ -845,3 +871,391 @@ Representation of an [AzureStorageBlobContainer](https://docs.microsoft.com/en-u
...
+|provisioningstate | The status of the Cosmos DB account at the time the operation was called.|
+|failoverpriority | The failover priority of the region.|
+|iszoneredundant | Flag to indicate whether or not this region is an AvailabilityZone region.|
+
+### Relationships
+
+- Azure Database Account has write permissions from, read permissions from and is associated with Azure CosmosDB Locations.
+
+        ```
+        (AzureCosmosDBAccount)-[WRITE_PERMISSIONS_FROM]->(AzureCosmosDBLocation)
```
achantavy, line 985: I'm a little confused about the wording of this since I am not familiar with Cosmos. What does it mean for a CosmosDBAccount to have read permissions from a CosmosDBLocation versus having it be associated with a CosmosDBLocation?

docs/schema/azure.md
kedarghule: Okay! I'll add the generic link that @ramonpetgrave64 suggested since Microsoft keeps updating their documentation and the link might break in the future.

docs/schema/azure.md
kedarghule: `read_locations` : An array that contains of the read locations enabled for the Cosmos DB account.

`write_locations`: An array that contains the write location for the Cosmos DB account.

`locations`: This is the one which creates the "associated with" relationship. An array that contains all of the locations enabled for the Cosmos DB account. Also, specifies a region in which the Azure Cosmos DB database account is deployed.

kedarghule:
@achantavy Made the changes. Let me know if there are any more :)

cartography/intel/azure/cosmosdb.py
kedarghule: Done!

docs/schema/azure.md
ramonpetgrave64: Maybe something like `CAN_READ` and `CAN_WRITE`?

kedarghule: Database accounts can't read a location. They will have read permission from a location. Likewise for write permissions.

ramonpetgrave64: What do you mean by "They will have read permission from a location"? For a particular database and location, do you mean the accounts will have permissions to read the data when accessing the database through that location?

How about `CAN_READ_FROM`?

kedarghule: Yes, you are right. Okay! I'll change it to CAN_READ_FROM and CAN_WRITE_TO.

kedarghule:
@achantavy and @ramonpetgrave64 All changes made :) Let me know if there are any more! :)

cartography/intel/azure/cosmosdb.py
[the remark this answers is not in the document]
achantavy: @mpurusottamc Let's continue the conversation on this in #166.

cartography/intel/azure/cosmosdb.py
achantavy: Fair points on copying prior patterns. Disclaimer: there are lots of prior code in the AWS modules that aren't exactly shining examples of great software :-)

In general, I think it's less bad to repeat code than it is to tightly couple and generalize functions too early ("write code that is easy to delete" and all that).

All that said, we're working to improve things by standardizing on the load steps in the near future: https://docs.google.com/document/d/1IZ12R3oROn11LcYj5XunokyOjJkKu-H2O1TEk065Dsk/edit#

Will not block for now.

achantavy (approved):
Thanks again for the contribution!

cartography/intel/azure/cosmosdb.py
kedarghule: Alright. I'll def check out this document :) Thanks for letting me know :)

kedarghule:
@achantavy Updated the latest in this branch. You can merge into master :)

kedarghule:
> Please address the comments about the `_load_...` discussion.

Done

ramonpetgrave64 (approved):
Thanks for your patience!

PR #579, opened 2021-03-22
Remove timer import

pchheda-lyft:
The timer import is wrong. I'm having trouble finding the right import. However, removing this for now will fix the problem.


cartography/scoped_stats_client.py
```diff
@@ -43,7 +42,7 @@ def incr(self, stat: str, count: int = 1, rate: float = 1.0) -> None:
         if self.is_enabled():
             self._client.incr(self._scope_prefix + "." + stat, count, rate)
 
-    def timer(self, stat: str, rate: float = 1.0) -> Timer:
+    def timer(self, stat: str, rate: float = 1.0):
```
achantavy, line 45: I'm ok with removing the return typehint, but have you tried  solving this with `def timer(...) -> Optional[Timer]`?

pchheda-lyft: The optional isn't the problem.  

from statsd.client.timer import Timer -> This import doesn't work
from statsd.client import Timer -> this works, but shows up that Timer is a protected element

achantavy: Did this break anything?

pchheda-lyft: from statsd.client.timer import Timer  -> this was the existing import. It doesn't break anything yet (since infraintelgraph isn't using it, yet).

But I was trying to use it in infraintelgraph and the unit tests break.

achantavy (approved):

PR #584, opened 2021-03-25
Add module level get stats

pchheda-lyft:
Summary of changes:
1. Added a module level get_stats method (static, you don't need an object to get this). Essentially, this allows us to add these lines anyplace and get the base stats handler with prefix as required

> from cartography.stats import get_stats
> stat_handler = get_stats("a")
> 

2. This can be further scoped as follows:

> scoped_stat_handler = stat_handler.get_stats("b") (This will lead to prefix being a.b)
> scoped_stat_handler.incr("metric") (will publish counter stat "a.b.metric" )
> 

3. When the sync is run, the underlying _stats_client is initialized. All ScopedStatsClient have element _client that refer to this reference.


cartography/stats.py
```diff
@@ -53,3 +53,24 @@ def timer(self, stat: str, rate: float = 1.0):
...
+# Global _stats_client
+# Will be set when cartography.config.statsd_enabled is True
+_stats_client: StatsClient = None
+
+
+def set_stats_client(stats_client):
+    """
+    This is used to set the module level stats client configured to talk with a statsd host
+    """
+    global _stats_client
```
achantavy, line 67: I'm scared of globals. Is this absolutely necessary? Can we accomplish the scenario you have in mind without this?

pchheda-lyft: Isn't cartography.util.stats_client also global? The _stats_client is exactly what cartography.util.stats_client used to be (statsd client). The set_stats_client is just a setter to initialize the  _stats_client (will happen as part of the sync run)

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg: Any reason we can't just create a stats client object in the cli/main, and pass it to sync? Or at least have a singleton stats client class  that we provide a handle to a la `cartography.util.stats_client`?

For example, we should only need to do this once:
```
        cartography.util.stats_client = StatsClient(
            host=config.statsd_host,
            port=config.statsd_port
        )
```
Then after we can call that with get_stats_client(prefix).

achantavy: Payal and I talked about this a bit yesterday. Internally we have other stats code that follows this same pattern of using a global variable for the stats client, so I'm fine with doing the same thing here.

> Any reason we can't just create a stats client object in the cli/main, and pass it to sync?

Do you mean adding an object to the config object, or adding another parameter to a sync_*/start_* function? The latter is a much, much larger refactor.

cartography/stats.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, client: StatsClient, prefix: str = None):
         self._client = client
         self._scope_prefix = prefix
 
-    def get_scoped_stats_client(self, scope: str):
+    def get_stats(self, scope: str):
```
dansec-lyft, line 20: I think this name should be more specific to what is returned. "get_stats_client" is more descriptive of what it's doing. Please add a type hint for the return object.

cartography/stats.py
```diff
@@ -53,3 +53,24 @@ def timer(self, stat: str, rate: float = 1.0):
...
+
+def set_stats_client(stats_client):
+    """
+    This is used to set the module level stats client configured to talk with a statsd host
+    """
+    global _stats_client
+    _stats_client = stats_client
+
+
+def get_stats(prefix: str) -> ScopedStatsClient:
```
dansec-lyft, line 71: Similar to the above comment but probably more important for readability since this will get called as a standalone method. Making it more descriptive for what is being returned would be helpful "get_stats_client" or similar.

cartography/stats.py
FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg: I was thinking of attaching to the config object. I'm totally fine with the current approach if you are though. Better we get things working now in a minimal way, and then we figure out  the bigger problem!

cartography/stats.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
pchheda-lyft, line 1: Needed this to add type hint on line 22 
See https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class

cartography/stats.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, client: StatsClient, prefix: str = None):
         self._client = client
         self._scope_prefix = prefix
 
-    def get_scoped_stats_client(self, scope: str):
+    def get_stats_client(self, scope: str) -> 'ScopedStatsClient':
```
pchheda-lyft, line 20: Added the type hint as a string. See https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class

cartography/stats.py
pchheda-lyft: Updated. Added the type hint as a string as well (https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class)

cartography/stats.py
pchheda-lyft: Updated.

achantavy (approved):

PR #600, opened 2021-04-03
Added lastupdated field to GCP DNS zones and record sets

jg10:
I noticed that when GCP DNS zones or record sets were removed in GCP they were still showing up in Cartography. After a bit of digging it turns out the GCP DNS zones and record sets were missing the lastupdated field that gets referenced in the gcp_dns_cleanup.json script. This PR adds them and I tested and the cleanup script will then successfully remove the deleted GCP DNS zones and record sets as intended.


marco-lancini (approved):
Thanks!

PR #601, opened 2021-04-07
Modify statsclient so that incr works

pchheda-lyft:
Problem: 
Currently in an intelmodule we add two lines 
`from cartography.stats import get_stats_client`
`stat_handler = get_stats_client(__name__)`
When this is done, the base stats client points to None, and hence the stats_client pointed to by the stat_handler is also None
Later when the sync config initializes a new StatsClient, the base stats stats client now points to this object, however, the stats_client pointed to by the stat_handler is still None. 

Point to note: the timeit util function is always called after the sync config  initializes a new StatsClient, and it uses the base stats client, hence those metrics are being published as is.

However, incr doesn't work in the intelmodule.

Existing approach 

Part1 in the screenshot below:
Part 2 in the screenshot below shows why a slightly different approach doesn't work as well
Image from iOS (3)

New approach which works

1. Base static ScopedStatsClient object is created (underlying StatsClient is None, _root is self, prefix is None)
2. When the intelmodule calls stats.get_stats_client, the returned ScopedStatsClient has prefix "intelmodule", _root is object from 1, underlying StatsClient is None
3. When the sync config is applied, and statsd is enabled, the underlying StatsClient for object (_client) from 1 is set to an actual StatsClient object configured to collect metrics
4. When the intelmodule is calling incr on object from 2, the ScopedStatsClient uses _root._client (_root is object from 1 and _root._client was set in step 3), and calls the StatsClient.incr (StatsClient is set! woohoo so the counter metric is published)

I've tested the change in Staging with our internal project, and publishing counter metrics works as expected.

### DIFF
--- a/cartography/stats.py
+++ b/cartography/stats.py
@@ -1,9 +1,4 @@
-import logging
-
 from statsd import StatsClient
-
-
-logger = logging.getLogger(__name__)
 
 
 class _StatsClientWrapper:
@@ -61,7 +56,6 @@ class ScopedStatsClient:
         if self.is_enabled():
             if self._scope_prefix:
                 stat = f"{self._scope_prefix}.{stat}"
-            logger.info(f"Stat being incremented is {stat}")
             self._client.get_client().incr(stat, count, rate)
 
     def timer(self, stat: str, rate: float = 1.0):
--- a/cartography/stats.py
+++ b/cartography/stats.py
@@ -40,7 +40,9 @@ class ScopedStatsClient:
                              The statsd server will take the sample rate into account for counters
         """
         if self.is_enabled():
-            self._client.incr(self._scope_prefix + "." + stat, count, rate)
+            if self._scope_prefix:
+                stat = f"{self._scope_prefix}.{stat}"
+            self._client.incr(stat, count, rate)
 
     def timer(self, stat: str, rate: float = 1.0):
         """
@@ -51,6 +53,8 @@ class ScopedStatsClient:
                              The statsd server will take the sample rate into account for counters
         """
         if self.is_enabled():
+            if self._scope_prefix:
+                stat = f"{self._scope_prefix}.{stat}"
             return self._client.timer(stat, rate)
         return None
 
--- a/cartography/stats.py
+++ b/cartography/stats.py
@@ -6,6 +6,20 @@ from statsd import StatsClient
 logger = logging.getLogger(__name__)
 
 
+class _StatsClientWrapper:
+    def __init__(self, client: StatsClient):
+        self._client = client
+
+    def set_client(self, client: StatsClient) -> None:
+        self._client = client
+
+    def get_client(self) -> StatsClient:
+        return self._client
+
+    def is_enabled(self) -> bool:
+        return self._client is not None
+
+
 class ScopedStatsClient:
     """
     A Proxy object for an underlying statsd client.
@@ -18,7 +32,7 @@ class ScopedStatsClient:
     newer_client.incr('bad') # Metric name = a.subsystem.bad
     """
 
-    def __init__(self, client: StatsClient, prefix: str = None):
+    def __init__(self, client: _StatsClientWrapper, prefix: str = None):
         self._client = client
         self._scope_prefix = prefix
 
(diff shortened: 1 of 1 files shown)


achantavy (dismissed):
Thanks for working on this!

cartography/stats.py
```diff
@@ -40,10 +28,19 @@ def get_stats_client(self, scope: str) -> 'ScopedStatsClient':
...
+
+        scoped_stats_client = ScopedStatsClient(prefix)
+        scoped_stats_client._parent = self
+        return scoped_stats_client
+
+    def get_base_scoped_stats_client(self) -> 'ScopedStatsClient':
+        base_client = self
+        while base_client._parent:
+            base_client = base_client._parent
+        return base_client
```
gkleiman, line 40:
```suggestion

    def get_root_client() -> 'ScopedStatsClient':
        client = ScopedStatsClient()
        client._root = client

        return client
```

cartography/stats.py
```diff
@@ -40,10 +28,19 @@ def get_stats_client(self, scope: str) -> 'ScopedStatsClient':
...
 
     def is_enabled(self) -> bool:
-        return self._client.is_enabled()
+        return self.get_base_scoped_stats_client()._client is not None
```
gkleiman, line 43:
```suggestion
        return self._root._client is not None
```

cartography/stats.py
```diff
@@ -69,16 +66,16 @@ def timer(self, stat: str, rate: float = 1.0):
...
 
     def set_stats_client(self, stats_client: StatsClient) -> None:
-        self._client.set_client(stats_client)
+        self.get_base_scoped_stats_client()._client = stats_client
 
 
 # Global _scoped_stats_client
 # Will be set when cartography.config.statsd_enabled is True
-_scoped_stats_client: ScopedStatsClient = ScopedStatsClient(_StatsClientWrapper(None))
+_scoped_stats_client: ScopedStatsClient = ScopedStatsClient()
```
gkleiman, line 78:
```suggestion
_scoped_stats_client: ScopedStatsClient = ScopedStatsClient.get_root_client()
```

cartography/stats.py
```diff
@@ -69,16 +66,16 @@ def timer(self, stat: str, rate: float = 1.0):
         if self.is_enabled():
             if self._scope_prefix:
                 stat = f"{self._scope_prefix}.{stat}"
-            return self._client.get_client().timer(stat, rate)
+            return self.get_base_scoped_stats_client()._client.timer(stat, rate)
         return None
 
     def set_stats_client(self, stats_client: StatsClient) -> None:
-        self._client.set_client(stats_client)
+        self.get_base_scoped_stats_client()._client = stats_client
```
gkleiman, line 73:
```suggestion
        self._root._client = stats_client
```

cartography/stats.py
```diff
@@ -69,16 +66,16 @@ def timer(self, stat: str, rate: float = 1.0):
         if self.is_enabled():
             if self._scope_prefix:
                 stat = f"{self._scope_prefix}.{stat}"
-            return self._client.get_client().timer(stat, rate)
+            return self.get_base_scoped_stats_client()._client.timer(stat, rate)
```
gkleiman, line 69:
```suggestion
            return self._root._client.timer(stat, rate)
```

cartography/stats.py
```diff
@@ -56,7 +53,7 @@ def incr(self, stat: str, count: int = 1, rate: float = 1.0) -> None:
         if self.is_enabled():
             if self._scope_prefix:
                 stat = f"{self._scope_prefix}.{stat}"
-            self._client.get_client().incr(stat, count, rate)
+            self.get_base_scoped_stats_client()._client.incr(stat, count, rate)
```
gkleiman, line 56:
```suggestion
            self._root._client.incr(stat, count, rate)
```

gkleiman (commented):
What do you think about storing a reference to the root client instead of to the parent?

That way you can avoid having to traverse the whole hierarchy whenever you increment a stat. See the suggestions for how it could look like.

gkleiman (approved):

Issue #559, opened 2021-03-08
AWS apigateway sync - TooManyRequests exception


achantavy:
*Title*: *One line description*

*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

APIGateway sync crashes with TooManyRequests for `get_rest_api_details`.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run apigateway sync on account that is experiencing API quota issues.

*Logs*:
> If applicable, copy and paste your console log with the failing stack trace.

```
File "{PATH}/cartography/cartography/intel/aws/apigateway.py", line 359, in sync_apigateway_rest_apis
for api_id, stage, certificate, resource, policy in stages_certificate_resources:
botocore.errorfactory.TooManyRequestsException: An error occurred (TooManyRequestsException) when calling the GetResources operation: Too Many Requests
result = method(*args, **kwargs)
File "{PATH}/cartography/cartography/intel/aws/apigateway.py", line 296, in load_rest_api_details
aws_requested_syncs=aws_requested_syncs, # Could be replaced later with per-account requested syncs
for page in response_iterator:
sync_apigateway_rest_apis(neo4j_session, boto3_session, region, current_aws_account_id, update_tag)
File "{PATH}/cartography/cartography/util.py", line 61, in timed
File "/code/venvs/venv/lib/python3.6/site-packages/botocore/client.py", line 357, in _api_call
raise error_class(parsed_response, operation_name)
resources = get_rest_api_resources(api, client)
File "{PATH}/cartography/cartography/intel/aws/apigateway.py", line 46, in get_rest_api_details
result = method(*args, **kwargs)
```

*Please complete the following information:*:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]

047a6bd1820e846e70bda4c6ddc18027da39e20a

*Additional context*

- Going forward we should think of a generic way of handling these API limits. AWS EMR sync handled it with sleeps: https://github.com/lyft/cartography/blob/047a6bd1820e846e70bda4c6ddc18027da39e20a/cartography/intel/aws/emr.py#L29 - maybe we could apply this one-off here as well.

cpollard0:
[...]

PR #603, opened 2021-04-09
#559 - Update AWS docs on using environment vars for retry behavior

mpurusottamc:
Added AWS Config with retry logic to fetch API Gateway resources. We have tested it on our own service deployment.

https://github.com/lyft/cartography/issues/559

cc: @cpollard0


achantavy (changes requested):
Thanks for taking this on! Could we use or modify https://github.com/lyft/cartography/blob/1583dbe6e08ad5402012714db9b2a652506f2b76/cartography/intel/aws/ec2/util.py#L5 instead?

cpollard0:
Have we considered using the AWS_MAX_ATTEMPTS environment variable, having boto3 pick it up, and not setting it in code? End users can use the setting that's best for them, and we could update the readme to give some guidance on how to and best practices. It'll also affect all clients, so no need to set and maintain it in a ton of places.

achantavy:
@cpollard0 Hm, good point actually. I wasn't actually aware of that. We should update documentation instead on boto3 config information available: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables

mpurusottamc:
@achantavy @cpollard0 I like the idea to use environment variables instead of assigning a static value. I can update the documentation with this information.

I see another instance of similar code at https://github.com/lyft/cartography/blob/master/cartography/intel/aws/elasticsearch.py#L41
`https://github.com/lyft/cartography/blob/master/cartography/intel/aws/elasticsearch.py#L41`

We should standardize to have just one set of config in the AWS intel root, instead of inside EC2 folder. What do you guys think?

achantavy:
> I can update the documentation with this information.

Thank you!

> We should standardize to have just one set of config in the AWS intel root, instead of inside EC2 folder. What do you guys think?

The fully correct solution would be to encourage using the env vars instead  and eventually remove these functions completely (can come in a future PR)

mpurusottamc:
@achantavy i have updated the docs

achantavy (approved):
Thank you!

PR #606, opened 2021-04-14
Add basic docker-compose support for development

ryan-lane:
This change adds very basic Dockerfile and docker-compose.yml files
for running cartography against a local neo4j.

Usage:

```bash
docker build -t lyft/cartography
docker-compose run cartography ...
```

I've purposely not included an entrypoint, so that custom sync
scripts can be called through run.

Configuration is possible via the `.compose` directory, which is
git ignored. Environment for cartography is in
`.compose/cartography/cartography.env`. neo4j config, logs, etc is
located at `.compose/neo4j/...`

Note that on initial start of the compose stack, it's necessary to
change the neo4j user's password through the neo4j UI.


ryan-lane:
I wasn't fully sure where in the docs we should stick this, so I didn't update them yet. Let me know!

achantavy:
Thanks for working on this!

> I wasn't fully sure where in the docs we should stick this, so I didn't update them yet. Let me know!

Let's make a new doc in `docs/setup/docker.md` to describe how to use this, and then we can link to it in README.md by making a `##`-heading called "Docker" placed right after the `## Install and configure` heading.

ryan-lane:
Ok. Done!

docs/containers/testing-with-docker.md
```diff
@@ -0,0 +1,54 @@
...
+entrypoint, so it's necessary to pass in the command being run. This
+also makes it possible to run a custom sync script, rather than only
+cartography.
+
+### Example
+
+`./compose/cartography/cartography.env`:
+```
+# See the cartography github configuration intel module docs
+GITHUB_KEY=BASE64ENCODEDKEY
```
achantavy, line 46: I don't know a lot about docker config patterns/practices. Is it possible to avoid asking the developer to save creds in a file?

ryan-lane: Ah, yeah, I think that can be done. Let me test it and push in a change.

docs/containers/testing-with-docker.md
```diff
@@ -0,0 +1,51 @@
...
+
+Configuration is possible via the `.compose` directory, which is
+git ignored. Environment for cartography is in
+`.compose/cartography/cartography.env`. neo4j config, logs, etc is
+located at `.compose/neo4j/...`
+
+### Notes
+
+* On initial start of the compose stack, it's necessary to
+change the neo4j user's password through the neo4j UI.
```
achantavy, line 36: [optional/I haven't tested in this context] Since this is strictly for dev testing, we could do something like this on startup to set the password to something:

```bash
if ((echo >/dev/tcp/localhost/7687) &>/dev/null); then echo "found neo4j" && curl -H "Content-Type: application/json" -X POST -d '{"password":"testtesttest"}' -u neo4j:neo4j http://localhost:7474/user/neo4j/password 
```

achantavy (dismissed):
Thanks for working on this!

@marco-lancini, what do you think? The container basically only makes Neo4j available and the docs make it clear that this is only for testing.

docs/containers/testing-with-docker.md
```diff
@@ -0,0 +1,51 @@
...
+* Neither the docker image, nor the docker-compose file define an
+entrypoint, so it's necessary to pass in the command being run. This
+also makes it possible to run a custom sync script, rather than only
+cartography.
+
+### Example
+
+```bash
+# See the cartography github configuration intel module docs
+export GITHUB_KEY=BASE64ENCODEDKEY
```
achantavy, line 46: I thought about this a bit more and I'm a bit concerned about asking a user to put their creds in the terminal which would get recorded in bash history.

AWS official docs [don't bother](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_environment.html) to deal with this (I suppose they assume the dev knows this), but since this is a security tool I feel like we should do better.

How do you feel about adding a disclaimer instructing the user to turn off history logging temporarily when setting creds with `set +o history` and then turning it on back again with `set -o history` ([ref](https://attack.mitre.org/techniques/T1552/003/))?

ryan-lane: I added disabling/renabling command history to the example.

I think it's good we're being good security citizens here, but IAM creds and github access keys normally live in files in the user's home directory, but no need to have them live in other spots on the filesystem :D

docs/containers/testing-with-docker.md
ryan-lane: I'm not opposed to this, but I'd like to followup with improvements like that in other PRs, based on other folks' feedback.

achantavy:
Hey @kmcquade, hope you've been well :), been a while since we talked about using docker to test this tool. What do you think about this test setup?

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /srv/cartography
+
+ENV PATH=/venv/bin:$PATH
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY . /srv/cartography
+
```
marco-lancini, line 12: Not sure if there is an appetite to harden this, but does it need to be run as `root`? 
Can we drop to an un-privileged user?

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+        interval: 30s
+        timeout: 10s
+        retries: 5
+  cartography:
+    image: lyft/cartography
+    init: true
+    restart: on-failure
+    depends_on:
+      - neo4j
+    env_file: ./.compose/cartography/cartography.env
```
marco-lancini, line 31: If we want to have a config stored in a file, it might be worth providing a sample to show what the expected syntax is

docs/containers/testing-with-docker.md
```diff
@@ -0,0 +1,58 @@
...
+# Testing with docker
+
+## Using the included docker-compose support
+
+### Usage
+
+```bash
+docker build -t lyft/cartography
+docker-compose up -d
+docker-compose run cartography ...
```
marco-lancini, line 23: Isn't cartography already started with the command above?

docs/containers/testing-with-docker.md
```diff
@@ -0,0 +1,58 @@
...
+export GITHUB_KEY=BASE64ENCODEDKEY
+# You need to set this after starting neo4j once, and resetting
+# the default neo4j password, which is neo4j
+export NEO4j_PASSWORD=...
+# Reenable bash command history
+set -o history
+# Start cartography dependencies
+docker-compose up -d
+# Run cartography
+docker-compose run -e GITHUB_KEY -e NEO4j_PASSWORD cartography cartography --github-config-env-var GITHUB_KEY --neo4j-uri bolt://neo4j:7687 --neo4j-password-env-var NEO4j_PASSWORD --neo4j-user neo4j
```
marco-lancini, line 57: In this doc we use environment variables, whereas in the docker-compose file we expect an `env` file.

marco-lancini:
> Thanks for working on this!
> 
> @marco-lancini, what do you think? The container basically only makes Neo4j available and the docs make it clear that this is only for testing.

@achantavy I'm personally not a fan of storing credentials/tokens within the folder of a repo itself.
What I would probably do is mount the `.aws` folder as a volume and then reference the profile I need as environment variable

docs/containers/testing-with-docker.md
ryan-lane: No, it's `docker-compose run cartography cartography ...`

ryan-lane:
Changed docs to use passed in environment, changed the user it runs as, and updated the docs.

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+      - NEO4J_dbms_memory_pagecache_size=1G
+      - NEO4J_dbms.memory.heap.initial_size=1G
+      - NEO4J_dbms_memory_heap_max__size=1G
+    healthcheck:
+        test: ["CMD", "curl", "-f", "http://localhost:7474"]
+        interval: 30s
+        timeout: 10s
+        retries: 5
+  cartography:
+    image: lyft/cartography
```
marco-lancini, line 26: Are we saying `lyft/cartography` exists on DockerHub?

docs/containers/testing-with-docker.md
```diff
@@ -0,0 +1,62 @@
...
```
marco-lancini, line 61: From this example, it might not be clear how the cartography container would then be able to authenticate to any cloud provider.
I see there is a `TODO` in line `34` above, might be worth fleshing it out

achantavy: Adding to this a bit - to provide AWS support we could employ something like this: https://github.com/achantavy/boto3-docker

Let's try to establish a pattern on how other modules can provide creds to the cartography container.

docker-compose.yml
ryan-lane: No. See the build instructions, which say to do `docker build -t lyft/cartography`. It's common to prefix an imagename with the repo's org, and since the Dockerfile exists in the repo, and this is the convention with other lyft repos, I followed that convension.

docs/containers/testing-with-docker.md
ryan-lane: It seems to me that in the configuration docs that everything except AWS passes environment into cartography. Only AWS seems to use a file. So, these docs cover every intel module except AWS.

ryan-lane: I don't have a good way to test the AWS intel module right now, which is why I left it as a TODO. It would be ideal for someone with the ability to test that to try out the docker-compose support and add that in.

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg (approved):
This is great

achantavy (approved):
🚀

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
This will play fantastic with localstack where we can check how the system works with a fake local aws!

PR #607, opened 2021-04-15
initial route table support

krisek:
The intention is to raise this as a draft PR to be able to discuss the implementation, test cases and documentation will come once the implementation is settled.


cartography/intel/aws/ec2/route_tables.py
```diff
@@ -0,0 +1,520 @@
...
+    """
+
+    neo4j_session.run(
+        ingest_default_associations, route_tables=data, update_tag=update_tag,
+        region=region, aws_account_id=aws_account_id,
+    )
+
+
+@timeit
+def load_routes(
```
achantavy, line 153: Ideally, rather than have neo4j itself perform control flow, we could try to transform the data into simpler lists of dicts that can be merged in a single `unwind`. Is there a way to ensure all items in a list are the same type of gateway to make things simpler?

But yeah in general, let's try to have as few merges per query as possible because we've found that neo4j can throw unpredictable errors when we throw enough data at it and if queries aren't exactly the way it wants them; I write about this a bit in https://github.com/lyft/cartography/pull/631 (in progress; let me know what you think there too).

achantavy (changes requested):
Looks solid - thanks for the docs and tests. Can you also add the screenshots you posted on Slack to the PR itself? They're really helpful.

I also left a comment about Neo4j queries, let me know if anything I wrote doesn't make sense and feel free to tag others on this too.

cartography/intel/aws/__init__.py
```diff
@@ -126,7 +126,15 @@ def _sync_multiple_accounts(
     for profile_name, account_id in accounts.items():
         logger.info("Syncing AWS account with ID '%s' using configured profile '%s'.", account_id, profile_name)
         common_job_parameters["AWS_ID"] = account_id
-        boto3_session = boto3.Session(profile_name=profile_name)
+
+        boto3_session = None
```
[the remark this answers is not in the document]
krisek: Commit dropped. (Hope it worked.)

tests/integration/cartography/intel/aws/ec2/test_ec2_route_tables.py
```diff
@@ -0,0 +1,184 @@
...
+        ("rtb-0fa0eb0f6130f6a2e", "rtb-0fa0eb0f6130f6a2e|172.154.90.0/AQ24", "vgw-0afe079e97f2c77f8"),
+        ("rtb-0fa0eb0f6130f6a2e", "rtb-0fa0eb0f6130f6a2e|172.183.21.203/32", "vgw-0afe079e97f2c77f8"),
+        ("rtb-01fbd15d4b6e8e14f", "rtb-01fbd15d4b6e8e14f|0.0.0.0/0", "igw-007fe6f07b9419e44"),
+        ("rtb-0d49710751f04455d", "rtb-0d49710751f04455d|0.0.0.0/0", "igw-007fe6f07b9419e44"),
+    }
+
+    # Fetch relationships
+    result = neo4j_session.run(
+        """
+        MATCH (rtb:EC2RouteTable)-[:ROUTE]-(r)-[:ASSOCIATED]-(gw)
```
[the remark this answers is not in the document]
krisek: Updated in #20a8aef 
gw is tricky (in other words won't fix): I mirrored here the logic of the AWS EC2Route flexible structure where gw node can be of several types (pcx connection, virtual gateway, internet gateway, etc.)

cartography/intel/aws/ec2/route_tables.py
krisek: :thinking: I thought we optimize on not processing the data too much before loading to neo4j. This would mean to rewrite the most complex part of the PR.

krisek:
Example queries and graphs

In an architecture with connectivity to various partners over VPN gateways where proxies (behind ELBs) are inserted into the path that expose partner services to cloud resources. 

The first query finds the ELB & ports belonging to a specific VPN gateway:

```
MATCH (n:AWSVPNGateway)-[:ASSOCIATED]->(r:EC2Route)<-[:ROUTE]-(rtb:EC2RouteTable) WHERE n.id IN ['vgw-12345'] 
OPTIONAL MATCH (rtb)<-[:ASSOCIATED]-(snet:EC2Subnet) 
OPTIONAL MATCH (snet)<-[:PART_OF_SUBNET]-(i:EC2Instance)
OPTIONAL MATCH (snet)<-[]-(lb:LoadBalancer)
OPTIONAL MATCH (lb)-[]->(listen:ELBListener)
RETURN n,r,rtb,snet,i,lb,listen
```

image

The next one is to opposite direction where I look up the VPN belonging to a specific ELB.

```
MATCH (n:LoadBalancer)-[]-(snet:EC2Subnet)-[:ASSOCIATED]-(rtb:EC2RouteTable)-[:ROUTE]-(route)-[]-(agw) 
WHERE n.dnsname IN ['internal-.....'] 
RETURN n, snet, rtb, route, agw
```

image

So the main path is `(EC2Subnet)-[]-(EC2RouteTable)-[]-(EC2Route)` and then from the (EC2Route)-[]-(Gateway|Peering|etc) we point to some destination

An other very handy one: find all blackholes in the environment

```
MATCH (r:EC2Route{state:'blackhole'}) RETURN count(r);
```

To check if a specific IP address reachable from a specific subnet (in different VPCs, connected over peerings) (here the source subnet & VPC are specified with a specific AwsTag)

```
MATCH (n:EC2PrivateIp{private_ip_address:'10.123.123.123'})--(nif:NetworkInterface)--(snet:EC2Subnet)--(vpc:AWSVpc)
OPTIONAL MATCH (t:Tag{value:'mysubnet'})-[:TAGGED]-(vpc_k8:AWSVpc)-[:MEMBER_OF_AWS_VPC]-(snet_cni:EC2Subnet)-[:TAGGED]-(tag_cni:Tag{value:'aws-cni'})
OPTIONAL MATCH (vpc)--(peer:AWSPeeringConnection)--(vpc_k8)
OPTIONAL MATCH (acidr:AWSCidrBlock)--(peer)--(bcidr:AWSCidrBlock)
OPTIONAL MATCH (snet_cni)-[]-(rtb:EC2RouteTable)-[]-(route:EC2Route)-[]-(peer)
OPTIONAL MATCH (snet)-[]-(rtb_2:EC2RouteTable)-[]-(route_2:EC2Route)-[]-(peer)
OPTIONAL MATCH (route)-[]-(cidr_r:AWSIpv4CidrBlock)
RETURN n,nif,snet,vpc, t, vpc_k8, snet_cni, tag_cni, peer, acidr, bcidr, rtb, route,rtb_2, route_2,cidr_r
```

image

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:bionic
```
achantavy, line 1: (Adding a comment so this PR doesn't get auto-closed by stalebot)

Just noticed this dockerfile - could we include this in a separate PR since it's not directly related to the route table work?

cartography/intel/aws/ec2/route_tables.py
achantavy: @krisek - My main reason for being hesitant on the long query at first was because Neo4j would throw ServiceUnavailable exceptions that were difficult to repro and understand. Now that I have a [viable fix](https://github.com/lyft/cartography/issues/522#issuecomment-885402304) for it when it happens, I'm ok with the query form as long as we call `neo4j_session.write_transaction()` so that we can get automatic retries in case the database connection times out for any reason.

Here's an example:

```python
def _load_findings_tx(tx: neo4j.Transaction, findings: List[Dict], update_tag: int) -> None:
    query = """
        query goes here
    """
    logger.info(f"Loading {len(findings)} findings to the graph")
    tx.run(query, Findings=findings, UpdateTag=update_tag)


def load(session: neo4j.Session, findings: LIst[Dict]) -> None:
    session.write_transaction(_load_findings_tx, findings, update_tag)
```

As long as load functions call queries this way, I think we'll avoid the exception more effectively - I'll update the intel module guide at some point.

PR #609, opened 2021-04-20
Add GCP VPN Gateways and VPN Tunnels

lgomezma:
This PR intends to add support for GCP VPN Gateways (for now limited to Target VPN Gateways) and VPN Tunnels. Also linking those with GCP Forwading Rules.


lgomezma:
@achantavy finally managed to make the linter pass. It looks like it was just space alignment :P

cartography/data/jobs/cleanup/gcp_compute_vpn_gateways_cleanup.json
```diff
@@ -0,0 +1,17 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (n:GCPVpnGateway) WHERE n.project_id = {GCP_PROJECT_ID} AND n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
achantavy, line 4: Would it be a lot of work to draw a path from the GCPVpnGateway to the GCPProject here instead of querying for n.projectid?

cartography/intel/gcp/compute.py
```diff
@@ -694,6 +808,155 @@ def load_gcp_forwarding_rules(neo4j_session: neo4j.Session, fwd_rules: List[Dict
             _attach_fwd_rule_to_vpc(neo4j_session, fwd, gcp_update_tag)
 
 
+@timeit
+def load_gcp_vpn_gateways(neo4j_session: neo4j.Session, vpngw_list: Dict, gcp_update_tag: int) -> None:
```
achantavy, line 812: [non-blocker] What kind of perf are you seeing with this job? It could be worth refactoring to do an `unwind .. merge`, but I won't block on this since I'm working on something to make the writes more standardized - will be trying in the coming weeks.

cartography/data/jobs/cleanup/gcp_compute_vpn_gateways_cleanup.json
lgomezma: I wanted to do the PROJECT ID standard part of the query in the cleanup the same way we do with the AWS account. This is to avoid the issue we have where we cleanup all the nodes of projects that haven't being synched yet on this run. I could probably make the way back to the GCPProject but I find this way more straightforward and easy to understand what's going on. Thoughts?

cartography/intel/gcp/compute.py
lgomezma: I actually tried to do that but decided against it cause I found it hard to do it that way and then individually check for each element to call separately the `_attach_vpn_gateway_to_fwd_rules` method. Do you have an example where `UNWIND` is used along with an `_attach` method?

lgomezma: @achantavy thoughts on this?

cartography/data/jobs/cleanup/gcp_compute_vpn_gateways_cleanup.json
lgomezma: @achantavy thoughts on this?

achantavy: I think we should follow the pattern in AWS modules where we check for 

```
(:GCPVpnGateway)<-[r:RESOURCE]-(:GCPProject)
```

I claim full responsibility for messing that up with the first GCP modules - sorry!

But yeah, all assets should be linked to an account/project/whatever-it-is-called-by-that-data-provider.

cartography/intel/gcp/compute.py
achantavy: The way you have it is fine; we'll fix it more broadly later.

achantavy (changes requested):
Hey @lgomezma, let's just change up the cleanup jobs. Can you also provide a sample screenshot of what your graph looks like after this change?

Sorry again for the slow reply.

Issue #605, opened 2021-04-13
Error when syncing multiple accounts: Unhandled exception during sync stage 'aws'


jbar462:
*Title*: Error when syncing multiple accounts: Unhandled exception during sync stage 'aws'

*Description*:
When I try to sync the multiple accounts that I have setup, using 'AWS_CONFIG_FILE=/root/.aws/config cartography --neo4j-uri bolt://localhost:7687 --aws-sync-all-profiles' cartography works for a little bit then errors and stops running. Seems to crash at the same stage of the sync each time, when "INFO:cartography.intel.aws.apigateway:Syncing AWS APIGateway Rest APIs for region 'us-east-1' in account 'xxxxxxxxxx'." Will paste full log of error below. Also something that caught my eye in the error log was that the final line shows "json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)"

*To Reproduce*:
run AWS_CONFIG_FILE=/root/.aws/config cartography --neo4j-uri bolt://localhost:7687 --aws-sync-all-profiles

*Logs*:
INFO:cartography.intel.aws.apigateway:Syncing AWS APIGateway Rest APIs for region 'us-east-1' in account 'XXXXXXXXXX'.
ERROR:cartography.sync:Unhandled exception during sync stage 'aws'
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/cartography/sync.py", line 73, in run
    stage_func(neo4j_session, config)
  File "/usr/local/lib/python3.7/site-packages/cartography/util.py", line 68, in timed
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/cartography/intel/aws/__init__.py", line 201, in start_aws_ingestion
    requested_syncs,
  File "/usr/local/lib/python3.7/site-packages/cartography/intel/aws/__init__.py", line 139, in _sync_multiple_accounts
    aws_requested_syncs=aws_requested_syncs,  # Could be replaced later with per-account requested syncs
  File "/usr/local/lib/python3.7/site-packages/cartography/intel/aws/__init__.py", line 57, in _sync_one_account
    RESOURCE_FUNCTIONS[func_name](**sync_args)
  File "/usr/local/lib/python3.7/site-packages/cartography/util.py", line 68, in timed
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/cartography/intel/aws/apigateway.py", line 369, in sync
[...]

PR #610, opened 2021-04-22
fixed json parsing bug

ryohare:
Fixed Json parsing bug for API Gateways reported in #605.

Issue was the JSON for the API Gateways where doubly escaped json and therefore could not be parsed via `loads`. Escaped the policy document json. Also discovered a data format change which necessitated converting a `expand` to and `append` for policies.

### DIFF
--- a/cartography/intel/aws/apigateway.py
+++ b/cartography/intel/aws/apigateway.py
@@ -296,7 +296,7 @@ def load_rest_api_details(
     for api_id, stage, certificate, resource, policy in stages_certificate_resources:
         parsed_policy = parse_policy(api_id, policy)
         if parsed_policy is not None:
-            policies.extend(parsed_policy)
+            policies.append(parsed_policy)
         if len(stage) > 0:
             for s in stage:
                 s['apiId'] = api_id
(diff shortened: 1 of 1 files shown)


achantavy:
@ryohare Thanks so much for the fix!

I think the CLA check is currently failing because you need to have all commits be made with the GitHub name that you signed the CLA with.

achantavy:
Is it possible to unit test this condition? Sample data is available here: https://github.com/lyft/cartography/blob/master/tests/data/aws/apigateway.py

ryan-ohare-nydig:
Sure, I'll clean this up a bit. So many github accounts I forget what exactly to set `git config user.name` too.

Let me look at adding in a unit test for it. I wanted to get the PR in for the issue thread since someone else was struggling with the issue I patched out for my own purposes.

Look for an update in the next few days.

ryohare:
So many github accounts, I forget sometimes what to set `git config user.name` too.

Let me look at the unit tests. No issue in adding some.

Wanted to get this PR opened for the issue thread as someone else was stuck with this issue I patched out a few weeks ago locally for my purposes.

ryohare:
Closing this and moving to #611

PR #611, opened 2021-04-22
#605: Fix AWS APIGateway json parsing bug

ryohare:
Fixed Json parsing bug for API Gateways reported in #605.

Issue was the JSON for the API Gateways where doubly escaped json and therefore could not be parsed via loads. Escaped the policy document json. Also discovered a data format change which necessitated converting a expand to and append for policies.

Continuation of PR #610


achantavy (approved):
Thanks very much for the fix!

PR #617, opened 2021-05-12
AWS KMS policy - log and skip failures

achantavy:
If we fail to retrieve KMS key policies, log a warning and continue the sync.

The way the code was originally written with `policy = None` seems to  indicate that the intended behavior is to return this value, rather than `raise`.


achantavy:
@mpurusottamc - thoughts?

kunaals (approved):

PR #619, opened 2021-05-13
Add attribute to config class, for custom intel module configuration

ryan-lane:
This change adds an attribute to the Config class that can be used by custom intel modules.

Though it's possible to inject attributes into an existing object, mypy will (rightly) complain that the attribute doesn't exist in the class. To support this more directly, I've added a dict attribute that can be use for custom intel modules, where cartography is configured through a custom script runner.


ryan-lane:
Nevermind. I'm able to handle this with a subclass of config...

PR #622, opened 2021-05-15
Add Tag support for LoadBalancer & LoadBalancerV2

patrickod:
Extend the `resourcegroupstaggingapi` intel module with support for Load Balancer(V2) <> Tag relationships

### DIFF
--- a/cartography/intel/aws/resourcegroupstaggingapi.py
+++ b/cartography/intel/aws/resourcegroupstaggingapi.py
@@ -41,12 +41,14 @@ def get_short_id_from_elb_arn(alb_arn: str) -> str:
     """
     return alb_arn.split('/')[-1]
 
-def get_short_id_from_alb_arn(alb_arn: str) -> str:
+
+def get_short_id_from_lb2_arn(alb_arn: str) -> str:
     """
-    Return the ALB name from the ARN
+    Return the (A|N)LB name from the ARN
     For example, for arn:aws:elasticloadbalancing:::loadbalancer/app/foo/ab123", return 'foo'.
-    :param arn: The ALB's full ARN
-    :return: The ALB's name
+    For example, for arn:aws:elasticloadbalancing:::loadbalancer/net/foo/ab123", return 'foo'.
+    :param arn: The (N|A)LB's full ARN
+    :return: The (N|A)LB's name
     """
     return alb_arn.split('/')[-2]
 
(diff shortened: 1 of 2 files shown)


cartography/intel/aws/resourcegroupstaggingapi.py
```diff
@@ -48,6 +66,8 @@ def get_bucket_name_from_arn(bucket_arn: str) -> str:
     'ec2:vpc': {'label': 'AWSVpc', 'property': 'id', 'id_func': get_short_id_from_ec2_arn},
     'ec2:transit-gateway': {'label': 'AWSTransitGateway', 'property': 'id'},
     'ec2:transit-gateway-attachment': {'label': 'AWSTransitGatewayAttachment', 'property': 'id'},
+    'elasticloadbalancing:loadbalancer': {'label': 'LoadBalancer', 'property': 'name', 'id_func': get_short_id_from_elb_arn},
+    'elasticloadbalancing:loadbalancer/app': {'label': 'LoadBalancerV2', 'property': 'name', 'id_func': get_short_id_from_alb_arn},
```
achantavy, line 70: [question] Hey, so I didn't write our ELBv2 code and I don't know load balancers in AWS really well, but I see that ELBv2s can either be have `loadbalancer/app` or `loadbalancer/net` in the resource name: https://docs.aws.amazon.com/service-authorization/latest/reference/list_elasticloadbalancingv2.html.

Our [schema doc](https://github.com/lyft/cartography/blob/master/docs/schema/aws.md#loadbalancerv2) says we support both ALBs and NLBs, and a quick look at the [code](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/ec2/load_balancer_v2s.py) seems to confirm this.

Should also add another line for

```python
'elasticloadbalancing:loadbalancer/net': {'label': 'LoadBalancerV2', 'property': 'name', 'id_func': get_short_id_from_nlb_arn},
```
?

achantavy (commented):
Thanks for the help! Just left a quick question.

achantavy (approved):
Thanks for the contribution!

achantavy:
Looks like you just gotta fix some nit picky CI check but looks good to me otherwise

patrickod:
thanks @achantavy I think I've squashed the last of the failures here.

achantavy (approved):

PR #623, opened 2021-05-17
Extended AWS Lambda Support

kedarghule:
Cartography currently supports populating only AWS Lambda Functions nodes in graph db. With this PR, following extensions are added to the already existing code - 

1) The code will now also populate resources like Lambda Function Aliases, Event Source Mappings and Lambda Layers.
2) The code will now ingest more attributes of the AWS Lambda Function resource, thereby providing more information.
3) Currently the code does not use the UNWIND approach, instead it uses loops to populate the AWS Lambda Function resources. This PR rectifies that.


cartography/intel/aws/lambda_function.py
```diff
@@ -29,47 +33,193 @@ def get_lambda_data(boto3_session: boto3.session.Session, region: str) -> List[D
...
     MERGE (owner)-[r:RESOURCE]->(lambda)
     ON CREATE SET r.firstseen = timestamp()
     SET r.lastupdated = {aws_update_tag}
-    WITH lambda
-    MATCH (role:AWSPrincipal{arn: {Role}})
+    WITH lambda, lf
+    MATCH (role:AWSPrincipal{arn: lf.Role})
     MERGE (lambda)-[r:STS_ASSUME_ROLE_ALLOW]->(role)
     ON CREATE SET r.firstseen = timestamp()
     SET r.lastupdated = {aws_update_tag}
```
kunaals, line 72: Context after the UNWIND should be indented

cartography/intel/aws/lambda_function.py
```diff
@@ -29,47 +33,193 @@ def get_lambda_data(boto3_session: boto3.session.Session, region: str) -> List[D
...
+    SET a.aliasname = alias.Name,
+    a.functionversion = alias.FunctionVersion,
+    a.description = alias.Description,
+    a.revisionid = alias.RevisionId,
+    a.lastupdated = {aws_update_tag}
+    WITH a, alias
+    MATCH (lambda:AWSLambda{id: alias.FunctionArn})
+    MERGE (lambda)-[r:KNOWN_AS]->(a)
+    ON CREATE SET r.firstseen = timestamp()
+    SET r.lastupdated = {aws_update_tag}
```
kunaals, line 155: Ditto indent here

cartography/intel/aws/lambda_function.py
```diff
@@ -29,47 +33,193 @@ def get_lambda_data(boto3_session: boto3.session.Session, region: str) -> List[D
...
+    for page in paginator.paginate(FunctionName=lambda_function['FunctionName']):
+        event_source_mappings.extend(page['EventSourceMappings'])
+
+    return event_source_mappings
+
+
+@timeit
+def get_lambda_function_details(
+        boto3_session: boto3.session.Session, data: List[Dict], region: str,
+) -> Generator[Any, Any, Any]:
```
kunaals, line 107: looks like this generator returns 4 things? also can you be more specific than just 'Any' here?

cartography/intel/aws/lambda_function.py
```diff
@@ -29,47 +33,193 @@ def get_lambda_data(boto3_session: boto3.session.Session, region: str) -> List[D
...
-        )
+        function_aliases = get_function_aliases(lambda_function, client)
+        event_source_mappings = get_event_source_mappings(lambda_function, client)
+        layers = lambda_function.get('Layers', [])
+        yield lambda_function['FunctionArn'], function_aliases, event_source_mappings, layers
+
+
+@timeit
+def load_lambda_function_details(
+        neo4j_session: neo4j.Session, lambda_function_details: List[Tuple[Any, Any, Any, Any]], update_tag: int,
```
kunaals, line 118: Can you be more specific than `Any, Any, Any, Any` in `List[Tuple[Any, Any, Any, Any]]`?

cartography/intel/aws/lambda_function.py
kedarghule: Ah I wasn't aware of having indentation after UNWIND. I'll make the change.

cartography/intel/aws/lambda_function.py
kedarghule: Done

cartography/intel/aws/lambda_function.py
kedarghule: Yeah. Made the change in the latest commit

kedarghule: @kunaals So I reverted to the 3 'Any's because for some reason it fails in the linter - `error: "Generator" expects 3 type arguments, but 4 given`. I think this is why in all other codes too irrespective of how many values Generator returns - only 3 have always been mentioned.

Linter error Link - https://github.com/lyft/cartography/pull/623/checks?check_run_id=2691387616

cartography/intel/aws/lambda_function.py
kedarghule: Yes, done

cartography/intel/aws/lambda_function.py
achantavy: Saw this on Stackoverflow: https://stackoverflow.com/a/42531262/565573

> The return type of generator functions can be annotated by the generic type Generator[yield_type, send_type, return_type] provided by typing.py module

So I think you need to do

```
-> Generator[Tuple[Type1, Type2, Type3, Type4], None, None]
```

Admittedly I have no idea what a `send_type` or `return_type` is in this context, but that's the expectation for the `Generator` type.

cartography/intel/aws/lambda_function.py
```diff
@@ -29,47 +33,194 @@ def get_lambda_data(boto3_session: boto3.session.Session, region: str) -> List[D
...
+        ingest_lambda_functions,
+        lambda_functions_list=data,
+        Region=region,
+        AWS_ACCOUNT_ID=current_aws_account_id,
+        aws_update_tag=aws_update_tag,
+    )
+
+
+@timeit
+def get_function_aliases(lambda_function: Dict, client: botocore.client.BaseClient) -> List[Any]:
```
achantavy, line 85: Wrap with `@aws_handle_regions`.

cartography/intel/aws/lambda_function.py
```diff
@@ -29,47 +33,194 @@ def get_lambda_data(boto3_session: boto3.session.Session, region: str) -> List[D
...
+    aliases: List[Any] = []
+    paginator = client.get_paginator('list_aliases')
+    for page in paginator.paginate(FunctionName=lambda_function['FunctionName']):
+        aliases.extend(page['Aliases'])
+
+    return aliases
+
+
+@timeit
+def get_event_source_mappings(lambda_function: Dict, client: botocore.client.BaseClient) -> List[Any]:
```
achantavy, line 95: Wrap with `@aws_handle_regions`.

cartography/intel/aws/lambda_function.py
```diff
@@ -29,47 +33,194 @@ def get_lambda_data(boto3_session: boto3.session.Session, region: str) -> List[D
...
+    event_source_mappings: List[Any] = []
+    paginator = client.get_paginator('list_event_source_mappings')
+    for page in paginator.paginate(FunctionName=lambda_function['FunctionName']):
+        event_source_mappings.extend(page['EventSourceMappings'])
+
+    return event_source_mappings
+
+
+@timeit
+def get_lambda_function_details(
```
achantavy, line 105: Wrap with `@aws_handle_regions`.

achantavy (changes requested):
Also remember to update https://github.com/lyft/cartography/blob/master/cartography/intel/aws/resources.py

cartography/intel/aws/lambda_function.py
kedarghule: Done

cartography/intel/aws/lambda_function.py
kedarghule: Done

cartography/intel/aws/lambda_function.py
kedarghule: Done

kedarghule:
> Also remember to update https://github.com/lyft/cartography/blob/master/cartography/intel/aws/resources.py

I'm sorry but I didn't understand what you're trying to say here. I didn't add a new module in AWS, just some extension in the previous code and since lambda is already mentioned here I thought there was no need of any updates here. Am I missing something?

Also, @achantavy changes made as per your review :)

cartography/data/jobs/cleanup/aws_import_lambda_cleanup.json
```diff
@@ -1,20 +1,50 @@
...
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:AWSLambdaEventSourceMapping)<-[:RESOURCE]-(:AWSLambda)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (:AWSLambdaEventSourceMapping)<-[r:RESOURCE]-(:AWSLambda)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DELETE (r) return COUNT(*) as TotalCompleted",
```
achantavy, line 34: Because all of these cleanup jobs run in order from top to bottom, by this point we might have deleted `:AWSLambda` nodes and the whole `(:AWSLambdaEventSourceMapping)<-[r:RESOURCE]-(:AWSLambda)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}})` path may not match anymore, resulting in orphaned event source mapping and layer nodes.

Therefore, we should move the event source mapping and layer cleanup jobs to the top of this file.
Does this make sense? Let me know if I can explain better.

cartography/intel/aws/lambda_function.py
achantavy: See https://stackoverflow.com/a/42531262/565573 - the type args for a generator are strange. See if you can make sense of that. If not, I won't block if typehints are not 100% correct here.

achantavy:
> > Also remember to update https://github.com/lyft/cartography/blob/master/cartography/intel/aws/resources.py
> 
> I'm sorry but I didn't understand what you're trying to say here. I didn't add a new module in AWS, just some extension in the previous code and since lambda is already mentioned here I thought there was no need of any updates here. Am I missing something?

Ah yeah you're right my bad, makes sense.

cartography/data/jobs/cleanup/aws_import_lambda_cleanup.json
kedarghule: Ah yes, Makes sense. That way it would be applicable to `(:AWSLambdaFunctionAlias)<-[r:KNOWN_AS]-(:AWSLambda)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}})` too right I think. I'll move AWSLambdaFunctionAlias cleanup job to the top as well.

kedarghule:
> > > Also remember to update https://github.com/lyft/cartography/blob/master/cartography/intel/aws/resources.py
> > 
> > 
> > I'm sorry but I didn't understand what you're trying to say here. I didn't add a new module in AWS, just some extension in the previous code and since lambda is already mentioned here I thought there was no need of any updates here. Am I missing something?
> 
> Ah yeah you're right my bad, makes sense.

No worries. I thought maybe I missed something 😆. Anyways, Changes made. Lemme know if any more changes are left or else you can merge :D

cartography/intel/aws/lambda_function.py
kedarghule: Okay I took a look at that. 

The `yield_type` will remain `Any` since I have two types in yield - str (for lambda_function['FunctionArn']) and List[Dict] (for the rest of them).

I'm not sure what `send_type` is. But what I can glean from the context is it is the value that we pass to another function within this function. I feel for our context it can be `Any` since we are passing different types to different functions in `get_lambda_function_details()`.

Finally, `return_type` is if we have a return statement. Since we don't it is `None`. So I'm turning the Generator to [Any, Any, None].

cartography/data/jobs/cleanup/aws_import_lambda_cleanup.json
```diff
@@ -1,20 +1,50 @@
...
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (:AWSAccount{id: {AWS_ID}})-[:RESOURCE]->(:AWSLambda)-[r:STS_ASSUME_ROLE_ALLOW]->(:AWSPrincipal) WITH r LIMIT {LIMIT_SIZE} DELETE r return COUNT(*) as TotalCompleted",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:AWSLambdaFunctionAlias)<-[:KNOWN_AS]-(:AWSLambda)<-[:RESOURCE]-(:AWSAccount{id: {AWS_ID}}) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
```
achantavy: Almost there!

Let's say we have this path

```
(:Alias{lastupdated:1})--(:Lambda:lastupdated:1})--(:Account{lastupdated:2})
```

Note that account is up to date (update_tag = 2) but Alias and Lambda are stale (update_tag=1).

As written, your cleanup job will run from top to bottom. The first statement will 
delete the stale Lambda, so the path will now look like

```
(:Alias{lastupdated:1})--  X --(:Account{lastupdated:2})
```

The problem now is that when we try to run your 3rd statement to clean up Aliases, the path `(:Alias)--(:Lambda)--(:Account)` will no longer match and that Alias above will be an orphaned node connected to nothing.

achantavy: Because of this, we will need to clean up longest paths first, and then shorter ones. Does this make sense?

achantavy (approved):
Woot

cartography/data/jobs/cleanup/aws_import_lambda_cleanup.json
kedarghule: Made the changes in the latest commit

PR #631, opened 2021-06-01
Introduce general purpose functions for graph writes

achantavy:
Follow-up from https://github.com/lyft/cartography/pull/629 and [our cartography write perf spec](https://docs.google.com/document/d/1IZ12R3oROn11LcYj5XunokyOjJkKu-H2O1TEk065Dsk/edit#heading=h.15nsmgmjaaml).

Introduces `merge_nodes()` and `merge_relationships()`. These are library functions designed to write nodes and relationships to the graph while following performance best practices described in https://neo4j.com/blog/bulk-data-import-neo4j-3-0/ and https://dzone.com/articles/tips-for-fast-batch-updates-of-graph-structures-wi.

Once we refactor the existing plugins to use these functions, we will have one central place for graph writes to happen. This has many benefits:
- We only need to update one place to improve write operations for the whole project. Further, Neo4j 3.5 is end-of-life by the end of 2021 so when we move to Neo4j 4.0, we will only need to update just one place instead of updating every single query in the whole project.
- Intelmodule authors will no longer need to write error prone cypher queries. The querybuilder functions handle the boilerplate tasks of setting up the UNWIND structure, transactions, batching, and update tags.

Feedback greatly appreciated here, especially since this is meant to be a core set of functions for cartography. I have included integration tests so you can see how to invoke them.


cartography/graph/loader.py
```diff
@@ -0,0 +1,134 @@
...
+    ingest_query = build_node_ingestion_query(node_label, node_property_map)
+
+    chunk = node_data_list[:batch_size]
+    cursor = 0
+    while len(chunk) > 0:
+        query_args = {
+            'UpdateTag': update_tag,
+            'DictList': chunk,
+        }
+        neo4j_session.write_transaction(_merge_chunk, ingest_query, query_args)
```
achantavy, line 53: @voutilad I implemented the first step in the spec you commented on - thanks very much for your help and input. Does this usage of `write_transaction()` look correct? I'm wondering if I need to explicitly commit the transaction or anything else. I copied from https://neo4j.com/docs/driver-manual/1.7/sessions-transactions/#driver-transactions-explicit and https://neo4j.com/docs/api/python-driver/1.7/transactions.html.

cartography/graph/loader.py
```diff
@@ -0,0 +1,134 @@
...
+            ... <optionally expands the relationship property map too> ...
+
+    :param neo4j_session: The Neo4j transaction object.
+    :param node_label_a: The label of node A.
+    :param search_property_a: the search key to used to search the graph to find node A. For
+    performance, this should be an indexed property. If your graph is large, querying on
+    non-indexed properties can cause your syncs to take **days** to run!
+    :param dict_key_a: The dict key on what value of `search_property_a` to search for.
+    :param node_label_b: The label of node B.
+    :param search_property_b: the search key to used to search the graph to find node B. For
```
roshinis78, line 98:
```suggestion
    :param search_property_b: the search key to use to search the graph to find node B. For
```

cartography/graph/loader.py
```diff
@@ -0,0 +1,134 @@
...
+        MATCH (a:`node_label_a`{`search_property_a`:item.`dict_key_a`})
+        MATCH (b:`node_label_b`{`search_property_b`:item.`dict_key_b})
+        MERGE (a)-[r:`rel_label`]->(b)
+        ON CREATE SET r.firstseen = timestamp()
+        SET r.lastupdated = {UpdateTag},
+            ... <optionally expands the relationship property map too> ...
+
+    :param neo4j_session: The Neo4j transaction object.
+    :param node_label_a: The label of node A.
+    :param search_property_a: the search key to used to search the graph to find node A. For
```
roshinis78, line 93:
```suggestion
    :param search_property_a: the search key to use to search the graph to find node A. For
```

roshinis78 (commented):
Just found some typos :)

cartography/graph/loader.py
voutilad: The `write_transaction` method on a `Session` is a managed transaction, so there's no reason to explicitly commit. It will be committed for you and even retried some number of times (3 I think) if there are transient errors like a lock timeout due to a competing transaction.

One thing I'd recommend is explicitly "consuming" the results inside the transaction work function. Technically if the query doesn't have a `RETURN` statement, it's probably ok, but it's a best practice. Not consuming results can lead to transactions being held open. You can either call `consume()` on the result or iterate over the values. (See https://neo4j.com/docs/api/python-driver/current/api.html#result)

achantavy: This is helpful, thanks! I added an update to consume the results. I can't think of a scenario where we would want to return data out of a `merge` query so I'm just having it debuglog counters here.

cartography/graph/loader.py
```diff
@@ -0,0 +1,144 @@
...
+            'UpdateTag': update_tag,
+            'DictList': chunk,
+            'NodeLabel': node_label,  # used for debug logs
+        }
+        neo4j_session.write_transaction(_merge_chunk, ingest_query, query_args)
+        cursor += batch_size
+        chunk = node_data_list[cursor:cursor + batch_size]
+
+
+def _merge_chunk(tx: neo4j.Transaction, ingest_query: str, query_args: Dict) -> None:
```
achantavy, line 59: @pchheda-lyft @gkleiman This might be a good place to emit stats since this is eventually where all writes will be performed. Can you have a skim and let me know your thoughts?

ruwaifaa:
Do we need something similar to remove stale nodes?

achantavy:
> Do we need something similar to remove stale nodes?

Very good call out. Definitely want to do this: 

1. We'd have one function to `detach delete` nodes in batches, and then another to `delete` relationships in batches.
1. Then a plugin author would write a cleanup function to put those pieces together in a way that makes sense.

I'd like it to be in a separate PR though.

cartography/graph/loader.py
```diff
@@ -0,0 +1,144 @@
...
+    UNWIND {DictList} AS item
+        MERGE (i:`node_label`{id:item.`node_property_map['id']`})
+        ON CREATE SET i.firstseen = timestamp()
+        SET i.lastupdated = {UpdateTag},
+        ... <expand the given node property map to set the other node fields> ...
+
+    :param neo4j_session: The Neo4j session
+    :param node_label: The label of the node to MERGE to Neo4j
+    :param node_property_map: Mapping from node field names to dict keys in `node_data_list`.
+    :param node_data_list: A list of dicts to be written as nodes to the graph.
```
kunaals, line 35: can we simplify these three parameters? in particular the `node_property_map` feels inelegant, mapping Neo4j field names to dict field names. I'd rather just have the dict field names be the same as Neo4j field names.

An idea in the same vein: instead of having all of these parameters be separate entities, what if this method just took a list of dataclass objects? The dataclass would have a name (`node_label`), named fields (replacing the `node_property_map`), and data for each of those fields (`node_data_list`). The field names in the dataclass could be the exact same names attached to the Neo4j objects. Dataclasses are pretty simple and elegant and are typically as efficient as dicts but with better type checks, syntax highlighting, and readability

cartography/graph/loader.py
```diff
@@ -0,0 +1,144 @@
...
+        ON CREATE SET r.firstseen = timestamp()
+        SET r.lastupdated = {UpdateTag},
+            ... <optionally expands the relationship property map too> ...
+
+    :param neo4j_session: The Neo4j transaction object.
+    :param node_label_a: The label of node A.
+    :param search_property_a: the search key to use to search the graph to find node A. For
+    performance, this should be an indexed property. If your graph is large, querying on
+    non-indexed properties can cause your syncs to take **days** to run!
+    :param dict_key_a: The dict key on what value of `search_property_a` to search for.
```
kunaals, line 105: i think search_property and dict_key are confusing given that we've referred to properties as fields before. what about `field_name` and `field_value`?

cartography/graph/loader.py
achantavy: > in particular the node_property_map feels inelegant, mapping Neo4j field names to dict field names. I'd rather just have the dict field names be the same as Neo4j field names.

Makes sense, although the `transform` step of every intelmodule would now need to change the key of each dictionary to be lowercase.

On dataclasses, I'm a fan. Before adding that functionality I wanted to share this PR first to show what is going on under the hood to get a better idea of how we would map things together. How would we represent relationships in dataclasses though? This current approach keeps that part fairly lightweight with just lists of dicts for nodes, and lists of dicts representing node-mapping pairs for relationships.

PR #635, opened 2021-06-10
Make run_analysis_job and run_cleanup_job read location configurable

ryan-lane:
When creating custom intel modules, it would be ideal to also be able to run analysis and cleanup jobs from a custom location as well.

This change updates the utils for running cleanup jobs and analysis jobs to add a package argument that will be passed into read_text. The default paths are kept as a default for the argument.


achantavy (approved):
This is great, thank you!

PR #636, opened 2021-06-11
Add default encryption data to s3 bucket resources

ryan-lane:
This change adds properties to S3 buckets related to default encryption, and updates the schema for these properties in the readme.

It also does a minor refactor of the module to consolidate some of the exception handling.

It also fixes a bug related to loading policy. Previously, if a bucket had no ACLs, the bucket policy also wouldn't be parsed/loaded. Now the policy is parsed/loaded whether or not an ACL exists.

It also updates mypy in pre-commit, to support python 3.9.


ryan-lane:
I removed a warning on "NoSuchBucketPolicy" because that behavior is expected (the bucket isn't deleted, but the bucket exists and has no policy)

ryan-lane:
I found a bug with this in production. Please hold review.

ryan-lane:
I verified the last commit correctly fixes the issue I ran into.

cartography/intel/aws/s3.py
```diff
@@ -102,21 +91,45 @@ def get_acl(bucket: Dict, client: botocore.client.BaseClient) -> Optional[str]:
...
+    return acl
+
+
+@timeit
+def get_encryption(bucket: Dict, client: botocore.client.BaseClient) -> Optional[str]:
+    """
+    Gets the S3 bucket default encryption configuration. Returns encryption configuration or None if not enabled
+    """
+    try:
+        encryption = client.get_bucket_encryption(Bucket=bucket['Name'])
```
achantavy, line 107: Just wondering, is there an AWS API that returns these all in batch? Like can we say "get me all the bucket encryption objects" and then it'll return a list of dicts for all buckets in our environment?

ryan-lane: Alas, you have to make a call per bucket :(

achantavy:
Can you update tests with this?

If it's too troublesome to wire that up, showing a screenshot of nodes with your expected attributes is super helpful too.

ryan-lane:
Oh, I didn't see there were tests for this. Yeah. I'll update them!

ryan-lane:
@achantavy done!

kedarghule (commented):
Tests look good.

cc: @achantavy

achantavy (approved):
Thanks @ryan-lane!

PR #639, opened 2021-06-14
Elb listener sslpolicy

ryan-lane:
This change adds ssl policy information to load balancers and v2 load balancers. Fixes https://github.com/lyft/cartography/issues/637


achantavy (approved):
Thanks for the contribution!

Issue #304, opened 2020-05-30
Map EC2 IAM Instance Profiles to IAM Roles


achantavy:
**Feature request template**

*Title*: *Map EC2 IAM Instance Profiles to IAM Roles*

*Description*:
> Describe your idea.  Please be detailed.  If a feature request, please
describe the desired behavior, what scenario it enables, and how it
would be used.

Cartography EC2 instance nodes have `iaminstanceprofiles`, which determine the IAM role that an EC2 instance is allowed to assume. To make Cartography even more powerful we should map instances to the IAM roles that they may assume.

It should look like this:
```
(:EC2Instance{iaminstanceprofile:"some-profile"})-[:STS_ASSUME_ROLE_ALLOW]->(:AWSRole{role:"some-role})
```

The idea here is that if an EC2 instance is compromised, it is straightforward to understand what other assets the attacker may access. From this point, we already have @SecPrez's previous work mapping roles to other assets, so the following path would also be possible:

```
(:EC2Instance{iaminstanceprofile:"some-profile"})-[:STS_ASSUME_ROLE_ALLOW]->(:AWSRole{role:"some-role})-[:CAN_READ]->(:S3Bucket)
```

Proposed algorithm pseudocode:
```
for each instance:
  role = aws_api.get_instance_profile(instance).roles[0]
  create_mapping("(instance)-->(role)")
```

[optional *Relevant Links*:]
> Any extra documentation required to understand the issue.
- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IamInstanceProfile.html
- https://docs.aws.amazon.com/zh_cn/cli/latest/reference/iam/get-instance-profile.html
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#IAM.Client.get_instance_profile

PR #646, opened 2021-06-21
Map EC2 Instance iaminstanceprofiles to IAM Roles (Issue #304)

phishoes:
test doesnt work: AWSRoles, EC2Instances, and AWSAccounts are made properly, but the load_ec2_instances() doesn't seem to connect AWSAccount to EC2Instance as a RESOURCE


cartography/intel/aws/__init__.py
```diff
@@ -207,6 +207,12 @@ def start_aws_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
         common_job_parameters,
     )
 
+    run_analysis_job(
```
achantavy, line 210: Because the analysis job itself relies on `AWS_ID` for the current AWS account ID, you will want to put this in `_sync_one_account` and not in `start_aws_ingestion`.

tests/data/aws/ec2/instances.py
```diff
@@ -509,6 +509,68 @@
             'ReservationId': 'r-03',
         },
     ],
+    'Roles': [
```
achantavy, line 512: Is `Roles` a field returned from `aws ec2 describe instances`? If not, please move it to tests/data/aws/iam/.

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,87 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
...
+    )
+    
+    data_instances = tests.data.aws.ec2.instances.DESCRIBE_INSTANCES['Reservations']
+    data_iam = tests.data.aws.ec2.instances.DESCRIBE_INSTANCES['Roles']
+
+    cartography.intel.aws.ec2.instances.load_ec2_instances(
+        neo4j_session, data_instances, TEST_REGION, TEST_ACCOUNT_ID, TEST_UPDATE_TAG,
+    )
+
+    cartography.intel.aws.iam.load_roles(
```
achantavy, line 128: Very thorough!

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,87 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
...
+    """
+    Ensure that iaminstanceprofiles are connected to their expected iam roles
+    """
+    neo4j_session.run(
+        """
+        MERGE (aws:AWSAccount{id: {aws_account_id}})
+        ON CREATE SET aws.firstseen = timestamp()
+        SET aws.lastupdated = {aws_update_tag}
+        """,
+        aws_account_id="OWNER_ACCOUNT_ID",
```
achantavy, line 108: I think this is the source of the bug you mentioned. Change `"OWNER_ACCOUNT_ID"` to `TEST_ACCOUNT_ID`.

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,87 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
...
+        aws_update_tag=TEST_UPDATE_TAG,
+    )
+    neo4j_session.run(
```
achantavy, line 111: Delete; no need to test a second account.

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,87 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
...
+
+    cartography.intel.aws.ec2.instances.load_ec2_instances(
+        neo4j_session, data_instances, TEST_REGION, TEST_ACCOUNT_ID, TEST_UPDATE_TAG,
+    )
+
+    cartography.intel.aws.iam.load_roles(
+        neo4j_session, data_iam, TEST_ACCOUNT_ID, TEST_UPDATE_TAG,
+    )
+
+    config = cartography.config.Config(
```
achantavy, line 132: `config` is not needed for this test; delete.

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,87 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
...
+        neo4j_uri='bolt://localhost:7687',
+        update_tag=TEST_UPDATE_TAG,
+        aws_sync_all_profiles=True,
+    )
+
+    common_job_parameters = {
+        "UPDATE_TAG": config.update_tag,
```
achantavy, line 139: Replace `config.update_tag` with `TEST_UPDATE_TAG` like in the other tests.

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,87 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
...
+        "permission_relationships_file": config.permission_relationships_file,
```
achantavy, line 140: Remove `'permission_relationships_file'`; this field is not necessary for this test.

cartography/data/jobs/analysis/aws_ec2_iaminstance.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "EC2 Instances assume IAM roles",
+    "statements": [
+        {
+            "__comment": "Create ASSUMES_ROLE realtionship from ec2 instance to the associated iaminstance iam role",
+            "query":"MATCH (aa:AWSAccount)-[:RESOURCE]->(i:EC2Instance)\nWITH SPLIT(i.iaminstanceprofile, '/')[-1] AS role_name, aa, i\nMATCH (aa)-[:RESOURCE]->(r:AWSRole)\nWHERE r.arn ENDS WITH role_name\nMERGE (i)-[:ASSUMES_ROLE]->(r)",
```
achantavy, line 6: Let's use the name `STS_ASSUMEROLE_ALLOW` since that's the name used in the IAM module for roles assuming other roles.

achantavy (changes requested):
Thank you for working on this and good job grokking our codebase! Left some suggestions that I think will help fix the bugs you mentioned to me.

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
phishoes: I need the account id to match the ones in the existing instances from previous tests. I didn't get the memo that `TEST_ACCOUNT_ID` was the same as the ID of AWSAccount. Do I have to l`oad_ec2_instances()` for each individual AWSAccount ID?

achantavy: I bring up that specific variable because it's available to you in the test so you might as well use it: https://github.com/lyft/cartography/blob/342c21ba622872651313989713f7d209006967f0/tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py#L7

I don't think this particular test needs more than one AWS account so we can just use this single ID.

phishoes: Okay, sounds good. I originally did those two strings because they were what was in the existing instances used for the other tests (among the other poorly named things in those test instances)

tests/data/aws/iam.py
```diff
@@ -142,3 +142,48 @@
         },
     ],
 }
+
+INSTACE_ROLES = {
```
achantavy, line 146: Nit: `INSTANCE`

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,59 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
         for n in nodes
     }
     assert actual_nodes == expected_nodes
+
+
+def test_ec2_iaminstanceprofiles(neo4j_session, *args):
+    """
+    Ensure that iaminstanceprofiles are connected to their expected iam roles
```
achantavy, line 100: Nit: "Ensure that EC2Instances are attached to the IAM Roles that they can assume due to their IAM instance profiles".

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,59 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
         for n in nodes
     }
     assert actual_nodes == expected_nodes
+
+
+def test_ec2_iaminstanceprofiles(neo4j_session, *args):
```
achantavy, line 98: Why `*args`?

tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py
```diff
@@ -90,3 +93,59 @@ def test_ec2_reservations_to_instances(neo4j_session, *args):
...
+    cartography.intel.aws.iam.load_roles(
+        neo4j_session, data_iam, TEST_ACCOUNT_ID, TEST_UPDATE_TAG,
+    )
+
+    common_job_parameters = {
+        "UPDATE_TAG": TEST_UPDATE_TAG,
+    }
+
+    run_analysis_job(
+        'aws_ec2_iaminstance.json',
```
achantavy, line 128: Let's rename this file to 'aws_ec2_iaminstanceprofile.json'

cartography/data/jobs/analysis/aws_ec2_iaminstance.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "EC2 Instances assume IAM roles",
+    "statements": [
+        {
+            "__comment": "Create STS_ASSUMEROLE_ALLOW realtionship from ec2 instance to the associated iaminstance iam role",
```
achantavy, line 5: Nit: How does something like "Create STS_ASSUMEROLE_ALLOW relationships from EC2 instances to the IAM roles they can assume via their iaminstanceprofiles" sound?

achantavy (changes requested):
Almost there. Left a couple of nits.

Please update the schema doc here: https://github.com/lyft/cartography/blob/master/docs/schema/aws.md#relationships-23 and here: https://github.com/lyft/cartography/blob/master/docs/schema/aws.md#relationships-11 (having the same content in two places is fine).

and please also update the PR title to include the issue number (#304) and make it more specific e.g. "Map EC2 Instance iaminstanceprofiles to IAM Roles".

achantavy (approved):
Ship it!

PR #651, opened 2021-07-01
GitHub: Improve logging when failing to parse Python requirements

aneeshusa:
A few improvements based on reviewing logs from an internal run yesterday:
- Entirely skip lines that are empty (after stripping comments);
  it's not useful to try to parse an empty line
  or to get a log line that we couldn't parse it.
- Include the original line in the message instead of the stripped one,
  so we can more easily debug which line is at fault.
- Use `exc_info=True` to log the exception
  instead of manually stringifying the exception message;
  this will provide a full stacktrace and enable filtering.
- Minimize the amount of code in the try/except,
  and the amount of verbiage in the log line.


achantavy (approved):
Thanks!!

PR #655, opened 2021-07-20
Add tag support for most supported AWS resources

ryan-lane:
This change adds support for syncing in tags for most of the currently supported AWS resources. It also adds some missing indexes.


ryan-lane:
It seems this API doesn't support IAM user, role, and group. I'd like to be able to sync tags for them. Would it be OK to fetch the tags via the IAM API? If so, should I stick that in the IAM code, rather than this resource? (I'll do this in a separate PR)

achantavy:
> It seems this API doesn't support IAM user, role, and group. I'd like to be able to sync tags for them. Would it be OK to fetch the tags via the IAM API? If so, should I stick that in the IAM code, rather than this resource? (I'll do this in a separate PR)

It seems like IAM is on this list: https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/supported-services.html

Was the resource name specified wrong, or maybe the way this module calls the API doesn't generalize perfectly?

ryan-lane:
It's on the list, but they have a more detailed list of what's supported
here:
https://docs.aws.amazon.com/ARG/latest/userguide/supported-resources.html#services-iam

Note that IAM user, role, and group aren't supported (very little from IAM
is).

achantavy:
> It's on the list, but they have a more detailed list of what's supported
here:
https://docs.aws.amazon.com/ARG/latest/userguide/supported-resources.html#services-iam
Note that IAM user, role, and group aren't supported (very little from IAM
is).

Ah, bummer. Looks good to me, thanks for the PR!

achantavy (approved):

PR #659, opened 2021-07-29
Add SQS support to intel modules

ryan-lane:
This change adds support for syncing in AWS SQS queues, adding a relation from the account to the queue, and a relation between queues and their dead letter queues. It also syncs tags for SQS queues.


ninjaSec7 (approved):

cartography/intel/aws/sqs.py
```diff
@@ -0,0 +1,140 @@
...
+import neo4j
+
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_sqs_queue_list(boto3_session: boto3.session.Session, region: str) -> List[str]:
```
kedarghule, line 16: Add `@aws_handle_regions` before this function. You'll also have to add `from cartography.util import aws_handle_regions` in the beginning.

cartography/intel/aws/sqs.py
```diff
@@ -0,0 +1,140 @@
...
+    client = boto3_session.client('sqs', region_name=region)
+    paginator = client.get_paginator('list_queues')
+    queues: List[Any] = []
+    for page in paginator.paginate():
+        queues.extend(page.get('QueueUrls', []))
+    return queues
+
+
+@timeit
+def get_sqs_queue_attributes(
```
kedarghule, line 26: Similarly, below `@timeit` and before the function definition, add `@aws_handle_regions`.

cartography/intel/aws/sqs.py
```diff
@@ -0,0 +1,140 @@
...
+        MATCH (queue:SQSQueue{id: relation.arn}), (deadletter:SQSQueue{id: relation.dead_letter_arn})
+        MERGE (queue)-[r:HAS_DEADLETTER_QUEUE]->(deadletter)
+        ON CREATE SET r.firstseen = timestamp()
+        SET r.lastupdated = {aws_update_tag}
+    """
+    neo4j_session.run(
+        attach_deadletter_to_queue,
+        Relations=data,
+        aws_update_tag=aws_update_tag,
+    )
```
kedarghule, line 120: @achantavy can you please take a quick look at this code section to see if everything is okay?

kedarghule (changes requested):
Took a quick look and requested a couple of changes.  Since a new intel module is being added, add the name here too - https://github.com/lyft/cartography/blob/master/README.md in the Supported Platforms section. Also, I forgot to mention this last time so add in the Secrets Manager name too :D

ryan-lane:
During a production run of this module, I found an exception I need to catch. Let me fix that really quick.

ryan-lane:
ok. done!

cartography/intel/aws/sqs.py
achantavy: On a quick pass it looks good to me!

kedarghule (approved):
Thanks for the contribution @ryan-lane :D Approving and merging now 🚀

PR #660, opened 2021-07-29
Sync AWS Secret Manager data from all regions

ryan-lane:
This change syncs secret manager data for all regions, rather than just the region that cartography is being run from.

### DIFF
--- a/cartography/intel/aws/secretsmanager.py
+++ b/cartography/intel/aws/secretsmanager.py
@@ -5,6 +5,7 @@ from typing import List
 import boto3
 import neo4j
 
+from cartography.util import aws_handle_regions
 from cartography.util import dict_date_to_epoch
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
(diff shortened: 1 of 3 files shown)


ninjaSec7 (dismissed):

cartography/intel/aws/secretsmanager.py
```diff
@@ -13,8 +13,8 @@
 
 
 @timeit
-def get_secret_list(boto3_session: boto3.session.Session) -> List[Dict]:
-    client = boto3_session.client('secretsmanager')
+def get_secret_list(boto3_session: boto3.session.Session, region: str) -> List[Dict]:
+    client = boto3_session.client('secretsmanager', region_name=region)
```
kedarghule, line 17: Add `@aws_handle_regions` before this function.

kedarghule (changes requested):
All looks good. One small change and then we can merge. 

PS:  Since a new intel module is being added, add the name here too - https://github.com/lyft/cartography/blob/master/README.md in the Supported Platforms section. I forgot about this last time. (I know I mentioned this in the SQS PR as well, sorry)

kedarghule (approved):
Looks good! Thank you for the contribution @ryan-lane :D

PR #663, opened 2021-08-06
Fix Azure library errors in azure-mgmt-sql

jg10:
This PR pins the version of azure-mgmt-sql to version 1.0.0 and updates the PyJWT library to match the referenced library version in azure-cli-core. 

As referenced [here](https://pypi.org/project/azure-mgmt-sql/) version 2 of the azure-mgmt-sql library introduced many breaking changes and I experienced our Azure sync failing when getting an SQL servers Threat Detection Policy which was one of the breaking changes made. 

I've tested this and it works again with the pinned version as otherwise it will get the latest version which is currently 3.0.1 and will not work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
         "azure-mgmt-cosmosdb>=6.0.0",
         "msrestazure >= 0.6.4",
         "azure-mgmt-storage>=16.0.0",
-        "azure-mgmt-sql==1.0.0",
+        "azure-mgmt-sql<=1.0.0",
         "azure-identity>=1.5.0",
     ],
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
         "packaging",
         "cryptography<3.4,>=3.2",
         "python-digitalocean>=1.16.0",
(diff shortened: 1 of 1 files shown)


mpurusottamc (dismissed):
Thanks for fixing the version to 1.0.0. This looks good.

achantavy:
Let's also file an issue to update the module to support the latest SQL library

setup.py
```diff
@@ -43,15 +43,15 @@
...
+        "PyJWT>=2.1.0",
         "adal>=1.2.4",
         "azure-cli-core>=2.12.0",
         "azure-mgmt-compute>=5.0.0",
         "azure-mgmt-resource>=10.2.0",
         "azure-mgmt-cosmosdb>=6.0.0",
         "msrestazure >= 0.6.4",
         "azure-mgmt-storage>=16.0.0",
-        "azure-mgmt-sql>=0.11.0",
+        "azure-mgmt-sql==1.0.0",
```
achantavy, line 54: Rather than `==` pin, is it possible to express this as a `<=` pin?

jg10: Yep sounds good @achantavy, I just made the change and tested it to be sure.

achantavy (approved):

PR #673, opened 2021-08-23
Add support for AWS config

ryan-lane:
This change adds basic support for AWS config. Currently it supports Configuration Recorders, Delivery Channels, and Config Rules. With these objects, it's possible to query to ensure config is enabled for an account/region (region has configuration recorder and delivery channel), and to see which config rules are enabled specifically for the account/region.

Since organizational rules can also be enabled for config, this support doesn't make it possible to see every config rule enabled, but support for that is cross-account/cross-region, and is out of scope for this PR.


achantavy (approved):
Looks really good! Thank you!

PR #675, opened 2021-08-23
Add basic security hub support

ryan-lane:
This change adds very basic support for security hub. Currently, it just adds a SecurityHub node type, which is associated with an account, if security hub is enabled for an account.


cartography/intel/aws/securityhub.py
```diff
@@ -0,0 +1,67 @@
...
+import boto3
+import neo4j
+
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
```
achantavy, line 15: Also decorate with `@aws_handle_regions` from `cartography.util`

cartography/intel/aws/securityhub.py
```diff
@@ -0,0 +1,67 @@
...
+def get_hub(boto3_session: boto3.session.Session) -> Dict:
+    client = boto3_session.client('securityhub')
+    try:
+        return client.describe_hub()
+    except client.exceptions.ResourceNotFoundException:
```
achantavy, line 20: Maybe we should log a warning here too? Up to you

cartography/intel/aws/securityhub.py
```diff
@@ -0,0 +1,67 @@
...
+
+@timeit
+def load_hub(
+    neo4j_session: neo4j.Session,
+    data: Dict,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    ingest_hub = """
+    WITH {Hub} AS hub
```
achantavy, line 32: Could we use `UNWIND` here?

achantavy (changes requested):
Some minor comments but looks good; thanks for the PR!

cartography/intel/aws/securityhub.py
ryan-lane: I think it'll throw an exception if it's not enabled for the account, so I think this is what we want.

cartography/intel/aws/securityhub.py
ryan-lane: There's only a single one per account/region.

cartography/intel/aws/securityhub.py
ryan-lane: This isn't regional. It's per account.

achantavy (approved):

achantavy (approved):

PR #687, opened 2021-09-14
Fix access of json string redrive policy in sqs

ryan-lane:
This change json loads the redrive policy from the SQS queue, which is a JSON string, though the boto docs indicate it's an object. It gets the deadletter queue from the loaded object.

Fixes https://github.com/lyft/cartography/issues/677


achantavy (approved):
Thanks for taking this on!

PR #692, opened 2021-09-21
Deprecate cleanup of AWS accounts

marco-lancini:
Currently, Cartography doesn't support the case of multiple runs ingesting different AWS Organizations,
as the most recent run will cleanup the AWS Accounts of all the other Organizations.

After discussing this with @achantavy we came to the following conclusions:

* The current idea of detach deleting everything attached to an `AWSAccount` is too broad
* Consumers of the graph should make use of the `lastupdated` fields to determine if something is stale or not
* In the case of multiple accounts, it is safer to err toward keep more data instead of broad deleting it
* We also have cleanup jobs for the individual resources already

This PR deprecates the current behaviour, and removes the cleanup job targeted to the `AWSAccount` node type.

### DIFF
--- a/cartography/data/jobs/cleanup/aws_account_cleanup.json
+++ b/cartography/data/jobs/cleanup/aws_account_cleanup.json
@@ -1,9 +1,0 @@
-{
-  "statements": [
-    {
-      "query": "MATCH (n:AWSAccount) WHERE n.lastupdated <> {UPDATE_TAG} WITH n LIMIT {LIMIT_SIZE} DETACH DELETE (n) return COUNT(*) as TotalCompleted",
-      "iterative": true,
-      "iterationsize": 100
-  }],
-  "name": "cleanup AWS Accounts"
-}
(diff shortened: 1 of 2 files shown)


achantavy:
Just wanted to add that this PR intends to support the scenario of running a sync on some accounts at a certain time of day and syncing the other accounts later on in the day without the second sync deleting all the accounts that were synced in the first one.

Explained a different way, suppose at 5 am cartography syncs AWS accounts A, B, and C, and then at 10am cartography syncs accounts A and B. The way the code is currently written, cartography will delete all assets from account C. Cartography makes the assumption that on any given run you will sync all the accounts that you care about. This assumption is not correct in the case if you have the constraint of needing to sync accounts A and B in the morning and account C later on in the day.

One side effect of this PR is that over time one's graph will have orphaned AWS accounts. As Marco mentioned, we already have cleanup jobs for the resources that belong to any given account, so we don't expect leaving behind stale accounts to yield a large footprint in one's graph.

achantavy (approved):

PR #694, opened 2021-09-23
Add ref to docker instructions

kunaals:
Docker compose is likely the easiest way to get up and running on one's laptop, though it has a limited number of intelmodules (only GH at time of writing)


achantavy (approved):

achantavy:
Thanks!

PR #695, opened 2021-09-26
Add HTTP retries for github API

skiptomyliu:
Github API is flakey where for some reason a 403 is returned every now and then.  Retrying the request goes through.

This PR adds a maximum of 5 retries before raising an exception.

### DIFF
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -85,7 +85,7 @@ def fetch_all(
             resp = fetch_page(token, api_url, organization, query, cursor)
         except requests.exceptions.Timeout:
             retry += 1
-            if retry <= retries:
+            if retry >= retries:
                 logger.error(
                     f"GitHub: Could not retrieve page of resource {resource_type} due to API timeout.",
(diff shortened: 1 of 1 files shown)


cartography/intel/github/util.py
```diff
@@ -88,11 +89,13 @@ def fetch_all(
...
-                exc_info=True,
-            )
-            raise
+            retry += 1
+            if retry <= retries:
+                logger.error(
+                    f"GitHub: Could not retrieve page of resource `{resource_type}` due to HTTP error.",
+                    exc_info=True,
+                )
+                raise
```
achantavy, line 98: I _think_ we need a `continue` here otherwise it will continue on to line 99 which will result in a KeyError because `resp` will be falsy.

```suggestion
                raise
            else:
                continue
```

achantavy (commented):
Thank you very much for helping with this! Left a question

achantavy (approved):
Thank you!

achantavy (approved):

PR #696, opened 2021-09-28
new boto errors

dwanderson-intel:
String checking always tricky.

`UnauthorizedOperation` is probably a bit broad to pass through, but, seemed to run fine enough for our account.

### DIFF
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -79,9 +79,11 @@ def timeit(method):
 def aws_handle_regions(func):
     ERROR_CODES = [
         'AccessDeniedException',
+        'AccessDenied',
         'UnrecognizedClientException',
         'InvalidClientTokenId',
         'AuthFailure',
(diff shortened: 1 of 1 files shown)


dwanderson-intel:
trying to merge into my fork, whoops.

achantavy:
@dwanderson-intel Have you seen these exception types? If so, we'd love to have your change upstream :-)

dwanderson-intel:
Thanks for the prompt follow-up!

I'll try to poke around and see - the renaming of the `AccessDenied` is maybe just a botocore version change; I'll double check what I have.

`UnauthorizedOperation` seems to be related to region-blocking that my company has across all AWS accounts, but I'm not sure how to make it more granular / appropriate to ignore (eg if someone didn't have the `SecurityAudit` IAM Policy, they might get blocked from a lot of things and we'd want to alert them).

PR #699, opened 2021-10-01
[draft] query and process go modules during repo sync

crockeo:
this draft PR introduces the ability to ingest go modules in cartography. go 1.17 [changes how the go module graph works and adds extra indirect dependencies into the `go.mod`](https://golang.org/doc/go1.17#go-command). this PR leverages that new behavior to collect requirements from more go modules!

this naive ingestion scheme will probably have to change before it's productionized but it's a good start 🙂

### DIFF
--- a/cartography/intel/github/repos.py
+++ b/cartography/intel/github/repos.py
@@ -76,6 +76,11 @@ GITHUB_ORG_REPOS_PAGINATED_GRAPHQL = """
                             text
                         }
                     }
+                    gomod:object(expression: "HEAD:go.mod") {
+                        ... on Blob {
+                            text
+                        }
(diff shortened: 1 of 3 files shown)


aneeshusa:
GitHub supports Go now in their dependency graph - we should definitely consider just using that instead of writing our own code: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems

aneeshusa:
That being said, I haven't looked at the GitHub API in detail yet - it may be the case that we can get more details (like what's indirect vs not) by parsing ourselves instead of using the GitHub API.

PR #711, opened 2021-10-19
Fix #652: Also sync ec2 images referenced in graph

### LINKED ISSUE #652: Fetch AWS EC2 AMIs based on what's in the graph
*Description*:

Fetching AMIs is difficult, because the default filter options for the API are roughly:

* AMIs I own
* AMIs managed by AWS
* AMIs managed by the community

Fetching all images takes nearly a minute per region, per account, and loads tens of thousands of images, which aren't being actively used.

It would be ideal to fetch the AMIs in a multi-step manner, which introspects the graph to determine which AMIs should be fetched:

1. Load EC2 instances, launch configurations, instance templates, etc.
2. Query the graph to find AMIs in use
3. Fetch/load AMIs based on what's in use.

I'm not sure at this point if cartography is doing graph introspection to determine what to fetch at this point, so I wanted to float the idea first.

ryan-lane:
This change syncs more AMIs into the graph, based on AMIs that are referenced in instance, launch configurations, and launch templates in the graph, per-account/region.

It also adds some attributes that weren't being pulled in.

### DIFF
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -89,9 +89,9 @@ def transform_ecr_repository_images(repo_data: Dict) -> List[Dict]:
     return repo_images_list
 
 
-@timeit
-def load_ecr_repository_images(
-        neo4j_session: neo4j.Session, repo_images_list: List[Dict], region: str, aws_update_tag: int,
+def _load_ecr_repo_img_tx(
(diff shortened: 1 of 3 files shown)


ryan-lane:
Fixes https://github.com/lyft/cartography/issues/652

PR #715, opened 2021-11-01
Deprecate cleanup of DigitalOcean accounts

marco-lancini:
## Summary

Following https://github.com/lyft/cartography/pull/692, which tackled the lack of support for different AWS Organizations, this PR ports the same  concept to the DigitalOcean ingestion.

## The problem

Currently, Cartography only supports the ingestion of one single DigitalOcean Account.

In scenarios where multiple DigitalOcean Accounts are at play (where each Account is ingested by its own Cartography Job), this means that the most recent run will delete all other accounts ingested by other runs.

## Solution proposed with this PR

Similar to https://github.com/lyft/cartography/pull/692, this PR deprecates the cleanup of DigitalOcean Accounts based on the `UPDATE_TAG`


kunaals (approved):
This seems fine for now. I think a better long term solution for this would be to make a call to DigitalOcean and get a list of all active accounts and delete orphaned accounts.

PR #718, opened 2021-11-05
Transform and load dependencies from setup.cfg

olivia-hong:
Previously, requirements.txt was the only file cartography loaded dependencies from.
This PR adds support for setup.cfg, which is where many Python libraries define dependencies;
specifically in `install_requires`, `setup_requires`, and `extras_require`

Also tested locally by running cartography, ensuring the data was loaded correctly, and checking that I could query dependencies from `setup.cfg`.

Note: the GH dependency graph unfortunately [does not support setup.cfg ingestion](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems) at the moment. Although setup.cfg support has been implement in [dependabot](https://github.com/dependabot/dependabot-core/pull/3423), the two projects [are completely separate with no timeline for merge](https://github.com/dependabot/dependabot-core/pull/2281#discussion_r516652962) at the moment.


crockeo (commented):
some initial thoughts

cartography/intel/github/repos.py
```diff
@@ -235,58 +244,123 @@ def _transform_collaborators(collaborators: Dict, repo_url: str, transformed_col
...
-            out_requirements_files.append({
-                "id": requirement_id,
-                "name": canon_name,
-                "specifier": spec,
-                "version": pinned_version,
-                "repo_url": repo_url,
-            })
+
+def _transform_setup_cfg_requirements(
+    setup_cfg_contents: Dict,
```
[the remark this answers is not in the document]
olivia-hong: The contents are whatever is returned from the graphql API. Right now that's a single key value pair where the value is the entire setup.cfg file as a string, which I don't think is worth defining static typing for.

cartography/intel/github/repos.py
```diff
@@ -235,58 +244,123 @@ def _transform_collaborators(collaborators: Dict, repo_url: str, transformed_col
...
+    :param repo_url: str: The URL of the GitHub repo.
+    :param out_requirements_files: Output array to append transformed results to.
+    :return: Nothing.
+    """
+    if setup_cfg_contents and setup_cfg_contents.get('text'):
+        text_contents = setup_cfg_contents['text']
+        setup_cfg = ConfigParser()
+        try:
+            setup_cfg.read_string(text_contents)
+        except MissingSectionHeaderError:
```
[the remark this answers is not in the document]
olivia-hong: thanks!

cartography/intel/github/repos.py
```diff
@@ -235,58 +244,123 @@ def _transform_collaborators(collaborators: Dict, repo_url: str, transformed_col
...
+    out_requirements_files: List[Dict],
+) -> None:
+    """
+    Performs data transformations for the setup.cfg file in a GitHub repo, if available.
+    :param setup_cfg_contents: Dict: Contains contents of a repo's setup.cfg file.
+    :param repo_url: str: The URL of the GitHub repo.
+    :param out_requirements_files: Output array to append transformed results to.
+    :return: Nothing.
+    """
+    if setup_cfg_contents and setup_cfg_contents.get('text'):
```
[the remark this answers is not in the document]
olivia-hong: optional and changed the function argument, good catch thanks!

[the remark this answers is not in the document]
crockeo: if you want job security through obfuscation...

`if (text_contents := (setup_cfg_contents or {}).get('text')) is not None:`

jk that's not even valid since we're not guaranteed to be running under python3.8

cartography/intel/github/repos.py
```diff
@@ -235,58 +242,124 @@ def _transform_collaborators(collaborators: Dict, repo_url: str, transformed_col
...
+        text_contents = setup_cfg_contents['text']
+        setup_cfg = ConfigParser()
+        try:
+            setup_cfg.read_string(text_contents)
+        except Error:
```
crockeo, line 280: `import configparser` and `configparser.Error` seems clearer here since `Error` is a fairly generic term

crockeo (approved):
code changes look good to me. let's also get a review from cartography owners as discussed over slack.

crockeo (dismissed):
re-approving after change and merge

tests/data/github/repos.py
```diff
@@ -32,6 +32,14 @@
         },
         'collaborators': {'edges': [], 'nodes': []},
         'requirements': {'text': 'cartography\nhttplib2<0.7.0\njinja2\nlxml\n-e git+https://example.com#egg=foobar\nhttps://example.com/foobar.tar.gz\npip @ https://github.com/pypa/pip/archive/1.3.1.zip#sha1=da9234ee9982d4bbb3c72346a6de940a148ea686\n'},  # noqa
+        'setupCfg': {
+            'text': '''
```
aneeshusa, line 36: nit: Some `textwrap.dedent` wouldn't hurt in this file similar to what we do in our internal repos

cartography/intel/github/repos.py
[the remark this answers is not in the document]
aneeshusa: nit: prefer an early return so you can eliminate a level of indentation. i.e.
```
if not setup_cfg_contents or 'text' not in setup_cfg_contents:
    return
text_contents = setup_cfg_contents['text']
...
```

cartography/intel/github/repos.py
```diff
@@ -121,7 +126,8 @@ def transform(repos_json: List[Dict]) -> Dict:
         _transform_repo_objects(repo_object, transformed_repo_list)
         _transform_repo_owners(repo_object['owner']['url'], repo_object, transformed_repo_owners)
         _transform_collaborators(repo_object['collaborators'], repo_object['url'], transformed_collaborators)
-        _transform_python_requirements(repo_object['requirements'], repo_object['url'], transformed_requirements_files)
+        _transform_requirements_txt(repo_object['requirements'], repo_object['url'], transformed_requirements_files)
+        _transform_setup_cfg_requirements(repo_object['setupCfg'], repo_object['url'], transformed_requirements_files)
```
aneeshusa, line 130: What happens if a dep is listed in both places (e.g. no bounds in setup.cfg but pinned in requirements.txt)?
What do we want to have happen (e.g. what would be most useful for our query patterns/the https://github.com/lyft/cartography/blob/master/docs/usage/samplequeries.md sample queries)?
This is likely worth a test case.

docs/schema/github.md
```diff
@@ -174,7 +174,8 @@ Representation of a single Programming Language [language object](https://develo
 
 ## Dependency::PythonLibrary
 
-Representation of a Python library as listed in a [requirements.txt](https://pip.pypa.io/en/stable/user_guide/#requirements-files) file.
+Representation of a Python library as listed in a [requirements.txt](https://pip.pypa.io/en/stable/user_guide/#requirements-files)
+or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) file
```
aneeshusa, line 178: Probably worth mentioning that we include everything from all of `install_requires`, `setup_requires`, and `extras_require`, folks might only think about `install_requires` at first.

aneeshusa (approved):

cartography/intel/github/repos.py
olivia-hong: If it's listed in both places and has different specifiers for each usage, cartography will create two separate nodes, which I think makes sense rather than any sort of "merging" logic. This allows users to query what version(s) are being used or perhaps find out that they are specifying a dependency in multiple files when it's not needed.

aneeshusa: That SGTM, thanks! Just wanted to check that we didn't have one "overwrite" the other

cartography/intel/github/repos.py
```diff
@@ -235,58 +241,125 @@ def _transform_collaborators(collaborators: Dict, repo_url: str, transformed_col
...
-    :param req_file_contents: str: The text contents of the requirements file.
+    Performs data transformations for the requirements.txt file in a GitHub repo, if available.
+    :param req_file_contents: Dict: The contents of the requirements.txt file.
     :param repo_url: str: The URL of the GitHub repo.
     :param out_requirements_files: Output array to append transformed results to.
     :return: Nothing.
     """
     if req_file_contents and req_file_contents.get('text'):
         text_contents = req_file_contents['text']
+        reqs_list = text_contents.split("\n")
```
FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg, line 258: Would prefer this named the same as the param in the function you  call below

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg:
Re testing, have you run this on a few setup.cfgs from some of  our bigger projects?

FAYiEKcbD0XFqF2QK2E4viAHg8rMm2VbjYKdjTg (dismissed):
Looks good; a few minor questions but good to go.

olivia-hong:
> Re testing, have you run this on a few setup.cfgs from some of our bigger projects?

I tested on repos with setup.cfg's that had requirements across install_requires, setup_requires, and extras_require.

aneeshusa (approved):

olivia-hong:
🚀

PR #726, opened 2021-11-22
Secure test_suite.yml workflow

varunsh-coder:
This PR makes two changes to the test_suite.yml workflow
1. Adds minimum token permissions for the `GITHUB_TOKEN`. This is a security best practice as per [GitHub Actions Hardening Guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#restricting-permissions-for-tokens) and is also part of the [Scorecards](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) checks.
2. Adds `step-security/harden-runner` to the jobs. This is a new GitHub Action that provides runtime security for GitHub hosted runner. As of now, it can be used to set allowed outbound traffic for workflow runs to prevent credential exfiltration. You can see the results from the run on the fork [here](https://github.com/varunsh-coder/cartography/runs/4287501915?check_suite_focus=true#step:3:3). I noticed outbound calls to security.ubuntu.com and azure.archive.ubuntu.com on port 80, which I will investigate. After the workflow has run a couple of times, and discovered outbound calls, I can create another PR to restrict outbound calls to allowed domains. 

@achantavy would love to pilot `step-security/harden-runner` on this repo to get feedback and improve the experience. Do let me know if you have any questions. Thanks!

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -15,7 +15,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
-      - uses: step-security/harden-runner@v0.1.1
+      - uses: step-security/harden-runner@v0.3.0
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
@@ -29,7 +29,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
-      - uses: step-security/harden-runner@v0.1.1
+      - uses: step-security/harden-runner@v0.3.0
       - uses: actions/checkout@v2
       # See https://debian.neo4j.com/
       - name: Neo4j setup
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -15,7 +15,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


marco-lancini:
Hi @varunsh-coder, are you able to provide some context around `step-security/harden-runner`?
I couldn't find a Github repo with its code, and a Google search returned no results

varunsh-coder:
> Hi @varunsh-coder, are you able to provide some context around `step-security/harden-runner`?
> I couldn't find a Github repo with its code, and a Google search returned no results

Hi @marco-lancini Sure. Here is the repo: https://github.com/step-security/harden-runner

Please let me know if you need any more info. Thanks!

varunsh-coder:
> Hi @varunsh-coder, are you able to provide some context around `step-security/harden-runner`?
> I couldn't find a Github repo with its code, and a Google search returned no results

@marco-lancini since you had asked this earlier, and it was in pilot, I had not provided it. Posted a blog about this today, sharing here: https://blog.stepsecurity.io/introducing-harden-runner-github-action-to-prevent-supply-chain-attacks-91a030b67a77

achantavy:
Closing this one out. I don't see the usecase for our repo yet but we can revisit this in the future.

PR #727, opened 2021-11-24
Better retries for github's flakey APIs

skiptomyliu:
- Github sometimes returns 502s with ChunkedEncodingError.  Retrying the requests makes it work.
- Minor bug fix where retry count was not being reset to 0 after a successful request.  This allows retrying per individual failed request, instead of using a global retry for all paginations.
- Minor refactor to remove duplicate code.

### DIFF
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -84,28 +84,24 @@ def fetch_all(
     while has_next_page:
         try:
             resp = fetch_page(token, api_url, organization, query, cursor)
+            retry = 0
         except requests.exceptions.Timeout:
             retry += 1
-            if retry >= retries:
-                logger.error(
-                    f"GitHub: Could not retrieve page of resource {resource_type} due to API timeout.",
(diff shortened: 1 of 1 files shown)


cartography/intel/github/util.py
```diff
@@ -84,31 +84,27 @@ def fetch_all(
...
             if retry >= retries:
                 logger.error(
                     f"GitHub: Could not retrieve page of resource `{resource_type}` due to HTTP error.",
                     exc_info=True,
                 )
                 raise
-            else:
+            elif retry > 0:
                 time.sleep(1 * retry)
-                continue
```
kunaals, line 108: I think we still need this continue to actually retry the request?

kunaals (changes requested):

cartography/intel/github/util.py
skiptomyliu: thanks for catching this

kunaals (approved):

PR #732, opened 2021-12-11
Update test-requirements for Python 3.8 support

ruwaifaa:
Python 3.8 in some cases can improve performance by more than 10% over Python 3.7 and earlier versions! There’s also many bugfixes and new features in the standard library; see the Python 3.8 [release notes](https://docs.python.org/3/whatsnew/3.8.html) for more info. We need to make sure that cartography runs with Python 3.8
This PR involves making sure that the packages work with Python 3.8 and unit tests run successfully

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 COPY . /srv/cartography
 
+# Installs pip supported by python3.8
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3.8 get-pip.py
 
 RUN pip install -e . && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,15 @@ WORKDIR /srv/cartography
 
 ENV PATH=/venv/bin:$PATH
 RUN apt-get update && \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -4,14 +4,15 @@ WORKDIR /srv/cartography
...
-    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.8-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
-
 COPY . /srv/cartography
 
-RUN pip3 install -e . && \
-    pip3 install -r test-requirements.txt
+RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3.8 get-pip.py
```
achantavy, line 12: Can you add a quick comment on why this layer is needed so we remember why later on?

achantavy (changes requested):

Dockerfile
ruwaifaa: Done

achantavy (approved):
Thanks!

PR #741, opened 2021-12-20
Add more lambda attributes, and make relationship between lambda and ECR

ryan-lane:
This change pulls in more lambda attributes, and makes a relationship between lambda and ECR, for lambdas that use images, and have a matching sha256.


kedarghule (approved):
Looks good! Thanks for the contribution and sorry for the delay in reviewing this!

PR #743, opened 2022-01-07
Add support for AWS ECS

ryan-lane:
This PR adds support for ECS. It supports: clusters, container instances, services, tasks, containers, task definitions and container definitions. It adds most relevant relationships, though some relationships listed in the attachments for the objects are not yet defined.

This isn't a complete abstraction. Followups will be necessary to add relationships between ECS and other AWS resources (like between EC2 instances and ECSContainerInstances, between tasks and EC2 networking, between ECS tasks and IAM, etc).


ryan-lane:
I need to followup to add integration tests, but otherwise this is generally feature complete for what I plan on implementing.

ryan-lane:
Ok. This is ready for review!

tests/integration/cartography/intel/aws/test_ecs.py
```diff
@@ -0,0 +1,244 @@
...
+        (
+            CLUSTER_ARN,
+            "test_cluster",
+            "ACTIVE",
+        ),
+    }
+
+    nodes = neo4j_session.run(
+        """
+        MATCH (n:ECSCluster)
```
achantavy, line 25: Can you please add tests to verify the relationships between the new nodes?

cartography/intel/aws/ecs.py
```diff
@@ -0,0 +1,590 @@
...
+    client = boto3_session.client('ecs', region_name=region)
+    paginator = client.get_paginator('list_clusters')
+    clusters: List[Any] = []
+    cluster_arns: List[Any] = []
+    for page in paginator.paginate():
+        cluster_arns.extend(page.get('clusterArns', []))
+    # TODO: also include attachment info, and make relationships between the attachements
+    # and the cluster.
+    includes = ['SETTINGS', 'CONFIGURATIONS']
+    for i in range(0, len(cluster_arns), 100):
```
achantavy, line 31: What's the chunking for? Throttling?

cartography/intel/aws/ecs.py
```diff
@@ -0,0 +1,590 @@
...
+from cartography.util import dict_date_to_epoch
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+@aws_handle_regions
+def get_ecs_clusters(boto3_session: boto3.session.Session, region: str) -> Tuple[List, List]:
```
achantavy, line 21: Nit: can you type out the extra characters for `Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]`? Same comment for the other typehints.

achantavy (changes requested):
Can you please add some tests to ensure the relationships between the new nodes are created as expected? Other than that looks really good to me; the rest of my comments are nits.

cartography/intel/aws/ecs.py
ryan-lane: The API takes a maximum of 100 arns in the argument

achantavy:
@kedarghule  - just removed you so that it doesn't spam the channel :) feel free to look if you have time too

docs/schema/aws.md
```diff
@@ -2841,3 +2855,281 @@ Representation of an AWS EC2 [Elastic IP address](https://docs.aws.amazon.com/AW
...
+| agent\_update\_status | The status of the most recent agent update. If an update wasn't ever requested, this value is NULL. |
+| registered\_at | The Unix timestamp for the time when the container instance was registered. |
+
+### Relationships
+
+- An ECSCluster has ECSContainerInstances
+
+        ```
+        (ECSCluster)-[HAS_CONTAINER_INSTANCE]->(ECSContainerInstance)
+        ```
```
kedarghule, line 2920: Please mention this relationship under `ECSCluster` too and follow the same pattern for the rest of the resources.

cartography/intel/aws/ecs.py
```diff
@@ -0,0 +1,590 @@
...
+                region,
+            )
+            load_ecs_tasks(
+                neo4j_session,
+                cluster_arn,
+                tasks,
+                region,
+                current_aws_account_id,
+                update_tag,
+            )
```
kedarghule, line 589: Not a blocker but just a suggestion: How about we add in a new function for these lines of code so that the sync() function remains short and only has the regions loop? Would just help improve the readability imo.

kedarghule (changes requested):

cartography/intel/aws/ecs.py
ryan-lane: I think if we were unit testing modules this would be a good step for testibility, but I actually think from a reading perspective it's easier to see the actions in a row. I'm not opposed to it, but I'd rather do this in a refactor later :)

kedarghule: Sure no problem!

kedarghule (dismissed):
Looks good to me! If @achantavy approves too, we can merge :)

kedarghule (approved):

achantavy (approved):

achantavy:
Looks good, thank you!

PR #744, opened 2022-01-12
Do not fail on pods without labels in k8s intel module

ryan-lane:


kedarghule (approved):
Thanks for the quick fix! :)

PR #745, opened 2022-01-12
Add support for versioning and public block access configs in S3 module

ryan-lane:
This change pulls in versioning and public access block information to S3 nodes.


kedarghule (approved):
Looks good! And sorry for taking time for this one!

PR #747, opened 2022-01-13
Extension of support for current GCP & Azure services

mpurusottamc:
For GCP, support for these additional services - IAM, API Gateway, Cloud KMS, Cloudrun, GCP SQL, Firestore, Bigtable, Cloud Function

For Azure, support for these additional services - Function App, Network, Resource Group, Tag, IAM, Key Vaults, AKS


mpurusottamc:
There are few commits which do not have proper Author info. And that's causing the CLA check to fail. We are trying to fix this. Once resolved, we will reopen the PR.

PR #774, opened 2022-03-01
Log errors in github graphql response

ryan-lane:
GraphQL queries generally always return with a 200 response, but put errors into the response. To log errors, we need to log them from the response.


achantavy (approved):
Thank you!

PR #775, opened 2022-03-03
Use the python driver neo4j>=4.0.0

ramonpetgrave64:
Upgrade the neo4j driver, improving stability.
The new 4.x drivers [should be backwards compatible](https://github.com/neo4j/neo4j-python-driver#neo4j-bolt-driver-for-python) with older neo4j 3.x database that we currently use.
```
... These drivers will also be compatible with the previous Neo4j release, although new server features will not be available.
```
Tests pass, but before we merge, ~~we need to make sure that logging still works correctly~~
https://github.com/lyft/cartography/blob/master/cartography/cli.py#L461
https://github.com/lyft/cartography/blob/master/cartography/driftdetect/cli.py#L242

The logger is now called `'bolt'`
https://github.com/neo4j/neo4j-python-driver/blob/8f708fe9374f7e6bdfdf84ff611f266a8a5e9344/neo4j/io/_bolt3.py#L62

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -458,7 +458,7 @@ def main(argv=None):
     logging.basicConfig(level=logging.INFO)
     logging.getLogger('botocore').setLevel(logging.WARNING)
     logging.getLogger('googleapiclient').setLevel(logging.WARNING)
-    logging.getLogger('neo4j.bolt').setLevel(logging.WARNING)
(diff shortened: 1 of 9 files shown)


achantavy (approved):
Please change the PR title to disambiguate from the neo4j database and the neo4j driver  -- maybe something like "Use neo4j driver >=4.0.0"

Issue #778, opened 2022-03-06
AWS: use transactions for resourcegroupstaggingapi load_tags()


achantavy:
*Bug Template**

*Title*: *One line description*

*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

The resourcegroupstaggingapi sync is using `neo4j_session.run()` instead of a managed transaction function (which comes with auto retries) and is crashing. 

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run cartography on a large enough graph.

*Logs*:
> If applicable, copy and paste your console log with the failing stack trace.

```
Traceback (most recent call last):
  File "/{PATH}/neo4j/io/_common.py", line 55, in _yield_messages
    buffer.receive(sock, 2)
  File "/{PATH}/neo4j/packstream.py", line 482, in receive
    n = sock.recv_into(view[self.used:end], end - self.used)
  File "/{PATH}/gevent/_socketcommon.py", line 688, in recv_into
    return self._sock.recv_into(*args)
ConnectionResetError: [Errno 104] Connection reset by peer

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/{PATH}/cartography/intel/aws/__init__.py", line 145, in _sync_multiple_accounts
    _sync_one_account(
  File "/{PATH}/cartography/intel/aws/__init__.py", line 72, in _sync_one_account
    RESOURCE_FUNCTIONS['resourcegroupstaggingapi'](**sync_args)
  File "/{PATH}/cartography/util.py", line 108, in timed
    result = method(*args, **kwargs)
  File "/{PATH}/cartography/intel/aws/resourcegroupstaggingapi.py", line 198, in sync
    load_tags(neo4j_session, tag_data, resource_type, region, update_tag)
  File "/{PATH}/cartography/util.py", line 108, in timed
    result = method(*args, **kwargs)
  File "/{PATH}/cartography/intel/aws/resourcegroupstaggingapi.py", line 156, in load_tags
    neo4j_session.run(
  File "/{PATH}/neo4j/work/simple.py", line 215, in run
    self._autoResult._run(
  File "/{PATH}/neo4j/work/result.py", line 114, in _run
    self._attach()
  File "/{PATH}/neo4j/work/result.py", line 203, in _attach
    self._connection.fetch_message()
[...]

PR #779, opened 2022-03-08
AWS: use transactions in resourcegrouptaggingapi

ramonpetgrave64:
Adresses #778

Upgrades the `load_tags` function to use neo4j transactions.

### DIFF
--- a/cartography/intel/aws/resourcegroupstaggingapi.py
+++ b/cartography/intel/aws/resourcegroupstaggingapi.py
@@ -128,9 +128,11 @@ def get_tags(boto3_session: boto3.session.Session, resource_types: List[str], re
     return resources
 
 
-@timeit
-def load_tags(
-    neo4j_session: neo4j.Session, tag_data: Dict, resource_type: str, region: str,
+def _load_tags_tx(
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/resourcegroupstaggingapi.py
```diff
@@ -153,14 +155,28 @@ def load_tags(
...
         Region=region,
     )
 
 
+@timeit
+def load_tags(
+    neo4j_session: neo4j.Session, tag_data: Dict, resource_type: str, region: str,
+    aws_update_tag: int,
+) -> None:
+    neo4j_session.write_transaction(
```
meng-han, line 171: curiosity: why is it write_transaction but not read_transaction(if there's such thing?), this method is load tags and sounds to be read only

ramonpetgrave64: There is a read_transaction. The difference is when there is a clustered environment, with different hosts designated for reads and different hosts designated for writes.

But load_tags does indeed write to the database.

meng-han (approved):

PR #781, opened 2022-03-10
Add basic support for AWS SSM

ryan-lane:


docs/root/modules/aws/schema.md
```diff
@@ -2975,3 +2988,81 @@ Representation of an AWS ECS [Container](https://docs.aws.amazon.com/AmazonECS/l
...
+- SSMInstancePatch is a resource under the AWS Account.
+
+        ```
+        (AWSAccount)-[RESOURCE]->(SSMInstancePatch)
+        ```
+
+- EC2Instances have SSMInstancePatches
+
+        ```
+        (EC2Instance)-[RESOURCE]->(SSMInstancePatch)
```
ramonpetgrave64, line 3067: I hesitate to call this relationship between instance and patch as RESOURCE. Maybe call it PATCH?

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@timeit
+def sync(
+    neo4j_session: neo4j.Session, boto3_session: boto3.session.Session, regions: List[str], current_aws_account_id: str,
+    update_tag: int, common_job_parameters: Dict,
+) -> None:
+    for region in regions:
+        logger.info("Syncing SSM for region '%s' in account '%s'.", region, current_aws_account_id)
+        running_instance_ids = get_running_instance_ids(neo4j_session, region, current_aws_account_id)
```
ramonpetgrave64, line 187: Can we also have patches and information for non-running instances.

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,192 @@
...
+from cartography.util import aws_handle_regions
+from cartography.util import dict_date_to_epoch
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_running_instance_ids(neo4j_session: neo4j.Session, region: str, current_aws_account_id: str) -> List[str]:
```
ramonpetgrave64, line 18: This means that the instances must exist in the graph before we can begin discovering SSM patches and information. I think it would be better sort of do the discovery from both directions:
1. query AWS for SSM Information and Patches
2. Load those into the graph as SSMInstanceInformation and SSMInstancePatch nodes
  1. Also use the instance_id to create what I like to call husk EC2Instance nodes, where they only have the id property. 
 
This is how we do it for EC2Instance and EC2Subnet.
https://github.com/lyft/cartography/blob/master/cartography/intel/aws/ec2/instances.py#L191
https://github.com/lyft/cartography/blob/master/cartography/intel/aws/ec2/subnets.py

docs/root/modules/aws/schema.md
ryan-lane: Oh, whoops. That name is wrong in the docs. I actually use `HAS_PATCH` here. I'll update this.

docs/root/modules/aws/schema.md
```diff
@@ -2975,3 +2988,81 @@ Representation of an AWS ECS [Container](https://docs.aws.amazon.com/AmazonECS/l
...
+- SSMInstanceInformation is a resource under the AWS Account.
+
+        ```
+        (AWSAccount)-[RESOURCE]->(SSMInstanceInformation)
+        ```
+
+- SSMInstanceInformation is a resource of an EC2Instance
+
+        ```
+        (EC2Instance)-[RESOURCE]->(SSMInstanceInformation)
```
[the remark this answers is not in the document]
ryan-lane: I can use `HAS_INFORMATION` instead.

cartography/intel/aws/ssm.py
ryan-lane: Yeah, I can switch this to not filter by instance state.

cartography/intel/aws/resources.py
```diff
@@ -48,6 +48,8 @@
     'ec2:launch_templates': sync_ec2_launch_templates,
     'ec2:autoscalinggroup': sync_ec2_auto_scaling_groups,
     'ec2:instance': sync_ec2_instances,
+    # This module must be included after ec2:instance, as uses instance
```
ramonpetgrave64, line 51: some copy pasta here. Say 
```
# This module `ec2:instance` must be included before `ssm`, as `ssm` uses EC2Instance data provided by this module.
```

cartography/intel/aws/resources.py
```diff
@@ -80,6 +82,8 @@
     'secretsmanager': secretsmanager.sync,
     'securityhub': securityhub.sync,
     'sqs': sqs.sync,
+    # This module must be included after ec2:instance, as uses instance
```
ramonpetgrave64, line 85: copy pasta. Maybe rephrase to the inverse of my above comment.

cartography/intel/aws/ssm.py
```diff
@@ -31,14 +31,14 @@ def get_running_instance_ids(neo4j_session: neo4j.Session, region: str, current_
...
     client = boto3_session.client('ssm', region_name=region)
     instance_information: List[Dict[str, Any]] = []
-    for i in range(0, len(running_instance_ids), 50):
-        running_instance_ids_chunk = running_instance_ids[i:i + 50]
+    for i in range(0, len(instance_ids), 50):
+        instance_ids_chunk = instance_ids[i:i + 50]
         instance_information_chunk = client.describe_instance_information(
-            Filters=[{"Key": "InstanceIds", "Values": running_instance_ids_chunk}],
+            Filters=[{"Key": "InstanceIds", "Values": instance_ids_chunk}],
             MaxResults=50,
```
ramonpetgrave64, line 42: 50 might be too low to to give out NextToken, but to be safe we should still paginate this.

cartography/intel/aws/ssm.py
ramonpetgrave64: I understand. That makes me want to add this directly to the ec2 module, just because they are so dependent. But I'm okay with `ssm` being by itself.

ramonpetgrave64 (changes requested):
Just the small fixes, plus tests, and then I think it will be ready to merge.

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,192 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_instance_ids(neo4j_session: neo4j.Session, region: str, current_aws_account_id: str) -> List[str]:
+    get_instances_query = """
+    MATCH (:AWSAccount{id: {AWS_ACCOUNT_ID}})-[:RESOURCE]->(i:EC2Instance)
+    WHERE i.region = {Region}
+    RETURN i.id
```
ramonpetgrave64, line 22: We may also need to paginate this query.

cartography/intel/aws/ssm.py
ryan-lane: @ramonpetgrave64 this is a bit of a weird API. It takes a list of instance IDs, and gives back a list of results for the instances. It seems that the maximum number of instance IDs it'll take in is 50, and the maximum results it'll give out is 50, so though it's pageable, I think effectively, it'll never page. I guess it doesn't hurt to be paranoid about it, and I can add a paginator in.

cartography/intel/aws/ssm.py
ryan-lane: I can't find a really simple way to do this, and all the pagination support in cartography itself is for writing, not querying. I think even with a very large number of results (100k+, which would be a very very large number of instances for a single region and one account), the query will be fast without pagination.

I'd happy to paginate it, but right now the only pagination support via cartography is for write queries. Maybe this is something we can followup, with library functions for doing queries like this?

ryan-lane:
Ok. I've added tests, and I believe I've handled the feedback!

cartography/intel/aws/ssm.py
ramonpetgrave64: I should have linked, and you're right we don't currently do pagination for neo4j queries. But it can be done with SKIP and LIMIT https://neo4j.com/docs/cypher-manual/current/clauses/skip/. 

But I won't block the PR on this, since even 100k+ results can return very quickly.

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,193 @@
...
+def load_instance_information(
+    neo4j_session: neo4j.Session,
+    data: List[Dict[str, Any]],
+    region: str,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    ingest_query = """
+    UNWIND {InstanceInformation} AS instance
+        MERGE (i:SSMInstanceInformation{id: instance.InstanceId})
```
ramonpetgrave64, line 78: non-blocker I think `activation_id` would be a better candidate for the node's id.
https://lyft.github.io/cartography/dev/writing-intel-modules.html#all-nodes-need-these-fields

tests/integration/cartography/intel/aws/test_ssm.py
```diff
@@ -0,0 +1,140 @@
...
+    assert actual_nodes == {"i-01"}
+
+    nodes = neo4j_session.run(
+        """
+        MATCH (:EC2Instance{id: "i-02"})-[:HAS_INFORMATION]->(n:SSMInstanceInformation)
+        RETURN n.id
+        """,
+    )
+    actual_nodes = {n["n.id"] for n in nodes}
+    assert actual_nodes == {"i-02"}
```
ramonpetgrave64, line 85: non-blocker: We should combine these two queries and asserts into one query and one assert.

tests/integration/cartography/intel/aws/test_ssm.py
```diff
@@ -0,0 +1,140 @@
...
+    assert actual_nodes == {"i-01-test.x86_64:0:4.2.46-34.amzn2"}
+
+    nodes = neo4j_session.run(
+        """
+        MATCH (:EC2Instance{id: "i-02"})-[:HAS_PATCH]->(n:SSMInstancePatch)
+        RETURN n.id
+        """,
+    )
+    actual_nodes = {n["n.id"] for n in nodes}
+    assert actual_nodes == {"i-02-test.x86_64:0:4.2.46-34.amzn2"}
```
ramonpetgrave64, line 140: Try to combine al these queries and asserts into one.

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,193 @@
...
+        ON CREATE SET r2.firstseen = timestamp()
+        SET r2.lastupdated = {aws_update_tag}
+    """
+    for ii in data:
+        ii["LastPingDateTime"] = dict_date_to_epoch(ii, "LastPingDateTime")
+        ii["RegistrationDate"] = dict_date_to_epoch(ii, "RegistrationDate")
+        ii["LastAssociationExecutionDate"] = dict_date_to_epoch(ii, "LastAssociationExecutionDate")
+        ii["LastSuccessfulAssociationExecutionDate"] = dict_date_to_epoch(ii, "LastSuccessfulAssociationExecutionDate")
+
+    neo4j_session.run(
```
ramonpetgrave64, line 119: non-blocker: make this into a transaction.

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,193 @@
...
+        ON CREATE SET r2.firstseen = timestamp()
+        SET r2.lastupdated = {aws_update_tag}
+    """
+    for p in data:
+        p["InstalledTime"] = dict_date_to_epoch(p, "InstalledTime")
+        # Split the comma separated CVEIds, if they exist, and strip
+        # the empty string from the list if not.
+        p["CVEIds"] = list(filter(None, p.get("CVEIds", "").split(",")))
+
+    neo4j_session.run(
```
ramonpetgrave64, line 167: non blocker, make this into a transaction

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+@timeit
+def get_instance_ids(neo4j_session: neo4j.Session, region: str, current_aws_account_id: str) -> List[str]:
+    get_instances_query = """
+    MATCH (:AWSAccount{id: {AWS_ACCOUNT_ID}})-[:RESOURCE]->(i:EC2Instance)
+    WHERE i.region = {Region}
+    RETURN i.id
+    """
+    results = neo4j_session.run(get_instances_query, AWS_ACCOUNT_ID=current_aws_account_id, Region=region)
```
ramonpetgrave64, line 24: non-blocker: make this into a transaction

ramonpetgrave64 (commented):
Some small things that I noticed after taking a third look. To me, these are non blockers, and I'm okay with merging as is. But I do want to give you a chance to address them.

ramonpetgrave64 (approved):

cartography/intel/aws/ssm.py
```diff
@@ -0,0 +1,193 @@
...
+def load_instance_patches(
+    neo4j_session: neo4j.Session,
+    data: List[Dict[str, Any]],
+    region: str,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    ingest_query = """
+    UNWIND {InstancePatch} AS patch
+        MERGE (p:SSMInstancePatch{id: patch._instance_id + "-" + patch.Title})
```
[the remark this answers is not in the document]
ryan-lane: kb_id isn't unique, unfortunately. That's why I went with title, which is roughly the kb_id and the underlying version.

cartography/intel/aws/ssm.py
ryan-lane: I'm mostly going off examples from other modules. Would it be good to have a PR that goes through and converts all the modules to using transactions? Maybe ideally using a common function that makes it a bit easier. From what I remember, it's a bit of a pain.

cartography/intel/aws/ssm.py
ryan-lane: `activation_id` isn't the unique ID for the record, as activations are another feature of SSM, and if it isn't enabled, that field is null. `instance_id` is unique and always set, as the instance information is one to one with instances.

ryan-lane:
I'd say let's merge it in, and followup on some of the rest of the issues in followup PRs.

PR #788, opened 2022-03-24
Add support for AWS Codeguru

t-maoshengye:
Co-authored-by: QiGuangxing <t-guangxing.qi@paypay-corp.co.jp>


cartography/intel/github/repos.py
```diff
@@ -17,6 +17,9 @@
 
 logger = logging.getLogger(__name__)
 
+# (ryan-lane): To workaround a bug in the graphql API (see:
+# https://github.community/t/user-node-is-null-if-email-is-private-or-unset/179323)
+# email has been removed from the collaborators->nodes portion of this query
```
ryan-lane, line 22: Ah. I think you've based this incorrectly. Your change should be based on master, not on paypay-main.

PR #790, opened 2022-03-24
Export types, set stricter mypy baseline

aneeshusa:
# Export types, set stricter mypy baseline

Lyft recently shipped a change to our internal GitHub code
that included a type error, which we should have been able to catch in CI.
Update Cartography to make this possible:
- Follow https://peps.python.org/pep-0561/ to export the type hints
  already present in the Cartography codebase,
  by adding a `py.typed` file and including it in our distributions.
- In our recent buggy change, the lack of typing on decorators (namely `@timeit`)
  meant that although all the functions involved had type hints,
  because they were decorated with `@timeit` mypy saw them typed as `Any`
  and was not able to warn us of a mismatch of argument types/count.
  So, add type annotations to `@timeit` and set `disallow_untyped_decorators = true`.
- Ratchet up the default mypy configuration to be more strict generally.
  We're not yet ready to set `strict = true` globally,
  and that option is unforunately only settable globally,
  not on a per-module basis
  so we cannot opt modules out.
  Set as many global strictness flags as feasible,
  since we cannot incrementally adopt them in a per-module basis.
  Fix all errors outside intel modules;
  for intel modules, `# type: ignore` if there are only 1 or 2 errors
  or else ignore errors on the entire module.

# Ratchet mypy on modules used by Lyft's GitHub ingestion

Lyft has additional changes planned to its GitHub code in the near term,
so ratchet up the strictness flags to catch any errors.
Note that we leave `disallow_any_generics = false`;
setting it to `true` would be more strict
but it's not clear it's yet worth it to add those types,
vs switching to e.g. TypedDict.

The `GraphJob.run_from_json` and `GraphJob.run_from_json_file`
methods lose the default value of `None` for their `parameters` argument;
this appears to be a breaking change for API callers not passing that argument.
However, the parameters argument is passed through to a dict.update call,
which does not accept `None` as an input,
so any callers not passing that argument would have failed at runtime anyways.

# Loosen typing on GraphJob json file_paths

The `file_path` parameter is passed directly to `open`,
which supports `typing.Union[str, bytes, os.PathLike]`,
and internally at Lyft we always pass a string for this parameter.

We need to parse out a shortname as a string from the path,
so we can't expand this type to include bytes;
only expand to be a Union of str and Path.

# Remove unneeded use of global stats client

Instead of manually scoping our stat, use `get_stats_client`.

### DIFF
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -19,7 +19,7 @@ import neo4j
 
 from cartography.graph.job import GraphJob
 from cartography.graph.statement import get_job_shortname
-from cartography.stats import _scoped_stats_client
+from cartography.stats import get_stats_client
 from cartography.stats import ScopedStatsClient
 
 if sys.version_info >= (3, 7):
@@ -114,11 +114,9 @@ def timeit(method: F) -> F:
     # Allow access via `inspect` to the wrapped function. This is used in integration tests to standardize param names.
     @wraps(method)
     def timed(*args, **kwargs):  # type: ignore
-        stats_client = _scoped_stats_client  # global client
+        stats_client = get_stats_client(method.__module__)
         if stats_client.is_enabled():
-            # Example metric name "cartography.intel.aws.iam.get_group_membership_data"
-            metric_name = f"{method.__module__}.{method.__name__}"
-            timer = stats_client.timer(metric_name)
+            timer = stats_client.timer(method.__name__)
             timer.start()
             result = method(*args, **kwargs)
             timer.stop()
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -3,10 +3,15 @@ import re
 import sys
 from functools import wraps
 from string import Template
+from typing import Any
+from typing import BinaryIO
+from typing import Callable
+from typing import cast
 from typing import Dict
 from typing import Iterable
 from typing import List
 from typing import Optional
+from typing import TypeVar
 from typing import Union
 
 import botocore
@@ -25,7 +30,12 @@ else:
 logger = logging.getLogger(__name__)
 
 
-def run_analysis_job(filename, neo4j_session, common_job_parameters, package='cartography.data.jobs.analysis'):
+def run_analysis_job(
+    filename: str,
+    neo4j_session: neo4j.Session,
+    common_job_parameters: Dict,
+    package: str = 'cartography.data.jobs.analysis',
+) -> None:
     GraphJob.run_from_json(
         neo4j_session,
         read_text(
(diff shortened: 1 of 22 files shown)


cartography/intel/aws/ec2/instances.py
```diff
@@ -121,7 +121,7 @@ def _load_ec2_instance_tx(
         instance_state: str,
         current_aws_account_id: str,
         region: str,
-        update_tag: str,
+        update_tag: int,
```
ramonpetgrave64, line 124: The update tag really should be either string or int.
https://lyft.github.io/cartography/dev/writing-intel-modules.html#handling-cartography-s-update-tag

cartography/intel/digitalocean/compute.py
```diff
@@ -15,7 +15,7 @@ def sync(
         neo4j_session: neo4j.Session,
         manager: Manager,
         projects_resources: dict,
-        digitalocean_update_tag: str,
+        digitalocean_update_tag: int,
```
ramonpetgrave64, line 18: The update tag really should be either string or int.
https://lyft.github.io/cartography/dev/writing-intel-modules.html#handling-cartography-s-update-tag

cartography/intel/digitalocean/management.py
```diff
@@ -24,7 +24,7 @@ def sync(
 def sync_projects(
         neo4j_session: neo4j.Session,
         manager: Manager,
-        digitalocean_update_tag: str,
+        digitalocean_update_tag: int,
```
ramonpetgrave64, line 27: The update tag really should be either string or int.
https://lyft.github.io/cartography/dev/writing-intel-modules.html#handling-cartography-s-update-tag

ramonpetgrave64 (dismissed):
Looks good. Just minor changes about update_tag type and how the stats client is imported.

cartography/util.py
```diff
@@ -86,20 +96,23 @@ def merge_module_sync_metadata(
...
     This decorator uses statsd to time the execution of the wrapped method and sends it to the statsd server.
     This is only active if config.statsd_enabled is True.
     :param method: The function to measure execution
     """
     # Allow access via `inspect` to the wrapped function. This is used in integration tests to standardize param names.
     @wraps(method)
-    def timed(*args, **kwargs):
-        stats_client = get_stats_client(None)
+    def timed(*args, **kwargs):  # type: ignore
+        stats_client = _scoped_stats_client  # global client
```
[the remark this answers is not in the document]
aneeshusa: `cartography.stats.get_stats_client` accepts only a `str` for prefix, passing a `None` value wouldn't be allowed by mypy.

We could allow it in mypy, but that doesn't make sense as:
- every `get_stats_client(None)` call will be creating a new proxy object unnecessarily
- we don't want folks to use the root stats client anyways, they should be used nested/scoped clients in general

It looks like the code does end up creating a scope (manually!!) below, let me see about using `get_stats_client` with an actual str argument.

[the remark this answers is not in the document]
aneeshusa: Pushed a commit to refactor this func and use `get_stats_client` in the right way.

skiptomyliu (dismissed):

skiptomyliu (approved):

aneeshusa:
Alex and Ramon are out today, so going to merge as-is and work with them to cut a release with these changes when they're back.

PR #791, opened 2022-03-25
Pull in more EMR data from the available describe cluster data

ryan-lane:
This change pulls in useful data that's already available on the EMR describe cluster data being pulled down.


kedarghule (approved):
Looks good! Thank you for the contribution!

PR #792, opened 2022-03-28
Fix typing of `update_tag` parameters

aneeshusa:
Automated via `rg 'update_tag: str' -l | xargs gsed -i 's/update_tag: str/update_tag: int/g'`.
Pulling this out of the [larger change to improve mypy configuration](https://github.com/lyft/cartography/pull/790).

### DIFF
--- a/cartography/intel/aws/dynamodb.py
+++ b/cartography/intel/aws/dynamodb.py
@@ -30,7 +30,7 @@ def get_dynamodb_tables(boto3_session: boto3.session.Session, region: str) -> Li
 @timeit
 def load_dynamodb_tables(
     neo4j_session: neo4j.Session, data: List[Dict], region: str, current_aws_account_id: str,
-    aws_update_tag: str,
+    aws_update_tag: int,
 ) -> None:
     ingest_table = """
     MERGE (table:DynamoDBTable{id: {Arn}})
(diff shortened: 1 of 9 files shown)


olivia-hong (dismissed):

ramonpetgrave64 (dismissed):
The update tag really should be either string or int.
https://lyft.github.io/cartography/dev/writing-intel-modules.html#handling-cartography-s-update-tag

achantavy:
`Union[str, int]` as a typehint could work.

When cartography was first built types weren't really in mind, so thanks for the help 😅  

As reference, here's the intent of the update tag field: https://lyft.github.io/cartography/dev/writing-intel-modules.html?highlight=update_tag#handling-cartography-s-update-tag

aneeshusa:
Discussed with Alex in a meeting internally - agreed to constrain update tags to always be timestamps represented as ints and standardize on `int` for the type hints here.

achantavy:
@marco-lancini @ryan-lane @mpurusottamc @SecPrez -- any concerns about enforcing the update_tag as an `int`? Previously this was meant to be any value to enable users to tag nodes with any value they wanted, but this is bad for dev exp since it stops us from having good static checks. I'm in favor of the change but want to hear others' thoughts.

ryan-lane:
@achantavy yes, I'd be in favor of this change. I was a bit surprised it was int|str because the tag is by convention set at the beginning of the run as an int timestamp.

achantavy (approved):
Thanks for taking this on and thanks for the input everyone :)

PR #794, opened 2022-04-01
Add intel module to sync CVE data from NIST

ryan-lane:
This change adds a cve intel module, which syncs CVE data from the NIST json feeds, using the v4 json format feeds.

This module is a bit special, since it doesn't have a cleanup job, and it introspects the graph to determine what it needs to sync.

CVE data is generally historical and isn't deleted. The data is updated, but is updated through a feed. Historical data is grouped by year, and updates are available in special feeds that are updated daily, and contain updates for the past 8 days. The yearly data is also updated daily, but assuming the yearly data has been synced, it's only necessary to pull in the recent and modified feeds after that point.

The module syncs each year, then syncs the recent and historical data. For each type (year, recent, modified), it records the sync via SyncMetadata. Prior to syncing the yearly data, the module checks to see if the yearly data has been synced by introspecting the graph for the SyncMetadata for the yearly data. It only syncs years that have not yet been recorded. This allows the module to normally only sync the modified and recent feeds.


ramonpetgrave64:
This is very interesting. 

Aside, the lack of a cleanup job is concerning, and we'll need to carefully consider the possibility of the graph growing too large too quickly.

ryan-lane:
Well, for this there's an upper bound on growth. Nodes are only added if
there's new CVEs. Also, there's no cleanup because CVEs are never deleted
from the source.

ryan-lane:
Not stale

achantavy (commented):
Can you please update 'supported platforms' under README.md too?

ryan-lane:
@achantavy done!

achantavy (approved):

PR #798, opened 2022-04-06
Add sync metadata to github, k8s and pagerduty modules

ryan-lane:
This change adds sync metadata to github, k8s, and pagerduty modules, for the top-level resources. Pagerduty doesn't have a top-level resource, because pagerduty doesn't expose it, so the top-level resource is considered the module, with `module` as the group_id.


ramonpetgrave64 (approved):

ramonpetgrave64:
I'm happy to see this being used

Issue #793, opened 2022-03-31
EC2 Snapshots gets lots of stuff, takes very long


danielsaporo:
I'm trying Cartography on an AWS account that is practically empty (just a few users), and noticed that it takes a while to run. What really stands out is that it takes **over 9 minutes** to get EBS snapshot data. And this account does not have (and never had) any EBS volumes in any regions.

Yet, when I debug `get_snapshots()` in `snapshots.py`, I see that `client.get_paginator('describe_snapshots')` returns many thousands of records despite the fact that there is no snapshot data in any region visible in the AWS Console.

So, what is [describe_snapshots](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_snapshots) returning exactly, and do we need all that?

ryan-lane:
Ah. It looks like that function returns all available snapshots. I think what we probably want here is to filter to snapshots owned by the current account. @achantavy @ramonpetgrave64, thoughts?

danielsaporo:
> Ah. It looks like that function returns all available snapshots. I think what we probably want here is to filter to snapshots owned by the current account.

Why does `describe_snapshots` return a lot of snapshots that don't belong to the current account? From the description I see:

> Describes the specified EBS snapshots available to you or all of the EBS snapshots available to you.
> 
> The snapshots available to you include public snapshots, private snapshots that you own, and private snapshots owned by other Amazon Web Services accounts for which you have explicit create volume permissions.

However I don't really understand this. I'm using a new AWS account which hasn't been granted access to any other account. I just hope AWS isn't returning a lot of stuff from other accounts that it shouldn't.

ryan-lane:
@danielsaporo There's public EBS snapshots for public datasets, and there's a lot of them. This is similar to public AMIs. See, for example, how I changed that to only pull AMIs from owned accounts, but then introspected the graph to find other images that should be fetched: https://github.com/lyft/cartography/blob/master/cartography/intel/aws/ec2/images.py#L17-L59

ramonpetgrave64:
Fixed in #799

danielsaporo:
Thanks for fixing this. What is the process to upgrade to a newer version of Cartography that includes this fix? Do I need to wait for some kind of release to happen first?

PR #799, opened 2022-04-06
filter ebs snapshots to just the current account (#793)

amlweems:
The `describe_snapshots` call returns approximately 40k public snapshots which creates extra nodes / relationships and generally causes performances issues with the graph. This PR filters snapshots by `OwnerId` `"self". This parameter is documented [here](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-snapshots.html) and returns snapshots owned or explicitly granted to the caller's account.

### DIFF
--- a/cartography/intel/aws/ec2/snapshots.py
+++ b/cartography/intel/aws/ec2/snapshots.py
@@ -32,7 +32,7 @@ def get_snapshots(boto3_session: boto3.session.Session, region: str, in_use_snap
     for page in paginator.paginate(OwnerIds=['self']):
         snapshots.extend(page['Snapshots'])
 
-    # fetch in-use snapshots not in self-owned snapshots
+    # fetch in-use snapshots not in self_owned snapshots
     self_owned_snapshot_ids = {s['SnapshotId'] for s in snapshots}
     other_snapshot_ids = set(in_use_snapshot_ids) - self_owned_snapshot_ids
     if other_snapshot_ids:
--- a/cartography/intel/aws/ec2/snapshots.py
+++ b/cartography/intel/aws/ec2/snapshots.py
@@ -25,17 +25,19 @@ def get_snapshots_in_use(neo4j_session: neo4j.Session, region: str, current_aws_
 
 @timeit
 @aws_handle_regions
-def get_snapshots(boto3_session: boto3.session.Session, region: str, snapshotIds: List[str]) -> List[Dict]:
+def get_snapshots(boto3_session: boto3.session.Session, region: str, in_use_snapshot_ids: List[str]) -> List[Dict]:
     client = boto3_session.client('ec2', region_name=region)
     paginator = client.get_paginator('describe_snapshots')
     snapshots: List[Dict] = []
     for page in paginator.paginate(OwnerIds=['self']):
         snapshots.extend(page['Snapshots'])
 
-    if not snapshotIds:
-        return snapshots
-    for page in paginator.paginate(SnapshotIds=snapshotIds):
-        snapshots.extend(page['Snapshots'])
+    # fetch in-use snapshots not in self-owned snapshots
+    self_owned_snapshot_ids = {s['SnapshotId'] for s in snapshots}
+    other_snapshot_ids = set(in_use_snapshot_ids) - self_owned_snapshot_ids
+    if other_snapshot_ids:
+        for page in paginator.paginate(SnapshotIds=list(other_snapshot_ids)):
+            snapshots.extend(page['Snapshots'])
     return snapshots
 
 
--- a/cartography/intel/aws/ec2/snapshots.py
+++ b/cartography/intel/aws/ec2/snapshots.py
@@ -13,12 +13,28 @@ logger = logging.getLogger(__name__)
 
 
 @timeit
+def get_snapshots_in_use(neo4j_session: neo4j.Session, region: str, current_aws_account_id: str) -> List[str]:
+    query = """
+    MATCH (:AWSAccount{id: {AWS_ACCOUNT_ID}})-[:RESOURCE]->(v:EBSVolume)
+    WHERE v.region = {Region}
+    RETURN v.snapshotid as snapshot
+    """
+    results = neo4j_session.run(query, AWS_ACCOUNT_ID=current_aws_account_id, Region=region)
+    return [r['snapshot'] for r in results]
+
+
+@timeit
 @aws_handle_regions
-def get_snapshots(boto3_session: boto3.session.Session, region: str) -> List[Dict]:
+def get_snapshots(boto3_session: boto3.session.Session, region: str, snapshotIds: List[str]) -> List[Dict]:
     client = boto3_session.client('ec2', region_name=region)
     paginator = client.get_paginator('describe_snapshots')
     snapshots: List[Dict] = []
     for page in paginator.paginate(OwnerIds=['self']):
+        snapshots.extend(page['Snapshots'])
+
+    if not snapshotIds:
+        return snapshots
+    for page in paginator.paginate(SnapshotIds=snapshotIds):
         snapshots.extend(page['Snapshots'])
     return snapshots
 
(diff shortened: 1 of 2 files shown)


ryan-lane:
I agree with this. It may be good as a followup later to also inspect the graph for EBS snapshots that aren't owned by the account, but are in use, to have fuller coverage.

ramonpetgrave64:
Yes, it may be a better idea to filter to also include EBSVolumes -> EBSSnapshots in use by owned EC2Instances
https://github.com/lyft/cartography/blame/21af12527c70c1467cba56d3b44448ff277fd82f/cartography/intel/aws/ec2/instances.py#L380

@kedarghule

kedarghule:
> @amlweems Thank you!
> 
> @ryan-lane Yes, it may be a better idea to filter to also include EBSVolumes -> EBSSnapshots in use by owned EC2Instances https://github.com/lyft/cartography/blame/21af12527c70c1467cba56d3b44448ff277fd82f/cartography/intel/aws/ec2/instances.py#L380
> 
> @kedarghule

Yes, this looks good! I agree with @ryan-lane that we could also have a future PR in which we can populate EBS Snapshots not owned by the account but in use.

I'd like @mpurusottamc to take a look at this as well since I had opened the PR for EBS Snapshots during my time at Cloudanix. 

Once the CLA check is okay, we can merge!

ramonpetgrave64:
@kedarghule I disagree, in that this PR should include both owned snapshots and snapshots in use by owned instances. This PR with just the filter for owned snapshots, I think, would be too much of a regression.

kedarghule:
> @kedarghule I disagree, in that this PR should include both owned snapshots and snapshots in use by owned instances. This PR with just the filter for owned snapshots, I think, would be too much of a regression.

Oh okay. I see your reasoning and I am inclined to agree. I do agree that we should show the snapshots in use by owned instances too.

mpurusottamc:
@kedarghule This change makes a lot of sense. It was a miss on our side during the initial checkin. Thanks @amlweems for correcting this.

ramonpetgrave64:
@amlweems Are you able to update this PR to include both owned snapshots and snapshots in use by owned instances?

amlweems:
@ramonpetgrave64 Sure thing! I've just pushed a commit that adds a check for snapshots in use and fetches those in addition to any self owned snapshots.

cartography/intel/aws/ec2/snapshots.py
```diff
@@ -12,13 +12,29 @@
...
     snapshots: List[Dict] = []
-    for page in paginator.paginate():
+    for page in paginator.paginate(OwnerIds=['self']):
+        snapshots.extend(page['Snapshots'])
+
+    if not snapshotIds:
+        return snapshots
+    for page in paginator.paginate(SnapshotIds=snapshotIds):
         snapshots.extend(page['Snapshots'])
     return snapshots
```
ramonpetgrave64, line 39: Change `snapshotIds` to `in_use_snapshot_ids`.

Change L35-39 to something like
```
...
self_owned_snapshot_ids = set(s['SnapshitId'] for s in snapshots)

# reduce a a smaller list of unfetched snapshots
other_snapshot_ids = set(in_use_snapshot_ids) - self_owned_snapshot_ids
if other_snapshot_ids:
    for page in paginator.paginate(SnapshotIds=list(other_snapshot_ids)):
        snapshots.extend(page['Snapshots']
return snapshots
```

cartography/intel/aws/ec2/snapshots.py
```diff
@@ -12,13 +12,29 @@
 logger = logging.getLogger(__name__)
 
 
+@timeit
+def get_snapshots_in_use(neo4j_session: neo4j.Session, region: str, current_aws_account_id: str) -> List[str]:
```
ramonpetgrave64, line 16: Please try and add an integration test for this.

ramonpetgrave64 (changes requested):

ramonpetgrave64:
> @ramonpetgrave64 Sure thing! I've just pushed a commit that adds a check for snapshots in use and fetches those in addition to any self owned snapshots.

Thank you! 
Just a small change needed, since it seems your current fix would fetch the same snapshots twice.
Also, please sign the CLA.

cartography/intel/aws/ec2/snapshots.py
amlweems: 👍 Added!

cartography/intel/aws/ec2/snapshots.py
amlweems: 👍 Updated!

amlweems:
@ramonpetgrave64 Thanks for the review! I've made both changes and signed the CLA.

tests/integration/cartography/intel/aws/ec2/test_ec2_snapshots.py
```diff
@@ -1,11 +1,46 @@
...
+    cartography.intel.aws.ec2.volumes.load_volumes(
+        neo4j_session,
+        transformed_data,
+        TEST_REGION,
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+    )
+
+    # Assert
+    expected_snapshots = {
```
ramonpetgrave64, line 35: non-blocker: I'd rather this be a list, since the function does return a list.

cartography/intel/aws/ec2/snapshots.py
```diff
@@ -12,14 +12,32 @@
...
-def get_snapshots(boto3_session: boto3.session.Session, region: str) -> List[Dict]:
+def get_snapshots(boto3_session: boto3.session.Session, region: str, in_use_snapshot_ids: List[str]) -> List[Dict]:
     client = boto3_session.client('ec2', region_name=region)
     paginator = client.get_paginator('describe_snapshots')
     snapshots: List[Dict] = []
-    for page in paginator.paginate():
+    for page in paginator.paginate(OwnerIds=['self']):
         snapshots.extend(page['Snapshots'])
+
+    # fetch in-use snapshots not in self-owned snapshots
```
ramonpetgrave64, line 35: nonblocker: self_owned, not self-owned

tests/integration/cartography/intel/aws/ec2/test_ec2_snapshots.py
```diff
@@ -1,11 +1,46 @@
...
+        "sn-01", "sn-02",
+    }
+
+    snapshots = cartography.intel.aws.ec2.snapshots.get_snapshots_in_use(neo4j_session, TEST_REGION, TEST_ACCOUNT_ID)
```
ramonpetgrave64, line 39: non blocker: actual_snapshots

ramonpetgrave64 (approved):
Looks good. A few smaller comments, but nothing to block. I'll approve for now. Do you want to address my comments before I merge?

ramonpetgrave64:
Also please update your branch with the latest from the main branch.

amlweems:
@ramonpetgrave64 👍 fixed the non-blockers and updated with main

tests/integration/cartography/intel/aws/ec2/test_ec2_snapshots.py
```diff
@@ -1,11 +1,46 @@
...
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+    )
+
+    # Assert
+    expected_snapshots = [
+        "sn-01", "sn-02",
+    ]
+
+    actual_snapshots = cartography.intel.aws.ec2.snapshots.get_snapshots_in_use(neo4j_session, TEST_REGION, TEST_ACCOUNT_ID)
```
ramonpetgrave64, line 39: One last thing for the linter
```suggestion
    actual_snapshots = cartography.intel.aws.ec2.snapshots.get_snapshots_in_use(
        neo4j_session, 
        TEST_REGION, TEST_ACCOUNT_ID
    )
```

ramonpetgrave64 (changes requested):

ramonpetgrave64 (approved):

ramonpetgrave64:
@amlweems Thank you again! Ideas, suggestions, and contributions are always welcome.

Issue #734, opened 2021-12-13
ServiceUnavailable in resourcegroupstaggingapi cleanup job


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

The resourcegroupstaggingapi is somehow failing to run even just one cleanup statement. I've observed it hanging for 2 hours and then finally giving up with a ServiceUnavailable exception.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

This is probably specific to our own environment but I imagine a large enough deployment of cartography could run into this issue too.

*Logs*:
> If applicable, copy and paste your console log with the failing stack trace.

```
Unhandled error while executing statement in job 'cleanup AWS Tags': Failed to read from defunct connection Address(host=$HOSTNAME, port=7687) (Address(host=$IP, port=7687))
```

Full stack trace

```
Traceback (most recent call last):
 [...]
  File "/path/to/cartography/intel/aws/__init__.py", line 139, in _sync_multiple_accounts
    _sync_one_account(
  File "/path/to/cartography/intel/aws/__init__.py", line 72, in _sync_one_account
    RESOURCE_FUNCTIONS['resourcegroupstaggingapi'](**sync_args)
  File "/path/to/cartography/util.py", line 68, in timed
    result = method(*args, **kwargs)
  File "/path/to/cartography/intel/aws/resourcegroupstaggingapi.py", line 194, in sync
    cleanup(neo4j_session, common_job_parameters)
  File "/path/to/cartography/util.py", line 68, in timed
    result = method(*args, **kwargs)
  File "/path/to/cartography/intel/aws/resourcegroupstaggingapi.py", line 175, in cleanup
    run_cleanup_job('aws_import_tags_cleanup.json', neo4j_session, common_job_parameters)
  File "/path/to/cartography/util.py", line 38, in run_cleanup_job
    GraphJob.run_from_json(
  File "/path/to/cartography/graph/job.py", line 111, in run_from_json
    job.run(neo4j_session)
  File "/path/to/cartography/graph/job.py", line 55, in run
    stm.run(neo4j_session)
  File "/path/to/cartography/graph/statement.py", line 69, in run
    self._run_iterative(session)
  File "/path/to/cartography/graph/statement.py", line 119, in _run_iterative
    result: neo4j.StatementResult = session.write_transaction(self._run_noniterative)
[...]

PR #812, opened 2022-04-14
Addresses #734: AWS: load_tags in batches

ramonpetgrave64:
Currently, we load AWSTag nodes in somewhat of a batched manner, grouped by resource types. But EC2Instances, for example, would have all of their tags loaded at once in one `write_transaction`.

Given the potential for AWSTag nodes to be highly connected, it is sensible to load these nodes in smaller batches, for a more stable sync of AWSTag nodes.

For example, if your organization has the practice of tagging resources by human owner, a single AWSTag node could be connected to thousands of other nodes.

This PR will load tags in batches of 100.

This may help is issue in #734


cartography/util.py
```diff
@@ -181,3 +183,19 @@ def dict_date_to_epoch(obj: Dict, key: str) -> Optional[int]:
 
 def camel_to_snake(name: str) -> str:
     return re.sub(r'(?<!^)(?=[A-Z])', '_', name).lower()
+
+
+def batch(items: Iterable, size=1000) -> List[List]:
+    '''
+    Takes an Iterable of items and returns a list of lists of the same items,
+     batched into chunks of the proivided `size`.
```
achantavy, line 191:
```suggestion
     batched into chunks of the provided `size`.
```

achantavy (approved):

Issue #808, opened 2022-04-13
Error while ingesting elastic_ip_addresses


marco-lancini:
`neobolt.exceptions.ClientError: Cannot merge node using null property value for id` is triggered while trying to ingest Elastic IPs, even if none were found

*Logs*:

```
INFO:cartography.intel.aws.ec2.elastic_ip_addresses:Loading 0 Elastic IP Addresses in ca-central-1.
ERROR:cartography.sync:Unhandled exception during sync stage 'aws'
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/cartography/sync.py", line 74, in run
    stage_func(neo4j_session, config)
  File "/usr/local/lib/python3.9/site-packages/cartography/util.py", line 114, in timed
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/cartography/intel/aws/__init__.py", line 220, in start_aws_ingestion
    _sync_multiple_accounts(
  File "/usr/local/lib/python3.9/site-packages/cartography/intel/aws/__init__.py", line 157, in _sync_multiple_accounts
    _sync_one_account(
  File "/usr/local/lib/python3.9/site-packages/cartography/intel/aws/__init__.py", line 60, in _sync_one_account
    RESOURCE_FUNCTIONS[func_name](**sync_args)
  File "/usr/local/lib/python3.9/site-packages/cartography/util.py", line 114, in timed
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/cartography/intel/aws/ec2/elastic_ip_addresses.py", line 99, in sync_elastic_ip_addresses
    load_elastic_ip_addresses(neo4j_session, addresses, region, current_aws_account_id, update_tag)
  File "/usr/local/lib/python3.9/site-packages/cartography/util.py", line 114, in timed
    return method(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/cartography/intel/aws/ec2/elastic_ip_addresses.py", line 73, in load_elastic_ip_addresses
    neo4j_session.run(
  File "/usr/local/lib/python3.9/site-packages/neo4j/__init__.py", line 503, in run
    self._connection.fetch()
  File "/usr/local/lib/python3.9/site-packages/neobolt/direct.py", line 419, in fetch
    return self._fetch()
  File "/usr/local/lib/python3.9/site-packages/neobolt/direct.py", line 461, in _fetch
    response.on_failure(summary_metadata or {})
[...]

PR #824, opened 2022-04-21
Fix 808: Error while ingesting elastic_ip_addresses

marco-lancini:
Fix for #808


ramonpetgrave64 (changes requested):
Look's good. But we also need to update the schema docs. 
https://github.com/lyft/cartography/blob/master/docs/root/modules/aws/schema.md#elasticipaddress

marco-lancini:
> Look's good. But we also need to update the schema docs. https://github.com/lyft/cartography/blob/master/docs/root/modules/aws/schema.md#elasticipaddress

Thanks for the pointer!
Since the release of the new docs website I couldn't find where the schema doc went :D

ramonpetgrave64 (approved):

ramonpetgrave64:
@marco-lancini Yes, I had some trouble finding it too. at ryan-lane

Issue #820, opened 2022-04-20
Analysis and cleanup jobs are inconsistently placed


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

By default, the [final](https://github.com/lyft/cartography/blob/f28fcd7436c17d9cbf32a31161f1d592d1d29f2e/cartography/sync.py#L173) cartography sync stage runs all analysis jobs located in the cartography/data/jobs/analysis folder. I noticed that there are cases where we call `run_analysis_job()` out of band such as when syncing [iaminstanceprofiles](https://github.com/lyft/cartography/blob/f28fcd7436c17d9cbf32a31161f1d592d1d29f2e/cartography/intel/aws/__init__.py#L77-L81) and analyzing [lambda-to-ecr relationships](https://github.com/lyft/cartography/blob/f28fcd7436c17d9cbf32a31161f1d592d1d29f2e/cartography/intel/aws/__init__.py#L83-L87), in [S3 acls](https://github.com/lyft/cartography/blob/f28fcd7436c17d9cbf32a31161f1d592d1d29f2e/cartography/intel/aws/s3.py#L229-L234), everything [here](https://github.com/lyft/cartography/blob/f28fcd7436c17d9cbf32a31161f1d592d1d29f2e/cartography/intel/aws/__init__.py#L228-L244), and probably others.

The problem with these one-off calls of analysis jobs is that by default they will all be run a second time when we reach the final sync stage. This is wasted work and adds time to the sync especially on a large graph.

To summarize,

1. We need to run analysis on the entire graph. We currently do this in the default final sync stage where all jobs in cartography/data/jobs/analysis are run.
2. We need to run analysis on a segment of the graph, for example a single AWS account or single GCP project. We currently do this with one-off calls to `run_analysis_job()`.
3. We need to avoid running any of these jobs twice. This could be accomplished by splitting cartography/data/jobs/analysis into separate folders, or other ideas.

*Please complete the following information:*:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]

0.56.0

ramonpetgrave64:
From @ryan-lane https://github.com/lyft/cartography/pull/826#issuecomment-1109419468
> Could we separate out the analysis jobs that are intended to run along with
> the code, from the ones that are intended to connect nodes between modules?
[...]

PR #826, opened 2022-04-25
#820: Remove redundant calls to analysis jobs

achantavy:
As seen in #820, the final sync of the default sync object already runs all analysis jobs that are placed in the cartography/data/jobs/analysis folder, but there are many places in the code where we call `run_analysis_job()` in a one-off way. This PR removes these redundant calls and updates the docs to explain the correct placement.


achantavy:
cc: @ryan-lane

cartography/intel/aws/__init__.py
```diff
@@ -74,18 +73,6 @@ def _sync_one_account(
...
         RESOURCE_FUNCTIONS['resourcegroupstaggingapi'](**sync_args)
 
-    run_analysis_job(
-        'aws_ec2_iaminstanceprofile.json',
-        neo4j_session,
-        common_job_parameters,
-    )
-
-    run_analysis_job(
-        'aws_lambda_ecr.json',
```
ramonpetgrave64, line 84: This one is already a problem for when we want to only sync one account, but we have data from other accounts present in the graph. In such a case, this job would remove data from the other accounts, probably unintended by the user.

Yes, let's remove this invocation from here, but update the job to be more like the `aws_eks_asset_exposure.json`, where we would remove all the `:HAS` relationships, regarldess of `lastupdated`, and then create new `:HAS` relationships.

achantavy: > Yes, let's remove this invocation from here, but update the job to be more like the aws_eks_asset_exposure.json, where we would remove all the :HAS relationships, regarldess of lastupdated, and then create new :HAS relationships.

Consistency of "delete all then recreate" vs "delete only stale" is a problem, but this change is out of scope for this PR imo. What do you think, do you think this is blocking?

ryan-lane:
I thought the analysis jobs that were in the code were expected to be run with the code (and are always run), and that folks set the `--analysis-job-directory` option to specify analysis jobs that are custom and not included with the code.

If we remove the analysis jobs from the code, and folks are using the directory option to point to another location for their own custom analysis jobs, will these analysis jobs ever run?

achantavy:
@ryan-lane Yeah, all are correct points, thanks for pointing out.

I'm trying to solve the problem where some of these analysis jobs (e.g. iaminstanceprofiles) are incorrectly scoped in that they update assets outside of the current sub-resource. I'll revise and keep all the points we've discussed in mind.

achantavy:
Alright I thought about this some more. None of the AWS analysis jobs that I've removed from `sync_one_account()` scope themselves to a single account with an `AWS_ID` parameter, so for each AWS account they will run against _all_ AWS assets, which is probably not an intended behavior.

To maintain the idea that some analysis jobs are built-in and should always be run, we should instead move these at the end of `sync_multiple_accounts()` so that we get the intended behavior.

ryan-lane:
@achantavy can we change them to be scoped to the single account? I'd really like to be able to split the sync of my accounts into numerous jobs, so I can run them in parallel. I was planning on running them async from my custom script, but the sync_multiple _accounts path is currently blocking me from doing that

achantavy:
To change them to be scoped to one account we will need to update the analysis job folder files here: https://github.com/lyft/cartography/tree/master/cartography/data/jobs/analysis.
Currently the default cartography sync object runs all files in this folder as its final stage

ryan-lane:
Could we separate out the analysis jobs that are intended to run along with
the code, from the ones that are intended to connect nodes between modules?
It may be good to put analysis and cleanup files for modules directly into
their modules.

To be honest I wasn't aware that the folder was run at the end of every run
and for anyone that changes the value of that option it's not how it runs.

ramonpetgrave64:
@ryan-lane I like that idea, and let's continue the conversation in #820. For now, I think we can merge this change.

ramonpetgrave64 (approved):
We will continue the discussion in #820 and address the comments in a later PR.

ryan-lane:
@ramonpetgrave64 if you merge this in, and I have the `--analysis-job-directory` option set (which I do, since I have custom analysis jobs), if this gets merged in, these analysis jobs will no longer run for me, right?

achantavy:
Closing this per last week's discussion in OSS sync 4/28.

PR #833, opened 2022-04-28
AWS: Tags: load tags by account id

ramonpetgrave64:
Previously, we would incorrectly map tags to resources in all AWS accounts, which is incorrect behavior. This PR ensures that we map to resources only in the current AWS account being synced.


achantavy (approved):
Looks good to me, thanks for catching and fixing!

PR #836, opened 2022-05-02
Extension of support for current Azure services

ganeshjagdale9545:
For Azure, support for these additional services - Function App, Network, Resource Group, Tag, IAM, Key Vaults, AKS


achantavy:
Hey thanks for the contribution, like we mentioned in https://github.com/lyft/cartography/pull/764 though can you please split this to one module per PR?

jychp:
This PR has since been split into multiple successive PRs and implemented in Cartography

Issue #842, opened 2022-05-12
Unhandled IllegalLocationConstraintException in s3


krisek:
*Title*: *Unhandled IllegalLocationConstraintException in s3*

*Description*:
> Sync process crashing with

```
cartography-loader-adhoc--1-9gbxc cartography ERROR:cartography.sync:Unhandled exception during sync stage 'aws'                                                                      
cartography-loader-adhoc--1-9gbxc cartography Traceback (most recent call last):                                                                                                      
cartography-loader-adhoc--1-9gbxc cartography   File "/usr/local/lib/python3.10/site-packages/cartography/sync.py", line 74, in run                                                   
cartography-loader-adhoc--1-9gbxc cartography     stage_func(neo4j_session, config)                                                                                                   
cartography-loader-adhoc--1-9gbxc cartography   File "/usr/local/lib/python3.10/site-packages/cartography/util.py", line 117, in timed                                                
cartography-loader-adhoc--1-9gbxc cartography     return method(*args, **kwargs)                                                                                                      
cartography-loader-adhoc--1-9gbxc cartography   File "/usr/local/lib/python3.10/site-packages/cartography/intel/aws/__init__.py", line 220, in start_aws_ingestion                    
cartography-loader-adhoc--1-9gbxc cartography     _sync_multiple_accounts(                                                                                                            
cartography-loader-adhoc--1-9gbxc cartography   File "/usr/local/lib/python3.10/site-packages/cartography/intel/aws/__init__.py", line 157, in _sync_multiple_accounts                
cartography-loader-adhoc--1-9gbxc cartography     _sync_one_account(                                                                                                                  
cartography-loader-adhoc--1-9gbxc cartography   File "/usr/local/lib/python3.10/site-packages/cartography/intel/aws/__init__.py", line 60, in _sync_one_account                       
cartography-loader-adhoc--1-9gbxc cartography     RESOURCE_FUNCTIONS[func_name](**sync_args)                                                                                          
cartography-loader-adhoc--1-9gbxc cartography   File "/usr/local/lib/python3.10/site-packages/cartography/util.py", line 117, in timed                                                
cartography-loader-adhoc--1-9gbxc cartography     return method(*args, **kwargs)                                                                                                      
cartography-loader-adhoc--1-9gbxc cartography   File "/usr/local/lib/python3.10/site-packages/cartography/intel/aws/s3.py", line 649, in sync                                         
[...]

PR #843, opened 2022-05-13
#842 s3 exception fix

krisek:
This is a rudimentary fix for #842, unfortunately I have no time to update tests or tidy up more, feel free to take it or leave it.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -128,6 +128,14 @@ class CLI:
             ),
         )
         parser.add_argument(
+            '--aws-best-effort-mode',
+            action='store_true',
+            help=(
+                'Enable AWS sync best effort mode when syncing AWS accounts. This will allow cartography to continue '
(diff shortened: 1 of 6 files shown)


krisek:
As usual, some tests fail for files which I didn't touch at all... it's all a mystery.

ryan-lane:
@krisek these test failures are related to your changes, and are lint failures that can be automatically fixed by running pre-commit locally. Can you please run that?

achantavy:
Hey @krisek, good to hear from you again. You can use `make test_lint` locally before pushing to catch this stuff before the CI complains. Looks like it doesn't like some paren stuff: https://github.com/lyft/cartography/runs/6430113666?check_suite_focus=true

achantavy:
Per https://github.com/lyft/cartography/issues/842#issuecomment-1125028860, I think it makes more sense to add IllegalLocationConstraints to common exceptions. If the issue persists then we can use this PR instead.

PR #845, opened 2022-05-17
Undo change making aws_handle_regions polymorphic

aneeshusa:
This reverts commit 9b33276124cd472b75a52861cb47a8df48319602,
which updated `aws_handle_regions` to be polymorphic in return type
by accepting a `default_return_value` of arbitrary shape.
This is not straightforwardly statically typeble, so undo this.

Instead, update the ECS module (which used this functionality)
to separate ECS cluster ARN fetching and ECS cluster info fetching,
as they don't need to be handled in the same function.

Keep most of the unit tests added in the original change,
but remove the one about support custom `default_return_value`.

Pre-requisite for https://github.com/lyft/cartography/pull/790.


ryan-lane (dismissed):

achantavy (approved):
LGTM, thanks for the fix

olivia-hong (approved):

PR #846, opened 2022-05-23
Add AWSOrganization node and relationship.

shinya-murakami-paypay:
- Added import of AWSOrganizations.
- Fixed AWSAccounts auto-discovery to capture detailed data other than ID and Name.

### DIFF
--- a/cartography/data/jobs/analysis/aws_account_to_organization_link.json
+++ b/cartography/data/jobs/analysis/aws_account_to_organization_link.json
@@ -1,0 +1,13 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (ao:AWSOrganization), (aa:AWSAccount) WHERE ao.master_account_id = aa.id MERGE (aa)-[r:MASTER]->(ao) ON CREATE SET r.firstseen = {UPDATE_TAG} SET r.lastupdated = {UPDATE_TAG}",
+      "iterative": false
+    },
+    {
+      "query": "MATCH (:AWSAccount)-[r:MASTER]->(:AWSOrganization) WHERE r.lastupdated <> {UPDATE_TAG} DELETE (r)",
+      "iterative": false
+    }
+  ],
+  "name": "Connect AWSOrganization and AWSAccount"
+}
--- a/cartography/data/jobs/cleanup/aws_import_organization_cleanup.json
+++ b/cartography/data/jobs/cleanup/aws_import_organization_cleanup.json
@@ -1,0 +1,15 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (o:AWSOrganization) WHERE o.lastupdated <> {UPDATE_TAG} WITH o LIMIT {LIMIT_SIZE} DETACH DELETE (o) return COUNT(*) as TotalCompleted",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (:AWSAccount)-[r:BELONG_TO]->(:AWSOrganization) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DETACH DELETE (r) return COUNT(*) as TotalCompleted",
+      "iterative": true,
+      "iterationsize": 100
+    }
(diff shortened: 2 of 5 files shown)


ramonpetgrave64:
@shinya-murakami-paypay Thanks! Please also sign the CLA when you get the chance.

shinya-murakami-paypay:
@ramonpetgrave64 I signed and rechecked the CLA.

achantavy:
Mind re-updating with master branch again? Sorry

cartography/intel/aws/organizations.py
```diff
@@ -86,6 +91,55 @@ def get_aws_accounts_from_botocore_config(boto3_session: boto3.session.Session)
...
+    for page in paginator.paginate():
+        accounts.extend(page['Accounts'])
+
+    # Filter out every account which is not in the ACTIVE status
+    # and select only the Id and Name fields
+    filtered_accounts: List[Dict] = [x for x in accounts if x['Status'] == 'ACTIVE']
+    return filtered_accounts
+
+
+def load_aws_organization(
```
achantavy, line 119: Thank you for the PR, are you able to please add tests for the load functions?

cartography/data/jobs/cleanup/aws_import_organization_cleanup.json
```diff
@@ -0,0 +1,15 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (o:AWSOrganization) WHERE o.lastupdated <> {UPDATE_TAG} WITH o LIMIT {LIMIT_SIZE} DETACH DELETE (o) return COUNT(*) as TotalCompleted",
```
achantavy, line 4: Can you also please update the schema docs?

cartography/data/jobs/cleanup/aws_import_organization_cleanup.json
```diff
@@ -0,0 +1,15 @@
...
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (:AWSAccount)-[r:BELONG_TO]->(:AWSOrganization) WHERE r.lastupdated <> {UPDATE_TAG} WITH r LIMIT {LIMIT_SIZE} DETACH DELETE (r) return COUNT(*) as TotalCompleted",
```
achantavy, line 9: Nit: The convention for resource hierarchy seems to be defined with `:RESOURCE`, so `(:AWSOrganization)-[:RESOURCE]->(:AWSAccount)-[:RESOURCE]->...`.

I admittedly don't have a very strong opinion on this but thought I'd bring it up for consideration.

cartography/intel/aws/organizations.py
```diff
@@ -86,6 +91,55 @@ def get_aws_accounts_from_botocore_config(boto3_session: boto3.session.Session)
     return d
 
 
+def describe_organization(boto3_session: boto3.session.Session) -> Dict:
```
achantavy, line 94: The convention is to name this with `get`, can we do something like this instead?

```suggestion
def get_organizations(boto3_session: boto3.session.Session) -> Dict:
```

cartography/data/jobs/analysis/aws_account_to_organization_link.json
```diff
@@ -0,0 +1,13 @@
+{
+  "statements": [
```
achantavy, line 2: Rather than do this as part of an analysis job, could you move it to the Python code? My thinking is that analysis jobs are sort of "extras" with cartography and things like the relationship between an organization and account seems foundational, so we should have it in code because not all users of cartography will opt to run analysis jobs.

I'm open to thoughts here though.

achantavy (changes requested):

cartography/data/jobs/analysis/aws_account_to_organization_link.json
mike-traff-paypay: @achantavy sorry we haven't been following this up.  We'll address the issues by the end of the week.

ramonpetgrave64:
not stale

PR #848, opened 2022-05-24
Bump mypy and add missing __init__.py files

aneeshusa:
Bump mypy to the latest version.

The `__init__.py` files are required for mypy's per-module configuration;
otherwise those modules aren't recognized due to namespace package support not being on.


aneeshusa:
Split out of https://github.com/lyft/cartography/pull/790 for easier review.

yabanci:
This linked doc suggests that `__init__.py` is not required when the `--namespace-packages` flag is used. Is the idea to avoid using it? I always add an `__init__.py` because I assume that is required to import modules but is there any reason not to include it?

yabanci (approved):

PR #850, opened 2022-05-31
Fix Cartography 0.59.0 release

VJftw:
At the moment, with `cartography==0.59.0` running `cartography -h` results in
```
Traceback (most recent call last):
  File "***/bin/cartography", line 8, in <module>
    sys.exit(main())
  File "***/cartography/cli.py", line 504, in main
    default_sync = cartography.sync.build_default_sync()
  File "***/cartography/sync.py", line 166, in build_default_sync
    ('crowdstrike', cartography.intel.crowdstrike.start_crowdstrike_ingestion),
AttributeError: module 'cartography.intel' has no attribute 'crowdstrike'
```
as it has not been imported in this `sync.py`

PR #851, opened 2022-05-31
Publish a dist docker image

VJftw:
Hi again!

We think having an official Cartography docker image would really make the process of upgrading our version of Cartography much easier :)

Here's PR which builds a docker image and publishes it to GitHub Container Registry upon release. Hopefully, it helps :)

Below are some examples which demonstrate this new functionality:
 - Verifying that the Cartography docker image builds in the _Test Suite_ workflow: https://github.com/VJftw/cartography/runs/6679405016?check_suite_focus=true
 - GitHub packages link to published Cartography docker images: https://github.com/VJftw/cartography/pkgs/container/cartography
 - Building and publishing a Cartography docker image upon release creation: https://github.com/VJftw/cartography/actions/runs/2418077517
 - Manually building and publishing a Cartography docker image for a branch via workflow dispatch: https://github.com/VJftw/cartography/actions/runs/2418049380

~~Note: This includes changes from #850 so that a working cartography binary can actually be built in the docker image.~~


.github/workflows/publish-to-ghcr.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
achantavy, line 29: I'm not super familiar with github actions - what does this do?

dist.Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM python:3-alpine as builder
```
achantavy, line 1: What is the difference between using this file versus our current Dockerfile? https://github.com/lyft/cartography/blob/master/Dockerfile

achantavy (changes requested):
Thanks for the PR. If having a built cartography image in github CR will help your deployment, I'm supportive :). Left a few questions on why the current Dockerfile is not used, as well as on some mechanics of github actions secrets.

.github/workflows/publish-to-ghcr.yml
VJftw: Good question! I've added a few inline code comments to try and explain this. I've also defined more [explicit permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#assigning-permissions-to-a-specific-job) to make it hopefully a bit secure and clearer as [this token has quite far reaching permissions by default](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) 😄

dist.Dockerfile
VJftw: I've tried to create a _securer_, _productionised_ Docker image here following the principles of [distroless images](https://github.com/GoogleContainerTools/distroless#why-should-i-use-distroless-images), and [some best practices published by hexops](https://github.com/hexops/dockerfile). I think this is a useful distinction to have between the current `./Dockerfile` which was initially implemented as a testing/development image in #606. 

The differences between this `./dist.Dockerfile` and `./Dockerfile` are:
 
1. We build a self-contained cartography binary using `pyinstaller`. Dependencies like `psutil` and `cryptography` need to be compiled which is why we install build tools like `gcc` and development headers for `openssl` etc..
2. We copy this self-contained cartography binary into a fresh [Alpine Linux](https://github.com/docker-library/docs/tree/master/alpine#what-is-alpine-linux) container which is commonly used as a minimal base OS. This results in a much smaller Docker image as unneeded compile-time dependencies aren't installed. Ideally, we'd love to use `scratch` but Python doesn't offer a way to statically compile into a single binary like Go or Rust do. Our self-contained binary from step 1. appears to depend on having `sh` and the same libc bindings that it was built with, so we're using Alpine Linux to both build and run cartography.
3. We configure the Docker container to run as a UID and GID that likely doesn't overlap with host users.

achantavy:
cc: @marco-lancini @ryan-lane @mpurusottamc @ramonpetgrave64

dist.Dockerfile
ryan-lane: It would be ideal to not use alpine. Its security is actually worse than using a slim python (ubuntu-based) image. There's a lot of reasons here, but the primary being:

* Alpine doesn't have a security vulnerability reporting process
* Its individual packages aren't stabilized and there's no way to report a vulnerability for them either (and they're community managed)

Past security, its use of musl is non-ideal because most python packages won't have wheels available, and we'll run into random incompatibilies due to musl itself.

The slim images, at this point, aren't much larger than alpine images, while having the good security teams/communities behind ubuntu and debian.

I'm fine with making some improvements to the existing Dockerfile, but ideally it would keep things simple and use a python-slim image and pip.

.github/workflows/publish-to-ghcr.yml
ryan-lane: basically, this will cause the image to be pushed to github's docker registry when a release is cut. Looks good.

dist.Dockerfile
ryan-lane: Well, I guess a benefit of using pyinstaller is that it makes it a lot easier to use a multi-stage build like this. I guess I'm in favor of that change, but please use the python-slim image rather than alpine.

ryan-lane: To followup a bit, I also think it's best to keep the separate dockerfile, like this, in dist, if we're going to use pyinstaller, as it makes it more difficult to do quick development (like bind mounting the cartography code in, so changes don't require rebuilds).

dist.Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3-slim
```
ryan-lane, line 1: Should we specify the 3 version here? I guess it should be whatever highest version we claim compatibility with? I think maybe that's 3.9?

VJftw: update to `python:3.9-slim` :)

achantavy:
Thanks for working on this. As a final step would you be able to update the docs on usage of the container? Maybe you could create a new section here: https://github.com/lyft/cartography/blob/master/docs/root/ops.md. Basically say that this is one option of running cartography in production.

achantavy (approved):

Issue #801, opened 2022-04-07
"string indices must be integers" error


El070422:
Hello everyone! Can you please help me with the issue? Cartography (0.53.0 and 0.54.0) successfully collected data for several aws accounts, everything worked fine. But data collection is not working for one of the accounts. We got the error:

ERROR:cartography.sync:Unhandled exception during sync stage 'aws'
Traceback (most recent call last):
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/sync.py", line 74, in run
    stage_func(neo4j_session, config)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/util.py", line 110, in timed
    return method(*args, **kwargs)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/intel/aws/__init__.py", line 208, in start_aws_ingestion
    _sync_multiple_accounts(
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/intel/aws/__init__.py", line 145, in _sync_multiple_accounts
    _sync_one_account(
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/intel/aws/__init__.py", line 57, in _sync_one_account
    RESOURCE_FUNCTIONS[func_name](**sync_args)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/util.py", line 110, in timed
    return method(*args, **kwargs)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/intel/aws/apigateway.py", line 376, in sync
    sync_apigateway_rest_apis(neo4j_session, boto3_session, region, current_aws_account_id, update_tag)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/util.py", line 110, in timed
    return method(*args, **kwargs)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/intel/aws/apigateway.py", line 366, in sync_apigateway_rest_apis
    load_rest_api_details(neo4j_session, stages_certificate_resources, current_aws_account_id, aws_update_tag)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/util.py", line 110, in timed
    return method(*args, **kwargs)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/intel/aws/apigateway.py", line 321, in load_rest_api_details
    _load_apigateway_certificates(neo4j_session, certificates, update_tag)
  File "/home/centos/.local/lib/python3.8/site-packages/cartography/util.py", line 110, in timed
[...]

PR #852, opened 2022-06-06
Fix for #801

El070422:
Extend method adds only keys to a list of certificates. Therefore, instead of a list of dictionaries, we just get a list of keys, and in the end we come to an error string indices must be integers. We should use append in the case, since we are adding a dictionary (certificate) into the certificates list.
The replacement have been tested on our prod environment without the error.

### DIFF
--- a/cartography/intel/aws/apigateway.py
+++ b/cartography/intel/aws/apigateway.py
@@ -309,7 +309,7 @@ def load_rest_api_details(
             resources.extend(resource)
         if certificate:
             certificate['apiId'] = api_id
-            certificates.extend(certificate)
+            certificates.append(certificate)
 
     # cleanup existing properties
     run_cleanup_job(


kedarghule:
> Hi @kedarghule, Can you tell me please what I should do to sign the CLA? As far as I understand different organizations have different standards regarding what they consider a legally binding agreement to a CLA.

You can see that the CLA check is pending. Just click on it and you'll have to e-sign the thing.

El070422:
@kedarghule, the line doesn't have active links. Maybe I don't have permissions or I'm looking for it in wrong place?
Снимок экрана 2022-06-08 в 18 03 14

kedarghule:
> @kedarghule, the line doesn't have active links. Maybe I don't have permissions or I'm looking for it in wrong place? Снимок экрана 2022-06-08 в 18 03 14

Try this: https://oss.lyft.com/cla

El070422:
> > @kedarghule, the line doesn't have active links. Maybe I don't have permissions or I'm looking for it in wrong place? Снимок экрана 2022-06-08 в 18 03 14
> 
> Try this: https://oss.lyft.com/cla

Ok, I have signed the link but the status is still pending:
Снимок экрана 2022-06-08 в 18 13 11
Снимок экрана 2022-06-08 в 18 13 29

PR #854, opened 2022-06-07
Correctly propagate exceptions to exit code

achantavy:
Currently if any of [these errors](https://github.com/lyft/cartography/blob/b3f29ed65a6bf04fe1e4ab19b529ce007368c5b4/cartography/sync.py#L118-L149) happen (e.g. we fail to connect to the neo4j database) then we return 1 and the exit code is not propagated up to the process -- the resulting returned exit code to the OS is still 0 (success). This PR fixes this and adds typehints to make this process a bit easier to read.

Related to https://github.com/lyft/cartography/pull/509.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -499,7 +499,7 @@ class CLI:
         try:
             return cartography.sync.run_with_config(self.sync, config)
         except KeyboardInterrupt:
-            return 130
+            return cartography.util.STATUS_KEYBOARD_INTERRUPT
 
 
 def main(argv=None):
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -368,7 +368,7 @@ class CLI:
         )
         return parser
 
-    def main(self, argv):
+    def main(self, argv) -> int:
         """
         Entrypoint for the command line interface.
 
@@ -517,4 +517,4 @@ def main(argv=None):
     logging.getLogger('neo4j.bolt').setLevel(logging.WARNING)
     argv = argv if argv is not None else sys.argv[1:]
     default_sync = cartography.sync.build_default_sync()
-    return CLI(default_sync, prog='cartography').main(argv)
+    sys.exit(CLI(default_sync, prog='cartography').main(argv))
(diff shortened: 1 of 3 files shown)


cartography/cli.py
```diff
@@ -517,4 +517,4 @@ def main(argv=None):
     logging.getLogger('neo4j.bolt').setLevel(logging.WARNING)
     argv = argv if argv is not None else sys.argv[1:]
     default_sync = cartography.sync.build_default_sync()
-    return CLI(default_sync, prog='cartography').main(argv)
+    sys.exit(CLI(default_sync, prog='cartography').main(argv))
```
achantavy, line 520: This is the main change of this PR.

cartography/cli.py
```diff
@@ -368,15 +368,15 @@ def _build_parser(self):
...
+    def main(self, argv: str) -> int:
         """
         Entrypoint for the command line interface.
 
         :type argv: string
         :param argv: The parameters supplied to the command line program.
         """
         # TODO support parameter lookup in environment variables if not present on command line
-        config: cartography.config.Config = self.parser.parse_args(argv)
+        config: argparse.Namespace = self.parser.parse_args(argv)
```
achantavy, line 379: We do some duck typing with cartography.config.Config so that it has the same fields as the object returned by parse_args, but teeechnically parse_args returns an argparse.Namespace object so I'll type-annotate it properly here.

cartography/sync.py
```diff
@@ -82,9 +91,10 @@ def run(self, neo4j_driver, config):
                     raise  # TODO this should be configurable
                 logger.info("Finishing sync stage '%s'", stage_name)
         logger.info("Finishing sync with update tag '%d'", config.update_tag)
+        return STATUS_SUCCESS
 
 
-def run_with_config(sync, config):
+def run_with_config(sync: Sync, config: Union[Config, argparse.Namespace]) -> int:
```
achantavy, line 97: Again, our duck typing implicitly handled the cartography Config object vs argparse Namespace arg here (which was pretty neat), but explicit is better

ramonpetgrave64 (approved):

PR #859, opened 2022-06-09
0.62.0

ramonpetgrave64:


VJftw:
Sorry, I've had a little anxiety - before releasing, can we double-check that the new https://github.com/lyft/cartography/actions/workflows/publish-to-ghcr.yml workflow works correctly. Are you able to trigger a `workflow dispatch` against `master`? It should push a `:master` tagged image. 

The anxiety is that it could fail on the checkout step as it doesn't have `permissions: { contents: read }`.

It may work though, the [Test Suite](https://github.com/lyft/cartography/actions/workflows/test_suite.yml) workflow seems to be able to checkout fine.

ryan-lane:
@VJftw I used the same workflow in our recently released neo4j dashboard and it worked without read permission.  https://github.com/paypay/seizu/blob/master/.github/workflows/publish.yml :)

ryan-lane:
@achantavy @ramonpetgrave64 after cutting a release, the image will be published, but it will probably be published as private. you'll need to make sure the github org is allowed to publish public packages, then after publishing the packages, you'll need to go to the packages via from the org, find the image, go to the image setting, and change the visibility to public.

ramonpetgrave64:
@VJftw @ryan-lane Thanks, I think it worked, and was already set to public.

achantavy (approved):

PR #860, opened 2022-06-10
Removing `return count(*) as TotalCompleted from AWS Cleanup files

kedarghule:
This PR removes the `return COUNT(*) as TotalCompleted from the AWS Cleanup files.

I'll open a separate PR for Azure. Thought would keep these PRs service specific.


achantavy (approved):
Thank you!!

PR #862, opened 2022-06-12
Removing `return count(*) as TotalCompleted from Azure Cleanup files

kedarghule:
This PR builds on https://github.com/lyft/cartography/pull/860 but for Azure cleanup files. 


ramonpetgrave64 (approved):
Thanks!

PR #863, opened 2022-06-14
Added support to run Azure for specific services

ganeshjagdale9545:
For Azure, support to run Azure for specific Services list like sql, storage, compute etc.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -177,6 +177,16 @@ class CLI:
             ),
         )
         parser.add_argument(
+            '--azure-requested-syncs',
+            type=str,
+            default=None,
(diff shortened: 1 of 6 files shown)


achantavy (changes requested):
Hi @ganeshjagdale9545 - thanks for working on this, definitely looks very useful for Azure users. Would you be able to please add tests to ensure that the command line switch behaves as expected?

As an (admittedly really janky) example, we do this in AWS here: https://github.com/lyft/cartography/blob/ecec80a1db1fc926c4786b04830ca974a9af94de/tests/integration/cartography/intel/aws/test_init.py#L117-L139

PR #865, opened 2022-06-15
Removing `return count(*) as TotalCompleted from GitHub, GSuite, Okta, Kubernetes and PagerDuty cleanup files

kedarghule:
This PR builds on https://github.com/lyft/cartography/pull/860 but for GitHub, GSuite, Okta, Kubernetes and PagerDuty cleanup files and is the last one for this task. (Hopefully all cleanup files are updated now)


achantavy (approved):
Thanks!

Issue #773, opened 2022-03-01
AWS Sync: Try all AWS accounts, and raise Exception afterwards


ramonpetgrave64:
*Title*: *Try all AWS accounts, and raise Exception afterwards*

*Description*:
Sometimes a single AWS account's sync could fail. In such situations, we more likely want to proceed with the rest of our accounts, but still be notified of the failure.

I propose to wrap `_sync_one_account` into a try-catch, invoke `logger.exception` for the failures, and then finally raise another exception listing the accounts that failed.

https://github.com/lyft/cartography/blob/master/cartography/intel/aws/__init__.py#L145

For example:
```
failed_accont_ids = []
for account_id in account_ids:
    try:
        _sync_one_account(account_id)
    except Exception as e:
        logger.exception(e)
        failed_accont_ids.append(account_id)
if failed_accont_ids:
    raise Exception(f'Syncs failed for accounts {failed_accont_ids}')
```

achantavy:
I agree on this approach and wanted to add that we should ensure that if there is an exception in one of the accounts, then execution stops and _does not_ continue on to the cleanup jobs. We cannot have an exception in the `get` phase cause all data to be deleted.

achantavy:
@kunaals @ryan-lane @mpurusottamc - thoughts?

ryan-lane:
Yes. This is a step in the direction I'd like to see. I'd really like to be able to run AWS accounts separately, and I think any work we'd need to make that work would also be needed for this.

mpurusottamc:
Love the idea. This definitely avoids disruption in a multi-account setup.

+1 @achantavy for your idea of avoiding clean up in case of an exception in one of the accounts. We have been bitten on this in the past :)

marco-lancini:
Hi all,
I'd like to revamp this discussion.

I'm currently trying to ingest a large-scale AWS Org, and it has been painful to get Cartography to complete even a single run.
With Orgs of this size, there is a high chance of facing resources with peculiar characteristics, which might ultimately led to a failed run. 

In my case, for the Org in question, just in the past week I had to open multiple issues (#800, #808, #809, #821) to add ad-hoc `try/catch` exceptions to multiple parts of the codebase, in order to get to a successful ingestion (and I'm not there yet, so more issues might come this way).

I understand the original idea of being conservative (especially for the cleanup jobs), 
but failing hard on a whole ingestion (oftentimes hours after it started) just because of an "odd" resource is becoming increasingly frustrating as a user.

[...]

PR #870, opened 2022-06-29
#773: --aws-best-effort-mode: Create aggregated exceptions for AWS accounts sync

erickduran:
Aggregating AWS account sync exceptions into one and raising until the loop finishes

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -128,6 +128,14 @@ class CLI:
             ),
         )
         parser.add_argument(
+            '--aws-best-effort-mode',
+            action='store_true',
+            help=(
+                'Enable AWS sync best effort mode when syncing AWS accounts. This will allow cartography to continue '
+                'syncing without raising an exception if one or more AWS accounts fail to sync, errors will be logged.'
+            ),
+        )
+        parser.add_argument(
             '--azure-sync-all-subscriptions',
             action='store_true',
             help=(
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -132,7 +132,7 @@ class CLI:
             action='store_true',
             help=(
                 'Enable AWS sync best effort mode when syncing AWS accounts. This will allow cartography to continue '
-                'syncing without raising an exception if one or more AWS accounts fail to sync, errors will be logged.'
+                'syncing other accounts and delay raising an exception until the very end.'
             ),
         )
         parser.add_argument(
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -143,7 +143,7 @@ def _sync_multiple_accounts(
     common_job_parameters: Dict[str, Any],
     config: Config,
     aws_requested_syncs: List[str] = [],
-) -> None:
+) -> bool:
     logger.info("Syncing AWS accounts: %s", ', '.join(accounts.values()))
     organizations.sync(neo4j_session, accounts, sync_tag, common_job_parameters)
 
@@ -188,6 +188,8 @@ def _sync_multiple_accounts(
     # up those nodes after all AWS accounts have been synced.
     if not failed_account_ids:
         run_cleanup_job('aws_post_ingestion_principals_cleanup.json', neo4j_session, common_job_parameters)
+        return True
+    return False
 
 
 @timeit
@@ -233,7 +235,7 @@ def start_aws_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
     if config.aws_requested_syncs:
         requested_syncs = parse_and_validate_aws_requested_syncs(config.aws_requested_syncs)
 
-    _sync_multiple_accounts(
+    sync_successful = _sync_multiple_accounts(
         neo4j_session,
         aws_accounts,
         config.update_tag,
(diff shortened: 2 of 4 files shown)


achantavy:
FYI: @ryan-lane @marco-lancini @mpurusottamc :)

ryan-lane (approved):
Looks great!

marco-lancini (dismissed):

cartography/intel/aws/__init__.py
```diff
@@ -144,21 +145,35 @@ def _sync_multiple_accounts(
...
+                aws_requested_syncs=aws_requested_syncs,  # Could be replaced later with per-account requested syncs
+            )
+        except Exception as e:
+            failed_account_ids.append(account_id)
+            exception_traceback = traceback.TracebackException.from_exception(e)
+            traceback_string = ''.join(exception_traceback.format())
+            exception_tracebacks.append(traceback_string)
+            continue
+
+    if failed_account_ids:
```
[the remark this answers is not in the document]
erickduran: We can add the a formatted string (e.g. `f'Account ID: {account_id}\n`) prior to each traceback since it's just a string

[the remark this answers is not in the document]
achantavy: Yeah I think that works for me

cartography/intel/aws/__init__.py
```diff
@@ -144,21 +145,35 @@ def _sync_multiple_accounts(
...
+        try:
+            _sync_one_account(
+                neo4j_session,
+                boto3_session,
+                account_id,
+                sync_tag,
+                common_job_parameters,
+                aws_requested_syncs=aws_requested_syncs,  # Could be replaced later with per-account requested syncs
+            )
+        except Exception as e:
```
ramonpetgrave64, line 167: To help with debugging, it's also helpful to do a `logger.exception(...)` here so that the timestamp of the exception can be recorded.

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -91,6 +92,29 @@ def test_start_aws_ingestion(mock_run_analysis, mock_sync_multiple, mock_orgs, m
     assert mock_run_analysis.call_count == 3
 
 
+@mock.patch('cartography.intel.aws.boto3.Session')
+@mock.patch('cartography.intel.aws.organizations.get_aws_accounts_from_botocore_config')
+@mock.patch.object(cartography.intel.aws, '_sync_one_account', return_value=None)
+@mock.patch.object(cartography.intel.aws, 'run_analysis_job')
+@mock.patch.object(cartography.intel.aws, 'run_cleanup_job')
+def test_start_aws_ingestion_raises_aggregated_exceptions(
```
ramonpetgrave64, line 100: Two more important things I think we should test for:

- The loop continues after a failure.
- The details like the Account ID and Exception class are logged.

ramonpetgrave64 (changes requested):

tests/integration/cartography/intel/aws/test_init.py
erickduran: I think the loop continuity is covered by this (since all iterations are throwing the exception as a side effect):
https://github.com/lyft/cartography/blob/43b4b95faf02adf728ebd81503364458e86ce5d8/tests/integration/cartography/intel/aws/test_init.py#L113
or do you have an additional scenario in mind?

Also this checks for the Exception class in the traceback, so maybe just add the account ID as discussed above (?):
https://github.com/lyft/cartography/blob/43b4b95faf02adf728ebd81503364458e86ce5d8/tests/integration/cartography/intel/aws/test_init.py#L112

cartography/intel/aws/__init__.py
achantavy: Just talked with @ramonpetgrave64 and this is what we should do:

- let's include the failing account ID and the timestamp next to each traceback.
- there is no need to do logger.exception on line 168.

The main idea is to collect all the exceptions as they happen and wait until the end before logging at `error` level. We only want to log at `error` level right before when the program is fatally failing.

cartography/intel/aws/__init__.py
```diff
@@ -144,21 +145,35 @@ def _sync_multiple_accounts(
...
+            failed_account_ids.append(account_id)
+            exception_traceback = traceback.TracebackException.from_exception(e)
+            traceback_string = ''.join(exception_traceback.format())
+            exception_tracebacks.append(traceback_string)
+            continue
```
achantavy, line 172: Talked with @ramonpetgrave64 and it might be good to make this behavior configurable:

maybe something like
```
        if config.aws_best_effort_mode: # 
            # append the traceback and continue
            # ...
        else:
            raise
```

What do you think @erickduran?

tests/integration/cartography/intel/aws/test_init.py
achantavy: @erickduran perfect

cartography/intel/aws/__init__.py
erickduran: Yeah I like that, on it

erickduran: Just added this, note that  following effects when using `aws_best_effort_mode` and there is at least one error:
- will still try to execute the `run_analysis` functions after `_sync_multiple_accounts` finishes
- will not perform `run_cleanup_job`

Let me know your thoughts

achantavy: Very good callouts. I like how you're quickly getting familiar with our technical  debt 😂 .

> will not perform run_cleanup_job

This is ok.

> will still try to execute the run_analysis functions after _sync_multiple_accounts finishes

I have an idea: What are your thoughts on having `_sync_multiple_accounts()` return True if all accounts succeeded and False otherwise? Then in `start_aws_ingestion()` we could do something like

```python
sync_successful = _sync_multiple_accounts()
...
if sync_successful:
  # run analysis jobs
```

cartography/cli.py
```diff
@@ -127,6 +127,14 @@ def _build_parser(self):
                 'respects the AWS CLI/SDK environment variables and does not override them.'
             ),
         )
+        parser.add_argument(
+            '--aws-best-effort-mode',
+            action='store_true',
+            help=(
+                'Enable AWS sync best effort mode when syncing AWS accounts. This will allow cartography to continue '
+                'syncing without raising an exception if one or more AWS accounts fail to sync, errors will be logged.'
```
achantavy, line 135:
```suggestion
                'syncing other accounts and delay raising an exception until the very end.'
```

achantavy (approved):
This makes me so happy 🍻

ramonpetgrave64 (changes requested):

PR #883, opened 2022-07-21
Implement client functions

achantavy:
Implements transaction helper functions so that other Python apps can import them and not worry about implementing their own transaction logic.

Implements an example function that takes advantage of this: `get_aws_admin_like_principals()`.

The idea is for us to formalize important queries by using cartography `load()` functions to integration test each client function so that we can quickly catch when breaking schema changes happen.

Next steps could include to use data classes, but I wanted to keep this relatively lightweight as dicts for now.

### DIFF
--- a/cartography/client/core/tx.py
+++ b/cartography/client/core/tx.py
@@ -9,6 +9,25 @@ import neo4j
 
 
 def read_list_of_values_tx(tx: neo4j.Transaction, query: str, **kwargs) -> List[Union[str, int]]:
+    """
+    Runs the given Neo4j query in the given transaction object and returns a list of either str or int. This is intended
+    to be run only with queries that return a list of a single field.
+
+    Example usage:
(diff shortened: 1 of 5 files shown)


cartography/client/core/tx.py
```diff
@@ -0,0 +1,50 @@
+from typing import Any
```
achantavy, line 1: @marco-lancini, @ryan-lane, @mpurusottamc - hey, so I've created this set of functions so that other Python clients might have an easier time pulling data from a cartography-based graph (since these functions handle the boilerplate transaction logic)

I'd like to write maybe 1-3 sample Python functions using these to show users what they can do. For example maybe I'd include something like `get_ec2_instances_by_account()`. Any ideas on what would be good examples? Anything from https://github.com/marco-lancini/cartography-queries/ or any other places that you've found particularly compelling when showing your projects to people? :)

ramonpetgrave64 (approved):

PR #884, opened 2022-07-21
ECR: add ecr test_cleanup_repositories

ramonpetgrave64:
Ensure that after the cleanup job runs, all ECRRepository nodes with a different UPDATE_TAG are removed from the AWSAccount node. We load 100 additional nodes, because the cleanup job is configured to run iteratively, processing 100 nodes at a time. So this test also ensures that iterative cleanups do work.


achantavy (approved):
I love this, thank you for adding this sanity check

PR #885, opened 2022-07-21
neo4j driver upgrade from 1.7.6 to >=4.4.4

ramonpetgrave64:
Upgrading the driver from 1.7.6 to 4.4.4

We had previously attempted this and did a rollback in #775 and #782. The mistake was that in `statement.py` we incorrectly change `summary()` to `_obtain_summary()`, when we should have changed it to `consume()`. 

This cause some jobs, especially cleanup jobs, to stop be incomplete, running after only a single iteration, and also pushing empty stats. In #884, we now have tests that confirm that cleanup jobs will correctly do multiple iterations as needed.

This time we are doing it correctly following all of the [breaking changes](https://neo4j.com/docs/api/python-driver/current/breaking_changes.html) in Neo4j's upgrade guidance.

Now that we are officially upgrading the driver, `patch_libraries()` is no longer needed within the `experimental_neo4j_4x_support.py`, so I've commented it out to still serve as a reference.


achantavy (approved):
🎉

PR #895, opened 2022-08-15
Generalize graph write queries

achantavy:


.pre-commit-config.yaml
```diff
@@ -39,7 +39,7 @@ repos:
     -   id: reorder-python-imports
         args: [--py3-plus]
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.950
+    rev: v0.981
```
achantavy, line 42: Because we need python/mypy#13398

achantavy:
Obsoleted by https://github.com/lyft/cartography/pull/1038

PR #905, opened 2022-08-25
Added changes docker-compose.yml

steve-solun:
Support GDS and APOC plugins + fix the networking issue with the localhost as described.

When you use localhost url inside a container, it doesn't treat it as the host localhost, but inside of the container. 
There are two options you can go about. One is to add:

`network_mode: host`
to the cartography service. Then the localhost points to the host machine and it should find Neo4j at localhost:7474.

Instead of specifying `--neo4j-uri=bolt://localhost:7687` you can specify `--neo4j-uri=bolt://neo4j:7687` to point at Neo4j instance container. The URL is identical to the service name.


docker-compose.yml
```diff
@@ -1,31 +1,54 @@
...
+      context: .
+      dockerfile: dist.Dockerfile
     init: true
     restart: on-failure
     depends_on:
       - neo4j
+    environment:
+      - EXPERIMENTAL_NEO4J_4X_SUPPORT=True 
+    volumes:
+      - /Users/stevesolun/.aws:/root/.aws/
```
achantavy, line 50: Is there a way to generalize this so that it doesn't use your home dir? :)

achantavy (commented):
Hi @steve-solun - thank you for the PR! Can you please sign the CLA?

I am not a Docker expert so I am actually going to ask for help from someone else who knows more than me here. At first glance these changes look very helpful.

achantavy:
Hi @VJftw, @ryan-lane - any chance you have cycles to help take a skim at the Docker changes?

docker-compose.yml
```diff
@@ -1,31 +1,54 @@
...
       - ./.compose/neo4j/logs:/logs
       - ./.compose/neo4j/plugins:/plugins
     environment:
-      # Raise memory limits
+      # Raise memory limits:
       - NEO4J_dbms_memory_pagecache_size=1G
       - NEO4J_dbms.memory.heap.initial_size=1G
       - NEO4J_dbms_memory_heap_max__size=1G
+      # Auth:
+      - NEO4J_AUTH=none
```
ryan-lane, line 23: 😭 I guess I understand that this makes thing easier, especially the test suites, but I really wish we injected the credentials into neo4j, rather than turning auth off.

docker-compose.yml
```diff
@@ -1,31 +1,54 @@
...
+        interval: 10s
         timeout: 10s
-        retries: 5
+        retries: 10
   cartography:
-    image: lyft/cartography
-    user: cartography
+    build:
+      context: .
+      dockerfile: dist.Dockerfile
```
ryan-lane, line 42: Downside here is that it requires people to build the image to use it, right?

dist.Dockerfile
```diff
@@ -2,12 +2,16 @@ FROM python:3.9-slim
 
 # the UID and GID to run cartography as
 # (https://github.com/hexops/dockerfile#do-not-use-a-uid-below-10000).
-ARG uid=10001
-ARG gid=10001
+ARG uid=0 #10001
+ARG gid=0 #10001
```
ryan-lane, line 6: Why are you switching this to 0? We're purposely setting the uid to something more secure.

docker-compose.yml
ryan-lane: Also, it should be possible to add this build info while also keeping the image/user settings, right?

docker-compose.yml
ryan-lane: yeah, I think `~/.aws:/cartography/.aws/` should work here.

ryan-lane (changes requested):
For the localhost issue... don't use localhost, use the neo4j url. We shouldn't be suggesting to people to use host network mode. Compose works by giving names to separate containers and networking between them. I think this is more of a documentation problem.

docker-compose.yml
steve-solun: @ryan-lane sure thing, I wanted it to be easier for my testing, but in a real dev/prod env we are going to inject username and password.

docker-compose.yml
steve-solun: I think yes, it worked for me, but let's check it out.

docker-compose.yml
steve-solun: @achantavy I have signed the CLA :) please check.

docker-compose.yml
ryan-lane: Yeah, for this development setup it's probably fine, because it's also what the cartography integration tests expect. It's not ideal, but probably the right thing to do for now.

steve-solun: @ryan-lane We can add in bold that it's your end user responsibility to add username and password.

docker-compose.yml
ryan-lane, line 50: I don't think this is resolved. It's still using your home directory for the volume mount :)

steve-solun: I will check again and re-PR it

steve-solun:
@achantavy @ryan-lane please check now.

docker-compose.yml
steve-solun: @ryan-lane what would you like to change? We need Neo and your functionality to work together. Please explain.

ryan-lane: Both the user and image should be left here. The user makes the container run as non-root, which is more secure. The image definition let's folks pull the image without needing to build it.

I'm not sure what this has to do with cartography and neo4j compatibility?

steve-solun: @ryan-lane Sorry didn't get your point. Please give me an example of what do you mean and how would you like me to change it and why. I think I lost you a bit in this case.

steve-solun: This is a working docker-compose setup, if someone wants to change anything, let them change or tell me exactly what should I change in order to finish this PR. Maybe we can add some explanations to the docs? Please advise how would you like to proceed.

steve-solun: @achantavy

ryan-lane:
You removed the user and image lines from the cartography container in the
compose file.

We want the container to run as the user defined in the image (defined by
number; its possible that the image needs to add a user explicitly rather
than just specifying the uid?). We don't want the container running as
root, because that is insecure.

If the image line is removed, then to use the docker compose setup, you'd
have to build the image to use it. With the image line defined, you can run
cartography and it'll pull the image from the registry.

I'm asking for you to keep those lines in. If it's not working with that,
let's fix things to make it work.

steve-solun:
@ryan-lane for some reason I am getting the following error:

```
docker compose up                                
[+] Running 1/9
 ⠇ neo4j Pulling                                                                                                                                    2.9s
   ⠿ a9fe95647e78 Already exists                                                                                                                    0.0s
   ⠦ 4015b6e8cc8d Pulling fs layer                                                                                                                  0.6s
   ⠦ fd343789f78f Pulling fs layer                                                                                                                  0.6s
   ⠦ f361af981f24 Pulling fs layer                                                                                                                  0.6s
   ⠦ 2b95c30758e7 Waiting                                                                                                                           0.6s
   ⠦ e6e9cfe06407 Waiting                                                                                                                           0.6s
   ⠦ fc07026944f6 Waiting                                                                                                                           0.6s
 ⠿ cartography Error                                                                                                                                2.9s
Error response from daemon: pull access denied for lyft/cartography, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

steve-solun:
And the end of docker-compose is looks like:

```
cartography:
    # build:
    #   context: .
    #   dockerfile: dist.Dockerfile
    image: lyft/cartography
    user: cartography
    init: true
    restart: on-failure
    depends_on:
      - neo4j
    environment:
      - EXPERIMENTAL_NEO4J_4X_SUPPORT=True 
    volumes:
      - /Users/stevesolun/.aws:/root/.aws/
```

steve-solun:
@ryan-lane no such repository on DockerHub, I will build and upload it there but maybe you guys want it to be official.

steve-solun:
@ryan-lane please check it out now :)

ryan-lane:
Ah, I think the official one is at github, not docker hub. The reference
needs to be fixed!

steve-solun:
Sorry it didn't work, I have posted the error. 
If you want me to change the reference, fix the image/image name.
@ryan-lane I think everything is there and we can just approve it to let people use it. Just guide me with image reference. Hopefully we will finish this long PR.

steve-solun:
@ryan-lane I have fixed the image, please have a look now. Added some examples for people to use out of the box.

docker-compose.yml
```diff
@@ -13,19 +13,37 @@ services:
...
     healthcheck:
         test: ["CMD", "curl", "-f", "http://localhost:7474"]
-        interval: 30s
+        interval: 10s
         timeout: 10s
-        retries: 5
+        retries: 10
   cartography:
-    image: lyft/cartography
+    image: ghcr.io/lyft/cartography:0.63.0
```
ryan-lane, line 38: No need to list the specific version. If it's unlisted, it'll use the latest released tag, which will let us avoid having to update this version tag everytime we cut a release.

ryan-lane:
Looks good, just one last piece of feedback about removing the version number from the image.

docker-compose.yml
steve-solun: Done @ryan-lane please merge.

dist.Dockerfile
```diff
@@ -8,6 +8,10 @@ ARG gid=10001
 COPY . /var/cartography
 WORKDIR /var/cartography
 
+RUN apt-get update --fix-missing
+RUN apt --allow-unauthenticated  update -y
+RUN apt-get install vim telnet curl gcc -y
```
ryan-lane, line 13: Wait, I hadn't noticed this before, why is vim, telnet, curl and gcc being added?

dist.Dockerfile
```diff
@@ -8,6 +8,10 @@ ARG gid=10001
 COPY . /var/cartography
 WORKDIR /var/cartography
 
+RUN apt-get update --fix-missing
+RUN apt --allow-unauthenticated  update -y
```
ryan-lane, line 12: I can't think of a good reason why `--allow-unauthenticated` would be needed here.

ryan-lane (approved):

ryan-lane:
@achantavy looks good to me. I can't properly approve/merge because something is going on with github and my owners invitation. can you merge?

ryan-lane:
@steve-solun can you fix the lint error? It's an issue with missing EOF character I believe.

ryan-lane (approved):

PR #911, opened 2022-09-05
Check boto3 session allows calling 'inspector2' client service

heryxpc:
Considering not all accounts might have 'inspector2' configured, this avoids breaking the AWS import module unless the service is available.


marco-lancini (approved):

heryxpc:
This might not be required if the latest boto3 client is being used. Closing this for now.

PR #913, opened 2022-09-07
convert github syntax

ramonpetgrave64:
Converts the query syntax for Github to use the new format.

See #914


ramonpetgrave64:
accidentally merged without review, but I've since reverted, and reopened in #916

PR #915, opened 2022-09-07
fix backoff in tests

ramonpetgrave64:
This patches out the `backoff` functionality from unit tests. For the test against `handle_aws_regions`, we intentionally introduce errors, but the `backoff` decorator will have the function retry for up to 600 seconds. This PR changes that behavior to skip retries.


ramonpetgrave64:
@rlukin @marco-lancini

marco-lancini (approved):

PR #929, opened 2022-09-09
AWS: fix link between users and their access keys

p-l-:
The links was created based on the AWSUser object's name, while several such objects may share the same name accross different AWSAccount objects. This PR uses the ARN to create the links, which is unique.

See discussion on Slack <https://lyftoss.slack.com/archives/CTZUQL0KX/p1662638519733589>


p-l-:
CLA should be OK now

ramonpetgrave64:
Thank you! Looks good, but just a small change left for the linter.

p-l-:
Thanks @ramonpetgrave64! Should be good now!

Issue #914, opened 2022-09-07
Neo4j 4.x Upgrade


ramonpetgrave64:
## Neo4j 4.x Upgrade

Neo4j 3.5 is at end-of-life. As of #912 , cartography now has official Neo4j 4.x support, and our tests on Github will use 4.4. 

### Differences

The most significant difference is that there is a new syntax for query parameters. Instead of `{my_param}`, we must now use `$my_param`. This requires us to change hundreds of files that we will break into several small PRs for us all to manage. 

Going forward:

1. All new PRs must use the new syntax
2. Several PRs will come in the following days to upgrade manageable sections of cartography code.
3. The auto-patcher will continue to be used, so that code using the older syntax will continue to work.
4. Once all code has been converted, we will remove the auto-patcher.

### PRs

- https://github.com/lyft/cartography/pull/916
- https://github.com/lyft/cartography/pull/917
- https://github.com/lyft/cartography/pull/918
- https://github.com/lyft/cartography/pull/919
- https://github.com/lyft/cartography/pull/920
- https://github.com/lyft/cartography/pull/922
- https://github.com/lyft/cartography/pull/923
- https://github.com/lyft/cartography/pull/924
- https://github.com/lyft/cartography/pull/925
- https://github.com/lyft/cartography/pull/926
- https://github.com/lyft/cartography/pull/927
- https://github.com/lyft/cartography/pull/928
- Documentation
  - https://github.com/lyft/cartography/pull/978
- AWS
  - https://github.com/lyft/cartography/pull/932
  - https://github.com/lyft/cartography/pull/933
  - https://github.com/lyft/cartography/pull/934
  - https://github.com/lyft/cartography/pull/935
  - https://github.com/lyft/cartography/pull/936
  - https://github.com/lyft/cartography/pull/937
  - https://github.com/lyft/cartography/pull/938
  - https://github.com/lyft/cartography/pull/939
  - https://github.com/lyft/cartography/pull/940
  - https://github.com/lyft/cartography/pull/941
  - https://github.com/lyft/cartography/pull/942
  - https://github.com/lyft/cartography/pull/943
  - https://github.com/lyft/cartography/pull/944
  - https://github.com/lyft/cartography/pull/945
  - https://github.com/lyft/cartography/pull/946
  - https://github.com/lyft/cartography/pull/947
  - https://github.com/lyft/cartography/pull/948
  - https://github.com/lyft/cartography/pull/949
  - https://github.com/lyft/cartography/pull/950
[...]

PR #932, opened 2022-09-09
convert s3 syntax

ramonpetgrave64:
Converts the query syntax for s3 to use the new format.

See #914


kedarghule (approved):
Looks good! :D

PR #933, opened 2022-09-09
convert ssm syntax

ramonpetgrave64:
Converts the query syntax for ssm to use the new format.

See #914


erickduran (approved):

erickduran:
@ramonpetgrave64 looks like this one needs rebasing

PR #935, opened 2022-09-09
convert ksm syntax

ramonpetgrave64:
Converts the query syntax for kms to use the new format.

See #914


kedarghule (approved):
Looks good! :D

PR #982, opened 2022-09-15
convert merge_module_sync_metadata_syntax

ramonpetgrave64:
See Neo4j 4.x Upgrade #914


cartography/util.py
```diff
@@ -93,12 +93,11 @@ def merge_module_sync_metadata(
             n:SyncMetadata, n.firstseen=timestamp()
         SET n.syncedtype='${synced_type}',
             n.grouptype='${group_type}',
-            n.groupid={group_id},
-            n.lastupdated={UPDATE_TAG}
```
meng-han, line 97: why is this update_tag in a different syntax with group_id?

cartography/util.py
```diff
@@ -93,12 +93,11 @@ def merge_module_sync_metadata(
...
+            n.groupid='${group_id}',
+            n.lastupdated=$UPDATE_TAG
     """)
     neo4j_session.run(
         template.safe_substitute(group_type=group_type, group_id=group_id, synced_type=synced_type),
-        group_id=group_id,
```
meng-han, line 101: deletion is intended?

ramonpetgrave64: Yes, it was already a bit redundant and did not need to be a query parameter, since it could be used as a Template parameter.

cartography/util.py
ramonpetgrave64: Here I use `group_id` as a Template parameter, wrapped again inside quotes. There's also a integ test that covers this change.

meng-han (approved):

PR #989, opened 2022-09-19
Kubernetes: support ingresses

p-l-:
Support includes rules & paths, and the links from paths to services, ultimately bringing links from an ingress to the targeted service.


tests/integration/cartography/intel/kubernetes/test_ingresses.py
```diff
@@ -0,0 +1,38 @@
+from cartography.intel.kubernetes.ingresses import load_ingresses
+from tests.data.kubernetes.ingresses import GET_INGRESSES_DATA
+from tests.integration.cartography.intel.kubernetes.test_services import test_load_services
+
+
+TEST_UPDATE_TAG = 123456789
+
+
+def test_load_ingresses(neo4j_session):
```
ramonpetgrave64, line 9: For the tests, also query for the relationships. This will make us a bit more confident about the test data.

ramonpetgrave64:
This is really cool! I'll take a closer look tomorrow.

tests/integration/cartography/intel/kubernetes/test_ingresses.py
p-l-: Done, thanks for your feedback!

cartography/intel/kubernetes/ingresses.py
```diff
@@ -0,0 +1,101 @@
...
+                            "port": p.backend.service.port.number,
+                        }
+                        for p in r.http.paths
+                    ],
+                }
+                for r in ingress.spec.rules
+            ],
+            "tls": ingress.spec.tls,
+        }
+        for ingress in client.networking.list_ingress_for_all_namespaces().items
```
ramonpetgrave64, line 54: We usually like to have `get_`, `transform_`, and `load_` in different functions.  Here, you combine the get and transform. I didn't correct this in your previous PR about KubernetesSecrets, but try to follow the pattern in this PR.

cartography/data/indexes.cypher
```diff
@@ -451,3 +451,10 @@ CREATE INDEX IF NOT EXISTS FOR (n:KubernetesSecret) ON (n.lastupdated);
 CREATE INDEX IF NOT EXISTS FOR (n:KubernetesService) ON (n.id);
 CREATE INDEX IF NOT EXISTS FOR (n:KubernetesService) ON (n.name);
 CREATE INDEX IF NOT EXISTS FOR (n:KubernetesService) ON (n.lastupdated);
+CREATE INDEX IF NOT EXISTS FOR (n:KubernetesIngress) ON (n.id);
+CREATE INDEX IF NOT EXISTS FOR (n:KubernetesIngress) ON (n.name);
+CREATE INDEX IF NOT EXISTS FOR (n:KubernetesIngress) ON (n.lastupdated);
+CREATE INDEX IF NOT EXISTS FOR (n:KubernetesIngressRule) ON (n.ingress_uid, n.host);
+CREATE INDEX IF NOT EXISTS FOR (n:KubernetesIngressRule) ON (n.lastupdated);
+CREATE INDEX IF NOT EXISTS FOR (n:KubernetesIngressRuleHttpPath) ON (n.ingress_uid, n.host, n.path, n.path_type);
```
ramonpetgrave64, line 459: This would be our first time using composite indexes. There could be some performance implications.
https://neo4j.com/docs/cypher-manual/4.1/administration/indexes-for-search-performance/#administration-indexes-single-vs-composite-index

Have you noticed any issues when you do the single indexes?

p-l-: No, it went really fine. I have been careful with the order to be future-usages-proof (as for now, we use all the keys when we use this index, as it is a kind of "id" of the rule or the path).

cartography/intel/kubernetes/ingresses.py
p-l-: I have not found a `transform_` function in the `kubernetes` module (I used other submodules as templates for my code).

Do you think I could do that in another PR? Maybe with other `kubernetes` submodules if appropriate?

ramonpetgrave64: The `get_` is meant only to do the api fetch. And the `transform_` is meant to munge your original list of dictionaries into another list of dictionaries that your `load_` will use.

cartography/data/indexes.cypher
p-l-: I could do that, but:
- compound indexes are meant exactly for that
- such `id` index would not then be suitable for lookups on just the ingress_uid, for example

cartography/intel/kubernetes/ingresses.py
p-l-: Yes I totally understand. What I meant is: this pattern seems not in use yet in the whole k8s submodule, so if you agree, I could do that in another PR, for all the components at once.

cartography/data/indexes.cypher
ramonpetgrave64: Please do that. 
Why do you think it would not be appropriate?

cartography/intel/kubernetes/ingresses.py
ramonpetgrave64: I know it's not used in this module, but it really should be used.

cartography/data/indexes.cypher
p-l-: What I'm saying is: compound indexes are done for that kind of use cases. They have the advantages of

- being straightforward (you just use them with the API)
- being usable with only some keys (you don't have to create a regexp to look for all rules within a given ingress, for example)

I see no advantage of creating a pseudo field with a string concatenation of different values. That's exactly where compound indexes are needed.

cartography/intel/kubernetes/ingresses.py
p-l-: That's done. Does that match what you expect?

cartography/intel/kubernetes/ingresses.py
```diff
@@ -0,0 +1,108 @@
...
+            ingress.deleted_at = k8ingress.deletion_timestamp,
+            ingress.tls = k8ingress.tls
+        WITH ingress, k8ingress.namespace as ns, k8ingress.cluster_uid as cuid, k8ingress.rules as k8rules
+        MATCH (cluster:KubernetesCluster {id: cuid})-[:HAS_NAMESPACE]->(space:KubernetesNamespace {name: ns})
+        MERGE (space)-[rel1:HAS_INGRESS]->(ingress)
+        ON CREATE SET rel1.firstseen = timestamp()
+        SET rel1.lastupdated = $update_tag
+        WITH space, ingress, k8rules
+        UNWIND k8rules as k8rule
+            MERGE (rule: KubernetesIngressRule {ingress_uid: ingress.id, host: k8rule.host})
```
ramonpetgrave64, line 82: Just like you did for `KubernetesIngress`, I still need you to create some suitable `id` for for `KubernetesIngressRule` and for `KubernetesIngressRuleHttpPath`.

cartography/data/indexes.cypher
ramonpetgrave64: The advantage is that we will have an `id` for these new nodes, just like we have for every other node in our index. The n=Neo4j docs also advise that users come up with their own Node ids, rather than leaving the ids unassigned.

You may keep the compound indexes, but you must have an appropriate `id` for the new nodes.

cartography/intel/kubernetes/ingresses.py
p-l-: I still cannot see how an `id` would be needed when `id`s, for rules and paths, are in fact composed indexes, but I won't have time to continue this discussion further anytime soon. Thanks for the review anyway.

PR #993, opened 2022-09-21
Add Option to Specify Request Timeout in PagerDuty API Session.

rana-awais-paypay:
As the title suggests.


ryan-lane:
What's the current behavior, if the timeout isn't specified? Is the default for the timeout None, because this change will cause None to be passed in.

cartography/intel/pagerduty/__init__.py
```diff
@@ -38,6 +38,8 @@ def start_pagerduty_ingestion(
         logger.info('PagerDuty import is not configured - skipping this module. See docs to configure.')
         return
     session = APISession(config.pagerduty_api_key)
+    if config.pagerduty_request_timeout:
```
ryan-lane, line 41: What if the timeout is set to 0?

ryan-lane:
Checked the requests docs. None is default. Please update the if statement for an explicit check on None, so 0 can be used as an argument

rana-awais-paypay:
> Checked the requests docs. None is default. Please update the if statement for an explicit check on None, so 0 can be used as an argument

While `requests` default is None, pagerduty session sets it to 5 seconds as default: https://pagerduty.github.io/pdpyras/#pdpyras.PDSession.timeout

Either way, checking against None makes sense and I've updated the check.

ryan-lane:
@rana-awais-paypay oh, please also update the docs: https://github.com/lyft/cartography/blob/master/docs/root/modules/pagerduty/config.md

rana-awais-paypay:
> @rana-awais-paypay oh, please also update the docs: https://github.com/lyft/cartography/blob/master/docs/root/modules/pagerduty/config.md

Done.

docs/root/modules/pagerduty/config.md
```diff
@@ -7,3 +7,4 @@ Follow these steps to analyze PagerDuty objects with Cartography.
 1. Prepare your PagerDuty API key.
     1. Generate your API token by following the steps from the PagerDuty [Generating API Keys documentation](https://support.pagerduty.com/docs/generating-api-keys)
     1. Populate an environment variable with the API key. You can pass the environment variable name via CLI with the `--pagerduty-api-key-env-var` parameter.
+    1. Populate an environment variable with the API session timeout. You can also set this via CLI with `--pagerduty-request-timeout` parameter.
```
ryan-lane, line 10:
```suggestion
    1. You can set the timeout for pagerduty requests via the CLI with `--pagerduty-request-timeout` parameter.
```

ryan-lane (approved):

ryan-lane:
Thanks for the PR!

Issue #994, opened 2022-09-28
Incorrect order for exception checking for function _is_common_exception in s3.py


trodery:
The function `_is_common_exception` in `s3.py` checks for the string `NoSuchBucket` in the exception prior to checking for the string `NoSuchBucketPolicy`. Since `NoSuchBucket` is a substring of `NoSuchBucketPolicy`, `NoSuchBucket` will always be matched.

The current function from [https://github.com/lyft/cartography/blob/master/cartography/intel/aws/s3.py#L181](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/s3.py#L181)

```python
@timeit
def _is_common_exception(e: Exception, bucket: Dict) -> bool:
    error_msg = "Failed to retrieve S3 bucket detail"
    if "AccessDenied" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - Access Denied")
        return True
    elif "NoSuchBucket" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - No Such Bucket")
        return True
    elif "AllAccessDisabled" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - Bucket is disabled")
        return True
    elif "EndpointConnectionError" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - EndpointConnectionError")
        return True
    elif "NoSuchBucketPolicy" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - NoSuchBucketPolicy")
        return True
    elif "ServerSideEncryptionConfigurationNotFoundError" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - ServerSideEncryptionConfigurationNotFoundError")
        return True
    elif "InvalidToken" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - InvalidToken")
        return True
    elif "NoSuchPublicAccessBlockConfiguration" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - NoSuchPublicAccessBlockConfiguration")
        return True
    elif "IllegalLocationConstraintException" in e.args[0]:
        logger.warning(f"{error_msg} for {bucket['Name']} - IllegalLocationConstraintException")
        return True
    return False
```

trodery:
Fixed in #995

PR #995, opened 2022-09-28
Fix exception checking order

trodery:
Fix exception checking order in function `_is_common_exception` so that `NoSuchBucketPolicy` is checked before `NoSuchBucket`


trodery:
This fixes issue #994

trodery:
The CLA has definitely been signed. I'm not sure why it says it isn't.

ramonpetgrave64:
It's working now.

PR #1007, opened 2022-10-07
Fixes #1006

### LINKED ISSUE #1006: relationship between lambda and AWSPrincipal is always deleted
*Title*: relationship between lambda and AWSPrincipal is always deleted

*Description*:
Currently when lambda sync is done, it creates an edge between the Lambda and the AWS Role this lambda would run as (STS_ASSUME_ROLE_ALLOW). 
In https://github.com/lyft/cartography/blob/master/cartography/intel/aws/lambda_function.py (line:71)
WITH lambda, lf
        MATCH (role:AWSPrincipal{arn: lf.Role})
        MERGE (lambda)-[r:STS_ASSUME_ROLE_ALLOW]->(role)
        SET r.lastupdated = $aws_update_tag

However, immediately after the sync, cleanup_lambda is called:
def cleanup_lambda(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
    run_cleanup_job('aws_import_lambda_cleanup.json', neo4j_session, common_job_parameters)

which seems to unconditionally delete the relationship:
    {
      "query": "MATCH (:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(:AWSLambda)-[r:STS_ASSUME_ROLE_ALLOW]->(:AWSPrincipal) WITH r LIMIT $LIMIT_SIZE DELETE r",
      "iterative": true,
      "iterationsize": 100
    }

*To Reproduce*:
After processing if you query for relationship between lambda and principal, you will get 0 result
MATCH (:AWSLambda)-[r:STS_ASSUME_ROLE_ALLOW]->(:AWSPrincipal)
return r

Is this deliberate ?

VishalAgarW:
Resolve relationship between lambda and AWSPrincipal is always deleted #1006

### DIFF
--- a/cartography/data/jobs/cleanup/aws_import_lambda_cleanup.json
+++ b/cartography/data/jobs/cleanup/aws_import_lambda_cleanup.json
@@ -31,7 +31,7 @@
       "iterationsize": 100
     },
     {
-      "query": "MATCH (:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(:AWSLambda)-[r:STS_ASSUME_ROLE_ALLOW]->(:AWSPrincipal) WITH r LIMIT $LIMIT_SIZE DELETE r",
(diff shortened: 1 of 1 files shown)


ramonpetgrave64:
Thank you!

ramonpetgrave64 (approved):

PR #1008, opened 2022-10-08
crowdstrike: remove arbitrary filter and limit

juju4:
Minor change to remove hardcoded settings that likely don't fit many environments


juju4:
Any comment on this one?

juju4:
it's possible to add a cli param but usually you would want to get all assets. imho, more a edge case

achantavy (commented):
@juju4 - sorry for the extremely delayed review, I'm finally able to catch up a bit.

From my brief read, it seems like endpoints.py is written specifically to load in CrowdstrikeHosts assuming that they are not on AWS EC2. If we remove the filter, then will the rest of the code still work? Will it fail if there are CrowdstrikeHosts that are not on AWS EC2?

Again, I'm not familiar with Crowdstrike, just doing a read and seeing that the filter may have been put there for a reason.

juju4:
Example filter from falconpy docs
https://www.falconpy.io/Service-Collections/Hosts.html#available-filters

empty filter is valid meaning no filter. It is working, used in production.
existing code restricts to 400 devices and only AWS_EC2 provider

juju4:

achantavy (approved):
Approving, let's open up an issue to expose the parameters in config.py.

chandanchowdhury (approved):

PR #1010, opened 2022-10-08
Add Rapid7 InsightVM source

juju4:
Add Ingestion of Rapid7 InsightVM assets as source. (follow-up of https://github.com/lyft/cartography/issues/1000)
Include initial test data and draft schema

Bugs
* fix import of array and dict, or needed values (addresses, configurations...) to get instance_id, subscription_d and resource_id which are required to make relationship with AzureVirtualMachine.
* fix mypy issues.

Pending
* AzureVirtualMachine relationship.
* import vulnerabilities' detections.
* may want cleanup retention to be configurable. Typically, EDR keep inventories for 30-45 days.


ramonpetgrave64:
This is really cool, and thank you!
I can see that this is still draft from you, but I'll try to comment more closely this week.

juju4:
Like mde, good for review and merge.

cartography/data/jobs/rapid7_import_cleanup.json
```diff
@@ -0,0 +1,10 @@
+{
```
ramonpetgrave64, line 1: duplicate file

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
...
+    Extract from rapid7 history, first seen/scan time
+    """
+    firstseen = history[0]["date"]
+    for entry in history:
+        if firstseen > entry["date"]:
+            firstseen = entry["date"]
+    return firstseen
+
+
+# pylint: disable=too-many-arguments,too-many-locals
```
ramonpetgrave64, line 171: is this necessary?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
...
+
+            data = resp.json()
+
+            total_resources = data["page"]["totalResources"]
+            if total_resources < limit:
+                limit = total_resources
+
+        except requests.HTTPError as exception:
+            logger.warning("GetAssets0: %s", resp.content)
+            logger.exception("GetAssets0: %s", exception)
```
ramonpetgrave64, line 235: Do you mean to have both of these log messages?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
...
+            if "resources" not in data:
+                logger.warning("data without resources: %s", data)
+
+        except Exception as exception:
+            logger.warning("GetAssets1: %s", resp.content)
+            logger.exception("GetAssets1 - Incomplete data: %s", exception)
+            return result_array
+
+        try:
+            # logger.debug("GetAssets count %s", len(data['resources']))
```
ramonpetgrave64, line 248:
```suggestion
```

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    firstseen = history[0]["date"]
+    for entry in history:
+        if firstseen > entry["date"]:
+            firstseen = entry["date"]
+    return firstseen
+
+
+# pylint: disable=too-many-arguments,too-many-locals
+def Rapid7Hosts(
```
ramonpetgrave64, line 172: use snake case

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    Get Rapid7 vulnerabilities
+
+    https://help.rapid7.com/insightvm/en-us/api/index.html#operation/getVulnerabilities
+    Vulnerabilities per host are returned with hosts
+    """
+
+    # TBD
+
+    return ["TBD"]
```
ramonpetgrave64, line 303: Still TBD?

cartography/intel/rapid7/vulnerabilities.py
```diff
@@ -0,0 +1,109 @@
...
+import neo4j
+
+from .util import Rapid7Hosts
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def sync_vulnerabilities(
```
ramonpetgrave64, line 14: Because `Rapid7VulnerabilitiesByHost` is still "TBD", most of this file is still unreachable code, right?

docs/schema/rapid7.md
```diff
@@ -0,0 +1,59 @@
+# Cartography - Rapid7 InsightVM Schema
```
ramonpetgrave64, line 1: move this file to docs/root/modules

docs/schema/rapid7.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Table of contents
+
+- [Rapid7Host](#rapid7host)
+
+## Rapid7Host
+
+Placeholder representation of a single [Rapid7 Host or asset](https://help.rapid7.com/insightvm/en-us/api/index.html#operation/getAssets). This node is the minimal data necessary to map an asset.
+
+Warning! Work In Progress.
```
ramonpetgrave64, line 15: Still work in progress?

docs/schema/rapid7.md
```diff
@@ -0,0 +1,59 @@
...
+* Azure Tenant contains one or more Rapid7 Hosts.
+```
+(AzureTenant)-[RESOURCE]->(Rapid7Host)
+```
+* Azure Subscription contains one or more Rapid7 Hosts.
+```
+(AzureSubscription)-[RESOURCE]->(Rapid7Host)
+```
+* Azure Virtual Machine is one single Rapid7 Host.
+* Similarly for other cloud providers and Onpremises.
```
ramonpetgrave64, line 59: Still TBD? Your changes do not create these relationships.

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
+"""
+cartography/intel/rapid7/util
+"""
+# pylint: disable=invalid-name,broad-except
+import json
+import logging
+import re
+from array import array
+
+import pandas
```
ramonpetgrave64, line 10: You'll need to add pandas as a requirement in setupp.py. But please try to accomplish this without using pandas.

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
+"""
+cartography/intel/rapid7/util
+"""
+# pylint: disable=invalid-name,broad-except
```
ramonpetgrave64, line 4: is this needed?

ramonpetgrave64 (changes requested):

.github/workflows/test_suite.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - master
+      - devel*
```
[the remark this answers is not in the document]
juju4: this  allows testing in my own repo and devel* branches independently of the PR.
Not sure how testing your own devel branch.

Removed

cartography/intel/rapid7/util.py
juju4: As said above, I reviewed files with pylint and black.
This avoid triggering pylint errors.

cartography/intel/rapid7/util.py
juju4: rapid7 API content is usually huge. that's one reason that I split it from exception message itself.

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,303 @@
...
+
+        except Exception as exc:
+            logger.warning("GetAssets3: %s", resp.content)
+            logger.exception("GetAssets3 - Incomplete data: %s", exc)
+            return result_array
+
+    # logger.debug("result_array %s", (result_array))
+
+    logger.info("Flattening data")
+    df_rapid7_tmp = pandas.json_normalize(result_array, sep="_", max_level=3)
```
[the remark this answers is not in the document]
juju4: I would prefer to keep pandas.
IMHO, that makes things simpler and more readable.
Also, I use it also for other sources ingestion to come

cartography/intel/rapid7/util.py
juju4: Function removed as the host vulnerabilities part is not on my priorities for now.

docs/schema/rapid7.md
juju4: On relationships, I would be interested into some advices.
It does not seem most of the sources of cartography are doing cross-sources relations.
But this is an important point to me.
- link Cloud VM with tools present on it.
- problem, tool may have not always the information that makes relation certain. In case of Azure
* best case is to have resourceid (mde, rapid7 but does not seem always retrieved)
* crowdstrike gets only subscriptionid, resourcegroup and hostname
* worst case, only have hostname

I started to make AzureVM relation based on resourceid and where not available AzureSubscription based on subscriptionid.
I removed the latter as I found it making graph less readable.
I was thinking to use PRESENT_IN if relation based on resourceid and MAYBE_PRESENT_IN if relation based just on hostname.

What do you think?

cartography/intel/rapid7/util.py
juju4: same than above. pylint usage (docstring, silence errors)

juju4:
As said, would need a bit of help on mypy error as less familiar with the tool.

Also, for some reason, CLA check is not green, but when going to https://oss.lyft.com/cla, it is marked signed and current. I did it a few weeks ago.

juju4:
Ah,CLA check is green now

cartography/intel/rapid7/util.py
juju4: is this fine or should I remove those?

cartography/intel/rapid7/util.py
juju4: input?

docs/schema/rapid7.md
juju4: feedback on this?

juju4:
on my side, linter looks all good.
only pyupgrade is failing but not related to code and did recent change to 2.31. 0 (3.2.0 works ok for me)

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,292 @@
...
+        try:
+            match = re.search(
+                "/subscriptions/([0-9a-f-].*?)/resourceGroups/",
+                resourceid,
+                flags=re.IGNORECASE,
+            )
+            if match:
+                sub = match.group(1)
+                return sub
+        except AttributeError as exception:
```
ramonpetgrave64, line 32: I don't think you need to catch this exception, since you already check if `match` is not `None`.

cartography/intel/rapid7/util.py
ramonpetgrave64: These concerns do seem valid, so please address them.

cartography/cli.py
```diff
@@ -393,6 +393,39 @@ def _build_parser(self):
...
+        )
+        parser.add_argument(
+            '--rapid7-server-url',
+            type=str,
+            default=None,
+            help=(
+                'The url of the Rapid7 InsightsVM server.'
+            ),
+        )
+        # FIXME! not passed correctly
```
ramonpetgrave64, line 420: fixme

cartography/cli.py
```diff
@@ -537,6 +570,15 @@ def main(self, argv: str) -> int:
                 ' and this option will be removed when code hard-coded syntax upgrades are completed',
             )
 
+        # Rapid7 config
```
ramonpetgrave64, line 573:
```suggestion
```

cartography/data/jobs/cleanup/rapid7_import_cleanup.json
```diff
@@ -0,0 +1,10 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (h:Rapid7Host) WHERE h.lastupdated <> $UPDATE_TAG WITH h LIMIT $LIMIT_SIZE DETACH DELETE (h)",
+      "iterative": true,
+      "iterationsize": 100
+    }
+  ],
+  "name": "cleanup rapid7"
+}
```
ramonpetgrave64, line 10: Also cleanup relationships between rapid7 and azure.

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,292 @@
...
+                azure = json.loads(line["value"])
+            else:
+                azure = line["value"]
+            instance_id = azure["instanceId"]
+
+    logger.warning("extract_rapid7_configurations_instanceid: %s", instance_id)
+    return instance_id
+
+
+def extract_rapid7_history_lastseen(history: list) -> str:
```
ramonpetgrave64, line 157: You can replace this with something like 
`min(entry['date'] for entry in history)`

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,292 @@
...
+    Extract from rapid7 history, last seen/scan time
+    """
+    lastseen = history[0]["date"]
+    for entry in history:
+        if lastseen < entry["date"]:
+            lastseen = entry["date"]
+    return lastseen
+
+
+def extract_rapid7_history_firstseen(history: list) -> str:
```
ramonpetgrave64, line 168: You can replace this with something like
max(entry['date'] for entry in history)

cartography/intel/rapid7/util.py
ramonpetgrave64: Try to address the pylint issues.

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def extract_rapid7_configurations_resourcegroup(configurations: Any[str, list]) -> str:
+    """
+    Extract from json, specific keyvalue pair: azure resource group
+
+    Input: dataframe row configurations column
+    """
+
+    if configurations is None or not isinstance(configurations, list):
```
ramonpetgrave64, line 92: Does the data returned from rapid7s api sometimes have empty fields?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,292 @@
...
+        return ""
+
+    rg = ""
+    for line in configurations:
+        if line["name"] == "azure":
+            logger.warning("line: %s", line)
```
ramonpetgrave64, line 98: Should this be a debug message, instead?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,292 @@
...
+
+    instance_id = ""
+    for line in configurations:
+        if line["name"] == "azure":
+            logger.warning("line: %s", line)
+            if isinstance(line["value"], str):
+                azure = json.loads(line["value"])
+            else:
+                azure = line["value"]
+            instance_id = azure["instanceId"]
```
ramonpetgrave64, line 151: Will and `"azure"` value always be an instance?

ramonpetgrave64 (changes requested):
Please do try to address the pylint comments.

cartography/intel/rapid7/util.py
juju4: done

cartography/cli.py
juju4: at this point, I'm not sure why not passed.

cartography/intel/rapid7/util.py
juju4: yes

cartography/intel/rapid7/util.py
juju4: if azure type, I believe there will be instanceid
but not all azure resources are correctly marked azure type

juju4:
I also have a mypy error that I'm not sure how to address
```
cartography/intel/rapid7/util.py:171: error: Unsupported target for indexed assignment ("Tuple[str, str, str, bool]")  [index]
```

docs/schema/rapid7.md
ramonpetgrave64: I don't have much context on how Azure works. But our preference is to only be certain about the relationships we create.

cartography/intel/rapid7/util.py
ramonpetgrave64: What kinds of situations would cause this HTTPError? Are you only meaning to cover regular network errors? If so, please remove this. But if you must log an exception, try to come up with a better message, rather than logging the entire data object.

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+    for line in configurations:
+        if line["name"] == "azure":
+            logger.debug("line: %s", line)
+            if isinstance(line["value"], str):
+                azure = json.loads(line["value"])
+            else:
+                azure = line["value"]
+            resource_group = extract_rg_from_resourceid(azure["resourceId"])
+
+    logger.warning("extract_rapid7_configurations_resourcegroup: %s", resource_group)
```
ramonpetgrave64, line 98: Why is this a warning? If resouregroup would produce too many messages, then perhaps it's more appropriate to have this as a debug message.

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+    sort: str = "",
+    nexpose_timeout: int = 5,
+) -> List:
+    """
+    Get Nexpose inventory full data to array
+    https://help.rapid7.com/insightvm/en-us/api/index.html#operation/getAssets
+
+    To filter, use https://help.rapid7.com/insightvm/en-us/api/index.html#operation/findAssets
+
+    Warning! if very large assets count (>10k, likely depending on assets details), it
```
ramonpetgrave64, line 163: How much memory in megabytes could this be?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+            else:
+                azure = line["value"]
+            subscription_id = extract_sub_from_resourceid(azure["resourceId"])
+            logger.warning(
+                "extract_rapid7_configurations_subscriptionid0: %s (%s)",
+                subscription_id,
+                azure["resourceId"],
+            )
+
+    logger.warning("extract_rapid7_configurations_subscriptionid: %s", subscription_id)
```
ramonpetgrave64, line 74: Should this be a warning?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+    Input: dataframe row configurations column
+    """
+
+    if configurations is None or not isinstance(configurations, list):
+        return ""
+
+    subscription_id = ""
+    for line in configurations:
+        if line["name"] == "azure":
+            logger.warning("line: %s", line)
```
ramonpetgrave64, line 62: Should this be a warning?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+            if isinstance(line["value"], str):
+                azure = json.loads(line["value"])
+            else:
+                azure = line["value"]
+            resource_id = azure["resourceId"]
+
+    logger.warning("extract_rapid7_configurations_resourceid: %s", resource_id)
```
ramonpetgrave64, line 122: Should this be a warning?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+    Input: dataframe row configurations column
+    """
+
+    if configurations is None or not isinstance(configurations, list):
+        return ""
+
+    resource_id = ""
+    for line in configurations:
+        if line["name"] == "azure":
+            logger.warning("line: %s", line)
```
ramonpetgrave64, line 115: Should this be a warning?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+    Input: dataframe row configurations column
+    """
+
+    if configurations is None or not isinstance(configurations, list):
+        return ""
+
+    instance_id = ""
+    for line in configurations:
+        if line["name"] == "azure":
+            logger.warning("line: %s", line)
```
ramonpetgrave64, line 139: Should this be a warning?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+            if isinstance(line["value"], str):
+                azure = json.loads(line["value"])
+            else:
+                azure = line["value"]
+            instance_id = azure["instanceId"]
+
+    logger.warning("extract_rapid7_configurations_instanceid: %s", instance_id)
```
ramonpetgrave64, line 146: Should this be a warning?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,246 @@
...
+    subscription_id = ""
+    for line in configurations:
+        if line["name"] == "azure":
+            logger.warning("line: %s", line)
+            if isinstance(line["value"], str):
+                azure = json.loads(line["value"])
+            else:
+                azure = line["value"]
+            subscription_id = extract_sub_from_resourceid(azure["resourceId"])
+            logger.warning(
```
ramonpetgrave64, line 68: Should this be a warning?

ramonpetgrave64 (changes requested):

cartography/intel/rapid7/util.py
juju4: Even w 48GB RAM, only up to 13k...
Added downloadreport api or local report options to avoid this

cartography/cli.py
juju4: fixed by using string type for cli argument. not sure why bool not passed correctly.

docs/schema/rapid7.md
juju4: the problem is always the same when dealing with multiple tools. is there a reliable key to use to join?
Azure resourceid could be but sadly not fetched reliably by most tools, at least in my context.
hostname/short_hostname is the default join key but hardly reliable (hostname reused, different case depending on tool, different length enforcement depending on tool - some enforce 15-chars windows limit...)

At this point, I'm plan to make relation base on resourceid or hostname/short_hostname.

cartography/intel/rapid7/util.py
juju4: removed resp.content

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,445 @@
...
+        nexpose_verify_cert2 = False
+        logger.warning("Requested to not verify certificate (%s)", nexpose_verify_cert)
+    else:
+        nexpose_verify_cert2 = True
+
+    count = total_resources = 0
+    size_interval = 500
+    result_array: List[Any] = []
+    df_rapid7_tmp = pandas.DataFrame()
+    while count < limit:
```
ramonpetgrave64, line 184: Why do we have a limit? Is it because of API ratelimitting? If that's the case, consider using the `@backoff` decorator.
https://github.com/lyft/cartography/blob/master/cartography/util.py#L206

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,445 @@
...
+
+    nexpose_verify_cert = authorization[3]
+    logger.info("nexpose_verify_cert: %s", nexpose_verify_cert)
+    if nexpose_verify_cert.lower() != "true":
+        nexpose_verify_cert2 = False
+        logger.warning("Requested to not verify certificate (%s)", nexpose_verify_cert)
+    else:
+        nexpose_verify_cert2 = True
+
+    count = total_resources = 0
```
ramonpetgrave64, line 180: It looks like `count` is unused.

cartography/intel/rapid7/util.py
ramonpetgrave64: Do we still need this in the docstring?

cartography/intel/rapid7/util.py
```diff
@@ -0,0 +1,445 @@
...
+                limit = data["page"]["totalResources"]
+
+        except requests.HTTPError as exception:
+            logger.exception("GetAssets0: %s", exception)
+            return result_array
+
+        if "resources" not in data:
+            logger.warning("data without resources: %s", data)
+
+        result_array = result_array + data["resources"]
```
ramonpetgrave64, line 222:
```suggestion
        result_array.extend(data["resources"])
```

ramonpetgrave64 (changes requested):

docs/schema/rapid7.md
juju4: more to add here?

cartography/intel/rapid7/util.py
juju4: fine to resolve?

cartography/intel/rapid7/util.py
juju4: No API rate-limiting AFAIK, more response body size and experience that it is more likely to fail if more depending on context. at least, not on doc https://help.rapid7.com/insightvm/en-us/api/index.html#operation/findAssets

Another reason to use the report option.

docs/root/modules/rapid7/rapid7.md
```diff
@@ -0,0 +1,74 @@
...
+JOIN fact_asset fa USING (asset_id)
+JOIN dim_operating_system dos USING (operating_system_id)
+JOIN dim_asset_unique_id daui USING (asset_id)
+```
+
+| Field | Description |
+|-------|--------------|
+| firstseen| Timestamp of when a sync job first discovered this node  |
+| lastupdated |  Timestamp of the last time the node was updated |
+| r7_id | id |
```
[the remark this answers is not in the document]
juju4: IMHO, it depends on each product which may have different output.
Typically for IP/MAC address, if multiple cards, there is no certainty that each product will pick the same. same for os product, version... no shared taxonomy sadly.

if still want to remove prefix, please say if everywhere or only part of fields:

tests/integration/cartography/rapid7/test_endpoints.py
```diff
@@ -0,0 +1,30 @@
...
+    expected_nodes = {
+        (
+            "282"
+        ),
+    }
+
+    nodes = neo4j_session.run(
+        """
+        MATCH (n:Rapid7Host)
+        RETURN n.id
```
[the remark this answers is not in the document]
juju4: did a quick attempt. testing needed

juju4:
Currently, this can get data from
- API getAssets
- local report file
- remote report file directly from vulnerability management server

I don't know if should keep the API one which has limitations mostly for large inventory.
After, there are probably some cases where it is better than report.
On my side, using remote report file now in a satisfying manner.

Mypy typing errors that I didn't solve (or was impacting functionalities), all related to extract_rapid7_configurations_* functions.
https://github.com/juju4/cartography/actions/runs/3862847617/jobs/6584667101#step:4:67
```
cartography/intel/rapid7/util.py:51: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
cartography/intel/rapid7/util.py:80: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
cartography/intel/rapid7/util.py:104: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
cartography/intel/rapid7/util.py:128: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
Found 4 errors in 1 file (checked 441 source files)
```

juju4:
outside of the points left above for which I would like inputs too.

PR #1012, opened 2022-10-20
Bor 3134 modify cartography to support updation of specific resource

ashwin-borneo:


ramonpetgrave64:
Thank you for this. We've been thinking about this idea, to sync an update a single node in the graph. Can you write a description or document explaining your approach?

cartography/intel/aws/resources.py
```diff
@@ -1,28 +1,12 @@
...
-from . import resourcegroupstaggingapi
-from . import route53
-from . import s3
-from . import secretsmanager
-from . import securityhub
-from . import sqs
-from . import ssm
+from cartography.intel.aws import s3_list_only
+
+from . import (apigateway, config, dynamodb, ecr, ecs, eks, elasticache,
```
ramonpetgrave64, line 5: Please avoid these kinds of format changes

PR #992, opened 2022-09-20
Bug fix #991

sachafaust:

PR #1019, opened 2022-11-04
Neo4j 4.x: disable auto-patcher, fix missed syntax changes

ramonpetgrave64:
Final wrapup of #914.
Here, we remove the autopatcher, and convert the last few bits of syntax that were missed after #992.


achantavy (approved):
Huzzah!

Issue #1024, opened 2022-11-08
Standardize and make ID fields consistent


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

As mentioned in #910, cartography node fields are inconsistent. For example,

>  in an AWSAccount you'd need to use id, whereas an AWSUser has a userid and arn but not an id.

*Please complete the following information:*:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]
0.65.0

> Add any other context about the problem here

This blocks #910.

danielsaporo:
I think the first step should be to list all the different node types and identify which property to use as a unique ID for each.

juju4:
consistency also applies to other field/tools.
hostname/short_hostname can have different case. hostname can be short or long depending on tool, or truncated to 15 characters for windows per netbios limitations)
mac address can be written in a variety of way too.
tags can be values or key-values

gokulyc:
Is it possible give unique id for all nodes & add separate field which describes that node(label/ description).

resilience-jychp:
@achantavy one solution for this problem (and especially #910) is to use computed community ID.

With new datamodel it will be easy to implement:
- add a flag for Field to indicate a field is part of community ID OR add a property to NodeSchema that list all property fields that are part of community ID
- when generating insert query compute this field value and hash the result (ex : concat field value and get the md5 hash of this concatenated value) and insert it under "community_id" field

I used this method in my previous job to handle partial node from network logs.

Today I use this to extract inventories from Cartography

achantavy:
As a brief update, I'd like to refactor more AWS assets to the new data model before working on ID standardization. Once things are in the new model, the standardization task will be a lot more straightforward and less error prone.

Issue #349, opened 2020-07-13
Design a way to more quickly write AWS sync plugins


achantavy:
## Design a way to more quickly write AWS sync plugins

*Description*:

> Describe your idea.  Please be detailed.  If a feature request, please
describe the desired behavior, what scenario it enables, and how it
would be used.

### Background

An intel module needs to perform a common set of steps:

1. Get data with a certain set of fields from an external API

2. Perform a set of data transformations such as generating unique IDs if vendor-provided ones (e.g. AWS ARNs) are not available.

3. Load new nodes and relationships to the graph and connect them to all related nodes such as AWS accounts and others

4. Make indexes for the new node types

5. Create integration tests to ensure the nodes and connections are loaded as expected

6. Update our schema documents

This is a lot of boilerplate code to write and there are many repeated sections across intel modules. 

### The idea

To improve our developer experience, we could create an abstraction layer that accepts as input what data types the developer wants to retrieve and their expected forms in the graph (steps 1-4). This will greatly improve our velocity on adding new plugins.

This issue will initially be scoped to AWS syncs and we will revisit this approach for the other modules.

### The task

Write a specification document on what this abstraction interface could look like and submit requests for comment.

achantavy:
Done! We will need to refactor existing code using #1038, but we have the capability now!

PR #1038, opened 2022-11-28
#349: Generalize graph write queries

achantavy:
Addresses #349, also a good step working toward #1024.

- Provides module authors with an automated way to load data to the graph so that they don't have to hand write Neo4j queries. Makes the process significantly less error prone by generating queries, and helps with reliability because transactions and batching are abstracted away.
- Defines schema objects so that we can automate index creation and cleanup job creation in a future PR.
- Demonstrates this functionality by having the AWS EMR sync adopt it.
- Allows module authors to connect relationships to nodes using one or more attributes.

Why not use an existing ORM? Existing ones:
- don't use `UNWIND` + `MERGE` for speed and batching
- don't allow configuring Neo4j driver details such as connection timeouts, which are absolutely necessary if you must send traffic across a load balancer
- don't run Neo4j managed transaction functions for automatic retries
- in some cases use their own custom driver (or even require you to run a separate webserver for IPC!) and not the official Neo4j driver


.pre-commit-config.yaml
```diff
@@ -39,7 +39,7 @@ repos:
     -   id: reorder-python-imports
         args: [--py3-plus]
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.950
+    rev: v0.981
```
achantavy, line 42: Bumping this to 0.981 because we need this fix: https://github.com/python/mypy/pull/13398

cartography/intel/aws/emr.py
```diff
@@ -48,53 +57,76 @@ def get_emr_describe_cluster(boto3_session: boto3.session.Session, region: str,
...
 def load_emr_clusters(
-    neo4j_session: neo4j.Session, cluster_data: List[Dict], region: str, current_aws_account_id: str,
-    aws_update_tag: int,
+        neo4j_session: neo4j.Session,
+        cluster_data: List[Dict],
+        region: str,
+        current_aws_account_id: str,
+        aws_update_tag: int,
 ) -> None:
-    query = """
```
achantavy, line 56: This PR refactors the EMR sync to demonstrate using the new functionality. Rather than hand write queries, module authors now define schema objects and call `build_ingestion_query()` and `load_graph_data()`.

cartography/graph/model.py
```diff
@@ -0,0 +1,234 @@
...
+        If False, looks for the property in the data dict.
+        Defaults to False.
+        """
+        self.name = name
+        self.set_in_kwargs = set_in_kwargs
+
+    def _parameterize_name(self) -> str:
+        return f"${self.name}"
+
+    def __repr__(self) -> str:
```
kunaals, line 59: is "item" meaningful in someway here as the prefix?

cartography/graph/model.py
```diff
@@ -0,0 +1,234 @@
...
+@dataclass
+class CartographyNodeProperties(abc.ABC):
+    """
+    Abstract base dataclass that represents the properties on a CartographyNodeSchema. This is intended to enforce that
+    all subclasses will have an id and a lastupdated field defined on their resulting nodes.
+    """
+    id: PropertyRef = field(init=False)
+    lastupdated: PropertyRef = field(init=False)
+
+    def __post_init__(self):
```
skiptomyliu, line 72: I don't think this is needed, it should throw a TypeError by default

```python
from abc import ABC, abstractmethod

class Giraffe(ABC):
    @abstractmethod
    def eat(self):
        pass
a = Giraffe()
```

```
  File "/private/tmp/test.py", line 7, in <module>
    a = Giraffe()
TypeError: Can't instantiate abstract class Giraffe with abstract method eat
```

ramonpetgrave64 (changes requested):
Still not yet finished reviewing, and I want to add a helper function to make it a bit easier to create a `CartographyNodeSchema `. like this, using the `types` library. But this can come later.

```
create_node_schema(
  label='EMRCluster',
  properties=create_node_properties(
    arn='ClusterArn',
    auto_terminate='AutoTerminate',
    ...
  ),
  ...
)
```

ramonpetgrave64 (approved):
Looks great! The type tricks I mentioned can come later.

tests/unit/cartography/graph/test_querybuilder_complex.py
```diff
@@ -0,0 +1,139 @@
...
+        ],
+    )
+
+
+def test_build_ingestion_query_complex():
+    # Act
+    query = build_ingestion_query(InterestingAssetSchema())
+
+    expected = """
+        UNWIND $DictList AS item
```
achantavy, line 103: @csanders-git brought up a good point - how do we handle the case where a given node only attaches to another given node _sometimes_ -- depending on the order, a null match will cause the other links to not be created.

ramonpetgrave64: Then we might use the `OPTIONAL MATCH`, but I'm not 100% sure yet.

cartography/graph/model.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class PropertyRef:
+    """
+    PropertyRefs represent properties on cartography nodes and relationships.
+
+    cartography takes lists of Python dicts and loads them to Neo4j. PropertyRefs allow our dynamically generated Neo4j
+    ingestion queries to set values for a given node or relationship property from (A) a field on the dict being
+    processed (PropertyRef.override=False, default), or (B) from a single variable provided by a keyword argument
+    (PropertyRef.override=True).
```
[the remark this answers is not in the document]
achantavy: Good catch, thank you

cartography/intel/aws/emr.py
```diff
@@ -48,53 +57,76 @@ def get_emr_describe_cluster(boto3_session: boto3.session.Session, region: str,
...
+    target_node_key_property_ref: PropertyRef = PropertyRef('AccountId', set_in_kwargs=True)
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: EMRClusterToAwsAccountRelProperties = EMRClusterToAwsAccountRelProperties()
+
+
+@dataclass
+class EMRClusterSchema(CartographyNodeSchema):
+    label: str = 'EMRCluster'
+    properties: EMRClusterNodeProperties = EMRClusterNodeProperties()
```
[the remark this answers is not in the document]
achantavy: Shouldn't we typehint the specific type and not the generic one?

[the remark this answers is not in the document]
ramonpetgrave64: I think it's helpful just as a clue to devs on what the supertype should be. When editing in an IDE, though, the effect is the same.

tests/unit/cartography/graph/test_querybuilder_complex.py
achantavy: I figured out how to do this using `CALL{ query1 UNION query2 UNION ... queryN }`

cartography/intel/aws/emr.py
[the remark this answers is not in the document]
achantavy: By doing this we're saying that `EMRClusterSchema.properties` will accept any `CartographyNodeProperties` type, which is not correct, so I think we should keep this as-is.

tests/integration/cartography/graph/test_querybuilder_rel_subsets.py
```diff
@@ -0,0 +1,140 @@
...
+from cartography.graph.querybuilder import build_ingestion_query
+from tests.data.graph.querybuilder.sample_data.partial_relationships import INTERESTING_NODE_WITH_PARTIAL_RELATIONSHIPS
+from tests.data.graph.querybuilder.sample_data.partial_relationships import MERGE_SUB_RESOURCE_QUERY
+from tests.data.graph.querybuilder.sample_data.partial_relationships import MERGE_WORLD_ASSET_QUERY
+from tests.data.graph.querybuilder.sample_models.interesting_asset import InterestingAssetSchema
+
+
+def test_load_graph_data_subset_of_relationships(neo4j_session):
+    """
+    Test load_graph_data() if a schema defines multiple relationships but only a subset of them are present in our data.
```
ramonpetgrave64, line 11:
```suggestion
    Test load_graph_data() if a schema defines multiple relationships but only a subset of them are possible in our data.
```

tests/integration/cartography/graph/test_querybuilder_rel_subsets.py
```diff
@@ -0,0 +1,140 @@
...
+    assert actual == expected
+
+    # Assert that the InterestingAsset to HelloAsset relationship does NOT exist
+    expected = {
+        ('interesting-node-id', None),
+    }
+    result = neo4j_session.run(
+        """
+        MATCH (n1:InterestingAsset)
+        OPTIONAL MATCH (n1)<-[:ASSOCIATED_WITH]-(n2:HelloAsset)
```
ramonpetgrave64, line 54: I think you meant to write an outward relationship direction here. Still, we can rewrite it this way:
```suggestion
        OPTIONAL MATCH (n1)--(n2:HelloAsset)
```

tests/integration/cartography/graph/test_querybuilder_rel_subsets.py
ramonpetgrave64, line 11:
```suggestion
    Test load_graph_data() if a schema defines multiple relationships but only a subset of them are possible to create in our data.
```

tests/integration/cartography/graph/test_querybuilder_rel_subsets.py
```diff
@@ -0,0 +1,140 @@
...
```
ramonpetgrave64, line 117: Same here about the outward relationship

ramonpetgrave64 (changes requested):
Looks good, but that one test case needs to fix.

ramonpetgrave64 (approved):

PR #1050, opened 2022-12-16
generalizequerywrites: helper shortcuts

ramonpetgrave64:
Adds helper functions to more easily make the classes used in `cartography.graph.model`.
Recreates the AWSLambda ingestion query with these helper functions.

Later, we can add things like automatically including a `lastupdated` PropertyRef


achantavy:
This is cool. Can you also add to the PR description or a comment on what implementing this would look like without the helper function so that we can compare?

ramonpetgrave64:
> This is cool. Can you also add to the PR description or a comment on what implementing this would look like without the helper function so that we can compare?

Yes, For comparison, we can look at the `SimpleNodeWithSubResourceSchema` built here manually
https://github.com/lyft/cartography/blob/36c6ed3ba5540a8975b371230b047249c559cf08/tests/data/graph/querybuilder/sample_models/simple_node.py#L46
, versus built with the helper functions:
https://github.com/lyft/cartography/blob/3921739be77714b7866f1eccc7b4038e1da0853b/tests/unit/cartography/graph/test_schema_builder.py#L21

You don't save lines of code, but organizing the code is visually easier for some people.

achantavy:
I personally like the look of option 1 better because it feels more "declarative" to me. For example if I want to see in code how a `SimpleNodeWithSubResourceSchema` is defined, it makes sense to me that there is a class with that name.

Being able to specify any string as the class name in option 2 feels more error prone to me because for example if I'm someone who doesn't know the internals of how `build_node_schema()` or `build_rel_properties()` work, then outside of actually running the code I have no warnings about picking a `cls_name` that already exists.

Open to other thoughts here.

achantavy:
Let's close this one out for now. We talked about this last month and I think it makes more sense to have these classes statically defined. Happy to revisit at any time though.

PR #1052, opened 2022-12-19
Automatic cleanup jobs

achantavy:
Uses the schema in #1038 to automatically create cartography cleanup jobs.

Refactors AWS EMR sync to use this new method and adds tests.

## Before

```json
{
    "statements": [
        {
            "query": "MATCH (:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(n:EMRCluster) WHERE n.lastupdated <> $UPDATE_TAG WITH n LIMIT $LIMIT_SIZE DETACH DELETE (n)",
            "iterative": true,
            "iterationsize": 100
        },
        {
            "query": "MATCH (:AWSAccount{id: $AWS_ID})-[r:RESOURCE]->(:EMRCluster) WHERE r.lastupdated <> $UPDATE_TAG WITH r LIMIT $LIMIT_SIZE DELETE r",
            "iterative": true,
            "iterationsize": 100
        }
    ],
    "name": "cleanup EMRCluster"
}
```

```python
def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
    run_cleanup_job('aws_import_emr_cleanup.json', neo4j_session, common_job_parameters)
```

## Now

```python
def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
    cleanup_job = GraphJob.from_node_schema(EMRClusterSchema(), common_job_parameters)
    cleanup_job.run(neo4j_session)
```


cartography/graph/job.py
```diff
@@ -86,6 +124,36 @@ def from_json(cls, blob: str, short_name: Optional[str] = None) -> 'GraphJob':
         name = data["name"]
         return cls(name, statements, short_name)
 
+    @classmethod
+    def from_node_schema(cls, node_schema: CartographyNodeSchema, parameters: Dict[str, Any]) -> 'GraphJob':
```
ramonpetgrave64, line 128: rename to `cleanup_job_from_node_schema`

achantavy: There is also a `GraphJob.from_json_file()`, so I think we should follow the naming convention

ramonpetgrave64 (changes requested):

cartography/graph/job.py
```diff
@@ -1,21 +1,60 @@
...
+    """
+    :param queries: A list of Neo4j queries with parameters indicated by leading '$' like $this.
+    :return: The set of all parameters across all given Neo4j queries.
+    """
+    parameter_set = set()
+    for query in queries:
+        as_template = Template(query)
+        params = _get_identifiers(as_template)
+        for param in params:
+            parameter_set.add(param)
```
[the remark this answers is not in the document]
achantavy: Can't extend a set; only lists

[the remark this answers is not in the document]
ramonpetgrave64: `update`, sorry.

cartography/graph/cleanupbuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        # Ensure that the selected rels actually exist on the node_schema
+        sub_resource_rel, other_rels = filter_selected_relationships(node_schema, selected_rels)
+
+    if not sub_resource_rel:
+        raise ValueError(
+            "Auto-creating a cleanup job for a node_schema without a sub resource relationship is not supported. "
+            f'Please check the class definition of "{node_schema.__class__.__name__}". If the optional `selected_rels` '
+            'param was specified to build_cleanup_queries(), then ensure that the sub resource relationship is '
+            'present.',
+        )
```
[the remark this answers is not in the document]
achantavy: Btw that's a good point, I'll write an explicit test to make sure we don't accidentally delete the sub resource.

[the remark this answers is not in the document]
achantavy: This test exists here: https://github.com/lyft/cartography/pull/1052/files#diff-df8d4a7aff3c6a18912248c0096c3f74dd9c799c93ee9e265955e9fc58c39bbaR105-R114

cartography/graph/cleanupbuilder.py
```diff
@@ -0,0 +1,321 @@
...
+    if other_rels:
+        for rel in other_rels.rels:
+            result.append(
+                _build_cleanup_node_query(node_schema, rel),
+            )
+            result.append(
+                _build_cleanup_rel_query(node_schema, rel),
+            )
+
+    if sub_resource_rel:
```
ramonpetgrave64, line 61: No need for this conditional, since we would raise an exception first.

cartography/graph/cleanupbuilder.py
```diff
@@ -0,0 +1,321 @@
...
+        # Make sure that the sub resource one is last in the list; order matters.
+        result.append(
+            _build_cleanup_node_query(node_schema),
```
ramonpetgrave64, line 64: I think we should be explicit about cleaning up the `sub_resource_rel`, since we will always require the node to have it. And it would simplify the logic a bit in `_build_cleanup_node_query()`

cartography/graph/cleanupbuilder.py
```diff
@@ -0,0 +1,290 @@
...
+        )
+    return query_template.safe_substitute(
+        node_label=node_schema.label,
+        sub_resource_label=node_schema.sub_resource_relationship.target_node_label,
+        sub_res_link=sub_res_link,
+        sub_res_match_clause=_build_match_clause(sub_res_node_matcher),
+    )
+
+
+def _build_cleanup_rel_query(
```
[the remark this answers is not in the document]
ramonpetgrave64: `_build_cleanup_node_query()` and `_build_cleanup_rel_query()` can be combined into a `build_node_and_rel_cleanup_queries()`, since we would always do them in pairs. They could just return the two as `(node_cleanup_query, rel_cleanup_query)`

tests/integration/cartography/intel/aws/test_emr.py
```diff
@@ -62,3 +67,34 @@ def test_load_emr_clusters_relationships(neo4j_session):
...
+    cleanup(
+        neo4j_session,
+        {
+            'UPDATE_TAG': TEST_UPDATE_TAG + 1,  # Simulate that is a new sync run so the prev update tag is obsolete now
+            'AccountId': TEST_ACCOUNT_ID,
+        },
+    )
+
+    # Assert: Expect no EMR clusters in the graph now
+    assert check_nodes(neo4j_session, 'EMRCluster', ['arn']) == set()
```
ramonpetgrave64, line 100: nit: a better test would also load nodes that should not be cleaned up.

tests/integration/cartography/graph/test_cleanupbuilder.py
```diff
@@ -0,0 +1,120 @@
...
+from tests.data.graph.querybuilder.sample_data.helloworld_relationships import INTERESTING_NODE_WITH_ALL_RELS
+from tests.data.graph.querybuilder.sample_data.helloworld_relationships import MERGE_HELLO_ASSET_QUERY
+from tests.data.graph.querybuilder.sample_data.helloworld_relationships import MERGE_SUB_RESOURCE_QUERY
+from tests.data.graph.querybuilder.sample_data.helloworld_relationships import MERGE_WORLD_ASSET_QUERY
+from tests.data.graph.querybuilder.sample_models.interesting_asset import InterestingAssetSchema
+from tests.integration.util import check_nodes
+from tests.integration.util import check_rels
+
+
+def test_cleanup_interesting_asset_end_to_end(neo4j_session):
```
ramonpetgrave64, line 14: One more end-to-end that `SubResource`, `InterestingAsset`, and `HelloAsset` were updated, but not `WorldAsset`?

ramonpetgrave64 (approved):
Looks good, but I really would like to see _build_cleanup_node_query() and _build_cleanup_rel_query() combined into one function.

cartography/graph/cleanupbuilder.py
```diff
@@ -0,0 +1,290 @@
...
+    if selected_relationship.direction == LinkDirection.INWARD:
+        rel_to_delete_template = Template("<-[r:$RelToDeleteLabel]-")
+    else:
+        rel_to_delete_template = Template("-[r:$RelToDeleteLabel]->")
+    rel_to_delete = rel_to_delete_template.safe_substitute(RelToDeleteLabel=selected_relationship.rel_label)
+
+    # Ensure the node is attached to the sub resource and delete the relationship
+    query_template = Template(
+        """
+        MATCH (src:$node_label)$sub_resource_link(:$sub_resource_label{$match_sub_res_clause})
```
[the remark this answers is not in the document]
ramonpetgrave64: I see it's already account-specific.

tests/integration/cartography/graph/test_cleanupbuilder.py
```diff
@@ -0,0 +1,120 @@
...
+    # Arrange: add these nodes and rels to the graph, all with lastupdated=1:
+    # (i:InterestingAsset{id:'interesting-node-id'})<-[:RELATIONSHIP_LABEL]-(:SubResource{id:sub-resource-id});
```
achantavy, line 16: TODO: Can we make a test case using a real world example instead of these fake nodes so that the test is more understandable?

cartography/graph/cleanupbuilder.py
```diff
@@ -0,0 +1,321 @@
...
+        DELETE r;
+    ```
+
+    Example 2: with a selected relationship
+    _build_cleanup_rel_query(
+        InterestingAssetSchema(),
+        InterestingAssetToHelloAssetRel(),
+    )
+    -->
+        MATCH (src:InterestingAsset)<-[:RELATIONSHIP_LABEL]-(:SubResource{id: $sub_resource_id})
```
achantavy, line 209: TODO: sketch out unifying this function with build_cleanup_node_query

ramonpetgrave64 (changes requested):
Last thing, I think about the example path

ramonpetgrave64 (approved):
👏

PR #1053, opened 2022-12-21
Region Parallelism within single resource

harshagw:
Currently, for a single resource, data is being updated for each region one after another. The other way is by using threading to allow region parallelism within resources while different resources run sequentially.

I performed a little experiment to check if it saved time. The data is as follows: 

Screenshot 2022-12-21 at 10 46 10 AM

Result: For each resource, the mean time to run the sync takes approximately 10 times less if we parallelly update the data across regions. 

Update: We cannot use a neo4J session across multiple threads. I created a neo4JsessionFactory which can create a new session for each new thread.


ryan-lane:
I think it would generally be better to use asyncio rather than threading. It's a lot easier to reason about coroutines than dealing with the thread locking behavior.

Also, I think it's probably best to start with concurrency from the account level, then move down into the resource level. I do also agree that on a per-resource/per-region basis, we could potentially save a lot of time by also concurrently handling regions.

The biggest reason to start with per-account is that currently, each account runs sequentially, and there are some resources in an account that can take 20+ minutes (and can't be handled concurrently per-region). It's also at a much higher level and touches far less code, so it's easier to reason about.

achantavy:
Hey, thanks for sharing your approach and data. I have a few thoughts:

- I think we should use multiprocessing instead of threads because we'd run into the GIL in Python (although I am admittedly not the most informed on this).

- On error handling, yes, stopping execution altogether is expected because if there are errors that we cannot handle then we should not continue to change graph data.

- On a brief skim, the session factory looks good.

- Like Ryan said, we should do parallel accounts.

- The number of simultaneous parallel processes should be configurable.

cc: @kunaals

ryan-lane:
Multiprocessing is a lot more complex, won't speed things up much more than
green threads, and will eat a lot more memory. Depending on how cartography
is being run (single CPU in k8s) it could actually be a lot slower than
using green threads. The vast majority of the time spent in cartography is
IO, rather than CPU. Using some form of green threads will speed things up
a lot with considerably less memory overhead. Effectively, whenever one
green thread does IO, it'll yield to the other threads.

harshagw:
> Like Ryan said, we should do parallel accounts.

I don't have multiple accounts to test running different accounts parallelly. I don't want to submit any code without testing it.

achantavy:
Implementing mocks in unit tests would probably be good enough. But yeah, I don't think parallel regions is the way to go here.

achantavy:
@harshagw  Going to close this one out since we've decided to do parallel accounts and not parallel regions. I did cherry-pick the Neo4j session factory from your branch over to draft PR https://github.com/lyft/cartography/pull/1138 so your name is on the commits (thank you!) -- will continue the work there.

PR #1054, opened 2022-12-21
Update dev docs on how to use new object model

achantavy:
Update the how-to-write-an-intel-module docs using the new cartography object model introduced in #1038.

### DIFF
--- a/docs/root/dev/writing-intel-modules.md
+++ b/docs/root/dev/writing-intel-modules.md
@@ -1,13 +1,15 @@
 # How to write a new intel module
 
-This doc contains guidelines on creating a Cartography intel module. If you want to add a new data type to Cartography,
-this is the guide for you!  It is fairly straightforward to copy the structure of an existing intel module and test it,
-but we'll share some best practices in this doc to save you some time. We look forward to receiving your PR!
+If you want to add a new data type to Cartography, this is the guide for you. We look forward to receiving your PR!
 
 ## Before getting started...
 
 Read through and follow the setup steps in [the Cartography developer guide](developer-guide.html). Learn the basics of
 running, testing, and linting your code there.
+
+## The fast way
+
+To get started coding without reading this doc, just copy the structure of our [AWS EMR module](https://github.com/lyft/cartography/blob/master/cartography/intel/aws/emr.py) and use it as an example. For a longer written explanation of the "how" and "why", read on.
 
 ## Configuration and credential management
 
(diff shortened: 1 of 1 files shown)


docs/root/dev/writing-intel-modules.md
```diff
@@ -103,52 +232,12 @@ All of these queries use indexes for faster lookup.
...
-
-    - Every `MERGE` query that creates a new relationship should look like this
-
-        ```cypher
-        ON CREATE SET r.firstseen = $UpdateTag
-        SET
-        r.lastupdated = $UpdateTag
-        ```
-
-#### Connecting different node types with the `_attach` pattern
```
ramonpetgrave64, line 142: I feel like their may be cases where we want to attach after loading. But in the cases I can think of, it's still best to transform the data and use a NodeSchema that properly defines those extra relationships.

docs/root/dev/writing-intel-modules.md
```diff
@@ -54,19 +58,144 @@ For the sake of consistency, if a field does not exist, set it to `None` and not
...
+```
+
+A `CartographyNodeProperties` object consists of [`PropertyRef`](https://github.com/lyft/cartography/blob/e6ada9a1a741b83a34c1c3207515a1863debeeb9/cartography/graph/model.py#L37) objects. `PropertyRefs` tell `querybuilder.build_ingestion_query()` where to find appropriate values for each field from the list of dicts.
+
+For example, `id: PropertyRef = PropertyRef('Id')` above tells the querybuilder to set a field called `id` on the `EMRCluster` node using the value located at key `'id'` on each dict in the list.
+
+As another example, `region: PropertyRef = PropertyRef('Region', set_in_kwargs=True)` tells the querybuilder to set a field called `region` on the `EMRCluster` node using a keyword argument called `Region` supplied to `cartography.client.core.tx.load_graph_data()`. `set_in_kwargs=True` is useful in cases where we want every object loaded by a single call to `load_graph_data()` to have the same value for a given attribute.
+
+
+#### Defining relationships
```
ramonpetgrave64, line 129: I think it's worth adding an example here on the `other_relationships`, even if it's only pointing to the test cases we have.

ramonpetgrave64 (approved):

docs/root/dev/writing-intel-modules.md
achantavy: Good point, and yes that's the approach I would encourage.

It's also possible to first generate a query using `selected_relationships` and then load to that with additional data, but I don't have an example ready for that and that sounds complicated and I'm not ready to encourage that pattern in the doc yet.

ramonpetgrave64 (approved):

PR #1061, opened 2022-12-28
Fix #1059: Use arn as id for managed policies

### LINKED ISSUE #1059: cartography treats managed IAM policies as if they were inline policies
*Description*:

cartography treats IAM managed policies as if they were inline policies.

*To Reproduce*:

Run the cartography IAM sync.

*Screenshots*:

**Expected behavior**:

If I run `match(s:AWSPolicy{name:"policy-name"})--(p:AWSPrincipal) return *  limit 30;`, I should get a graph with a one-to-many relationship from policy to principals (assuming that 'policy-name' is of type 'managed' and not 'inline'). It should look like this:

```
  (     policy   )
  /        |     \
(prin1) (prin2) (prin3)
```

In total I should see 1 policy and N principals attached to it.

**Actual behavior**:

I run `match(s:AWSPolicy{name:"policy-name"})--(p:AWSPrincipal) return *  limit 30;` and see N policies and N principals:
image

0.69.0

*Additional context*:

This is because all calls to load managed policies all get [funneled](https://github.com/lyft/cartography/blob/585fc0667feda4cf0d42ebe428cc3ab36fe6432e/cartography/intel/aws/iam.py#L661) to [load_policy_data()](https://github.com/lyft/cartography/blob/585fc0667feda4cf0d42ebe428cc3ab36fe6432e/cartography/intel/aws/iam.py#L594-L600), which handles the data like an inline policy: it generates the policy ID based on the principal that it is attached to.

I'm pretty sure this is a bug since managed policies should not have unique IDs based on the principals that they are attached to.

achantavy:
Fixes #1059: Uses arn as id for managed policies instead of embedding the principal arn in as part of the id.

Adds previously missing unit and integration tests to make sure principal-to-policy connections are loaded correctly.

Before:
image

After:
image


cartography/intel/aws/iam.py
```diff
@@ -108,11 +112,11 @@ def get_user_managed_policy_data(boto3_session: boto3.session.Session, user_list
...
     for user in user_list:
         name = user["UserName"]
-        arn = user["Arn"]
+        user_arn = user["Arn"]
         resource_user = resource_client.User(name)
         try:
-            policies[arn] = {
-                p.policy_name: p.default_version.document["Statement"]
+            policies[user_arn] = {
+                p.arn: p.default_version.document["Statement"]
```
achantavy, line 119: There are 2 main changes of this PR. This is one change..

cartography/intel/aws/iam.py
```diff
@@ -145,11 +149,11 @@ def get_role_managed_policy_data(boto3_session: boto3.session.Session, role_list
...
     for role in role_list:
         name = role["RoleName"]
-        arn = role["Arn"]
+        role_arn = role["Arn"]
         resource_role = resource_client.Role(name)
         try:
-            policies[arn] = {
-                p.policy_name: p.default_version.document["Statement"]
+            policies[role_arn] = {
+                p.arn: p.default_version.document["Statement"]
```
achantavy, line 156: ..and this is the other.

cartography/intel/aws/iam.py
```diff
@@ -76,10 +80,10 @@ def get_group_managed_policy_data(boto3_session: boto3.session.Session, group_li
     policies = {}
     for group in group_list:
         name = group["GroupName"]
-        arn = group["Arn"]
+        group_arn = group["Arn"]
         resource_group = resource_client.Group(name)
-        policies[arn] = {
-            p.policy_name: p.default_version.document["Statement"]
+        policies[group_arn] = {
+            p.arn: p.default_version.document["Statement"]
```
achantavy, line 86: We are using the arn as the key in the dict returned by `get_group_managed_policy_data()`, instead of the policy name. This lets us dedupe managed policies.

ramonpetgrave64 (dismissed):
Looks good. Can you add tests for a Role's Managed Policies.

ramonpetgrave64 (approved):

PR #1063, opened 2022-12-30
Fix integration tests related to timezone

achantavy:
These integration tests work on our CI but not on my local laptop with PST time. This PR makes them work on both now.


cartography/intel/aws/securityhub.py
```diff
@@ -25,7 +25,7 @@ def get_hub(boto3_session: boto3.session.Session) -> Dict:
 
 def transform_hub(hub_data: Dict) -> None:
     if 'SubscribedAt' in hub_data and hub_data['SubscribedAt']:
-        subbed_at = datetime.datetime.strptime(hub_data['SubscribedAt'], "%Y-%m-%dT%H:%M:%S.%fZ")
```
achantavy, line 28: Under the old method, the test expected `1606993517` but my laptop returned `1607022317`.

cartography/intel/aws/securityhub.py
```diff
@@ -25,7 +25,7 @@ def get_hub(boto3_session: boto3.session.Session) -> Dict:
...
+        subbed_at = parser.parse(hub_data['SubscribedAt'])
```
achantavy, line 28: With this, both my laptop and the CI return the expected value of `1606993517`

tests/integration/cartography/intel/aws/ec2/test_auto_scaling_groups.py
```diff
@@ -20,7 +20,7 @@ def test_load_launch_configurations(neo4j_session, *args):
         (
             "example",
             "arn:aws:autoscaling:us-east-1:000000000000:launchConfiguration:00000000-0000-0000-0000-000000000000:launchConfigurationName/example",  # noqa:E501
-            "1632221734.0",
+            "1632221734",
```
achantavy, line 23: Let me know if getting rid of the decimal point is a problem

ramonpetgrave64 (approved):
Works great.

PR #1065, opened 2022-12-30
Fix #1064: handle `firstseen` as a reserved word

### LINKED ISSUE #1064: EMRCluster node don't have firstseen property
*Title*: EMRCluster node does not have firstseen property anymore

*Description*:
As part of https://github.com/lyft/cartography/pull/1038 change, EMRCluster node no longer has "firstseen" property. The problem is that "firstseen" is defined as explicit property in EMRClusterNodeProperties class 
firstseen: PropertyRef = PropertyRef('firstseen')

during query building phase, this ends up getting translated as:
 SET\n                i.lastupdated = $lastupdated,\ni.arn = item.ClusterArn,\ni.auto_terminate = item.AutoTerminate,\ni.autoscaling_role = item.AutoScalingRole,\ni.custom_ami_id = item.CustomAmiId,\n**i.firstseen = item.firstseen,**

however since the item doesn't have firstseen property, it gets reset to None and the original timestamp (set via ON CREATE SET) is lost.

*To Reproduce*:
1. Run cartography against an account that has EMR setup and then review the EMRCluster node.

achantavy:
Fixes #1064: Prevents `firstseen` from being defined as part of CartographyNodeProperties and CartographyRelProperties objects because the querybuilder automatically sets it. Adds tests for EMR.

I couldn't add a unit test for the `__post_init__` because if I create a CartographyNodeProperties with a `firstseen` field, the `__post__init__` gets run when the dataclass object is defined, which causes the TypeError to throw before any tests are able to run.


ramonpetgrave64:
Should we do something similar for `lastupdated`?

achantavy:
> Should we do something similar for lastupdated?


No, because `lastupdated` is part of the `SET` clause and not the `ON CREATE SET` clause, so it should be in the propertyschema.

ramonpetgrave64 (approved):

PR #1069, opened 2023-01-02
Add test helpers check_nodes() and check_rels()

achantavy:
This helps make writing repetitive tests a bit less annoying: the module author no longer needs to remember how to manipulate or index into a neo4j.Result object to check if their expected rels and nodes are connected properly.

### Nodes
Before:
```python
nodes = neo4j_session.run(
    """
    MATCH (r:EMRCluster) RETURN r.arn;
    """,
)
actual_nodes = {n['r.arn'] for n in nodes}

```

After:
```python
assert check_nodes(neo4j_session, 'EMRCluster', ['arn']) == expected_nodes
```
-------

### Relationships

Before:
```python
# Fetch relationships
result = neo4j_session.run(
    """
    MATCH (n1:AWSAccount)-[:RESOURCE]->(n2:EMRCluster) RETURN n1.id, n2.arn;
    """,
)
actual = {
    (r['n1.id'], r['n2.arn']) for r in result
}

assert actual == expected
```

After:
```python
assert check_rels(
    neo4j_session,
    'AWSAccount',
    'id',
    'EMRCluster',
    'arn',
    'RESOURCE',
    rel_direction_right=True,
) == expected
```

### DIFF
--- a/tests/integration/util.py
+++ b/tests/integration/util.py
@@ -15,7 +15,8 @@ def check_nodes(neo4j_session: neo4j.Session, node_label: str, attrs: List[str])
     tuples.
     """
     if not attrs:
-        return None
+        raise ValueError("`attrs` passed to check_nodes() must have at least one element.")
+
     attrs = ", ".join(f"n.{attr}" for attr in attrs)
     query_template = Template("MATCH (n:$NodeLabel) RETURN $Attrs")
     result = neo4j_session.run(
--- a/tests/integration/util.py
+++ b/tests/integration/util.py
@@ -1,5 +1,8 @@
 from string import Template
-from typing import Set, Optional, List, Any
+from typing import Any
+from typing import List
+from typing import Optional
+from typing import Set
 from typing import Tuple
 
 import neo4j
@@ -16,7 +19,7 @@ def check_nodes(neo4j_session: neo4j.Session, node_label: str, attrs: List[str])
     attrs = ", ".join(f"n.{attr}" for attr in attrs)
     query_template = Template("MATCH (n:$NodeLabel) RETURN $Attrs")
     result = neo4j_session.run(
-        query_template.safe_substitute(NodeLabel=node_label, Attrs=attrs)
+        query_template.safe_substitute(NodeLabel=node_label, Attrs=attrs),
     )
     return {tuple(row.values()) for row in result}
 
--- a/tests/integration/util.py
+++ b/tests/integration/util.py
@@ -1,0 +1,57 @@
+from string import Template
+from typing import Set, Optional, List, Any
+from typing import Tuple
+
+import neo4j
+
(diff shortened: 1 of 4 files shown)


tests/integration/util.py
```diff
@@ -0,0 +1,60 @@
...
+    Helper function to test a given relationship between node 1 and node 2.
+    Returns the result of a neo4j match query as a set of tuples when given the node labels of 2 nodes, an attribute to
+    return from each of them, the label of the relationship between them, and the direction of that relationship.
+    :param neo4j_session: The neo4j session
+    :param node_1_label: The label of the first node to check
+    :param node_1_attr: The attribute of the first node to check
+    :param node_2_label: The label of the second node to check
+    :param node_2_attr: The attribute of the second node to check
+    :param rel_label: The str label of the relationship between node 1 and node 2.
+    :param rel_direction_left: The direction of the node is to the left (default=True). Else it is to the right.
```
kunaals, line 46: I feel like the more natural direction to think of relations is going to the right, like in the example you gaveL: `(n1:AWSAccount)-[:RESOURCE]->(n2:EMRCluster)`. Thoughts on making this rel_direction_right?

ramonpetgrave64: Or reuse the `LinkDirection.INWARD`?

tests/integration/util.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def check_nodes(neo4j_session: neo4j.Session, node_label: str, attrs: List[str]) -> Optional[Set[Tuple[Any, ...]]]:
+    """
+    Helper function for checking nodes in cartography integration tests.
+    Returns the result of a neo4j match query on the given node label and the given list of attributes as a set of
+    tuples.
+    """
+    if not attrs:
+        return None
```
ramonpetgrave64, line 18: Maybe through an exception here, like `ValueError`.

ramonpetgrave64 (approved):

tests/integration/util.py
achantavy: INWARD and OUTWARD don't really work here because those are relatively to a specific NodeSchema. Here there is no "anchor point" so we have to use left or right, if that makes sense.

tests/unit/cartography/test_util.py
```diff
@@ -92,3 +94,9 @@ def test_batch(mocker):
     assert actual == expected
     # Also check for empty input
     assert batch([], 3) == []
+
+
+def test_check_nodes_empty_list_raises_exc():
```
achantavy, line 99: This has to be in tests/unit because if it is in tests/integration, pytest.raises() never gets caught for some reason.

kunaals (approved):

ramonpetgrave64 (approved):

PR #1071, opened 2023-01-05
feat: add oauth support to gsuite

resilience-jychp:
Allow gsuite to use oauth credentials.

Legacy method is used by default for backward compatibility.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -417,7 +417,7 @@ class CLI:
             type=str,
             default='GSUITE_GOOGLE_APPLICATION_CREDENTIALS',
             help=(
-                'The method used by GSuite to authenticate. delegated is the legacy one.'
+                'The name of environment variable containing secrets for GSuite authentication.'
             ),
         )
         return parser
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -403,6 +403,23 @@ class CLI:
                 'The crowdstrike URL, if using self-hosted. Defaults to the public crowdstrike API URL otherwise.'
             ),
         )
+        parser.add_argument(
+            '--gsuite-auth-method',
+            type=str,
+            default='delegated',
+            choices=['delegated', 'oauth'],
+            help=(
+                'The method used by GSuite to authenticate. delegated is the legacy one.'
+            ),
+        )
+        parser.add_argument(
+            '--gsuite-tokens-env-var',
+            type=str,
+            default='GSUITE_GOOGLE_APPLICATION_CREDENTIALS',
+            help=(
+                'The method used by GSuite to authenticate. delegated is the legacy one.'
+            ),
+        )
         return parser
 
     def main(self, argv: str) -> int:
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -549,6 +549,13 @@ class CLI:
         else:
             config.crowdstrike_client_secret = None
 
+        # GSuite config
+        if config.gsuite_tokens_env_var:
+            logger.debug(f"Reading config string for GSuite from environment variable {config.gsuite_tokens_env_var}")
+            config.gsuite_config = os.environ.get(config.gsuite_tokens_env_var)
+        else:
+            config.github_config = None
+
         # Run cartography
         try:
             return cartography.sync.run_with_config(self.sync, config)
--- a/cartography/intel/gsuite/__init__.py
+++ b/cartography/intel/gsuite/__init__.py
@@ -1,11 +1,11 @@
+import base64
+import json
 import logging
 import os
 from collections import namedtuple
-import json
-import base64
-import httplib2
 
 import googleapiclient.discovery
+import httplib2
 import neo4j
 from googleapiclient.discovery import Resource
 from oauth2client.client import ApplicationDefaultCredentialsError
@@ -61,10 +61,9 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
         "UPDATE_TAG": config.update_tag,
     }
 
-    # Legacy delegated method
-    if config.gsuite_auth_method == 'delegated':
+    if config.gsuite_auth_method == 'delegated':  #  Legacy delegated method
         try:
-            credentials = GoogleCredentials.from_stream(os.environ.get(config.gsuite_tokens_env_var))
+            credentials = GoogleCredentials.from_stream(config.gsuite_config)
             credentials = credentials.create_scoped(OAUTH_SCOPE)
             credentials = credentials.create_delegated(os.environ.get('GSUITE_DELEGATED_ADMIN'))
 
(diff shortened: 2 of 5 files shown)


resilience-jychp:
@achantavy linter issues should be fixed now

docs/root/modules/gsuite/config.md
```diff
@@ -15,3 +15,77 @@ Follow these steps to analyze GSuite users and groups with Cartography.
...
+
+1. Using Oauth
+
+    1. Create an App on [Google Cloud Console](https://console.cloud.google.com/)
+    1. Refer to follow documentation if needed:
+        1. https://developers.google.com/admin-sdk/directory/v1/quickstart/python
+        1. https://developers.google.com/workspace/guides/get-started
+        1. https://support.google.com/a/answer/7281227?hl=fr
+    1. Download credentials file
+    1. Use helper script for OAuth flow to obtain refresh_token
```
achantavy, line 27:
```suggestion
    1. Use helper script below for OAuth flow to obtain refresh_token
```

docs/root/modules/gsuite/config.md
```diff
@@ -15,3 +15,77 @@ Follow these steps to analyze GSuite users and groups with Cartography.
...
+    1. Serialize needed secret
+        ```
```
achantavy, line 29:
```suggestion
        ```python
```

docs/root/modules/gsuite/config.md
```diff
@@ -15,3 +15,77 @@ Follow these steps to analyze GSuite users and groups with Cartography.
...
+1. Using Oauth
+
+    1. Create an App on [Google Cloud Console](https://console.cloud.google.com/)
+    1. Refer to follow documentation if needed:
+        1. https://developers.google.com/admin-sdk/directory/v1/quickstart/python
+        1. https://developers.google.com/workspace/guides/get-started
+        1. https://support.google.com/a/answer/7281227?hl=fr
+    1. Download credentials file
+    1. Use helper script for OAuth flow to obtain refresh_token
+    1. Serialize needed secret
```
achantavy, line 28: Let's either add a step here telling the user to delete the credentials.json file that was saved to disk, or add a step in the script below that deletes the file.

docs/root/modules/gsuite/config.md
```diff
@@ -15,3 +15,77 @@ Follow these steps to analyze GSuite users and groups with Cartography.
...
+
+from google_auth_oauthlib.flow import InstalledAppFlow
+from googleapiclient.discovery import build
+
+
+scopes = ["https://www.googleapis.com/auth/admin.directory.userreadonly", "https://www.googleapis.com/auth/admin.directory.group.readonly", "https://www.googleapis.com/auth/admin.directory.group.member"]
+
+print('Go to https://console.cloud.google.com/ > API & Services > Credentials and download secrets')
+project_id = input('Provide your project ID:')
+client_id = input('Provide you Client ID:')
```
achantavy, line 54:
```suggestion
client_id = input('Provide your client ID:')
```

docs/root/modules/gsuite/config.md
```diff
@@ -15,3 +15,77 @@ Follow these steps to analyze GSuite users and groups with Cartography.
...
+client_secret = input('Provide you Client Secret:')
```
achantavy, line 55:
```suggestion
client_secret = input('Provide your client secret:')
```

cartography/intel/gsuite/__init__.py
```diff
@@ -62,23 +61,50 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
...
-            (
-                "Unable to initialize GSuite creds. If you don't have GSuite data or don't want to load "
-                'Gsuite data then you can ignore this message. Otherwise, the error code is: %s '
-                'Make sure your GSuite credentials are configured correctly, your credentials file (if any) is valid. '
-                'For more details see README'
-            ),
-            e,
-        )
-        return
+    if config.gsuite_auth_method == 'delegated':  #  Legacy delegated method
```
achantavy, line 64:
```suggestion
    if config.gsuite_auth_method == 'delegated':  # Legacy delegated method
        logger.info('Attempting to authenticate to GSuite using legacy delegated method')
```

cartography/intel/gsuite/__init__.py
```diff
@@ -62,23 +61,50 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
...
+                    'Gsuite data then you can ignore this message. Otherwise, the error code is: %s '
+                    'Make sure your GSuite credentials file (if any) is valid. '
+                    'For more details see README'
+                ),
+                e,
+            )
+            return
+    elif config.gsuite_auth_method == 'oauth':
+        auth_tokens = json.loads(str(base64.b64decode(config.gsuite_config).decode()))
+        logger.info('SO GOOD SO FAR')
```
achantavy, line 84:
```suggestion
        logger.info('Attempting to authenticate to GSuite using OAuth')
```

cartography/intel/gsuite/__init__.py
```diff
@@ -62,23 +61,50 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
...
+            credentials = GoogleCredentials(
+                None,
+                auth_tokens['client_id'],
+                auth_tokens['client_secret'],
+                auth_tokens['refresh_token'],
+                None,
+                auth_tokens['token_uri'],
+                'Cartography',
+            )
+            credentials.refresh(httplib2.Http())
```
achantavy, line 95: Just wondering, what does this do?

cartography/intel/gsuite/__init__.py
```diff
@@ -62,23 +61,50 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
...
-        )
-        return
+    if config.gsuite_auth_method == 'delegated':  #  Legacy delegated method
+        try:
+            credentials = GoogleCredentials.from_stream(config.gsuite_config)
+            credentials = credentials.create_scoped(OAUTH_SCOPE)
+            credentials = credentials.create_delegated(os.environ.get('GSUITE_DELEGATED_ADMIN'))
+
+        except ApplicationDefaultCredentialsError as e:
+            logger.debug('Error occurred calling GoogleCredentials.get_application_default().', exc_info=True)
```
achantavy, line 71:
```suggestion
```

I realize this code was left over from before; we can omit it now I think.

cartography/intel/gsuite/__init__.py
```diff
@@ -62,23 +61,50 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
...
+                auth_tokens['client_secret'],
+                auth_tokens['refresh_token'],
+                None,
+                auth_tokens['token_uri'],
+                'Cartography',
+            )
+            credentials.refresh(httplib2.Http())
+            credentials = credentials.create_scoped(OAUTH_SCOPE)
+        except ApplicationDefaultCredentialsError as e:
+            logger.debug('Error occurred calling GoogleCredentials.get_application_default().', exc_info=True)
```
achantavy, line 98:
```suggestion
```

Same here.

cartography/intel/gsuite/__init__.py
```diff
@@ -62,23 +61,50 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
...
+            logger.error(
+                (
+                    "Unable to initialize GSuite creds. If you don't have GSuite data or don't want to load "
+                    'Gsuite data then you can ignore this message. Otherwise, the error code is: %s '
+                    'Make sure your GSuite credentials are configured correctly, your credentials are valid. '
+                    'For more details see README'
+                ),
+                e,
+            )
```
achantavy, line 107:
```suggestion
            )
            return
```

I think we want to return out of this function if this exception is hit, which will then continue on to run the other syncs.

achantavy (changes requested):
Thanks for the PR! Left some nits and questions but overall looks good to me.

cartography/intel/gsuite/__init__.py
resilience-jychp: We just store refresh token this step will request an access token from it.

achantavy (dismissed):
Thank you for the PR!

achantavy:
Just gotta remove a whitespace and remove the `on match set` and I think the CI will pass

achantavy (approved):
Thank you!

cartography/cli.py
```diff
@@ -403,6 +403,23 @@ def _build_parser(self):
...
+            help=(
+                'The method used by GSuite to authenticate. delegated is the legacy one.'
+            ),
+        )
+        parser.add_argument(
+            '--gsuite-tokens-env-var',
+            type=str,
+            default='GSUITE_GOOGLE_APPLICATION_CREDENTIALS',
+            help=(
+                'The method used by GSuite to authenticate. delegated is the legacy one.'
```
achantavy, line 420: Crap, I missed one thing in my review actually - what should the docstring here be? This is the same text message for `--gsuite-auth-method`.

Sorry for the back and forth.

achantavy (changes requested):

achantavy (approved):

PR #1075, opened 2023-01-07
Add Azure Monitor source

juju4:
Add Ingestion of Azure Monitor aka Azure Log Analytics aka Sentinel as source.


jychp:
Done in https://github.com/cartography-cncf/cartography/pull/1994

PR #1078, opened 2023-01-07
Add sumologic source

juju4:
Sumologic log platform
https://www.sumologic.com/


cartography/data/jobs/cleanup/sumologic_import_cleanup.json
```diff
@@ -0,0 +1,10 @@
+{
+  "statements": [
+    {
+      "query": "WITH datetime()-duration('P7D') AS threshold MATCH (h:SumologicHost) WHERE h.lastupdated < threshold WITH h LIMIT $LIMIT_SIZE DETACH DELETE (h)",
```
ramonpetgrave64, line 4: Can you tell us how this works?

docs/root/modules/sumologic/schema.md
```diff
@@ -0,0 +1,18 @@
...
+.. _sumologic_schema:
+
+### Sumologic core platform
+
+Representation of a system sending logs to Sumologic core platform. Be aware that this may vary depending on your environment.
+Default code expects _sourceCategory to be formatted as BU/DC/SOURCE_TYPE aka business unit, datacenter or location, and source type. Source type as SERVER or NXLOG_* to match targeted systems (linux, windows...). Some fields may need to be set through [Field Extraction Rules](https://help.sumologic.com/docs/manage/field-extractions/)
+
+| Field | Description |
+|-------|-------------|
+|tool_first_seen| Timestamp of when first available logs for host is available since first sync|
```
ramonpetgrave64, line 12: I think you meant to add the schema for the hosts?

cartography/data/jobs/cleanup/sumologic_import_cleanup.json
juju4: This is the same thing as before but cleaning only entries lastupdated older than 7 days as per https://github.com/lyft/cartography/issues/1015

docs/root/modules/sumologic/schema.md
juju4: I don't understand the comment.
On tool_first_seen, this is limited by the duration given to sumologic query. if query only last 7d log, the earliest first seen is only 7d old, even if logging platform can have more.
But querying on long time period is expensive. And if the platform aka cartography server lives long enough, it will get there.

juju4:
outside of the points left above for which I would like inputs too.

PR #1079, opened 2023-01-12
Add Hexnode MDM source

jychp:
This source will pull users/devices/policies from Hexnode MDM.

It adds some Nodes:
HexnodeUser
HexnodeDevice
HexnodeDeviceGroup
HexnodePolicy

It uses Human as pivot with other modules (GSuite, GitHub, HiBob ...)


resilience-jychp:
Transition to new datamodel is blocked by : https://github.com/lyft/cartography/issues/1131

PR #1080, opened 2023-01-12
Populate AWS Tags for IAM roles

meng-han:
According to AWS support, `resourcegroupstaggingapi` does not work for IAM role yet and currently there's no ETA for it. This is a temporary workaround so that AWS tags can be populated for AWS roles via IAM client/resource instead of resourcegroupstaggingapi

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -164,6 +164,7 @@ def get_role_managed_policy_data(boto3_session: boto3.session.Session, role_list
             )
     return policies
 
+
 @timeit
 def get_role_tag_data(boto3_session: boto3.session.Session, role_list: List[Dict]) -> List[Dict]:
     resource_client = boto3_session.resource('iam')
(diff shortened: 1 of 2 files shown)


cartography/intel/aws/iam.py
```diff
@@ -162,6 +164,30 @@ def get_role_managed_policy_data(boto3_session: boto3.session.Session, role_list
...
+    """
+    resource_client = boto3_session.resource('iam')
+    all_tags: List[Dict] = []
+    for role in role_list:
+        name = role["RoleName"]
+        role_arn = role["Arn"]
+        resource_role = resource_client.Role(name)
+        role_tags = {
+            'ResourceARN': role_arn,
+            'Tags': resource_role.tags
```
meng-han, line 186: if tags are empty, should probably not add it to the list at all

cartography/intel/aws/iam.py
```diff
@@ -721,6 +749,24 @@ def sync_role_inline_policies(
...
+    resource_type = 'iam:role'
+    logger.info("Syncing IAM role tags for account '%s'.", current_aws_account_id)
+    tag_data = get_role_tag_data(boto3_session, data["Roles"])
+    transform_tags(tag_data, resource_type)
+    logger.info(f"Loading {len(tag_data)} tags for rsource type {resource_type}")
+    load_tags(
+        neo4j_session=neo4j_session,
+        tag_data=tag_data,
+        resource_type=resource_type,
+        region='', #TODO: fix region, iam is global, this can be optional?
```
meng-han, line 765: region

meng-han:
in favor of https://github.com/lyft/cartography/pull/1081

PR #1081, opened 2023-01-12
Populate AWS Tags for IAM roles

meng-han:
Alternative approach: https://github.com/lyft/cartography/pull/1080

According to AWS support, resourcegroupstaggingapi does not work for IAM role yet and currently there's no ETA for it. This is a temporary workaround so that AWS tags can be populated for AWS roles via IAM client/resource instead of resourcegroupstaggingapi


achantavy (approved):
Looks good, thank you!

PR #1083, opened 2023-01-13
Add Lastpass source

resilience-jychp:
This source will pull users from Lastpass

Only one new node:
- Lastpass

It uses Human as pivot with other modules (GSuite, GitHub, HiBob ...)


cartography/data/jobs/analysis/lastpass_human_link.json
```diff
@@ -0,0 +1,13 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (human:Human), (user:LastpassUser) WHERE human.email = user.email MERGE (human)-[r:IDENTITY_LASTPASS]->(user) ON CREATE SET r.firstseen = $UPDATE_TAG SET r.lastupdated = $UPDATE_TAG",
+      "iterative": false
+    },
+    {
+      "query": "MATCH (:Human)-[r:IDENTITY_LASTPASS]->(:LastpassUser) WHERE r.lastupdated <> $UPDATE_TAG WITH r LIMIT $LIMIT_SIZE DELETE (r) return COUNT(*) as TotalCompleted",
```
ramonpetgrave64, line 8: use `DETACH DELETE`, but also put this statement into the cleanup job.

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,85 @@
...
+    return result
+
+def load(
+    neo4j_session: neo4j.Session,
+    data: List[Dict],
+    update_tag: int,
+) -> None:
+
+    query = """
+    UNWIND $UserData as user
```
ramonpetgrave64, line 57: non blocker: try to use our new method for automatically generating these queries.
https://lyft.github.io/cartography/dev/writing-intel-modules.html#load

cartography/data/jobs/analysis/lastpass_human_link.json
ramonpetgrave64: I think you're right.

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,72 @@
...
+@timeit
+def transform(api_result: dict) -> List[Dict]:
+    result: List[dict] = []
+    for uid, user in api_result['Users'].items():
+        n_user = user.copy()
+        n_user['id'] = int(uid)
+        n_user['created'] = int(dt_parse.parse(user['created']).timestamp() * 1000)
+        n_user['last_pw_change'] = int(dt_parse.parse(user['last_pw_change']).timestamp() * 1000)
+        try:
+            n_user['last_login'] = int(dt_parse.parse(user['last_login']).timestamp() * 1000)
```
ramonpetgrave64, line 52: Can the value be an empty string, or is it no-empty and unparseable?

cartography/cli.py
```diff
@@ -420,6 +420,22 @@ def _build_parser(self):
...
+        parser.add_argument(
+            '--lastpass-cid-env-var',
+            type=str,
+            default=None,
+            help=(
+                'The name of environment variable containing the Lastpass CID for authentication.'
+            ),
+        )
+        parser.add_argument(
+            '--lastpass-provhash-env-var',
```
ramonpetgrave64, line 432: Doe the lastpass docs have some defaults that make sense to be put here?

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@timeit
+def transform(api_result: dict) -> List[Dict]:
+    result: List[dict] = []
+    for uid, user in api_result['Users'].items():
+        n_user = user.copy()
+        n_user['id'] = int(uid)
+        for k in ('created', 'last_pw_change', 'last_login'):
+            if n_user[k] == '':
```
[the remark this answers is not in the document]
ramonpetgrave64: But if the string is empty in `n_user[k]`, would it also be empty in `user[k]`? Does the `parse(user[k]`) raise an exception?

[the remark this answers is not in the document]
resilience-jychp: Sure it raises an Exception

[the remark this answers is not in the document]
ramonpetgrave64: It seems like lines 50 and 51 are actually unnecessary because the exception to be raised is unhandled.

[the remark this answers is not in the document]
resilience-jychp: On the contrary is because the error is hunandled that we have to do this check

A user who never logged in will have an empty last_login, resulting on a hunandled exception that will crash the tool.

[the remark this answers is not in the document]
achantavy: From your testing, the Lastpass API returns empty strings?

If this is the case, I think it is ok to do 

```python
n_user[k] = int(dt_parse.parse(n_user[k]).timestamp() * 1000) if n_user[k] else None
```
because the Neo4j driver will automatically leave that field as unset on the node, and because the empty string is Falsy in Python.

So overall, I think we can we write this function as 

```python
@timeit
def transform(api_result: Dict[str, Any]) -> List[Dict[str, Any]]:
    result: List[Dict[str, Any]] = []
    for uid, user in api_result['Users'].items():
        n_user = user.copy()
        n_user['id'] = int(uid)
        for k in ('created', 'last_pw_change', 'last_login'):
            n_user[k] = int(dt_parse.parse(n_user[k]).timestamp() * 1000) if n_user[k] else None
        result.append(n_user)
    return result
```

docs/root/modules/lastpass/config.md
```diff
@@ -0,0 +1,9 @@
+## Lastpass Configuration
+
+.. _lastpass_config:
+
+Follow these steps to analyze Lastpass objects with Cartography.
+
+1. Prepare your Lastpass CID & ProvHash key.
+    1. Get your CID (account number) and ProvHash from Lastpass [Where can I find the CID and API secret?](https://support.lastpass.com/help/where-can-i-find-the-cid-and-api-secret)
```
achantavy, line 8: Curious, what privileges does this provhash grant? Is the authz configurable?

achantavy (changes requested):

cartography/intel/lastpass/users.py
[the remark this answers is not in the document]
resilience-jychp: @ramonpetgrave64 my bad I just checked my code again and there is a missing 'else' (or continu in the if condition) ... I known understand your point ;)

@achantavy your solution is more elegant thant mine

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,70 @@
...
+@timeit
+def get(lastpass_cid: str, lastpass_provhash: str) -> dict:
+    payload = {
+        'cid': lastpass_cid,
+        'provhash': lastpass_provhash,
+        'cmd': 'getuserdata',
+        'data': None,
+    }
+    session = Session()
+    req = session.post('https://lastpass.com/enterpriseapi.php', data=payload, timeout=20)
```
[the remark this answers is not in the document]
resilience-jychp: Done
An improvment will be to make this configurable (see https://github.com/lyft/cartography/issues/1072)

docs/root/modules/lastpass/config.md
resilience-jychp: Sadly it's Lastpass ... so no configuration/restriction (do not judge me, Lastpass is a Legacy app, working on new solution ;) )

The API is mainly designed to allow provisioning (add/edit/delet user & groups)
You can add/access enterprise level shared access
Thanksfully you can not access user's secrets

cartography/data/jobs/analysis/lastpass_human_link.json
```diff
@@ -0,0 +1,13 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (human:Human), (user:LastpassUser) WHERE human.email = user.email MERGE (human)-[r:IDENTITY_LASTPASS]->(user) ON CREATE SET r.firstseen = $UPDATE_TAG SET r.lastupdated = $UPDATE_TAG",
```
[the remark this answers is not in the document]
resilience-jychp: I juste made a copy of GSuite/GitHub analysis job, but you are right, using CartographySchema is a better way.

But I have some issues, please refer to https://github.com/lyft/cartography/issues/1106

Think the best way is an helper function (like you did for DNS).

cartography/intel/lastpass/schema.py
```diff
@@ -0,0 +1,34 @@
...
+    applications: PropertyRef = PropertyRef('applications')
+    attachments: PropertyRef = PropertyRef('attachments')
+    password_reset_required: PropertyRef = PropertyRef('password_reset_required')
+    multifactor: PropertyRef = PropertyRef('multifactor')
+
+
+@dataclass(frozen=True)
+class LastpassUserSchema(CartographyNodeSchema):
+    label: str = 'LastpassUser'
+    properties: LastpassUserNodeProperties = LastpassUserNodeProperties()
```
achantavy, line 34: @resilience-jychp - hey, one more thought - what would make sense to use for a Lastpass sub resource? If there is a concept of an "organization" or "tenant" or something similar then we would link the Lastpass user to that so that cleanup jobs become more safe.

resilience-jychp:
Sure I can do that.
Tenant Node will be pretty empty but that make sense.

Le mar. 14 févr. 2023, 07:11, Alex Chantavy ***@***.***> a
écrit :

cartography/data/jobs/analysis/lastpass_human_link.json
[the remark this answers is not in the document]
resilience-jychp: I replaced this with a CartographyRelSchema

resilience-jychp:
@achantavy 

I migrate to new model architecture, but not sure how auto_cleanup work, so I kept the old way.

resilience-jychp:
@achantavy this PR should be ready for new model, waiting for you comments before working on other ones

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,80 @@
...
+    for uid, user in api_result['Users'].items():
+        n_user = user.copy()
+        n_user['id'] = int(uid)
+        for k in ('created', 'last_pw_change', 'last_login'):
+            n_user[k] = int(dt_parse.parse(user[k]).timestamp() * 1000) if user[k] else None
+        result.append(n_user)
+    return result
+
+
+def load(
```
achantavy, line 58: Can you update this to use

```python
load(neo4j_session, LastpassTenantSchema(), data, tenant_id=lastpass_cid, lastupdated=update_tag)
```

and 

```python
load(neo4j_session, LastpassUserSchema(), data, tenant_id=lastpass_cid, lastupdated=update_tag)
```

instead of generating the query and calling `load_graph_data`?

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,80 @@
...
+logger = logging.getLogger(__name__)
+# Connect and read timeouts of 60 seconds each; see https://requests.readthedocs.io/en/master/user/advanced/#timeouts
+_TIMEOUT = (60, 60)
+
+
+@timeit
+def sync(
+    neo4j_session: neo4j.Session,
+    lastpass_provhash: str,
+    common_job_parameters: Dict[str, Any],
```
achantavy, line 25: I think we should include `update_tag` and `tenant_id` as separate parameters to `sync` in addition to including them as keys in `common_job_parameters`. The AWS modules do similar.

achantavy (changes requested):
Almost there! Left comments on some function signature changes and using `cartography.client.core.tx.load`. It's looking good!

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,80 @@
...
+) -> None:
+    users = get(lastpass_provhash, common_job_parameters)
+    formated_users = transform(users)
+    load(neo4j_session, formated_users, common_job_parameters)
```
achantavy, line 29: Oh, this is also missing a cleanup function. Just need to add one to cleanup the users. No need to clean up the tenant.

```python
def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict[str, Any]) -> None:
    GraphJob.from_node_schema(LastpassUserSchema(), common_job_parameters).run(neo4j_session)
```

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,80 @@
...
+    }
+    session = Session()
+    req = session.post('https://lastpass.com/enterpriseapi.php', data=payload, timeout=_TIMEOUT)
+    req.raise_for_status()
+    return req.json()
+
+
+@timeit
+def transform(api_result: dict) -> List[Dict]:
+    result: List[dict] = []
```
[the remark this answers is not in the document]
resilience-jychp: Hum I disagree with this one, result is a List[Dict[str, Any]]

cartography/intel/lastpass/users.py
resilience-jychp: Hum ... I can make the change but do not realy like that, seems very confusing for me :
https://github.com/lyft/cartography/blob/672b5fe4918fef562d7816d8e9833269c801156f/cartography/intel/aws/emr.py#L96

Here you use update_tag for load function and common_parameters for cleanup.
IMHO if for any reason in the code those two parameters become different you will have a uggly cleanup very hard to debug

tests/integration/cartography/intel/lastpass/test_users.py
```diff
@@ -0,0 +1,38 @@
+import cartography.intel.lastpass.users
+import tests.data.lastpass.users
+
+
+TEST_UPDATE_TAG = 123456789
+
+
+def test_load_lastpass_users(neo4j_session):
```
ramonpetgrave64, line 8: also add a test for humans and tenants

ramonpetgrave64 (changes requested):
Looks great, just a couple test cases to add.

cartography/intel/lastpass/users.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@timeit
+def transform(api_result: Dict[str, Any]) -> List[Dict[str, Any]]:
+    result: List[Dict[str, Any]] = []
+    for uid, user in api_result['Users'].items():
+        n_user = user.copy()
+        n_user['id'] = int(uid)
+        for k in ('created', 'last_pw_change', 'last_login'):
+            n_user[k] = int(dt_parse.parse(user[k]).timestamp() * 1000) if user[k] else None
```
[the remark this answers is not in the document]
resilience-jychp: Yes, empty string are check by "if user[k]" statement
The k is always in the API result, but with None or Empty String value

cartography/intel/lastpass/users.py
achantavy: I agree that this is a problem but all the other syncs currently follow that pattern, and update_tag is never written to (fingers crossed that it stays this way).

In future I would like to do a larger refactor to eliminate the possibility of those two parameters becoming different. This could involve:

(A) make `common_job_parameters` become a formal object where `update_tag` and `tenant_id` can become their own fields 

or

(B) Get rid of `common_job_parameters` altogether and just use parameters from the function itself

or

(C) something else


However for right now I think this module should keep the existing pattern.

tests/integration/cartography/intel/lastpass/test_users.py
achantavy: @resilience-jychp - can you let `sync()` be the function under test and patch out the results of `get()` with `tests.data.lastpass.users.LASTPASS_USERS`? 

This way as the Lastpass module grows you don't need to add more transform + load calls and you can just add more test cases.

Here's an example of doing this: https://github.com/lyft/cartography/blob/ab1584f325c8a9c424c1358f2e482b9c4df2c7ea/tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py#L20-L37

tests/integration/cartography/intel/lastpass/test_users.py
```diff
@@ -0,0 +1,38 @@
...
+
```
achantavy, line 9: As one more nitpicky suggestion, can you comment the test with 

```python
# Arrange
...

# Act
sync(...)

# Assert
```

As mentioned on Slack, we will use this PR to be the main example that we use for new contributors to copy so I am being a bit extra right now :)

resilience-jychp:
@achantavy @ramonpetgrave64 changes made

docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
         timeout: 10s
         retries: 10
   cartography:
-    image: ghcr.io/lyft/cartography
+    image: lyft/cartography
```
achantavy, line 37: What's this do?

resilience-jychp: Good catch, I changed the docker-compose file to run tests with local build.
I revert this change immediately.

ramonpetgrave64 (dismissed):

ramonpetgrave64 (approved):

achantavy (approved):

achantavy:
Thank you @resilience-jychp !

PR #1095, opened 2023-01-21
Fix S3 policy test

phishoes:
test_load_s3_policies sometimes fails on CI because the neo4j query returns the results in different orders. This PR fixes the test by checking for list membership instead of list equivalence.


tests/integration/cartography/intel/aws/test_s3.py
```diff
@@ -153,7 +153,9 @@ def test_load_s3_policies(neo4j_session, *args):
         )
         for n in nodes
     ]
-    assert actual_nodes == expected_nodes
+    assert len(actual_nodes) == 3
```
achantavy, line 156: While you're in here, it might be a good opportunity to refactor it to use `check_rels()` and `check_nodes()`, but this works too.

achantavy (approved):

PR #1096, opened 2023-01-23
Organize data model files under cartography.models

achantavy:
As discussed on https://lyftoss.slack.com/archives/CTZUQL0KX/p1674037104828659: it'd be easier to discover schemas and other data types if they are located in one place.

- Node-related objects in cartography/graph/model are moved to cartography/models/nodes.py
- Relationship-related objects in cartography/graph/model.py are moved to cartography/models/relationships.py
- Objects and functions applicable to both are moved to cartography/models/common.py
- Model objects for cartography/intel/aws/emr.py are moved to cartography/models/aws/emr.py (thankfully there is only one real file to refactor here since we're catching this early enough).

### DIFF
--- a/cartography/data/jobs/cleanup/aws_import_tags_cleanup.json
+++ b/cartography/data/jobs/cleanup/aws_import_tags_cleanup.json
@@ -121,6 +121,16 @@
             "iterationsize": 100
         },
         {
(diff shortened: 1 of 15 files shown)


achantavy:
cc: @resilience-jychp

ramonpetgrave64 (approved):

PR #1107, opened 2023-02-02
Automatic indexes

achantavy:
Use cartography data model to enable automatic index creation. This is important because missing indexes can lead to extremely poor performance.

**Before**: manually update indexes.cypher
```cypher
CREATE INDEX IF NOT EXISTS FOR (n:EMRCluster) ON (n.id);
CREATE INDEX IF NOT EXISTS FOR (n:EMRCluster) ON (n.arn);
CREATE INDEX IF NOT EXISTS FOR (n:EMRCluster) ON (n.lastupdated);
```

---

**After**: cartography will automatically create indexes for all fields necessary to make joins efficient on other nodes.

```python
def load_emr_clusters(...) -> None:
...
    load( ... )
```

Optionally update the node schema object with `extra_index=True` for fields that you explicitly want indexes for.
```python
@dataclass(frozen=True)
class EMRClusterNodeProperties(CartographyNodeProperties):
    arn: PropertyRef = PropertyRef('ClusterArn', extra_index=True)
```


cartography/intel/aws/emr.py
```diff
@@ -60,9 +61,8 @@ def load_emr_clusters(
         aws_update_tag: int,
 ) -> None:
     logger.info("Loading EMR %d clusters for region '%s' into graph.", len(cluster_data), region)
-
+    ensure_indexes(neo4j_session, EMRClusterSchema())
```
ramonpetgrave64, line 64: Can you try to sneak this into the `NodeSchema`'s `__init__`, so that we don't have to call this? If not, maybe call this within `build_ingestion_query` or `load_graph_data`?

cartography/graph/querybuilder.py
```diff
@@ -374,3 +375,56 @@ def build_ingestion_query(
...
+    rel_schemas = []
+    if node_schema.sub_resource_relationship:
+        rel_schemas.extend([node_schema.sub_resource_relationship])
+    if node_schema.other_relationships:
+        rel_schemas.extend(node_schema.other_relationships.rels)
+    result.extend([
+        index_template.safe_substitute(
+            TargetNodeLabel=rs.target_node_label,
+            TargetAttribute=target_key,
+        ) for rs in rel_schemas for target_key in asdict(rs.target_node_matcher).keys()
```
ramonpetgrave64, line 419: non-blcoker: This one is a little tricky to read, so I recommend we make it into the usual nested loop.

ramonpetgrave64 (changes requested):
Looks good, but if it doesn't make sense to automatically call ` ensure_indexes ` somehwere else, I'll approve.

docs/root/dev/writing-intel-modules.md
```diff
@@ -83,7 +82,7 @@ def load_emr_clusters(
 ```
 
 
-`load_graph_data()` requires an `ingestion_query` to be generated from `CartographyNodeSchema` and `CartographyRelSchema` objects. [cartography.graph.querybuilder.build_ingestion_query()](https://github.com/lyft/cartography/blob/e6ada9a1a741b83a34c1c3207515a1863debeeb9/cartography/graph/querybuilder.py#L312) does just that: it accepts those schema objects as input and returns a well-formed and optimized cypher query.
```
ramonpetgrave64, line 86: Modules won't be directly calling `load_graph_data` anymore?

cartography/client/core/tx.py
```diff
@@ -210,3 +213,49 @@ def load_graph_data(
...
+
+    for query in queries:
+        if not query.startswith('CREATE INDEX IF NOT EXISTS'):
+            raise ValueError('Query provided to `ensure_indexes()` does not start with "CREATE INDEX IF NOT EXISTS".')
+        neo4j_session.run(query)
+
+
+def load(
+        neo4j_session: neo4j.Session,
+        node_schema: CartographyNodeSchema,
```
ramonpetgrave64, line 239: If this function is going to be the only entrypoint for 'load_graph_data', then I hesitate this to make this a required argument, because not everything will have a proper NodeSchema, but we still like the managed transaction aspect of `load_graph_data` for other modules to begin using, if we can't easily onboard them to NodeSchemas.

Maybe rename it to `load_graph_data_from_node_schema`?

docs/root/dev/writing-intel-modules.md
achantavy: Correct, I updated the doc. They will be calling `load()`. It's early enough to make a change like this before more modules start using it

cartography/client/core/tx.py
achantavy: They can still call `load_graph_data()` directly, that's a public entrypoint.

achantavy: I think calling it `load()` makes it like we're saying "this is the way that we intend for all of this to eventually look"

ramonpetgrave64 (approved):

cartography/intel/aws/emr.py
```diff
@@ -45,11 +43,9 @@ def get_emr_describe_cluster(boto3_session: boto3.session.Session, region: str,
...
         cluster_details = response['Cluster']
     except botocore.exceptions.ClientError as e:
-        logger.warning(
-            "Could not run EMR describe_cluster due to boto3 error %s: %s. Skipping.",
-            e.response['Error']['Code'],
-            e.response['Error']['Message'],
-        )
+        code = e.response['Error']['Code']
+        msg = e.response['Error']['Message']
+        logger.warning(f"Could not run EMR describe_cluster due to boto3 error {code}: {msg}. Skipping.")
```
achantavy, line 48: Figured I'd sneak in a change here to make the string formatting more readable since we expect new module authors to be copying this

cartography/client/core/tx.py
```diff
@@ -238,24 +238,17 @@ def load(
...
-    You would want to specify your own if you are calling `build_ingestion_query()` yourself and using the
-    `selected_relationships` argument.
     :param dict_list: The data to load to the graph represented as a list of dicts.
     :param kwargs: Allows additional keyword args to be supplied to the Neo4j query.
     :return: None
     """
     ensure_indexes(neo4j_session, node_schema)
-    if not ingestion_query:
-        ingestion_query = build_ingestion_query(node_schema)
+    ingestion_query = build_ingestion_query(node_schema)
```
achantavy, line 253: @ramonpetgrave64 - what you think? I figure we can add back the ability to specify an `ingestion_query` later on if we find problems

docs/root/dev/writing-intel-modules.md
```diff
@@ -68,22 +68,20 @@ def load_emr_clusters(
...
     )
 
 ```
 
 
-`load_graph_data()` requires an `ingestion_query` to be generated from `CartographyNodeSchema` and `CartographyRelSchema` objects. [cartography.graph.querybuilder.build_ingestion_query()](https://github.com/lyft/cartography/blob/e6ada9a1a741b83a34c1c3207515a1863debeeb9/cartography/graph/querybuilder.py#L312) does just that: it accepts those schema objects as input and returns a well-formed and optimized cypher query.
+The `load()` function's `ingestion_query` argument is a string generated from `CartographyNodeSchema` and `CartographyRelSchema` objects. [cartography.graph.querybuilder.build_ingestion_query()](https://github.com/lyft/cartography/blob/e6ada9a1a741b83a34c1c3207515a1863debeeb9/cartography/graph/querybuilder.py#L312) does just that: it accepts those schema objects as input and returns a well-formed and optimized cypher query.
 
 
 #### Defining a node
```
ramonpetgrave64, line 87: Add a note about using the `extra_index` argument

ramonpetgrave64 (changes requested):
Almost there!

ramonpetgrave64 (approved):

PR #1111, opened 2023-02-07
 Fix #1102: Correctly transform policy statements in-place

### LINKED ISSUE #1102: IAM statement not being transformed properly
*Title*: IAM Statement is not being transformed properly

*Description*:
In  iam.py: _transform_policy_statements method if the original statement is not a list, it is transposed as a list:
    if not isinstance(statements, list):
        statements = [statements]

However the original dictionary value is not updated, in transform_policy_data:
statements = _transform_policy_statements(
                statements, policy_id,
            )

it should be:
policy_statement_map[policy_key] =  _transform_policy_statements(
                statements, policy_id,
            )

*To Reproduce*:
> inline policy statement value happens to be only this (note this is not a list but just a dict object) prior to transformation
{'Effect': 'Allow', 'Action': 'secretsmanager:GetSecretValue', 'Resource': 'arn:aws:secretsmanager:XXXXX:XXXXXXXX:secret:*'}

*Logs*:
 File "/var/cartography/cartography/intel/aws/iam.py", line 854, in load_policy_statements
statement_data_id=statement_data['id'],
 TypeError: string indices must be integers

achantavy:
Fixes #1102 - previously `transform_policy_data()` assigned the output of `_transform_policy_statements()` to a temporary variable when it intended to update the given dict in-place. This PR fixes the bug and adds a test.

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -540,11 +540,12 @@ def transform_policy_data(policy_map: Dict, policy_type: str) -> None:
     for principal_arn, policy_statement_map in policy_map.items():
         logger.debug(f"Transforming IAM {policy_type} policies for principal {principal_arn}")
         for policy_key, statements in policy_statement_map.items():
-            policy_id = transform_policy_id(principal_arn, policy_type, policy_key) \
-                if policy_type == PolicyType.inline.value else policy_key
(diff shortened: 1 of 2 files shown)


ryan-lane (approved):
LGTM

Issue #1106, opened 2023-02-02
CartographySchema with Node shared by multiple intel


resilience-jychp:
*Example:*
- GoogleSuite use Human to link GoogleAccount to a "Person" with analysis job `(h)-[r:IDENTITY_GSUITE]->(user)`
- GitHub use Human to link GitHubAccount to a "Person" with analysis job `(h)-[r:IDENTITY_GITHUB]->(user)`
- I wanted to do the same think with Lastpass (https://github.com/lyft/cartography/pull/1083)

I'm stuck with the CartographySchema implementation :

## Problem 1 - multiple definition
Even with the schema reorganization (https://github.com/lyft/cartography/pull/1096) the Human node must be redefined several times that could lead to inconsistant fields
Add a cross_intel/common/wellknown list of Node in this schema will lead to a PropertyRef issue (field can change between intel sources)
**Suggestion: add a notion of "constraint/schema" verified by the CI which checks the consistency of the schema (the nodes are defined several times but consistency is guaranteed)**

## Problem 2 - auto clean
In my usage, there is not one execution of the script but several (which can be triggered by Terraform CIs, crons etc ...)
With a CartographySchemaNode, autoclean will erasing the Humans created by another Intel
_ex: GitHub will update a small subset of Human, but maybe sooner GoogleSuite updated more Humans nodes that will be purged by autoclean on the "GitHub execution"_
**Suggestion: add a cli parameter: autoclean_max_age (0 by default) and purge for lastupdate < NOW - autoclean_max_age instead of lastupdate != NOW**

achantavy:
Hey, thank you for the thoroughly written issue.

## Regarding problem 1

I don't think the Human node will need to be defined in multiple places. The [GSuiteUser-to-Human link](https://github.com/lyft/cartography/blob/81902b23fa80e4ba5332ba00b4477e3a556d5eb7/cartography/data/jobs/analysis/gsuite_human_link.json#L4) makes that connection if GSuiteUser.email == Human.email. Let me see if I can roughly sketch out what the model will look like.


```python
@dataclass(frozen=True)
class GSuiteUser(CartographyNodeSchema):
    label: str = 'GSuiteUser'
    properties: ... omitted ...
    sub_resource_relationship: ... omitted ...
    other_relationships: Optional[OtherRelationships] = OtherRelationships(
        GSuiteUserToHumanRel(),
    )

@dataclass(frozen=True)
class GSuiteUserToHumanRel(CartographyRelSchema):
[...]

PR #1112, opened 2023-02-07
#1106: Test we can load data with subset of node schema fields

achantavy:
Follow-up of https://github.com/lyft/cartography/issues/1106#issuecomment-1420335812: 

This PR adds a formal test to validate this claim:

> The cartography data model also correctly handles the case where we try to ingest data that has only a subset of the fields defined on its schema.

Put another way, we test that if a node schema has more fields defined than the dict being loaded, then we correctly treat those missing fields as None.

### DIFF
--- a/tests/integration/cartography/graph/test_querybuilder_missing_fields_in_data.py
+++ b/tests/integration/cartography/graph/test_querybuilder_missing_fields_in_data.py
@@ -6,7 +6,7 @@ from tests.integration.util import check_nodes
 
 def test_load_missing_fields_in_data(neo4j_session):
     # Act: load our sample data where some items only have `property1` defined and other items only have `property2`
-    # defined.
+    # defined. `SimpleNodeSchema` includes both `property1` and `property2`.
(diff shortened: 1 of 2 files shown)


tests/integration/cartography/graph/test_querybuilder_missing_fields_in_data.py
```diff
@@ -0,0 +1,19 @@
+from cartography.client.core.tx import load
+from tests.data.graph.querybuilder.sample_data.simple_node_missing_fields import SIMPLE_NODE_MISSING_PROPS
+from tests.data.graph.querybuilder.sample_models.simple_node import SimpleNodeSchema
+from tests.integration.util import check_nodes
+
+
+def test_load_missing_fields_in_data(neo4j_session):
+    # Act: load our sample data where some items only have `property1` defined and other items only have `property2`
+    # defined.
```
ramonpetgrave64, line 9:
```suggestion
    # Act: load our sample data where some items only have `property1` defined and other items only have `property2`
    # defined. `SimpleNodeSchema` expects both `property1` and `property2`.
```

ramonpetgrave64 (approved):

ramonpetgrave64 (approved):

PR #1117, opened 2023-02-13
Allow user to specify which sync modules to run from CLI

achantavy:
The Cartography CLI currently runs all modules regardless of whether creds are configured for them or not. We've instructed users wanting to configure this behavior can implement a custom sync command as a workaround: https://lyft.github.io/cartography/dev/developer-guide.html#implementing-custom-sync-commands.

This PR makes it so that users can control this behavior from the CLI if they want.

This makes it easier for us to include extra functionality that may not be directly related to a sync job within cartography, e.g. vuln scanners. Future work will add sub-parsers to the CLI to make configuration cleaner.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -3,13 +3,13 @@ import getpass
 import logging
 import os
 import sys
-from typing import Optional, List
+from typing import Optional
 
 import cartography.config
 import cartography.sync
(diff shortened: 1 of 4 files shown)


shaunwebb-db:
When will this feature be available?

PR #1119, opened 2023-02-17
Data model: allow case-insensitive property match

achantavy:
Updates PropertyRef class and querybuilder to allow for case insensitive property matching when creating new relationships. Note that this is not supported for sub resource relationships.

Also fixes a bug where our generated query outputs `WITH i, item\nWITH i, item` if no sub resource relationship is defined but an OtherRelationship is.

### DIFF
--- a/cartography/graph/querybuilder.py
+++ b/cartography/graph/querybuilder.py
@@ -168,9 +168,7 @@ def _build_attach_sub_resource_statement(sub_resource_link: Optional[Cartography
     sub_resource_attach_template = Template(
         """
         WITH i, item
-        OPTIONAL MATCH (j:$SubResourceLabel)
-        WHERE
-            $WhereClause
+        OPTIONAL MATCH (j:$SubResourceLabel{$MatchClause})
         WITH i, item, j WHERE j IS NOT NULL
         $RelMergeClause
         ON CREATE SET r.firstseen = timestamp()
@@ -190,7 +188,7 @@ def _build_attach_sub_resource_statement(sub_resource_link: Optional[Cartography
 
     attach_sub_resource_statement = sub_resource_attach_template.safe_substitute(
         SubResourceLabel=sub_resource_link.target_node_label,
-        WhereClause=_build_where_clause_for_rel_match("j", sub_resource_link.target_node_matcher),
+        MatchClause=_build_match_clause(sub_resource_link.target_node_matcher),
         RelMergeClause=rel_merge_clause,
         SubResourceRelLabel=sub_resource_link.rel_label,
         set_rel_properties_statement=_build_rel_properties_statement('r', rel_props_as_dict),
--- a/cartography/graph/querybuilder.py
+++ b/cartography/graph/querybuilder.py
@@ -117,15 +117,16 @@ def _build_where_clause_for_rel_match(node_var: str, matcher: TargetNodeMatcher)
     :return: a Neo4j where clause
     """
     match = Template("$node_var.$key = $prop_ref")
-    imatch = Template("toLower($node_var.$key) = toLower($prop_ref)")
+    case_insensitive_match = Template("toLower($node_var.$key) = toLower($prop_ref)")
 
     matcher_asdict = asdict(matcher)
 
     result = []
     for key, prop_ref in matcher_asdict.items():
-        prop_line = match.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
         if prop_ref.ignore_case:
-            prop_line = imatch.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
+            prop_line = case_insensitive_match.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
+        else:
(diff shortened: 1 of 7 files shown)


cartography/graph/querybuilder.py
```diff
@@ -109,6 +109,27 @@ def _build_match_clause(matcher: TargetNodeMatcher) -> str:
...
+    match = Template("$node_var.$key = $prop_ref")
+    imatch = Template("toLower($node_var.$key) = toLower($prop_ref)")
+
+    matcher_asdict = asdict(matcher)
+
+    result = []
+    for key, prop_ref in matcher_asdict.items():
+        prop_line = match.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
+        if prop_ref.ignore_case:
+            prop_line = imatch.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
```
ramonpetgrave64, line 128:
```suggestion
    case_insensitive_match = Template("toLower($node_var.$key) = toLower($prop_ref)")

    matcher_asdict = asdict(matcher)

    result = []
    for key, prop_ref in matcher_asdict.items():
        if prop_ref.ignore_case:
            prop_line = case_insensitive_match.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
        else:
            prop_line = match.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
```

cartography/models/core/common.py
```diff
@@ -22,10 +22,22 @@ def __init__(self, name: str, set_in_kwargs=False, extra_index=False):
...
             Defaults to False.
+        :param ignore_case: If True, performs a case-insensitive match when comparing the value of this property during
+        relationship creation. Defaults to False. This only has effect as part of a TargetNodeMatcher, and this is not
+        supported for the sub resource relationship.
+            Example on why you would set this to True:
+            GitHub usernames can have both uppercase and lowercase characters, but GitHub itself treats usernames as
+            case-insensitive. Suppose your company's internal personnel database stores GitHub usernames all as
+            lowercase. If you wanted to map your company's employees to their GitHub identities, you would need to
+            perform a case-insensitive match between your company's record of a user's GitHub username and your
+            cartography catalog of GitHubUser nodes. Therefore, you would need `ignore_case=False` in the PropertyRef
```
ramonpetgrave64, line 33:
```suggestion
            cartography catalog of GitHubUser nodes. Therefore, you would need `ignore_case=True` in the PropertyRef
```

cartography/graph/querybuilder.py
```diff
@@ -109,6 +109,27 @@ def _build_match_clause(matcher: TargetNodeMatcher) -> str:
     return ', '.join(match.safe_substitute(Key=key, PropRef=prop_ref) for key, prop_ref in matcher_asdict.items())
 
 
+def _build_where_clause_for_rel_match(node_var: str, matcher: TargetNodeMatcher) -> str:
```
ramonpetgrave64, line 112: nonblocker: I'd rename this to something like `_build_where_clause_with_match_clause`

tests/data/graph/querybuilder/sample_models/fake_emps_githubusers.py
```diff
@@ -0,0 +1,44 @@
...
+@dataclass(frozen=True)
+class FakeEmpToGitHubUserRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class FakeEmpToGitHubUser(CartographyRelSchema):
+    target_node_label: str = 'GitHubUser'
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'username': PropertyRef('github_username', ignore_case=True)},
```
ramonpetgrave64, line 23: side note: the new changes make it a bit easier to start matching against multiple properties.

ramonpetgrave64 (changes requested):

cartography/models/core/common.py
achantavy: Damn, good eye

tests/data/graph/querybuilder/sample_models/fake_emps_githubusers.py
achantavy: I guess it makes it more flexible, though I don't want it to grow too much more beyond what it's already at.

We already support matching multiple attrs - see
https://github.com/lyft/cartography/blob/c4310e5b068ff85b9e2054eda115581ea1484883/tests/integration/cartography/graph/test_querybuilder_match_on_multiple_attrs.py#LL19C40-L19C40
and
https://github.com/lyft/cartography/blob/c4310e5b068ff85b9e2054eda115581ea1484883/tests/data/graph/querybuilder/sample_models/multiple_attr_match.py#L26-L31

ramonpetgrave64: That's right. I somehow forgot.

ramonpetgrave64 (approved):

PR #1124, opened 2023-02-21
Support args via config file

achantavy:
[ ] address TODO items
[ ] update docs on how to use a config file and the CLI args

Relies on https://github.com/lyft/cartography/pull/1123 and https://github.com/lyft/cartography/pull/1117.


achantavy:
Probably will use dynaconf or something similar instead of rolling our own

PR #1130, opened 2023-02-27
Implement Okta throttle to respect API rate limits

tmsteere:
See https://github.com/lyft/cartography/issues/1129. 

This update adds a rate limit-checking function to okta.utils. This function validates the available rate limit against the maximum rate limit and waits until the rate limit is reset if less than 10% of the rate limit remains. All of the applicable Okta ingest functions have been updated to check the rate limit before continuing. 

This fixes an existing bug where rate limit bypasses were silently ignored with a catch at debug level and a crash bug where rate limits were not considered at all during groups ingestion.

### DIFF
--- a/cartography/intel/okta/utils.py
+++ b/cartography/intel/okta/utils.py
@@ -1,7 +1,8 @@
 # Okta intel module - utility functions
+import time
 from okta.framework import PagedResults
 from okta.framework.ApiClient import ApiClient
-
+from requests import Response
 
 def is_last_page(response: PagedResults) -> bool:
     """
@@ -28,3 +29,22 @@ def create_api_client(okta_org: str, path_name: str, api_key: str) -> ApiClient:
     )
 
     return api_client
+
+def check_rate_limit(response: Response) -> int:
+    """
+    Checks if we are about to hit the rate limit and waits until reset if so
+    :param response: server response
+    :return: int representing time slept    
+    """
+    sleep_time = 0
+    rate_limit_threshold = 0.1
+    remaining = response.headers.get('x-rate-limit-remaining')
+    limit = response.headers.get('x-rate-limit-limit')
+    if remaining is not None and limit is not None:
+        if  (int(remaining) / int(limit)) < rate_limit_threshold:
+            reset_time = response.headers.get('x-rate-limit-reset')
+            if reset_time is not None:
+                sleep_time = int(reset_time) - int(time.time())
+                if sleep_time > 0:
+                    time.sleep(sleep_time)
+    return sleep_time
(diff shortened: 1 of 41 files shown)


cartography/intel/okta/utils.py
```diff
@@ -28,3 +34,27 @@ def create_api_client(okta_org: str, path_name: str, api_key: str) -> ApiClient:
...
+
+
+def check_rate_limit(response: Response) -> int:
+    """
+    Checks if we are about to hit the rate limit and waits until reset if so
+    :param response: server response
+    :return: int - time waited
+    """
+    sleep_time = 0
+    rate_limit_threshold = 0.1
```
achantavy, line 46: If an Okta customer for whatever reason has a very high API rate-limit threshold, then a ratio set too high will cause them to always hit our circuit breaker of 1 minute, effectively always raising that exception for them.

I think the better approach would be to catch the `OktaError` exception, check if it has the "API call exceeded rate limit due to too many requests." string, and then sleep until `x-rate-limit-reset`. I realize it might be too difficult to structure the code to do that right now though.

The real answer would be to upgrade the Okta SDK  to the latest version which seems to [handle rate limiting options](https://github.com/okta/okta-sdk-python#rate-limiting) for us.

That said, to unblock your scenario I am ok with going with this immediate bandaid fix, so I'd recommend writing `check_rate_limit()` like this:

```python
def check_rate_limit(response: Response) -> None:
    """
    Checks if we are about to hit the rate limit and waits until reset if so
    :param response: server response
    """
    rate_limit_threshold = 0.1

    remaining = response.headers.get('x-rate-limit-remaining')
    limit = response.headers.get('x-rate-limit-limit')
    reset_time = response.headers.get('x-rate-limit-reset')

    if remaining and limit and reset_time:
        if (int(remaining) / int(limit)) < rate_limit_threshold:
            sleep_time_seconds = int(reset_time) - int(time.time())
            if sleep_time_seconds <= 0:
                # A negative sleep time does not make sense so treat it the same as a 0 sleep time
                return
            if sleep_time_seconds > 60:
                raise ValueError(
                    f"Okta API limit exceeded. Sleep time of {sleep_time_seconds} would exceed one minute. Crashing "
                    f"Okta sync to avoid blocking."
                )
            logger.warning(f"Okta rate limit threshold reached. Waiting {sleep_time_seconds} seconds.")
            time.sleep(sleep_time_seconds)
```

This way
- we have fewer levels of `if`.
- `logger.exception` does not raise an exception, it only logs the exception object if one was raised, so we should instead raise the exception there.
- I removed the return value because it's not being used anywhere. This will require changes to the unit tests: we'd use the `mock_sleep` passed to the test function and call `mock_sleep.assert_called_with(<YOUR_VALUE_HERE>)` to verify that we're sleeping with the expected values.

---

Then in the mid to longer term, help updating the Okta library would be appreciated (-:

achantavy (changes requested):

cartography/intel/okta/utils.py
tmsteere: Okta API rate limits are in one-minute increments, so no matter how high their API limit is, a one-minute circuit breaker should catch any failure state.

achantavy (approved):
Thank you!

PR #1134, opened 2023-03-10
Fixes #1133: AWS: IAM: Roles: merge roles by principal first

### LINKED ISSUE #1133: When syncing multiple accounts, Cartography can incorrectly represent a single AWSRole as 2 nodes
*Description*:

When syncing multiple accounts starting with a fresh graph, depending on the order of the accounts being synced, Cartography may incorrectly represent a single AWS role as a AWSPrincipal node and a separate AWSRole node - both with the same arn. The expected behavior is for a single node to be created with labels AWSPrincipal and AWSRole.

*To Reproduce*:

Suppose we have AWS accounts with ids 123 and 456 such that

1. 123 has role 'datateam' and 456 has role 'terraform'.

2. 'arn:aws:iam::456:role/terraform' allows arn:aws:iam::123:role/datateam' to assume it.

That is, 'arn:aws:iam::456:role/terraform' has trust relationship document

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "",
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "arn:aws:iam::123:role/datateam",
                ]
            },
            "Action": "sts:AssumeRole"
        }
    ]
}
```

Start with a blank graph. Configure the cartography sync to sync account 456 first, and account 123 after.

## Expected behavior

At the end of this, we will have the following paths:

```cypher
(:AWSPrincipal:AWSRole{arn:"arn:aws:iam:456::role/terraform"})-[:TRUSTS_AWS_PRINCIPAL]->(:AWSPrincipal:AWSRole{arn:"arn:aws:iam:123::role/datateam"}))
```

## Actual behavior

The following paths are created:

```cypher
(:AWSPrincipal:AWSRole{arn:"arn:aws:iam:456::role/terraform"})-[:TRUSTS_AWS_PRINCIPAL]->(:AWSPrincipal{arn:"arn:aws:iam:123::role/datateam"}))

(:AWSRole{arn:"arn:aws:iam:123::role/datateam"}) // <-- this node is disconnected and has same arn as above!
```

0.75.1 

The offending line of code is here: https://github.com/lyft/cartography/blob/d94ef53cbf5b23ecf132d472a8952a9cad1b89ce/cartography/intel/aws/iam.py#L325

We should also fix this similar pattern too: https://github.com/lyft/cartography/blob/d94ef53cbf5b23ecf132d472a8952a9cad1b89ce/cartography/intel/aws/redshift.py#L117

We should add an integration test to reproduce the error above and validate a fix.

ramonpetgrave64:
Fixes #1133 

This PR:

- When merging AWSRoles, first merges as an AWSPrincipal and then adds the AWSRole label.
- Modifies a test to verify the fix.

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -309,11 +309,15 @@ def load_roles(
     neo4j_session: neo4j.Session, roles: List[Dict], current_aws_account_id: str, aws_update_tag: int,
 ) -> None:
     ingest_role = """
-    MERGE (rnode:AWSRole{arn: $Arn})
-    ON CREATE SET rnode:AWSPrincipal, rnode.roleid = $RoleId, rnode.firstseen = timestamp(),
-    rnode.createdate = $CreateDate
-    ON MATCH SET rnode.name = $RoleName, rnode.path = $Path
-    SET rnode.lastupdated = $aws_update_tag
+    MERGE (rnode:AWSPrincipal{arn: $Arn})
+    ON CREATE SET rnode.firstseen = timestamp()
+    SET
+        rnode:AWSRole,
+        rnode.roleid = $RoleId,
+        rnode.createdate = $CreateDate,
+        rnode.name = $RoleName,
+        rnode.path = $Path,
+        rnode.lastupdated = $aws_update_tag
     WITH rnode
     MATCH (aa:AWSAccount{id: $AWS_ACCOUNT_ID})
     MERGE (aa)-[r:RESOURCE]->(rnode)
--- a/tests/integration/cartography/intel/aws/iam/test_iam.py
+++ b/tests/integration/cartography/intel/aws/iam/test_iam.py
@@ -6,6 +6,7 @@ import tests.data.aws.iam
 from cartography.cli import CLI
 from cartography.config import Config
 from cartography.sync import build_default_sync
+from tests.integration.util import check_nodes
 
 TEST_ACCOUNT_ID = '000000000000'
 TEST_REGION = 'us-east-1'
@@ -64,15 +65,50 @@ def test_load_groups(neo4j_session):
     )
 
 
-def test_load_roles(neo4j_session):
-    data = tests.data.aws.iam.LIST_ROLES['Roles']
+def _get_principal_role_nodes(neo4j_session):
+    '''
+    Get AWSPrincipal node tuples (rolearn, arn) that have arns with substring `:role/`
+    '''
+    return {
+        (roleid, arn)
+        for (roleid, arn) in check_nodes(neo4j_session, 'AWSPrincipal', ['roleid', 'arn'])
+        if ':role/' in arn # filter out other Principals nodes, like the ec2 service princiapl
+    }
 
+
+def test_load_roles(neo4j_session):
+    '''
+    Ensures that we load AWSRoles without duplicating against AWSPrincipal nodes
+    '''
+    assert set() == _get_principal_role_nodes(neo4j_session)
+    data = tests.data.aws.iam.LIST_ROLES['Roles']
+    expected_principals = {
+        (None, item['Arn'], )
+        for item in data
+    }
+    # Load a the roles as Principals, initially.
+    neo4j_session.run(
+        '''
+        UNWIND $data as item
+            MERGE (p:AWSPrincipal{arn: item.Arn})
+        ''',
+        data=data
+    )
+    actual_principals = _get_principal_role_nodes(neo4j_session)
+    assert expected_principals == actual_principals
+    assert set() == check_nodes(neo4j_session, 'AWSRole', ['arn'])
+    # Load the roles normally
     cartography.intel.aws.iam.load_roles(
         neo4j_session,
         data,
         TEST_ACCOUNT_ID,
         TEST_UPDATE_TAG,
     )
+    # Ensure that the new AWSRoles are merged into pre-existing AWSPrincipal nodes,
+    # and we do not have duplicate AWSPrincipal nodes.
+    role_nodes = check_nodes(neo4j_session, 'AWSRole', ['roleid', 'arn'])
+    principal_nodes = _get_principal_role_nodes(neo4j_session)
+    assert role_nodes == principal_nodes
 
 
 def test_load_roles_creates_trust_relationships(neo4j_session):


tests/integration/cartography/intel/aws/iam/test_iam.py
```diff
@@ -64,15 +65,50 @@ def test_load_groups(neo4j_session):
...
         neo4j_session,
         data,
         TEST_ACCOUNT_ID,
         TEST_UPDATE_TAG,
     )
+    # Ensure that the new AWSRoles are merged into pre-existing AWSPrincipal nodes,
+    # and we do not have duplicate AWSPrincipal nodes.
+    role_nodes = check_nodes(neo4j_session, 'AWSRole', ['roleid', 'arn'])
+    principal_nodes = _get_principal_role_nodes(neo4j_session)
+    assert role_nodes == principal_nodes
```
achantavy, line 111: Put the actual value on the right-hand side; Pycharm (and I think Pytest) seems to expect that assertions have the expected values on the LHS and the actual values on the RHS

tests/integration/cartography/intel/aws/iam/test_iam.py
```diff
@@ -64,15 +65,50 @@ def test_load_groups(neo4j_session):
...
     cartography.intel.aws.iam.load_roles(
         neo4j_session,
         data,
         TEST_ACCOUNT_ID,
         TEST_UPDATE_TAG,
     )
+    # Ensure that the new AWSRoles are merged into pre-existing AWSPrincipal nodes,
+    # and we do not have duplicate AWSPrincipal nodes.
+    role_nodes = check_nodes(neo4j_session, 'AWSRole', ['roleid', 'arn'])
+    principal_nodes = _get_principal_role_nodes(neo4j_session)
```
achantavy, line 110: To make this test easier to read, I think we might want to use hardcoded values like this:

```python
assert {
 ('AROA00000000000000000', 'arn:aws:iam::000000000000:role/example-role-0'),
 ('AROA00000000000000001', 'arn:aws:iam::000000000000:role/example-role-1'),
 ('AROA00000000000000002', 'arn:aws:iam::000000000000:role/example-role-2'),
 ('AROA00000000000000003', 'arn:aws:iam::000000000000:role/example-role-3'),
} == check_nodes(neo4j_session, 'AWSRole', ['roleid', 'arn'])
```

tests/integration/cartography/intel/aws/iam/test_iam.py
```diff
@@ -64,15 +65,50 @@ def test_load_groups(neo4j_session):
...
+    Ensures that we load AWSRoles without duplicating against AWSPrincipal nodes
+    '''
+    assert set() == _get_principal_role_nodes(neo4j_session)
     data = tests.data.aws.iam.LIST_ROLES['Roles']
-
+    expected_principals = {
+        (None, item['Arn'])
+        for item in data
+    }
+    # Load a the roles as Principals, initially.
```
achantavy, line 89:
```suggestion
    # Arrange: Load the roles as bare Principals without other labels. This replicates the case where we discover a
    # role from another account via an AssumeRolePolicy document or similar ways. See #1133.```

File: cartography/intel/aws/iam.py [hunk 4]
```diff
@@ -309,11 +309,15 @@ def load_roles(
     neo4j_session: neo4j.Session, roles: List[Dict], current_aws_account_id: str, aws_update_tag: int,
 ) -> None:
     ingest_role = """
-    MERGE (rnode:AWSRole{arn: $Arn})
-    ON CREATE SET rnode:AWSPrincipal, rnode.roleid = $RoleId, rnode.firstseen = timestamp(),
-    rnode.createdate = $CreateDate
-    ON MATCH SET rnode.name = $RoleName, rnode.path = $Path
-    SET rnode.lastupdated = $aws_update_tag
+    MERGE (rnode:AWSPrincipal{arn: $Arn})
```
achantavy, line 312: Moving things from the `ON CREATE SET` and the `ON MATCH SET` is good.

However I don't think we should merge by the principal first. From #1133, the only situations in cartography where we create these "bare/husk/whatever a good name for them is" principal nodes are

1. When we [parse AssumeRoleDocuments](https://github.com/lyft/cartography/blob/d94ef53cbf5b23ecf132d472a8952a9cad1b89ce/cartography/intel/aws/iam.py#L325) -- in this case, we should definitely set the `AWSRole` label there, because assumerole only happens for roles and not users or groups.
2. When we [handle redshift IAM role attachments](https://github.com/lyft/cartography/blob/d94ef53cbf5b23ecf132d472a8952a9cad1b89ce/cartography/intel/aws/redshift.py#L117) -- in this case, we should also definitely set the `AWSRole` label there because the field in the API data is called ["IamRoleArn"](https://github.com/lyft/cartography/blob/d94ef53cbf5b23ecf132d472a8952a9cad1b89ce/cartography/intel/aws/redshift.py#L126), so it wouldn't make sense to keep that as a bare principal.

achantavy (changes requested):

cartography/intel/aws/iam.py
ramonpetgrave64: We probably should be setting those as Roles, but It's also account roots, users, and federated identities that can assume roles.

- https://aws.amazon.com/blogs/security/how-to-use-trust-policies-with-iam-roles/
- https://stackoverflow.com/questions/66414774/can-iam-user-assume-a-role-if-the-user-is-not-described-in-the-trust-policy-of

tests/integration/cartography/intel/aws/iam/test_iam.py
```diff
@@ -64,15 +65,50 @@ def test_load_groups(neo4j_session):
...
+    return {
+        (roleid, arn)
+        for (roleid, arn) in check_nodes(neo4j_session, 'AWSPrincipal', ['roleid', 'arn'])
+        if ':role/' in arn  # filter out other Principals nodes, like the ec2 service princiapl
+    }
+
+
 def test_load_roles(neo4j_session):
+    '''
+    Ensures that we load AWSRoles without duplicating against AWSPrincipal nodes
```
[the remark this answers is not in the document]
ramonpetgrave64: Changed.

tests/integration/cartography/intel/aws/iam/test_iam.py
ramonpetgrave64: changed

tests/integration/cartography/intel/aws/iam/test_iam.py
ramonpetgrave64: added

tests/integration/cartography/intel/aws/iam/test_iam.py
ramonpetgrave64: changed

cartography/intel/aws/iam.py
achantavy: We talked about this:

- We aren't going to parse ARNs yet to determine the specific principal type. This will come in a future PR.
- When syncing trust relationships, we don't know the type of the trusted principal unless we parse the ARN. Since we aren't doing that yet, we will continue to merge with the generic label -- that is, we will continue to do `merge :AWSPrincipal` instead of `merge :<specific label>`.

Therefore when we reach sync_roles, we will do `MERGE :AWSPrincipal` to solve #1133.

achantavy (approved):

PR #1135, opened 2023-03-14
Fix EBS cleanup job typo

achantavy:
Fix: changes EBS volume cleanup job from `ATTACHED_TO` => `ATTACHED_TO_EC2_INSTANCE`.

---

This is already annotated in docs here as `ATTACHED_TO_EC2_INSTANCE`: https://github.com/lyft/cartography/blob/master/docs/root/modules/aws/schema.md#relationships-69

and this is created in the code itself as `ATTACHED_TO_EC2_INSTANCE`: https://github.com/lyft/cartography/blob/2903dbe5693efe43dcd578397ef2d864571d5946/cartography/intel/aws/ec2/volumes.py#L82

So the cleanup job here is wrong.

### DIFF
--- a/cartography/data/jobs/cleanup/aws_import_ec2_instances_cleanup.json
+++ b/cartography/data/jobs/cleanup/aws_import_ec2_instances_cleanup.json
@@ -50,17 +50,17 @@
       "iterationsize": 100
     },
     {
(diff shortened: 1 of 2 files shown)


achantavy:
Obsolete after https://github.com/cartography-cncf/cartography/pull/1214

PR #1136, opened 2023-03-14
Only run aws analysis jobs if dependent syncs finished

achantavy:
AWS analysis jobs now declare what sync resources must have succeeded before they are allowed to run. This ensures that we don't run analysis on partial or incorrect graph data.

### DIFF
--- a/tests/unit/cartography/test_util.py
+++ b/tests/unit/cartography/test_util.py
@@ -145,7 +145,7 @@ def test_run_analysis_and_ensure_deps_no_requirements(mock_run_analysis_job: moc
     # Act
     run_analysis_and_ensure_deps(
         'aws_foreign_accounts.json',
-        set(),  # This job has no requirements
+        {'iam'},
         requested_syncs,
         common_job_parameters,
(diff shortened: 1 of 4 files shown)


tests/unit/cartography/test_util.py
```diff
@@ -92,3 +95,65 @@ def test_batch(mocker):
...
+    requested_syncs = {
+        'ec2:instance',
+        'iam',
+        'resourcegroupstaggingapi',
+    }
+
+    # Act
+    run_analysis_and_ensure_deps(
+        'aws_foreign_accounts.json',
+        set(),  # This job has no requirements
```
ramonpetgrave64, line 148: This is okay, but I'd rather this have something inside.
```suggestion
        {'iam'},
```

ramonpetgrave64 (approved):

ramonpetgrave64:
This is great!

ramonpetgrave64 (dismissed):

ramonpetgrave64 (approved):

PR #1138, opened 2023-03-14
[DRAFT] Sync AWS accounts in parallel

achantavy:


cartography/intel/aws/__init__.py
```diff
@@ -153,43 +198,32 @@ def _sync_multiple_accounts(
...
+                    sync_one_account_runner,
+                    profile_name,
+                    account_id,
+                    boto3_session,
+                    neo4j_thread_session,
+                    sync_tag,
+                    common_job_parameters,
+                    aws_requested_syncs,
+                    aws_best_effort_mode,
+                ).result()
```
alowde-ps, line 223: I believe this will wait for each execution in turn (consecutive execution) as the call to the `result()` method of the Future returned by `executor.submit` will wait for the future to complete.
To get concurrent execution I believe you would want to submit the sync functions in a loop as is done here, collect the returned Future objects in a list and then call `concurrent.futures.wait` to wait for all of the syncs to be complete (or iterate over `concurrent.futures.as_completed` if you want to report sync statuses as they complete).

PR #1146, opened 2023-03-20
EC2 Instance sync => new data model

achantavy:
Refactors EC2 instance sync to use the cartography data model. Adds previously missing tests.

This is not a breaking change as the ID conventions are still the same as before (e.g. using EC2Instance instanceid for id instead of arn), even if they are inconsistent with other modules.

A follow-up PR will standardize IDs using ARNs.


cartography/intel/aws/ec2/instances.py
```diff
@@ -21,410 +29,343 @@
...
+                    sg_list.append(
+                        {
+                            'GroupId': group['GroupId'],
+                            'InstanceId': instance_id,
+                        },
+                    )
 
+            for network_interface in instance['NetworkInterfaces']:
+                for security_group in network_interface['Groups']:
+                    network_interface_list.append({
```
achantavy, line 125: @resilience-jychp - remember how we talked about cardinality? This is how I'm dealing with it in a transform function. In this case, an instance can have multiple network interfaces, and each of those can have multiple security groups, so I'll just add them each as their own entry to the list.

cartography/models/aws/ec2/securitygroups.py
```diff
@@ -0,0 +1,64 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class EC2SecurityGroupNodeProperties(CartographyNodeProperties):
+    # arn: PropertyRef = PropertyRef('Arn', extra_index=True) # TODO decide on this
```
achantavy, line 16: Followed up on in https://github.com/lyft/cartography/pull/1147

cartography/intel/aws/ec2/instances.py
resilience-jychp: @achantavy did the same, not totaly convinced it's the best way to do it, but definitively do the job

resilience-jychp (dismissed):
This model migration seems OK
I have a much better understanding of the role of the SubResource relationship now

Did not check, but maybe module doc need to be updated after this migration ?

cartography/intel/aws/ec2/instances.py
```diff
@@ -21,410 +29,343 @@
...
-            ON CREATE SET r.firstseen = timestamp()
-            SET r.lastupdated = $update_tag
-    """
-    tx.run(
-        query,
-        Interfaces=instance_data['NetworkInterfaces'],
-        InstanceId=instance_data['InstanceId'],
-        update_tag=update_tag,
-    )
+def transform_ec2_instances(
```
achantavy, line 38: @ramonpetgrave64 - what you think about this big function, should I split it?

tests/integration/cartography/intel/aws/test_ssm.py
```diff
@@ -1,36 +1,35 @@
...
+        boto3_session,
+        [TEST_REGION],
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+        {'UPDATE_TAG': TEST_UPDATE_TAG, 'AWS_ID': TEST_ACCOUNT_ID},
     )
 
 
-def test_load_instance_information(neo4j_session):
+@patch.object(cartography.intel.aws.ec2.instances, 'get_ec2_instances', return_value=DESCRIBE_INSTANCES['Reservations'])
```
ramonpetgrave64, line 29: nonblocker: same comment about patches

cartography/intel/aws/ec2/instances.py
ramonpetgrave64: I'd rather that return such a big tuple. Maybe convert it to a namedtuple, so we don't have to be so careful about the ordering?

It's a big one, but if we break it into many functions, we would be repeating a bit of code to  dig deep into the objects. I'm okay with it either way.

cartography/intel/aws/ec2/instances.py
```diff
@@ -21,410 +29,343 @@
...
-                mapping['InstanceId'] = instance["InstanceId"]
-                instance_ebs_volumes_list.append(mapping)
-    return instance_ebs_volumes_list
-
-
-def _load_ec2_instance_ebs_tx(
-        tx: neo4j.Transaction,
-        ebs_data: List[Dict[str, Any]],
-        update_tag: int,
+def load_ec2_instance_nodes(
```
ramonpetgrave64, line 268: copy pasta. duplicated at `load_instances()`

cartography/models/aws/ec2/keypairs.py
```diff
@@ -0,0 +1,64 @@
...
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class EC2KeyPairNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef('KeyPairArn')
+    arn: PropertyRef = PropertyRef('KeyPairArn', extra_index=True)
+    keyname: PropertyRef = PropertyRef('KeyName')
+    region: PropertyRef = PropertyRef('Region', set_in_kwargs=True)
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
```
ramonpetgrave64, line 20: This schema doesn't include other props like keyfingerprint. I understand that we can really only get that data in a separate `sync_ec2_key_pars`, but should we add the property now or later, if we intend to reuse this class?

cartography/intel/aws/ec2/instances.py
```diff
@@ -21,410 +29,343 @@
...
-@timeit
-def load_ec2_instance_network_interfaces(neo4j_session: neo4j.Session, instance_data: Dict, update_tag: int) -> None:
-    neo4j_session.write_transaction(_load_ec2_instance_net_if_tx, instance_data, update_tag)
+            if instance.get("KeyName"):
+                key_name = instance["KeyName"]
+                key_pair_arn = f'arn:aws:ec2:{region}:{current_aws_account_id}:key-pair/{key_name}'
+                keypair_list.append({
+                    'KeyPairArn': key_pair_arn,
+                    'KeyName': key_name,
+                    'InstanceId': instance_id,
```
ramonpetgrave64, line 111: Strange to set an instanceId here, since 1 keypair can be used with many instances, but it will only be used for the TargetNodeMatcher, and duplicates keypair dicts will still be merged to onto one node.

cartography/models/aws/ec2/keypairs.py
```diff
@@ -0,0 +1,64 @@
...
+    rel_label: str = "SSH_LOGIN_TO"
+    properties: EC2KeyPairToEC2InstanceRelProperties = EC2KeyPairToEC2InstanceRelProperties()
+
+
+@dataclass(frozen=True)
+class EC2KeyPairSchema(CartographyNodeSchema):
+    label: str = 'EC2KeyPair'
+    properties: EC2KeyPairNodeProperties = EC2KeyPairNodeProperties()
+    sub_resource_relationship: EC2KeyPairToAWSAccount = EC2KeyPairToAWSAccount()
+    other_relationships: OtherRelationships = OtherRelationships(
```
ramonpetgrave64, line 60: Can you also convert the ec2_key_pair_sync, just so we have a good example of re-using NodeSchemas from multiple directions?  syncing instances to discover key pairs,  then syncing key pairs independently, while reusing the same NodeSchema class.

cartography/models/aws/ec2/keypairs.py
achantavy: I followed up on this in https://github.com/lyft/cartography/pull/1147.

cartography/models/aws/ec2/keypairs.py
achantavy: I followed up on this in https://github.com/lyft/cartography/pull/1147 -- this PR is already pretty huge, can we defer this until #1147?

cartography/intel/aws/ec2/instances.py
achantavy: Good callout, I'll look more closely.

cartography/intel/aws/ec2/instances.py
achantavy: Damn, good catch thank you

cartography/intel/aws/ec2/instances.py
achantavy: We talked about this a bit offline - I think this will be solved when we refactor the keypair sync to the new model

Additioanlly, an EC2 instance schema itself only defines a single EC2 keypair so I think we're good here: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Instance.html

tests/integration/cartography/intel/aws/test_resourcegroupstaggingapi.py
```diff
@@ -14,19 +17,28 @@
...
+        boto3_session,
+        [TEST_REGION],
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+        {'UPDATE_TAG': TEST_UPDATE_TAG, 'AWS_ID': TEST_ACCOUNT_ID},
     )
 
 
-def test_transform_and_load_ec2_tags(neo4j_session):
+@patch.object(cartography.intel.aws.ec2.instances, 'get_ec2_instances', return_value=DESCRIBE_INSTANCES['Reservations'])
```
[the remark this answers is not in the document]
achantavy: Hm, I'm going to defer this one, we can refactor the test in a follow-up.

achantavy:
> Did not check, but maybe module doc need to be updated after this migration ?

Nah, this PR keeps all the fields the same so should be good.

ramonpetgrave64 (approved):

PR #1147, opened 2023-03-22
Standardize on ARN for ID with ec2 instances, keypairs, nics, securitygroups, subnets , volumes

achantavy:
Also update data model to support nodes being updated in multiple places.

Relies on https://github.com/lyft/cartography/pull/1146.

Partially addresses #1024.


achantavy:
this is stale and out of date and doesn't quite work. will need a better thought out approach since some aws assets dont have arns so it wouldnt work  to standardize on arn for them.

PR #1154, opened 2023-04-06
EBSVolume => New data model; Allow node attr updates from multiple intel modules

achantavy:
Refactors EBS Volume details to new data model.

Additionally, fixes a bug that prevented a node from being updated correctly by multiple intel modules.

**Expected behavior**:
After refactoring the EBS Volume sync to the new model, we can run the EC2 instance sync and the EBS volume sync one after the other and see the `deleteontermination` field on the EBSVolumes that are attached to EC2 instances.

**Actual behavior**:
After refactoring, when I run the EC2 instance sync and the EBS volume sync one after the other, the `deleteontermination` field is not present.

**Root cause**:
The EC2 instance sync loads in :EC2Instances with :EBSVolumes attached based on the output of `describe-ec2-instances`. This data includes a field called `deleteontermination` on the :EBSVolume. `deleteontermination` is only set during the EC2 instance sync.

Next when the EBS Volume sync runs, it enriches existing :EBSVolumes with additional fields from data retrieved with `describe-ebs-volumes` but because this API call does not include `deleteontermination`, we will overwrite the previous values of `deleteontermination` with `None`, thus removing the property from all :EBSVolumes.

**Fix**:
This PR updates the `PropertyRef` implementation to use `COALESCE()` so that any existing values already defined for a given node property are preserved. This is quirky so I have also tried to thoroughly update documentation.


cartography/intel/aws/util/arns.py
```diff
@@ -0,0 +1,18 @@
+from typing import Optional
+
+
+def build_arn(
```
ramonpetgrave64, line 4: 👍

cartography/intel/aws/ec2/volumes.py
```diff
@@ -26,90 +28,76 @@ def get_volumes(boto3_session: boto3.session.Session, region: str) -> List[Dict]
...
+            'State': volume['State'],
+            'OutpostArn': volume['OutpostArn'],
+            'SnapshotId': volume['SnapshotId'],
+            'Iops': volume['Iops'],
+            'FastRestored': volume['FastRestored'],
+            'MultiAttachEnabled': volume['MultiAttachEnabled'],
+            'VolumeType': volume['VolumeType'],
+            'VolumeId': volume_id,
+            'KmsKeyId': volume['KmsKeyId'],
+        })
```
ramonpetgrave64, line 52: No need for parenthesis
```suggestion
        raw_vol = {
            'Arn': build_arn('ec2', current_aws_account_id, 'volume', volume_id, region),
            'AvailabilityZone': volume['AvailabilityZone'],
            'CreateTime': volume['CreateTime'],
            'Encrypted': volume['Encrypted'],
            'Size': volume['Size'],
            'State': volume['State'],
            'OutpostArn': volume['OutpostArn'],
            'SnapshotId': volume['SnapshotId'],
            'Iops': volume['Iops'],
            'FastRestored': volume['FastRestored'],
            'MultiAttachEnabled': volume['MultiAttachEnabled'],
            'VolumeType': volume['VolumeType'],
            'VolumeId': volume_id,
            'KmsKeyId': volume['KmsKeyId'],
        }
```

cartography/intel/aws/ec2/volumes.py
```diff
@@ -26,90 +28,76 @@ def get_volumes(boto3_session: boto3.session.Session, region: str) -> List[Dict]
...
+
+        if not active_attachments:
+            result.append(raw_vol)
+            continue
+
+        for attachment in active_attachments:
+            vol_with_attachment = raw_vol.copy()
+            vol_with_attachment['InstanceId'] = attachment['InstanceId']
```
ramonpetgrave64, line 60: We do this because volumes can be attached to multiple instances?
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volumes-multi.html

cartography/intel/aws/ec2/volumes.py
```diff
@@ -26,90 +28,76 @@ def get_volumes(boto3_session: boto3.session.Session, region: str) -> List[Dict]
...
     for volume in volumes:
-        volume['VolumeArn'] = f"arn:aws:ec2:{region}:{current_aws_account_id}:volume/{volume['VolumeId']}"
-        volume['CreateTime'] = str(volume['CreateTime'])
-    return volumes
+        attachments = volume.get('Attachments', [])
+        active_attachments = [a for a in attachments if a['State'] == 'attached']
+
+        volume_id = volume['VolumeId']
+        raw_vol = ({
+            'Arn': build_arn('ec2', current_aws_account_id, 'volume', volume_id, region),
```
ramonpetgrave64, line 38: Except for `Arn`, should we bother being explicit about the rest of the props?
We could do 
```
raw_vol = {
'Arn': build_arn('ec2', current_aws_account_id, 'volume', volume_id, region),,
**volume
}
```

cartography/models/core/common.py
```diff
@@ -34,25 +45,51 @@ def __init__(self, name: str, set_in_kwargs=False, extra_index=False, ignore_cas
...
-        On the other hand if set_in_kwargs is True, then the value will instead come from kwargs passed to
-        querybuilder.build_ingestion_query(). This is used for things like applying the same update tag to all nodes of
-        a given run.
-        """
-        return f"item.{self.name}" if not self.set_in_kwargs else self._parameterize_name()
+        # Attention: This implementation detail is quirky and deserves the following essay to explain.
+        # This function ensures that the Neo4j query sets the value of this node/relationship property by checking the
+        # following sources in order:
+        # 1. the current dict being processed. That is, we check if dict `item` contains a key called `self.name` with
+        # a non-None value. If so, we are done. Else continue and check the next source.
```
ramonpetgrave64, line 79:
```suggestion
        # a non-None value. If so, we use that value. Else continue and check the next source.
```

cartography/models/core/common.py
```diff
@@ -34,25 +45,51 @@ def __init__(self, name: str, set_in_kwargs=False, extra_index=False, ignore_cas
...
+        # 2. An existing value on the node itself. That is, we convert self.name to lowercase and check if that is non
+        # null on the Neo4j node `i` already. If None, then the property is not set (this a Neo4j driver behavior).
+        # This means we make an ASSUMPTION that the property name set on the node is the lowercase version of self.name.
```
ramonpetgrave64, line 82: We might need to enforce that PropertyRef variables on NodeSchemas are all lowercase. And also enforce a standard between the variable names and the PropertyRef inputs.

```
myprop = PropertyRef('MyProp')
```

instead of 

```
mycoolprop = PropertyRef('MyProp')
```

ramonpetgrave64 (changes requested):

cartography/models/core/common.py
achantavy: I agree. Or we need to find another way to allow different intel modules to update the same node types using the same data schema objects. I don't know, can you think of a better way to solve this?

ramonpetgrave64: For now, another way would be to use subclasses. The ec2/volumes module can have a base version of `EBSVolumeNodeProperties` that does not include `deleteontermination`, while the ec2/instances module has a subclass of `EBSVolumeNodeProperties` that does include that extra property.

```
@dataclass(frozen=True)
class EBSVolumeNodePropertiesFromInstances(EBSVolumeNodeProperties):
    deleteontermination: PropertyRef = PropertyRef('DeleteOnTermination')

@dataclass(frozen=True)
class EBSVolumeSchemaFromInstances(EBSVolumeSchema):
    properties: EBSVolumeNodePropertiesFromInstances = EBSVolumeNodePropertiesFromInstances()
```

This way only the ec2/instances module can overwrite the `deleteontermination` property.

One problem with that: What if we wanted the subclass to have fewer props that the the base class? Then we would need something that could make the `asdict` method preclude the undesired props.

cartography/intel/aws/ec2/volumes.py
achantavy: Good point, I don't have a strong opinion here. I think being explicit is nice because it lets a reader see what fields are being processed.

achantavy:
Going to close this PR, re-file it as an issue, and try a different approach using separate classes for properties. I think this solution is too quirky and complicated.

PR #1155, opened 2023-04-10
fix OktaUserFactor cleanup okta_import_cleanup.json

ramonpetgrave64:
This cleanup job for OktaUserFactor uses the same label `n` twice, once for OktaUser and again for OktaUserFactor in the same query.

Neo4j will not error, but the query does not return anything, so cleanups will not happen.

```
MATCH (:OktaOrganization)-[:RESOURCE]->(n:OktaUser)-[:FACTOR]->(n:OktaUserFactor) return n
```

This PR fixes it to just `(n:OktaUserFactor)`

### DIFF
--- a/cartography/data/jobs/cleanup/okta_import_cleanup.json
+++ b/cartography/data/jobs/cleanup/okta_import_cleanup.json
@@ -1,7 +1,7 @@
 {
   "statements": [
     {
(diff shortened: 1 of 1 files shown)


achantavy (approved):
Thank you thank you

PR #1158, opened 2023-04-14
Update Dockerfile to focal and reorder steps

resilience-jychp:
This PR includes :
- upgrade from bionic to focal (bionic is EOL this month)
- reorder some step to allow layer cach for pip dependencies

Focal use python3.8, so everything should be OK.

Reordering steps allow pip dependencies install before all source code is copied (this is quite anoying for local tests ...)


achantavy:
It looks reasonable to me. @ryan-lane knows more a lot more about docker than me though - what do you think?

Dockerfile
```diff
@@ -7,15 +7,18 @@ RUN apt-get update && \
...
 # Installs pip supported by python3.8
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3.8 get-pip.py
 
+# Installs python dependencies
+COPY setup.py test-requirements.txt ./
 RUN pip install -e . && \
     pip install -r test-requirements.txt
 
+# Install cartography
+COPY . /srv/cartography
```
ramonpetgrave64, line 20: Thanks for taking care to upgrade. Can you explain what this line is for?

resilience-jychp: Before : 
copying source code was done before installing dependencies, so any change will trigger dependencies install

After :
first only requirements are copied, so the docker layer is cached
then files are copied, so change in files wont trigger a new dependencies install

achantavy (approved):

PR #1163, opened 2023-04-22
feat: add Trivy AWS ECR scan support

achantavy:
Adds support to ingest [Trivy](https://trivy.dev/latest/) JSON scan results to cartography from an S3 bucket.

Graph
image

- https://www.youtube.com/watch?v=F4EFHK21Et0 

### Checklist

Running successfully with no prefix
Screenshot 2025-06-15 at 9 36 23 PM

Running successfully with a prefix
Screenshot 2025-06-15 at 9 36 43 PM

----

Future improvements:
- Accumulating: currently we load the findings per image to Neo4j. It would likely be faster to accumulate all the scan results in memory and then write to neo4j at once. This can come in future.

This is a huge PR, but for what it's worth (lol), 1400 lines are just test data, and several hundred more are test code.


cartography/intel/aws/ecr.py
```diff
@@ -107,9 +107,12 @@ def load_ecr_repositories(
 def transform_ecr_repository_images(repo_data: Dict) -> List[Dict]:
     """
     Ensure that we only load ECRImage nodes to the graph if they have a defined imageDigest field.
+    Process repositories in a consistent order to handle overlapping image digests deterministically.
```
achantavy, line 110: This is to make the tests happy because sometimes the ordering gets messed up. I don't think there's much impact

cartography/intel/trivy/scanner.py
```diff
@@ -0,0 +1,391 @@
...
+        ecr_images: List of ECR image tuples (region, image_tag, image_uri, repo_name, image_digest)
+        boto3_session: boto3 session for dependency injection
+
+    Returns:
+        List of tuples (region, image_tag, image_uri, repo_name, image_digest, s3_object_key)
+        for images that have corresponding scan results in S3
+    """
+    s3_client = boto3_session.client("s3")
+
+    # Use the graph data to create a mapping of image URIs to their metadata for efficient lookup from the s3 listing
```
achantavy: I'm not particularly happy about this. But here's what I'm trying to do:

A graph with ECR images in it gives us image URI + image digest. An s3 bucket with trivy json files in it gives us image URIs. We don't know the digest of the image.

So, we do some data wrangling: 

1. we query the graph to find `ECRImage` nodes to get their image digests and the URIs of their attached `ECRRepositoryImage` nodes. 

2. And then each item in s3 has the naming convention "<image URI>.json", we take that image URI, and then read the json in that file to create the nodes and relationships.

We could just do (2), but then we wouldn't be able to attach packages and scan findings to the `ECRImage` nodes. They would have to go to the `ECRRepositoryImage` nodes. 

For background, the design choice to split ECRImage and ECRRepositoryImage was made in case we wanted to faithfully represent images with different binary contents and different tags living in different repos.

cartography/sync.py
```diff
@@ -70,6 +71,7 @@
         "semgrep": cartography.intel.semgrep.start_semgrep_ingestion,
         "snipeit": cartography.intel.snipeit.start_snipeit_ingestion,
         "tailscale": cartography.intel.tailscale.start_tailscale_ingestion,
+        "trivy": cartography.intel.trivy.start_trivy_ingestion,
```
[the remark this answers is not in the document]
achantavy: good bot

docs/root/modules/trivy/schema.md
```diff
@@ -0,0 +1,91 @@
...
+    ```
+
+### Package
+Representation of a package installed in a container image.
+
+| Field | Description |
+|-------|-------------|
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | Unique identifier for the package (format: version|name) |
```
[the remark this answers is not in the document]
achantavy: older cartography convention. not going to change unfortunately.

docs/root/modules/trivy/schema.md
```diff
@@ -0,0 +1,91 @@
...
+| installed_version | Version of the installed package |
+| name | Name of the package |
+| version | Version of the package (same as installed_version) |
```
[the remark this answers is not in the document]
achantavy: older cartography convention, not going to change this

docs/root/modules/trivy/schema.md
```diff
@@ -0,0 +1,91 @@
...
+    ```
+
+### TrivyFix
+Representation of a fix for a vulnerability.
+
+| Field | Description |
+|-------|-------------|
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | Unique identifier for the fix (format: version|name) |
```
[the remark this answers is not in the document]
achantavy: ditto

tests/data/trivy/trivy_sample.py
```diff
@@ -0,0 +1,1431 @@
+TRIVY_SAMPLE = {
```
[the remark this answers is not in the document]
heryxpc: I agree with this comment @achantavy, could you load it as a json file instead?
this is a nit, not super critical.

cartography/intel/trivy/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+def get_scan_targets(
+    neo4j_session: Session,
+    account_ids: list[str] | None = None,
+) -> list[tuple[str, str, str, str, str]]:
+    """
+    Return list of ECR images from all accounts in the graph as tuples with shape (region, image_tag, image_uri,
+    repo_name, image_digest).
+    """
+    if not account_ids:
+        aws_accounts = list_accounts(neo4j_session)
```
heryxpc, line 29: Why the default is scanning all AWS accounts and not the opposite?
I can see an scenario when you run this module without understanding the consequences of running scans in a very large number of images.

achantavy: To clarify, this module does not do any scanning -- it only loads scan result data to the graph.

Let me draw this out to make it clearer: 
Blank diagram - Page 2 (1)

To write Trivy nodes to the graph, we need scan results to be in S3, and we need ECR nodes to be in the graph. In the diagram above, red represents data that comes from a scan result stored in S3, blue respresents ECR nodes in the graph, and purple represents those two sources blending together.

`get_scan_targets()` simply lists all available images in the graph so that if a Trivy scan result exists for it in S3, then we can map the image to the scan result. 

So, if there is no scan result available for a given ECRImage, then no edge will be drawn.

achantavy: Oh, and going back to the example above, we see that because only one of those images exists in both the scan results bucket _and_ the graph, it is the only piece of data that gets drawn to the graph

cartography/intel/trivy/scanner.py
achantavy: Update: I think I can massively simplify this. The Trivy source data contains a `RepoDigests` field that matches up with the digest in an ECRImage. If I do this, I won't need to do this data wrangling and I won't need this additional dict.

achantavy: Fixed now!

cartography/cli.py
```diff
@@ -637,6 +637,24 @@ def _build_parser(self):
                 "Required if you are using the Anthropic intel module. Ignored otherwise."
             ),
         )
+        parser.add_argument(
```
jychp, line 640: I think we'll have more intel modules requiring entry files in the future, and we'll need to provide a more general way to handle that — but no need to overthink it for now.

cartography/sync.py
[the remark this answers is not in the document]
achantavy: @jychp - actually, what should I do here? I forgot about the dynamic behavior

cartography/intel/trivy/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+    Return list of ECR images from all accounts in the graph.
+    """
+    if not account_ids:
+        aws_accounts = list_accounts(neo4j_session)
+    else:
+        aws_accounts = account_ids
+
+    ecr_images: set[str] = set()
+    for account_id in aws_accounts:
+        for _, _, image_uri, _, _ in get_ecr_images(neo4j_session, account_id):
```
achantavy, line 35: Kind of an ugly workaround to use an existing client function without modifying it and breaking an interface. In future can replace with its own dedicated function

cartography/sync.py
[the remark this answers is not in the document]
jychp: Let's keep the current approach with explicit declarations for now.
The redundancy—introduced accidentally since we agreed not to use dynamic loading—will be removed in a separate PR.

cartography/models/trivy/package.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@dataclass(frozen=True)
+class TrivyPackageToImage(CartographyRelSchema):
+    target_node_label: str = "ECRImage"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("ImageDigest")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "DEPLOYED"
```
jychp, line 38: I generally prefer direct verbs (e.g., Image → deploys → Package), but that’s more a personal preference than an absolute rule.

cartography/intel/trivy/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+        images_in_graph: Set of ECR images in the graph
+        json_files: Set of S3 object keys for JSON files
+        trivy_s3_prefix: S3 prefix path containing scan results
+
+    Returns:
+        List of tuples (image_uri, s3_object_key)
+    """
+    intersection = []
+    prefix_len = len(trivy_s3_prefix)
+    for s3_object_key in json_files:
```
jychp, line 57: I think we can reduce complexity by using sets and built-in intersection operator on sets, but not important here due to the few amount of data

jychp (approved):
LGTM

cartography/models/trivy/package.py
achantavy: Yeah, will do going forward. Keeping it this way to not break other users of early versions of this code

cartography/intel/trivy/__init__.py
achantavy: Yeah, better to use the built-ins, though this already uses sets so the time complexity is optimized. Could further optimize by not appending a uri + key and just use a key instead

PR #1167, opened 2023-05-05
Add support for GitHub teams and team-repos

achantavy:
Testing performed

- Integration tests
- Manually ran CLI

### DIFF
--- a/cartography/intel/github/__init__.py
+++ b/cartography/intel/github/__init__.py
@@ -6,6 +6,7 @@ import neo4j
 from requests import exceptions
 
 import cartography.intel.github.repos
+import cartography.intel.github.teams
 import cartography.intel.github.users
 from cartography.config import Config
 from cartography.util import timeit
@@ -45,6 +46,14 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
                 auth_data['token'],
                 auth_data['url'],
                 auth_data['name'],
+            )
+            cartography.intel.github.teams.sync_github_teams(
+                neo4j_session,
+                common_job_parameters,
+                auth_data['token'],
+                auth_data['url'],
+                auth_data['name'],
+                config.update_tag,
             )
         except exceptions.RequestException as e:
             logger.error("Could not complete request to the GitHub API: %s", e)
--- a/cartography/intel/github/__init__.py
+++ b/cartography/intel/github/__init__.py
@@ -53,7 +53,6 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
                 auth_data['token'],
                 auth_data['url'],
                 auth_data['name'],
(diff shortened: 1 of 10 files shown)


resilience-jychp (commented):
Needed feature, thx for the job.
Some non blocking comments.

May be it can be confusing to have a module with both old way & new model queries ...

cartography/models/github/teams.py
```diff
@@ -0,0 +1,112 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class GitHubTeamNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef('url')
```
ramonpetgrave64, line 16: nonblocker: I feel like id should maybe be `{org}/{name}`, but i see that url was also used for the id of Repos

tests/integration/cartography/intel/github/test_teams.py
```diff
@@ -0,0 +1,114 @@
...
+TEST_JOB_PARAMS = {'UPDATE_TAG': TEST_UPDATE_TAG}
+TEST_GITHUB_URL = "https://fake.github.net/graphql/"
+FAKE_API_KEY = 'asdf'
+
+
+@patch.object(cartography.intel.github.teams, '_get_team_repos', return_value=GH_TEAM_REPOS)
+@patch.object(cartography.intel.github.teams, 'get_teams', return_value=GH_TEAM_DATA)
+def test_sync_github_teams(mock_teams, mock_team_repos, neo4j_session):
+    # Arrange
+    _ensure_local_neo4j_has_test_data(neo4j_session)
```
ramonpetgrave64, line 21: Non blocker: add one more test org to make sure teams aren't accidentally added to that other org.

cartography/intel/github/__init__.py
```diff
@@ -46,5 +47,13 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
                 auth_data['url'],
                 auth_data['name'],
             )
+            cartography.intel.github.teams.sync_github_teams(
+                neo4j_session,
+                common_job_parameters,
+                auth_data['token'],
+                auth_data['url'],
+                auth_data['name'],
+                config.update_tag,
```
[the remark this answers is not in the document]
ramonpetgrave64: I think we can take away this parameter instead, since the update tag is within common_job_parameters.

ramonpetgrave64 (changes requested):

ramonpetgrave64 (approved):

PR #1168, opened 2023-05-08
Add more detailed listeners, rules, actions to AWS ELBv2 load balancers.

alowde-ps:
Record rules, conditions and actions in the form (ELBV2Listener)-[ELBV2_LISTENER_RULE]-(ELBV2Action). This allows us to analyze the path of requests that are intelligently routed based on HTTP request parameters by AWS Application Load Balancers.

### DIFF
--- a/cartography/intel/aws/ec2/load_balancer_v2s.py
+++ b/cartography/intel/aws/ec2/load_balancer_v2s.py
@@ -22,7 +22,28 @@ def get_load_balancer_v2_listeners(client: botocore.client.BaseClient, load_bala
     for page in paginator.paginate(LoadBalancerArn=load_balancer_arn):
         listeners.extend(page['Listeners'])
 
+    # Technically listeners don't have names, but the AWS console titles them as 'protocol':'port' so we'll do the same.
+    for listener in listeners:
+        listener['name'] = listener['Protocol'] + ":" + str(listener['Port'])
     return listeners
+
+
+@timeit
+@aws_handle_regions
+def get_load_balancer_v2_rules(client: botocore.client.BaseClient, listener_arn: str) -> List[Dict]:
+    paginator = client.get_paginator('describe_rules')
+    rules: List[Dict] = []
+    for page in paginator.paginate(ListenerArn=listener_arn):
+        rules.extend(page['Rules'])
+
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/ec2/load_balancer_v2s.py
```diff
@@ -209,6 +365,146 @@ def load_load_balancer_v2_listeners(
...
+            result['source_ip'] = '|'.join(cnd['SourceIpConfig']['Values'])
+    return result
+
+
+def action_ingest_statement(type: str, additional: str) -> str:
+    statement = """
+    MATCH (rule:ELBV2Rule{id: $RuleId})-[ATTACHED_TO]->(listener:ELBV2Listener)
+    WITH listener, rule
+    UNWIND $Actions as action
+        MERGE (listener)-[rr:ELBV2_LISTENER_RULE]->(actionNode:ELBV2Action:""" + type + """{id: action.id})
```
achantavy, line 391: Rather than keep the same information in two places - in the ELBV2Rule _and_ this relationship - I'd rather just keep it in one place and prefer keeping the attributes in the ELBV2Rule node.

There is also an `ATTACHED_TO` relationship between `ELBV2Rule`s and `ELBV2Listeners`. I think having both `ATTACHED_TO` and `ELBV2_LISTENER_RULE` is a bit crowded so we should decide on keeping just one.

achantavy (changes requested):
This is a very cool PR and I'm looking forward to having this visibility. I mentioned this a while ago in a meeting but I'd like to have new plugins use the new data model. This way you get automatic cleanup jobs and it's less error prone to create new relationships.

Adopting the new data model means that the `transform()` functions will be doing more work to adjust the data into a workable shape for the querybuilder. One adjustment that'll need to happen is with things like this:

```cypher
MATCH (rule:ELBV2Rule{id: $RuleId})-[ATTACHED_TO]->(listener:ELBV2Listener)
... attach something to the listener ...
```

Here, we match on a Rule and then attach more data to the attached listener. To adapt to the new data model, we'd have to return an ID for the attached listener. This is a bit hard to explain in a comment but let me know when you get to this point and we can have a call. Let me know if you can't work on this and I'll see how far I can get adapting it. Might be easiest to do one new data type at a type.

Sorry for the slowness!

PR #1172, opened 2023-05-12
add duo intel module

ramonpetgrave64:
Adds a Duo intel module

- users, also linking to humans
- groups
- endpoints
- tokens
- webauthncredentials


cartography/intel/duo/users.py
```diff
@@ -0,0 +1,112 @@
...
+            'notes': user['notes'],
+            'phones': [dumps(phone) for phone in user['phones']],
+            'realname': user['realname'],
+            'status': user['status'],
+            'tokens': [dumps(token) for token in user['tokens']],
+            'u2ftokens': [dumps(u2ftoken) for u2ftoken in user['u2ftokens']],
+            'user_id': user['user_id'],
+            'username': user['username'],
+            'webauthncredentials': [
+                dumps(webauthncredential)
```
achantavy, line 72: Making sure I understand: this takes the full json and make it a string?

docs/root/modules/duo/config.md
```diff
@@ -0,0 +1,10 @@
+## Duo Configuration
+
+.. _duo_config:
+
+Follow these steps to analyze BigFix objects with Cartography.
```
achantavy, line 5:
```suggestion
Follow these steps to analyze Duo objects with Cartography.
```

achantavy (changes requested):

cartography/intel/duo/users.py
ramonpetgrave64: yes

achantavy: I guess this is ok for now. Eventually maybe it should be a node but I won't block on this.

docs/root/modules/duo/schema.md
```diff
@@ -0,0 +1,178 @@
...
+- A DuoUser has multiple DuoEndpoints
+
+    ```
+    (DuoUser)-[ENDPOINT_DUO]->(DuoEndpoint)
+    ```
+
+- A DuoUser is an identity to a Human
+
+    ```
+    (DuoUser)-[IDENTITY_DUO]->(Human)
```
achantavy, line 123: I think the convention should be for `:IDENTITY_DUO` to point _at_ the DuoUser to say that "This Human's Duo Identity is this -->"

docs/root/modules/duo/schema.md
```diff
@@ -0,0 +1,178 @@
...
+- An DuoApiHost contains DuoUsers
+
+    ```
+    (DuoApiHost)-[RESOURCE]->(DuoUser)
+    ```
+
+- A DuoUser is part of multiple DuoGroups.
+
+    ```
+    (DuoGroup)-[GROUP_DUO]->(DuoUser)
```
achantavy, line 111: If we keep the GROUP_DUO name, we should point toward the group.

Else we should rename the rel label to something like "MEMBER_OF_DUO_GROUP" since this currently is pointing at the DuoUser

cartography/intel/duo/users.py
```diff
@@ -0,0 +1,112 @@
...
+            'alias4': user['alias4'],
+            'created': user['created'],
+            'email': user['email'],
+            'firstname': user['firstname'],
+            'is_enrolled': user['is_enrolled'],
+            'last_directory_sync': user['last_directory_sync'],
+            'last_login': user['last_login'],
+            'lastname': user['lastname'],
+            'notes': user['notes'],
+            'phones': [dumps(phone) for phone in user['phones']],
```
achantavy, line 64: Can we exclude the phone number? I think phone number may be included in other places too.

We made a choice in the Okta sync to make sure a user phone number is not stored in the graph.

achantavy (changes requested):

docs/root/modules/duo/schema.md
ramonpetgrave64: renamed to MEMBER_OF_DUO_GROUP

docs/root/modules/duo/schema.md
ramonpetgrave64: fixed the comment

cartography/intel/duo/users.py
ramonpetgrave64: excluded

cartography/models/duo/user.py
```diff
@@ -81,7 +81,7 @@ class DuoGroupToDuoUserRel(CartographyRelSchema):
         {'group_id': PropertyRef('group_id')},
     )
     direction: LinkDirection = LinkDirection.INWARD
```
achantavy, line 83: We might want to point this the other direction. For example with MEMBER_OF_OKTA_GROUP we do `(:OktaUser)-[:MEMBER_OF_OKTA_GROUP]->(:OktaGroup)`. Right now this is pointing to the DuoUser and not the DuoGroup.

ramonpetgrave64: changed

cartography/intel/duo/users.py
ramonpetgrave64: Added. And changed the direction of MEMBER_OF_DUO_GORUP

achantavy (approved):
Looks good to me, thank you for working on this! As a future note let's please not make 3k line PRs a regular thing :p

PR #1175, opened 2023-05-18
Fix missing Instance/NetworkInterface/Groups error

heryxpc:


cartography/intel/aws/ec2/instances.py
```diff
@@ -65,85 +65,91 @@ def transform_ec2_instances(reservations: List[Dict[str, Any]], region: str, cur
...
+                    for group in instance["SecurityGroups"]:
+                        sg_list.append(
+                            {
+                                'GroupId': group['GroupId'],
+                                'InstanceId': instance_id,
+                            },
+                        )
+
+                for network_interface in instance['NetworkInterfaces']:
+                    for security_group in network_interface.get('Groups', []):
```
heryxpc, line 125: This is the actual change that avoids the error.
The rest is just indentation after the added try/except.

cartography/intel/aws/ec2/instances.py
```diff
@@ -65,85 +65,90 @@ def transform_ec2_instances(reservations: List[Dict[str, Any]], region: str, cur
...
+                        if 'VolumeId' in mapping['Ebs']:
+                            instance_ebs_volumes_list.append({
+                                'InstanceId': instance_id,
+                                'VolumeId': mapping['Ebs']['VolumeId'],
+                                'DeleteOnTermination': mapping['Ebs']['DeleteOnTermination'],
+                                # 'SnapshotId': mapping['Ebs']['SnapshotId'],  # TODO check on this
+                            })
+            except Exception as e:
+                logger.error(f"Failed to process instance {instance_id}. Error: {e}")
+                raise e
```
heryxpc, line 149: Added log to identify instance failing.

heryxpc:
👀 @ramonpetgrave64 @achantavy @lyft/vulnerability-management

cartography/intel/aws/ec2/instances.py
khanhldt: Generally not the best practice to add an extra log. You can wrap the exception by another one to clarify the error.

```
except Exception as e:
  raise Exception(f"Failed to process instance {instance_id}", e)
```

Sometimes it's also preferred to have a named exception but i think it might not be necessary in this case.

heryxpc: Yeah, I can wrap it to extend the data with the instance id.

serge-wq:
Beat me to it! :D

ramonpetgrave64 (approved):

heryxpc:
💨

PR #1186, opened 2023-06-14
Improve drift detection perf, tests, and add some typehints

achantavy:
Uses a set membership check instead of iterating through each item in the list of results to detect drift.

Adds unit test to ensure order of results does not matter in the drift detection check.

Also adds some typehints.

### DIFF
--- a/cartography/driftdetect/detect_deviations.py
+++ b/cartography/driftdetect/detect_deviations.py
@@ -3,16 +3,18 @@ import os
 
 from marshmallow import ValidationError
 
+from cartography.driftdetect.model import State
 from cartography.driftdetect.reporter import report_drift
 from cartography.driftdetect.serializers import ShortcutSchema
 from cartography.driftdetect.serializers import StateSchema
 from cartography.driftdetect.storage import FileSystem
 from cartography.driftdetect.util import valid_directory
+from cartography.driftdetect.config import GetDriftConfig
 
 logger = logging.getLogger(__name__)
 
 
-def run_drift_detection(config):
+def run_drift_detection(config: GetDriftConfig) -> None:
     try:
         if not valid_directory(config.query_directory):
             logger.error("Invalid Drift Detection Directory")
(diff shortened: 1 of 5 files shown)


cartography/driftdetect/get_states.py
```diff
@@ -90,7 +92,13 @@ def run_get_states(config):
                 logger.exception(err)
 
 
-def get_query_state(session, query_directory, state_serializer, storage, filename):
+def get_query_state(
+        session: neo4j.Session,
+        query_directory: str,
+        state_serializer: StateSchema,
+        storage,
```
achantavy, line 99: Not adding a typehint here because it seems like we're doing this ducktyping thing where we expect the `storage` object to have a write() function and other methods but I don't want to necessarily restrict it to _just_ the existing `storage.FileSystem` object.

cartography/driftdetect/detect_deviations.py
```diff
@@ -92,10 +96,12 @@ def compare_states(start_state, end_state):
     :return: list of tuples of differences between states in the form (dictionary, State object)
     """
     differences = []
+    # Use set for faster membership check
+    start_state_results = {tuple(res) for res in start_state.results}
```
achantavy, line 100: Main change of this PR. Spend more memory for faster lookup.

cartography/driftdetect/detect_deviations.py
```diff
@@ -92,10 +96,12 @@ def compare_states(start_state, end_state):
...
     for result in end_state.results:
-        if result in start_state.results:
+        if tuple(result) in start_state_results:
             continue
-        drift = []
+        drift: List[Union[str, List[str]]] = []
```
achantavy, line 104: I'm not a fan of how the list can have more than one type (and mypy isn't particularly happy about this either) but I don't want to break anything

ramonpetgrave64 (approved):

PR #1192, opened 2023-06-23
#1206: add async to aws:s3 and aws:ecr

ramonpetgrave64:
Add some helper functions for refactoring existing modules to be more async and use them for aws:s3 and aws:ecr.
In a test on one of our larger aws accounts, we've seen ~90% reduction in sync time for s3 and ecr.

- `to_async` wraps a regular synchronous function so that it can be used within async functions in a non-blocking fashion
- `so_sync` 
  1. takes a series are [Awaitables](https://docs.python.org/3.8/library/asyncio-task.html#asyncio-awaitables) (e.g., container objects returned when invoking async functions)
  2. schedules them all to be run simultaneously
  3. blocks until they are all finished
  4. returns all the results in the same order

There is a caveat when running these functions from a Jupyter notebook.
You must do a workaround, but homefully in future versions of asyncio this will not be needed
```
    # import nest_asyncio
    # nest_asyncio.apply()
```

### Testing
Existing unit tests mostly cover these changes, but I also did some manual testing against a real AWS account.


cartography/util.py
```diff
@@ -297,3 +301,31 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
...
+
+    NOTE: to use this in a Jupyter notebook, you need to do:
+    # import nest_asyncio
+    # nest_asyncio.apply()
+    '''
+    call = partial(func, *args, **kwargs)
+    return asyncio.get_event_loop().run_in_executor(None, call)
+
+
+def to_sync(*awaitables: Awaitable[Any]) -> Any:
```
achantavy, line 323:
```suggestion
def to_synchronous(*awaitables: Awaitable[Any]) -> Any:
```

cartography/intel/aws/ecr.py
```diff
@@ -148,9 +151,12 @@ def sync(
         logger.info("Syncing ECR for region '%s' in account '%s'.", region, current_aws_account_id)
         image_data = {}
         repositories = get_ecr_repositories(boto3_session, region)
-        for repo in repositories:
-            repo_image_obj = get_ecr_repository_images(boto3_session, region, repo['repositoryName'])
+
+        async def async_get_images(repo: Dict[str, Any]) -> None:
```
achantavy, line 155: For someone like me who does not know how `async` works, can you briefly describe what's going on here? E.g. how many calls to `get_ecr_repository_images` are happening in parallel, is it configurable, etc?

achantavy:
Also: I just played with this and it is indeed very slick 👍

cartography/util.py
```diff
@@ -297,3 +301,31 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
         items[i: i + size]
         for i in range(0, len(items), size)
     ]
+
+
+def to_async(func: Callable, *args: Any, **kwargs: Any) -> asyncio.Future:
```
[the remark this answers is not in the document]
ramonpetgrave64: changed

cartography/intel/aws/ecr.py
ramonpetgrave64: changed

cartography/intel/aws/ecr.py
```diff
@@ -139,6 +142,21 @@ def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
     run_cleanup_job('aws_import_ecr_cleanup.json', neo4j_session, common_job_parameters)
 
 
+def _get_image_data(boto3_session: boto3.session.Session, region: str, repositories: List[Dict]) -> Dict:
```
achantavy, line 145:
```suggestion
def _get_image_data(boto3_session: boto3.session.Session, region: str, repositories: List[Dict[str, Any]]) -> Dict[str, Any]:
```

while you're in here might as well add the extra chars

cartography/intel/aws/ecr.py
```diff
@@ -139,6 +142,21 @@ def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
...
+    '''
+    Given a list of repositories, get the image data for each repository,
+     return as a mapping from repositoryUri to image object
```
achantavy, line 148:
```suggestion
    return as a mapping from repositoryUri to image object
```

cartography/util.py
```diff
@@ -297,3 +302,47 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
...
+
+
+def to_asynchronous(func: Callable[..., R], *args: Any, **kwargs: Any) -> Awaitable[R]:
+    '''
+    Returns a Future that will run a function in the default threadpool.
+    Helper until we start using pytohn 3.9's asyncio.to_thread
+
+    Calls are also wrapped within a backoff decorator to handle throttling errors.
+
+    example:
```
achantavy, line 314: Good that you included an example, but since this is going to be a core function, I think we should have detailed docs on what each of the params is.

cartography/util.py
```diff
@@ -297,3 +302,47 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
         items[i: i + size]
         for i in range(0, len(items), size)
     ]
+
+
+def to_asynchronous(func: Callable[..., R], *args: Any, **kwargs: Any) -> Awaitable[R]:
+    '''
+    Returns a Future that will run a function in the default threadpool.
+    Helper until we start using pytohn 3.9's asyncio.to_thread
```
achantavy, line 310:
```suggestion
    Helper until we start using python 3.9's asyncio.to_thread
```

cartography/util.py
```diff
@@ -297,3 +302,47 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
...
+    # nest_asyncio.apply()
+    '''
+    CartographyThrottlingException = type('CartographyThrottlingException', (Exception,), {})
+    throttling_error_codes = ['LimitExceededException', 'Throttling']
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> R:
+        try:
+            return func(*args, **kwargs)
+        except botocore.exceptions.ClientError as error:
```
achantavy, line 329: This func is located in cartography.util. If we decide to use this for other sync jobs (non-AWS), will we add all of the different exception types here too, like will GCP exceptions be here too? Might end up having a lot of `excepts`.

cartography/util.py
```diff
@@ -297,3 +302,47 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
...
+    example:
+    future = to_async(my_func, my_arg, my_arg2)
+    to_sync(future)
+
+    NOTE: to use this in a Jupyter notebook, you need to do:
+    # import nest_asyncio
+    # nest_asyncio.apply()
+    '''
+    CartographyThrottlingException = type('CartographyThrottlingException', (Exception,), {})
+    throttling_error_codes = ['LimitExceededException', 'Throttling']
```
achantavy, line 323: This set of functions is located in cartography.util. Other libraries might use different exception names, so it might get unwieldy to have AWS, GCP, Azure names for throttling exceptions all in one place. I wonder if it might make sense to have this in cartography.aws.util (or whatever the equivalent is) first until we have an example of doing this with another module.

cartography/util.py
```diff
@@ -297,3 +302,47 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
...
+    call = partial(wrapped, *args, **kwargs)
+    return asyncio.get_event_loop().run_in_executor(None, call)
+
+
+def to_synchronous(*awaitables: Awaitable[Any]) -> Any:
+    '''
+    Waits for the Awaitable(s) to complete and returns their result(s).
+    See https://docs.python.org/3.8/library/asyncio-task.html#asyncio-awaitables
+
+    example:
```
achantavy, line 345: Same thing about explicitly documenting the parameters.

achantavy (changes requested):

cartography/util.py
ramonpetgrave64: thanks

cartography/util.py
ramonpetgrave64: a

ramonpetgrave64: added

cartography/util.py
ramonpetgrave64: added

cartography/util.py
```diff
@@ -297,3 +302,87 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
...
+    to_synchronous(future)
+
+    NOTE: to use this in a Jupyter notebook, you need to do:
+    # import nest_asyncio
+    # nest_asyncio.apply()
+    '''
+    CartographyThrottlingException = type('CartographyThrottlingException', (Exception,), {})
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> R:
```
achantavy, line 353: [nonblock][question]  What's the typevar R do?

ramonpetgrave64: It binds the return type of `func` so that mypy can understand that the wrapped function is supposed to still return type `R`.

cartography/util.py
ramonpetgrave64: added as the normal :param : spec

achantavy (approved):
Awesome thank you!

PR #1196, opened 2023-07-05
Update usage docs and roadmap links

achantavy:
Adds docs on 
- discovering what other nodes are connected to a given node type
- discovering what node properties are present on a given node type

Updates docs on
- roadmap link
- making syncmetadata docs more discoverable


resilience-jychp (approved):
Very useful update. Thx.

PR #1204, opened 2023-07-08
Set to None the gsuite token instead of github one

heryxpc:


resilience-jychp (approved):
Legit. Fix a bug introduced by myself.

Issue #1210, opened 2023-07-14
Schema: node attribute updates from multiple intel modules do not work


achantavy:
*Description*:
In #1154 I tried to refactor EBSVolumes to the new data model but ran into the following problem:

**Expected behavior**:
After refactoring the EBS Volume sync to the new model, I ran the EC2 instance sync and the EBS volume sync one after the other and see the `deleteontermination` field on the EBSVolumes that are attached to EC2 instances.

**Actual behavior**:
After refactoring, when I run the EC2 instance sync and the EBS volume sync one after the other, the `deleteontermination` field is not present.

**Root cause**:
The EC2 instance sync loads in :EC2Instances with :EBSVolumes attached based on the output of `describe-ec2-instances`. This data includes a field called `deleteontermination` on the :EBSVolume. `deleteontermination` is only set during the EC2 instance sync.

Next when the EBS Volume sync runs, it enriches existing :EBSVolumes with additional fields from data retrieved with `describe-ebs-volumes` but because this API call does not include `deleteontermination`, we will overwrite the previous values of `deleteontermination` with `None`, thus removing the property from all :EBSVolumes.

#1154 tries to solve this by updating the `PropertyRef` implementation to use `COALESCE()` so that any existing values already defined for a given node property are preserved. This is very quirky and complicated and _probably_ going to introduce lots of new problems. Instead, I'd like to explore doing this with separate classes for properties derived from different API responses.

This needs to be solved in order for the new data model to work.

PR #1214, opened 2023-07-14
#1210: EBSVolume => new data model, Allow node attr updates from multiple intel modules

achantavy:
See #1210 for full context.

#1154 tried to solve this problem by updating the querybuilder but this was too complex and would not generalize well.

This solution is simpler where we use different property classes for each API response so that we don't overwrite properties on a node set by another sync job.

This PR can be reviewed commit-by-commit:
- c0d9ac4cc35fd93e260964cfa41a813e2a2032a7 shows a repro of the error with a failing integration test.
- facb63bcbac6b68eec0fd2ea3f6b0550ac40eb10 shows the solution using multiple classes.


cartography/models/aws/ec2/volumes.py
```diff
@@ -61,3 +77,32 @@ class EBSVolumeSchema(CartographyNodeSchema):
             EBSVolumeToEC2Instance(),
         ],
     )
+
+
+@dataclass(frozen=True)
+class EBSVolumeInstanceProperties(CartographyNodeProperties):
```
ramonpetgrave64, line 83: Looking at the sample `DESCRIBE_INSTANCES`, the only properties about Volumes that are returned are just `volumeid` and `deleteontermination`, right?
https://github.com/lyft/cartography/blob/facb63bcbac6b68eec0fd2ea3f6b0550ac40eb10/tests/data/aws/ec2/instances.py#L390-L395

tests/integration/cartography/intel/aws/ec2/test_ec2_volumes.py
```diff
@@ -3,123 +3,147 @@
...
-    )
 
-    # Act: Load Test Volumes
-    data = tests.data.aws.ec2.volumes.DESCRIBE_VOLUMES
-    transformed_data = cartography.intel.aws.ec2.volumes.transform_volumes(data, TEST_REGION, TEST_ACCOUNT_ID)
+@patch.object(cartography.intel.aws.ec2.instances, 'get_ec2_instances', return_value=DESCRIBE_INSTANCES['Reservations'])
+@patch.object(cartography.intel.aws.ec2.volumes, 'get_volumes', return_value=DESCRIBE_VOLUMES)
+def test_sync_ebs_volumes_e2e(mock_get_vols, mock_get_instances, neo4j_session):
+    # Arrange
+    neo4j_session.run('MATCH (n) DETACH DELETE n;')
```
ramonpetgrave64, line 60: We shouldn't need this line, because the fixture should automatically do this.

tests/integration/cartography/intel/aws/ec2/test_ec2_volumes.py
```diff
@@ -3,123 +3,147 @@
...
         TEST_ACCOUNT_ID,
         TEST_UPDATE_TAG,
+        {'UPDATE_TAG': TEST_UPDATE_TAG, 'AWS_ID': TEST_ACCOUNT_ID},
     )
 
-    # Assert
-    expected = {
-        (TEST_ACCOUNT_ID, 'v-01'),
-        (TEST_ACCOUNT_ID, 'v-02'),
+    # Assert that deleteontermination is set by sync_ec2_instances
```
ramonpetgrave64, line 74:
```suggestion
    # Assert that deleteontermination is set by sync_ec2_instances. The encrypted property isn't returned by this API.
```

tests/integration/cartography/intel/aws/ec2/test_ec2_volumes.py
```diff
@@ -3,123 +3,147 @@
...
+        ('i-01', 'vol-0df'),
+        ('i-02', 'vol-03'),
+        ('i-03', 'vol-09'),
+        ('i-04', 'vol-04'),
     }
-    actual = {
-        (r['n1.id'], r['n2.id']) for r in result
+
+    # Assert that the account to volume rels exist
+    assert check_rels(
```
ramonpetgrave64, line 136: We should maybe do this check after syncing instances, but before syncing the volumes

ramonpetgrave64 (dismissed):
Looks great

tests/integration/cartography/intel/aws/ec2/test_ec2_volumes.py
achantavy: https://github.com/lyft/cartography/blob/48f50caed6ac889225e94611b1315f6a95c19470/tests/integration/conftest.py#L12

fixture scope is set to module so it doesn't actually haha. If we change it to 'function' then it'll break all the other tests

ramonpetgrave64: Alright

cartography/models/aws/ec2/volumes.py
achantavy: Good point, will update

ramonpetgrave64 (approved):

cartography/models/aws/ec2/volumes.py
```diff
@@ -61,3 +77,31 @@ class EBSVolumeSchema(CartographyNodeSchema):
...
+    """
+    EBS Volume properties as known by an EC2 instance.
+    The EC2 instance API response includes a `deleteontermination` field and the volume id.
+    """
+    arn: PropertyRef = PropertyRef('Arn', extra_index=True)
+    id: PropertyRef = PropertyRef('VolumeId')
+    volumeid: PropertyRef = PropertyRef('VolumeId', extra_index=True)
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+    deleteontermination: PropertyRef = PropertyRef('DeleteOnTermination')
```
achantavy, line 92: @serge-wq [comment 1 of 2] - ah here we go. Here an EBS volume as known by an EC2 instance has a `deleteontermination` field...

cartography/models/aws/ec2/volumes.py
```diff
@@ -13,10 +13,23 @@
...
     region: PropertyRef = PropertyRef('Region', set_in_kwargs=True)
     lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
-    deleteontermination: PropertyRef = PropertyRef('DeleteOnTermination')
+    availabilityzone: PropertyRef = PropertyRef('AvailabilityZone')
+    createtime: PropertyRef = PropertyRef('CreateTime')
+    encrypted: PropertyRef = PropertyRef('Encrypted')
+    size: PropertyRef = PropertyRef('Size')
+    state: PropertyRef = PropertyRef('State')
+    outpostarn: PropertyRef = PropertyRef('OutpostArn')
+    snapshotid: PropertyRef = PropertyRef('SnapshotId')
```
achantavy, line 27: @serge-wq [comment 2 of 2] - .. but the EBSVolume by itself doesn't have that field. It also has fields that are not known by the EBSVolumeInstance.

The description of this PR shows more details on this decision.

Issue #1216, opened 2023-07-14
Refactor AWS SSM to new data model


achantavy:
Task: refactor the AWS SSM module to the new data model.

The SSMInstanceInformation node connects to EC2Instance nodes via instance ID and not ARN. Per #1024, we want to standardize on ARN and not instance ID. Before we do that though, we should make sure all data types involved are using the cartography data model.

PR #1217, opened 2023-07-15
#1216: AWS SSM->new data model

achantavy:
Refactors AWS SSM module to use cartography's data model.


cartography/data/indexes.cypher
```diff
@@ -321,10 +321,6 @@ CREATE INDEX IF NOT EXISTS FOR (n:SpotlightVulnerability) ON (n.host_info_local_
 CREATE INDEX IF NOT EXISTS FOR (n:SpotlightVulnerability) ON (n.lastupdated);
 CREATE INDEX IF NOT EXISTS FOR (n:SQSQueue) ON (n.id);
 CREATE INDEX IF NOT EXISTS FOR (n:SQSQueue) ON (n.lastupdated);
-CREATE INDEX IF NOT EXISTS FOR (n:SSMInstanceInformation) ON (n.id);
```
achantavy, line 324: No longer need this in the file since these are automatically handled

cartography/data/jobs/cleanup/aws_import_ssm_cleanup.json
```diff
@@ -1,25 +0,0 @@
-{
-  "statements": [
```
achantavy, line 2: No longer need these since they're automatically handled

cartography/intel/aws/ssm.py
```diff
@@ -65,6 +81,16 @@ def get_instance_patches(
     return instance_patches
 
 
+def transform_instance_patches(data_list: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    for p in data_list:
+        p["Id"] = f"{p['_instance_id']}-{p['Title']}"
```
ramonpetgrave64, line 86: The title can have hyphen and spaces. I think we should use the kbid instead.
```suggestion
        p["Id"] = f"{p['_instance_id']}/{p['KBId']}"
```

ramonpetgrave64: Non-blocker: we could move the somewhat out-of-band transformation on line 79 to a loop within `sync()`.

- patch["_instance_id"] = instance_id

to something like 
```
...
for instance_id in instance_ids:
  data = transform_instance_patches(instance_id, data)
  load_instance_patches(neo4j_session, data, region, current_aws_account_id, update_tag)
...
```

ramonpetgrave64 (dismissed):
We may want to change the way id is composed, but otherwise it's good.

ryan-lane (approved):
LGTM. I really love the new schema-based approach for loading/transforming data!

cartography/intel/aws/ssm.py
achantavy: The old way of forming an ID used this way. Let's keep it the same so that this is a refactor that doesn't change something important like an id.

achantavy: For the out-of-band thing, good idea, we can include that later on

ramonpetgrave64 (approved):

Issue #1218, opened 2023-07-15
Refactor AWS EC2 network interface module to new data model


achantavy:
Makes #1024 easier.

We'd like to join on an EC2 instance using ARN instead of the instanceid, and many different node types are attached to EC2 instances including network interfaces.

PR #1219, opened 2023-07-17
refactor: #1218, fix: #1152

### LINKED ISSUE #1152: EC2PrivateIP, NetworkInterface, and EBSVolume cleanup jobs are leaving behind orphan nodes
*Description*:

The EC2PrivateIp, NetworkInterface, and EBSVolume cleanup job is leaving behind orphaned nodes. I have observed this on other jobs too and will be filing separate issues for each of them.

*To Reproduce*:

Run the EC2 syncs over several days. Run
```cypher
match(n:EC2PrivateIp) where not (n)--() return distinct n.lastupdated, datetime({epochSeconds: n.lastupdated}), count(n) order by n.lastupdated 
```
Expectation:
We get back 0 orphaned records.

Actual:
We get a large number of orphaned records on each day.

0.76.0

achantavy:
Refactors EC2 Network Interfaces and Private IPs to the cartography data model. Fixes #1152 since the data model includes automatic cleanup jobs.


cartography/data/indexes.cypher
```diff
@@ -93,8 +93,6 @@ CREATE INDEX IF NOT EXISTS FOR (n:DOProject) ON (n.lastupdated);
 CREATE INDEX IF NOT EXISTS FOR (n:EBSSnapshot) ON (n.id);
 CREATE INDEX IF NOT EXISTS FOR (n:EBSSnapshot) ON (n.lastupdated);
 CREATE INDEX IF NOT EXISTS FOR (n:EC2KeyPair) ON (n.keyfingerprint);
-CREATE INDEX IF NOT EXISTS FOR (n:EC2PrivateIp) ON (n.id);
-CREATE INDEX IF NOT EXISTS FOR (n:EC2PrivateIp) ON (n.lastupdated);
```
ramonpetgrave64, line 97: Looks like we never had indexes for EC2NetworkInterface

cartography/models/aws/ec2/networkinterface_instance.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Selection of properties of a network interface as known by an EC2 instance
+    """
+    # arn: PropertyRef = PropertyRef('Arn', extra_index=True) TODO use arn; issue #1024
+    id: PropertyRef = PropertyRef('NetworkInterfaceId')
+    status: PropertyRef = PropertyRef('Status')
+    mac_address: PropertyRef = PropertyRef('MacAddress')
+    description: PropertyRef = PropertyRef('Description')
+    private_dns_name: PropertyRef = PropertyRef('PrivateDnsName')
+    private_ip_address: PropertyRef = PropertyRef('PrivateIpAddress')
```
ramonpetgrave64, line 25: But I think you meant to remove this `private_ip_address` form this particular schema?

cartography/models/aws/ec2/subnet_networkinterface.py
```diff
@@ -0,0 +1,87 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class EC2SubnetNetworkInterfaceNodeProperties(CartographyNodeProperties):
```
ramonpetgrave64, line 17: Since many of the properties of all these classes will be shared. I think it better to create and use base classes.

ramonpetgrave64 (changes requested):

cartography/models/aws/ec2/networkinterface_instance.py
achantavy: What you mean?

`private_ip_address` is present as a field on the node here: https://github.com/lyft/cartography/blob/830b8944879a01f52b21ee12b6fddf245f9733cb/cartography/intel/aws/ec2/network_interfaces.py#L51 so I'm making sure it's present in the refactor.

Should I remove it?

ramonpetgrave64: From the PR description I thought you meant that the property only comes from `describe_network_interfaces`

cartography/models/aws/ec2/subnet_networkinterface.py
achantavy: Eh, not a fan. I created a discussion here: https://github.com/lyft/cartography/discussions/1238

Let's continue the conversation there since I think this is something worth fleshing out in detail.

cartography/intel/aws/ec2/network_interfaces.py
```diff
@@ -27,256 +41,215 @@ def get_network_interface_data(boto3_session: boto3.session.Session, region: str
...
+                'RequesterManaged': network_interface['RequesterManaged'],
+                'SourceDestCheck': network_interface['SourceDestCheck'],
+                'Status': network_interface['Status'],
+                'SubnetId': network_interface['SubnetId'],
+                'ElbV1Id': elb_v1_id,
+                'ElbV2Id': elb_v2_id,
+            },
+        )
+        if network_interface.get('PrivateIpAddresses'):
+            for private_ip_address in network_interface['PrivateIpAddresses']:
```
ramonpetgrave64, line 85: Non blocker: chane this to 
```suggestion
       for private_ip_address in network_interface.get('PrivateIpAddresses', []):
```

ramonpetgrave64 (dismissed):

cartography/models/aws/ec2/networkinterfaces.py
```diff
@@ -1,5 +1,9 @@
 from dataclasses import dataclass
 
+from cartography.models.aws.ec2.networkinterface_instance import EC2NetworkInterfaceToAWSAccount
+from cartography.models.aws.ec2.networkinterface_instance import EC2NetworkInterfaceToEC2Instance
+from cartography.models.aws.ec2.networkinterface_instance import EC2NetworkInterfaceToEC2SecurityGroup
+from cartography.models.aws.ec2.networkinterface_instance import EC2NetworkInterfaceToEC2Subnet
```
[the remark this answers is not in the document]
achantavy: Do you think that I should redefine these rel classes rather than import them?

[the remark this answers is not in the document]
ramonpetgrave64: I think we can leave it for now, but if the API responses change and refactoring becomes problematic, we should then redefine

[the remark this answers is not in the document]
achantavy: Yeah, makes sense. I think it's ok to continue to import the relationship schemas since I don't think it's super likely for the ID fields that they rely on to change. We'll revisit if this ends up being a problem.

ramonpetgrave64 (approved):

cartography/models/aws/ec2/networkinterface_instance.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Selection of properties of a network interface as known by an EC2 instance
+    """
+    # arn: PropertyRef = PropertyRef('Arn', extra_index=True) TODO use arn; issue #1024
+    id: PropertyRef = PropertyRef('NetworkInterfaceId')
+    status: PropertyRef = PropertyRef('Status')
+    mac_address: PropertyRef = PropertyRef('MacAddress', extra_index=True)
+    description: PropertyRef = PropertyRef('Description')
+    private_dns_name: PropertyRef = PropertyRef('PrivateDnsName', extra_index=True)
+    private_ip_address: PropertyRef = PropertyRef('PrivateIpAddress', extra_index=True)
```
achantavy, line 25: @serge-wq - [comment 1 of 2] here's an example of what we were talking about earlier: a network interface as known by an EC2 instance has a private_ip_address...

achantavy: actually never mind, wait let me find a better example

PR #1224, opened 2023-07-19
Adds support to query Semgrep API to ingest SCA vulns

heryxpc:
Adds a new Schema and Intel Job to query Semgrep Enterprise API and ingest Semgrep Supply Chain (SSC) findings.
The schema connects a Semgrep Deployment with an id specific to a customer in Semgrep Enterprise to an SCA finding and location as a sub resource relationships. It also connects to a Github repository to match findings against where they were found. Each finding can have a location with the specific lines of code where the vulnerable dependency is being used.
SemgrepCartographyfinal


cartography/cli.py
```diff
@@ -500,6 +500,15 @@ def _build_parser(self):
                 'The Duo api hostname'
             ),
         )
+        parser.add_argument(
+            '--semgrep-env-var',
```
ramonpetgrave64, line 504: Rename to semgrep-app-token-env-var

cartography/intel/semgrep/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+import neo4j
+
+from cartography.config import Config
+from cartography.intel.semgrep.findings import sync
+from cartography.stats import get_stats_client
+from cartography.util import timeit
+
+
+logger = logging.getLogger(__name__)
+stat_handler = get_stats_client(__name__)
```
ramonpetgrave64, line 12: unused variable

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+from cartography.models.semgrep.findings import SemgrepDeploymentSchema
+from cartography.models.semgrep.findings import SemgrepSCAFindingSchema
+from cartography.models.semgrep.findings import SemgrepSCALocationSchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_deployment(semgrep_app_token: str) -> List[Dict[str, Any]]:
```
ramonpetgrave64, line 20: Since the deployment will only be a single object, I think we should refactor to refactor to not return a list.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    deployment = {}
+    deployment_url = "https://semgrep.dev/api/v1/deployments"
+    headers = {
+        "Content-Type": "application/json",
+        "Authorization": f"Bearer {semgrep_app_token}",
+    }
+    response = requests.get(deployment_url, headers=headers)
+    response.raise_for_status()
+    try:
```
ramonpetgrave64, line 33: I would remove all of the try/except branches. If anything fails, they will throw their own exceptions.

cartography/models/semgrep/findings.py
```diff
@@ -0,0 +1,129 @@
...
+    dependency_file_url: PropertyRef = PropertyRef('dependencyFileLocation_url')
+    scan_time: PropertyRef = PropertyRef('openedAt')
+
+
+@dataclass(frozen=True)
+class SemgrepDeploymentProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef('id')
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+    name: PropertyRef = PropertyRef('name')
+    slug: PropertyRef = PropertyRef('slug')
```
ramonpetgrave64, line 41: Add extra indexes

cartography/models/semgrep/findings.py
```diff
@@ -0,0 +1,129 @@
...
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'fullname': PropertyRef('repositoryName')},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "FOUND_IN"
+    properties: SemgrepSCAFindingToGithubRepoRelProperties = SemgrepSCAFindingToGithubRepoRelProperties()
+
+
+@dataclass(frozen=True)
+class SemgrepDeploymentSchema(CartographyNodeSchema):
```
ramonpetgrave64, line 79: I recommend putting each node typer into its own file

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+        params = {}
+        if cursor:
+            params = {"cursor": cursor}
+        try:
+            response = requests.get(sca_url, params=params, headers=headers)
+            response.raise_for_status()
+            if prev_request == response.request.url:
+                logger.warning(
+                    "Duplicate request detected. Breaking the loop to avoid infinite requests.",
+                )
```
ramonpetgrave64, line 79: I'm guessing this check was needed during initial development, but the situation shouldn't happen anymore with the way you've structured the loop?

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+    headers = {
+        "Content-Type": "application/json",
+        "Authorization": f"Bearer {semgrep_app_token}",
+    }
+
+    while has_more:
+        params = {}
+        if cursor:
+            params = {"cursor": cursor}
+        try:
```
ramonpetgrave64, line 73: No need for the try/excepts anymore

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+        sca_vuln["reachableIf"] = vuln["advisory"]["reachableIf"]
+        sca_vuln["exposureType"] = vuln["exposureType"]
+        dependency = f"{vuln['matchedDependency']['name']}|{vuln['matchedDependency']['versionSpecifier']}"
+        sca_vuln["matchedDependency"] = dependency
+        sca_vuln["dependencyFileLocation_path"] = vuln["dependencyFileLocation"]["path"]
+        sca_vuln["dependencyFileLocation_url"] = vuln["dependencyFileLocation"]["url"]
+        # Optional fields
+        cves = vuln.get("advisory", {}).get("references", {}).get("cveIds")
+        if len(cves) > 0:
+            # Take the first CVE
```
ramonpetgrave64, line 126: Why only the first cve?

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+            sca_vuln["cveId"] = vuln["advisory"]["references"]["cveIds"][0]
+        if vuln.get('closestSafeDependency'):
+            dep_fix = f"{vuln['closestSafeDependency']['name']}|{vuln['closestSafeDependency']['versionSpecifier']}"
+        else:
+            dep_fix = None
+        sca_vuln["closestSafeDependency"] = dep_fix
+        ref_urls = vuln["advisory"].get("references", {}).get("urls", [])
+        if ref_urls:
+            sca_vuln["ref_urls"] = ",".join(ref_urls)
```
ramonpetgrave64, line 135: non-blocker: this node property's value can also be an array.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+) -> None:
+    logger.info(f"Loading {len(vulns)} Semgrep SCA vulns info into the graph.")
+    load(
+        neo4j_session,
+        SemgrepSCAFindingSchema(),
+        vulns,
+        lastupdated=update_tag,
+        DEPLOYMENT_ID=deployment_id,
+    )
+    for vuln in vulns:
```
ramonpetgrave64, line 181: For better performance, all the usages should be extracted out first, so they can all be loaded once.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+            dep_fix = f"{vuln['closestSafeDependency']['name']}|{vuln['closestSafeDependency']['versionSpecifier']}"
+        else:
+            dep_fix = None
+        sca_vuln["closestSafeDependency"] = dep_fix
+        ref_urls = vuln["advisory"].get("references", {}).get("urls", [])
+        if ref_urls:
+            sca_vuln["ref_urls"] = ",".join(ref_urls)
+        sca_vuln["openedAt"] = vuln.get("openedAt", None)
+        usages_list = []
+        for usage in vuln.get("usages", []):
```
ramonpetgrave64, line 138: I think it better to have all the usages in their own separate list.
Each usage should also have the additional property sca_buld_id, so we can form the relationship.
This function will then return the two lists: vulns, usages

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,217 @@
...
+                SCA_ID=vuln["id"],
+            )
+
+
+@timeit
+def cleanup(
+    neo4j_session: neo4j.Session, common_job_parameters: Dict[str, Any],
+) -> None:
+    logger.info("Running Semgrep SCA findings cleanup job.")
+    cleanup_job = GraphJob.from_node_schema(
```
ramonpetgrave64, line 197: Also run cleanups for the SemgrepSCALocationSchema

tests/unit/cartography/intel/semgrep/test_findings.py
```diff
@@ -0,0 +1,168 @@
...
+        assert e.response.status_code == 500
+
+
+def test_transform_sca_vulns():
+    # Arrange
+    raw_vulns = SCA_RESPONSE["vulns"]
+    # Act
+    vulns = transform_sca_vulns(raw_vulns)
+    # Assert
+    assert vulns[0]["id"] == "yourorg/yourrepo|ssc-92af1d99-4fb3-4d4e-a9f4-d57572cd6590"
```
ramonpetgrave64, line 135: would be maybe easier to do an equality check against a rendered dict

ramonpetgrave64 (changes requested):

cartography/intel/semgrep/findings.py
heryxpc: Can I at least leave the `RequestException` case? I would like to see the 400's and 500's at the logs early rather than wait until it breaks all the sync.

cartography/intel/semgrep/findings.py
heryxpc: It was a safe guard, as nothing avoids to call `https://semgrep.dev/api/sca/deployments/{deployment_id}/vulns?cursor=lastcursor`.
Nevertheless, tested retrieving all the vulns without it and it worked. Removing it.

cartography/intel/semgrep/findings.py
heryxpc: From the tests I ran, it returns either 0 or 1 CVE that array.
Having a single CVE is useful because we want to match other scanner results 1 to 1.
In the future maybe we can use the CVE module https://github.com/lyft/cartography/tree/master/cartography/intel/cve, but that would lead to more complex relationships in order to connect a CVE from other scanner to each Semgrep finding.

cartography/intel/semgrep/findings.py
heryxpc: Does the [querybuilder](https://github.com/lyft/cartography/blob/f657a0f73c0cc449a27d8a490f2b86d66d31951d/cartography/graph/querybuilder.py) supports passing arrays?

cartography/intel/semgrep/findings.py
heryxpc: I saw your comment below regarding perfomance.
The number of usages is very small, but if that fits better the load all the same resource in batch paradigm I can change it.

cartography/models/semgrep/findings.py
heryxpc: I don't think we need to index deployment based on name and Slug, since it will be a single node, but let me know if there is something besides performance that I'm not seeing.

cartography/models/semgrep/findings.py
heryxpc: Done

cartography/intel/semgrep/findings.py
heryxpc: Moved it to it's own load function.

cartography/intel/semgrep/findings.py
heryxpc: Good callout.

tests/unit/cartography/intel/semgrep/test_findings.py
heryxpc: Updated it.

cartography/intel/semgrep/findings.py
heryxpc: Hey @ramonpetgrave64, after adding this check it breaks the integration tests. I cannot pass a single id to the `common_job_parameters` since the id changes per sub-resource node. Can I skip it?

heryxpc: Fixed it with the new relationship connecting deployment and locations.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,231 @@
...
+
+@timeit
+def sync(
+    neo4j_sesion: neo4j.Session,
+    semgrep_app_token: str,
+    update_tag: int,
+    common_job_parameters: Dict[str, Any],
+) -> None:
+    logger.info("Running Semgrep SCA findings sync job.")
+    semgrep_deployment = get_deployment(semgrep_app_token)
```
achantavy, line 216: Nit and non-blocker: consider splitting to sync_sca_vulns and sync_sca_usages to make it cleaner.

achantavy (changes requested):
Leaving suggestions on how to button this up. Looks good overall!

Can you please also add schema docs, config docs, and the Readme so that users know how to use it, how to set it up, and that we have Semgrep coverage?

cartography/intel/semgrep/findings.py
heryxpc: Wow, first time that I will see list as node properties, will give it a try!

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    Transforms the raw SCA vulns response from Semgrep API into a list of dicts
+    that can be used to create the SemgrepSCAFinding nodes.
+    """
+    vulns = []
+    usages = []
+    for vuln in raw_vulns:
+        sca_vuln: Dict[str, Any] = {}
+        # Mandatory fields
+        unique_id = f"{vuln['repositoryName']}|{vuln['advisory']['ruleId']}"
```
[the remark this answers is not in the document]
heryxpc: An SCA vuln is a finding specific to a repo.

I was thinking to use the `groupKey` (see [1](https://github.com/lyft/vulnmgmtinternal/blob/26cb73f61ecf45f9d7fa88286a692926e1c51107/semgrep_api/sample-response/sca-deployments-vulns.json#L71), [2](https://github.com/lyft/vulnmgmtinternal/blob/26cb73f61ecf45f9d7fa88286a692926e1c51107/semgrep_api/sample-response/sca-deployments-vulns.json#L152) and [3](https://github.com/lyft/vulnmgmtinternal/blob/26cb73f61ecf45f9d7fa88286a692926e1c51107/semgrep_api/sample-response/sca-deployments-vulns.json#L233) ). It does a similar thing, identifies a repository via an internal [repository id](https://github.com/lyft/vulnmgmtinternal/blob/26cb73f61ecf45f9d7fa88286a692926e1c51107/semgrep_api/sample-response/sca-deployments-vulns.json#L52), a [rule id](https://github.com/lyft/vulnmgmtinternal/blob/26cb73f61ecf45f9d7fa88286a692926e1c51107/semgrep_api/sample-response/sca-deployments-vulns.json#L6) which can apply to multiple repositories and a [exposure/reachability](https://github.com/lyft/vulnmgmtinternal/blob/26cb73f61ecf45f9d7fa88286a692926e1c51107/semgrep_api/sample-response/sca-deployments-vulns.json#L51) status.

I used the repositoryName to make it simple to identify what repository had the issue and the rule id to make it easier to search against the rules definition that Semgrep provides.
Does that makes sense?

cartography/intel/semgrep/findings.py
heryxpc: This endpoint is not yet documented on their public docs: https://semgrep.dev/api/v1/docs/#section/Introduction. I assume they made it an array for future cases.

The way we intend to use this is 1 SCA finding = 1 vuln at most (there are cases where there is no CVE).
Adding multiple CVEs to each finding will then cause us to search
```
MATCH (other:OtherCVEScanner)
MATCH (sca:SCAFinding)-[:FINDING_TO_CVE]-(cve:SCACVE)
WHERE other.id = cve.id
```
Considering there aren't cases yet where `FINDING_TO_CVE` would be more than 1, I thing we can save the complexity of adding an extra relationship.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,229 @@
...
+    deployment = {}
+    deployment_url = "https://semgrep.dev/api/v1/deployments"
+    headers = {
+        "Content-Type": "application/json",
+        "Authorization": f"Bearer {semgrep_app_token}",
+    }
+    response = requests.get(deployment_url, headers=headers)
+    response.raise_for_status()
+    try:
+        if response.status_code == 200:
```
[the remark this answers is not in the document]
heryxpc: Removed.

cartography/intel/semgrep/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    neo4j_session: neo4j.Session, config: Config,
+) -> None:
+    common_job_parameters = {
+        "UPDATE_TAG": config.update_tag,
+    }
+    if not config.semgrep_app_token:
+        logger.info('Semgrep import is not configured - skipping this module. See docs to configure.')
+        return
+    if not config.github_config:
+        logger.info('GitHub import is not configured - No relationship with GitHub repositories will be created.')
```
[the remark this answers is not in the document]
heryxpc: Done.

tests/unit/cartography/intel/semgrep/test_findings.py
```diff
@@ -0,0 +1,168 @@
...
+    try:
+        get_sca_vulns("your_semgrep_app_token", "yourorgid")
+    except RequestException as e:
+        # Assert
+        assert e.args[0] == "Server Error"
+        assert e.request == mock_requests
+        assert e.response.status_code == 500
+
+
+def test_transform_sca_vulns():
```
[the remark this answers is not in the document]
heryxpc: All this test is gone.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,229 @@
...
+    Gets the deployment associated with the passed Semgrep App token.
+    param: semgrep_app_token: The Semgrep App token to use for authentication.
+    """
+    deployment = {}
+    deployment_url = "https://semgrep.dev/api/v1/deployments"
+    headers = {
+        "Content-Type": "application/json",
+        "Authorization": f"Bearer {semgrep_app_token}",
+    }
+    response = requests.get(deployment_url, headers=headers)
```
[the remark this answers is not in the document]
heryxpc: Added same timeout as GitHub API.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,229 @@
...
+            deployment["id"] = data["deployments"][0]["id"]
+            deployment["name"] = data["deployments"][0]["name"]
+            deployment["slug"] = data["deployments"][0]["slug"]
+    except requests.RequestException as e:
+        logger.error(
+            "Could not complete request to the deployments Semgrep API: %s",
+            deployment_url,
+            exc_info=e,
+        )
+        raise e
```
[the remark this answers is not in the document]
heryxpc: No longer at the code.

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    while has_more:
+        params = {}
+        if cursor:
+            params = {"cursor": cursor}
+        try:
+            response = requests.get(sca_url, params=params, headers=headers)
+            response.raise_for_status()
+            if response.status_code == 200:
```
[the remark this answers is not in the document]
heryxpc: Removed.

heryxpc:
> Leaving suggestions on how to button this up. Looks good overall!
> 
> Can you please also add schema docs, config docs, and the Readme so that users know how to use it, how to set it up, and that we have Semgrep coverage?

PR comments addressed and added docs.
I wasn't sure about the README but added a reference and a page that links to the docs once they are deployed.

docs/root/modules/semgrep/config.md
```diff
@@ -0,0 +1,9 @@
+## Duo Configuration
```
achantavy, line 1:
```suggestion
## Semgrep Configuration
```

cartography/intel/semgrep/findings.py
achantavy: Adding a new relationship is super easy with the data model now ;)

Once we define a "canonical" CVE node in the data model, it'll make sense to include it but I see your point.

cartography/intel/semgrep/findings.py
achantavy: I forget what this conversation thread was originally about since GitHub's commenting seems to have moved it to a different line.

In general though let's try to use lists in Neo4j nodes as little as possible since it's easier to test for relationship connections. Using lists does make sense though for things where it's too heavyweight to create a whole new node type.

achantavy (approved):

cartography/intel/semgrep/findings.py
```diff
@@ -0,0 +1,220 @@
...
+        sca_vuln["transitivity"] = vuln.get("transitivity", None)
+        cves = vuln.get("advisory", {}).get("references", {}).get("cveIds")
+        if len(cves) > 0:
+            # Take the first CVE
+            sca_vuln["cveId"] = vuln["advisory"]["references"]["cveIds"][0]
+        if vuln.get('closestSafeDependency'):
+            dep_fix = f"{vuln['closestSafeDependency']['name']}|{vuln['closestSafeDependency']['versionSpecifier']}"
+        else:
+            dep_fix = None
+        sca_vuln["closestSafeDependency"] = dep_fix
```
ramonpetgrave64, line 114:
```suggestion
        if vuln.get('closestSafeDependency'):
            dep_fix = f"{vuln['closestSafeDependency']['name']}|{vuln['closestSafeDependency']['versionSpecifier']}"
            sca_vuln["closestSafeDependency"] = dep_fix
```

cartography/models/semgrep/deployment.py
```diff
@@ -0,0 +1,19 @@
...
+from cartography.models.core.nodes import CartographyNodeProperties
+from cartography.models.core.nodes import CartographyNodeSchema
+
+
+@dataclass(frozen=True)
+class SemgrepDeploymentProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef('id')
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+    name: PropertyRef = PropertyRef('name')
+    slug: PropertyRef = PropertyRef('slug')
```
ramonpetgrave64, line 13:
```suggestion
    name: PropertyRef = PropertyRef('name', extra_index=True)
    slug: PropertyRef = PropertyRef('slug', extra_index=True)
```

cartography/models/semgrep/findings.py
```diff
@@ -0,0 +1,80 @@
...
+    cve_id: PropertyRef = PropertyRef('cveId', extra_index=True)
+    reachability_check: PropertyRef = PropertyRef('reachability')
+    reachability_condition: PropertyRef = PropertyRef('reachableIf')
+    reachability: PropertyRef = PropertyRef('exposureType')
+    transitivity: PropertyRef = PropertyRef('transitivity')
+    dependency: PropertyRef = PropertyRef('matchedDependency')
+    dependency_fix: PropertyRef = PropertyRef('closestSafeDependency')
+    ref_urls: PropertyRef = PropertyRef('ref_urls')
+    dependency_file: PropertyRef = PropertyRef('dependencyFileLocation_path')
+    dependency_file_url: PropertyRef = PropertyRef('dependencyFileLocation_url')
```
ramonpetgrave64, line 33: These might also be good candidates for indexes, querying all findings associated with a file, for example

docs/root/modules/semgrep/schema.md
```diff
@@ -0,0 +1,91 @@
...
+| Field | Description |
+|-------|--------------|
+| firstseen | Timestamp of when a sync job first discovered this node  |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | Unique id identifying the location of the finding |
+| path | Path of the file where the usage was discovered |
+| start_line | Line where the usage starts |
+| start_column | Column where the usage starts |
+| end_line | Line where the usage ends |
+| end_column | Column where the usage ends |
```
ramonpetgrave64, line 90: In the NodeSchema, you do the abbreviations end_col and start_col.

ramonpetgrave64 (changes requested):

cartography/models/semgrep/findings.py
```diff
@@ -0,0 +1,80 @@
...
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class SemgrepSCAFindingNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef('id')
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+    rule_id: PropertyRef = PropertyRef('ruleId', extra_index=True)
+    repository: PropertyRef = PropertyRef('repositoryName', extra_index=True)
+    summary: PropertyRef = PropertyRef('title')
```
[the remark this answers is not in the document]
heryxpc: Added an extra index.

cartography/models/semgrep/findings.py
heryxpc: Added index to both of them.

docs/root/modules/semgrep/schema.md
heryxpc: Autocomplete typo, fixed it.

achantavy (approved):

ramonpetgrave64 (approved):

PR #1227, opened 2023-07-26
AWS: Extend aws_handle_regions to cover IAM get functions

alowde-ps:
This change adds the aws_handle_regions decorator to the IAM get functions so that we can safely skip over API calls that the caller is unauthorized to make.

To support this the aws_handle_functions decorator has been extended to return an empty type based on the signature of the calling function, and the `AWSGetFunc` type is now a union of Callable[..., List] and Callable[..., Dict[Any, Any].

This has been tested in action and passes mypy validation of the type changes.

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -12,6 +12,7 @@ import neo4j
 from cartography.intel.aws.permission_relationships import parse_statement_node
 from cartography.intel.aws.permission_relationships import principal_allowed_on_resource
 from cartography.stats import get_stats_client
+from cartography.util import aws_handle_regions
 from cartography.util import merge_module_sync_metadata
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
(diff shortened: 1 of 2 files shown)


ramonpetgrave64:
@alowde-ps While you're getting the tests to work, I've been thinking a bit more on this. What you're proposing sounds like a best-effort mode, which I think is a significant shift to how we've done things before. Personally I'd rather this be tied into our existing cli switch `--aws-best-effort-mode` somehow. wdyt @achantavy 

Also, I see that you put the decorator on functions like `get_group_managed_policy_data(boto3_session: boto3.session.Session, group_list: List[Dict])` which are meant to fetch for many items in the provided list. From our conversation, I think your use-case was when any single item is Denied. In this case, it's less appropriate to do `@handle_regions` for the entire list of items, and better to refactor the individual fetches into a separate function to receive the decorator.

achantavy:
Closing this one out since a lot of the changes would be tied up under a broader best-effort-mode

PR #1233, opened 2023-08-22
fix: GSuite data only populated on match

resilience-jychp:
On Gsuite Intel, User and Group properties are only populated on existing node (ON MATCH).
On first run only ID is populated.

This PR fix this issue.

### DIFF
--- a/cartography/intel/gsuite/api.py
+++ b/cartography/intel/gsuite/api.py
@@ -130,6 +130,14 @@ def load_gsuite_groups(neo4j_session: neo4j.Session, groups: List[Dict], gsuite_
         UNWIND $GroupData as group
         MERGE (g:GSuiteGroup{id: group.id})
         ON CREATE SET
+        g.group_id = group.id,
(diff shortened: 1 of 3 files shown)


cartography/intel/gsuite/api.py
```diff
@@ -130,6 +130,14 @@ def load_gsuite_groups(neo4j_session: neo4j.Session, groups: List[Dict], gsuite_
         UNWIND $GroupData as group
         MERGE (g:GSuiteGroup{id: group.id})
         ON CREATE SET
+        g.group_id = group.id,
```
achantavy, line 133: Instead of having these as `ON CREATE SET` I think it makes sense to have them in `SET`.

`ON CREATE SET` is mostly for things like ID fields and firstseen.

achantavy (changes requested):
Thanks!

cartography/intel/gsuite/api.py
resilience-jychp: Indeed.

PR #1249, opened 2023-09-23
fix: looks like a typo - network interfaces with attached elb wont lo…

petersutter5:
fix: looks like a typo - network interfaces with attached elb wont load into neo4j because they are tuple instead of string


petersutter5:
Your Signature
Current version of the CLA: [1.0](https://oss.lyft.com/cla/CLA/1.0)

Username

petersutter5

CLA Version

Signed and current (1.0)

PR #1243, opened 2023-09-19
refactor(okta): Update Okta Module to use Models

csanders-git:
### Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Refactoring (no functional changes)
- [x] Documentation update

### Summary

Upgrade the Okta connector from the legacy `okta` 0.x client to Okta Python SDK 3.4.4 and migrate its graph writes to Cartography's declarative data model.

This change:

- Uses one configured Okta SDK client with consistent pagination and error propagation.
- Migrates organizations, users, groups, applications, factors, trusted origins, and their relationships to `CartographyNodeSchema` models.
- Adds Okta user types, user and group administrator roles, and authenticators.
- Preserves AWS SAML and AWS IAM Identity Center group-to-role mappings.
- Adds ontology labels for tenants, user accounts, user groups, permission roles, and third-party applications.
- Uses canonical `MEMBER_OF` relationships while retaining `MEMBER_OF_OKTA_GROUP` as a deprecated compatibility edge until v1.0.0.
- Prevents authenticator provider secrets from being stored in Neo4j.
- Includes a temporary compatibility shim for valid application responses that Okta SDK 3.4.4 rejects.
- Replaces handwritten Okta indexes and cleanup logic with schema-managed indexes and cleanup where possible.

### Related issues or links

- Original compatibility discussion: https://github.com/cartography-cncf/cartography/pull/1243#issuecomment-5351995038
- Okta SDK user-type response modeling: https://github.com/okta/okta-sdk-python/issues/535
- Okta SDK application validation fixes: https://github.com/okta/okta-sdk-python/pull/546
- Okta SDK application validation fixes: https://github.com/okta/okta-sdk-python/pull/574

### Breaking changes

The graph model now uses ontology-backed node labels and canonical relationships:

- `OktaAdministrationRole` is replaced by `OktaUserRole` and `OktaGroupRole`, connected with `HAS_ROLE`.
- The deprecated `Human`/`IDENTITY_OKTA` integration is replaced by `User`/`HAS_ACCOUNT` through the `UserAccount` ontology mapping.
- `MEMBER_OF` replaces `MEMBER_OF_OKTA_GROUP`. Both relationships are written until v1.0.0 to preserve compatibility.

Migration cleanup removes obsolete nodes and relationships after their replacements are available.

### How was this tested?

[...]

PR #1259, opened 2023-09-27
Adding support for okta roles

csanders-git:
Admin roles are critical to understanding what users can do WRT to Okta. this PR adds support for showing which users/groups/apps are assigned which okta admin roles.


jychp:
Included in #1243

PR #1262, opened 2023-10-02
Fix docker-compose, Dockerfile, and container documentation

achantavy:
Docker-compose fixes
- correctly points cartography container's NEO4J_URL at the Neo4j container instead of localhost. This was not working at all before.
- uses the cartography image built locally instead of the one in ghcr. Per testing-with-docker.md, this is the intended path.

Dockerfile fixes
- Updates Dockerfile permissions so that the main test user has permissions to write to the .coverage file for unit and integration tests.

Documentation fixes
- Updates docker instructions to clarify that the main dockerfile is meant for testing purposes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,21 +4,25 @@ WORKDIR /srv/cartography
 
 ENV PATH=/venv/bin:$PATH
 RUN apt-get update && \
-    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.8-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.8-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl make && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
 # Installs pip supported by python3.8
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3.8 get-pip.py
 
+# Create cartography user so that we can give it ownership of the directory later for unit&integ tests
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -4,21 +4,25 @@ WORKDIR /srv/cartography
...
-    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.8-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.8-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl make && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
 # Installs pip supported by python3.8
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3.8 get-pip.py
 
+# Create cartography user so that we can give it ownership of the directory later for unit&integ tests
+RUN groupadd cartography && \
```
achantavy, line 15: Moved this up earlier so that we can set ownership properly on the `COPY`

Dockerfile
```diff
@@ -4,21 +4,25 @@ WORKDIR /srv/cartography
 
 ENV PATH=/venv/bin:$PATH
 RUN apt-get update && \
-    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.8-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.8-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl make && \
```
achantavy, line 7: so that we can run `make test` in the container

ramonpetgrave64 (approved):

PR #1265, opened 2023-10-09
Semgrep Intel module updates, scoped analysis jobs

heryxpc:
* Consume new official public API https://semgrep.dev/api/v1/docs/#tag/SupplyChainService
* Updated `SemgrepSCAFinding.id` to use the unique value identifier provided by Semgrep.
* Connect `CVE` nodes to `SemgrepSCAFinding`
* Connect `Dependency` to `SemgrepSCAFinding`
* Add analysis job to set a `:SemgrepSCAFinding.reachability_risk` to measure risk based on Impact (Severity) and Likelihood (Reachability). Used Appendix I from [NIST SP 800-30 Guide for conducting risk assessments](https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-30r1.pdf) as reference. 
* Introduces scoped analysis jobs


docs/root/modules/semgrep/schema.md
```diff
@@ -73,6 +73,18 @@ Represents a [Semgre Supply Chain](https://semgrep.dev/docs/semgrep-supply-chain
...
+    ```
+    (:SemgrepSCAFinding)-[:AFFECTS]->(:Dependency)
+    ```
+
+- A SemgrepSCAFinding linked to a CVE (optional)
+
+    ```
+    (:SemgrepSCAFinding)<-[:LINKED_TO]-(:CVE)
+    ```
+
```
ramonpetgrave64, line 87: Also update the schema of CVE and Dependency, if they exist

cartography/intel/semgrep/findings.py
```diff
@@ -52,26 +52,43 @@ def get_sca_vulns(semgrep_app_token: str, deployment_id: str) -> List[Dict[str,
     param: deployment_id: The Semgrep deployment id to use for retrieving SCA vulns.
     """
     all_vulns = []
-    sca_url = f"https://semgrep.dev/api/sca/deployments/{deployment_id}/vulns"
+    sca_url = f"https://semgrep.dev/api/v1/deployments/{deployment_id}/ssc-vulns"
     has_more = True
-    cursor = ""
+    cursor: Dict = {}
```
ramonpetgrave64, line 57:
```suggestion
    cursor: Dict[str, str] = {}
```

cartography/intel/semgrep/findings.py
```diff
@@ -52,26 +52,43 @@ def get_sca_vulns(semgrep_app_token: str, deployment_id: str) -> List[Dict[str,
...
     }
 
     while has_more:
-        params = {}
+        request_data = {
+            "deploymentId": deployment_id,
+            "pageSize": 100,
+            "exposure": ["UNREACHABLE", "REACHABLE", "UNKNOWN_EXPOSURE"],
+            "refs": ["_default"],
+        }
```
ramonpetgrave64, line 70: Move this outside the loop

cartography/intel/semgrep/findings.py
```diff
@@ -52,26 +52,43 @@ def get_sca_vulns(semgrep_app_token: str, deployment_id: str) -> List[Dict[str,
...
+            request_data = {
+                "deploymentId": deployment_id,
+                "pageSize": 100,
+                "exposure": ["UNREACHABLE", "REACHABLE", "UNKNOWN_EXPOSURE"],
+                "refs": ["_default"],
+                "cursor": {
+                    "vulnOffset": cursor["vulnOffset"],
+                    "issueOffset": cursor["issueOffset"],
+                },
+            }
```
ramonpetgrave64, line 81:
```suggestion
            request_data.update({
                "cursor": {
                    "vulnOffset": cursor["vulnOffset"],
                    "issueOffset": cursor["issueOffset"],
                },
            })
```

cartography/intel/semgrep/findings.py
```diff
@@ -86,8 +103,7 @@ def transform_sca_vulns(raw_vulns: List[Dict[str, Any]]) -> Tuple[List[Dict[str,
     for vuln in raw_vulns:
         sca_vuln: Dict[str, Any] = {}
         # Mandatory fields
-        unique_id = f"{vuln['repositoryName']}|{vuln['advisory']['ruleId']}"
-        sca_vuln["id"] = unique_id
+        sca_vuln["id"] = vuln["groupKey"]
```
ramonpetgrave64, line 106: non-blocker: This changes the schema for id. Could you instead introduce groupKey as another property?

cartography/models/semgrep/findings.py
```diff
@@ -68,6 +68,40 @@ class SemgrepSCAFindingToGithubRepoRel(CartographyRelSchema):
...
+class SemgrepSCAFindingToCVERelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:SemgrepSCAFinding)<-[:LINKED_TO]-(:CVE)
+class SemgrepSCAFindingToCVERel(CartographyRelSchema):
+    target_node_label: str = 'CVE'
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'id': PropertyRef('cveId')},
```
ramonpetgrave64, line 98: Is the property cveId or id?
https://github.com/lyft/cartography/blob/f3f237cb9625f26205ead1da971d5fb7cf3127b0/cartography/intel/cve/feed.py#L45

cartography/models/semgrep/findings.py
```diff
@@ -68,6 +68,40 @@ class SemgrepSCAFindingToGithubRepoRel(CartographyRelSchema):
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "LINKED_TO"
```
ramonpetgrave64, line 101: rename to something else like CAUSES

cartography/models/semgrep/findings.py
heryxpc: It's `cveId`, since its taken from https://github.com/lyft/cartography/blob/297434ebb8f6398ce9e56fea2cab4ceb92bf06f6/cartography/intel/semgrep/findings.py#L125

cartography/models/semgrep/findings.py
heryxpc: Sorry to push back, but `:CVE` nodes do not cause `:SemgrepSCAFinding`.
A `:SemgrepSCAFinding` is a result from a scan that might or might not have attached CVE information.
Hence, the CVE is not causing the finding, but rather the CVE is connected to the finding.
Other findings from different tools might be as well connected to the same CVE.

cartography/intel/semgrep/findings.py
heryxpc: Good callout, let me update.

cartography/data/jobs/analysis/semgrep_sca_risk_analysis.json
```diff
@@ -0,0 +1,66 @@
...
+    "name": "Semgrep SCA findings reachability risk analysis based on likelihood and impact. Impact = Severity, Likelihood = reachability + reachability_check",
+    "statements": [
+        {
+            "__comment__": "ARCHIVED -> Risk = Info",
+            "query": "MATCH (g:GitHubRepository{archived:true})<-[:FOUND_IN]-(s:SemgrepSCAFinding) SET s.reachability_risk = 'INFO' return COUNT(*) as TotalCompleted",
+            "iterative": false
+        },
+        {
+            "__comment__": "reachability = UNREACHABLE -> Risk = Info",
+            "query": "MATCH (s:SemgrepSCAFinding{reachability:'UNREACHABLE'}) SET s.reachability_risk = 'INFO' return COUNT(*) as TotalCompleted",
```
ramonpetgrave64, line 11: since there is already a 1:1 mapping between finding and repo, it seems like all of these calculations can be done from python in the "transform" method.

heryxpc: Is this a recommendation or a must?
TL;DR it's in cypher because that is what other module do.

I followed https://lyft.github.io/cartography/dev/writing-analysis-jobs.html?highlight=analysis and examples at https://github.com/lyft/cartography/blob/master/cartography/data/jobs/analysis to put any extra information based on an analysis, not from the original source, into the ingested nodes. The data is saved into the same node and as part of the same sync, but required an analysis to determine the new attribute.

If you have any other examples that follow that pattern (analyze, calculate and add data during transform step) I could follow them to keep the approach consistent in the codebase.

cartography/data/jobs/analysis/semgrep_sca_risk_analysis.json
```diff
@@ -0,0 +1,66 @@
+{
+    "name": "Semgrep SCA findings reachability risk analysis based on likelihood and impact. Impact = Severity, Likelihood = reachability + reachability_check",
+    "statements": [
+        {
+            "__comment__": "ARCHIVED -> Risk = Info",
```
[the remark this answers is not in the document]
heryxpc: Sure, added it https://github.com/lyft/cartography/pull/1265/files#diff-275b8d4c17a33754368587a25b15b91173e0f912d708df582ed35ff8f9480702R53

cartography/data/jobs/analysis/semgrep_sca_risk_analysis.json
ramonpetgrave64: I won't require it, but I think it will all be much more performant, understandable and testable if done in python, since it wouldn't need IO to the database.

ramonpetgrave64 (approved):

cartography/intel/semgrep/findings.py
```diff
@@ -206,6 +221,7 @@ def sync(
     vulns, usages = transform_sca_vulns(raw_vulns)
     load_semgrep_sca_vulns(neo4j_sesion, vulns, semgrep_deployment["id"], update_tag)
     load_semgrep_sca_usages(neo4j_sesion, usages, semgrep_deployment["id"], update_tag)
+    run_analysis_job('semgrep_sca_risk_analysis.json', neo4j_sesion, common_job_parameters)
```
achantavy, line 224: One thing to keep in mind: this analysis job is currently not scoped to run for a specific Semgrep tenant or specific Github tenant, so the statements where it deletes properties will happen for _all_ Github tenants and _all_ Semgrep tenants.

I think we should

1. edit the analysis job so that we only set or remove properties for Semgrep objects in the current Semgrep tenant.
2. introduce the idea of a scoped analysis job where we can define a set of analysis queries to run fora given tenant.


For item 2 we would make the following updates:

I. cartography/util.py:

```python
def run_scoped_analysis_job(
    filename: str,
    neo4j_session: neo4j.Session,
    common_job_parameters: Dict,
) -> None:
    """
    Enriches existing graph data scoped to a given sub resource - e.g. the current AWS account.
    Runs the queries in the cartography.data.jobs.scoped_analysis directory for the given `filename`. View the queries
    in cartography.data.jobs.scoped_analysis for specifics.
    """
    GraphJob.run_from_json(
        neo4j_session,
        read_text('cartography.data.jobs.scoped_analysis', filename),
        common_job_parameters,
        get_job_shortname(filename),
    )
```

II. add cartography/data/jobs/scoped_analysis/ as a new folder and move the semgrep_sca_risk_analysis.json file to it.

I played with this idea before but [never actually merged it](https://github.com/lyft/cartography/pull/1219/commits/3c7b0bd25b0798b71c00d48be0f970dd8c118031#diff-24cb7d746e097729d20949fb290b122181beebc03590e9ef937d2519614f32b8).

---

For the sake of giving an complete explanation, I'm going to write more details here but I do not think we should build it this way. Feel free to ignore if you're trying to just get on with the PR.

Now, if having this analysis job run for all tenants is the intended behavior, then we can have it the way you had it because all cartography analysis jobs eventually get run at the end of the sync (the [call itself](https://github.com/lyft/cartography/blob/2fa2cb0aa54b74844d9711c8bd30f759bf315f9f/cartography/sync.py#L56), the [details](https://github.com/lyft/cartography/blob/2fa2cb0aa54b74844d9711c8bd30f759bf315f9f/cartography/intel/analysis.py#L30-L31)). I actually don't think we should do it this way though because if you mix and match sync stages as part of a [custom sync](https://lyft.github.io/cartography/dev/developer-guide.html?highlight=custom%20sync#implementing-custom-sync-commands) you'll have to end up writing an internal shim to run the analysis job, so this can be messy.

cartography/data/jobs/analysis/semgrep_sca_risk_analysis.json
achantavy: I really like analysis jobs because they let us simplify the graph in powerful ways, but they also scare me because they can update and remove fields in ways that are hard to understand if not done carefully. 

Converting an analysis job to run in Python code with a transform function with the data model gives us nice conveniences like automatic cleanups of stale attributes and relationships, but it's also inflexible and can marginally slow things down since all the IO will be handled by Python and not the database.

So I think we should keep things as an analysis job in this PR and ensure that we test for the right things and create the right patterns so that other analysis jobs can copy them.

Thanks for being patient as we figure this out.

achantavy (changes requested):
I also just noticed semgrep wasn't added as a sync stage here: https://github.com/lyft/cartography/blob/2fa2cb0aa54b74844d9711c8bd30f759bf315f9f/cartography/sync.py#L55

cartography/data/jobs/analysis/semgrep_sca_risk_analysis.json
achantavy: >I won't require it, but I think it will all be much more performant, understandable and testable if done in python, since it wouldn't need IO to the database.

Pretty sure that for perf it's generally better to keep things done by the DB and having data hit Python makes it slower. We are willing to take the perf hit though if it makes things more testable and maintainable (like with introducing the data model). However in this case I think being able to run neo4j queries to do analysis is useful and we should find ways to build guardrails around them.

tests/unit/cartography/intel/github/test_github.py
```diff
@@ -26,7 +27,9 @@ def test_fetch_all_handles_retries(
     '''
     # Arrange
     exception = HTTPError
-    mock_fetch_page.side_effect = exception('my-error')
+    response = Response()
```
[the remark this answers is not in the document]
heryxpc: Linter error, the next lines needed a `Response` object otherwise mypy complains.

tests/integration/cartography/intel/semgrep/test_findings.py
```diff
@@ -238,3 +268,68 @@ def test_sync(mock_get_sca_vulns, mock_get_deployment, neo4j_session):
...
+        (
+            "132465::::ssc-92af1d99-4fb3-4d4e-a9f4-d57572cd6590::reachable",
+            "REACHABLE",
+            "MANUAL_REVIEW_REACHABLE",
+            "HIGH",
+            "MEDIUM",
+        ),
+    }
+    assert (
+        check_nodes(
```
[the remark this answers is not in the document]
heryxpc: Good callout, will update.

heryxpc:
> 'duo': cartography.intel.duo.start_duo_ingestion,

TY, I assume I didn't notice because I was using a custom sync.

cartography/intel/semgrep/findings.py
heryxpc: This is actually interesting, I was assuming a single tenant but it can clearly apply to multiple ones and the analysis wouldn't change (unless we add an external piece of information for the analysis).

So, we can apply the same rules to each Semgrep deployment and shouldn't affect the state of the graph.

One extra piece of information is that we query the deployment based on the `SEMGREP_APP_TOKEN`. It always returns a single tenant based on their current documentation: https://semgrep.dev/api/v1/docs/#tag/Deployment/operation/semgrep_app.saas.handlers.deployment.openapi_list_deployments
> Currently available auth scope does not extend over more than 1 deployment. This endpoint returns that single deployment your token can access. The endpoint additionally returns links towards related resources available on this web API.

So, I will go with option 1, considering currently it will:
1. Always be a single deployment per sync
2. Use the deployment to only update the `reachability_risk` property for `(:SemgrepDeployment{id:DEPLOYMENT_D. lastupdated: UPDATE_TAG})-[:RESOURCE]->(:SCASemgrepFinding)` nodes.

docs/root/modules/semgrep/schema.md
```diff
@@ -45,11 +45,12 @@ Represents a [Semgre Supply Chain](https://semgrep.dev/docs/semgrep-supply-chain
 | summary | A short title summarizing of the finding |
 | description | Description of the vulnerability. |
 | package_manager | The ecosystem of the dependency where the finding was discovered (e.g. pypi, npm, maven) |
-| severity | Severity of the finding (e.g. CRITICAL, HIGH, MEDIUM, LOW) |
+| severity | Severity of the finding based on Semgrep analysis (e.g. CRITICAL, HIGH, MEDIUM, LOW) |
 | cve_id | CVE id of the vulnerability from NVD. Check [cve_schema](../cve/schema.md) |
 | reachability_check | Whether the vulnerability reachability is confirmed, not confirmed or needs to be manually confirmed |
 | reachability_condition | Description of the reachability condition (e.g. reachable if code is used in X way) |
 | reachability | Whether the vulnerability is reachable or not |
+| reachability_risk | Risk of the vulnerability based on likelihod given by reachability (e.g. CRITICAL, HIGH, MEDIUM, LOW). Check [semgrep_sca_risk_analysis.json](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/analysis/semgrep_sca_risk_analysis.json) |
```
achantavy, line 53: One last change: can you update the URL and include a short explanation on how this is determined?

achantavy (approved):

PR #1266, opened 2023-10-11
Fix #1198: handle github ratelimit

### LINKED ISSUE #1198: Github: ratelimit handling
*Title*: *Github Ratelimit handinling*

*Description*:
Github's APIs have a ratelimit. Since some Github Orgs can be rather large, that means a cartography sync can oftentimes hit the ratelimit. But we they provide us with a Grapqhl object so we can check the when the ratelimit resets.
* https://docs.github.com/en/graphql/overview/resource-limitations#returning-a-calls-rate-limit-status

One suggestion: implement the check within our `fetch_page` function, perhaps sleeping until the reset time.
* https://github.com/lyft/cartography/blob/master/cartography/intel/github/util.py#L54C5-L54C15

*Relevant Links*:
* https://docs.github.com/en/graphql/overview/resource-limitations#returning-a-calls-rate-limit-status
* https://github.com/lyft/cartography/blob/master/cartography/intel/github/teams.py#L91

ramonpetgrave64:
Github has a requests-per-hour ratelimit that resets per hour.

### Changes

- when the `remaining` is lower than a threshold of `500`, delays github graphql requests until the `reset` time.
- changes the retry mode from constant to exponential backoff
- change the page size of the repos query to `50` from `100`. The github API would very frequently timeout, but is much better now.

### Testing

- added unit tests
- manual testing

### DIFF
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -36,7 +36,6 @@ def handle_rate_limit_sleep(token: str) -> None:
     rate_limit_obj = response_json['resources']['graphql']
     remaining = rate_limit_obj['remaining']
     threshold = _GRAPHQL_RATE_LIMIT_REMAINING_THREASHOLD
-    print(remaining)
     if remaining > threshold:
         return
     reset_at = datetime.fromtimestamp(rate_limit_obj['reset'], tz=tz.utc)
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -1,6 +1,9 @@
 import json
 import logging
 import time
+from datetime import datetime
+from datetime import timedelta
+from datetime import timezone as tz
 from typing import Any
 from typing import Dict
 from typing import List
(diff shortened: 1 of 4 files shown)


cartography/intel/github/util.py
```diff
@@ -10,16 +13,42 @@
...
     nodes: List[Dict[str, Any]]
     edges: List[Dict[str, Any]]
 
 
+def handle_rate_limit_sleep(token: str) -> None:
+    '''
+    Check the remaining rate limit and sleep if remaining is below threshold
+    :param token: The Github API token as string.
+    '''
+    response = requests.get('https://api.github.com/rate_limit', headers={'Authorization': f"token {token}"})
```
achantavy, line 33: Rather than add another call, would it be possible to use the `rateLimit` object in the graphql response object itself? https://github.com/lyft/cartography/blob/f3f237cb9625f26205ead1da971d5fb7cf3127b0/cartography/intel/github/teams.py#L91-L96

This would let us save an API call for each page.

ramonpetgrave64: That's why I tried originally, but if the graphql ratelimit is already fully consumed, then the query would not return a response. Calling the rest endpoint consumes a different ratelimit, `core` I believe. Test data shows that the limit for core is 60, but it's actually >5k for authenticated requests.

cartography/intel/github/util.py
```diff
@@ -10,16 +13,42 @@
 
 import requests
 
+
 logger = logging.getLogger(__name__)
 # Connect and read timeouts of 60 seconds each; see https://requests.readthedocs.io/en/master/user/advanced/#timeouts
 _TIMEOUT = (60, 60)
+_GRAPHQL_RATE_LIMIT_REMAINING_THREASHOLD = 500
```
skiptomyliu, line 20:
```suggestion
_GRAPHQL_RATE_LIMIT_REMAINING_THRESHOLD = 500
```
typo

ramonpetgrave64: Thanks. fixed

cartography/intel/github/util.py
achantavy: From conversation we had offline, if we start the sync without knowing that the `remaining` quota is already 0, then the graphql response won't even return and the code will crash.

tests/unit/cartography/intel/github/test_github.py
```diff
@@ -20,5 +32,59 @@ def test_fetch_all_handles_retries(mock_fetch_page: Mock) -> None:
...
+    '''
+    # Arrange
+    mock_datetime.fromtimestamp = datetime.fromtimestamp
+    now = datetime(year=2040, month=1, day=1, hour=19, minute=0, second=0, tzinfo=tz.utc)
+    mock_datetime.now = Mock(return_value=now)
+    reset = (now + timedelta(minutes=47)).timestamp()
+    # sleep for one extra minute for safety
+    expected_sleep_seconds = timedelta(minutes=48).seconds
+
+    # above threoshold
```
achantavy, line 60:
```suggestion
    # above threshold
```

tests/unit/cartography/intel/github/test_github.py
```diff
@@ -20,5 +32,59 @@ def test_fetch_all_handles_retries(mock_fetch_page: Mock) -> None:
...
+    resp_0 = deepcopy(RATE_LIMIT_RESPONSE_JSON)
+    resp_0['resources']['graphql']['remaining'] = _GRAPHQL_RATE_LIMIT_REMAINING_THRESHOLD + 1
+    resp_0['resources']['graphql']['reset'] = reset
+
+    # below threhold
```
achantavy, line 65:
```suggestion
    # below threshold
```

tests/unit/cartography/intel/github/test_github.py
```diff
@@ -20,5 +32,59 @@ def test_fetch_all_handles_retries(mock_fetch_page: Mock) -> None:
...
+    # Assert
+    mock_datetime.now.assert_not_called()
+    mock_sleep.assert_not_called()
+
+    # reset mocks
+    mock_datetime.reset_mock()
+    mock_sleep.reset_mock()
+
+    # Act
+    # mock_requests_get.return_value = Mock(json=Mock(return_value=resp_1))
```
achantavy, line 86:
```suggestion
```

Remove?

achantavy (approved):

PR #1268, opened 2023-10-18
Fixed bug with iterating over a non existing object

austincomstockzoom:
If there's an issue with getting the user, such as a nonexistent user, the dictionary will be overwritten with None and cause a key error upon attempting to look up the access keys.


achantavy (commented):
Hi @austincomstockzoom - thanks for the PR.

Can you please provide more context on the problem that this solves? Specifically, can you include a crash dump  or details on the expected behavior vs actual behavior?

Our IAM code is pretty crufty and there's a lot I'd like to fix in it, so I'd like to talk about the specific problems that it has first before writing code, since I think a lot of my review suggestions would not pertain to things in your code, if that makes sense.

PR #1278, opened 2023-11-29
LoadBalancerV2 target groups: include port and protocol to EC2Instance

heryxpc:
AWS api exposes a port and protocol per TargetGroup (see https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_TargetGroup.html)

Using those values to identify which EC2 instances are exposed by which protocols and ports based on LoadBalancer configuration.

### DIFF
--- a/cartography/intel/aws/ec2/load_balancer_v2s.py
+++ b/cartography/intel/aws/ec2/load_balancer_v2s.py
@@ -159,6 +159,7 @@ def load_load_balancer_v2_target_groups(
     MERGE (elbv2)-[r:EXPOSE]->(instance)
     ON CREATE SET r.firstseen = timestamp()
     SET r.lastupdated = $update_tag
+        r.port = $PORT, r.protocol = $PROTOCOL
     WITH instance
     MATCH (aa:AWSAccount{id: $AWS_ACCOUNT_ID})
     MERGE (aa)-[r:RESOURCE]->(instance)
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/ec2/load_balancer_v2s.py
```diff
@@ -159,6 +159,7 @@ def load_load_balancer_v2_target_groups(
     MERGE (elbv2)-[r:EXPOSE]->(instance)
     ON CREATE SET r.firstseen = timestamp()
     SET r.lastupdated = $update_tag
+        r.port = $PORT, r.protocol = $PROTOCOL
```
achantavy, line 162: You missed a comma. Needs to be

```
    SET r.lastupdated = $update_tag,
        r.port = $PORT, r.protocol = $PROTOCOL
```

Also, is the ARN of the target group available? If so, maybe we should include the ARN of the target group as `r.target_group_arn` so that it's easily searchable.

achantavy: Oh also let's remember to update the schema doc

cartography/intel/aws/ec2/load_balancer_v2s.py
```diff
@@ -157,8 +157,10 @@ def load_load_balancer_v2_target_groups(
     ingest_instances = """
     MATCH (elbv2:LoadBalancerV2{id: $ID}), (instance:EC2Instance{instanceid: $INSTANCE_ID})
     MERGE (elbv2)-[r:EXPOSE]->(instance)
-    ON CREATE SET r.firstseen = timestamp()
-    SET r.lastupdated = $update_tag
+    ON CREATE SET r.firstseen = timestamp(),
+                  r.target_group_arn = $TARGET_GROUP_ARN
```
achantavy, line 161: Is it possible for the target_group_arn to change? To be safe, let's move this to the SET clause.

achantavy (approved):

PR #1279, opened 2023-12-10
Update to ECRImage ID Mapped into AWS Inspector Findings to correctly map to ECR Image nodes 

udm17:
Correct id to match against the ECR image to form relationship

### DIFF
--- a/cartography/intel/aws/inspector.py
+++ b/cartography/intel/aws/inspector.py
@@ -85,7 +85,7 @@ def transform_inspector_findings(results: List[Dict[str, Any]]) -> Tuple[List[Di
         if f['resources'][0]['type'] == "AWS_EC2_INSTANCE":
             finding['instanceid'] = f['resources'][0]['id']
         if f['resources'][0]['type'] == "AWS_ECR_CONTAINER_IMAGE":
-            finding['ecrimageid'] = f['resources'][0]['id']
+            finding['ecrimageid'] = f['resources'][0]['id'].split('/')[2]
         if f['resources'][0]['type'] == "AWS_ECR_REPOSITORY":
             finding['ecrrepositoryid'] = f['resources'][0]['id']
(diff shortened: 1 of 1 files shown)


udm17:
The ID of the ECR Image mapped is the same as the image digest while the image ecrimageid obtained in Inspector is the arn. This change allows the id to be matched correctly when creating the relationship

achantavy:
Hi @udm17 thanks for the PR. Can you please describe the testing you performed for this PR? It'd be helpful to show a "before" and "after" description, maybe with a screenshot.

udm17:
Hi @achantavy. I've attached a few screenshot below for reference. The ECR Image id is the same as the digest, which begins with sha:~ identifier. On the other hand, the one found via boto in the Inspector module is in the arn format, containing the arn begining, the repo and finally, the id/digest value. 

The proposed change is a simple string split in the Inspector repo to extract the id from the arn and use that for the relationship formation matching.  

ID_before_change
mapping_after_change
mapping_before_change
id_after_change

achantavy (approved):
Ah I understand. Currently the ECRImage-to-AWSInspectorFinding relationship does not work because [we use image digests for the id of an ECRImage node](https://github.com/lyft/cartography/blob/86c79370f923ce6450fed085aed9932bec89b101/cartography/intel/aws/ecr.py#L109) but we [join](https://github.com/lyft/cartography/blob/86c79370f923ce6450fed085aed9932bec89b101/cartography/models/aws/inspector/findings.py#L105C1-L106C1) InspectorFindings to ECRImages [using the full arn](https://github.com/lyft/cartography/blob/86c79370f923ce6450fed085aed9932bec89b101/cartography/intel/aws/inspector.py#L88). This PR puts the image digest there so that the relationship gets created as intended. 

I will approve and merge in this PR for now, but we will eventually change the id fields of each AWS node to be the arn. We'll announce and document this when it happens.

PR #1280, opened 2024-01-12
Update CVE intel module to use NVD CVE API v2.0

heryxpc:
Issue https://github.com/lyft/cartography/issues/1275

Update the CVE intel module to the new [API 2.0 for vulnerabilities](https://nvd.nist.gov/developers/vulnerabilities).

Followed the [API 2.0 workflow](https://nvd.nist.gov/developers/api-workflows), but kept the year sync for compatibility with previous module version. Added as well delay among API calls to keep among [rate limits established](https://nvd.nist.gov/developers/start-here#divRateLimits).

Finally, switched to the Cartography ORM [model](https://github.com/lyft/cartography/tree/master/cartography/models/core) adding a new `CVEFeed` node representing the `:RESOURCE` parent of all `CVE` nodes.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -336,9 +336,9 @@ class CLI:
         parser.add_argument(
             '--nist-cve-url',
             type=str,
-            default='https://nvd.nist.gov/feeds/json/cve/1.1',
+            default='https://services.nvd.nist.gov/rest/json/cves/2.0/',
             help=(
-                'The base url for the NIST CVE data. Default = https://nvd.nist.gov/feeds/json/cve/1.1'
+                'The base url for the NIST CVE data. Default = https://services.nvd.nist.gov/rest/json/cves/2.0/'
             ),
         )
         parser.add_argument(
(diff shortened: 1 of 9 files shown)


cartography/cli.py
```diff
@@ -685,6 +693,13 @@ def main(self, argv: str) -> int:
         else:
             config.semgrep_app_token = None
 
+        # CVE feed config
+        if config.cve_api_key_env_var:
+            logger.debug(f"Reading NVD CVE API key environment variable {config.cve_api_key_env_var}")
```
AgoraSecurity, line 698: Leaving API Keys in logs is not a good security practice. I suggest either to remove the key completely or at least remove / hide most of the characters.

AgoraSecurity (commented):
Feedback on logs

cartography/cli.py
heryxpc: Agree. On this case it's not an API key that is being logged, but the ENV_VAR holding it to follow the same pattern from previous API key configurations.

AgoraSecurity (approved):
LGTM

PR #1286, opened 2024-02-22
Fix ghpages python version string

achantavy:
Similar issue as this: https://github.com/actions/setup-python/issues/401#issuecomment-1848265537

`3.10` gets interpreted as `3.1` when I want it to be `"3.10"`.

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.10
+          python-version: "3.10"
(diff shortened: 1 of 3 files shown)


.github/workflows/sphinx.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.10
+          python-version: "3.10"
```
heryxpc, line 27: Fix templating is broken, I wouldn't spend too much time on updating it: https://github.com/lyft/cartography/issues/1284

heryxpc (dismissed):

kledo-lyft (approved):

Issue #1293, opened 2024-04-05
Kandji Device support


chandanchowdhury:
**Feature request template**

*Title*: *Kandji Device*

*Description*:
Ingest Kandji device inventory similar to Jamf.

[optional *Relevant Links*:]
https://api-docs.kandji.io/#ecf2a684-45f5-49ad-b3d7-2da346b2927e

*Additional context*:
PR incoming.

PR #1294, opened 2024-04-05
Kandji device suport

chandanchowdhury:
#1293


achantavy (changes requested):
Hi @chandanchowdhury, thank you for the PR!

Before we move forward, can you please try using the data model to define your objects? Our guide here: https://lyft.github.io/cartography/dev/writing-intel-modules.html#load. We're trying to have all new modules going forward use it. It makes things _a lot_ less error prone in production since it handles transactions and automatically creates cleanup jobs.

Reach out on Slack if it's easier to chat there, or let me know too if you wanna ask questions live at our next meeting.

chandanchowdhury:
Thank you @achantavy for reviewing the PR.
I have added the model.

Question: Are you able to squash the unverified commits or have any tip on how to fix those?

PR #1295, opened 2024-04-12
Kandji device

chandanchowdhury:
Add support for Kandji device.

#1293


chandanchowdhury:
@achantavy Requesting review of this PR

cartography/models/kandji/device.py.backup
```diff
@@ -0,0 +1,52 @@
+from dataclasses import dataclass
```
achantavy, line 1: Let's remove this file for now

tests/integration/cartography/intel/kandji/test_kandji.py
```diff
@@ -0,0 +1,27 @@
+import cartography.intel.kandji
+import tests.data.kandji.devices
+
+TEST_UPDATE_TAG = 123456789
+
+
+def test_load_kandji_devices_data(neo4j_session):
```
achantavy, line 7: You can copy this test: https://github.com/lyft/cartography/blob/7601060cca2bd73105dfde059659f2d985958ad1/tests/integration/cartography/intel/aws/test_emr.py#L25-L41

Using `check_nodes()` will help save some typing.


Also since _load_devices is the function being tested here, we shouldn't wrap it in another function. We normally use this _ensure_blah_blah() pattern in cases where we need the test graph to be pre-populated with other data but the actual function under test is something else.

tests/integration/cartography/intel/kandji/test_kandji.py
```diff
@@ -0,0 +1,27 @@
...
+        'f27bcd08-f653-4930-83b0-51970e923b98',
+    }
+    nodes = neo4j_session.run(
+        """
+        MATCH (d:KandjiDevice) RETURN d.device_id;
+        """,
+    )
+    actual_nodes = {n['d.device_id'] for n in nodes}
+    assert actual_nodes == expected_nodes
+
```
achantavy, line 22: We should also test if KandjiDevices are attached to other nodes using `check_rels()`. Example: https://github.com/lyft/cartography/blob/7601060cca2bd73105dfde059659f2d985958ad1/tests/integration/cartography/intel/aws/test_emr.py#L62-L69

cartography/intel/kandji/devices.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@timeit
+def _get_devices(kandji_base_uri: str, kandji_token: str) -> List[Dict]:
+    return call_kandji_api("/api/v1/devices", kandji_base_uri, kandji_token)
+
+
+@timeit
+def _load_devices(data: Dict, neo4j_session: neo4j.Session, update_tag: int) -> None:
+    ingest_kandji_device = """
```
achantavy, line 24: Instead of writing the query to ingest the data, you can use `cartography.client.core.tx.load()` on your `KandjiDeviceSchema`. Example: https://github.com/lyft/cartography/blob/11b53d5c79eb073c103109ab9edb987be6057eb9/cartography/intel/lastpass/users.py#L68-L81

We should also introduce the idea of a Kandji account or tenant because it's needed for cartography automatic cleanup jobs.

achantavy (changes requested):
Thank you for the PR! Appreciate you taking the time and effort to adopt cartography patterns used in other modules.

For this PR, let's focus on introducing a Kandji device node and a Kandji account/tenant node. Having a tenant-like node is needed for cartography automatic cleanup jobs to work so that the data in the graph is always up to date. You can copy this PR as a good example: https://github.com/lyft/cartography/pull/1083/files

Let's also defer adding the crowdstrike relationship and the analysis jobs to separate PRs.

Basically just copy that example PR and tag me again when this is ready. @ me on Slack if you need faster answers to fast questions; this way we can be better at the back and forth needed sometimes in these reviews.

Thanks again.

cartography/models/kandji/device.py
```diff
@@ -0,0 +1,48 @@
...
+    serial_number: PropertyRef = PropertyRef('serial_number', extra_index=True)
+
+
+@dataclass(frozen=True)
+class KandjiTenantToKandjiDeviceRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KandjiTenant)<-[:DEVICE]-(:KandjiDevice)
```
achantavy, line 33: Since the relationship label is called `DEVICE`, the convention is to point _to the device_. Like this:

```
(:KandjiTenant)-[:DEVICE]->(:KandjiDevice)
```

So you would do `LinkDirection.INWARD`.

docs/root/modules/kandji/schema.md
```diff
@@ -0,0 +1,33 @@
...
+|-------|-------------|
+| id | Kandji Tenant id|
+
+### KandjiDevice
+
+Representation of a Kandji device.
+
+|Field | Description|
+|-------|-------------|
+|device_id | Kandji device id|
```
achantavy, line 19: `id` here is `device_id`, right?

achantavy (changes requested):
Almost there, left a note about relationship direction convention. Once that is done, we can merge!

This is looking super clean btw

cartography/models/kandji/device.py
chandanchowdhury: Apologies for the confusion Alex!
The correct relationship is
```
(:KandjiDevice)-[:ENROLLED_TO]->(:KandjiTenant)
```
I had updated the doc but forgot to update the comment (fixed now).

Does the relationship makes sense now?

achantavy (approved):
Thank you for your PR! Appreciate you taking the time to use the patterns and adjust

chandanchowdhury:
Thank you @achantavy for reviewing the changes and walking me through the steps!

Issue #1016, opened 2022-10-30
setup.py install is deprecated


juju4:
**Bug Template**

*Title*: setup.py install is deprecated

*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

* https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
* https://peps.python.org/pep-0517/
* https://github.com/microsoft/LightGBM/issues/5061

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.
```
$ dpkg -l python3 python3-setuptools
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name               Version        Architecture Description
+++-==================-==============-============-=========================================================================
ii  python3            3.10.6-1~22.04 amd64        interactive high-level object-oriented language (default python3 version)
ii  python3-setuptools 59.6.0-1.2     all          Python3 Distutils Enhancements
$ git clone https://github.com/lyft/cartography.git
Cloning into 'cartography'...
remote: Enumerating objects: 11267, done.
remote: Total 11267 (delta 0), reused 0 (delta 0), pack-reused 11267
Receiving objects: 100% (11267/11267), 5.89 MiB | 17.33 MiB/s, done.
Resolving deltas: 100% (8068/8068), done.
$ cd cartography/
$ python3 setup.py install
running install
/usr/lib/python3/dist-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/usr/lib/python3/dist-packages/setuptools/command/easy_install.py:158: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/usr/lib/python3/dist-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: 0.1.43ubuntu1 is an invalid version and will not be supported in a future release
  warnings.warn(
/usr/lib/python3/dist-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: 1.1build1 is an invalid version and will not be supported in a future release
[...]

PR #1300, opened 2024-04-27
Switch to pyproject.toml based build

chandanchowdhury:
As mentioned in #1016  `setup.py` based builds are being deprecated in favor of `pyproject.toml` based builds.

Important notes:
* The build detects the package version dynamically using Git commit and tags, so had to remove `.git` from `.dockerignore` for the Python package build to succeed during Docker build.
* However, the Docker build using `Docker.multistage` is multi-stage. The `builder` stage builds the Python package from source code. The next `runner` stage installs the package built in `builder` stage.
* Changed `.github/workflows/test_suite.yml` to use the new `Dockerfile.multistage`. __Need to update `publish-to-ghcr.yml` when we are satisfied with the multi-stage Docker build__.
* Added `Pipfile` and `Pipfile.lock` in case someone wants to start using `pipenv` or we ourselves do so in future to keep our development environment clean.


chandanchowdhury:
Review request

@achantavy

.dockerignore
```diff
@@ -1,4 +1,10 @@
 build/
 *.egg-info/
 venv/
-.git/
+# .git/ # Required for pyproject based automatic version detection
```
heryxpc, line 4: Having the `.git` here was added due to a security finding (https://github.com/lyft/cartography/pull/1296).
Wondering if this can be changed to `.git/config` without breaking the build instead, as that file is the one with the issue.
A bit more context: 
* https://docs.github.com/en/actions/security-guides/automatic-token-authentication
* https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#stealing-the-jobs-github_token

chandanchowdhury: Hi @heryxpc thank you for the review.
Yes, that is a known issue and the risk is handled using multi stage docker build.

First stage `builder` gets all the contents but only uses them to build the python package which does not contain any of the Git files. 
The second stage `runner` only gets the build Python package and then install it. You can check the COPY command: https://github.com/lyft/cartography/pull/1300/files#diff-c66a06dd3419530dcc94979d2ffbd6afa551e1e0cdb4dacd3d9ccbea08d600bdR30

I went ahead and tried your suggestion of ignoring `.git/config`. The build passed locally.

chandanchowdhury: Github build has been successful too.

achantavy:
Closing this one since we migrated the pyproject.toml in https://github.com/cartography-cncf/cartography/pull/1311

PR #1301, opened 2024-05-01
File based Python logging config

### LINKED ISSUE #1213: Feature Request: Add time to logs/output
If I run Cartography from the terminal, I can see output for each step, but it's missing a timestamp. I can't quite see how long individual steps or indeed the overall run took without setting up a stats server. This can be solved by simply prepending the current timestamp when writing log output.

chandanchowdhury:
Will resolve #1213 along with supporting more customization.


chandanchowdhury:
Requesting review @achantavy

cartography/cli.py
```diff
@@ -575,6 +583,9 @@ def main(self, argv: str) -> int:
         # TODO support parameter lookup in environment variables if not present on command line
         config: argparse.Namespace = self.parser.parse_args(argv)
         # Logging config
+        if config.logging_config:
```
achantavy, line 586: Can you add a comment linking to the Python doc for this API? This way readers can know the correct format for a logging config file.

logging.cfg
```diff
@@ -0,0 +1,65 @@
+[logger_root]
```
achantavy, line 1: Add a comment indicating that this is a sample logging config file to be used with the `--logging-config` argument

chandanchowdhury:
Messed up the DCO fix, will open a new one.

PR #1303, opened 2024-05-10
GitHub intel updates to error handling

serge-wq:
## Summary

We've found constant errors in the GitHub module when the response data does not have the expected schema. These updates, while not always prevent the crash, they will provide more insight into what happened.

### DIFF
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -13,6 +13,7 @@ from typing import Tuple
 
 import requests
 
+
 logger = logging.getLogger(__name__)
 # Connect and read timeouts of 60 seconds each; see https://requests.readthedocs.io/en/master/user/advanced/#timeouts
 _TIMEOUT = (60, 60)
@@ -80,12 +81,12 @@ def call_github_api(query: str, variables: str, token: str, api_url: str) -> Dic
 
 
 def fetch_page(
-        token: str,
-        api_url: str,
-        organization: str,
-        query: str,
-        cursor: Optional[str] = None,
-        **kwargs: Any,
+            token: str,
+            api_url: str,
+            organization: str,
+            query: str,
+            cursor: Optional[str] = None,
+            **kwargs: Any,
 ) -> Dict[str, Any]:
     """
     Return a single page of max size 100 elements from the Github api_url using the given `query` and `cursor` params.
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -174,7 +174,7 @@ def fetch_all(
             logger.warning(
                 f'Got no "data" attribute in response: {resp}. '
                 f'Stopping requests for organization: {organization} and '
-                f'resource_type: {resource_type}'
+                f'resource_type: {resource_type}',
             )
             has_next_page = False
             continue
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -139,6 +139,7 @@ def fetch_all(
     """
     cursor = None
     has_next_page = True
+    org_data: Dict[str, Any] = {}
     data: PaginatedGraphqlData = PaginatedGraphqlData(nodes=[], edges=[])
     retry = 0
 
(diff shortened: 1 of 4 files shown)


cartography/intel/github/util.py
```diff
@@ -81,12 +80,12 @@ def call_github_api(query: str, variables: str, token: str, api_url: str) -> Dic
...
-        organization: str,
-        query: str,
-        cursor: Optional[str] = None,
-        **kwargs: Any,
+    token: str,
+    api_url: str,
+    organization: str,
+    query: str,
+    cursor: Optional[str] = None,
+    **kwargs: Any,
```
kledo-lyft, line 88: What's the current convention in this repo? 

Maybe we should set up `black` on this repo. 
Changes like this can confuse the git history.

cartography/intel/github/util.py
```diff
@@ -170,6 +170,15 @@ def fetch_all(
...
             continue
 
+        if 'data' not in resp:
+            logger.warning(
+                f'Got no "data" attribute in response: {resp}. '
+                f'Stopping requests for organization: {organization} and '
+                f'resource_type: {resource_type}',
+            )
+            has_next_page = False
+            continue
```
kledo-lyft, line 180: is this a known issue that could happen? It's like a key in a dictionary that we expect but doesn't exist; and that can happen in many other places. Wonder if we know more about this particular case to suspect it might happen here.

cartography/intel/github/util.py
```diff
@@ -180,6 +189,14 @@ def fetch_all(
 
         cursor = resource['pageInfo']['endCursor']
         has_next_page = resource['pageInfo']['hasNextPage']
-
-    org_data = {'url': resp['data']['organization']['url'], 'login': resp['data']['organization']['login']}
```
kledo-lyft, line 184: I assume the change here is that: in the past, this information is retrieved in the last. 
We change this to collect this info on every page if not already collected.

kledo-lyft (commented):
Are we able to use this version using the git commit hash without having to merge?

cartography/intel/github/util.py
heryxpc: `org_data` is mandatory to perform the full ingestion of Github info to Cartography: https://github.com/lyft/cartography/blob/fa1be115c80e83cefe0e31be5d738b5cda9a0887/cartography/intel/github/users.py#L116
I believe below code ensures that it exists on any page, otherwise it breaks the full process before starting ingestion.

cartography/intel/github/util.py
heryxpc: `data` is the root element from GH GraphQL API.
Unfortunately, there is no official documentation explicitly describing the response 🙃 .
But, in case it doesn't exists, the code cannot get any data to ingest.

heryxpc (dismissed):
LGTM.
Like @khanhldt mentioned, might be good to skip the formatting changes on this PR and see if we can put `black` formatter to the whole repo on a separate PR.

cartography/intel/github/util.py
serge-wq: As @heryxpc said, GrahpQL response usually contain `data` as the root attribute in the response which contains the actual requested data. This is a convention I've seen across GraphQL, not only for GitHub. Another thing that I've seen across GraphQL APIs is that if there is no data available, the attributes are returned as undefined, null or not even present. That's why this check here ill help us know what happened.

serge-wq: Also, this issue is what has been breaking the GitHub crons in our infra. As it's unclear in which page it happens, I'm just breaking out the loop gracefully and making sure we at least got the org_data.

serge-wq:
@heryxpc @khanhldt styling updates were applied by the linter if I remove them checks will fail. Can I get another look?

khanhldt (approved):
Some of the linter changes are quite arbitrary as I tested this locally.
I'm not sure why, but don't have much time to look into this.
--
We should really set up Black on this

heryxpc (approved):

PR #1307, opened 2024-06-01
Add support for Okta group->AWS SSO role rel

achantavy:
AWS SSO role names are weird because they look like `AWSReservedSSO_myrolename_<somehash>`. This caused our awssaml module to not draw links from Okta groups to these SSO roles correctly. This PR updates the module with the correct string comparisons to do this.

Screenshot showing that this works:
Screenshot 2024-06-03 at 1 49 03 PM


jfernandoz (approved):
💯

PR #1310, opened 2024-06-13
Fix NoneType error in describe_images by filtering out None values in…

austincomstockzoom:
This merge request fixes an issue where the `describe_images` function in `cartography/intel/aws/ec2/images.py` encountered a NoneType error. The error occurred because None values were being included in the image_ids list, which is not allowed by the AWS SDK. In one particular account of several hundred tested, an image ID was returned that was None. I was able to verify that this allows all of my accounts to run ingestion successfully.


achantavy:
Thanks for the fix, looks good to me, should be good to merge after you sign the CLA

austincomstockzoom:
Check again, thanks

achantavy:
Thank you!

achantavy (approved):

PR #1312, opened 2024-06-17
#1302: Refactor EC2 LaunchTemplate to use data model

### LINKED ISSUE #1302: EC2 launch template sync TransactionOutOfMemoryError
*Description*:

```
neo4j.exceptions.ClientError: {code: Neo.ClientError.General.TransactionOutOfMemoryError} {message: The allocation of an extra 152.7 MiB would use more than the limit 2.0 GiB. Currently using 1.9 GiB. dbms.memory.transaction.max_size threshold reached}
```

Crash in the EC2 launch template sync.

*To Reproduce*:

Have enough assets that the EC2 launch template transaction runs out of memory.

0.90.0

achantavy:
Fixes #1302: refactors EC2 launch template sync to use data model. This way, writes to the graph are automatically batched and write failures are retried.


achantavy:
cc: @heryxpc @chandanchowdhury

cartography/intel/aws/ec2/launch_templates.py
```diff
@@ -1,115 +1,134 @@
...
+
+
+def transform_launch_template_versions(versions: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    result: list[dict[str, Any]] = []
+    for version in versions:
+        current = version.copy()
+
+        # Reformat some fields
+        current['Id'] = f"{version['LaunchTemplateId']}-{version['VersionNumber']}"
+        current['CreateTime'] = str(int(version['CreateTime'].timestamp()))
```
heryxpc, line 78: Is `CreateTime` always available with an  ISO 8601 format or is it safer to use a `.get` and return a null value in case it's missing?

achantavy: Always available. 

Side note: neo4j datetime support has improved a lot since this module was first written so we don't even need this `str(int())` thing anymore, but for the sake of not breaking any interfaces and keeping the existing behavior I will defer that change to a future PR.

cartography/intel/aws/ec2/launch_templates.py
```diff
@@ -1,115 +1,134 @@
...
-            template_versions.extend(versions["LaunchTemplateVersions"])
-        template["_template_versions"] = template_versions
     return templates
 
 
+def transform_launch_templates(templates: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    result: list[dict[str, Any]] = []
+    for template in templates:
+        current = template.copy()
+        current['CreateTime'] = str(int(current['CreateTime'].timestamp()))
```
heryxpc, line 32: Is it safe to consider `CreateTime` a always present ISO 8601 format date or is it safer to use a `.get` and return a null value in case it's missing?

heryxpc (approved):
LGTM, left a nit comment, no rush to apply it.

Issue #1316, opened 2024-06-17
AWS Subnets aren't linked to AWS VPCs on import


nelsonda:
*Title*: AWS subnets aren't linked to AWS VPCs on import

*Description*:
After running a Cartography against a fresh neo4j instance, VPCs aren't connected to subnets and vice versa.

*To Reproduce*:
Run cartography with an AWS account with VPCs & subnets. Present in current master branch: [098d8ca](https://github.com/lyft/cartography/commit/098d8ca5f4bb172944338dad9df797a36e23485a)

*Logs*:
N/A 

*Screenshots*:

![Screenshot 2024-06-17 at 12 35 02 PM](https://github.com/lyft/cartography/assets/930653/aa8d378d-0695-4c20-b9d4-591bb78ae46d)
![Screenshot 2024-06-17 at 12 39 08 PM](https://github.com/lyft/cartography/assets/930653/875fffee-cba4-4cb0-8f3e-0795a51888b7)

*Please complete the following information:*:
- Cartography release version or commit hash: [098d8ca](https://github.com/lyft/cartography/commit/098d8ca5f4bb172944338dad9df797a36e23485a)
- Python version: 3.12.3
- OS (feel free to omit this if you don't think it's relevant to your issue): OSX (but probably not relevant!)

PR #1318, opened 2024-06-18
For bug #1316, removed an underscore from AWS subnet's datamodel.

nelsonda:


achantavy:
Thanks for the help and for this PR, as discussed I finished this in #1320

PR #1319, opened 2024-06-18
Handle flaky CI failures in integration tests

achantavy:
See https://github.com/neo4j/neo4j/issues/13404 for context.

The neo4j 4.x container in our CI occasionally fails integration tests because we create indexes quickly without giving the database enough time for those actions to settle (roughly speaking).

This PR uses pytest-rerunfailures to retry flaky tests so that we don't need to manually re-run them.

Credit to @heryxpc for figuring this out.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -69,3 +69,8 @@ ignore_errors = true
 
 [coverage:report]
 fail_under = 30
+
+[tool:pytest]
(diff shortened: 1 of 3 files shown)


achantavy:
cc: @chandanchowdhury FYI this should fix the CI problems

heryxpc (approved):

PR #1324, opened 2024-06-28
Adding token generation before each service API

soumyadipDe:
Fix for https://github.com/lyft/cartography/issues/1323


chandanchowdhury:
I wonder if the solution would simpler if we

* Move all `_get_<Type>_resource` to their individual files. Example: `_get_compute_resource` into `cartography/intel/gcp/compute.py`
* Pass `credentials` along with `project_id` to all type specific code so it can do everything necessary

Basically, `__init__.py`, finds all projects and then for each project triggers type specific code which does the necessary work and handle all situations.

This way 
* we avoid needing to add `_sync_single_project_<Type>` in `__init__.py`
* Make testing each type simple as the individual code takes care of everything related to a project and type

Hope I have not confused things.

soumyadipDe:
I tried the suggestion but as feared it ended up with many changes to the code. Creating another PR to check - https://github.com/lyft/cartography/pull/1325

This is still raw and I need to test the change, but wanted to get early review if this is a desired approach to this.

chandanchowdhury:
@soumyadipDe just checking if you had a chance to test the changes? It looks good to me.

chandanchowdhury (approved):

PR #1157, opened 2023-04-12
Create pull_request_template.md

ramonpetgrave64:
Adds some guidance on how to edit/create intel modules

PR #1331, opened 2024-07-11
AWSmappingfix-ELBv1&v2<--ENI

udm17:
Some testing lead me to find that the mapping between Loadbalancer V1 and V2 was not happening correctly and in response to that, I delved into the code for the mapping. 

1. The regex used to distinguish between V1 and V2 had the variable assignment inverted ie v1's were getting assigned to elbv2_id attribute and vice versa 

2. Creating a dynamic dnsname and then matching for LoadbalancerV2 was problematic as well since the dynamically created names were always partial matches for the dnsname. The elb_match[2] attribute was never right to match to the correct node. (the regex returns the id of the loadbalancer while the dnsname always has a different value)

created using carto -> elb_match[1]-elb_match[2].elb.region.amazonaws.com
actual dnsname -> elb_match[1]-elb_match[2]-9574787.region.elb.amazonaws.com

Fix -> 

1. Swapping about the variables lead to successful mapping between LBv1 and the Network interfaces 

2. For LBV2, rather than create a dns name, I suggest using a combination of name and region for the loadbalancerv2 as the combination needs to be unique across the AWS account.

### DIFF
--- a/cartography/intel/aws/ec2/network_interfaces.py
+++ b/cartography/intel/aws/ec2/network_interfaces.py
@@ -54,11 +54,11 @@ def transform_network_interface_data(data_list: List[Dict[str, Any]], region: st
         elb_v2_id = None
         elb_match = re.match(r'^ELB (?:net|app)/([^\/]+)\/(.*)', network_interface.get('Description', ''))
         if elb_match:
-            elb_v1_id = f'{elb_match[1]}-{elb_match[2]}.elb.{region}.amazonaws.com'
+            elb_v2_id = elb_match[1]
         else:
             elb_match = re.match(r'^ELB (.*)', network_interface.get('Description', ''))
             if elb_match:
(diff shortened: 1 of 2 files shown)


chandanchowdhury:
Hey @udm17, thank you for the PR :)

Does #996 explain/fix the issue you found?

If not, do you mind opening an Issue and attach that to this PR (to follow the process being established by Ramon via #1157).

achantavy:
@udm17, thanks for digging and for the PR. It's been a while since I've looked at the ELB code.

It'd be ideal to write a test that catches this problem, but I also realize this is an old-ish part of the code where it'd probably be faster for me to add those than someone else at this time.

So, in place of that, would you be able to please post screenshots of what the relevant nodes and rels looked like before your change and after your change? This way it's more obvious to see the bug.

Issue #1321, opened 2024-06-21
Sync crashes when we try to describe a launch template version on a template that no longer exists


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

Sync crashes when calling describe launch template versions because we try to describe a template that may no longer exist.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run a sync on an account where launch templates are being spun up and down (I think).

Proposed fix: add error handling to skip over these cases.

*Logs*:
> If applicable, copy and paste your console log with the failing stack trace.

```
Traceback (most recent call last):
  File "cartography/sync.py", line 191, in run_with_config
    return sync.run(neo4j_driver, config)
  File "cartography/sync.py", line 113, in run
    stage_func(neo4j_session, config)
  File "cartography/util.py", line 192, in timed
    result = method(*args, **kwargs)
  File "cartography/intel/aws/__init__.py", line 298, in start_aws_ingestion
    sync_successful = _sync_multiple_accounts(
  File "cartography/intel/aws/__init__.py", line 196, in _sync_multiple_accounts
    raise Exception('\n'.join(exception_tracebacks))
Exception: 2024-06-20 16:17:40.827071 - Exception for account ID: XXX
Traceback (most recent call last):
  File "cartography/intel/aws/__init__.py", line 169, in _sync_multiple_accounts
    _sync_one_account(
  File "cartography/intel/aws/__init__.py", line 64, in _sync_one_account
    RESOURCE_FUNCTIONS[func_name](**sync_args)
  File "cartography/util.py", line 192, in timed
    result = method(*args, **kwargs)
  File "cartography/intel/aws/ec2/launch_templates.py", line 143, in sync_ec2_launch_templates
    versions = get_launch_template_versions(boto3_session, templates, region)
  File "cartography/util.py", line 192, in timed
    result = method(*args, **kwargs)
  File "backoff/_sync.py", line 105, in retry
    ret = target(*args, **kwargs)
  File "cartography/util.py", line 279, in inner_function
    return func(*args, **kwargs)
  File "cartography/intel/aws/ec2/launch_templates.py", line 67, in get_launch_template_versions
    for versions in v_paginator.paginate(LaunchTemplateId=template['LaunchTemplateId']):
  File "botocore/paginate.py", line 269, in __iter__
[...]

PR #1332, opened 2024-07-17
#1321: Refactor get_launch_template_versions to skip race condition.

heryxpc:
See https://github.com/lyft/cartography/issues/1321

### DIFF
--- a/cartography/intel/aws/ec2/launch_templates.py
+++ b/cartography/intel/aws/ec2/launch_templates.py
@@ -28,7 +28,7 @@ def get_launch_templates(boto3_session: boto3.session.Session, region: str) -> l
                 template_versions = get_launch_template_versions_by_template(boto3_session, template, region)
             except Exception as e:
                 logger.warning(
-                    f"Failed to get launch template versions for {template['LaunchTemplateId']}",
+                    f"Failed to get launch template versions for {template['LaunchTemplateId']}: {e}",
                     exc_info=True,
                 )
                 template_versions = []
--- a/cartography/intel/aws/ec2/launch_templates.py
+++ b/cartography/intel/aws/ec2/launch_templates.py
@@ -3,6 +3,7 @@ from typing import Any
 
 import boto3
 import neo4j
+from botocore.exceptions import ClientError
 
 from .util import get_botocore_config
 from cartography.client.core.tx import load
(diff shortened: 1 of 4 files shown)


cartography/intel/aws/ec2/launch_templates.py
```diff
@@ -22,7 +22,19 @@ def get_launch_templates(boto3_session: boto3.session.Session, region: str) -> l
     paginator = client.get_paginator('describe_launch_templates')
     templates: list[dict[str, Any]] = []
     for page in paginator.paginate():
-        templates.extend(page['LaunchTemplates'])
+        paginated_templates = page['LaunchTemplates']
+        for template in paginated_templates:
+            try:
```
achantavy, line 27: 2 options:

Option 1. **Simplest**: I think we should remove the try+catch here. The previous version of the code (https://github.com/lyft/cartography/blob/21bd5a379bb5a992d6e540c4b70c9228fa2f15c4/cartography/intel/aws/ec2/launch_templates.py#L24-L29) did not have the try+catch and did not have the race condition (that I introduced :x).

The only situation where we should be catching the base `Exception` is in "top level" orchestrator code. There are too many things that can go wrong 

Option 2. **more time-consuming**: catch the `botocore.exceptions.ClientError` and check specifically for that `InvalidLaunchTemplateId.NotFound` error code. I started on this briefly but got frustrated when I needed to add in boto3-stubs as a dependency to properly unit test it. Boto3 does all kinds of dynamic metaprogramming stuff that make it hard to unit test.

Given that I don't recall Ryan's old code throwing the race condition, I think removing the try+catch is best. We probably don't have that race condition anymore because the call to retrieve template versions happens closer to the call to retrieve templates.

cartography/intel/aws/ec2/launch_templates.py
```diff
@@ -55,20 +67,24 @@ def load_launch_templates(
 
 @timeit
 @aws_handle_regions
-def get_launch_template_versions(
+def get_launch_template_versions_by_template(
         boto3_session: boto3.session.Session,
-        templates: list[dict[str, Any]],
+        template: dict[str, Any],
```
achantavy, line 72: non-block: consider passing in the template_id as a str to this function since I don't think the whole object is necessary.

cartography/intel/aws/ec2/launch_templates.py
```diff
@@ -22,7 +22,19 @@ def get_launch_templates(boto3_session: boto3.session.Session, region: str) -> l
...
+            try:
+                template_versions = get_launch_template_versions_by_template(boto3_session, template, region)
+            except Exception as e:
+                logger.warning(
+                    f"Failed to get launch template versions for {template['LaunchTemplateId']}: {e}",
+                    exc_info=True,
+                )
+                template_versions = []
+            # Using a key not defined in latest boto3 documentation
+            template['GetVersions'] = template_versions
```
achantavy, line 36: Rather than tack on this in a field on the template, how about we let `get_launch_templates` return a tuple of 2 things: 

1. the list of templates
2. the list of template versions

I realize that the original code did it this way, but I think it's cleaner to have multiple return values in this case. There's no need to have the version data embedded in the template dict itself. 

By doing this we would also no longer need the `get_launch_template_versions` below (the one that just goes through each template and grabs each version and puts it in a list).

achantavy (changes requested):

cartography/intel/aws/ec2/launch_templates.py
heryxpc: Like we discussed, we need the `Exception` in cases where the environment is so dynamic that it can create and destroy launch templates in milliseconds.
I can catch only the `botocore.exceptions.ClientError`, but that is too specific to https://github.com/lyft/cartography/issues/1321, I would prefer to keep it simple and manage all sort of scenarios skipping the launch templates not found.

achantavy: Ah, in our convo I mentioned that I am ok with the try+except, but we shouldn't catch the base `Exception`. There are other things that can fail in the code block and we'd run the risk of burying other things.

achantavy (approved):

achantavy (approved):

PR #1336, opened 2024-07-22
Fix #1334: add retry + sleep to github team-repo sync

### LINKED ISSUE #1334: github sync nonetype error when listing repo-teams
**Bug Template**

*Description*:

GitHub sync crashes when listing repo-teams with a Nonetype error;

*To Reproduce*:

Not sure why this suddenly started failing.

*Logs*:

Crash:
```
  File "PATH/cartography/intel/github/teams.py", line 174, in sync_github_teams
    team_repos = _get_team_repos_for_multiple_teams(teams_paginated.nodes, organization, github_url, github_api_key)
  File "PATH/cartography/util.py", line 192, in timed
  File "PATH/cartography/intel/github/teams.py", line 63, in _get_team_repos_for_multiple_teams
    repo_urls = [t['url'] for t in team_repos.nodes] if team_repos.nodes else []
  File "PATH/cartography/intel/github/teams.py", line 63, in <listcomp>
    repo_urls = [t['url'] for t in team_repos.nodes] if team_repos.nodes else []
```

0.93.0

achantavy:
Adds a retry with sleep to `_get_team_repos_for_multiple_teams` because we have seen the GitHub GraphQL API sometimes return None for fields that are not supposed to be None.

- https://github.com/lyft/cartography/issues/1334

Added unit tests.

### Checklist

- [x] Update/add unit or integration tests

If you are modifying or implementing an intel module:
- [ ] Update the [schema](https://github.com/lyft/cartography/tree/master/docs/root/modules) and [readme](https://github.com/lyft/cartography/blob/master/docs/schema/README.md)
- [ ] Use the NodeSchema [data model](https://lyft.github.io/cartography/dev/writing-intel-modules.html#defining-a-node)

### DIFF
--- a/cartography/intel/github/teams.py
+++ b/cartography/intel/github/teams.py
@@ -49,7 +49,7 @@ def get_teams(org: str, api_url: str, token: str) -> Tuple[PaginatedGraphqlData,
 
 @timeit
 def _get_team_repos_for_multiple_teams(
-        team_raw_data: List[Dict[str, Any]],
+        team_raw_data: list[dict[str, Any]],
         org: str,
         api_url: str,
         token: str,
--- a/cartography/intel/github/teams.py
+++ b/cartography/intel/github/teams.py
@@ -53,15 +53,15 @@ def _get_team_repos_for_multiple_teams(
         org: str,
         api_url: str,
         token: str,
-) -> dict[str, set[RepoPermission]]:
-    result: dict[str, set[RepoPermission]] = {}
+) -> dict[str, list[RepoPermission]]:
+    result: dict[str, list[RepoPermission]] = {}
     for team in team_raw_data:
         team_name = team['slug']
         repo_count = team['repositories']['totalCount']
 
         if repo_count == 0:
             # This team has access to no repos so let's move on
-            result[team_name] = set()
+            result[team_name] = []
             continue
 
         repo_urls = []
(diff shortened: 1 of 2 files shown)


cartography/intel/github/teams.py
```diff
@@ -45,26 +49,53 @@ def get_teams(org: str, api_url: str, token: str) -> Tuple[PaginatedGraphqlData,
...
+
+        max_tries = 5
+
+        for current_try in range(1, max_tries + 1):
+            team_repos = _get_team_repos(org, api_url, token, team_name)
+
+            try:
+                # The `or []` is because `.nodes` can be None. See:
+                # https://docs.github.com/en/graphql/reference/objects#teamrepositoryconnection
+                for repo in team_repos.nodes or []:
```
heryxpc, line 78: Trying to understand the Python magic here.
If `team_repos.nodes == None` then the for condition will break before assigning any element from the iterable to `repo`?

heryxpc (dismissed):

cartography/intel/github/teams.py
achantavy: If `team_repos.nodes` is None, then the for loop will iterate through the empty list. See this:

```python
>>> derp = {'a': [1,2,3], b: None}
>>> for r in derp['a'] or []:
...   print(r)
...
1
2
3
>>> for r in derp['b'] or []:
...   print(r)
...
# prints nothing
```

heryxpc (approved):

PR #1342, opened 2024-07-25
Update Semgrep intel module to use the /findings API endpoint

heryxpc:
### Summary
Semgrep is consolidating their API  for both SCA and SAST findings: https://semgrep.dev/api/v1/docs/#tag/Finding/operation/semgrep_app.core_exp.findings.handlers.issue.openapi_list_recent_issues deprecating the one we use https://semgrep.dev/api/v1/deployments/1111/ssc-vulns.

This replaces the call to the new API, updates the mapping based on the new info retrieved, adds some guard rails to prevent issues during data ingestion and tackles read time-outs to the new API.

### Related issues or links
None

### Before the change
Screenshot 2024-07-24 at 7 21 42 p m

### After the change
Screenshot 2024-07-24 at 7 22 35 p m

### Log trace
Screenshot 2024-07-24 at 7 25 19 p m

### Checklist

- [x ] Include console log trace showing what happened before and after your changes.

- [ ] Use the NodeSchema [data model](https://lyft.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).

### DIFF
--- a/cartography/data/jobs/scoped_analysis/semgrep_sca_risk_analysis.json
+++ b/cartography/data/jobs/scoped_analysis/semgrep_sca_risk_analysis.json
@@ -13,47 +13,47 @@
         },
         {
             "__comment__": "not possible to identify if reachable && version specifier is the only flag of the vulnerability (likelihood = rare) && severity in [low, medium, high] -> Risk = Info",
-            "query": "MATCH (g:GitHubRepository{archived:false})<-[:FOUND_IN]-(s:SemgrepSCAFinding{reachability:'UNKNOWN_EXPOSURE', reachability_check:'VERSION_SPECIFIER', lastupdated:$UPDATE_TAG})<-[:RESOURCE]-(:SemgrepDeployment{id:$DEPLOYMENT_ID}) WHERE s.severity IN ['LOW', 'MEDIUM', 'HIGH'] SET s.reachability_risk = 'INFO' return COUNT(*) as TotalCompleted",
+            "query": "MATCH (g:GitHubRepository{archived:false})<-[:FOUND_IN]-(s:SemgrepSCAFinding{reachability:'UNREACHABLE', reachability_check:'NO REACHABILITY ANALYSIS', lastupdated:$UPDATE_TAG})<-[:RESOURCE]-(:SemgrepDeployment{id:$DEPLOYMENT_ID}) WHERE s.severity IN ['LOW', 'MEDIUM', 'HIGH'] SET s.reachability_risk = 'INFO' return COUNT(*) as TotalCompleted",
             "iterative": false
         },
         {
             "__comment__": "not possible to identify if reachable && version specifier is the only flag of the vulnerability (likelihood = rare) && severity = critical -> Risk = Low",
-            "query": "MATCH (g:GitHubRepository{archived:false})<-[:FOUND_IN]-(s:SemgrepSCAFinding{reachability:'UNKNOWN_EXPOSURE', reachability_check:'VERSION_SPECIFIER', lastupdated:$UPDATE_TAG})<-[:RESOURCE]-(:SemgrepDeployment{id:$DEPLOYMENT_ID}) WHERE s.severity = 'CRITICAL' SET s.reachability_risk = 'LOW' return COUNT(*) as TotalCompleted",
+            "query": "MATCH (g:GitHubRepository{archived:false})<-[:FOUND_IN]-(s:SemgrepSCAFinding{reachability:'UNREACHABLE', reachability_check:'NO REACHABILITY ANALYSIS', lastupdated:$UPDATE_TAG})<-[:RESOURCE]-(:SemgrepDeployment{id:$DEPLOYMENT_ID}) WHERE s.severity = 'CRITICAL' SET s.reachability_risk = 'LOW' return COUNT(*) as TotalCompleted",
             "iterative": false
         },
         {
-            "__comment__": "manual review required to confirm && version specifier is the only flag of the vulnerability (likelihood = possible) && severity in [low, medium] -> Risk = Low",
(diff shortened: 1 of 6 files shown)


cartography/data/jobs/scoped_analysis/semgrep_sca_risk_analysis.json
heryxpc, line 1: This is mentioned in a comment, but findings types are documented at https://semgrep.dev/docs/semgrep-supply-chain/overview#types-of-semgrep-supply-chain-findings

achantavy (approved):
If we haven't already, should show this off to Semgrep too. Maybe at some point they'll release an official Semgrep module (-:

cartography/intel/semgrep/findings.py
```diff
@@ -110,48 +143,64 @@ def transform_sca_vulns(raw_vulns: List[Dict[str, Any]]) -> Tuple[List[Dict[str,
...
+                usage_dict["findingId"] = hash(url.split("github.com/")[-1])
+                usage_dict["path"] = usage["location"]["path"]
+                usage_dict["startLine"] = usage["location"]["start_line"]
+                usage_dict["startCol"] = usage["location"]["start_col"]
+                usage_dict["endLine"] = usage["location"]["end_line"]
+                usage_dict["endCol"] = usage["location"]["end_col"]
+                usage_dict["url"] = url
+                usages.append(usage_dict)
+            vulns.append(sca_vuln)
+        except Exception as e:
```
achantavy, line 201: You're ok catching on anything and swallowing data errors? 

Something to consider: If the data ends up malformed the sync will continue

achantavy (approved):

cartography/intel/semgrep/findings.py
```diff
@@ -48,60 +50,91 @@ def get_deployment(semgrep_app_token: str) -> Dict[str, Any]:
...
 
+    logger.info(f"Retrieved {len(all_vulns)} Semgrep SCA vulns in {page} pages.")
     return all_vulns
 
 
+def _get_vuln_class(vuln: Dict) -> str:
+    vulnerability_classes = vuln["rule"].get("vulnerability_classes", [])
+    if vulnerability_classes:
+        return vulnerability_classes[0]
+    return "Other"
```
kledo-lyft, line 107: any particular reason / concern with only getting the first vuln class out of potentially many?

cartography/intel/semgrep/findings.py
```diff
@@ -48,60 +50,91 @@ def get_deployment(semgrep_app_token: str) -> Dict[str, Any]:
...
+        else:
+            return "UNREACHABLE"
+    return None
+
+
+def _build_vuln_url(vuln: str) -> str | None:
+    if 'CVE' in vuln:
+        return f"https://nvd.nist.gov/vuln/detail/{vuln}"
+    if 'GHSA' in vuln:
+        return f"https://github.com/advisories/{vuln}"
```
kledo-lyft, line 134: nit: could key casing be a concern?

cartography/intel/semgrep/findings.py
```diff
@@ -110,48 +143,64 @@ def transform_sca_vulns(raw_vulns: List[Dict[str, Any]]) -> Tuple[List[Dict[str,
...
+                usage_dict["startLine"] = usage["location"]["start_line"]
+                usage_dict["startCol"] = usage["location"]["start_col"]
+                usage_dict["endLine"] = usage["location"]["end_line"]
+                usage_dict["endCol"] = usage["location"]["end_col"]
+                usage_dict["url"] = url
+                usages.append(usage_dict)
+            vulns.append(sca_vuln)
+        except KeyError as e:
+            logger.warning(f"Error transforming Semgrep SCA vuln {vuln}: {e}")
+            continue
```
kledo-lyft, line 203: are all the keys required? Why is it a warning log? 

like, if we look at the results in Cartography and it might not be obvious that something might have failed, so how can we get notified and debug?

heryxpc:
> If we haven't already, should show this off to Semgrep too. Maybe at some point they'll release an official Semgrep module (-:

I shared it with them when first released, but there haven't been much engagement.
There aren't many integrations with the Semgrep API that I found at Github, so maybe we can promote Cartography + NeoDash as a good tool to surface and report their findings.

cartography/intel/semgrep/findings.py
heryxpc: I'm only getting the first to build a summary of the finding using the package + class.
In the past they retrieved this summary in the API, but the attribute is gone in the new version.
I tried to match what Semgrep UI displays as title for a particular finding, like:
> gevent: Improper Privilege Management

cartography/intel/semgrep/findings.py
heryxpc: We are upper casing the value we pass to this function:
```
        if vuln.get("vulnerability_identifier"):
              vuln_id = vuln["vulnerability_identifier"].upper()
              sca_vuln["cveId"] = vuln_id
              sca_vuln["ref_urls"] = [_build_vuln_url(vuln_id)]
```
If we use this function more generic, we could consider upper casing the input here.

cartography/intel/semgrep/findings.py
heryxpc: I know, I don't feel great about it but it mitigates the problem where there are data issues coming from the API and it doesn't block the full ingestion for such problems.

What would you suggest to tackle the problem when the API is missing information we have as mandatory to ingest (like reachability, repository or severity)?

kledo-lyft: what about using `error` log instead? So the ingestion doesn't get blocked and we can still get notified?

if there are data that are optional, the decision about what to do them might need to be a little more intentional
e.g. if missing reachability, just default to `unknown` may be?

PR #1358, opened 2024-09-24
Add docs on cartography in production and what it is and is not

achantavy:
Adds documentation on 
- what cartography is and isn't
- how to build prod applications around cartography
- clarify which steps are test machine steps vs production setup steps


chandanchowdhury (approved):
Thank you Alex :)

PR #1359, opened 2024-09-25
Fix broken links

achantavy:
Fixes broken links in the docs.


docs/root/modules/index.rst
```diff
@@ -1,5 +1,4 @@
 .. toctree::
-    :hidden:
```
achantavy, line 2: Removing this so that the table of contents shows up in the rendered modules github page

chandanchowdhury (approved):
Thank you Alex!

Issue #1363, opened 2024-09-26
Dockerfile --local can only be used inside a git repository


hatchetation:
**Dockerfile --local can only be used inside a git repository**

This is a clone of closed issue #1347, opened by @kmshar . 

I believe the proximate cause may be the security-related change introduced in #1296 . 

*Description*:
> Fails to build with default configuration.

*To Reproduce*:
> 1. Update `Dockerfile` to use `ubuntu:jammy` to resolve an unrelated problem with `python3.10-dev` being unavailable in `ubuntu:focal`
> 2. Build docker using `docker build -t lyft/cartography .`

*Logs*:
```
 => [ 9/10] RUN git config --global --add safe.directory /srv/cartography                                                         0.1s
 => ERROR [10/10] RUN /usr/bin/git config --local user.name "cartography"                                                         0.1s
------
 > [10/10] RUN /usr/bin/git config --local user.name "cartography":
0.123 fatal: --local can only be used inside a git repository
------
Dockerfile:33
--------------------
  31 |     # and the user that is going to run the unit&integ tests.
  32 |     RUN git config --global --add safe.directory /srv/cartography
  33 | >>> RUN /usr/bin/git config --local user.name "cartography"
  34 |
--------------------
ERROR: failed to solve: process "/bin/sh -c /usr/bin/git config --local user.name \"cartography\"" did not complete successfully: exit code: 128
```

>  - Cartography commit hash 2420ef65a1327025931edaa20a26a58f53651e55
>  - Python version: [3.10]
>  - OS: Docker on MacOS

chandanchowdhury:
Background: https://github.com/lyft/cartography/blob/master/.github/workflows/publish-to-ghcr.yml#L47

@achantavy I do not know the history, so need your input on if we should replace https://github.com/lyft/cartography/blob/master/Dockerfile with https://github.com/lyft/cartography/blob/master/dist.Dockerfile ?

achantavy:
Hey @hatchetation some clarifying questions - are you trying to spin up cartography in production or is this for playing around on a test machine?

For my own cartography testing I don't use docker, I create venvs and `pip install` things to it, so tbh I'm unfamiliar with how Docker is normally used in OSS Python projects or what are the expected ergonomics/scenarios/outcomes, but I do want to make this as smooth as possible - let me know as many details as you can.

If it's easier I have a thread on slack here: https://lyftoss.slack.com/archives/CTZUQL0KX/p1727375916891719, either way works.

[...]

PR #1364, opened 2024-10-02
#1363, #1347: Make docker-compose the preferred way to try cartography

### LINKED ISSUE #1347: Dockerfile --local can only be used inside a git repository
**Dockerfile --local can only be used inside a git repository**

*Description*:
>  Fails to build with default configuration.

*To Reproduce*:
> Build docker using `docker build -t lyft/cartography .`

*Logs*:

```
=> ERROR [10/10] RUN /usr/bin/git config --local user.name "cartography"                                                                                                         0.2s
------
 > [10/10] RUN /usr/bin/git config --local user.name "cartography":
0.203 fatal: --local can only be used inside a git repository
------
Dockerfile:33
--------------------
  31 |     # and the user that is going to run the unit&integ tests.
  32 |     RUN git config --global --add safe.directory /srv/cartography
  33 | >>> RUN /usr/bin/git config --local user.name "cartography"
  34 |
--------------------
ERROR: failed to solve: process "/bin/sh -c /usr/bin/git config --local user.name \"cartography\"" did not complete successfully: exit code: 128

```

*Screenshots*:

cartography-docker-failed-build

achantavy:
Addresses #1363 and #1347.
- Updates all relevant documentation to prefer running docker-compose for end-users.
- Provides documentation on how to use docker and docker-compose for developers.

### Checklist

#### Proof that this works
- Running an AWS sync using docker-compose
docker-compose-run

  CLI used:
  ```
docker-compose run -e AWS_PROFILE=1234_testprofile -e AWS_DEFAULT_REGION=us-east-1 cartography --neo4j-uri bolt://cartography-neo4j-1:7687
  ```


achantavy:
This still needs more work, Docker is weird.

achantavy:
cc: @hatchetation -- I think this addresses the doc bugs you brought up. I think it makes sense to have docker-compose be the main way to try out cartography.

Dockerfile
```diff
@@ -1,33 +1,20 @@
...
+# (https://github.com/hexops/dockerfile#do-not-use-a-uid-below-10000).
+ARG uid=10001
+ARG gid=10001
 
-ENV PATH=/venv/bin:$PATH
-RUN apt-get update && \
-    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends python3.10-dev python3-pip python3-setuptools openssl libssl-dev gcc pkg-config libffi-dev libxml2-dev libxmlsec1-dev curl make git && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+COPY . /var/cartography
```
chandanchowdhury, line 8: For Production docker image, we must install Cartography from PyPI.

Dockerfile
```diff
@@ -1,33 +1,20 @@
...
+WORKDIR /var/cartography
+ENV HOME=/var/cartography
 
-# Installs pip supported by python3.10
-RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3.10 get-pip.py
+RUN pip install -U -e .
```
chandanchowdhury, line 12: Production install should not use `--editable`.

chandanchowdhury (changes requested):

Dockerfile
achantavy: This is the way it already is right now though: https://github.com/lyft/cartography/blob/master/dist.Dockerfile

The other one that I removed is this way too: https://github.com/lyft/cartography/blob/6a6e2778204938cbd51add75827f7aaacef31526/Dockerfile#L20

achantavy: Are you saying that we should make it 

```Dockerfile
RUN pip install .
```

instead? I think that works.

chandanchowdhury: ```pip install cartography```

Docker steps should be the exact commands that we will use to setup a production environment on a bare Linux host, it is just that the image has everything the environment made up of such as Python, Cartography package installed with all its production dependencies only (no package that are required for dev/test).

achantavy: Mm, maybe I will add back the dev.Dockerfile then.

dev.Dockerfile will run an editable install

Dockerfile will install the latest release

.dockerignore
```diff
@@ -1,4 +1,3 @@
 build/
 *.egg-info/
 venv/
-.git/
```
achantavy, line 4: This is ok to remove because the Dockerfile that gets pushed to ghcr.io no longer copies in the current working directory. Rather, it does a pip install.

docker-compose.yml
```diff
@@ -33,19 +38,39 @@ services:
...
+    environment:
+      # Point to the neo4j service defined in this docker-compose file.
+      - NEO4J_URL=bolt://cartography-neo4j-1:7687
+
+# Intended to run local automated tests, custom sync scripts, and local changes.
+  cartography-dev:
+    # See dev instructions: we assume that you have built this with
+    # `docker build -t lyft/cartography-dev . -f dev.Dockerfile`.
+    # Do not push this image remotely!
+    image: lyft/cartography-dev
```
chandanchowdhury, line 63: Not a blocker.
You can add [build](https://docs.docker.com/reference/compose-file/build/#illustrative-example) to automatically build a new image (if code changed) during start-up.

chandanchowdhury (approved):

Issue #990, opened 2022-09-19
AWS SSO Identity Store records not currently read


jcmadick:
### Discussed in https://github.com/lyft/cartography/discussions/930

<div type='discussions-op-text'>

<sup>Originally posted by **jcmadick** September  9, 2022</sup>
AWS has released an update to AWS SSO (now IAM Identity Center) that exposes access to users using SSO for AWS access.

https://docs.aws.amazon.com/singlesignon/latest/userguide/identities.html

These users and groups are not included in the normal AWS Security Audit permissions and are not pulled through the current cartography code.  In my case, most of the relationships I'm interested in analyzing are tied to the SSO users and groups.  I have pulled a list of users and a list of groups from our AWS SSO.  I do not see anything in the fields to create or tie relationships.  I've attached a sanitized copy of a user record and a group record

I had to have permissions added to my sp to allow read access to users and groups in the identity store (unique from the normal identity store.)  From AWSCLI, the command is "aws identitystore list-users --identity-store-id d-*******"  There are other list options including groups and group membership.

I don't think integrating the pulls would be a challenge, but finding the relationships is where I'm having challenges.</div>
[AWS_SSO_Group_json.txt](https://github.com/lyft/cartography/files/9601283/AWS_SSO_Group_json.txt)
[AWS_SSO_User_json.txt](https://github.com/lyft/cartography/files/9601284/AWS_SSO_User_json.txt)

achantavy:
Somewhat related, but we now map AWS SSO roles to Okta groups after https://github.com/lyft/cartography/pull/1307

kunaals:
We now have identity center groups and represent all object types within Identity Center! Closing this out

PR #1380, opened 2024-11-10
Added in Identity Center Nodes + relationship mapping

SecPrez:
**Summary**
Mapped in [AWS Identity Center](https://aws.amazon.com/iam/identity-center/) and the access it provides to AWS accounts.
New Nodes: (AWSIdentityCenter), (AWSPermissionSet), (AWSSSOUser)
New Relationships:
(AWSAccount)-[RESOURCE]->(AWSIdentityCenter)
(AWSIdentityCenter)-[HAS_PERMISSION_SET]->(AWSPermissionSet)
(AWSSSOUser)<-[ALLOWED_BY]-(AWSRole)
(OktaUser)<-[CAN_ASSUME_IDENTITY]-(AWSSSOUser)
(AWSPermissionSet)-[ASSIGNED_TO_ROLE]->(AWSRole)

image
image

**Console Trace**
INFO:cartography.intel.aws.identitycenter:Syncing Identity Center instances for region us-east-1 INFO:cartography.intel.aws.identitycenter:Loading 1 Identity Center instances for region us-east-1 INFO:cartography.intel.aws.identitycenter:Loading 32 permission sets for instance arn:aws:sso:::instance/ssoins-72237a0dcb8c6df7 in region us-east-1 INFO:cartography.intel.aws.identitycenter:Loading 777 permission set role assignments INFO:cartography.intel.aws.identitycenter:Loading 803 SSO users for identity store d-906747a0b9 in region us-east-1 INFO:cartography.intel.aws.identitycenter:Getting role assignments for 803 users INFO:cartography.intel.aws.identitycenter:Loading 24292 role assignments INFO:cartography.intel.aws.identitycenter:Syncing Identity Center instances for region us-east-2 INFO:cartography.intel.aws.identitycenter:Loading 0 Identity Center instances for region us-east-2 INFO:cartography.intel.aws.identitycenter:Syncing Identity Center instances for region us-west-1 INFO:cartography.intel.aws.identitycenter:Loading 0 Identity Center instances for region us-west-1 INFO:cartography.intel.aws.identitycenter:Syncing Identity Center instances for region us-west-2 INFO:cartography.intel.aws.identitycenter:Loading 0 Identity Center instances for region us-west-2 INFO:cartography.graph.statement:Completed aws_import_identity_center_cleanup statement #1 INFO:cartography.graph.statement:Completed aws_import_identity_center_cleanup statement #2 INFO:cartography.graph.statement:Completed aws_import_identity_center_cleanup statement #3 INFO:cartography.graph.statement:Completed aws_import_identity_center_cleanup statement #4 INFO:cartography.graph.statement:Completed aws_import_identity_center_cleanup statement #5 INFO:cartography.graph.statement:Completed aws_import_identity_center_cleanup statement #6

**Related issues or links**

Fixes - https://github.com/cartography-cncf/cartography/issues/990

Checklist

[ x ] Update/add unit or integration tests.
[ X ] Include a screenshot showing what the graph looked like before and after your changes.
[ X ] Include console log trace showing what happened before and after your changes.

[ x ] Update the [schema](https://github.com/lyft/cartography/tree/master/docs/root/modules) and [readme](https://github.com/lyft/cartography/blob/master/docs/schema/README.md).

[ X ] Use the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).

### DIFF
--- a/cartography/intel/aws/identitycenter.py
+++ b/cartography/intel/aws/identitycenter.py
@@ -1,0 +1,371 @@
+import logging
+from typing import Dict
+from typing import List
+
+import boto3
+import neo4j
+
+from cartography.client.core.tx import load
+from cartography.models.aws.identitycenter.identitycenter import IdentityCenterInstanceSchema
+from cartography.models.aws.identitycenter.permissionsets import PermissionSetSchema
+from cartography.models.aws.identitycenter.ssouser import SSOUserSchema
+from cartography.util import aws_handle_regions
+from cartography.util import run_cleanup_job
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+@aws_handle_regions
+def get_identity_center_instances(boto3_session: boto3.session.Session, region: str) -> List[Dict]:
+    """
+    Get all AWS IAM Identity Center instances in the current region
+    """
+    client = boto3_session.client('sso-admin', region_name=region)
+    instances = []
+
+    try:
+        paginator = client.get_paginator('list_instances')
+        for page in paginator.paginate():
+            instances.extend(page.get('Instances', []))
+    except client.exceptions.ClientError as e:
+        logger.warning(f"Failed to get Identity Center instances in region {region}: {e}")
+        return []
+
+    return instances
+
+
+@timeit
+def load_identity_center_instances(
+    neo4j_session: neo4j.Session,
+    instance_data: List[Dict],
+    region: str,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    """
+    Load Identity Center instances into the graph
+    """
+    logger.info(f"Loading {len(instance_data)} Identity Center instances for region {region}")
+    load(
+        neo4j_session,
+        IdentityCenterInstanceSchema(),
+        instance_data,
+        lastupdated=aws_update_tag,
+        Region=region,
+        AWS_ID=current_aws_account_id,
+    )
+
+
+@timeit
+def cleanup_identity_center_instances(
+    neo4j_session: neo4j.Session,
+    common_job_parameters: Dict,
+) -> None:
+    """
+    Remove stale Identity Center instance nodes and relationships
+    """
+    run_cleanup_job(
+        'aws_import_identity_center_cleanup.json',
+        neo4j_session,
+        common_job_parameters,
+    )
+
+
+@timeit
+def get_permission_sets(boto3_session: boto3.session.Session, instance_arn: str, region: str) -> List[Dict]:
+    """
+    Get all permission sets for a given Identity Center instance
+    """
+    client = boto3_session.client('sso-admin', region_name=region)
+    permission_sets = []
+
+    try:
+        paginator = client.get_paginator('list_permission_sets')
+        for page in paginator.paginate(InstanceArn=instance_arn):
+            # Get detailed info for each permission set
+            for arn in page.get('PermissionSets', []):
+                try:
+                    details = client.describe_permission_set(
+                        InstanceArn=instance_arn,
+                        PermissionSetArn=arn,
+                    )
+                    permission_sets.append(details.get('PermissionSet', {}))
+                except client.exceptions.ClientError as e:
+                    logger.warning(f"Failed to get details for permission set {arn}: {e}")
+    except client.exceptions.ClientError as e:
+        logger.warning(f"Failed to get permission sets for instance {instance_arn} in region {region}: {e}")
+        return []
+
+    return permission_sets
+
+
+@timeit
+def get_permission_set_roles(
+    boto3_session: boto3.session.Session,
+    instance_arn: str,
+    permission_set_arn: str,
+    region: str,
+) -> List[Dict]:
+    """
+    Get all accounts associated with a given permission set
+    """
+    client = boto3_session.client('sso-admin', region_name=region)
+    accounts = []
+
+    try:
+        paginator = client.get_paginator('list_accounts_for_provisioned_permission_set')
+        for page in paginator.paginate(InstanceArn=instance_arn, PermissionSetArn=permission_set_arn):
+            accounts.extend(page.get('AccountIds', []))
+    except client.exceptions.ClientError as e:
+        logger.warning(f"Failed to get roles for permission set {permission_set_arn}: {e}")
+        return []
+
+    return accounts
+
+
+@timeit
+def load_permission_sets(
+    neo4j_session: neo4j.Session,
+    permission_sets: List[Dict],
+    instance_arn: str,
+    region: str,
+    aws_update_tag: int,
+) -> None:
+    """
+    Load Identity Center permission sets into the graph
+    """
+    logger.info(f"Loading {len(permission_sets)} permission sets for instance {instance_arn} in region {region}")
+
+    load(
+        neo4j_session,
+        PermissionSetSchema(),
+        permission_sets,
+        lastupdated=aws_update_tag,
+        InstanceArn=instance_arn,
+        Region=region,
+    )
+
+
+@timeit
+def get_permission_set_role_assignments(
+    boto3_session: boto3.session.Session,
+    instance_arn: str,
+    permission_sets: List[Dict],
+    region: str,
+) -> List[Dict]:
+    """
+    Get role assignments for Identity Center permission sets
+    """
+    role_assignments = []
+    for ps in permission_sets:
+        permission_set_arn = str(ps.get('PermissionSetArn'))
+        accounts = get_permission_set_roles(boto3_session, instance_arn, permission_set_arn, region)
+        for account_id in accounts:
+            arn = f"arn:aws:iam::{account_id}:role/aws-reserved/sso.amazonaws.com/AWSReservedSSO_{ps.get('Name')}"
+            role_assignments.append({
+                'PermissionSetArn': ps.get('PermissionSetArn'),
+                'RoleArn': arn,
+            })
+    return role_assignments
+
+
+@timeit
+def load_permission_set_role_assignments(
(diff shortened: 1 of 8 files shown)


SecPrez:
@achantavy Hey Alex ... how are you today?

cartography/intel/aws/identitycenter.py
```diff
@@ -0,0 +1,371 @@
...
+    neo4j_session: neo4j.Session,
+    role_assignments: List[Dict],
+    aws_update_tag: int,
+) -> None:
+    """
+    Load role assignments into the graph
+    """
+    logger.info(f"Loading {len(role_assignments)} role assignments")
+    if role_assignments:
+        neo4j_session.run(
```
achantavy, line 288: Let's use the data model here. I think we can do this by creating a `SSOUserToRole` relationship.

We can have a `transform` step that shapes the data returned from `get_sso_users()` and `get_role_assignments()` into what we need.

This way we don't have to use `neo4j_session.run()` (since this can fail and does not retry automatically), and the data model does batching and automatic cleanups and all that.

Let's get on a call if you're stuck with this. I'm also happy to push changes up to your branch if that's easier.

cartography/data/jobs/cleanup/aws_import_identity_center_cleanup.json
```diff
@@ -0,0 +1,36 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (n:AWSIdentityCenterInstance) WHERE n.lastupdated <> $UPDATE_TAG WITH n LIMIT $LIMIT_SIZE DETACH DELETE (n) RETURN COUNT(*) as TotalDeleted",
```
achantavy, line 4: Since the sync job runs one account at a time, this will currently delete all other AWSIdentityCenterInstance nodes in other accounts when we actually want to only delete the stale ones in the current account.

So since you've already built a data mode for this, you don't need to have a manually written cleanup job anymore and you can just do something like this:

https://github.com/cartography-cncf/cartography/blob/bf8d6670a9687bfb5dc40776af37b8a30cf2c910/cartography/intel/aws/ec2/instances.py#L307-L310

It will automatically generate and run the cleanup queries for you so you don't have to worry about any mistakes.

cartography/data/indexes.cypher
```diff
@@ -305,8 +305,7 @@ CREATE INDEX IF NOT EXISTS FOR (n:SpotlightVulnerability) ON (n.host_info_local_
 CREATE INDEX IF NOT EXISTS FOR (n:SpotlightVulnerability) ON (n.lastupdated);
 CREATE INDEX IF NOT EXISTS FOR (n:SQSQueue) ON (n.id);
 CREATE INDEX IF NOT EXISTS FOR (n:SQSQueue) ON (n.lastupdated);
-CREATE INDEX IF NOT EXISTS FOR (n:User) ON (n.arn);
-CREATE INDEX IF NOT EXISTS FOR (n:User) ON (n.lastupdated);
+CREATE INDEX IF NOT EXISTS FOR (n:UserAccount) ON (n.id);
```
achantavy, line 308: Since SSOUserSchema uses the data model, you don't need to manually update indexes for this. The index will be automatically ensured for you: https://github.com/cartography-cncf/cartography/blob/bf8d6670a9687bfb5dc40776af37b8a30cf2c910/cartography/graph/querybuilder.py#L424.

Under the hood, a function called `ensure_indexes()` runs everytime you call `load()` on a schema object: https://github.com/cartography-cncf/cartography/blob/bf8d6670a9687bfb5dc40776af37b8a30cf2c910/cartography/client/core/tx.py#L252

cartography/intel/aws/identitycenter.py
```diff
@@ -0,0 +1,371 @@
...
+    neo4j_session: neo4j.Session,
+    role_assignments: List[Dict],
+    aws_update_tag: int,
+) -> None:
+    """
+    Load Identity Center permission set role assignments into the graph
+    """
+    logger.info(f"Loading {len(role_assignments)} permission set role assignments")
+
+    neo4j_session.run(
```
achantavy, line 185: Same thing here about using the data model and a transform step.

cartography/models/aws/identitycenter/ssouser.py
```diff
@@ -0,0 +1,50 @@
...
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'id': PropertyRef('ExternalId')},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "CAN_ASSUME_IDENTITY"
+    properties: SSOUserToOktaUserRelProperties = SSOUserToOktaUserRelProperties()
+
+
+@dataclass(frozen=True)
+class SSOUserSchema(CartographyNodeSchema):
```
achantavy, line 42: Add in a sub_resource_relationship with label `RESOURCE`  to the AWSAccount. This makes it so that the autocleanup queries are easy to run.

cartography/models/aws/identitycenter/permissionsets.py
```diff
@@ -0,0 +1,44 @@
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "HAS_PERMISSION_SET"
+    properties: PermissionSetToInstanceRelProperties = PermissionSetToInstanceRelProperties()
+
+
+@dataclass(frozen=True)
+class PermissionSetSchema(CartographyNodeSchema):
+    label: str = 'AWSPermissionSet'
+    properties: PermissionSetProperties = PermissionSetProperties()
+    sub_resource_relationship: PermissionSetToInstance = PermissionSetToInstance()
```
achantavy, line 44: sub_resource_relationship is intended to point to the AWSAccount (yeah, it's not a great name, I have an issue open to rename it to something like "tenant_relationship"). It is used for autocleanup jobs so that we only delete stale nodes in the current account.

cartography/models/aws/identitycenter/identitycenter.py
```diff
@@ -0,0 +1,44 @@
...
+    identity_store_id: PropertyRef = PropertyRef('IdentityStoreId')
+    arn: PropertyRef = PropertyRef('InstanceArn')
+    created_date: PropertyRef = PropertyRef('CreatedDate')
+    id: PropertyRef = PropertyRef('InstanceArn')
+    status: PropertyRef = PropertyRef('Status')
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class IdentiyCenterToAwsAccountRelProperties(CartographyRelProperties):
```
achantavy, line 24: Typo: should be `IdentityCenterToAwsAccountRelProperties`.

achantavy (changes requested):
Thank you! Gave it a first pass and left comments about the data model. If something doesn't make sense let's get on a call. I can also push to your branch to show you what I mean.

cartography/data/jobs/cleanup/aws_import_identity_center_cleanup.json
SecPrez: Smart move. I will update

cartography/data/indexes.cypher
SecPrez: I need this because I added an abstract label of UserAccount and applied to all OktaUsers. SSOUsers can map to many IDP providers, so I did not want to make AWSSSOUsers dependent on a particular IDP provider. Since Okta and other IDP provider users are not using the new load method, I need to add the index here.

cartography/models/aws/identitycenter/identitycenter.py
SecPrez: Pending fix

cartography/data/indexes.cypher
achantavy: Right, and that case is taken care of because `SSOUserSchema` object has `UserAccount` in `extra_node_labels`.

So when you do 

```python
    load(
        neo4j_session,
        SSOUserSchema(),
        users,
     ...
    )
```

`load()` will call `ensure_indexes()` under the hood, and then `ensure_indexes()` will call `build_create_index_queries()`: 

https://github.com/cartography-cncf/cartography/blob/bf8d6670a9687bfb5dc40776af37b8a30cf2c910/cartography/graph/querybuilder.py#L424-L430

This will make sure that we run `CREATE INDEX IF NOT EXISTS` for every node label + `id` field. It also looks at all cases where TargetNodeMatchers are trying to match on other nodes and other fields and then makes sure an index exists for those.

So, all indexes are automatically taken care of. The indexes.cypher file is only for cases where we haven't had time to refactor older code.

SecPrez: You are correct that AWSSSOUser is also :UserAccount, but that is unimportant. I need :OktaUser to also be a :UserAccount. Since OktaUser is created the old way I added the index the old way. If you don't want the index to be updated the old way I can just wait for AWSSSOUser sync to happen but then I need to wait for the index to rebuild when it should have been built during OktaUser sync.

cartography/models/aws/identitycenter/ssouser.py
SecPrez: pending

cartography/data/indexes.cypher
achantavy: Gotcha, ok, that works then

cartography/intel/aws/identitycenter.py
achantavy: Understood. I took a stab at trying to allow fuzzy matching in the data model: https://github.com/cartography-cncf/cartography/pull/1383 - let me know if this helps

achantavy (approved):
Thanks for this! looks good. I need to figure out how to support some of the weird role mapping stuff with the model but that can come in a follow-up

PR #1383, opened 2024-11-13
Add support for fuzzy case-insensitive match in PropertyRef

achantavy:
We often need to connect 2 nodes based on a fuzzy match like in https://github.com/cartography-cncf/cartography/pull/1380#discussion_r1837801778. This PR adds a `fuzzy_and_ignore_case` option to the `PropertyRef` so that the final rendered query performs a `CONTAINS` in a case-insensitive way instead of an exact match.

https://github.com/cartography-cncf/cartography/pull/1380

### Checklist

### DIFF
--- a/tests/data/graph/querybuilder/sample_models/fake_emps_githubusers_fuzzy.py
+++ b/tests/data/graph/querybuilder/sample_models/fake_emps_githubusers_fuzzy.py
@@ -1,0 +1,44 @@
+from dataclasses import dataclass
+
+from cartography.models.core.common import PropertyRef
+from cartography.models.core.nodes import CartographyNodeProperties
+from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class FakeEmp2ToGitHubUserRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class FakeEmp2ToGitHubUser(CartographyRelSchema):
+    target_node_label: str = 'GitHubUser'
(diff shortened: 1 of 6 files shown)


tests/data/graph/querybuilder/sample_models/fake_emps_githubusers_fuzzy.py
```diff
@@ -0,0 +1,44 @@
...
+@dataclass(frozen=True)
+class FakeEmp2ToGitHubUserRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class FakeEmp2ToGitHubUser(CartographyRelSchema):
+    target_node_label: str = 'GitHubUser'
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'username': PropertyRef('github_username', fuzzy_and_ignore_case=True)},
```
achantavy, line 23: @SecPrez: this will allow us to connect 2 nodes in a "fuzzy" way, lemme know if this helps

tests/unit/cartography/graph/test_querybuilder_simple.py
```diff
@@ -90,3 +91,35 @@ def test_build_ingestion_query_case_insensitive_match():
...
+                i.lastupdated = $lastupdated,
+                i.email = item.email,
+                i.github_username = item.github_username
+
+        WITH i, item
+        CALL {
+            WITH i, item
+            OPTIONAL MATCH (n0:GitHubUser)
+            WHERE
+                toLower(n0.username) CONTAINS toLower(item.github_username)
```
achantavy, line 113: @SecPrez this is what the final rendered query will look like

achantavy:
@ramonpetgrave64 - you might be not available but just cc'ing you here anyway since i think you have the most context on this kind of data model change

achantavy:
@heryxpc lemme know what you think too

heryxpc (approved):

ramonpetgrave64 (approved):
looks good. 

In the future we may want to allow the sync module developer to add custom match templates, for more easily enabling other kinds of matches

```
email_dot_matcher = Template("replace($node_var.$key), ".", "") = replace($prop_ref, ".", "")"
```

to allow matching "s.f.seiei@example.com" to "sfseiei@example.com"

And then pass it to the schema with 

```
make_target_node_matcher(..., email_dot_matcher)
```

achantavy:
@ramonpetgrave64 so interesting to see the PropertyRef evolve

PR #1389, opened 2024-11-19
Fix IAM policy naming for inline policies with empty string Sid field.

SecPrez:
### Summary
  > AWS is breaking its own rules for some roles it has created in IdentityCenter. IdentityCenter created roles with inline policies that can have multiple statements, all with the Sid of an empty string. This means cartography will only load the last statement as they will all have the same AWSPolicyStatement.id. This change checks if the Sid is an empty string and uses the index numbering if it is encountered. 

### Checklist


achantavy (approved):
Ideally there'd be a test included for this or something to show the before and after but the change looks reasonable and small enough to me. Thank you!

PR #1391, opened 2024-11-20
Github "Direct" User Repo Access

danbrauer:
### Summary

This PR adds to the Github graph, adding repo access that is granted to all users who have a 'direct' affiliation to the repo.

Cartography currently [does](https://cartography-cncf.github.io/cartography/modules/github/schema.html#id6) already map some direct user repo access, but only for collaborators with an 'outside' affiliation to the repo.  This PR broadens that to include all collaborators, aka anybody with a 'direct' affiliation.  This follows Github's naming for these concepts, as seen [here](https://docs.github.com/en/graphql/reference/enums#collaboratoraffiliation).

In case it is unclear or for people newer to Github, note: this is focusing on access users are granted directly to a repo, as opposed to via a team.  Access granted via team is outside the scope of this PR.

We think this is a valuable addition to the graph for a few reasons, including:
1. Our analysts want few-to-no users to be granted access directly to repos, on the thinking that managing access via teams can make access easier to automate (ie with ABAC/RBAC type logic) and to audit.  Graphing direct-access, regardless of whether a user is within the org or outside it, will help highlight who to clean up.
2. Longer term, we eventually want to know, from the graph, _all_ access a user has.  This PR is a step in that direction.  (In a future PR, we hope to add a user-team membership relationship.  Since Cartography maps team-repo access rel, we could then have a user-team-repo graph, and that would complete the picture of user access in Github.)

Cartography AMPS User Direct Repo Access (3)

#### Screencaps

**A REPO WITH OUTSIDE COLLABORATORS**

BEFORE
CollabsBefore

AFTER
CollabsAfter

**A REPO WITH NON-OUTSIDE COLLABORATORS**

BEFORE
(no results, because these sorts of users are not graphed)
Screenshot 2024-11-21 at 5 54 37 PM

AFTER
Screenshot 2024-11-21 at 5 54 25 PM

**GENERAL COUNTS TO GIVE A SENSE OF CONNECTIONS NOW THERE**

BEFORE
UserRepoRelsBefore

AFTER
UserRepoRelsAfter

### Related issues or links

None

### Checklist

- [X] Include a screenshot showing what the graph looked like before and after your changes.

- [X] Update the [schema](https://github.com/lyft/cartography/tree/master/docs/root/modules) and [readme](https://github.com/lyft/cartography/blob/master/docs/schema/README.md).

**N/A/** If you are implementing a new intel module:

### DIFF
--- a/cartography/intel/github/repos.py
+++ b/cartography/intel/github/repos.py
@@ -1,6 +1,6 @@
 import configparser
 import logging
-from dataclasses import dataclass
+from collections import namedtuple
 from string import Template
 from typing import Any
 from typing import Dict
@@ -20,16 +20,17 @@ from cartography.util import timeit
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=False)
-class UserAffiliationAndRepoPermission:
-    """
-    Representation of a user's permission level and affiliation to a GitHub repo. See:
-    - Permission: https://docs.github.com/en/graphql/reference/enums#repositorypermission
-    - Affiliation: https://docs.github.com/en/graphql/reference/enums#collaboratoraffiliation
-    """
-    user: Dict
-    permission: str  # WRITE, MAINTAIN, ADMIN, etc
-    affiliation: str  # OUTSIDE, DIRECT
+# Representation of a user's permission level and affiliation to a GitHub repo. See:
+# - Permission: https://docs.github.com/en/graphql/reference/enums#repositorypermission
+# - Affiliation: https://docs.github.com/en/graphql/reference/enums#collaboratoraffiliation
+UserAffiliationAndRepoPermission = namedtuple(
+    'UserAffiliationAndRepoPermission',
+    [
+        'user',  # Dict
+        'permission',  # 'WRITE', 'MAINTAIN', 'ADMIN', etc
+        'affiliation',  # 'OUTSIDE', 'DIRECT'
+    ],
+)
 
 
 GITHUB_ORG_REPOS_PAGINATED_GRAPHQL = """
--- a/cartography/intel/github/repos.py
+++ b/cartography/intel/github/repos.py
@@ -1,10 +1,8 @@
 import configparser
 import logging
-from collections import namedtuple
 from dataclasses import dataclass
-from time import sleep
 from string import Template
-from typing import Any, Tuple
+from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -15,11 +13,11 @@ from packaging.requirements import Requirement
 from packaging.utils import canonicalize_name
 
 from cartography.intel.github.util import fetch_all
+from cartography.intel.github.util import PaginatedGraphqlData
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
-
 
 
 @dataclass(frozen=False)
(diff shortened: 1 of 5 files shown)


danbrauer:
As noted in the PR desc, this is partial and not yet working. Tests and docs are not yet updated, and I haven't reviewed my own code.  I am sharing it early because I wanted to ask a couple of questions, which I will share in slack soon.

cartography/intel/github/repos.py
```diff
@@ -12,11 +13,25 @@
...
+@dataclass(frozen=False)
+class UserAffiliationAndRepoPermission:
+    """
+    Representation of a user's permission level to a GitHub repo and affiliation type to the target GitHub org. See:
+    - Permission: https://docs.github.com/en/graphql/reference/enums#repositorypermission
+    - Affiliation: https://docs.github.com/en/graphql/reference/enums#collaboratoraffiliation
+    """
+    user: Dict
+    permission: str  # WRITE, MAINTAIN, ADMIN, etc
+    affiliation: str  # OUTSIDE, DIRECT
```
danbrauer, line 32: I was led to creating this from a [similar thing](https://github.com/cartography-cncf/cartography/blob/master/cartography/intel/github/teams.py#L20) done in teams.py.  There, a namedtuple is used.  Here, I used a dataclass.  Is that confusing, or is it okay?

tests/data/github/repos.py
```diff
@@ -103,52 +110,8 @@
...
-                },
-                {
-                    'url': 'https://github.com/roshinis78',
-                    'login': 'roshinis78',
-                    'name': 'Roshini Saravanakumar',
-                    'email': 'ro@example.com',
-                    'company': 'ExampleCo',
-                },
-            ],
-        },
```
danbrauer, line 151: I largely replicated this data as is below, tho I did change the permissions levels so that they vary instead of all being 'WRITE'

danbrauer:
This should be ready for review.  I do still intend to do a review of my own for typos and the like but short of stuff like that, I think this is ready.

cartography/intel/github/repos.py
```diff
@@ -12,11 +13,25 @@
...
+@dataclass(frozen=False)
+class UserAffiliationAndRepoPermission:
+    """
+    Representation of a user's permission level and affiliation to a GitHub repo. See:
+    - Permission: https://docs.github.com/en/graphql/reference/enums#repositorypermission
+    - Affiliation: https://docs.github.com/en/graphql/reference/enums#collaboratoraffiliation
+    """
+    user: Dict
+    permission: str  # WRITE, MAINTAIN, ADMIN, etc
+    affiliation: str  # OUTSIDE, DIRECT
```
danbrauer, line 32: I was led to creating this from a [similar thing](https://github.com/cartography-cncf/cartography/blob/master/cartography/intel/github/teams.py#L20) done in teams.py.  There, a namedtuple is used.  Here, I used a dataclass.  I could go back to using a namedtuple if that'd be preferred?

achantavy: Yeah let's use a namedtuple instead so that it matches

achantavy (approved):
This is great, thank you!

Left a minor note about changing the dataclass to a namedtuple to be consistent, but otherwise this is great.

danbrauer:
Updated to named tuple.  Will be back on Tue and can merge if all looks good (or you can merge if you want in my absence!)

achantavy:
Thank you!

Including a note for future searchers: we deferred adding the data model here. See this thread for details: https://cloud-native.slack.com/archives/C080M2LRLDA/p1732137038521919

cartography/intel/github/repos.py
```diff
@@ -111,34 +226,52 @@ def get(token: str, api_url: str, organization: str) -> List[Dict]:
     return repos.nodes
 
 
-def transform(repos_json: List[Dict]) -> Dict:
+def transform(
+    repos_json: List[Dict], direct_collaborators: dict[str, List[UserAffiliationAndRepoPermission]],
+    outside_collaborators: dict[str, List[UserAffiliationAndRepoPermission]],
```
cmm-lyft, line 231: @achantavy this resulted in a breaking change as it is changing the signature of the function without providing a default value nor allowing to pass None as param.
Should we update this code? or this should be enforced for next versions?

achantavy: Is the breaking change in the Lyft internal deployment of cartography? If so share the stack trace with me internally. I thought I accounted for the changes; sorry, I probably missed something.

PR #1392, opened 2024-11-27
EC2 network ACL: Use get method to access dict keys to prevent exception

cmm-lyft:
### Summary
Correcting error:
```
File "/code/venvs/venv/lib/python3.10/site-packages/cartography/intel/aws/ec2/network_acls.py", line 72, in transform_network_acl_data
    'CidrBlock': rule['CidrBlock'],
KeyError: 'CidrBlock'
```

### Checklist

### DIFF
--- a/cartography/intel/aws/ec2/network_acls.py
+++ b/cartography/intel/aws/ec2/network_acls.py
@@ -69,7 +69,8 @@ def transform_network_acl_data(
                 direction = 'egress' if rule['Egress'] else 'inbound'
                 transformed_rule = {
                     'Id': f"{network_acl['NetworkAclId']}/{direction}/{rule['RuleNumber']}",
-                    'CidrBlock': rule['CidrBlock'],
+                    'CidrBlock': rule.get('CidrBlock'),
+                    'Ipv6CidrBlock': rule.get('Ipv6CidrBlock'),
                     'Egress': rule['Egress'],
                     'Protocol': rule['Protocol'],
                     'RuleAction': rule['RuleAction'],
--- a/cartography/models/aws/ec2/network_acl_rules.py
+++ b/cartography/models/aws/ec2/network_acl_rules.py
@@ -21,6 +21,7 @@ class EC2NetworkAclRuleNodeProperties(CartographyNodeProperties):
     fromport: PropertyRef = PropertyRef('FromPort')
     toport: PropertyRef = PropertyRef('ToPort')
     cidrblock: PropertyRef = PropertyRef('CidrBlock')
+    Ipv6CidrBlock: PropertyRef = PropertyRef('Ipv6CidrBlock')
     egress: PropertyRef = PropertyRef('Egress')
     rulenumber: PropertyRef = PropertyRef('RuleNumber')
     ruleaction: PropertyRef = PropertyRef('RuleAction')
--- a/docs/root/modules/aws/schema.md
+++ b/docs/root/modules/aws/schema.md
@@ -1324,18 +1324,19 @@ Representation of an AWS Elastic Container Registry [Repository](https://docs.aw
 Representation of an AWS [EC2 Network ACL Rule Entry](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkAclEntry.html)
 For additional explanation see https://docs.aws.amazon.com/vpc/latest/userguide/nacl-rules.html.
 
-| Field          | Description                                                                                 |
-|----------------|---------------------------------------------------------------------------------------------|
-| **id**         | The ID of this rule: `{network_acl_id}/{egress or inbound}/{rule_number}`                   |
-| network_acl_id | The ID of the network ACL that this belongs to                                              |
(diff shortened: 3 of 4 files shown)


cmm-lyft:
👀  @achantavy

cmm-lyft:
👀  @heryxpc @serge-wq @khanhldt

cartography/intel/aws/ec2/network_acls.py
```diff
@@ -68,12 +68,12 @@ def transform_network_acl_data(
...
-                    'Protocol': rule['Protocol'],
-                    'RuleAction': rule['RuleAction'],
-                    'RuleNumber': rule['RuleNumber'],
+                    'Id': f"{network_acl['NetworkAclId']}/{direction}/{rule.get('RuleNumber')}",
+                    'CidrBlock': rule.get('CidrBlock'),
+                    'Egress': rule.get('Egress'),
+                    'Protocol': rule.get('Protocol'),
+                    'RuleAction': rule.get('RuleAction'),
+                    'RuleNumber': rule.get('RuleNumber'),
                     # Add pointer back to the nacl to create an edge
```
kledo-lyft, line 77: switching to `get` comes with the question: what's the default value? And what's the implication of having such broken / default data in our DB?

I think the data model for this is defined here:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-networkaclentry.html

CIDR block is indicated as "Conditional", and it looks like either this field or `Ipv6CidrBlock` must be present. So we should update our code here to take one or another instead.

kledo-lyft: Probably worth reviewing the rest of the parameters too.

cmm-lyft: I wouldn't consider this data as "broken". We trust the incoming data from the AWS ACLs API, considering the optional fields. Any required transformations or enrichments can be performed ad hoc by the client consuming this data.

serge-wq: I think the question here is whether cartography should provide an accurate representation of the AWS data and its constraints.

If so, then we should adhere to what @kledo-lyft and provide some kind of validation to spot the cases where we don't get any of the two values. If not, then it's fine to just store the data _as is_ and let the cartography consumers think about this.

I think the first option is better. Even though it can mean more maintenance work, consumers can be sure the data is as accurate as possible to AWS and it will also make easier to spot any bug when the data we're storing doesn't follow these constraints.

achantavy: First, sorry for the crash. It's annoying that AWS APIs are inconsistent and sometimes leave out entire fields.

>I think the data model for this is defined here:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-networkaclentry.html

CIDR block is indicated as "Conditional", and it looks like either this field or Ipv6CidrBlock must be present. So we should update our code here to take one or another instead.

@kledo-lyft - good catch. I was referring to https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkAclEntry.html when I wrote this code. It looks like the cloudformation docs are better. 


This fix is simple:

1. do a `.get('CidrBlock')`.
2. do a `.get('Ipv6CidrBlock')`.
3. leave everything else the same (unless we know there are others where the API leaves it out).
4. update https://github.com/cartography-cncf/cartography/blob/774d67b560ea50035fe6f8ed72795ec0c3d45cf7/cartography/models/aws/ec2/network_acl_rules.py#L23 to have an ipv6cidrblock field

This works because the neo4j python driver treats Nones as if the field does not even exist, so assuming that the aws conditional is correct, our end result is that the node written to the graph will have either cidrblock or ipv6cidrblock but not both.

---

As seen here, cidrblock and ipv6cidrblock are both not id fields so it's ok for these to be None.
https://github.com/cartography-cncf/cartography/blob/774d67b560ea50035fe6f8ed72795ec0c3d45cf7/cartography/models/aws/ec2/network_acl_rules.py#L23

achantavy (changes requested):

achantavy (approved):

cartography/models/aws/ec2/network_acl_rules.py
```diff
@@ -21,6 +21,7 @@ class EC2NetworkAclRuleNodeProperties(CartographyNodeProperties):
     fromport: PropertyRef = PropertyRef('FromPort')
     toport: PropertyRef = PropertyRef('ToPort')
     cidrblock: PropertyRef = PropertyRef('CidrBlock')
+    Ipv6CidrBlock: PropertyRef = PropertyRef('Ipv6CidrBlock')
     egress: PropertyRef = PropertyRef('Egress')
```
kledo-lyft, line 25: @cmm-lyft  a bit late to comment but probably more consistent to use lowercase `ip`. If we haven't cut a release, maybe can still do a quick fix?

achantavy: oh yeah good catch. lol sorry i got too eager with the approval

PR #1393, opened 2024-11-29
Increasing timeout ratios for CVE API requests

cmm-lyft:
### Summary
Based on tests conducted semi-automated, recently https://services.nvd.nist.gov/rest/json/cves/2.0/ response time has increased up to 100,000 ms, and the latest executions have raised 503 error, responding after 20 to 30 retries with 3 seconds intervals:

`Failed to get CVE data from NIST NVD API 503 : <html><body><h1>503 Service Unavailable</h1> No server is available to handle this request. </body></html>`

Increasing the `CONNECT_AND_READ_TIMEOUT` to 120 seconds, the MAX_RETRIES to 30 and including sleep time on Exception to mitigate the insidense of 503 error.

The max execution time of the request logic (worst case scenario) would be of 7 minutes (120 of request connection, 120 of request timeout, 30 retries with 6 sleep time as max).

### Checklist


cmm-lyft:
👀  @achantavy @heryxpc @serge-wq @khanhldt

cartography/intel/cve/feed.py
```diff
@@ -22,9 +22,9 @@
 
 logger = logging.getLogger(__name__)
 
-MAX_RETRIES = 3
-# Connect and read timeouts of 60 seconds each; see https://requests.readthedocs.io/en/master/user/advanced/#timeouts
-CONNECT_AND_READ_TIMEOUT = (60, 60)
+MAX_RETRIES = 30
+# Connect and read timeouts of 120 seconds each; see https://requests.readthedocs.io/en/master/user/advanced/#timeouts
+CONNECT_AND_READ_TIMEOUT = (120, 120)
```
kledo-lyft, line 27: 503 is not related to (connect, read) timeout config, no? Any reason to do those changes?

ideally, i would suggest to use http adapter with backoff. If that's complex to change, at least try to do similar thing in code (doubling the sleep time every next attempt).

cmm-lyft: I thought the same, but executing tests I observed that is the current behavior. 
You can manually try with this link: https://services.nvd.nist.gov/rest/json/cves/2.0/
execute the request several times from your browser (even without passing credentials / token), and you will reproduce that 503 error.
the response is super slow, and in certain cases (from my tests) it keeps returning the error even after +20 retries:
image

kledo-lyft: I see. It looks like a timeout that occurred on their reverse proxy or load balancer. The configuration we have here is for the client side to cut off the connection, I suspect it make a difference.

I tried the `curl` command and it failed with 503 fairly fast, then 1 time it successfully waited for 10 seconds before returning the whole data. Usually connect timeout doesn't need to be very long, while read timeout can be longer to wait for the data to come back. Still maybe you don't really need to do those changes. If you really wish so, I would suggest something like (30, 120) instead.

And yes, backoff retry is better because if everyone keeps bombarding their service won't be able to recover.

cartography/intel/cve/feed.py
```diff
@@ -98,6 +98,9 @@ def _call_cves_api(url: str, api_key: str | None, params: Dict[str, Any]) -> Dic
             retries += 1
             if retries >= MAX_RETRIES:
                 raise
+            # Exponential backoff
+            sleep_time *= 2
+            time.sleep(sleep_time)
```
kledo-lyft, line 103: so a total of 3 * 2^5 or 6 * 2^5 seconds when API key is provided?

That seems low. Do we need higher number?

cmm-lyft: why   2^5?
condition at L99 is >= , so it will actually iterate over 6. 
3 * 2^6 = 192
3 * 2^6 = 384
I think those times are high enough. 👀

kledo-lyft: heh, let's try to enumerate it.
it's incremented by 1 before the check and stopped when the value reaches 6
```
retries variable value -> sleep time
1 -> 3
2 -> 6
3 -> 12
4 -> 24
5 -> 48
6 -> stopped (hit >= MAX_RETRIES)
```

kledo-lyft (approved):

cmm-lyft:
👀 @achantavy

achantavy (approved):

PR #1395, opened 2024-12-04
Github "Immediate" User Team Membership

danbrauer:
### Summary

This PR adds to the Github graph, adding user membership in teams, for users who are 'immediate' members of a team.

In case it is unclear or for people newer to Github, note: this is focusing on ['immediate'](https://docs.github.com/en/graphql/reference/enums#teammembershiptype) membership to a team, meaning a member is in the team directly as opposed to being in a child team.  A user could be considered a member of a team if they are members of a child team, but this PR maps only 'immediate' membership.  (In a follow-up PR we'd like to add child teams to the graph, which we think will complete the membership picture.)

We think this is a valuable addition to the graph because our broad intent is to understand all access a user has, and (at least in our org) most access to repos is granted via team.  If we do not know who is in the team, then, we do not know who has access.

Cartography AMPS User Direct Team Membership

#### Screencaps

**EXAMPLE USER LOOKUP**

BEFORE
(empty result because nothing exists)
Screenshot 2024-12-03 at 5 07 00 PM

AFTER
Screenshot 2024-12-03 at 5 06 04 PM

**OVERVIEW OF COUNTS OF EACH TYPE**

BEFORE
(empty result because nothing exists)
Screenshot 2024-12-03 at 5 09 03 PM

AFTER
Screenshot 2024-12-03 at 5 08 44 PM

### Related issues or links

None

### Checklist

**N/A** If you are implementing a new intel module:

### DIFF
--- a/cartography/intel/github/teams.py
+++ b/cartography/intel/github/teams.py
@@ -17,7 +17,10 @@ from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
 
+# A team's permission on a repo: https://docs.github.com/en/graphql/reference/enums#repositorypermission
 RepoPermission = namedtuple('RepoPermission', ['repo_url', 'permission'])
+# A team member's role: https://docs.github.com/en/graphql/reference/enums#teammemberrole
+UserRole = namedtuple('UserRole', ['user_url', 'role'])
 
 
 @timeit
@@ -33,6 +36,9 @@ def get_teams(org: str, api_url: str, token: str) -> Tuple[PaginatedGraphqlData,
                         url
                         description
                         repositories(first: 100) {
+                            totalCount
+                        }
+                        members(first: 100, membership: IMMEDIATE) {
                             totalCount
                         }
                     }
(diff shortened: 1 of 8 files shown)


achantavy (approved):
Looks good, thank you!

I will open a task to support cleanup jobs for node schemas that aren't attached to a tenant (theres a slack thread on this somewhere).

Right now this code does not support users moving between teams because there is no cleanup job defined for GitHub team-to-user relationships.

Give me a little bit and I'll try to add that feature before the next release.

For now, can you let me know your thoughts on my non-blocking comments?

cartography/intel/github/teams.py
```diff
@@ -35,6 +38,9 @@ def get_teams(org: str, api_url: str, token: str) -> Tuple[PaginatedGraphqlData,
                         repositories(first: 100) {
                             totalCount
                         }
+                        members(first: 100, membership: IMMEDIATE) {
```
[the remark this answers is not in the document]
danbrauer: I thought it was required, even though it didn't seem necessary.  I could've sworn in other queries elsewhere I got errors without it, tho it seems like here for this it isn't needed.  I will remove it tmmw.

(Oh also, I was following the pattern in the line above, where `repositories` does this, specifying `first: 100` even tho it's only looking for a count.  I could see if that can be cleaned up as well.)

cartography/intel/github/teams.py
```diff
@@ -142,10 +148,106 @@ def _get_team_repos(org: str, api_url: str, token: str, team: str) -> PaginatedG
     return team_repos
 
 
+def _get_team_users_for_multiple_teams(
```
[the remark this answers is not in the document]
danbrauer: Here I was copying the existing pattern for the pre-existing function in this file, `_get_team_repos_for_multiple_teams`.  So I can look into how the decorator works and see if I understand how to apply it, but then I'd want to do the same in the other function, so there isn't a confusing difference between two very similar funcs.

cartography/intel/github/teams.py
```diff
@@ -142,10 +148,106 @@ def _get_team_repos(org: str, api_url: str, token: str, team: str) -> PaginatedG
...
+
+            except TypeError:
+                # Handles issue #1334
+                logger.warning(
+                    f"GitHub returned None when trying to find user or role data for team {team_name}.",
+                    exc_info=True,
+                )
+                if current_try == max_tries:
+                    raise RuntimeError(f"GitHub returned a None member url for team {team_name}, retries exhausted.")
+                sleep(current_try ** 2)
```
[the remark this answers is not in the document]
danbrauer: Here again I was copying the pre-existing function in this file, `_get_team_repos_for_multiple_teams`.  I can add a log in both, so they're consistent.

danbrauer:
I took a quick look and replied.  I can try to do the changes tmmw morning.  Two of them seems clear (cleaning up the graphql; adding a log statement).  The other one, about using a backoff decorator, is less clear but I am happy to look into it when back at my desk.  Thank you!

danbrauer:
I pushed the two easy updates.  I will look at the third, doing a 'backoff', in a bit.

danbrauer:
Using a backoff decorator seems simple enough.  It doesn't save many lines of code, but it does feel better.  I am encountering some typing issues around the decorated function, and then will have to understand how to update existing testing to ensure the retries act as desired / decorator is wired up right.  I have to step away now but can come back to it later today.

cartography/intel/github/teams.py
```diff
@@ -23,6 +23,18 @@
...
+def backoff_handler(details: Dict) -> None:
+    """
+    Custom backoff handler for GitHub calls in this module.
+    """
+    team_name = details['kwargs'].get('team_name') or 'not present in kwargs'
+    updated_details = {**details, 'team_name': team_name}
+    logger.warning(
+        "Backing off {wait:0.1f} seconds after {tries} tries. Calling function {target} for team {team_name}"
+        .format(**updated_details),
+    )
```
danbrauer, line 35: Wasn't sure how valuable this was over just using the backoff_handler over in util.py.  If you think it's not worth it, I can get rid of it.

cartography/intel/github/teams.py
```diff
@@ -70,40 +82,27 @@ def _get_team_repos_for_multiple_teams(
...
+            # https://docs.github.com/en/graphql/reference/objects#teamrepositoryconnection
+            for repo in team_repos.nodes or []:
+                repo_urls.append(repo['url'])
+            # The `or []` is because `.edges` can be None.
+            for edge in team_repos.edges or []:
+                repo_permissions.append(edge['permission'])
+
+        retries_with_backoff(get_teams_repos_inner_func, TypeError, 5, backoff_handler)(
+            org=org, api_url=api_url, token=token, team_name=team_name,
+            repo_urls=repo_urls, repo_permissions=repo_permissions,
```
danbrauer, line 104: 1. I had to wrap the inner func in another func to work through some mypy typechecks.  (Also side benefit is it allows `retries_with_backoff` to be a reusable thing.)
2. I used kwargs instead of args to make the backoff_handler able to take some info out of the args easily, for more specific logging.

tests/unit/cartography/intel/github/test_teams.py
```diff
@@ -123,8 +123,8 @@ def test_get_team_users_happy_path(mock_get_team_users):
 
 
 @patch('cartography.intel.github.teams._get_team_repos')
-@patch('cartography.intel.github.teams.sleep')
-def test_get_team_repos_github_returns_none(mock_sleep, mock_get_team_repos):
+@patch('cartography.intel.github.teams.backoff_handler', spec=True)
```
danbrauer, line 126: I didn't _need_ to replace 'sleep' with 'backoff_handler', we could've had a fine test without it.  But, it did give a nice extra affirmation that the decorator is wired in correctly, because we can tell that it does its backoff handling the expected number of times.

The `spec=True` was needed, if I understand right, to fool the decorator into thinking it had a backoff handler func it could use.  Without this, with just a MagicMock, it wasn't being invoked.  With this, it was being invoked exactly as expected.

tests/unit/cartography/intel/github/test_teams.py
```diff
@@ -134,7 +134,7 @@ def test_get_team_repos_github_returns_none(mock_sleep, mock_get_team_repos):
     mock_get_team_repos.return_value = mock_team_repos
 
     # Assert we raise an exception
-    with pytest.raises(RuntimeError):
+    with pytest.raises(TypeError):
```
danbrauer, line 137: TypeError here rather than RuntimeError because in the old code, it explicitly threw a RuntimeError, whereas in the new code, it throws whatever the underlying error is, which in this case is TypeError.

cartography/intel/github/teams.py
```diff
@@ -32,7 +47,10 @@ def get_teams(org: str, api_url: str, token: str) -> Tuple[PaginatedGraphqlData,
                         slug
                         url
                         description
-                        repositories(first: 100) {
+                        repositories {
```
danbrauer, line 50: removing the 'first: 100' here because it was never necessary

danbrauer:
@achantavy This should be ready for review again.  Happy to discuss the use of the decorator (or anything else) if you had something different in mind.

achantavy:
Spoke with Daniel, we are going to merge this in for now and it will be part of a release candidate. Will fast follow with a change to add automatic cleanup jobs.

PR #1397, opened 2024-12-06
Github Child Team Membership

danbrauer:
### Summary

This PR adds to the Github graph, adding child team members of teams.

This is very similar to, and can be considered a follow-up to the prior PR #1395, which added 'immediate' user members of teams.  Now this PR is adding child-teams.  Between these two, the graph can now answer the question of who is a member of a team, either directly or via child teams.

We think this is a valuable addition to the graph because our broad intent is to understand all access a user has.  Since access is frequently granted to teams, we need the complete picture of team membership.

Cartography AMPS Team Child Team Membership

#### Screencaps

**EXAMPLE ALL TEAMS LOOKUP**

BEFORE
(nothing is returned because the relationship does not exist)
Screenshot 2024-12-06 at 11 23 54 AM

AFTER
(details not visible here but hopefully this related a sense of the relationships)
Screenshot 2024-12-06 at 11 53 11 AM

**EXAMPLE SINGLE TEAM LOOKUP**

BEFORE
(nothing is returned because the relationship does not exist)
Screenshot 2024-12-06 at 12 44 55 PM

AFTER
Screenshot 2024-12-06 at 12 44 32 PM

#### Note on loops / cyclic graphs

GitHub does not seem to allow the creation of loops, eg `Team A <-- Team B <-- Team A`, so this PR has no handling or guarding against that sort of thing.  

### Related issues or links

None.

### Checklist

**Not applicable** If you are implementing a new intel module:

### DIFF
--- a/cartography/intel/github/teams.py
+++ b/cartography/intel/github/teams.py
@@ -172,6 +172,21 @@ def _get_team_repos(org: str, api_url: str, token: str, team: str) -> PaginatedG
     return team_repos
 
 
+def _get_teams_users_inner_func(
+        org: str, api_url: str, token: str, team_name: str,
+        user_urls: List[str], user_roles: List[str],
+) -> None:
+    logger.info(f"Loading team users for {team_name}.")
+    team_users = _get_team_users(org, api_url, token, team_name)
+    # The `or []` is because `.nodes` can be None. See:
+    # https://docs.github.com/en/graphql/reference/objects#teammemberconnection
+    for user in team_users.nodes or []:
+        user_urls.append(user['url'])
+    # The `or []` is because `.edges` can be None.
+    for edge in team_users.edges or []:
+        user_roles.append(edge['role'])
+
+
 def _get_team_users_for_multiple_teams(
         team_raw_data: list[dict[str, Any]],
         org: str,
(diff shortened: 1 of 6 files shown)


danbrauer:
a few tweaks from #1406 spilled over into here, and I rebased.  I need to review a tiny bit more and change a couple of small details to line up with the style made in that PR.  Will update here later today.

cartography/intel/github/teams.py
```diff
@@ -166,6 +172,21 @@ def _get_team_repos(org: str, api_url: str, token: str, team: str) -> PaginatedG
...
+) -> None:
+    logger.info(f"Loading team users for {team_name}.")
+    team_users = _get_team_users(org, api_url, token, team_name)
+    # The `or []` is because `.nodes` can be None. See:
+    # https://docs.github.com/en/graphql/reference/objects#teammemberconnection
+    for user in team_users.nodes or []:
+        user_urls.append(user['url'])
+    # The `or []` is because `.edges` can be None.
+    for edge in team_users.edges or []:
+        user_roles.append(edge['role'])
```
danbrauer, line 187: not strictly related to the intent of this PR, but I made this change to keep in line with the style change made in #1404, where a different 'inner_func' in this file was pulled out of the function it was inside of.

tests/unit/cartography/intel/github/test_teams.py
```diff
@@ -122,10 +151,37 @@ def test_get_team_users_happy_path(mock_get_team_users):
...
+        ],
+    }
+
+    # Assert that we did not retry because this was the happy path
+    mock_get_child_teams.assert_called_once_with('test-org', 'https://api.github.com', 'test-token', 'team1')
+
+
 @patch('cartography.intel.github.teams._get_team_repos')
 @patch('cartography.intel.github.teams.backoff_handler', spec=True)
-def test_get_team_repos_github_returns_none(mock_backoff_handler, mock_get_team_repos, mock_sleep):
```
danbrauer, line 128: I removed the mock_sleep again because sleep is no longer used/imported in this logic, so there's no need to mock it.  it _used_ to be used for the retry logic, but that's been replaced by the `@backoff` annotation.

danbrauer:
**UPDATE** this is rebased and sanity tested.  Should be ready for review/merge.

achantavy (approved):

danbrauer:
🙇 thank you @achantavy!  just a reminder I am unable to merge PRs myself so if this is good to go, please do the honors.

PR #1402, opened 2024-12-11
Set outside_collaborators new func argument as optional to avoid breaking changes

cmm-lyft:


cmm-lyft:
👀  @achantavy @heryxpc  @khanhldt

Issue #1374, opened 2024-10-31
Github user.role is potentially inaccurate in graphs with multiple github orgs?


danbrauer:
In the process of working on [pr 1373](https://github.com/cartography-cncf/cartography/pull/1373) I noticed what seems like an edge case for users who are in multiple Github organizations: imagine a user who is an 'ADMIN' to Org1, and a 'MEMBER' to Org2.  The user node's ‘role’ property will say either 'ADMIN' or 'MEMBER', so, it will be incorrect with respect to one of the orgs.

This was not an issue for my PR specifically but in thinking through it, it did make me think of an alternative graph that might be clearer?  Cartography could graph these relationships like so: 

> (User)-[MEMBER_OF|ADMIN_OF|UNAFFILIATED]-(Org)

And then the user.role property would goes away (because its info would now be encoded in the type of the relationship between user and org).

I haven't looked into implementing this yet, because I wasn't sure if it's of any interest, or if maybe there is a reason for the way things are currently done.  It would be a 'breaking' change, since it removes a property from a node and would change the relationship type of some users (i.e. admins currently are MEMBER_OF but would become ADMIN_OF).

Does this make sense and, if we wanted to make this update in one of our upcoming sprints, would that be welcome?  It's not critical for us, and we have other changes we'd like to make, but it might be something I can do.

>  - Cartography release version 0.93
>  - Python version: 3.12

achantavy:
Mm, yeah everything you wrote sounds correct. The code right now does not support the idea of a user having different 'role' properties for different orgs. How about a schema like this:

1. A user belongs to a github org. This is the "tenant relationship". cartography convention is to have one label be that specific relationship so that we can determine what org a user belongs to in a uniform way -- that is, without needing to account for multiple possibilities of label values connecting back to the org.

```cypher
(:GitHubOrganization)<-[:MEMBER_OF]-(:GitHubUser)
```

2. Some users can be admins of github orgs:

```cypher
(:GitHubOrganization)<-[:ADMIN_OF|-(:GitHubUser)
```

So in the case where a user is an admin, we would have 1 MEMBER_OF edge + 1 ADMIN_OF edge, so it would be a bit redundant, but that's fine. It'd look something like

```cypher

(:GitHubUser)-[:MEMBER_OF]->(:GitHubOrganization)
          |                 ^
           -[:ADMIN_OF]----|
```

Thoughts?

cc: @heryxpc @ramonpetgrave64 @chandanchowdhury

[...]

PR #1403, opened 2024-12-12
GitHub: introduce ADMIN_OF relationship between users and orgs

danbrauer:
### Summary

This PR fixes the issue described in issue #1374—please see the issue and discussion there for details, but here it is in summary:

The 'role' property on a GitHubUser node expresses whether a user is an 'ADMIN' (aka Owner) or a 'MEMBER' of a GitHubOrganization.  For a graph with multiple orgs, however, the role property will reflect only one of the user-org relationships.  If the user is an admin in one org and a member in another, for example, the 'role' property will be wrong for one of those relationships.  To fix this, this PR shifts the role to be expressed in the relationship between GitHubUser and GitHubOrganization.

Below are a few before/after screen caps to show intent and the PR working.

#### **USER IS AN OWNER OF ONE ORG, MEMBER OF ANOTHER**

BEFORE
Here the user is graphed as 'MEMBER_OF' to both orgs.  The 'role' property (not pictured) is set to reflect one of those relationships, so it is partly wrong.
Screenshot 2024-12-09 at 5 22 15 PM

AFTER
The relationship is now updated to reflect that the user is an admin of one of the orgs.  And, as suggested by @achantavy  in #1374, admins now get two user-org relationships "so that we can determine what org a user belongs to in a uniform way".
Screenshot 2024-12-09 at 5 08 04 PM

#### **MEMBERS AND UNAFFILIATEDS REMAIN THE SAME**

BEFORE & AFTER
This user is handy for demonstration purposes because they are a member of one org, and unaffiliated to another (the user in question is an enterprise owner, and is not a member of the second org).  These relationships remain untouched by this PR, and so the before and after results for the query shown are identical.
Screenshot 2024-12-09 at 5 07 20 PM

#### **RELATIONSHIP COUNTS**

BEFORE
Only MEMBER_OF and UNAFFILIATED is graphed, and to know who is an ADMIN you have to look at a property:
Screenshot 2024-12-11 at 8 44 21 PM
Screenshot 2024-12-11 at 9 03 52 PM
(Note there are 16 admins, but also note how I had to specify folks who are 'UNAFFILIATED' in my query.  If I hadn't done that, I would've gotten a higher number, as some of those folks are admins but of a different org.  So the result would have been wrong.)

AFTER
In addition to the exiting relationships that were there before, there are 16 new ADMIN_OF relationships for every node that _had_ an ADMIN role property.
Screenshot 2024-12-11 at 8 41 56 PM

### Related issues or links

- #1374 

### Checklist

**NOT APPLICABLE** you are implementing a new intel module:

### DIFF
--- a/cartography/intel/github/users.py
+++ b/cartography/intel/github/users.py
@@ -144,10 +144,13 @@ def transform_users(user_data: List[Dict], owners_data: List[Dict], org_data: Di
 
     users_dict = {}
     for user in user_data:
+        # all members get the 'MEMBER_OF' relationship
         processed_user = deepcopy(user['node'])
-        processed_user['role'] = user['role']
         processed_user['hasTwoFactorEnabled'] = user['hasTwoFactorEnabled']
         processed_user['MEMBER_OF'] = org_data['url']
+        # admins get a second relationship expressing them as such
+        if user['role'] == 'ADMIN':
+            processed_user['ADMIN_OF'] = org_data['url']
         users_dict[processed_user['url']] = processed_user
 
     owners_dict = {}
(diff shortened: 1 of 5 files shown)


achantavy (approved):
LGTM

PR #1405, opened 2024-12-12
nonetype handling for repo collaborators

danbrauer:
### Summary

Fix for #1404

### Related issues or links

- #1404

### Checklist

Provide proof that this works (this makes reviews move faster). Please perform one or more of the following:
- [x] Update/add unit or integration tests.
- [x] Include a screenshot showing what the graph looked like before and after your changes.
- [ ] Include console log trace showing what happened before and after your changes.

**Not applicable** If you are changing a node or relationship:
- [ ] Update the [schema](https://github.com/lyft/cartography/tree/master/docs/root/modules) and [readme](https://github.com/lyft/cartography/blob/master/docs/schema/README.md).

**Not applicable** If you are implementing a new intel module:
- [ ] Use the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).

PR #1406, opened 2024-12-13
0.96.1: Fix #1404: Handle transient errors from GitHub repo collaborators sync

### LINKED ISSUE #1404: github sync nonetype error when listing collaborators
*Title*: 
GitHub sync crashes when listing collaborators with a Nonetype error;

*Description*:

Similar as #1334.

*To Reproduce*:

Run github sync. I guess this becomes more pronounced on a large enough graph or in conditions when your github quota is low or some other condition is flaky.

*Logs*:

Crash

```
  ...
  File "{PATH}app/intelmodules/github.py", line 321, in sync_repos
    direct_collabs = _get_repo_collaborators_for_multiple_repos(
  File "{PATH}/cartography/intel/github/repos.py", line 160, in _get_repo_collaborators_for_multiple_repos
    (affiliation == 'DIRECT' and repo['directCollaborators']['totalCount'] == 0)):
```

*Screenshots*:

0.96.0

cc: @danbrauer - any chance you have time to help here? maybe adding a retry + sleep of some sort, either with your decorator technique or the older technique. np if you're busy!

achantavy:
### Summary

Fix for #1404. Made this as a separate PR from https://github.com/cartography-cncf/cartography/pull/1405 since I can't push up to the etsy fork.

### Related issues or links

- #1404

### Checklist

**Not applicable** If you are changing a node or relationship:

**Not applicable** If you are implementing a new intel module:

### DIFF
--- a/cartography/intel/github/repos.py
+++ b/cartography/intel/github/repos.py
@@ -14,6 +14,8 @@ from packaging.utils import canonicalize_name
 
 from cartography.intel.github.util import fetch_all
 from cartography.intel.github.util import PaginatedGraphqlData
+from cartography.util import backoff_handler
+from cartography.util import retries_with_backoff
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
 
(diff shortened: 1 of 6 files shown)


cartography/intel/github/repos.py
```diff
@@ -162,14 +164,42 @@ def _get_repo_collaborators_for_multiple_repos(
...
+        def get_repo_collaborators_inner_func(
+                org: str,
+                api_url: str,
+                token: str,
+                repo_name: str,
+                affiliation: str,
+                collab_users: List[dict[str, Any]],
+                collab_permission: List[str],
+        ) -> None:
+            logger.info(f"Loading {affiliation} collaborators for repo {repo_name}.")
```
kledo-lyft, line 179: nit thing: i'm not entirely sure the need and effect of having this func defined inside a for-loop. I don't immediately see the need for it since all the params are passed in it seems like. If so, we can move it out to the package level for clarity

kledo-lyft (approved):
Left a comment but nothing blocking. If it's not possible to move the func out, then you may consider just using variables already defined at the func's scope, like org, collab_users, etc. instead of passing them in again via parameters.

achantavy:
> Left a comment but nothing blocking. If it's not possible to move the func out, then you may consider just using variables already defined at the func's scope, like org, collab_users, etc. instead of passing them in again via parameters.

@kledo-lyft - Good feedback, thanks. How's it now? I think it _might_ be ok to do something like this:

```
@retries_with_backoff(TypeError, 5, backoff_handler)
def _get_repo_collaborators_for_multiple_repos(...):
    ...
    _get_repo_collaborators_inner_func(...)
    ...
```

Think it'd be worth restructuring like this?

PR #1407, opened 2024-12-13
cve_feed: manage http request and add request log

kledo-lyft:
**NOTE**: Turn on "Hide whitespaces" to review

### Summary

We got some memory problem recently especially after NVD API becomes more unstable and retries are more often. This change attempts to reduce the memory leak while many requests are created, also logging the requests to help debug easier as there might be more improvements needed for this to work nicer.

* Added a log statement to indicate when the NIST NVD API is being called, which includes the URL and parameters being used.
* Wrap the `requests.get` call to use a context manager (`with` statement) to ensure the response is properly closed after use.

### DIFF
--- a/cartography/intel/cve/feed.py
+++ b/cartography/intel/cve/feed.py
@@ -86,11 +86,13 @@ def _call_cves_api(url: str, api_key: str | None, params: Dict[str, Any]) -> Dic
     results: Dict[Any, Any] = dict()
 
     while params["resultsPerPage"] > 0 or params["startIndex"] < totalResults:
+        logger.info(f"Calling NIST NVD API at {url} with params {params}")
         try:
-            res = requests.get(
+            with requests.get(
                 url, params=params, headers=headers, timeout=CONNECT_AND_READ_TIMEOUT,
-            )
-            res.raise_for_status()
+            ) as res:
+                res.raise_for_status()
+                data = res.json()
         except requests.exceptions.HTTPError:
             logger.error(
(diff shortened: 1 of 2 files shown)


cartography/intel/cve/feed.py
```diff
@@ -85,30 +85,32 @@ def _call_cves_api(url: str, api_key: str | None, params: Dict[str, Any]) -> Dic
...
+            logger.info(f"Calling NIST NVD API at {url} with params {params}")
+            try:
+                res = session.get(
+                    url, params=params, headers=headers, timeout=CONNECT_AND_READ_TIMEOUT,
+                )
+                res.raise_for_status()
+                data = res.json()
+            except requests.exceptions.HTTPError:
+                logger.error(
+                    f"Failed to get CVE data from NIST NVD API {res.status_code} : {res.text}",
```
cmm-lyft, line 99: I don't think the spikes of memory are falling in this exception case. 
Spikes are not raising an explicit HTTP exception, instead a runtime one (I guess, but I haven't find any logs from kibana / OpenSearch).

cmm-lyft (changes requested):
Could you please elaborate more how is this related to the memory usage spikes?

achantavy (approved):

cartography/intel/cve/feed.py
kledo-lyft: discussing offline since it's related to Lyft

PR #1409, opened 2024-12-13
0.96.2: Fix #1404: allow GitHub sync to continue when we can't list collaborators

### LINKED ISSUE #1404: github sync nonetype error when listing collaborators

achantavy:
Allows the GitHub sync to continue if the user is unable to list repository collaborators.

### Checklist

### DIFF
--- a/cartography/intel/github/repos.py
+++ b/cartography/intel/github/repos.py
@@ -7,6 +7,7 @@ from typing import Dict
 from typing import List
 from typing import Optional
 
+import backoff
 import neo4j
 from packaging.requirements import InvalidRequirement
 from packaging.requirements import Requirement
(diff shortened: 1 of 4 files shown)


kledo-lyft (approved):

cartography/intel/github/repos.py
```diff
@@ -140,22 +140,40 @@ def _get_repo_collaborators_inner_func(
...
-    logger.info(f"Loading {affiliation} collaborators for repo {repo_name}.")
-    collaborators = _get_repo_collaborators(token, api_url, org, repo_name, affiliation)
+) -> dict[str, list[UserAffiliationAndRepoPermission]]:
+    result: dict[str, list[UserAffiliationAndRepoPermission]] = {}
+
+    for repo in repo_raw_data:
+        repo_name = repo['name']
+        repo_url = repo['url']
+
+        if ((affiliation == 'OUTSIDE' and repo['outsideCollaborators']['totalCount'] == 0) or
```
achantavy, line 154: Moved this check to the inner function because the repo object can be None so this should also be wrapped in a retry.

PR #1410, opened 2024-12-14
cve_feed: Correct the retry and sleep time

kledo-lyft:
### Summary

The [current code](https://github.com/cartography-cncf/cartography/blob/4d53bce6d9f3f6703b709b70071cbcc36820a5bb/cartography/intel/cve/feed.py#L73-L113) uses of variable `sleep_time` for both retries and sleep between requests. Furthermore, the `sleep_time` was not reset in the next request, meaning that if the first request increases it to 16 seconds, the second request will continue to take that up to every bigger number (the `retries` count is reset though which makes it worse).

Two changes here:
- The sleep between requests is set to be a dedicated variable `sleep_between_requests`.
- I decided to rewrite the request retry logic more properly using HttpAdapter's retry policy.

Unfortunately, I tried pretty hard to see if we can keep `test_call_cves_api_with_error` which tests the retry logic but it's not very feasible. Also it doesn't make a lot of sense to test something managed by the Session object itself. So I had to drop it.

### Testing

I think existing integ test can confirm the code is still working. The retry change itself will need some manual review to see if it makes sense.

### DIFF
--- a/cartography/intel/cve/feed.py
+++ b/cartography/intel/cve/feed.py
@@ -101,7 +101,7 @@ def _call_cves_api(url: str, api_key: str | None, params: Dict[str, Any]) -> Dic
     with requests.Session() as session:
         _configure_session(session)
         while params["resultsPerPage"] > 0 or params["startIndex"] < totalResults:
-            logger.error(f"Calling NIST NVD API at {url} with params {params}")
+            logger.info(f"Calling NIST NVD API at {url} with params {params}")
             res = session.get(url, params=params, headers=headers, timeout=CONNECT_AND_READ_TIMEOUT)
             res.raise_for_status()
             data = res.json()
--- a/cartography/intel/cve/feed.py
+++ b/cartography/intel/cve/feed.py
@@ -11,10 +11,7 @@ from typing import List
 from typing import Optional
 
 import neo4j
-import requests
 from requests import Session
-from requests.adapters import HTTPAdapter
-from urllib3 import Retry
 
 from cartography.client.core.tx import load
 from cartography.client.core.tx import read_list_of_values_tx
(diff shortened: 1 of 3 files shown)


cartography/intel/cve/feed.py
```diff
@@ -68,49 +70,46 @@ def _map_cve_dict(cve_dict: Dict[Any, Any], data: Dict[Any, Any]) -> None:
...
-                    f"Failed to get CVE data from NIST NVD API {res.status_code} : {res.text}",
-                )
-                retries += 1
-                if retries >= MAX_RETRIES:
-                    raise
-                # Exponential backoff
-                sleep_time *= 2
-                time.sleep(sleep_time)
-                continue
+            logger.error(f"Calling NIST NVD API at {url} with params {params}")
```
kledo-lyft, line 104: or mistake

cartography/intel/cve/feed.py
```diff
@@ -68,49 +70,46 @@ def _map_cve_dict(cve_dict: Dict[Any, Any], data: Dict[Any, Any]) -> None:
...
     else:
-        sleep_time = DELAYED_SLEEP_TIME  # Sleep for 6 seconds between each request to avoid rate limiting
+        sleep_between_requests = DELAYED_SLEEP_TIME
         logger.warning(
-            f"No NIST NVD API key provided. Increasing sleep time to {sleep_time}.",
+            f"No NIST NVD API key provided. Increasing sleep time to {sleep_between_requests}.",
         )
     results: Dict[Any, Any] = dict()
 
     with requests.Session() as session:
```
achantavy, line 101: Instead of creating the session here, maybe we could accept the session as an argument to _call_cves_api like dependency injection. This way _configure_session() can be called somewhere else.

I'm not familiar with this code but from the snippet you pasted in your description, it looks like _call_cves_api is pretty complicated and does a lot of things, so if we need to gut it and redo it, I'm supportive.

heryxpc (approved):
LGTM
Probably the missing test only adds coverage to one single line ` res.raise_for_status()`, that can be skipped if we trust  how `urlib3.Retry` works.

cartography/intel/cve/feed.py
kledo-lyft: the _call_cves_api is now fairly simple now after this change:
https://github.com/cartography-cncf/cartography/blob/fcf6cb11682db1b20eb10f00b536d9a5cd8800b0/cartography/intel/cve/feed.py#L85-L113

about further optimization of the session handling, i guess the most optimal place is to initiate the session in the `start_cve_ingestion` function in the `__init__.py`, yes?

achantavy: > the _call_cves_api is now fairly simple now after this change:

Agree, you improved it a lot.

> about further optimization of the session handling, i guess the most optimal place is to initiate the session in the start_cve_ingestion function in the __init__.py, yes?

Yeah I think so

kledo-lyft: I moved the session to the higher level and passing it into feed.py calls.
There are some more small refactoring improvements to the code in the __init__.py.

I thought of having a singleton session for simplicity but noticed all the calls can further be parallelized using asyncio at some point, and singleton session would require a lot of changes again if we ever wanted to do that. With this version, it should be simpler to do so.

kledo-lyft: well, i turned on auto-merge, it flagged that it's still waiting for your review
then i clicked on update branch, and it updated the branch then merged..

if you want any  more changes, let me know i can do in another PR then.

PR #1411, opened 2024-12-17
Opal Intel module

SecPrez:
Intel module to map in Opal Resource, what those resources grant access, and map who can automatically and manually approve themselves to those resources.

image

### Checklist

- [ x] Update/add unit or integration tests.

- [ x] Update the [schema](https://github.com/lyft/cartography/tree/master/docs/root/modules) and [readme](https://github.com/lyft/cartography/blob/master/docs/schema/README.md).

- [ x] Use the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).

### DIFF
--- a/cartography/data/jobs/cleanup/opal_aws_cleanup.json
+++ b/cartography/data/jobs/cleanup/opal_aws_cleanup.json
@@ -1,0 +1,30 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (n:OpalResource) WHERE n.lastupdated <> $UPDATE_TAG WITH n LIMIT $LIMIT_SIZE DETACH DELETE n",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:OktaGroup)-[r:CAN_AUTO_APPROVE_ACCESS]->(m:OpalResource) WHERE r.lastupdated <> $UPDATE_TAG WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:OktaGroup)-[r:CAN_MANUALLY_APPROVE_ACCESS]->(m:OpalResource) WHERE r.lastupdated <> $UPDATE_TAG WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:OpalResource)-[r:PROVIDES_ACCESS_TO]->(m:AWSRole) WHERE r.lastupdated <> $UPDATE_TAG WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "iterative": true,
+      "iterationsize": 100
+    },
+    {
+      "query": "MATCH (n:OpalResource)-[r:PROVIDES_ACCESS_TO]->(m:OktaGroup) WHERE r.lastupdated <> $UPDATE_TAG WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "iterative": true,
+      "iterationsize": 100
+    }
+  ],
+  "name": "Opal AWS cleanup"
+}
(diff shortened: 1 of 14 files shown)


cartography/intel/opal/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+    opal_to_okta_map = {}
+    for group in groups:
+        opal_group_id = group.get('group_id')
+        okta_group_id = group.get('remote_id')
+        if opal_group_id and okta_group_id:
+            opal_to_okta_map[opal_group_id] = okta_group_id
+    return opal_to_okta_map
+
+
+def get_owner_ids_to_okta_groups_map(
```
achantavy, line 79: Would it make sense to move this and get_opal_groups_to_okta_groups_map to opal_okta.py?

cartography/intel/opal/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    Parse a group dict and add the resource_type field.
+    """
+    group["resource_id"] = group.get("group_id")
+    group["resource_type"] = "OPAL_GROUP"
+    group["remote_resource_name"] = group.get("remote_name")
+    return group
+
+
+def get_all_groups(client: opal.api.GroupsApi) -> List[Dict[str, Any]]:
```
achantavy, line 44: Can we move this to opal_common.py? I'm thinking this file should pretty much only have the start_* and other orchestration functions.

cartography/intel/opal/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+        owner_id = owner.get('owner_id')
+        source_group_id = owner.get('source_group_id')
+        if owner_id and source_group_id:
+            okta_group_id = opal_to_okta_map.get(source_group_id)
+            if okta_group_id:
+                owner_ids_to_okta_groups_map[owner_id] = okta_group_id
+    return owner_ids_to_okta_groups_map
+
+
+def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict[str, Any]) -> None:
```
achantavy, line 105: Can we move this to opal_aws.py?

cartography/data/jobs/cleanup/opal_aws_cleanup.json
```diff
@@ -0,0 +1,30 @@
+{
+  "statements": [
+    {
+      "query": "MATCH (n:OpalResource) WHERE n.lastupdated <> $UPDATE_TAG WITH n LIMIT $LIMIT_SIZE DETACH DELETE n",
```
achantavy, line 4: Let's use the OpalResource model that you created so that we can remove this manual cleanup job. To do this we can introduce a fake "tenant" like in this LastPass PR: https://github.com/cartography-cncf/cartography/pull/1083/files#diff-3b2215e87b2aaf55796fc63a223403d163b97d5bcbb6931d5886802a1f554cd8.

cartography/intel/opal/opal_common.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Load the mapping of OktaGroup to OpalResources they can manually approve access for into Neo4j.
+    """
+    logger.info(f"Loading {len(manual_approved_access)} manual approved access relationships")
+    neo4j_session.run(
+        """
+        UNWIND $manual_approved_access AS access
+        MATCH (g:OktaGroup {id: access.okta_group_id})
+        MATCH (r:OpalResource {id: access.resource_id})
+        MERGE (g)-[rel:CAN_MANUALLY_APPROVE_ACCESS]->(r)
```
achantavy, line 127: Same comment here about making this a rel schema defined on the OpalResource model and then using a read_transaction function to get a list of the IDs you need.

achantavy (changes requested):
Thanks for putting this together, I'm excited. Left some comments about using the data model to represent some of the OpalResource relationships, and how we'd need to add some read_transaction() functions to grab lists of IDs from the graph to make that possible.

I think it's also worth adding in a "fake tenant" node for the OpalResource so that we can automatically clean up the OpalResources -- this is done with the LastPass sync here: https://github.com/cartography-cncf/cartography/pull/1083.

Happy to get on a call if there's things that don't make sense, also happy to help with making these changes too, just let me know.

Issue #1413, opened 2024-12-20
Automatic cleanup jobs for Github Teams and Users


danbrauer:
**Feature request template**

*Automatic cleanup jobs for Github Teams and Users*

*Description*:

As mentioned by @achantavy in this [slack thread](https://cloud-native.slack.com/archives/C080M2LRLDA/p1734712842634609?thread_ts=1733319441.601799&cid=C080M2LRLDA) among other places, Cartography wants to implement cleanup jobs for the following:
* github teams and users:
  * the MEMBER_OF_TEAM relationship
  * the MEMBER relationship
  * the MAINTAINER relationship
* github users and orgs
  * the ADMIN_OF relationship

*Relevant Links*:
* Related PR where this is being worked on: https://github.com/cartography-cncf/cartography/pull/1399
* a [slack thread](https://cloud-native.slack.com/archives/C080M2LRLDA/p1734712842634609?thread_ts=1733319441.601799&cid=C080M2LRLDA) where this issue came up

achantavy:
This is complete as of https://github.com/cartography-cncf/cartography/pull/1419

PR #1419, opened 2024-12-27
#1413: Support automatic cleanups for NodeSchemas without a sub resource

achantavy:
#1413 

Removes a restriction so that cartography's data model can now automatically clean up NodeSchemas that don't have a tenant relationship.

In these cases, the nodes will not be deleted even if they are considered stale, but their stale relationships will be.

We do it this way because cartography syncs assets one tenant at a time, and stale nodes can only be safely deleted if they are tied to a tenant so that we do not erroneously delete nodes attached to other tenants.

- https://github.com/lyft/cartography/issues/1413

We discussed this in the above issue but I'll summarize a bit here: in the case of GitHub, GitHubUsers don't have a true "tenant" relationship with their organization because they can exist independent of their organizations. In this case, it makes sense for a GitHubUser to have a node schema that does not have a tenant rel. If we want to list out all of the users in an organization, then we simply query for the organization and its attached users.

### Checklist


cartography/data/indexes.cypher
```diff
@@ -161,12 +161,8 @@ CREATE INDEX IF NOT EXISTS FOR (n:GCPSubnet) ON (n.id);
 CREATE INDEX IF NOT EXISTS FOR (n:GCPSubnet) ON (n.lastupdated);
 CREATE INDEX IF NOT EXISTS FOR (n:GCPVpc) ON (n.id);
 CREATE INDEX IF NOT EXISTS FOR (n:GCPVpc) ON (n.lastupdated);
-CREATE INDEX IF NOT EXISTS FOR (n:GitHubOrganization) ON (n.id);
```
achantavy, line 164: No longer needed anymore because GitHubOrganizations and GitHubUsers have been migrated to use the data model

achantavy:
cc: @danbrauer

tests/unit/cartography/graph/test_cleanupbuilder.py
```diff
@@ -120,14 +122,49 @@ def test_get_params_from_queries():
...
+        ''',
+        '''
+        MATCH (n:GitHubUser)
+        MATCH (n)-[r:ADMIN_OF]->(:GitHubOrganization)
+        WHERE r.lastupdated <> $UPDATE_TAG
+        WITH r LIMIT $LIMIT_SIZE
+        DELETE r;
+        ''',
+    ]
+    assert clean_query_list(actual_queries) == clean_query_list(expected_queries)
```
kunaals, line 150: nit: fodder for a different PR - might be better to compare sets here instead of lists

kunaals (approved):

PR #1424, opened 2024-12-29
Update AWS EC2 keypair sync to use the new data model

achantavy:
Updates EC2 key pair sync to use the data model.

Fixes an issue that created duplicate key pairs because the older version of the key pair sync had extra node labels but the EC2KeyPairInstance model did not:
image

### Checklist

### DIFF
--- a/cartography/intel/aws/ec2/key_pairs.py
+++ b/cartography/intel/aws/ec2/key_pairs.py
@@ -7,7 +7,7 @@ import neo4j
 
 from .util import get_botocore_config
 from cartography.graph.job import GraphJob
-from cartography.models.aws.ec2.keypairs import EC2KeypairInstanceSchema
+from cartography.models.aws.ec2.keypair_instance import EC2KeypairInstanceSchema
 from cartography.util import aws_handle_regions
 from cartography.util import timeit
 
--- a/cartography/intel/aws/ec2/key_pairs.py
+++ b/cartography/intel/aws/ec2/key_pairs.py
@@ -1,13 +1,14 @@
 import logging
+from typing import Any
 from typing import Dict
-from typing import List
 
 import boto3
 import neo4j
 
 from .util import get_botocore_config
+from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
-from cartography.models.aws.ec2.keypair_instance import EC2KeypairInstanceSchema
+from cartography.models.aws.ec2.keypair import EC2KeyPairSchema
 from cartography.util import aws_handle_regions
 from cartography.util import timeit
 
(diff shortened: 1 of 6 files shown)


achantavy:
@ramonpetgrave64 - tagged you on this one since this was the very first module you ever contributed and you have the most context :-)

cartography/intel/aws/ec2/key_pairs.py
```diff
@@ -16,42 +17,44 @@
...
+        region: str,
+        current_aws_account_id: str,
+) -> list[dict[str, Any]]:
+    transformed_key_pairs = []
+    for key_pair in key_pairs:
+        key_name = key_pair["KeyName"]
+        transformed_key_pairs.append({
+            'KeyPairArn': f'arn:aws:ec2:{region}:{current_aws_account_id}:key-pair/{key_name}',
+            'KeyName': key_name,
+            'KeyFingerprint': key_pair.get("KeyFingerprint"),
```
kunaals, line 36: Are there every situations where we wouldn't see a fingerprint? My intuition is no

kunaals: And as such maybe we shouldn't use a `.get` so that we fail loudly?

cartography/intel/aws/ec2/key_pairs.py
```diff
@@ -16,42 +17,44 @@
...
-        )
+    # Load EC2 keypairs as known by describe-key-pairs
+    load(
+        neo4j_session,
+        EC2KeyPairSchema(),
+        data,
+        Region=region,
+        AWS_ID=current_aws_account_id,
+        lastupdated=update_tag,
+    )
```
kunaals, line 57: I recognize that convention within Cartography is to have a `load_ec2_blah` method but this wrapping of `load` feels superfluous. Is the only benefit of wrapping this invocation of `load` the @timeit on `load_ec2_key_pairs`?

cartography/models/aws/ec2/keypair_instance.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@dataclass(frozen=True)
+class EC2KeyPairInstanceToAWSAccount(CartographyRelSchema):
+    target_node_label: str = 'AWSAccount'
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'id': PropertyRef('AWS_ID', set_in_kwargs=True)},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
```
kunaals, line 36:
```suggestion
    rel_label: str = 'RESOURCE'
```

cartography/models/aws/ec2/keypair_instance.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@dataclass(frozen=True)
+class EC2KeyPairInstanceToEC2Instance(CartographyRelSchema):
+    target_node_label: str = 'EC2Instance'
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'id': PropertyRef('InstanceId')},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "SSH_LOGIN_TO"
```
kunaals, line 52:
```suggestion
    rel_label: str = 'SSH_LOGIN_TO'
```

cartography/intel/aws/ec2/key_pairs.py
achantavy: Sometimes yeah. AWS APIs can be inconsistent and sometimes return None on mandatory fields. Since fingerprint isn't used as a key anywhere, it's ok for it to be null. If it as for the keyname or arn then yeah we should fail loudly and not use .get().

describe-ec2-instances returns keypair information from a given ec2 instance and that [only returns an arn and keyname](https://github.com/cartography-cncf/cartography/blob/f08d85223b40b9f25a2a4ab6b26327e3d5b79a43/cartography/models/aws/ec2/keypair_instance.py#L17-L21), so having a fingerprint is an extra bit of information that only describe-key-pairs knows.

cartography/intel/aws/ec2/key_pairs.py
achantavy: True true. When refactoring I kept the convention of having the get+transform+load even though I built a standard loader function. This is mostly out of convention but it's a bit extra. Kind of helps with adding a logger call and as you mentioned having a timeit decorator. This is sort of an informal "interface" on a sync module where we have to have a get+transform+load even when the load itself is mostly a wrapper for a lower level call.

cartography/models/aws/ec2/keypair_instance.py
achantavy: Maybe we should adopt a `black` or `pre-commit` rule that automatically handles the quote stuff instead of handling this one by one. I think there's inconsistencies all over the place and there's some situations to switch between `"` and `'` 😅

cartography/intel/aws/ec2/key_pairs.py
ramonpetgrave64: There's also the analysis job to attach keypairs to other keypairs, and mark them as user-uploaded.
 * https://github.com/cartography-cncf/cartography/blob/aacaba9fa1b4f74db8eb6fcb8c3e6dd4b5588809/cartography/data/jobs/analysis/aws_ec2_keypair_analysis.json#L19-L28

Aside, We can probably build this functionality into NodeSchema.

ramonpetgrave64 (approved):

cartography/intel/aws/ec2/key_pairs.py
achantavy: Another aside, I'm wondering if there is a good general purpose way to abstract it so that an analysis job writer doesn't have to write the cleanup step.

PR #1433, opened 2025-01-02
Skip database calls in core.tx.load() if data list is empty

achantavy:
If the list passed to load() is empty, we now return early to save time from checking indexes, generating a query string, and talking to the database.

Also fixes this for AWS' resourcegroupstaggingapi sync.

Since many cartography users likely have nothing in many regions, we might as well save ourselves some network and database calls.

### Checklist


ramonpetgrave64 (approved):
Looks good. The calls are nested within batch(), so empty lists won’t necessarily trigger database connections. But it’s still good to skip creating the indexes when unnecessary.

PR #1437, opened 2025-01-06
CI: Wait for PyPI to publish before building Dockerfile

achantavy:
When making a new release, we 
1. publish the release to PyPI
2. install it from PyPI to the Dockerfile
3. publish that Dockerfile to GHCR

After publishing to PyPI however we need a few more seconds for the package to be available. Currently we skip over this and the latest release is not available, causing the build to fail.
image
https://github.com/cartography-cncf/cartography/actions/runs/12577359450/job/35054696773

This PR makes it so that we now wait 10 seconds for PyPI before attempting to build the image.

### DIFF
--- a/.github/workflows/publish-to-ghcr-and-pypi.yml
+++ b/.github/workflows/publish-to-ghcr-and-pypi.yml
@@ -73,10 +73,12 @@ jobs:
           # for more detailed information.
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      # Get the cartography version from the GitHub release event metadata
-      - name: Extract version
+      # Get the cartography version from the GitHub release event metadata and sleep to wait for PyPI to finish publishing
+      - name: Extract version and sleep
         id: version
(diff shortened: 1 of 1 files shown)


chandanchowdhury (changes requested):
Setting up job dependency using [needs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) is a better solution.

achantavy:
> Setting up job dependency using [needs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) is a better solution.

If I understand correctly, that doesn't solve it though because right now the jobs already do happen one and a time and in order, so the GHCR step already happens right after the PyPI step is done, but there is a variable amount of time after the PyPI publish step until the package is actually available and servable from PyPI servers. A more elegant way would be to add a curl or something, but I didn't want to overengineer this.

chandanchowdhury (approved):

Issue #1439, opened 2025-01-14
Correctly assign AWS account in AWS Inspector findings


heryxpc:
**Bug Template**

*Title*: *Correctly assign AWS account in AWS Inspector findings*

*Description*:
Current behavior assigns AWS Inspector resources to the account that allowed access to the sync.
However, in case Inspector is configured with a delegated administrator for Amazon Inspector, the finding can be found on a different AWS account.
The expected behavior should always create findings like:
`(a:AWSAccount)-[:RESOURCE]->(n:AWSInspectorFinding)-[:AFFECTS]->(i:EC2Instance)<-[:RESOURCE]-(a)`

*To Reproduce*:
Run AWS Inspector sync module after a delegated administrator is configured to gather Inspector findings from another account.

*Logs*:
N/A

*Screenshots*:
Expected
![Screenshot 2025-01-13 at 6 18 13 p m](https://github.com/user-attachments/assets/de793678-3889-4555-98b0-bb8123b54fee)

Error cases
![Screenshot 2025-01-13 at 6 22 23 p m](https://github.com/user-attachments/assets/0711149f-36fa-4c72-a65e-2c55486c04a1)

*Please complete the following information:*:
>  - Cartography release version or commit hash [0.98.0rc5]
>  - Python version: [3.10.0]
>  - OS (feel free to omit this if you don't think it's relevant to your issue): [e.g. Ubuntu bla bla, OSX bla bla]

*Additional context*:
Might not cause an issue, but `(:AWSAccount)-[:RESOURCE]->(:AWSInspectorFinding)` is a relationship used to build automated clean up jobs. Since the resource was synced using a different AWSAccount, it might be a problem during cleanup.

achantavy:
Ah, I understand. One possible solution would be for the AWS Inspector finding to not be attached to an account.

The downside here is that the Inspector node won't be automatically cleaned up, but it will get detached from all of its associations. We could mitigate this by having an out of band process clean up these "island" Inspector Findings after a given amount of time.

The benefit is that the Inspector Finding would only be attached to instances that are relevant to it.

Lemme know your thoughts and we can talk in detail.

heryxpc:
> The downside here is that the Inspector node won't be automatically cleaned up, but it will get detached from all of its associations. We could mitigate this by having an out of band process clean up these "island" Inspector Findings after a given amount of time.
> The benefit is that the Inspector Finding would only be attached to instances that are relevant to it.

[...]

PR #1442, opened 2025-01-17
Inspector to account relationship for delegated accounts

heryxpc:
Add relationship to `:AWSAccount` and `:AWSInspectorFinding` when the finding is ingested via a delegated account.

- https://github.com/cartography-cncf/cartography/issues/1439
- https://docs.aws.amazon.com/inspector/latest/user/admin-member-relationship.html
- https://docs.aws.amazon.com/organizations/latest/userguide/services-that-can-integrate-inspector2.html

### Checklist

### DIFF
--- a/docs/root/modules/aws/schema.md
+++ b/docs/root/modules/aws/schema.md
@@ -192,16 +192,17 @@ Representation of an AWS [Inspector Finding](https://docs.aws.amazon.com/inspect
     ```
     (AWSInspectorFinding)-[:AFFECTS]->(ECRImage)
     ```
-- AWSInspectorFinding may be delegated from AWSAccounts.
-
-        ```
-        (AWSAccount)-[DELEGATE]->(AWSInspectorFinding)
-        ```
-- AWSInspectorFindings belong to AWSAccounts.
+- AWSInspectorFindings managed by AWSAccount.
 
         ```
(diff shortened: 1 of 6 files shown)


docs/root/modules/aws/schema.md
```diff
@@ -192,7 +192,11 @@ Representation of an AWS [Inspector Finding](https://docs.aws.amazon.com/inspect
     ```
     (AWSInspectorFinding)-[:AFFECTS]->(ECRImage)
     ```
+- AWSInspectorFinding may be delegated from AWSAccounts.
```
achantavy, line 195: Can we add a detail that describes how this is different from the :RESOURCE relationship?

heryxpc: Sure, let me add the reference regarding delegated accounts and their findings.

heryxpc: While trying to document this clearly I noticed that `DELEGATE` was not a great word to describe the relationship and moved it to `MEMBER` since it is closer to the definition from AWS: https://docs.aws.amazon.com/inspector/latest/user/admin-member-relationship.html
> The delegated administrator can also view and retrieve information for their own account and any associated member. 
...
> A member account can view and retrieve information about their account in Amazon Inspector, while settings for their account are managed by the delegated administrator.

achantavy (approved):

PR #1446, opened 2025-02-02
Refactor AWS API gateway to use data model

achantavy:
- Refactors API gateway to use data models, adding sub resource relationships for api gateway stages, certificates, and resources to allow for automatic cleanups.
- Adds tests for the sync* function, ensures that existing ones work.

Screenshot 2025-02-02 at 3 08 40 PM

This PR can be (roughly) reviewed commit by commit to see how each function was refactored one by one while keeping the tests working.

### Checklist


cartography/intel/aws/apigateway.py
```diff
@@ -107,222 +112,145 @@ def get_rest_api_policy(api: Dict, client: botocore.client.BaseClient) -> Any:
     return policy
```
kunaals, line 112: GH wont let me comment on line 107 but `get_rest_api_policy()` doesn't use the boto client param, can update function header

kunaals (approved):

PR #1448, opened 2025-02-03
CHORE: Update GitHub actions dependencies

jychp:
While investigating why the latest Scorecard workflows failed (https://github.com/cartography-cncf/cartography/actions/runs/13084015053 and https://github.com/cartography-cncf/cartography/actions/runs/13090489797), I noticed that several GitHub Actions were outdated, particularly:
- **Upload Artifact** in OSSF Scorecard has been deprecated since November 2024.
- **Cache V2** in Sphinx relies on an API that has been unavailable since February 1st, 2025.
I checked, and we use these actions in a fairly standard way, with minimal breaking changes. These updates should not impact the workflows.

I also think it would be beneficial to enable **Dependabot**, but I haven't included it in this PR while we discuss the topic.

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -1,14 +1,12 @@
----
-# https://github.com/marketplace/actions/ossf-scorecard-action
+# Based on: https://github.com/marketplace/actions/ossf-scorecard-action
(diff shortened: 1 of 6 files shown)


chandanchowdhury (approved):
Fantastic! Thank you Jeremy :)

Issue #1349, opened 2024-08-28
[Feature Request] Add OpenSSF Scorecard information


heryxpc:
*Title*: *Ingest OpenSSF scorecard information*

*Description*:
Create a new node `OpenSSFScorecard` with Github's scorecard checks for every project ingested.
An OpenSSF scorecard is a serious of checks that can be useful to surface the security posture of a project.
See for example https://scorecard.dev/viewer/?uri=github.com/lyft/cartography
Checks can be used as an overall score or in an individual form, for example to determine if the project is actively maintained (which could reflect if cadence of security fixes).

The information can be queried using a public API, for example https://api.securityscorecards.dev/projects/github.com/lyft/cartography

This could be a separate intel module or a submodule from https://github.com/lyft/cartography/blob/master/cartography/intel/github

[optional *Relevant Links*:]
https://openssf.org/projects/scorecard/
https://openssf.org/blog/2024/04/17/beyond-scores-with-openssf-scorecard-granular-structured-results-for-custom-policy-enforcement/
https://api.securityscorecards.dev/

PR #1450, opened 2025-02-03
CHORE: add OpenSSF badges & pin dependencies versions

jychp:
### Summary
This PR:

- Adds OpenSSF badges (Security ScoreCard & BestPractices) to improve project security visibility.
- Adds a badge to display the status of the latest build workflow.
- Pins GitHub Actions versions using SHA (OpenSSF Security compliance).
- Pins the base image versions in Dockerfiles (OpenSSF compliance).

_Notes: This PR does not pin Python dependencies. Doing so requires defining governance around dependency updates, particularly regarding Dependabot integration. A separate discussion should be initiated to address this._

### Related Issues
- https://github.com/cartography-cncf/cartography/issues/1349

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # This is a thin distribution of the cartography software.
 # It is published at ghcr.io.
-FROM python:3.10-slim
+FROM python:3.10-slim@sha256:a636f5aafba3654ac4d04d7c234a75b77fa26646fe0dafe4654b731bc413b02f
 
 # Default to ''. Overridden with a specific version specifier e.g. '==0.98.0' by build args or from GitHub actions.
 ARG VERSION_SPECIFIER
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -4,7 +4,7 @@
 # - This dockerfile will get called with .cache as a volume mount.
 # - The current working directory on the host building this container
 #   is the cartography source tree from github.
-FROM python:3.10-slim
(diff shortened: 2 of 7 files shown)


dev.Dockerfile
```diff
@@ -4,7 +4,7 @@
 # - This dockerfile will get called with .cache as a volume mount.
 # - The current working directory on the host building this container
 #   is the cartography source tree from github.
-FROM python:3.10-slim
+FROM python:3.10-slim@sha256:a636f5aafba3654ac4d04d7c234a75b77fa26646fe0dafe4654b731bc413b02f
```
achantavy, line 7: What's the best practice for updating the pin on the image?

achantavy (approved):

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # This is a thin distribution of the cartography software.
 # It is published at ghcr.io.
-FROM python:3.10-slim
+FROM python:3.10-slim@sha256:a636f5aafba3654ac4d04d7c234a75b77fa26646fe0dafe4654b731bc413b02f
```
achantavy, line 3: Same here with the pin, what's the best practice on updating?

dev.Dockerfile
jychp: I'm used to doing this with Dependabot.
It works really well, including for pinned versions.
You just need to tweak the default configuration a bit (to avoid getting 20k PRs per week, etc.).
I'm currently preparing a PR to enable Dependabot.

PR #1451, opened 2025-02-03
chore: add dependabot configuration

jychp:
### Summary
This PR adds Dependabot configuration for updating dependencies in GitHub Actions, Dockerfiles, and Python libraries.

Updates will be checked weekly, with PRs:
- Prefixed with `chore:`
- Labeled as `dependencies`

**GitHub Actions**
- Minor and patch updates will be grouped into a single PR
- Major updates will be handled in dedicated PRs

**Python**
- Minor and patch updates will be grouped into a single PR
- Major updates will be handled in dedicated PRs
- The strategy is "loose": requirements files will only be updated if necessary (e.g., if the current version constraint is `^1.0.0` and already includes the new version `1.2.0`, no update will be made)

**Docker**
- Only patch updates will be applied to maintain the image on Python 3.10

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,7 +24,7 @@ updates:
     target-branch: "master"
     commit-message:
       # Prefix all commit messages with "chore: "
-      prefix: "chore"
+      prefix: "chore: "
     groups:
       # Group all minor and patch updates for Python
       # Keep individual updates for major updates
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -47,7 +47,7 @@ updates:
     target-branch: "master"
     commit-message:
       # Prefix all commit messages with "chore: "
-      prefix: "chore"
+      prefix: "chore: "
     # Ignore major & minor updates to stay on python:3.10
     # TODO: Remove when ready to upgrade to python:3.11
     ignore:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -49,6 +49,6 @@ updates:
     # TODO: Remove when ready to upgrade to python:3.11
     ignore:
       - dependency-name: "*"
-      - update-types:
+        update-types:
         - version-update:semver-major
         - version-update:semver-minor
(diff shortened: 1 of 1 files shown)


achantavy:
Thanks for taking this on. We should merge this in as soon as possible, but before we do I wanted to share one skeleton in the closet that I know of: The okta sync relies on an alpha version of the Okta Python client. Upgrading Okta sync to use the up to date version will cause things to break. As long as we don't touch that, everything else should be ok.

jychp:
@achantavy good catch, I added an explicit ignore block for this one with a comment that link the issue.

When dependabot will create the first PR, the tests will give us more information on broken things and I will be able to update the configuration accordingly (an create the issues to adapt the code ^^')

.github/dependabot.yml
```diff
@@ -0,0 +1,57 @@
...
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    target-branch: "master"
+    commit-message:
+      # Prefix all commit messages with "chore: "
+      prefix: "chore"
```
kunaals, line 11: Looking at this example looks like we should include the ": " as well https://github.com/dependabot/dependabot-core/issues/10627
```suggestion
      prefix: "chore: "
```

.github/dependabot.yml
```diff
@@ -0,0 +1,57 @@
...
+          - "patch"
+  # Maintain dependencies for Python
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    target-branch: "master"
+    commit-message:
+      # Prefix all commit messages with "chore: "
+      prefix: "chore"
```
kunaals, line 27:
```suggestion
      prefix: "chore: "
```

.github/dependabot.yml
```diff
@@ -0,0 +1,57 @@
...
+      - dependency-name: "okta"
+  # Maintain dependencies for Docker
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    target-branch: "master"
+    commit-message:
+      # Prefix all commit messages with "chore: "
+      prefix: "chore"
```
kunaals, line 50:
```suggestion
      prefix: "chore: "
```

kunaals (commented):
Just one nit :)

kunaals (approved):

PR #1465, opened 2025-02-10
feat: allow update schema without updating nodes

jychp:
### Summary
This PR adds an `only_relationships` flag to the `build_ingestion_query()` function.  
The goal is to update the relationships between nodes without risking overwriting their properties.  

Here’s the use case where I need this:  
- I retrieve users from GitHub  
- I retrieve repositories from GitHub  
- I also retrieve User/Repository links  

To avoid a complex transformation that could be hard to maintain, I believe the simplest approach is:  
- One ingestion call with all user data  
- One ingestion call with all repository data  
- One ingestion call on Repositories (which contains the link) with minimal information (Repo ID + User ID as defined in the model)  

The flag allow to run this last ingestion call without deleting all properties of Repo node

### DIFF
--- a/cartography/graph/querybuilder.py
+++ b/cartography/graph/querybuilder.py
@@ -387,7 +387,8 @@ def build_ingestion_query(
         if property_ref is None:
             raise ValueError(
                 f"build_ingestion_query() failed: The CartographyNodeSchema {node_schema.__class__.__name__} does not "
-                "have a 'lastupdated' property. Please ensure that the 'lastupdated' property is defined on the node.")
+                "have a 'lastupdated' property. Please ensure that the 'lastupdated' property is defined on the node.",
+            )
         query_template = Template(
             """
             UNWIND $DictList AS item
--- a/cartography/graph/querybuilder.py
+++ b/cartography/graph/querybuilder.py
@@ -353,6 +353,7 @@ def filter_selected_relationships(
 def build_ingestion_query(
         node_schema: CartographyNodeSchema,
         selected_relationships: Optional[Set[CartographyRelSchema]] = None,
+        only_relationships: bool = False,
 ) -> str:
     """
     Generates a Neo4j query from the given CartographyNodeSchema to ingest the specified nodes and relationships so that
@@ -364,6 +365,8 @@ def build_ingestion_query(
     If selected_relationships is None (default), then we create a query using all RelSchema specified in
     node_schema.sub_resource_relationship + node_schema.other_relationships.
     If selected_relationships is the empty set, we create a query with no relationship attachments at all.
+    :param only_relationships: If True, the query will only attach relationships and will not update node properties.
+    This is usefull when you only have partial node data and you want to attach relationships to existing nodes.
     :return: An optimized Neo4j query that can be used to ingest nodes and relationships.
     Important notes:
     - The resulting query uses the UNWIND + MERGE pattern (see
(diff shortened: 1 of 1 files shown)


achantavy:
How can we ensure that `only_relationships` is only used when properties of a node have not changed? I don't want to allow for the possibility where the `update_tag` is current but the node properties on multiple nodes are out of date.

I'm not quite understanding the scenario, is the PR trying to optimize for compute time or something else?

cc: @ramonpetgrave64 - might have some thoughts since you've been deep in the weeds here before

jychp:
Indeed, after more in depth tests, may be I can do it differently, will keep this PR as draft until I figure out if it's realy needed or if it's only due to a bad model choice I made.

ramonpetgrave64:
> The flag allow to run this last ingestion call without deleting all properties of Repo node

Is it true that old properties are removed with the query builders, or the cleanup jobs? If it's only the cleanup jobs that remove the old properties, maybe your flag can somehow make the cleanup jobs skip.

achantavy:
No, old properties are preserved except in the case of analysis jobs

achantavy:
Sorry, which statement in that file? I read through but I'm not sure which one.

Cleanup jobs only delete stale nodes and stale relationships. They never touch properties of a node or properties of a relationship. 

The only case where properties are changed in a manner similar to this is analysis jobs. These add properties to a node or relationship -- for example with internet exposure ([full doc](https://cartography-cncf.github.io/cartography/dev/writing-analysis-jobs.html)), we [remove some attributes](https://github.com/cartography-cncf/cartography/blob/e1925497f1213ef02c8887187a42d875e10d5386/cartography/data/jobs/analysis/aws_ec2_asset_exposure.json#L4-L19) before running the analysis; this is a kind of "freshness step".

jychp:
My understanding was there are 2 nodes because in some cases the API does not return "has_mfa_enable", so the value will be lost (if it's present at the first query but not atthe second one).

But you said that updating a node does not "delete" a property if the property is missing from the data dict right ?

So there is no reason for having this two nodes (at least regarding the nodes properties)

ramonpetgrave64:
@achantavy What if we addd `coalesce` to the querybuilder. This means that truly stale properties never get deleted. 

^ I feel like we discussed this but still decided on having it like the way you said it:  "GitHubUsers as known by API $Y", as well as "GitHubUsers as known by API $X".

jychp:
@achantavy OK, so I understood well how it's work

I think this flag could be usefull when we have:
- one endpoint /entityA listing NodeA (with full properties id, propA, probB)
- one endpoint /entityB listing NodeB (with full properties id, propC, propD, propE)
- one endpoint /x listing NodeA with linked NodeB with only partial properties (lets say only id for each nodes)

With the current design I need to call the third endpoint and do a painfull tranformation
With this flag I can just make a call to the build function with partial data (only IDs, missing properties wont be overridden)

jychp:
@achantavy, maybe this is not the best approach.  

Here’s an example of the issue I encountered while migrating GitHub repositories to the new data model:  
- Users are already synced.  
- When adding repositories, I need to duplicate each dictionary for every relationship to an owner or a direct/outside collaborator.  
  - Example: For `RepoA` with `CollabA` and `CollabB`, I have to make two queries: `(RepoA, CollabA)` and `(RepoA, CollabB)`.  
- This limitation is acceptable for a small number of nodes, but in my case, uploading 120 repositories requires **21,932 queries** to import all collaborators and owners.  

Maybe we need a query that **only creates relationships** without updating existing nodes.  

For example, in GitHub:  
- `GitHubRepository` nodes already exist.  
- `GitHubUser` nodes already exist.  
- I believe we could construct a query that links **N repositories to N users** in a single command by iterating over two lists.  

I used GitHub as an example, but this issue will occur for all **highly connected** nodes.  

What do you think?

jychp:
Close in favor of #1502

Issue #1304, opened 2024-05-13
Setup black as the formatter for the repo


serge-wq:
*Description*:

The goal is to set up black as the code formatter for the whole repo to have a more standard way of formatting it. Current linter configuration throws what it seems to be arbitrary errors during development.

PR #1466, opened 2025-02-10
chore: better PEP8 formating

jychp:
### Summary
This PR includes several maintenance tasks:
- Updating the links to the repository after the migration (from Lyft to CNCF).
- Adding Black as the code formatter, which I believe is the most popular formatter for Python. It is much more consistent than what was previously in place (and having clean code that conforms to PEP8 is important!).
- Removing unnecessary hooks (since they are now included in Black) and fixing compatibility issues.
- Formatting the code.

**NOTE**: The PR is quite large due to the formatting changes, so I strongly recommend reviewing the PR commit by commit.

### Related issues or links
- https://github.com/cartography-cncf/cartography/issues/1304


achantavy:
Hey, thanks for taking on the big heavy debt tasks.

Can we defer the poetry decision to a separate PR? (Tbh I really love `uv` but don't want to start a holy war here haha.)

That said I totally agree with using `black`. 

The other big area of debt includes one by one removing the ignore_errors = True in setup.cfg. I say one by one because doing more than one at once will cause us to introduce too many changes.

jychp:
Ok @achantavy , let's go to the octagon to debate UV vs Poetry :)  
More seriously, since version 2.x, Poetry no longer requires special configurations (notably, dependencies can be defined at the project level, as is currently the case). Supporting Poetry does not undermine the current setup nor the possibility of using UV later.

.pre-commit-config.yaml
```diff
@@ -13,31 +13,19 @@ repos:
...
     rev: v3.19.1
     hooks:
     -   id: pyupgrade
         args: [--py36-plus]
--   repo: https://github.com/asottile/add-trailing-comma
-    rev: v3.1.0
-    hooks:
-    -   id: add-trailing-comma
-        args: [--py36-plus]
--   repo: https://github.com/asottile/reorder_python_imports
```
achantavy, line 36: I like the reorder_python_imports for this reason: https://github.com/asottile/reorder-python-imports?tab=readme-ov-file#why-this-style

How does black behave?

achantavy (approved):

PR #1468, opened 2025-02-11
Handle each image id separately

heryxpc:
### Summary
Previous code handled any client exception at `client.describe_images(ImageIds=[image_ids])['Images']` as a full failure.
If for any reason (e.g. the image id is no longer found) the API call failed, all the id's from the set were skipped.

This PR splits the calls one per image id. There is a small performance degradation due to latency, but it was not perceptible during my tests with +200 images.

This was tried to fix at 
- https://github.com/cartography-cncf/cartography/pull/1441
But it was reverted since didn't fix the root cause.

Logs after:
```
2025-02-11 14:26:50,379 - cartography.intel.aws.ec2.images - INFO - Syncing images for region 'us-east-1' in account '277829364062'.
2025-02-11 14:27:00,234 - cartography.intel.aws.ec2.images - WARNING - Failed retrieve image id ami-03be86ff62fe082ad for region - us-east-1. Error - An error occurred (InvalidAMIID.NotFound) when calling the DescribeImages operation: The image id '[ami-03be86ff62fe082ad]' does not exist
2025-02-11 14:27:02,763 - cartography.intel.aws.ec2.images - INFO - Syncing images for region 'us-east-2' in account '277829364062'.
2025-02-11 14:27:03,143 - cartography.intel.aws.ec2.images - INFO - Syncing images for region 'us-west-1' in account '277829364062'.
2025-02-11 14:27:03,561 - cartography.intel.aws.ec2.images - INFO - Syncing images for region 'us-west-2' in account '277829364062'.
```

Example error before the change using https://github.com/cartography-cncf/cartography/pull/1441:
```
Failed to retrieve public images for region - us-east-1. Error - An error occurred (InvalidAMIID.NotFound) when calling the DescribeImages operation: The image ids '[ami-0d4e924e774f7d54a, ami-0b3126f23ad61defc, ami-0644dffc72cee3960, ami-09ca115a458c43d7a, ami-064a36b42c8203d8d, ami-073cf29d5b46bb2ba, ami-0cdc91d6f6d45591f, ami-03fe03bbcbfde1c1e, ami-0cfd0e38c0448c98c, ami-0af76f69b0f541ca0, ami-04c6260c832afa746, ami-0c346f96131f30a3b, ami-0fb9c9fc5db3e5a84, ami-07a6b49ef075534f9, ami-06dfde5c1a402f710, ami-072309c05e170f9a4]' do not exist
```

### Checklist

### DIFF
--- a/cartography/intel/aws/ec2/images.py
+++ b/cartography/intel/aws/ec2/images.py
@@ -8,6 +8,7 @@ import neo4j
 from botocore.exceptions import ClientError
 
 from cartography.client.core.tx import load
+from cartography.client.core.tx import read_list_of_values_tx
 from cartography.graph.job import GraphJob
 from cartography.intel.aws.ec2.util import get_botocore_config
 from cartography.models.aws.ec2.images import EC2ImageSchema
@@ -20,21 +21,27 @@ logger = logging.getLogger(__name__)
 @timeit
 def get_images_in_use(neo4j_session: neo4j.Session, region: str, current_aws_account_id: str) -> List[str]:
     get_images_query = """
+    CALL {
     MATCH (:AWSAccount{id: $AWS_ACCOUNT_ID})-[:RESOURCE]->(i:EC2Instance)
-    WHERE i.region = $Region
-    RETURN DISTINCT(i.imageid) as image
-    UNION
+    WHERE i.region = $Region AND i.imageid IS NOT NULL
+    RETURN i.imageid AS image
+    UNION ALL
     MATCH (:AWSAccount{id: $AWS_ACCOUNT_ID})-[:RESOURCE]->(lc:LaunchConfiguration)
-    WHERE lc.region = $Region
-    RETURN DISTINCT(lc.image_id) as image
-    UNION
+    WHERE lc.region = $Region AND lc.image_id IS NOT NULL
+    RETURN lc.image_id AS image
+    UNION ALL
     MATCH (:AWSAccount{id: $AWS_ACCOUNT_ID})-[:RESOURCE]->(ltv:LaunchTemplateVersion)
-    WHERE ltv.region = $Region
-    RETURN DISTINCT(ltv.image_id) as image
+    WHERE ltv.region = $Region AND ltv.image_id IS NOT NULL
+    RETURN ltv.image_id AS image
+    }
+    RETURN DISTINCT image;
     """
-    results = neo4j_session.run(get_images_query, AWS_ACCOUNT_ID=current_aws_account_id, Region=region)
-    images = {r['image'] for r in results if r['image']}
-    return list(images)
+    result = read_list_of_values_tx(
+        neo4j_session, get_images_query,
+        AWS_ACCOUNT_ID=current_aws_account_id, Region=region,
+    )
+    images = [str(image) for image in result]
+    return images
 
 
 @timeit
--- a/cartography/intel/aws/ec2/images.py
+++ b/cartography/intel/aws/ec2/images.py
@@ -49,7 +49,7 @@ def get_images(boto3_session: boto3.session.Session, region: str, image_ids: Lis
         logger.warning(f"Failed retrieve self owned images for region - {region}. Error - {e}")
     images.extend(self_images)
     if image_ids:
-        ids_retrieved = {image['ImageId'] for image in images}
+        self_image_ids = {image['ImageId'] for image in images}
         ids_pending = [id for id in image_ids if id not in ids_retrieved]
         # Go one by one to avoid losing all images if one fails
         for image in ids_pending:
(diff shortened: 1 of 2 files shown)


cartography/intel/aws/ec2/images.py
```diff
@@ -33,33 +33,31 @@ def get_images_in_use(neo4j_session: neo4j.Session, region: str, current_aws_acc
     RETURN DISTINCT(ltv.image_id) as image
     """
     results = neo4j_session.run(get_images_query, AWS_ACCOUNT_ID=current_aws_account_id, Region=region)
-    images = []
-    for r in results:
-        images.append(r['image'])
-    return images
+    images = {r['image'] for r in results if r['image']}
+    return list(images)
```
heryxpc, line 37: This avoids a further duplicates check as well as `NoneType` check.

cartography/intel/aws/ec2/images.py
```diff
@@ -33,33 +33,31 @@ def get_images_in_use(neo4j_session: neo4j.Session, region: str, current_aws_acc
...
-            for image in images_in_use:
-                if image["ImageId"] not in _ids:
-                    images.append(image)
     except ClientError as e:
-        logger.warning(f"Failed retrieve images for region - {region}. Error - {e}")
+        logger.warning(f"Failed retrieve self owned images for region - {region}. Error - {e}")
+    images.extend(self_images)
+    if image_ids:
+        ids_retrieved = {image['ImageId'] for image in images}
+        ids_pending = [id for id in image_ids if id not in ids_retrieved]
```
achantavy, line 53: I would move this set membership logic to the for-loop to save a minor bit of memory and time.

```
for image_id in ids_pending:
  if image_id in ids_retrieved:
    continue
  ... do the rest ...
```

cartography/intel/aws/ec2/images.py
```diff
@@ -33,33 +33,31 @@ def get_images_in_use(neo4j_session: neo4j.Session, region: str, current_aws_acc
...
+        # Go one by one to avoid losing all images if one fails
+        for image in ids_pending:
+            try:
+                public_images = client.describe_images(ImageIds=[image])['Images']
+                images.extend(public_images)
+            except ClientError as e:
+                logger.warning(f"Failed retrieve non-self image for region - {region}. Error - {e}")
```
achantavy, line 60: I think we should log the specific image id here

achantavy (changes requested):

cartography/intel/aws/ec2/images.py
heryxpc: Since each query is independent because of the `UNION` the case could exist where the same image is used by both Launch Templates and EC2 images.
There is a further validation in `get_images` but using the `set` reduces the chances of repeated images.
There was also the case of retuning a `NULL` which translates into a `None` and caused a bug in the past.

achantavy: Non-blocker, but I think ideally this filtering logic should be pushed up to the database as much as possible to reduce code complexity.

Maybe we do something like

```cypher
CALL {
    MATCH (:AWSAccount{id: $AWS_ACCOUNT_ID})-[:RESOURCE]->(i:EC2Instance)
    WHERE i.region = $Region AND i.imageid IS NOT NULL
    RETURN i.imageid AS image
    UNION ALL
    MATCH (:AWSAccount{id: $AWS_ACCOUNT_ID})-[:RESOURCE]->(lc:LaunchConfiguration)
    WHERE lc.region = $Region AND lc.image_id IS NOT NULL
    RETURN lc.image_id AS image
    UNION ALL
    MATCH (:AWSAccount{id: $AWS_ACCOUNT_ID})-[:RESOURCE]->(ltv:LaunchTemplateVersion)
    WHERE ltv.region = $Region AND ltv.image_id IS NOT NULL
    RETURN ltv.image_id AS image
}
WITH COLLECT(DISTINCT image) AS images
RETURN CASE WHEN SIZE(images) > 0 THEN images ELSE [] END AS unique_images;
```
- The 'is not null' at each step ensures that we're not adding 'None' to the list
- The `collect` makes sure that we de-dupe the ids
- The `case` statement makes it so that we always return a list

achantavy (approved):

cartography/intel/aws/ec2/images.py
heryxpc: I tested on a preproduction environment and it works, returning the same number of unique ids. The cost on ~ 1000 unique images of
```
WITH COLLECT(DISTINCT image) AS images
RETURN CASE WHEN SIZE(images) > 0 THEN images ELSE [] END AS unique_images;
```
is around 500 milliseconds.
I'm not sure about the performance improvement vs the `set` comprenhesion, but I can add the filter suggested to keep it all within the DB.

PR #1470, opened 2025-02-25
Use `pip install` with `--no-cache-dir` inside the Dockerfile

PeterDaveHello:
### Summary

This will help minimize the Docker image size by avoiding pip's package cache. Which is kind of Dockerfile best practices ;)

### Checklist


achantavy (approved):
Looks good to me, thank you!

PR #1471, opened 2025-02-26
ECR: add describe_images to get image size information.

heryxpc:
### Summary
I have a use case where I need the image size to take decisions on a future workflow.
I'm not so sure on the performance overhead, but wanted to give it a shot and see what other people think about getting the image details using [describe_images](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecr/client/describe_images.html) call.
`describe_image` was previously removed along with `describe_image_scan_findings` to improve performance(https://github.com/cartography-cncf/cartography/pull/468), but this PR only adds `describe_image`.

Screenshot 2025-02-27 at 9 39 01 a m

### Related issues or links
N/A

### Checklist

### DIFF
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -27,6 +27,7 @@ def get_ecr_repositories(boto3_session: boto3.session.Session, region: str) -> L
         ecr_repositories.extend(page['repositories'])
     return ecr_repositories
 
+
 @timeit
 @aws_handle_regions
 def get_ecr_repository_images(boto3_session: boto3.session.Session, region: str, repository_name: str) -> List[Dict]:
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -27,7 +27,8 @@ def get_ecr_repositories(boto3_session: boto3.session.Session, region: str) -> L
         ecr_repositories.extend(page['repositories'])
     return ecr_repositories
 
-
+@timeit
+@aws_handle_regions
 def get_ecr_repository_images(boto3_session: boto3.session.Session, region: str, repository_name: str) -> List[Dict]:
     logger.debug("Getting ECR images in repository '%s' for region '%s'.", repository_name, region)
     client = boto3_session.client('ecr', region_name=region)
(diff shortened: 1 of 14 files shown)


cartography/intel/aws/ecr.py
```diff
@@ -28,15 +28,16 @@ def get_ecr_repositories(boto3_session: boto3.session.Session, region: str) -> L
     return ecr_repositories
 
 
-@timeit
-@aws_handle_regions
```
achantavy, line 32: Probably still want the `@aws_handle_regions` decorator here for the error handling I think

cartography/intel/aws/ecr.py
```diff
@@ -103,7 +104,8 @@ def _load_ecr_repo_img_tx(
         ON CREATE SET ri.firstseen = timestamp()
         SET ri.lastupdated = $aws_update_tag,
             ri.tag = repo_img.imageTag,
-            ri.uri = repo_img.repo_uri + COALESCE(":" + repo_img.imageTag, '')
+            ri.uri = repo_img.repo_uri + COALESCE(":" + repo_img.imageTag, ''),
+            ri.image_size = repo_img.imageSizeInBytes
```
achantavy, line 108: [non-block] While you're here, are there other fields we'd like to include from https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_DescribeImages.html?

Maybe media manifest type, image pushed at, and last recorded pull time?

achantavy (changes requested):

cartography/intel/aws/ecr.py
heryxpc: Sure, I can add those extra fields.

heryxpc: Done

docs/root/modules/aws/schema.md
```diff
@@ -1379,6 +1379,11 @@ This way, more than one `ECRRepositoryImage` can reference/be connected to the s
 | tag | The tag applied to the repository image, e.g. "latest" |
 | uri | The URI where the repository image is stored |
 | **id** | same as uri |
+| image_size | The size of the image in bytes |
```
achantavy, line 1382:
```suggestion
| image_size_bytes | The size of the image in bytes |
```

cartography/intel/aws/ecr.py
```diff
@@ -103,7 +105,12 @@ def _load_ecr_repo_img_tx(
         ON CREATE SET ri.firstseen = timestamp()
         SET ri.lastupdated = $aws_update_tag,
             ri.tag = repo_img.imageTag,
-            ri.uri = repo_img.repo_uri + COALESCE(":" + repo_img.imageTag, '')
+            ri.uri = repo_img.repo_uri + COALESCE(":" + repo_img.imageTag, ''),
+            ri.image_size = repo_img.imageSizeInBytes,
```
achantavy, line 109:
```suggestion
            ri.image_size_bytes = repo_img.imageSizeInBytes,
```

achantavy (approved):

PR #1480, opened 2025-03-06
Paginate and extract image tag when using describe_images at ECR module

heryxpc:
### Summary
Follow up from https://github.com/cartography-cncf/cartography/pull/1471/.
[describe_images](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecr/client/describe_images.html) returns `imageTags` as an array of tags, compared to the single `imageTag`returned by [list_images](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecr/client/list_images.html).
Added as well more data to include a couple of the new fields to the integration test.

### Related issues or links
- https://github.com/cartography-cncf/cartography/pull/1471

### Checklist

### DIFF
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -41,10 +41,10 @@ def get_ecr_repository_images(boto3_session: boto3.session.Session, region: str,
         describe_response = describe_paginator.paginate(repositoryName=repository_name, imageIds=image_ids)
         for response in describe_response:
             image_details = response['imageDetails']
-            if len(image_details) != len(image_ids):
-                raise ValueError("Expected the same number of image details as image IDs")
-            for detail in image_details:
-                detail['imageTag'] = detail['imageTags'][0]
+            image_details = [
+                {**detail, 'imageTag': detail['imageTags'][0]} if detail.get('imageTags') else detail
+                for detail in image_details
+            ]
             ecr_repository_images.extend(image_details)
     return ecr_repository_images
 
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -33,13 +33,19 @@ def get_ecr_repositories(boto3_session: boto3.session.Session, region: str) -> L
 def get_ecr_repository_images(boto3_session: boto3.session.Session, region: str, repository_name: str) -> List[Dict]:
     logger.debug("Getting ECR images in repository '%s' for region '%s'.", repository_name, region)
     client = boto3_session.client('ecr', region_name=region)
-    paginator = client.get_paginator('list_images')
+    list_paginator = client.get_paginator('list_images')
     ecr_repository_images: List[Dict] = []
-    for page in paginator.paginate(repositoryName=repository_name):
+    for page in list_paginator.paginate(repositoryName=repository_name):
         image_ids = page['imageIds']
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/ecr.py
```diff
@@ -33,13 +33,19 @@ def get_ecr_repositories(boto3_session: boto3.session.Session, region: str) -> L
...
-            response = client.describe_images(repositoryName=repository_name, imageIds=image_ids)
-            ecr_repository_images.extend(response['imageDetails'])
+        describe_paginator = client.get_paginator('describe_images')
+        describe_response = describe_paginator.paginate(repositoryName=repository_name, imageIds=image_ids)
+        for response in describe_response:
+            image_details = response['imageDetails']
+            if len(image_details) != len(image_ids):
+                raise ValueError("Expected the same number of image details as image IDs")
+            for detail in image_details:
+                detail['imageTag'] = detail['imageTags'][0]
```
cmm-lyft, line 47: let's adding a validation here, in case imageTags is not present in detals, or is empty.

heryxpc: I removed the exception handling entirely and added the check for existence of imageTags inside a list comprehension to make it more readable.

cartography/intel/aws/ecr.py
```diff
@@ -33,13 +33,21 @@ def get_ecr_repositories(boto3_session: boto3.session.Session, region: str) -> L
...
-    for page in paginator.paginate(repositoryName=repository_name):
+    for page in list_paginator.paginate(repositoryName=repository_name):
         image_ids = page['imageIds']
-        if image_ids:
-            response = client.describe_images(repositoryName=repository_name, imageIds=image_ids)
-            ecr_repository_images.extend(response['imageDetails'])
+        if not image_ids:
+            continue
+        describe_paginator = client.get_paginator('describe_images')
+        describe_response = describe_paginator.paginate(repositoryName=repository_name, imageIds=image_ids)
```
heryxpc, line 43: I sid some tests and this call only adds a few millis to each page from `list_paginator` and, since the response to `describe_images` is also paginated, the number of calls to AWS API is very reduced.

cartography/intel/aws/ecr.py
```diff
@@ -33,13 +33,21 @@ def get_ecr_repositories(boto3_session: boto3.session.Session, region: str) -> L
...
+        for response in describe_response:
+            image_details = response['imageDetails']
+            image_details = [
+                {**detail, 'imageTag': detail['imageTags'][0]} if detail.get('imageTags') else detail
+                for detail in image_details
+            ]
```
heryxpc, line 49: @cmm-lyft how does this check looks like?

tests/data/aws/ecr.py
```diff
@@ -26,47 +26,88 @@
         },
     ],
 }
-
+DESCRIBE_IMAGES = {
```
heryxpc, line 29: Adding this to have sample data from [boto3 ec3.describe_images](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecr/client/describe_images.html).

cmm-lyft (approved):
LGTM! 🚀

achantavy (approved):

PR #1502, opened 2025-03-31
feat: Support one-to-many in data model, implement iaminstanceprofiles

### LINKED ISSUE #1365: [Issue report]Misuse of iaminstanceprofile in aws_ec2_iaminstance.json due to missing EC2InstanceProfile object
*Title*: *Misuse of iaminstanceprofile in aws_ec2_iaminstance.json due to missing EC2InstanceProfile object*

*Description*:
> In aws_ec2_iaminstance.json (found [here](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/analysis/aws_ec2_iaminstance.json)), Cartography incorrectly assumes a direct connection between the iaminstanceprofile name and the IAM Role. This assumption is not accurate, as AWS uses IamInstanceProfile as a distinct resource that links to a single IAM Role.

The misuse happens because Cartography does not currently create a separate EC2InstanceProfile object, leading to incorrect modeling of the relationship between EC2 Instances and IAM Roles.

Expected behavior:

A separate EC2InstanceProfile object should be created and properly linked to EC2 Instances.
The relationship between the EC2InstanceProfile and its associated IAM Role should be accurately represented, avoiding the incorrect assumption about the name correlation.

[optional *Relevant Links*:]
> [here](https://github.com/lyft/cartography/blob/master/cartography/data/jobs/analysis/aws_ec2_iaminstance.json)

achantavy:
Addresses discussion in #1465 and in this [slack thread](https://cloud-native.slack.com/archives/C080M2LRLDA/p1743350053575729) by adding support for one-to-many relationships in the Cartography data model.

Puts it to use to fix #1365 by adding correct support for iaminstanceprofiles. We now properly represent that AWS roles and EC2 instances have a many-to-many association.

- https://github.com/lyft/cartography/issues/1365

### Checklist

image

This PR looks bigger than it actually is because of test data and tests.


cartography/models/aws/iam/instanceprofile.py
```diff
@@ -0,0 +1,67 @@
...
+@dataclass(frozen=True)
+class InstanceProfileToAWSRoleRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef('lastupdated', set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class InstanceProfileToAWSRole(CartographyRelSchema):
+    target_node_label: str = 'AWSRole'
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {'arn': PropertyRef('Roles', one_to_many=True)},
```
achantavy, line 53: This is the main change of this pr

..

kind of a lie but kind of not at the same time :)

tests/unit/cartography/graph/test_querybuilder_onetomany.py
```diff
@@ -0,0 +1,48 @@
...
+            MERGE (i)<-[r:RESOURCE]-(j)
+            ON CREATE SET r.firstseen = timestamp()
+            SET
+                r.lastupdated = $lastupdated
+
+            UNION
+            WITH i, item
+            OPTIONAL MATCH (n0:AWSRole)
+            WHERE
+                n0.arn IN item.Roles
```
achantavy, line 35: this is how we generate a query that performs one-to-many writes

cartography/data/jobs/scoped_analysis/aws_ec2_iaminstanceprofile.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "EC2 Instances assume IAM roles",
```
achantavy, line 2: previously we assumed that ec2instance-->iamrole if the names matched. this is very incorrect, so we now implement iaminstanceprofiles correctly, and we also keep the same interface of still letting the user query for ec2 instances that can assume roles

cartography/graph/querybuilder.py
```diff
@@ -128,6 +133,9 @@ def _build_where_clause_for_rel_match(node_var: str, matcher: TargetNodeMatcher)
             prop_line = case_insensitive_match.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
         elif prop_ref.fuzzy_and_ignore_case:
             prop_line = fuzzy_and_ignorecase_match.safe_substitute(node_var=node_var, key=key, prop_ref=prop_ref)
+        elif prop_ref.one_to_many:
```
jychp, line 136: I do not know if there are some cases when we want (or even if it's possible) one_to_many with others flags but your implementation make theses flags incompatibles.

IMHO we should add a test before with a warning/error if there is incompatible flags or at least document it

cartography/models/core/common.py
```diff
@@ -50,6 +51,7 @@ def __init__(
         self.extra_index = extra_index
         self.ignore_case = ignore_case
         self.fuzzy_and_ignore_case = fuzzy_and_ignore_case
+        self.one_to_many = one_to_many
```
jychp, line 54: Please update doc string with this new param (and specify that it is incompatible with ignore_case, fuzzy_and_ignore_case)

achantavy: Done

cartography/graph/querybuilder.py
achantavy: Added a condition

PR #1512, opened 2025-04-02
Adding logs on 502 error for debugging in Github requests calls

cmm-lyft:
### Summary
These changes are necessary to simplify and clarify the debugging and triaging of pipeline errors caused by 502 HTTP errors when making requests to the GitHub API.

### Checklist

### DIFF
--- a/cartography/intel/github/util.py
+++ b/cartography/intel/github/util.py
@@ -70,6 +70,14 @@ def call_github_api(query: str, variables: str, token: str, api_url: str) -> Dic
         # Add context and re-raise for callers to handle
         logger.warning("GitHub: requests.get('%s') timed out.", api_url)
         raise
+
+    if response.status_code == 502:
+        logger.error(
+            "GitHub API returned 502 Bad Gateway error. This often indicates GitHub's internal issues. "
+            "Response headers: %s, Response body: %s",
+            dict(response.headers),
+            response.text,
+        )
     response.raise_for_status()
     response_json = response.json()
     if "errors" in response_json:


cmm-lyft:
👀  @heryxpc @achantavy 
This is a minor change, and shouldn't be impactful. WDYT?

cartography/intel/github/util.py
```diff
@@ -70,6 +70,14 @@ def call_github_api(query: str, variables: str, token: str, api_url: str) -> Dic
         # Add context and re-raise for callers to handle
         logger.warning("GitHub: requests.get('%s') timed out.", api_url)
         raise
+
+    if response.status_code == 502:
+        logger.error(
```
heryxpc, line 75:
```suggestion
        logger.warning(
```
Github API errors are logged as warning above/below this code.

cartography/intel/github/util.py
```diff
@@ -70,6 +70,14 @@ def call_github_api(query: str, variables: str, token: str, api_url: str) -> Dic
...
+            "GitHub API returned 502 Bad Gateway error. This often indicates GitHub's internal issues. "
+            "Response headers: %s, Response body: %s",
+            dict(response.headers),
+            response.text,
+        )
```
heryxpc, line 80: The `response.raise_for_status()`will log downstream the exception, but I agree that logging the body and headers will give more details on the actual error to troubleshoot.

heryxpc (approved):
Left a few comments, but LGTM.

cartography/intel/github/util.py
achantavy: @cmm-lyft can you provide an example log message created by this PR? I'm somewhat concerned with how large the message will be when logging a whole set of response headers.

cartography/intel/github/util.py
achantavy: +1. error-level logs in cartography are used to indicate fatals.

achantavy (changes requested):
Hey @cmm-lyft, can you give a bit more context on what you're trying to fix? I can help figure out what are the right things to have as logs and metrics.

achantavy:
Hey @cmm-lyft , I think the changes could make sense but I'd just like a bit more context, feel free to start a slack thread if that's easier

PR #1518, opened 2025-04-08
feat: add serial number and remove duplicate assignment

chandanchowdhury:
## Summary by mrge
Added serial number property to CrowdstrikeHost nodes and fixed a duplicate property assignment. This enables relationship mapping between Crowdstrike devices and other services like Kandji.

**Bug Fixes**
- Removed duplicate assignment of the `cid` property that was being set twice.

**New Features**
- Added `serial_number` property to CrowdstrikeHost nodes from device data.

### Summary
The Crowdstrike module currently does not fetch device serial which is essential for creating relationship with other services such as Kandji.

### Checklist

Before
There were no `serial_number` property to show.

After
Property `serial_number` shows up
image


achantavy (approved):
Thank you!

achantavy:
We can have a follow-up PR to make a relationship between crowdstrike and kandji

PR #1520, opened 2025-04-09
[Fix #1519] Error handling for launch templates sync

serge-wq:
## Summary by mrge
Added error handling for AWS launch templates that no longer exist during synchronization. This prevents process interruption when short-lived launch templates disappear between discovery and version fetching.

**Bug Fixes**
- Added try/except block to catch and handle missing launch template errors
- Added specific error handling for `InvalidLaunchTemplateId.NotFound` errors
- Added warning log when templates no longer exist instead of failing
- Preserved normal exception handling for other types of errors

This PR is part of a [stack](https://docs.mrge.io/overview), managed by [mrge](https://mrge.io):

* `master` (default branch)
* [#1143: Use arg to select aws session type (#1142)](https://github.com/cartography-cncf/cartography/pull/1143)
* **[#1520: [Fix #1519] Error handling for launch templates sync](https://github.com/cartography-cncf/cartography/pull/1520)** ⬅️ Current PR ([View on mrge](https://mrge.io/pr/cartography-cncf/cartography/pull/1520))

---

### Summary

Added error handling to the function that fetches the launch template versions so that the process doesn't get interrumpted when we can't find the launch template anymore. 

Some setups will have very short lived launch templates that will cause this part to fail if the launch template previously fetched doesn't exist anymore. More details in the bug issue. 

### Checklist

Provide proof that this works (this makes reviews move faster). Please perform one or more of the following:
- [ ] Update/add unit or integration tests.
- [ ] Include a screenshot showing what the graph looked like before and after your changes.
- [ ] Include console log trace showing what happened before and after your changes.

If you are changing a node or relationship:
- [ ] Update the [schema](https://github.com/lyft/cartography/tree/master/docs/root/modules) and [readme](https://github.com/lyft/cartography/blob/master/docs/schema/README.md).

If you are implementing a new intel module:
- [ ] Use the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).

PR #1522, opened 2025-04-10
Prevent loading of non-existent launch templates

serge-wq:
### Summary
Some launch templates can be short lived and be removed between `get_launch_templates` and `get_launch_template_versions` which caused the latter to fail with a `InvalidLaunchTemplateId.NotFound`. The previous fix only handled the error but we were still loading non-existent templates into the graph. This changes filters out the templates as an attempt to load cleaner data.

How logs look like during a run:
Screenshot 2025-04-10 at 4 38 33 p m

How graph looks like:
Screenshot 2025-04-10 at 4 46 34 p m

### Related issues or links
https://github.com/cartography-cncf/cartography/pull/1520

### Checklist


achantavy (approved):
Thanks!

ishaanverma:
i know its kinda late to have a look at this change since it got merged. but having the `get_launch_template_versions` function return a tuple instead of a list will result in a `ValueError` whenever an exception is caught by the `aws_hangle_regions` decorator since that decorator returns an empty list that can't be unpacked

ishaanverma:
in a situation like this where the template doesn't exist between the two calls, perhaps its fine to let the template version fetch fail and still load the template into the graph and whenever the job is run again that stale template will get deleted

serge-wq:
thanks for flagging! You are right totally missed the return type for the decorator there. I like the idea of consolidating at the transform function.

serge-wq:
Want to roll back this one in the meantime @ishaanverma?

achantavy:
If it's easier I'm open to just failing forward and fixing it since this has all been in pre-releases and not a real release

ishaanverma:
yup, happy to let it stay and patch

PR #1523, opened 2025-04-10
Refactor CrowdStrike host to new data model

achantavy:
Fixes a problem where a crowdstrike host's serial number only gets set when the node is created. 

I figured I'd just refactor the host to use the data model while I was looking at it.

### Checklist


cartography/models/crowdstrike/hosts.py
```diff
@@ -0,0 +1,50 @@
+# cartography/models/crowdstrike/host_schema.py
```
kunaals, line 1: I'd prefer to delete this comment given that it's just the file path?
```suggestion
```

kunaals (approved):
1 nit, otherwise LGTM

Issue #1519, opened 2025-04-09
Launch template not found when fetching versions


serge-wq:
**Bug Template**

*Title*: *Launch template not found when fetching versions*

*Description*:
AWS sync job is failing because of the inability to fetch versions for specific launch templates. 

We manage launch templates automatically, which makes some of them very short lived. I think that some of the templates fetched here: https://github.com/cartography-cncf/cartography/blob/e7dd92bfa0fc3542dfe6761510f4e210938953b9/cartography/intel/aws/ec2/launch_templates.py#L20 are no longer present when the execution reaches this part https://github.com/cartography-cncf/cartography/blob/e7dd92bfa0fc3542dfe6761510f4e210938953b9/cartography/intel/aws/ec2/launch_templates.py#L57

*To Reproduce*:
Hard to reproduce as it is very dependent on launch templates state and management.  

*Logs*:
```
Giving up inner_function(...) after 2 tries (botocore.exceptions.ClientError: An error occurred (InvalidLaunchTemplateId.NotFound) when calling the DescribeLaunchTemplateVersions operation: The specified launch template, with template ID <template-id>, does not exist.)
```

*Screenshots*:
> N/A

*Please complete the following information:*:
- Cartography release version or commit hash: 0.100.0
- Python version: 3.10

PR #1528, opened 2025-04-11
Consolidate launch templates and their versions outside the fetch functions

serge-wq:
### Summary
According to the comments in #1522, we will use a different approach to select which templates will be loaded in the end:
- Rolled back last changes in #1522 
- Used a simpler approach by just checking `LaunchTemplateId` to see which templates don't have any versions and filter them out. 

### Related issues or links
Related issue: #1519

### Testing

As before I've run locally the sync without a problem and also I'm adding some screenshots of my local neo4j instance showing:
- A sample of some launch templates linked to their versions.
- No launch template is added without a version

Screenshot 2025-04-11 at 5 41 58 p m
Screenshot 2025-04-11 at 5 39 10 p m
Screenshot 2025-04-11 at 5 28 48 p m

### Checklist


serge-wq:
What do you think? @ishaanverma

tests/unit/cartography/intel/aws/ec2/test_launch_templates.py
```diff
@@ -97,127 +103,167 @@ def test_get_launch_template_versions_by_template_success():
...
 
 
 @patch('cartography.intel.aws.ec2.launch_templates.logger')
-@patch('cartography.intel.aws.ec2.launch_templates.get_launch_template_versions_by_template')
-def test_get_launch_template_versions_none_found(mock_get_by_template, mock_logger):
+@patch('cartography.intel.aws.ec2.launch_templates.cleanup')
+@patch('cartography.intel.aws.ec2.launch_templates.load_launch_template_versions')
+@patch('cartography.intel.aws.ec2.launch_templates.transform_launch_template_versions')
+@patch('cartography.intel.aws.ec2.launch_templates.load_launch_templates')
+@patch('cartography.intel.aws.ec2.launch_templates.transform_launch_templates')
```
achantavy, line 203: Rather than patch out the transforms, is it enough to patch the `get` functions? I feel like that would make it simpler. If not, then I get it

cartography/intel/aws/ec2/launch_templates.py
```diff
@@ -176,11 +164,18 @@ def sync_ec2_launch_templates(
     for region in regions:
         logger.info(f"Syncing launch templates for region '{region}' in account '{current_aws_account_id}'.")
         templates = get_launch_templates(boto3_session, region)
-        versions, found_templates = get_launch_template_versions(boto3_session, region, templates)
+        versions = get_launch_template_versions(boto3_session, region, templates)
+
+        # Ensure we only load templates for which we found versions
+        # (handles templates deleted between the two API calls)
+        valid_template_ids = {v['LaunchTemplateId'] for v in versions}
+        existing_templates = [t for t in templates if t['LaunchTemplateId'] in valid_template_ids]
```
achantavy, line 172: Is it possible for this be moved into the `transform_launch_templates()` function? Ideally the sync function only does a get+transform+load+cleanup; but completely understand if that ends up breaking some interface or making something else much more complicated

ishaanverma:
lgtm! agree with @achantavy we can probably move that logic into the `transform_launch_templates` function

achantavy (approved):

PR #1535, opened 2025-04-17
Add integration tests for AWS VPC and IGW

achantavy:
Adds integ tests for AWS VPC and IGW to unblock #1530.

Paying back some debt.

### Checklist

### DIFF
--- a/tests/data/aws/ec2/internet_gateways.py
+++ b/tests/data/aws/ec2/internet_gateways.py
@@ -1,0 +1,13 @@
+TEST_INTERNET_GATEWAYS = [
+    {
+        'Attachments': [{'State': 'available', 'VpcId': 'vpc-038cf'}],
+        'InternetGatewayId': 'igw-013cb',
+        'OwnerId': '12345',
+        'Tags': [],
+    },
+    {
+        'Attachments': [{'State': 'available', 'VpcId': 'vpc-0f510'}],
+        'InternetGatewayId': 'igw-0387',
+        'OwnerId': '12345',
+    },
+]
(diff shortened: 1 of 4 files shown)


tests/integration/cartography/intel/aws/ec2/test_internet_gateways.py
```diff
@@ -0,0 +1,87 @@
...
+        neo4j_session,
+        boto3_session,
+        [TEST_REGION],
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+        {'UPDATE_TAG': TEST_UPDATE_TAG, 'AWS_ID': TEST_ACCOUNT_ID},
+    )
+
+    # Act
+    sync_internet_gateways(
```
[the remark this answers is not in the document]
achantavy: not necessary. we verify data correctness. testing for mocks is brittle as hell

tests/data/aws/ec2/internet_gateways.py
```diff
@@ -0,0 +1,13 @@
+TEST_INTERNET_GATEWAYS = [
+    {
+        'Attachments': [{'State': 'available', 'VpcId': 'vpc-038cf'}],
+        'InternetGatewayId': 'igw-013cb',
+        'OwnerId': '12345',
+        'Tags': [],
+    },
```
kunaals, line 7: Is it worth adding some tests where the IGWs are in states besides available?

jychp (approved):
LGTM but not tested

tests/data/aws/ec2/internet_gateways.py
achantavy: Discussed in person - we will ingest the igw regardless of states and we will store the state on the node

PR #1536, opened 2025-04-18
Use new data model for Kubernetes intel module

ishaanverma:
change the kubernetes intel module to use the cartography data model and introduce a new schema for kubernetes. this is a potentially breaking change, since this changes the relationships between the nodes and a few properties of the nodes.

the old schema looked like this:
current

the new schema looks like this (the pink relations show sub-resource relationships):
proposed

the goal of the new schema is to better mirror the k8s object model and allow for easy extensibility to add more resource types in the future.  

Screenshot after running E2E on a minikube cluster:
Screenshot 2025-06-26 at 11 23 56 PM

### Checklist


cartography/intel/kubernetes/secrets.py
```diff
@@ -1,60 +1,113 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 @timeit
-def sync_secrets(
-    session: Session,
-    client: K8sClient,
+def get_secrets(client: K8sClient) -> List[V1Secret]:
+    items = k8s_paginate(client.core.list_secret_for_all_namespaces, limit=1)
```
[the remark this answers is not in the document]
ishaanverma: good bot

cartography/intel/kubernetes/util.py
```diff
@@ -1,51 +1,164 @@
...
     if date:
-        return int(date.strftime("%s"))
+        return int(date.timestamp())
     return None
+
+
+def k8s_paginate(
+    list_func: Callable,
+    **kwargs: Any,
+) -> list[dict[str, Any]]:
```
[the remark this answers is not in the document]
ishaanverma: in the kubernetes python client there is no top-level base model class that all kubernetes resource models like V1Pod, V1Namespace, etc. inherit from so it's hard to define a type here. but we can be sure that returned value will contain a list of dict[str, Any]

cartography/intel/kubernetes/__init__.py
```diff
@@ -3,39 +3,55 @@
...
 from cartography.intel.kubernetes.util import get_k8s_clients
-from cartography.util import run_cleanup_job
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
 
 
 @timeit
 def start_k8s_ingestion(session: Session, config: Config) -> None:
+    if not config.update_tag:
```
jychp, line 19: Not sure this will ever trigger as update_tag as a default value

cartography/models/kubernetes/clusters.py
```diff
@@ -0,0 +1,27 @@
...
+    compiler: PropertyRef = PropertyRef("compiler")
+    platform: PropertyRef = PropertyRef("platform")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesClusterSchema(CartographyNodeSchema):
+    label: str = "KubernetesCluster"
+    properties: KubernetesClusterNodeProperties = KubernetesClusterNodeProperties()
+    sub_resource_relationship = None
```
jychp: This statement is useless.

By the way if you want to use an automatic cleanup job you can add the new flag:
`    scoped_cleanup: bool = False`

cartography/models/kubernetes/namespaces.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@dataclass(frozen=True)
+class KubernetesNamespaceNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("uid")
+    name: PropertyRef = PropertyRef("name", extra_index=True)
+    creation_timestamp: PropertyRef = PropertyRef("creation_timestamp")
+    deletion_timestamp: PropertyRef = PropertyRef("deletion_timestamp")
+    status_phase: PropertyRef = PropertyRef("status_phase")
+    cluster_name: PropertyRef = PropertyRef(
```
jychp, line 20: Is the extra_index realy needed here ?
Node is already linked to a Cluster Node

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class KubernetesContainerNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("uid")
+    name: PropertyRef = PropertyRef("name", extra_index=True)
+    image: PropertyRef = PropertyRef("image", extra_index=True)
+    namespace: PropertyRef = PropertyRef("namespace", extra_index=True)
+    cluster_name: PropertyRef = PropertyRef(
```
jychp, line 20: Is the extra_index really necessary as the node is already connected to a cluster node ?

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+        "CLUSTER_NAME", set_in_kwargs=True, extra_index=True
+    )
+    pod_name: PropertyRef = PropertyRef("pod_name", extra_index=True)
```
jychp: Same question here because the container is already connected to a pod node

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+class KubernetesContainerToKubernetesNamespace(CartographyRelSchema):
+    target_node_label: str = "KubernetesNamespace"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "cluster_name": PropertyRef("CLUSTER_NAME", set_in_kwargs=True),
+            "name": PropertyRef("namespace"),
+        }
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
```
jychp: `RESOURCE` label should be reserved for 'sub_resource' relationships.

I think `CONTAINS` can be a good alternative here.

cartography/models/kubernetes/pods.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@dataclass(frozen=True)
+class KubernetesPodNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("uid")
+    name: PropertyRef = PropertyRef("name", extra_index=True)
+    status_phase: PropertyRef = PropertyRef("status_phase")
+    creation_timestamp: PropertyRef = PropertyRef("creation_timestamp")
+    deletion_timestamp: PropertyRef = PropertyRef("deletion_timestamp")
+    namespace: PropertyRef = PropertyRef("namespace", extra_index=True)
```
jychp, line 21: Is the extra_index really needed as the node is already connected to a namespace node ?

cartography/models/kubernetes/pods.py
```diff
@@ -0,0 +1,80 @@
...
+    labels: PropertyRef = PropertyRef("labels")
+    cluster_name: PropertyRef = PropertyRef(
```
jychp, line 23: same question here

cartography/models/kubernetes/pods.py
```diff
@@ -0,0 +1,80 @@
...
+class KubernetesPodToKubernetesNamespace(CartographyRelSchema):
+    target_node_label: str = "KubernetesNamespace"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "cluster_name": PropertyRef("CLUSTER_NAME", set_in_kwargs=True),
+            "name": PropertyRef("namespace"),
+        }
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
```
jychp: `RESOURCE` label should be reserved for 'sub_resource' relationships.

I think `CONTAINS` can be a good alternative here.

cartography/models/kubernetes/secrets.py
```diff
@@ -0,0 +1,79 @@
...
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class KubernetesSecretNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("uid")
+    name: PropertyRef = PropertyRef("name", extra_index=True)
+    creation_timestamp: PropertyRef = PropertyRef("creation_timestamp")
+    deletion_timestamp: PropertyRef = PropertyRef("deletion_timestamp")
+    namespace: PropertyRef = PropertyRef("namespace", extra_index=True)
```
jychp, line 20: Do we need this extra index ? The node is already connected to a namespace node.

cartography/models/kubernetes/secrets.py
```diff
@@ -0,0 +1,79 @@
...
+    owner_references: PropertyRef = PropertyRef("owner_references")
+    type: PropertyRef = PropertyRef("type")
+    cluster_name: PropertyRef = PropertyRef(
```
jychp, line 23: Same question here

cartography/models/kubernetes/secrets.py
```diff
@@ -0,0 +1,79 @@
...
+class KubernetesSecretToKubernetesNamespace(CartographyRelSchema):
+    target_node_label: str = "KubernetesNamespace"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "cluster_name": PropertyRef("CLUSTER_NAME", set_in_kwargs=True),
+            "name": PropertyRef("namespace"),
+        }
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
```
jychp: `RESOURCE` label should be reserved for 'sub_resource' relationships.

I think `CONTAINS` can be a good alternative here.

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+    status_image_id: PropertyRef = PropertyRef("status_image_id")
+    status_image_sha: PropertyRef = PropertyRef("status_image_sha")
+    status_ready: PropertyRef = PropertyRef("status_ready")
+    status_started: PropertyRef = PropertyRef("status_started")
+    status_state: PropertyRef = PropertyRef("status_state")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesContainerToKubernetesNamespaceProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesContainerToKubernetesNamespaceRelProperties` (naming convention)

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesContainerToKubernetesPodProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesContainerToKubernetesPodRelProperties` (naming convention)

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesContainer)<-[:RESOURCE]-(:KubernetesNamespace)
+class KubernetesContainerToKubernetesNamespace(CartographyRelSchema):
```
jychp: Name should be `KubernetesContainerToKubernetesNamespaceRel` (naming convention)

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: KubernetesContainerToKubernetesNamespaceProperties = (
+        KubernetesContainerToKubernetesNamespaceProperties()
+    )
+
+
+@dataclass(frozen=True)
+# (:KubernetesContainer)<-[:CONTAINS]-(:KubernetesPod)
+class KubernetesContainerToKubernetesPod(CartographyRelSchema):
```
jychp: Name should be `KubernetesContainerToKubernetesPodRel` (naming convention)

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "CONTAINS"
+    properties: KubernetesContainerToKubernetesPodProperties = (
+        KubernetesContainerToKubernetesPodProperties()
+    )
+
+
+@dataclass(frozen=True)
+class KubernetesContainerToKubernetesClusterProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesContainerToKubernetesClusterRelProperties` (naming convention)

cartography/models/kubernetes/containers.py
```diff
@@ -0,0 +1,109 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesContainer)<-[:RESOURCE]-(:KubernetesCluster)
+class KubernetesContainerToKubernetesCluster(CartographyRelSchema):
```
jychp: Name should be `KubernetesContainerToKubernetesClusterRel` (naming convention)

cartography/models/kubernetes/namespaces.py
```diff
@@ -0,0 +1,51 @@
...
+        "cluster_name", set_in_kwargs=True, extra_index=True
+    )
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesNamespaceToKubernetesClusterProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesNamespaceToKubernetesClusterRelProperties` (naming convention)

cartography/models/kubernetes/namespaces.py
```diff
@@ -0,0 +1,51 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesNamespace)<-[:RESOURCE]-(:KubernetesCluster)
+class KubernetesNamespaceToKubernetesCluster(CartographyRelSchema):
```
jychp: Name should be `KubernetesNamespaceToKubernetesClusterRel` (naming convention)

cartography/models/kubernetes/pods.py
```diff
@@ -0,0 +1,80 @@
...
+    labels: PropertyRef = PropertyRef("labels")
+    cluster_name: PropertyRef = PropertyRef(
+        "CLUSTER_NAME", set_in_kwargs=True, extra_index=True
+    )
+    node: PropertyRef = PropertyRef("node")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesPodToKubernetesNamespaceProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesPodToKubernetesNamespaceRelProperties` (naming convention)

cartography/models/kubernetes/pods.py
```diff
@@ -0,0 +1,80 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesPod)<-[:RESOURCE]-(:KubernetesNamespace)
+class KubernetesPodToKubernetesNamespace(CartographyRelSchema):
```
jychp: Name should be `KubernetesPodToKubernetesNamespaceRel` (naming convention)

cartography/models/kubernetes/pods.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: KubernetesPodToKubernetesNamespaceProperties = (
+        KubernetesPodToKubernetesNamespaceProperties()
+    )
+
+
+@dataclass(frozen=True)
+class KubernetesPodToKubernetesClusterProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesPodToKubernetesClusterRelProperties` (naming convention)

cartography/models/kubernetes/pods.py
```diff
@@ -0,0 +1,80 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesPod)<-[:RESOURCE]-(:KubernetesCluster)
+class KubernetesPodToKubernetesCluster(CartographyRelSchema):
```
jychp: Name should be `KubernetesPodToKubernetesClusterRel` (naming convention)

cartography/models/kubernetes/secrets.py
```diff
@@ -0,0 +1,79 @@
...
+    owner_references: PropertyRef = PropertyRef("owner_references")
+    type: PropertyRef = PropertyRef("type")
+    cluster_name: PropertyRef = PropertyRef(
+        "CLUSTER_NAME", set_in_kwargs=True, extra_index=True
+    )
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesSecretToKubernetesNamespaceProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesSecretToKubernetesNamespaceRelProperties` (naming convention)

cartography/models/kubernetes/secrets.py
```diff
@@ -0,0 +1,79 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesSecret)<-[:RESOURCE]-(:KubernetesNamespace)
+class KubernetesSecretToKubernetesNamespace(CartographyRelSchema):
```
jychp: Name should be `KubernetesSecretToKubernetesNamespaceRel` (naming convention)

cartography/models/kubernetes/secrets.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: KubernetesSecretToKubernetesNamespaceProperties = (
+        KubernetesSecretToKubernetesNamespaceProperties()
+    )
+
+
+@dataclass(frozen=True)
+class KubernetesSecretToKubernetesClusterProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesSecretToKubernetesClusterRelProperties` (naming convention)

cartography/models/kubernetes/secrets.py
```diff
@@ -0,0 +1,79 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesSecret)<-[:RESOURCE]-(:KubernetesCluster)
+class KubernetesSecretToKubernetesCluster(CartographyRelSchema):
```
jychp: Name should be `KubernetesSecretToKubernetesClusterRel` (naming convention)

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class KubernetesServiceNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("uid")
+    name: PropertyRef = PropertyRef("name", extra_index=True)
+    creation_timestamp: PropertyRef = PropertyRef("creation_timestamp")
+    deletion_timestamp: PropertyRef = PropertyRef("deletion_timestamp")
+    namespace: PropertyRef = PropertyRef("namespace", extra_index=True)
```
jychp, line 20: Do we really need that index ? The node is already connected to an other node representing that field value.

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+    selector: PropertyRef = PropertyRef("selector")
+    type: PropertyRef = PropertyRef("type")
+    load_balancer_ip: PropertyRef = PropertyRef("load_balancer_ip")
+    load_balancer_ingress: PropertyRef = PropertyRef("load_balancer_ingress")
+    cluster_name: PropertyRef = PropertyRef(
```
jychp, line 26: Do we really need that index ? The node is already connected to an other node representing that field value.

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+        "CLUSTER_NAME", set_in_kwargs=True, extra_index=True
+    )
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesServiceToKubernetesClusterProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesServiceToKubernetesClusterRelProperties` (naming convention)

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesService)<-[:RESOURCE]-(:KubernetesCluster)
+class KubernetesServiceToKubernetesCluster(CartographyRelSchema):
```
jychp: Name should be `KubernetesServiceToKubernetesClusterRel` (naming convention)

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+class KubernetesServiceToKubernetesNamespace(CartographyRelSchema):
+    target_node_label: str = "KubernetesNamespace"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "cluster_name": PropertyRef("CLUSTER_NAME", set_in_kwargs=True),
+            "name": PropertyRef("namespace"),
+        }
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
```
jychp: `RESOURCE` label should be reserved for 'sub_resource' relationships.

I think `CONTAINS` can be a good alternative here.

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+    properties: KubernetesServiceToKubernetesClusterProperties = (
+        KubernetesServiceToKubernetesClusterProperties()
+    )
+
+
+@dataclass(frozen=True)
+class KubernetesServiceToKubernetesNamespaceProperties(CartographyRelProperties):
```
jychp: Name should be `KubernetesServiceToKubernetesNamespaceRelProperties` (naming convention)

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesService)<-[:RESOURCE]-(:KubernetesNamespace)
+class KubernetesServiceToKubernetesNamespace(CartographyRelSchema):
```
jychp: Name should be `KubernetesServiceToKubernetesNamespaceRel` (naming convention)

cartography/models/kubernetes/services.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+@dataclass(frozen=True)
+class KubernetesServiceToKubernetesPodProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesService)-[:TARGET]->(:KubernetesPod)
+class KubernetesServiceToKubernetesPod(CartographyRelSchema):
```
jychp: Name should be `KubernetesServiceToKubernetesPodRel` (naming convention)

docs/root/modules/kubernetes/schema.md
jychp, line 1: Refer to the comments on `RESOURCE` vs `CONTAINS` on model description files.

tests/integration/cartography/intel/kubernetes/test_clusters.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Assert
+    expected_nodes = {
+        (KUBERNETES_CLUSTER_IDS[0],),
+        (KUBERNETES_CLUSTER_IDS[1],),
+    }
+    assert check_nodes(neo4j_session, "KubernetesCluster", ["id"]) == expected_nodes
+
+
+# cleaning up the kubernetes cluster node is currently not supported
```
jychp, line 29: Use `scoped_cleanup: bool = False` on the NodeSchema definition and you will be able to cleanup

cartography/intel/kubernetes/clusters.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    load(
+        neo4j_session,
+        KubernetesClusterSchema(),
+        cluster_data,
+        lastupdated=update_tag,
+    )
+
+
+# cleaning up the kubernetes cluster node is currently not supported
```
jychp, line 64: Use `scoped_cleanup: bool = False` in the NodeSchema definition and you will be able to cleanup

jychp (changes requested):
Some comments on naming convention but look good to me overall (not tested)

cartography/models/kubernetes/clusters.py
ishaanverma: removed, and added scoped cleanup

cartography/models/kubernetes/containers.py
ishaanverma: done

cartography/models/kubernetes/pods.py
ishaanverma: done

cartography/models/kubernetes/secrets.py
ishaanverma: done

cartography/models/kubernetes/containers.py
ishaanverma: done

cartography/models/kubernetes/containers.py
ishaanverma: done

cartography/models/kubernetes/containers.py
ishaanverma: done

cartography/models/kubernetes/namespaces.py
ishaanverma: done

cartography/models/kubernetes/pods.py
ishaanverma: done

cartography/models/kubernetes/pods.py
ishaanverma: done

cartography/models/kubernetes/secrets.py
ishaanverma: done

cartography/models/kubernetes/secrets.py
ishaanverma: done

cartography/models/kubernetes/services.py
ishaanverma: done

cartography/models/kubernetes/services.py
ishaanverma: done

cartography/models/kubernetes/services.py
ishaanverma: done

cartography/models/kubernetes/services.py
ishaanverma: done

docs/root/modules/kubernetes/schema.md
ishaanverma: sticking with `RESOURCE` for sub resource relationships and `CONTAINS` for the other relationships

cartography/models/kubernetes/containers.py
ishaanverma: done

cartography/models/kubernetes/containers.py
ishaanverma: done

cartography/models/kubernetes/containers.py
ishaanverma: done

cartography/models/kubernetes/namespaces.py
ishaanverma: done

cartography/models/kubernetes/pods.py
ishaanverma: done

cartography/models/kubernetes/pods.py
ishaanverma: done

cartography/models/kubernetes/secrets.py
ishaanverma: done

cartography/models/kubernetes/secrets.py
ishaanverma: done

cartography/models/kubernetes/services.py
ishaanverma: done

cartography/models/kubernetes/services.py
ishaanverma: done

ishaanverma:
re: `extra_index` properties on the nodes. even though the node is connected to another node that has the same value, the extra index just helps with queries like:

```cypher
(n:KubernetesPod{cluster_name: 'xyz'}) RETURN n
(n:KubernetesPod{namespace: 'abc'}) RETURN n
```

and i think the extra index should also help with the ingestion for cases like:

```python
@dataclass(frozen=True)
# (:KubernetesContainer)<-[:CONTAINS]-(:KubernetesPod)
class KubernetesContainerToKubernetesPodRel(CartographyRelSchema):
    target_node_label: str = "KubernetesPod"
        {
            "cluster_name": PropertyRef("CLUSTER_NAME", set_in_kwargs=True),
            "namespace": PropertyRef("namespace"),
            "id": PropertyRef("pod_id"),
        }
    )
    direction: LinkDirection = LinkDirection.INWARD
    rel_label: str = "CONTAINS"
    properties: KubernetesContainerToKubernetesPodRelProperties = (
        KubernetesContainerToKubernetesPodRelProperties()
    )
```

where the target node matcher relies on matching the namespace property of a node

ishaanverma:
the cleanup jobs are still something i need to look into. i think in some cases they dont seem to work correctly... maybe im missing something

ishaanverma:
had a chance to look at the `scoped_cleanup` flag. in this case setting `scoped_cleanup` to `False` will not trigger a cleanup because of this condition in `cartography/jobs/cleanupbuilder.py` -> `build_cleanup_queries`:

```python
if (
    not node_schema.sub_resource_relationship
    and not node_schema.other_relationships
):
    return []
```

The `KubernetesClusterSchema` does not have any relationships defined so the automatic cleanup job will just return empty queries.

ishaanverma:
also wanted to point out another scenario, that took me a while to understand. 

let's say you have two clusters `A` and `B`. if you run the cartography intel job only on cluster `A` first, then run the intel job only on cluster `B`. The second time you run it, no nodes will be cleaned up in cluster `A` because the cleanup queries rely on a `CLUSTER_ID` parameter. for e.g. the cleanup queries generated by the cleanup job will be:

```cypher
MATCH (n:KubernetesNamespace)<-[s:RESOURCE]-(:KubernetesCluster{id: $CLUSTER_ID})
WHERE n.lastupdated <> $UPDATE_TAG
WITH n LIMIT $LIMIT_SIZE
DETACH DELETE n;

MATCH (n:KubernetesNamespace)<-[s:RESOURCE]-(:KubernetesCluster{id: $CLUSTER_ID})
WHERE s.lastupdated <> $UPDATE_TAG
WITH s LIMIT $LIMIT_SIZE
DELETE s  
```

so the cleanup is just scoped to the cluster currently being ingested. which makes sense, you could execute multiple cluster ingestion jobs that run in parallel without interfering with each other.

jychp:
Regarding the cleanup you're right, my bad, I didn’t check whether your node had any relationships.

As for the extra index, I usually go with a query like `(Pod)<-[:HAS]-(Namespace {id: x})`, but if you prefer the solution with the extra index, I’m totally fine with that.

jychp (approved):
LGTM

Thanks for the refactoring work this will be really helpful, even for other modules (like cloud providers with managed K8s clusters).

PR #1553, opened 2025-05-06
Human node rework

jychp:
This PR standardizes the way `Human` nodes are managed.

It is the first in a series aimed at introducing other abstract nodes following the same model (UserGroup, Device, ComputeInstance, Database, etc.).

Currently, the concept of a Human is handled inconsistently across integrations:
- GSuite maps Users to Humans via an analysis job (but does not create the node).
- Duo and LastPass define a relationship via the data model but do not create the node.
- Okta creates both the node and the relationship via an explicit (legacy) query.
- Some intel modules import users but do not map them to Humans (Entra, GitHub, SnipeIT, OCI, PagerDuty).

This PR standardizes the following:
- LastPass (migrated)
- Duo (migrated)
- Entra (added)
- GitHub (added)
- SnipeIT (added)

The following modules will be migrated once they are updated to the new data model:
- GSuite
- Okta
- OCI
- PagerDuty


cartography/models/common/helpers.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def _load_abstracted_nodes(
+    neo4j_session: neo4j.Session,
+    schema: Type[CartographyNodeSchema],
+    update_tag: int,
+    data: List[Dict[str, Any]],
+    mapping: Dict[str, str],
+    id_field: str = "id",
+):
```
achantavy, line 23: Add a comment about how this function standardizes the ways that an email address can be referred to

achantavy (commented):
I understand and agree with the summary you outlined:

> GSuite maps Users to Humans via an analysis job (but does not create the node).
Duo and LastPass define a relationship via the data model but do not create the node.
Okta creates both the node and the relationship via an explicit (legacy) query.
Some intel modules import users but do not map them to Humans (Entra, GitHub, SnipeIT, OCI, PagerDuty).

Basically the majority of these modules assume that a Human node already exists, so they `match` on it and then draw an edge to it from a `*user` node.

I also understand that `load_human_from_users()` `merge`s in Human nodes when given just an email address or some minimal bit of information. This is an interesting pattern.

Zooming out a bit, I'll share a bit of background on why it is this way. At Lyft, the `Human` node referred to our internal definition of a team member, so there was an internal module there that served as the "authority" for a `Human` node. So, all of these other modules that came later have these rough edges where they assume that a `Human` node already exists in the graph.

This is an uncommon usecase, but I am slightly concerned about how this PR interacts with deployments similar to Lyft's where someone has created a module that ingests `Human` nodes based on their own team-member service data. Put another way (to hopefully be more clear), if we have a module that ingests in Human nodes without using any tricks used in this PR, will there be unwanted side effects in data quality if the code in this PR runs? We could potentially test for that.

I'm also slightly concerned about adding another overarching pattern required for all Cartography syncs: this PR asks all module authors to remember to add in a call to `load_human_from_users()` for all cases where a user email is potentially present in the data. I'd call this a "bottom-up" approach.

However, on one hand, I think this is really neat because it makes cartography do some analysis to highlight the humans in a graph, and then we can traverse them and quickly see the human identities.

On the other hand, I wonder if instead of doing it this way, should we be instead advocating for a "top-down" approach where a team-member module would dictate "these are the team member emails, and I know that they have identities in Duo, Entra, SnipeIT, etc".

Also, it's possible for email addresses to match here for non-human identities and then we would be mislabelling them by giving them the `Human` label with this code.

Let me think on this a bit, let me know what you think too

jychp:
Oh, I see. I think this comes from the fact that "Human" is actually being used more as a notion of an employee. Which isn’t a problem, but it definitely shifts the approach.

If "Human" is indeed an employee, then yes it should only be created from a source of truth (like the IDP or an HR module).

What do you think about adding a CLI argument `--idp-module` that takes the name of one of the modules? If it's provided, Humans are created from that module; otherwise, the module just links Users to existing Humans.

That also opens up the possibility of detecting all external or undeclared accounts by querying User nodes that aren't linked to any Human. I love that.

As for the implementation approach, I think we need a hybrid model:
* Standard modules should declare relationships in the usual way.
* Only modules that can be considered IDPs (and are passed via `--idp-module`) should be allowed to create nodes (either using the provided helper or their own method).

PR #1556, opened 2025-05-09
fix: isort was missing from precommit config

jychp:
### Summary
precommit config file with `isort` was not commited in #1466


chandanchowdhury (approved):
Thank you :)

PR #1557, opened 2025-05-09
Switch to uv for package management

jychp:
### Summary

This PR introduces the use of **uv** for dependency management.

It takes the approach of committing the uv.lock file (which contains the resolved dependencies, is updated either by Dependabot or manually by contributors) to ensure consistent dependency resolution across:
-  contributors' machines
- test workflows
- the build CI

_Note: Using pinned dependencies is considered a best practice and is a prerequisite for the project's acceptance in the next stages of the CNCF process._

### DIFF
--- a/docs/root/dev/developer-guide.md
+++ b/docs/root/dev/developer-guide.md
@@ -19,17 +19,21 @@ If you prefer docker instead, skip this and scroll down to [these instructions](
 
 1. **Perform an editable install of the cartography source code**
 
-    Run `cd cartography` and then `pip install -e .` (yes, actually type the period into the command line) to install Cartography from source to the current venv.
+    Run `cd cartography` and then `uv sync` (yes, actually type the period into the command line) to install Cartography from source to the current venv.
 
 4. **Run from source**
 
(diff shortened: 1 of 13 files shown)


docs/root/dev/developer-guide.md
```diff
@@ -40,19 +45,19 @@ If you prefer docker instead, skip this and scroll down to [these instructions](
     `export "NEO4J_URL=<your_neo4j_instance_bolt_url:your_neo4j_instance_port>"`
 
 1. **Run tests using `make`**
-    - `make test_lint` runs [pre-commit](https://pre-commit.com) linting against the codebase.
-    - `make test_unit` runs the unit test suite.
+    - `uv run make test_lint` runs [pre-commit](https://pre-commit.com) linting against the codebase.
```
achantavy, line 48: Would it make sense to instead update the Makefile? If I understand correctly, if we updated the Makefile then we wouldn't need to prefix `make` with `uv run`

jychp: Sure it makes sense. I will make the change.

jychp:
docker build time devided by 3: 98s -> 31s
image size reduced by 25%: 1GB to 725Mo

achantavy (approved):

PR #1563, opened 2025-05-13
chore: add tests to enforce schema convention

jychp:
### Summary

This PR adds unit tests (`<path to file>`) to enforce data schema conventions. The tests validate the following:

* Python object naming conventions:

  * Node classes end with `Schema`
  * Relationship classes end with `Rel`
  * Node properties classes end with `Properties`
  * Relationship properties classes end with `RelProperties`
* All `sub_resource_relationship`s must have the label `RESOURCE` and direction `INWARD`
* Each module must have **at least one** root node (i.e., a node with no `sub_resource_relationship`)
* Each module must have **at most one** root node

> ⚠️ During the data model migration phase, these tests will only emit warnings and are non-blocking.

### Notes

* While the naming conventions might seem like nitpicking, they greatly improve overall code consistency, readability, and make AI-assisted code completion much more relevant and effective.

* I also renamed all affected objects in this PR to align with the conventions. The only executable code added is the test itself.

✅ These changes are non-breaking and do **not** impact the underlying database schema or the end users.

### DIFF
--- a/cartography/models/aws/apigateway.py
+++ b/cartography/models/aws/apigateway.py
@@ -24,21 +24,21 @@ class APIGatewayRestAPINodeProperties(CartographyNodeProperties):
 
 
 @dataclass(frozen=True)
-class APIGatewayRestAPIToAwsAccountRelProperties(CartographyRelProperties):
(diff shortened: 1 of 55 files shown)


achantavy (approved):
Big help, thank you!

PR #1564, opened 2025-05-13
add seed for demo DB

jychp:
### Summary
This PR introduces a new `demo` module that seeds a Neo4j database with mock data.

At this stage, the dataset is based on our integration test fixtures. Each seed can be customized or extended with different data, so there's some code duplication for now — but it's intentional to allow more flexibility in shaping demo datasets going forward.

I opted to seed the data using raw code instead of exporting from an existing Neo4j instance. While this approach requires a bit more upfront setup, it's much easier to maintain over time (no need to re-export the DB every time the dataset changes). This script can later be integrated into CI to ensure the demo database stays aligned with the latest version of the tool.

Can be simply run with `uv run python3 -m demo`. Credentials are configured with `NEO4J_URL` var env.

> ⚠️ I edited some data from the test to bring some consistency between modules (and allow Human nodes to be build accross modules for instance)

### Not covered by this PR
- AWS: will be done in a dedicated PR due to the volume of data
- Following modules will be added after they have been migrated to the new data model migration:
     - GCP
     - Gsuite
     - Jamf
     - Kubernetes
     - OCI
     - Okta
     - PagerDuty
- Data adaptation for a more connected graph will be added/modified (this will be done step by step alongside the Ontology implementation)

image


demo/seeds/bigfix.py
```diff
@@ -0,0 +1,34 @@
+from unittest.mock import patch
+
+import cartography.intel.bigfix.computers
+import tests.data.bigfix.computers
+from demo.seeds.base import Seed
+
+BIGFIX_ROOT_URL = "https://bigfixroot.example.com"
+BIGFIX_USER = "myuser"
+BIGFIX_PASSWORD = "mypassword"
```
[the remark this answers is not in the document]
jychp: fake credentials

demo/seeds/bigfix.py
```diff
@@ -0,0 +1,34 @@
+from unittest.mock import patch
+
+import cartography.intel.bigfix.computers
+import tests.data.bigfix.computers
```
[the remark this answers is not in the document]
jychp: demo will not be deployed on production, acceptable tradeoff for now

demo/seeds/openai.py
```diff
@@ -0,0 +1,116 @@
+from unittest.mock import Mock
+from unittest.mock import patch
+
+import cartography.intel.openai.adminapikeys
+import cartography.intel.openai.apikeys
+import cartography.intel.openai.projects
+import cartography.intel.openai.serviceaccounts
+import cartography.intel.openai.users
+import tests.data.openai.adminapikeys
```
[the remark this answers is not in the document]
jychp: demo will not be deployed on production, acceptable tradeoff for now

tests/integration/cartography/intel/semgrep/test_findings.py
```diff
@@ -97,7 +97,7 @@ def test_sync_findings(mock_get_sca_vulns, mock_get_deployment, neo4j_session):
         "moment|2.29.2",
         "moment|2.29.4",
         "package-lock.json",
-        "https: //github.com/org/repository/blob/commit_id/package-lock.json#L14373",
+        "https: //github.com/simpsoncorp/sample_repo/blob/commit_id/package-lock.json#L14373",
```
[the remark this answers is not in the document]
kunaals: @jychp this seems legit?

demo/__main__.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    count = result.single()["count"]
+    if count > 0:
+        if force_flag:
+            logger.warning(
+                "Force flag is set. Proceeding to clear the database without confirmation."
+            )
+        else:
+            print(
+                "Database already contains none demo data., Are you sure you want to continue? (y/N)"
```
kunaals:
```suggestion
                "Database already contains demo data. Are you sure you want to continue? (y/N)"
```

demo/__main__.py
```diff
@@ -0,0 +1,177 @@
...
+    humans = [
+        {"email": "mbsimpson@simpson.corp"},
+        {"email": "hjsimpson@simpson.corp"},
+        {"email": "lmsimpson@simpson.corp"},
+        {"email": "bjsimpson@simpson.corp"},
+    ]
+    neo4j_session.run(
+        "UNWIND $data as item MERGE (h:Human{email: item.email})",
+        data=humans,
+    )
```
kunaals: @jychp thoughts on bringing abstract node types into this demo at the same time that we bring them into Cartography itself? I feel like bringing the Human node type into this demo now doesn't do it justice and may confuse users at this time

demo/seeds/duo.py
```diff
@@ -0,0 +1,111 @@
...
+        self._seed_groups(mock_client)
+        self._seed_users(mock_client)
+
+    def _seed_hosts(self) -> None:
+        cartography.intel.duo.api_host.sync_duo_api_host(
+            self.neo4j_session,
+            {
+                "DUO_API_HOSTNAME": API_HOSTNAME,
+                "UPDATE_TAG": self.update_tag,
+            },
```
kunaals: nit: this stanza repeated multiple times in file

demo/seeds/semgrep.py
```diff
@@ -0,0 +1,116 @@
...
+        ON CREATE SET dep.firstseen = timestamp()
+        SET dep.lastupdated = $update_tag
+        """,
+        dep_id="moment|2.29.2",
+        update_tag=1234,
+    )
+
+
+# TODO: Replace with actual CVE data
+def create_cve_nodes(neo4j_session) -> None:
```
kunaals, line 47: Can we use the CVE data / module here instead of creating the CVE node by hand? Maybe even use the CVE seed?

kunaals (commented):
Looks great, left some minor notes that Id like to discuss and after that I think its good to merge!

demo/__main__.py
```diff
@@ -0,0 +1,177 @@
...
+    # Set up Neo4j connection
+    neo4j_driver = neo4j.GraphDatabase.driver(NEO4J_URL)
+    neo4j_session = neo4j_driver.session()
+
+    UPDATE_TAG = 0
+
+    # Config
+    config = Config(
+        neo4j_uri=NEO4J_URL,
+    )
```
kunaals:
```suggestion
NEO4J_URL = os.environ.get("NEO4J_URL", "bolt://localhost:7687")
NEO4J_USER = os.environ.get("NEO4J_USER")
NEO4J_PASSWORD = os.environ.get("NEO4J_PASSWORD")

logger = logging.getLogger(__name__)


def main(force_flag: bool) -> None:
    # Set up Neo4j connection
    if NEO4J_USER and NEO4J_PASSWORD:
        neo4j_driver = neo4j.GraphDatabase.driver(
            NEO4J_URL,
            auth=neo4j.basic_auth(NEO4J_USER, NEO4J_PASSWORD),
        )
    else:
        neo4j_driver = neo4j.GraphDatabase.driver(NEO4J_URL)
    neo4j_session = neo4j_driver.session()

    UPDATE_TAG = 0

    # Config
    config = Config(
        neo4j_uri=NEO4J_URL,
        neo4j_user=NEO4J_USER,
        neo4j_password=NEO4J_PASSWORD,
    )
```
Making the demo more usable might be nice especially if it's someone's first day using Cartography. Picking up username and password if folks have it in their env would help, they may have changed username and password on startup or might be using aura to run the demo.

kunaals: (somewhat selfish since I was using a non-default neo4j password 😄 )

demo/__main__.py
```diff
@@ -0,0 +1,177 @@
...
+                return
+
+    # Clear the previous database
+    logger.info("Clearing the existing database...")
+    neo4j_session.run("MATCH (n) DETACH DELETE n;")
+
+    # Create indexes
+    logger.info("Creating indexes...")
+    create_indexes.run(neo4j_session, config)
+
```
kunaals:
```suggestion

    # Wait for indexes to finish building to avoid "insanely frequent schema changes"
    # error when we immediately start writing data.
    logger.info("Waiting for indexes to become online...")
    try:
        neo4j_session.run("CALL db.awaitIndexes()")
    except neo4j.exceptions.Neo4jError:
        # Some Neo4j versions require a timeout argument.
        neo4j_session.run("CALL db.awaitIndexes('600s')")

```
When running this for the first time I ran into an error where Cartography started writing before indexes were created, this change fixed that 

```
neo4j.exceptions.DatabaseError: {code: Neo.DatabaseError.Statement.ExecutionFailed} {message: Could not compile query due to insanely frequent schema changes}
```

demo/__main__.py
```diff
@@ -0,0 +1,177 @@
...
+    # TODO: Okta after data model migration
+    logger.info("    loading OpenAI")
+    OpenAISeed(neo4j_session, UPDATE_TAG).run()
+    # TODO: PagerDuty after data model migration
+    logger.info("    loading Semgrep")
+    SemgrepSeed(neo4j_session, UPDATE_TAG).run()
+    logger.info("    loading SnipeIT")
+    SnipeitSeed(neo4j_session, UPDATE_TAG).run()
+    logger.info("    loading Tailscale")
+    TailscaleSeed(neo4j_session, UPDATE_TAG).run()
```
kunaals:
```suggestion
    import importlib
    import pkgutil
    import inspect

    from demo import seeds
    seed_classes: list[type] = []

    for _, mod_name, _ in pkgutil.iter_modules(seeds.__path__):
        if mod_name == "base":
            continue
        module = importlib.import_module(f"demo.seeds.{mod_name}")
        for _, obj in inspect.getmembers(module, inspect.isclass):
            # A valid seed is a concrete subclass declared inside demo.seeds.*
            if (
                issubclass(obj, (Seed, AsyncSeed))
                and obj not in (Seed, AsyncSeed)
                and obj.__module__.startswith("demo.seeds")
            ):
                seed_classes.append(obj)

    # Preserve deterministic order for reproducibility
    seed_classes.sort(key=lambda cls: cls.__name__)

    logger.info("Loading demo data via %d seed classes...", len(seed_classes))
    for seed_cls in seed_classes:
        logger.info("    loading %s", seed_cls.__name__)
        try:
            seed_cls(neo4j_session, UPDATE_TAG).run()
        except Exception:
            logger.exception("Seed %s failed", seed_cls.__name__)
```
This might be a little overcomplicated, but it makes it so that you just have to add a seed file and don't have to register it again here. This way we can add custom seed files like you mention in the PR description, but we keep things DRY here. Tested it and it works

kunaals: Oh and we'll have to add an import `from demo.seeds.base import Seed, AsyncSeed`

demo/__main__.py
achantavy: +1, let's hold off on adding Human nodes until then

demo/__main__.py
jychp: I'll update the message to "Database already contains data."
The main goal of this check is to reduce the risk of accidentally pushing data to the production DB if someone forgets to set the correct environment variable.

demo/__main__.py
jychp: yeah, the point of this PR is to offer me an easy env to work on abstract node/ontology

but sure we could remove this old feature for now

demo/seeds/semgrep.py
jychp: Yes, that’s what I had in mind, but it’ll take a bit more time, so I was planning to address it in a separate PR if that works for you. That’s why I left the TODO comment :)

demo/__main__.py
jychp: I believe this is caused by some modules still using the old schema, otherwise, this should already be handled by the create_indexes helper.
I’ve added it to the demo anyway.

demo/__main__.py
jychp: Oh I love that, much more elegant, and very similar to what I suggested for dynamic module intel loading.

demo/seeds/semgrep.py
kunaals: sg!

kunaals (approved):
Looks good!

Issue #1477, opened 2025-03-04
Missing Assets/Resource


Mubashir-ahmed:
Hello team, I am currently mapping alerts generated on my AWS account to their resource. However I am unable to find any suitable asset for following.

1- AwsEc2VpcEndpointService
2- AwsSsmDocument
3- AwsSnsTopic
4- AwsS3AccountPublicAccessBlock
5- AwsWafv2WebAcl
6- AwsWafv2RuleGroup
7- AwsWafv2WebAcl
8- AwsCertificateManagerCertificate
9- AwsAthenaWorkGroup
10- AwsBackupBackupPlan
11- AwsBackupBackupVault
12- AwsCloudFormationStack
13- AwsCloudFrontDistribution
14- AwsCloudTrailTrail
15- AwsCloudWatchAlarm
16- AwsLogsLogGroup
17- AwsEventsEventbus
18- AwsEventSchemasRegistry
19- AwsGlueDataCatalog
20- AwsGuardDutyDetector
21- AwsRdsEventSubscription

Please look into this. If they are present in cartography, please let me know their name. If not present please add them with arn so that they can be mapped easily (can provide sample arn if need be). Thanks

PS: I am using prowler for running cloud security check

achantavy:
Great list, thanks for the feature request. All the resources we support are here: https://cartography-cncf.github.io/cartography/modules/aws/index.html

Happy to accept PRs, you can follow these guides: https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html

But as time goes on we will aim to add all of these.

Which are most important for you right now?

arujjval:
Hi, can I start working on this?

D3VISRI:
Hello Sir, I am new to open source. I thought I would give this issue a try, but the checklist (like the CloudWatch issue) is still open. However, when I went through the codebase, it seems to be already implemented. Am I right?

yogeshvema:
hlo sir can i start working on this ??

PR #1570, opened 2025-05-17
feat: Add S3AccountPublicAccessBlock support to AWS module

AdityaPandeyCN:
### Summary
I have added S3AccountPublicAccessBlock Support to the AWS module. I am working on adding the remaining intel modules and prioritised this specific one because I feel this is important from the security point of view.

### Related issues or links

- https://github.com/cartography-cncf/cartography/issues/1477

### Checklist


AdityaPandeyCN:
I got a type error after using the aws_handle_regions decorator with my function so removed it,thinking of a different approach now.

cartography/intel/aws/s3accountpublicaccessblock.py
```diff
@@ -0,0 +1,153 @@
...
+def get_account_public_access_block(
+    boto3_session: boto3.session.Session, region: str
+) -> List[Dict]:
+    """
+    Get the S3 Account Public Access Block settings for a region.
+    Returns a list containing at most one item (the block configuration).
+    """
+    client = boto3_session.client("s3control", region_name=region)
+    try:
+        account_id = boto3_session.client("sts").get_caller_identity()["Account"]
```
achantavy, line 34: Are we sure the account id is a required parameter for `get_public_access_block()`? 

If so, we shouldn't need to use get-caller-identity to get the current account. From the `sync_*()` function, the `common_job_parameters` variable has a key called `AWS_ID` that stores the current AWS account ID being synced. We can pass that value into this `get_()` function.

cartography/intel/aws/s3accountpublicaccessblock.py
```diff
@@ -0,0 +1,150 @@
...
+from cartography.stats import get_stats_client
+from cartography.util import merge_module_sync_metadata
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+stat_handler = get_stats_client(__name__)
+
+
+@timeit
+def get_account_public_access_block(
```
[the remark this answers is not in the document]
achantavy: Whats the type error that you get? Ideally we don't need to reimplement the exception handling that the decorator provides

cartography/intel/aws/s3accountpublicaccessblock.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    for region in regions:
+        logger.info(
+            f"Syncing S3 Account Public Access Block for {region} in account {current_aws_account_id}"
+        )
+
+        public_access_blocks = get_account_public_access_block(boto3_session, region)
+        transformed_data = []
+
+        for public_access_block in public_access_blocks:
```
achantavy, line 126: Move this loop into `transform_account_public_access_block()` function so that the `sync_()` function doesn't need to do this looping

cartography/intel/aws/s3accountpublicaccessblock.py
```diff
@@ -0,0 +1,153 @@
...
+            transformed = transform_account_public_access_block(
+                public_access_block,
+                region,
+                current_aws_account_id,
+            )
+            transformed_data.append(transformed)
+
+        if transformed_data:
```
achantavy, line 134: The `if` is not necessary here. If the list is empty, then the call to `load()` under the hood will not cause any database transactions.

achantavy (changes requested):
Big thanks for the help! Left small suggestions.

Also remember to update the README.md's AWS supported items list

cartography/intel/aws/s3accountpublicaccessblock.py
AdityaPandeyCN: > Are we sure the account id is a required parameter for `get_public_access_block()`?
> I referenced this https://docs.aws.amazon.com/cli/latest/reference/s3control/get-public-access-block.html#:~:text=To%20list%20public%20block%20access%20settings%20for%20an%20account 
> If so, we shouldn't need to use get-caller-identity to get the current account. From the `sync_*()` function, the `common_job_parameters` variable has a key called `AWS_ID` that stores the current AWS account ID being synced. We can pass that value into this `get_()` function.
This feels correct,will make the changes

AdityaPandeyCN: I referenced this for adding accound id a parameter https://docs.aws.amazon.com/cli/latest/reference/s3control/get-public-access-block.html#:~:text=To%20list%20public%20block%20access%20settings%20for%20an%20account

achantavy (approved):

achantavy:
Thank you!!!

Issue #1552, opened 2025-05-05
Support additional AWS resource types from CloudGoat scenarios


kunaals:
**Description**:
> Cartography is missing a number of AWS resource types that are required to fully ingest and visualize CloudGoat scenarios.  Adding these will enable users to run through CloudGoat tutorials end‑to‑end and get immediate insight into attacker infrastructure.
>
> The resources we currently lack support for include (but are not limited to):
> - `aws_ssm_parameter`
> - `aws_codebuild_project`
> - `aws_cloudtrail_trail`
> - `aws_cloudwatch_log_group`
> - `aws_cloudwatch_log_metric_filter`
> - `aws_cloudwatch_metric_alarm`
> - `aws_sns_topic`
> - `aws_sns_topic_subscription`
> - `aws_s3_object`
> - `aws_ecs_cluster`
> - `aws_ecs_task_definition`
> - `aws_ecs_service`
> - `aws_efs_file_system`
> - `aws_efs_mount_target`
> - `aws_efs_access_point`
> - `aws_cloudwatch_event_rule`
> - `aws_cloudwatch_event_target`
> - `aws_lambda_permission`
> - `aws_route_table_association`
> - `aws_glue_connection`
> - `aws_glue_job`
> - `aws_s3_bucket_notification`
> - `aws_s3_bucket_public_access_block`
> - `aws_secretsmanager_secret_version`
> - `aws_api_gateway_method`
> - `aws_api_gateway_integration`
> - `aws_api_gateway_deployment`
> - `aws_cognito_user_pool`
> - `aws_cognito_user_pool_client`
> - `aws_cognito_identity_pool_roles_attachment`
> - `aws_lambda_event_source_mapping`
>
> Ideally each resource should:
> 1. Be added to the Cartography schema with properties needed to reconstruct common attacker/defender workflows.  
> 2. Have ingestion logic in the AWS plugin, including pagination, ARNs, tags, and relationships.  
> 3. Include examples in the docs and sample queries

[optional **Relevant Links**:]
> - CloudGoat “codebuild_secrets” scenario: https://github.com/RhinoSecurityLabs/cloudgoat/tree/master/scenarios/codebuild_secrets  
> - CloudGoat “detection_evasion” scenario: https://github.com/RhinoSecurityLabs/cloudgoat/blob/master/scenarios/detection_evasion/  
[...]

PR #1571, opened 2025-05-18
feat:  Add Secret manager secret version support 

AdityaPandeyCN:
### Summary
Added support for AWS Secrets Manager Secret Versions which allows Cartography to ingest and track version information for AWS Secrets Manager secrets.

### Related issues or links

### Checklist


cartography/intel/aws/secretsmanager.py
```diff
@@ -76,6 +76,107 @@ def cleanup_secrets(neo4j_session: neo4j.Session, common_job_parameters: Dict) -
...
+    region: str,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    """
+    Load secret versions into Neo4j.
+    """
+
+    ingest_secret_versions = """
+    UNWIND $SecretVersions as version
```
achantavy, line 115: For new modules, use the data model. It is ok to attach assets that were created using the data model to assets that were not.

cartography/models/aws/secretsmanager/secret_version.py
```diff
@@ -0,0 +1,115 @@
...
+    rel_label: str = "ENCRYPTED_BY"
+    properties: SecretsManagerSecretVersionRelProperties = (
+        SecretsManagerSecretVersionRelProperties()
+    )
+    # Only create this relationship if KmsKeyId exists
+    conditional_match_property: str = "KmsKeyId"
+
+
+@dataclass(frozen=True)
+class SecretsManagerSecretVersionSchema(CartographyNodeSchema):
```
achantavy, line 98: This doesn't appear to be used anywhere

cartography/intel/aws/secretsmanager.py
```diff
@@ -76,6 +81,115 @@ def cleanup_secrets(neo4j_session: neo4j.Session, common_job_parameters: Dict) -
...
+    region: str,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    """
+    Load secret versions into Neo4j using the data model.
+    """
+    logger.info("Loading %d secret versions", len(data))
+
+    formatted_data = []
```
achantavy, line 140: Move these adjustments to a transform() function

cartography/intel/aws/secretsmanager.py
```diff
@@ -76,6 +81,115 @@ def cleanup_secrets(neo4j_session: neo4j.Session, common_job_parameters: Dict) -
...
+
+@timeit
+def cleanup_secret_versions(
+    neo4j_session: neo4j.Session, common_job_parameters: Dict
+) -> None:
+    """
+    Clean up secret versions that are no longer present.
+    """
+    run_cleanup_job(
+        "aws_import_secret_versions_cleanup.json",
```
achantavy, line 187: Use the node schema, don't create new cleanup job files

cartography/intel/aws/secretsmanager.py
```diff
@@ -76,6 +81,115 @@ def cleanup_secrets(neo4j_session: neo4j.Session, common_job_parameters: Dict) -
...
+                if "CreatedDate" in version and not isinstance(
+                    version["CreatedDate"], datetime.datetime
+                ):
+                    version["CreatedDate"] = datetime.datetime.fromtimestamp(
+                        version["CreatedDate"], tz=datetime.timezone.utc
+                    )
+
+            versions.extend(page["Versions"])
+
+    except client.exceptions.ResourceNotFoundException:
```
achantavy, line 113: No need to handle these exceptions here because you're using the decorator -- no need to use any `try`+`except`.

tests/integration/cartography/intel/aws/test_secretsmanager.py
```diff
@@ -63,3 +66,105 @@ def test_load_load_secrets(neo4j_session, *args):
         for n in nodes
     }
     assert actual_nodes == expected_nodes
+
+
+def test_load_secret_versions(neo4j_session, *args):
```
achantavy, line 71: Don't test the `load()` by itself, test the `sync()`

tests/integration/cartography/intel/aws/test_secretsmanager.py
```diff
@@ -63,3 +66,105 @@ def test_load_load_secrets(neo4j_session, *args):
...
+            "arn:aws:secretsmanager:us-east-1:000000000000:secret:test-secret-1-000000:version:11111111-1111-1111-1111-111111111111",
+            "arn:aws:secretsmanager:us-east-1:000000000000:secret:test-secret-1-000000",
+            "11111111-1111-1111-1111-111111111111",
+            ("AWSPREVIOUS",),
+            "us-east-1",
+            1397585689,
+        ),
+    }
+
+    nodes = neo4j_session.run(
```
achantavy, line 134: Use `check_rels()` and `check_nodes()` instead. IMO `check_rels()` is more important.

achantavy (changes requested):
Hey thanks for following up, I'm wondering though, why is this so different from your earlier PR https://github.com/cartography-cncf/cartography/pull/1570/files? Your earlier PR had all of the patterns completely correct and this one is very different. Copy the patterns from your previous ones. If you're using AI, don't be shy about yelling at the AI until it gets it right :-)

AdityaPandeyCN:
Yeah sorry I think I messed up the structure while trying to understand the already implemented secret manager, I was importing the wrong load function the whole time which wasted some time. 

Also,I don't want to disturb you again and again so will tag you when I feel this is complete. Thankyou

d-aggarwal (commented):
Hi @AdityaPandeyCN , thank you for your work on this PR!

Just a small note regarding the `SecretsManagerSecretVersion` schema — according to the [[Boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/list_secret_version_ids.html)], the `ARN` refers to the secret itself, not to an individual version. This means that all versions under the same secret will share the same ARN.

You might consider adjusting the schema or clarifying the field to reflect this.

AdityaPandeyCN:
Can you have a look @achantavy

cartography/intel/aws/secretsmanager.py
```diff
@@ -85,12 +182,54 @@ def sync(
...
+
         load_secrets(neo4j_session, secrets, region, current_aws_account_id, update_tag)
+
+        for secret in secrets:
+            logger.info(
+                f"Getting versions for secret {secret.get('Name', 'unnamed')} ({secret['ARN']})"
+            )
+            secret_versions = get_secret_versions(boto3_session, region, secret["ARN"])
+
+            if secret_versions:
```
achantavy, line 206: Is it possible to go without this if-else logic? Ideally we don't need to add if-else nesting as much as possible in the sync functions.

cartography/models/aws/secretsmanager/secret_version.py
```diff
@@ -0,0 +1,116 @@
...
+
+@dataclass(frozen=True)
+class SecretsManagerSecretVersionToKMSKeyRel(CartographyRelSchema):
+    """
+    Relationship between Secret Version and its KMS key
+    Only created when KmsKeyId is present
+    """
+
+    target_node_label: str = "AWSKMSKey"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
```
achantavy, line 86: Can you please add a `check_rels()` test for this? 

You could do [this](https://github.com/cartography-cncf/cartography/blob/b886e0cf6b80bf69a5ed2901d1369f826f482560/tests/integration/cartography/intel/aws/test_kms.py#L10-L17) to  write in KMS Keys in the `arrange` step, run your code in the `act` step, and then finally test for the relationship in the `assert` step.

tests/integration/cartography/intel/aws/test_secretsmanager.py
```diff
@@ -1,65 +1,186 @@
...
+    sync(
+        neo4j_session,
+        boto3_session,
+        [TEST_REGION],
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+        {"UPDATE_TAG": TEST_UPDATE_TAG, "AWS_ID": TEST_ACCOUNT_ID},
+    )
+
+    secrets = neo4j_session.run(
```
achantavy, line 174: For consistency, could use `check_nodes()` and assert where it returns `set()` for empty set

achantavy (changes requested):
Almost there! thanks for addressing my other feedback. left comments on adding a test to the kms key

AdityaPandeyCN:
I have addressed the suggestions.

achantavy (approved):
LGTM thank you so much!

PR #1576, opened 2025-05-20
feat: Add SSMParameter resource to AWS module

Khushmagrawal:
This commit adds support for ingesting AWS SSM Parameters as a supported resource for the AWS Module.

### Related issues or links
(#1552)

### Checklist


tests/integration/cartography/intel/aws/test_ssm.py
```diff
@@ -165,3 +166,91 @@ def test_load_instance_patches(mock_get_instances, neo4j_session):
...
+            ),
+        ),
+    }
+
+    query = """
+    MATCH (p:SSMParameter)
+    RETURN p.id, p.name, p.type, p.lastmodifieddate, p.description, p.allowedpattern, p.policies_json
+    ORDER BY p.name // Optional: Order to make comparison easier if set iteration order isn't guaranteed
+    """
+    nodes = neo4j_session.run(query)
```
achantavy, line 223: Please use `check_rels()` and `check_nodes()` instead of neo4j_session.run()

cartography/models/aws/ssm/parameters.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@dataclass(frozen=True)
+class SSMParameterToKMSKeyRelRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+@dataclass(frozen=True)
+class SSMParameterToKMSKeyRel(CartographyRelSchema):
+    target_node_label: str = 'AWSKmsKey'
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
```
achantavy, line 60: This is awesome, can you please add an integration test to ensure this relationship gets created? This would involve adding in test data from KMSKeys, so in the test you would sync in test KMSKey data, then run the tests for SSMParameters , and then verify that the rel exists.

If this is too hard or out of scope for this PR, feel free to remove this relationship for now and have it come in a follow-up PR.

achantavy (changes requested):
Thanks so much for your PR! Left some comments about testing for the KMSKey relationship. If that ends up being too hard, let's defer that for a future PR.

cartography/models/aws/ssm/parameters.py
achantavy: Oh, specifically you could do [this](https://github.com/cartography-cncf/cartography/blob/b886e0cf6b80bf69a5ed2901d1369f826f482560/tests/integration/cartography/intel/aws/test_kms.py#L10-L17) to write in KMS Keys in the arrange step, run your code in the act step, and then finally test for the relationship in the assert step.

achantavy:
Side note, you can run the linter locally before pushing code to your PR: see https://cartography-cncf.github.io/cartography/dev/developer-guide.html#automated-testing

cartography/models/aws/ssm/parameters.py
Khushmagrawal: Added the integration test to verify the `ENCRYPTED_WITH` relationship.

jychp (changes requested):
thx for the PR, it's look very nice
some comment (somes a nitpicking more for information)

tests/integration/cartography/intel/aws/test_ssm.py
```diff
@@ -1,16 +1,22 @@
...
 import tests.data.aws.ssm
 from cartography.intel.aws.ec2.instances import sync_ec2_instances
 from tests.data.aws.ec2.instances import DESCRIBE_INSTANCES
 from tests.integration.cartography.intel.aws.common import create_test_account
+from tests.integration.util import check_nodes
+from tests.integration.util import check_rels
 
 TEST_ACCOUNT_ID = "000000000000"
 TEST_REGION = "us-east-1"
+TEST_REGION_FOR_KMS = "eu-west-1"
```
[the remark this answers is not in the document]
Khushmagrawal: Hey! Actually in the test data of KMS, region is `eu-west-1`  which was different from test region previously used in SSM and KMS keys are region-specific by design(although it does have multi region keys) I thought it made sense to keep the test region consistent for both KMS and SSM Parameter. That’s why I introduced this variable to align their regions.
Should I change the SSM data to `us-east-1`?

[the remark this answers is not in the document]
jychp: Ok there is a lack of consistency at the AWS level, will handle that later.

jychp (approved):
LGTM

jychp:
Kudos for your contribution !

PR #1584, opened 2025-05-23
feat: Add s3_object  resource to the aws module

AdityaPandeyCN:
### Summary
This PR adds  s3_objects support to the aws module. I will be making it  more complete like using all the model classes I have contructed, I also feel I can make the file naming and structure better so will be also working on that.

### Checklist

### DIFF
--- a/cartography/intel/aws/s3_objects.py
+++ b/cartography/intel/aws/s3_objects.py
@@ -58,7 +58,7 @@ def get_s3_objects_for_bucket(
     return objects
 
 
-@timeit
+
 def transform_s3_objects(
     objects: List[Dict],
     bucket_name: str,
--- a/cartography/intel/aws/s3_objects.py
+++ b/cartography/intel/aws/s3_objects.py
@@ -1,0 +1,250 @@
+import logging
+from datetime import datetime, timedelta, timezone
+from typing import Dict, List, Optional
+
+import boto3
+import neo4j
+
+from cartography.client.core.tx import load
+from cartography.graph.job import GraphJob
+from cartography.models.aws.s3.s3_object import S3ObjectSchema
+from cartography.stats import get_stats_client
+from cartography.util import aws_handle_regions
+from cartography.util import dict_date_to_epoch
+from cartography.util import merge_module_sync_metadata
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+stat_handler = get_stats_client(__name__)
+
+# Configuration constants
+DEFAULT_MAX_OBJECTS_PER_BUCKET = 10000
+
+
+@timeit
+@aws_handle_regions
+def get_s3_objects_for_bucket(
+    boto3_session: boto3.session.Session,
+    bucket_name: str,
+    region: str,
+    max_objects: int = DEFAULT_MAX_OBJECTS_PER_BUCKET,
+    fetch_owner: bool = False,
+) -> List[Dict]:
+    """
+    Get S3 objects for a specific bucket with pagination limit.
+    
+    Args:
+        boto3_session: AWS session
+        bucket_name: S3 bucket name
+        region: AWS region
+        max_objects: Maximum objects to return per bucket
+        fetch_owner: Whether to fetch owner information
+    """
+    client = boto3_session.client('s3', region_name=region)
+    objects = []
+    
+    paginator = client.get_paginator('list_objects_v2')
+    page_iterator = paginator.paginate(
+        Bucket=bucket_name,
+        FetchOwner=fetch_owner,
+        PaginationConfig={'MaxItems': max_objects}
+    )
+    
+    for page in page_iterator:
+        if 'Contents' in page:
+            objects.extend(page['Contents'])
+            
+    logger.info(f"Found {len(objects)} objects in bucket {bucket_name}")
+    return objects
+
+
+@timeit
+def transform_s3_objects(
+    objects: List[Dict],
+    bucket_name: str,
+    bucket_arn: str,
+    region: str,
+    aws_account_id: str,
+) -> List[Dict]:
+    """
+    Transform S3 objects to match our data model.
+    Based on actual ListObjectsV2 response fields.
+    """
+    transformed_objects = []
+    
+    for obj in objects:
+        # Skip folder markers (0-byte objects ending with /)
+        if obj.get('Size', 0) == 0 and obj.get('Key', '').endswith('/'):
+            continue
+            
+        transformed = {
+            'Key': obj['Key'],
+            'ARN': f"{bucket_arn}/{obj['Key']}",
+            'BucketName': bucket_name,
+            'BucketARN': bucket_arn,
+            'Size': obj.get('Size', 0),
+            'StorageClass': obj.get('StorageClass', 'STANDARD'),
+            'LastModified': dict_date_to_epoch(obj, 'LastModified'),
+            'ETag': obj.get('ETag', '').strip('"'),
+            'Region': region,
+        }
+        
+        # Add owner fields if present (only when FetchOwner=true)
+        if 'Owner' in obj:
+            transformed['OwnerId'] = obj['Owner'].get('ID')
+            transformed['OwnerDisplayName'] = obj['Owner'].get('DisplayName')
+            
+        # Add checksum algorithm if present
+        if 'ChecksumAlgorithm' in obj:
+            transformed['ChecksumAlgorithm'] = obj['ChecksumAlgorithm']
+            
+        # Add restore status for archived objects (Glacier)
+        if 'RestoreStatus' in obj:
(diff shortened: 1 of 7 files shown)


achantavy:
Ooh this one is super interesting!

@kunaals @jychp @heryxpc

jychp:
@achantavy interesting but can lead to very very long runs :)

@AdityaPandeyCN you did not choose the easiest one.

Adding the `DEFAULT_MAX_OBJECTS_PER_BUCKET = 10000` is a good security, but I think we want to be able to configure that using the CLI.

You will have to work on following files:
- cli.py
- config.py

I think we want:
- a default value
- `0` will disable object sync

Disabling object sync could be very important in a restricted environment (if their is any risk that object name contains PII/ePHI etc ...)

(I have a doubt with `0`, setting a limit to 0 is usually for unlimited, but want to try to avoid adding a `--enable-object-sync`)

(@achantavy feel free to comment/object those requirements)

achantavy:
But it's crazy useful

jychp:
Yeah, that will allow so many new analyses use cases

AdityaPandeyCN:
Thankyou for the comments, I am thinking of adding a s3-object-limit flag to control S3 object syncing. It'll default to 10000 objects per bucket, but users can set their own limit or disable it completely (with 0) for sensitive data environments. I'll start with the CLI and config changes. What do you think?

AdityaPandeyCN:
I feel this is ready for review.

cartography/intel/aws/s3_objects.py
```diff
@@ -0,0 +1,266 @@
...
+            neo4j_session,
+            transformed_objects,
+            bucket_region,
+            current_aws_account_id,
+            update_tag,
+        )
+
+        total_objects_synced += len(transformed_objects)
+
+    logger.info(f"Total S3 objects synced: {total_objects_synced}")
```
jychp, line 240: Not sure having a query to put metrics in logs is worth it.

cartography/models/aws/s3/s3_object.py
```diff
@@ -0,0 +1,80 @@
...
+    id: PropertyRef = PropertyRef("ARN")
+    arn: PropertyRef = PropertyRef("ARN", extra_index=True)
+    key: PropertyRef = PropertyRef("Key", extra_index=True)
+    bucket_name: PropertyRef = PropertyRef("BucketName", extra_index=True)
+    size: PropertyRef = PropertyRef("Size")
+    storage_class: PropertyRef = PropertyRef("StorageClass")
+    last_modified: PropertyRef = PropertyRef("LastModified", extra_index=True)
+    etag: PropertyRef = PropertyRef("ETag")
+    # Owner fields (only present if FetchOwner=true)
+    owner_id: PropertyRef = PropertyRef("OwnerId")
```
jychp, line 29: Can we make a relationship from that ID ?

cartography/models/aws/s3/s3_object.py
```diff
@@ -0,0 +1,80 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class S3ObjectToS3BucketRel(CartographyRelSchema):
+    target_node_label: str = "S3Bucket"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("BucketARN")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
```
jychp, line 55: We prefer `verbs`, an `INWARD` relationship whith `STORES` or `HAS` label is more coherent with the rest of the datamodel

tests/integration/cartography/intel/aws/test_s3object.py
```diff
@@ -0,0 +1,290 @@
...
+    )
+
+    expected_arns = {
+        ("arn:aws:s3:::test-bucket/documents/report.pdf",),
+        ("arn:aws:s3:::test-bucket/images/logo.png",),
+        ("arn:aws:s3:::test-bucket/archive/old-data.zip",),
+    }
+    assert check_nodes(neo4j_session, "S3Object", ["arn"]) == expected_arns
+
+    expected_rels = {
```
jychp, line 62: please test the Bucket-Object relationship too

tests/integration/cartography/intel/aws/test_s3object.py
```diff
@@ -0,0 +1,290 @@
...
+def test_sync_s3_objects_glacier_restore(mock_get_objects, neo4j_session):
+    """
+    Test syncing Glacier objects with restore status.
+    """
+    boto3_session = MagicMock()
+    create_test_account(neo4j_session, TEST_ACCOUNT_ID, TEST_UPDATE_TAG)
+
+    neo4j_session.run("MATCH (n:S3Object) DETACH DELETE n")
+
+    neo4j_session.run(
```
jychp, line 139: define an `_ensure_neo4j_has_tests_buckets` at the top of the file to avoid code duplication

tests/integration/cartography/intel/aws/test_s3object.py
```diff
@@ -0,0 +1,290 @@
...
+    assert result["owner_id"] == "owner-id-123"
+    assert result["owner_display_name"] == "test-owner"
+
+
+@patch.object(
+    cartography.intel.aws.s3_objects,
+    "get_s3_objects_for_bucket",
+    return_value=tests.data.aws.s3object.EMPTY_BUCKET_OBJECTS,
+)
+def test_sync_s3_objects_disabled(mock_get_objects, neo4j_session):
```
jychp, line 239: You're not actually testing that sync is disabled here.
Since you're mocking with an empty result list, the test will pass regardless of the limit value, so it doesn't validate the intended behavior.

tests/integration/cartography/intel/aws/test_s3object.py
```diff
@@ -0,0 +1,290 @@
...
+    )
+
+    mock_get_objects.assert_called()
+
+    call_args = mock_get_objects.call_args_list
+    for call in call_args:
+
+        assert call[0][3] == 0
+
+    nodes = neo4j_session.run(
```
jychp, line 284: Consider using check_nodes instead of a raw query

tests/integration/cartography/intel/aws/test_s3object.py
```diff
@@ -0,0 +1,290 @@
...
+        {
+            "UPDATE_TAG": TEST_UPDATE_TAG,
+            "AWS_ID": TEST_ACCOUNT_ID,
+            "aws_s3_object_max_per_bucket": 0,
+        },
+    )
+
+    mock_get_objects.assert_called()
+
+    call_args = mock_get_objects.call_args_list
```
jychp, line 279: Not sure what this code is doing, you should add a comment to clarify its purpose.
Also, it's unclear whether this logic is actually relevant or necessary

AdityaPandeyCN: I was trying to check if the function was called with limit=0, but was doing it in a really unclear way. I've removed it completely. The new test is much simpler, it just checks that when we set the limit to 0, the sync is actually disabled. Thanks for pointing this out

cartography/cli.py
```diff
@@ -256,6 +256,19 @@ def _build_parser(self):
                 " If not specified, cartography by default will run all AWS sync modules available."
             ),
         )
+        parser.add_argument(
+            "--aws-s3-object-max-per-bucket",
```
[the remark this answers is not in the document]
AdityaPandeyCN: done!

jychp:
Hi @AdityaPandeyCN can your handle the merge conflict ?
Sorry for the delay in PR review, was very busy recently.

cartography/intel/aws/s3_objects.py
```diff
@@ -0,0 +1,262 @@
...
+            Region=region,
+            AWS_ID=aws_account_id,
+        )
+
+    # Create owner relationships if owner information is present
+    owner_query = """
+    MATCH (o:S3Object{lastupdated: $update_tag})
+    WHERE o.owner_id IS NOT NULL
+    WITH o
+    MERGE (owner:AWSPrincipal{id: o.owner_id})
```
achantavy: let's not do this, see my other comment

cartography/models/aws/s3/s3_object.py
achantavy: > The issue is that the AWSPrincipal node might not exist yet.

This is fine, just run the graph sync twice and the owner will exist then :).

Create a relationship with a target node matcher

achantavy (changes requested):

achantavy:
@AdityaPandeyCN let me actually have a try at this, I'll push to this branch

AdityaPandeyCN:
Thankyou @achantavy I have checked edits from maintainers box, Do I have to do anything more?

AdityaPandeyCN:
I have addressed your suggestions.

PR #1586, opened 2025-05-26
feat: Added APIGatewayMethod ingestion and relationship

Khushmagrawal:
### Summary
This PR adds ingestion and relationships for AWS API Gateway Methods. Relationships include:
- (:APIGatewayResource)-[:RESOURCE]->(:APIGatewayMethod) 
- (:APIGatewayMethod)-[:INVOKES]->(:AWSLambda) 
- (:APIGatewayMethod)-[:ACCESSES]->(:S3Bucket) 
- (:APIGatewayMethod)-[:ACCESSES]->(:DynamoDBTable) 

### Related issues or links
( #1552 )

### Checklist

### DIFF
--- a/cartography/intel/aws/apigateway.py
+++ b/cartography/intel/aws/apigateway.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import re
 from typing import Any
 from typing import Dict
 from typing import List
@@ -15,6 +16,7 @@ from policyuniverse.policy import Policy
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.models.aws.apigateway import APIGatewayRestAPISchema
+from cartography.models.aws.apigateway_method import APIGatewayMethodSchema
 from cartography.models.aws.apigatewaycertificate import (
     APIGatewayClientCertificateSchema,
 )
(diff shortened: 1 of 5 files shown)


jychp (changes requested):
`sub_resource` (with label `RESOURCE`) should represent the relationship between the tenant (in this case, the `AWSAccount`) and the node, whenever applicable.

In this case, you should:

- [ ] Rename this relationship to `CONTAINS` (or use a more appropriate type, if needed)
- [ ] Move this relationship to `other_relationships`
- [ ] Create a new relationship pointing to the `AWSAccount` node
- [ ] Set this new relationship as the `sub_resource` relationship

Khushmagrawal:
> `sub_resource` (with label `RESOURCE`) should represent the relationship between the tenant (in this case, the `AWSAccount`) and the node, whenever applicable.
> 
> In this case, you should:
> 
> * [ ]  Rename this relationship to `CONTAINS` (or use a more appropriate type, if needed)
> * [ ]  Move this relationship to `other_relationships`
> * [ ]  Create a new relationship pointing to the `AWSAccount` node
> * [ ]  Set this new relationship as the `sub_resource` relationship

Oh yes, it should always represent the higher level parent so added and corrected the relationships, also updated the `cleanup()` (maintaining the order) and updated tests accordingly.

Khushmagrawal:
Also, there can be a relationship from APIGatewayMethod to AWSVpc via APIGatewayVpcLink, which would look like this:
(:APIGatewayMethod)-[:CONNECTS_VIA]->(:APIGatewayVpcLink)-[:CONNECTS_TO]->(:AWSVpc)
However, this requires first defining the `APIGatewayVpcLink` ([AWS_doc])(https://docs.aws.amazon.com/apigateway/latest/api/API_VpcLink.html)

I am planning to implement it in a future PR to keep this one focused.

cartography/intel/aws/apigateway.py
```diff
@@ -239,6 +390,8 @@ def transform_rest_api_details(
         if len(resource) > 0:
             for r in resource:
                 r["apiId"] = api_id
+                r["region"] = region
```
jychp, line 393: Why did you add this in the transformation ?

Region & Account ID are available as kwargs and are not needed to be set in the dict, right ?

jychp (changes requested):
Thanks for your contribution!
I think there are several places in your code where things could be simplified — and simpler is always better 🙂

If you're using AI tools like GitHub Copilot or Cody (which is totally fine, we're strong advocates of using AI), just make sure to provide enough context and guidance. For example, you can explicitly ask to keep things simple or reference existing files as examples to align with the project's conventions.

achantavy:
Hey @Khushmagrawal, are you available to follow up on the review comments? Totally get it if not, things get busy, just let us know

Khushmagrawal:
Hey @achantavy,  I'm currently tied up with some other commitments and likely won’t be able to address the review comments for another month. Would it make sense to close the PR for now and revisit it later? happy to follow whatever works best.

Khushmagrawal:
Closing this as #1819 implemented this. Thanks @d-aggarwal

PR #1603, opened 2025-05-31
feat: allow node_schema without sub resources to be cleaned up

achantavy:
### Summary
Enable cleanup jobs for node_schema objects without sub resource objects -- adds a `scoped_cleanup` flag to the node_schema. This allows a great deal of flexibility for future modules.

This PR is best understood by reading the tests.

### Checklist

  - Added new integration test `test_cleanup_unscoped_node_end_to_end` in `tests/integration/cartography/graph/test_cleanupbuilder_unscoped.py`
  - Test verifies the complete cleanup flow for unscoped nodes
  - Test includes assertions for node deletion, relationship cleanup, and preservation of unrelated nodes
  - Not applicable 
  - Not applicable 

  - Not applicable 

  - Not applicable

### DIFF
--- a/cartography/graph/cleanupbuilder.py
+++ b/cartography/graph/cleanupbuilder.py
@@ -15,6 +15,7 @@ from cartography.models.core.relationships import TargetNodeMatcher
 def build_cleanup_queries(node_schema: CartographyNodeSchema) -> List[str]:
     """
     Generates queries to clean up stale nodes and relationships from the given CartographyNodeSchema.
+    Properly handles cases where a node schema has a scoped cleanup or not.
     Note that auto-cleanups for a node with no relationships is not currently supported.
 
     Algorithm:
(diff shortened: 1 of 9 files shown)


cartography/graph/cleanupbuilder.py
```diff
@@ -53,15 +55,24 @@ def build_cleanup_queries(node_schema: CartographyNodeSchema) -> List[str]:
...
 
-    result = _build_cleanup_node_and_rel_queries(
-        node_schema,
-        node_schema.sub_resource_relationship,
-    )
+    result = []
+    if not node_schema.scoped_cleanup:
+        result = [_build_unscoped_cleanup_node_query(node_schema)]
+    # keep mypy happy by ensuring that the sub_resource_rel is not None here
+    elif node_schema.sub_resource_relationship:
```
jychp, line 62: I think having a `sub_resource_relationship` should supersede the `scoped_cleanup` flag, since it's the classic approach and aligns with how we want things to work by default.

If someone sets both `sub_resource_relationship` and `scoped_cleanup=False`, we should enforce scoped cleanup and emit a warning log to inform the user.

jychp: Just inverse conditions blocks order

jychp: and may be we should add a test for this edge case

cartography/graph/cleanupbuilder.py
```diff
@@ -94,6 +105,46 @@ def _build_cleanup_rel_query_no_sub_resource(
     )
 
 
+def _build_match_statement_for_cleanup(node_schema: CartographyNodeSchema) -> str:
+    """
+    Helper function to build a MATCH statement for a given node schema for cleanup.
+    """
+    if not node_schema.sub_resource_relationship and not node_schema.scoped_cleanup:
```
jychp, line 112: That's a perfect example of the comment above: inconsistent behavior if someone set the flag to false and define a subresource;
- line 62 will generate an unscoped MATCH
- here it will generate a scoped MATCH

cartography/graph/cleanupbuilder.py
```diff
@@ -185,6 +222,80 @@ def _build_cleanup_node_and_rel_queries(
...
+        $delete_action_clause
+        """,
+    )
+    return query_template.safe_substitute(
+        match_statement=_build_match_statement_for_cleanup(node_schema),
+        delete_action_clause=delete_action_clause,
+    )
+
+
+def _build_cleanup_rel_queries_unscoped(
```
jychp, line 258: for clarity you should keep the same naming pattern across that file:
- build_**unscoped**_cleanup_node_query
- build_cleanup_rel_requeries_**unscoped**

jychp: BTW, I don't understand why you introduced this function, since you already modified the existing `_build_cleanup_node_and_rel_queries` to include a conditional match. These two changes seem redundant.

jychp (changes requested):
Thanks for the PR, this is going to be super helpful!
Just a few questions/requests, but they should be quick to address :)

cartography/graph/cleanupbuilder.py
achantavy: i tried repurposing but the logic is different enough where it's even more of a mess to have in one place with all the special casing for whether something has a sub resource or not. it's easier to handle as separate functions. there probably is an even cleaner way of doing it but right now i figure it'll be eventually easier to consolidate rather than have me try to make something tightly packed right now

jychp (approved):
LGTM !

Issue #1594, opened 2025-05-28
Bug in Entra OU module


kunaals:
**Bug Template**
*Description*:
When running the Entra module in 0.103.0, I get the following error that seems to have come from [this PR](https://github.com/cartography-cncf/cartography/pull/1538).
`AttributeError: 'AdministrativeUnitsRequestBuilder' object has no attribute 'request'`

*To Reproduce*:
Just run the Entra sync

*Please complete the following information:*:
- Cartography release version: 0.103.0
- Python version: 3.13

*Logs*:
```
Traceback (most recent call last):
  File "/Users/kunaalsikka/src/subimage/subimage-backend/.venv/lib/python3.13/site-packages/cartography/sync.py", line 131, in run
    stage_func(neo4j_session, config)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kunaalsikka/src/subimage/subimage-backend/.venv/lib/python3.13/site-packages/cartography/util.py", line 202, in timed
    return method(*args, **kwargs)
  File "/Users/kunaalsikka/src/subimage/subimage-backend/.venv/lib/python3.13/site-packages/cartography/intel/entra/__init__.py", line 61, in start_entra_ingestion
    asyncio.run(main())
    ~~~~~~~~~~~^^^^^^^^
  File "/Users/kunaalsikka/.local/share/uv/python/cpython-3.13.2-macos-aarch64-none/lib/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/Users/kunaalsikka/.local/share/uv/python/cpython-3.13.2-macos-aarch64-none/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/kunaalsikka/.local/share/uv/python/cpython-3.13.2-macos-aarch64-none/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/Users/kunaalsikka/src/subimage/subimage-backend/.venv/lib/python3.13/site-packages/cartography/intel/entra/__init__.py", line 51, in main
    await sync_entra_ous(
    ...<6 lines>...
    )
[...]

PR #1605, opened 2025-06-02
fix: Entra OU Module: .request() was removed from MS SDK

shyammukund:
Fixed Bug in EntraOU Module shown in https://github.com/cartography-cncf/cartography/issues/1594

The bug was caused by an incompatible change in the Microsoft Graph Python SDK where the .request() method was removed from AdministrativeUnitsRequestBuilder. Previous code called client.directory.administrative_units.request() which no longer exists in newer SDK versions. Code is now updated in cartography/intel/entra/ou.py so that the updated SDK API pattern is used while maintaining the same OU synchronization functionality. Also updated the entra docs to include required permissions. 

Screenshot 2025-06-02 at 11 16 42 AM

### Test Passing after Fixing Bug
Screenshot 2025-06-02 at 12 44 07 PM

### Graph After Changes
Screenshot 2025-06-02 at 12 44 25 PM

### DIFF
--- a/cartography/intel/entra/ou.py
+++ b/cartography/intel/entra/ou.py
@@ -22,19 +22,21 @@ async def get_entra_ous(client: GraphServiceClient) -> list[AdministrativeUnit]:
     Get all OUs from Microsoft Graph API with pagination support
     """
     all_units: list[AdministrativeUnit] = []
-    
+
     # Initialize first page request
     current_request = client.directory.administrative_units
-    
+
     while current_request:
         try:
             response = await current_request.get()
             if response and response.value:
                 all_units.extend(response.value)
-                
+
                 # Handle next page using OData link
                 if response.odata_next_link:
-                    current_request = client.directory.administrative_units.with_url(response.odata_next_link)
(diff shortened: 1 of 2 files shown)


achantavy:
Hey @shyammukund, can you please show that this works? Example, maybe a screenshot showing the error happening and then a screenshot after showing the error no longer there?

shyammukund:
@achantavy Do you mean a screenshot of the cli ouput passing or a screenshot update of the neo4j database?

achantavy:
> @achantavy Do you mean a screenshot of the cli ouput passing or a screenshot update of the neo4j database?

Screenshots of cartography actually running on Entra. First one showing that it crashed, second one showing that it didn't.

--

(Side note, I'm a bit confused about the integration test failure you shared here: if the int test failed, then CI should have blocked your change from being merged in)

shyammukund:
@achantavy I misunderstood what you meant about showing tests passing (thought you meant integration tests). have gone ahead and uploaded the requested screenshots as well as the neo4j graph showing that it is now working.

PR #1612, opened 2025-06-07
Address config option mismatch in SnipeIT authentication flow

emmanuel-ferdman:
### Summary
This PR addresses a configuration bug in the CLI where the SnipeIT integration was incorrectly referencing `config.kandji_token` instead of the intended `config.snipeit_token`. The fix ensures that the correct token is used for SnipeIT-related operations, improving authentication accuracy and preventing potential misconfigurations. I believe this bug happened due to copy-paste from the Kandji section.

### Checklist


jychp (approved):
Good catch.
Did you have any impact on run time ?
The only one I seen is Kandji not running due to API key reset if you launch the CLI without providing `selected-modules`

PR #1617, opened 2025-06-09
Add AGENTS.md to make cartography more editable by agents

achantavy:
I, for one, welcome our new robot overlords.

Adds an AGENTS.md file to make cartography more easily editable by AI coding assistants.

https://github.com/cartography-cncf/cartography/pull/1614 is an example of a PR that was created in 1 or 2 steps with an AI agent.

OpenAI Codex should pick up on this automatically. For other agents you may need to reference this file explicitly.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,885 @@
+# AGENTS.md: Cartography Intel Module Development Guide
+
+> **For AI Coding Assistants**: This document provides comprehensive guidance for understanding and developing Cartography intel modules. It contains codebase-specific patterns, architectural decisions, and implementation details necessary for effective AI-assisted development within the Cartography project.
+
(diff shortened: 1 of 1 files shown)


chandanchowdhury:
High level observation: Will it help to have `Data Model: Defining Nodes and Relationships {#data-model}` before `Main Entry Point (`__init__.py`)` ?

achantavy:
> High level observation: Will it help to have `Data Model: Defining Nodes and Relationships {#data-model}` before `Main Entry Point (`**init**.py`)` ?

In practice, didn't make a difference. It goes in order: 

get + transform + load + cleanup, and then it only needs to know how to do the data model part at the load phase.

chandanchowdhury (approved):

PR #1627, opened 2025-06-12
Updates dev.Dockerfile to do an editable install of cartography with uv

achantavy:
Updates dev.Dockerfile to do an editable install of cartography source tree with uv.

dev.Dockerfile is intended for lint, unit, and integration testing of cartography, so an editable install is necessary.

As shown in the video below, this change also fixes debug breakpoint behavior in IDEs like Pycharm.

### Checklist

Before
https://github.com/user-attachments/assets/933c5292-2c2c-440b-8200-1c23a4ea6b30

After
https://github.com/user-attachments/assets/64a099a2-982d-4031-a72a-117d12327af9

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,4 @@ venv/
 .compose
 .cache
 .local
+.venv
--- a/dev-entrypoint.sh
+++ b/dev-entrypoint.sh
@@ -3,8 +3,5 @@ while ! git rev-parse --is-inside-work-tree >/dev/null 2>&1; do
   echo "Waiting for Git to be ready..."
   sleep 1
 done
-# Activate the virtual environment
-# This allow to used the virtual environment without having to specify `uv run`
-. .venv/bin/activate
 # Pass control to main container command
 exec "$@"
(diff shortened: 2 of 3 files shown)


achantavy:
@jychp -- I'm open for feedback here haha. Future PR should probably fix the pyproject.toml so that uv install actually works

achantavy:
cc: @d-aggarwal -- this might help with your uv struggles

.dockerignore
```diff
@@ -4,3 +4,4 @@ venv/
 .compose
 .cache
 .local
+.venv
```
achantavy, line 7: Prevent local .venv folder from being copied into the Docker build

jychp (dismissed):
LGTM

Personally, I prefer using uv sync rather than activating the virtual environment manually each time but that’s just a personal preference, and this PR doesn’t prevent that workflow.

jychp (changes requested):
You can use `RUN uv pip install -e . --system` in docker file to install cartography in editable mode. I think this was your initial intention, right ?

achantavy:
I think we can revisit later, I think this hasn't been a pain point for a while

PR #1628, opened 2025-06-12
feat: #1552 Add EfsMountTarget and EfsFileSystem as resource (also fixes #1609)

### LINKED ISSUE #1609: EFS sync fails because a required parameter is not specified.
*Description*:

AWS EFS sync fails because param is not specified.

*To Reproduce*:

*Logs*:

```
the below code yields this problem

INFO:backoff:Backing off inner_function(...) for 0.1s (botocore.errorfactory.BadRequest: An error occurred (BadRequest) when calling the DescribeMountTargets operation: Must specify exactly one mutually exclusive parameter.)

```

*Screenshots*:

0.104.0rc3

- https://github.com/cartography-cncf/cartography/blob/601c9601fbd256dc8f5452c45106e5ad31670715/cartography/intel/aws/efs.py#L27

- https://docs.aws.amazon.com/efs/latest/ug/API_DescribeMountTargets.html

To retrieve this information, we cannot just call a paginator on it, we need a parameter.

d-aggarwal:
Added EfsMountTarget as resource and fixed mount target fetch logic 

### Related issues or links
[#1552](
https://github.com/cartography-cncf/cartography/issues/1552) 

### Checklist

logs_efs

Relationship between EfsFileSystem and EfsMountTarget

image


cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,112 @@
...
+from cartography.models.aws.efs.mount_target import EfsMountTargetSchema
+from cartography.util import aws_handle_regions
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+@aws_handle_regions
+def get_efs_file_system_ids(boto3_session: boto3.Session, region: str) -> List[str]:
```
achantavy: Ah, so this is the main change over https://github.com/cartography-cncf/cartography/pull/1573/files?

achantavy (approved):
Thank you for the follow-up! Feel free to merge whenever

cartography/intel/aws/efs.py
achantavy: Actually, now that I think about it, based on these docs

- https://docs.aws.amazon.com/efs/latest/ug/API_DescribeFileSystems.html
- https://docs.aws.amazon.com/efs/latest/ug/API_DescribeMountTargets.html
- https://docs.aws.amazon.com/efs/latest/ug/whatisefs.html


would it make sense to include the file system as its own node, and then attach the mount target to the file system?

achantavy: Yeah I think that's the right way. I get that you're trying to keep the PR small (thank you!) but it feels a bit weird to ingest the mount target without ingesting the file system. The correct order of PRs might have been to do file system first and then mount target, but now that we're here, that's ok. I think it'd be a good idea to add the file systems to this PR too. File systems are also covered as part of #1552

d-aggarwal: Thanks for the review! Yeah, it felt a bit off to me as well..
You're right — we should add a node for the file system and create a relationship between the mount target and the file system.
Just to confirm: there won’t be any change to the current commits — we’ll still keep the get_efs_file_system_ids function in mocktarget, correct?
I’ll go ahead and add the node and relationship in this PR itself!

achantavy: > Just to confirm: there won’t be any change to the current commits — we’ll still keep the get_efs_file_system_ids function in mocktarget, correct?

I would update the tests.

## Option 1: separate tests

Have a file systems test that does this:

1. Arrange: <if needed> set up mocks and such
2. Act: run sync_efs_file_systems()
3. Assert that the nodes and rels are expected

And then for the mount targets tests do:

1. Arrange: <if needed> set up mocks and such, make sure that sync_efs_file_systems() is called in this step since it is required for syncing mount targets
2. Act: run sync_mount_targets()
3. Assert that the nodes and rels are expected, including that mount targets are connected to the efs file system nodes


## Option 2: one combined test

1. Arrange: set up mocks and such
2. Act: call sync_efs_file_systems() and sync_mount_targets(). In this option, both sync_*() functions are considered to be the functions under test
3. Assert that the nodes and rels are written as expected


Either way works, up to you

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,199 @@
...
+            tranformed_file_systems,
+            region,
+            current_aws_account_id,
+            update_tag,
+        )
+
+        mountTargets = get_efs_mount_targets(fileSystems, boto3_session, region)
+        mount_target_data: List[Dict[str, Any]] = []
+        for mountTarget in mountTargets:
+            mount_target_data.append(mountTarget)
```
achantavy: Why not just pass `mountTargets` directly into `load_efs_mount_targets`? This additional step uses unnecessary extra memory

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,199 @@
...
+    Transform SNS topic data for ingestion
+    """
+    transformed_file_systems = []
+    for file_system in fileSystems:
+        transformed_file_system = {
+            "FileSystemId": file_system["FileSystemId"],
+            "FileSystemArn": file_system["FileSystemArn"],
+            "OwnerId": file_system.get("OwnerId"),
+            "CreationToken": file_system.get("CreationToken"),
+            "CreationTime": dict_date_to_epoch(
```
achantavy: Don't use dict_date_to_epoch. Keep the value as is. I added a note about this here: https://github.com/cartography-cncf/cartography/blob/master/AGENTS.md#date-handling

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,199 @@
...
+                file_system, file_system.get("CreationTime", "")
+            ),
+            "LifeCycleState": file_system.get("LifeCycleState"),
+            "Name": file_system.get("Name"),
+            "NumberOfMountTargets": file_system.get("NumberOfMountTargets"),
+            "SizeInBytesValue": file_system.get("SizeInBytes", {}).get("Value"),
+            "SizeInBytesTimestamp": dict_date_to_epoch(
```
achantavy: Same here https://github.com/cartography-cncf/cartography/blob/master/AGENTS.md#date-handling

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    transformed_file_systems = []
+    for file_system in fileSystems:
+        transformed_file_system = {
+            "FileSystemId": file_system["FileSystemId"],
+            "FileSystemArn": file_system["FileSystemArn"],
+            "OwnerId": file_system.get("OwnerId"),
+            "CreationToken": file_system.get("CreationToken"),
+            "CreationTime": dict_date_to_epoch(
+                file_system, file_system.get("CreationTime", "")
```
achantavy: Don't use empty string. do `objectname.get('fieldname')` instead of `objectname.get('fieldname', '')`.

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,199 @@
...
+            "SizeInBytesTimestamp": dict_date_to_epoch(
+                file_system, file_system.get("SizeInBytes", {}).get("Timestamp", "")
+            ),
+            "PerformanceMode": file_system.get("PerformanceMode"),
+            "Encrypted": file_system.get("Encrypted"),
+            "KmsKeyId": file_system.get("KmsKeyId"),
+            "ThroughputMode": file_system.get("ThroughputMode"),
+            "AvailabilityZoneName": file_system.get("AvailabilityZoneName"),
+            "AvailabilityZoneId": file_system.get("AvailabilityZoneId"),
+            "FileSystemProtection": file_system.get("FileSystemProtection", {}).get(
```
achantavy, line 61: don't use empty string

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,199 @@
...
+            "CreationToken": file_system.get("CreationToken"),
+            "CreationTime": dict_date_to_epoch(
+                file_system, file_system.get("CreationTime", "")
+            ),
+            "LifeCycleState": file_system.get("LifeCycleState"),
+            "Name": file_system.get("Name"),
+            "NumberOfMountTargets": file_system.get("NumberOfMountTargets"),
+            "SizeInBytesValue": file_system.get("SizeInBytes", {}).get("Value"),
+            "SizeInBytesTimestamp": dict_date_to_epoch(
+                file_system, file_system.get("SizeInBytes", {}).get("Timestamp", "")
```
achantavy: Don't use empty string

achantavy (changes requested):
Thanks for the follow-up, left some small comments on the transform

d-aggarwal:
i have fixed the issues, kindly review @achantavy

d-aggarwal:
Hi @jychp , when you get a chance, kindly review this PR. Thanks!

docs/root/modules/aws/schema.md
```diff
@@ -3279,6 +3279,63 @@ Representation of an AWS ECS [Container](https://docs.aws.amazon.com/AmazonECS/l
...
     (ECSTask)-[HAS_CONTAINER]->(ECSContainer)
     ```
+
+### EfsFileSystem
+Representation of an AWS [EFS File System](https://docs.aws.amazon.com/efs/latest/ug/API_FileSystemDescription.html)
+| Field | Description |
+|-------|-------------|
+| **id** | The ID of the file system, assigned by Amazon EFS |
```
jychp, line 3320: Can you add the region field you updated in the model ?

docs/root/modules/aws/schema.md
```diff
@@ -3279,6 +3279,63 @@ Representation of an AWS ECS [Container](https://docs.aws.amazon.com/AmazonECS/l
...
+   (AWSAccount)-[RESOURCE]->(EfsFileSystem)
+   ```
+
+### EfsMountTarget
+Representation of an AWS [EFS Mount Target](https://docs.aws.amazon.com/efs/latest/ug/API_MountTargetDescription.html)
+| Field | Description |
+|-------|-------------|
+| **id** | System-assigned mount target ID |
+| arn | System-assigned mount target ID |
+| fileSystem_id | The ID of the file system for which the mount target is intended |
```
jychp, line 3352: Same here, can you add the region field ?

jychp (changes requested):
LGTM

Minor change in doc and we will be able to merge it, thanks

d-aggarwal:
> LGTM

I have added region in docs of schema

Issue #1591, opened 2025-05-27
Analysis job failure: aws_s3acl_analysis.json


jychp:
*Description*:
The `aws_s3acl_analysis.json` analysis job fail if AWS module is not called in the same execution.

The analysis job seems to rely on a AWS_ID parameter that is only presents in the COMMON_JOB_PARAMETERS if the aws intel populate him from config.

*To Reproduce*:
```
uv run cartography --selected-modules analysis --analysis-job-directory ./cartography/data/jobs/analysis/
```

*Logs*:
```
ERROR:cartography.graph.job:Unhandled error while executing statement in job 'AWS S3 Acl exposure analysis': {code: Neo.ClientError.Statement.ParameterMissing} {message: Expected parameter(s): AWS_ID}
ERROR:cartography.intel.analysis:An exception occurred while executing discovered analysis job: cartography/data/jobs/analysis/aws_s3acl_analysis.json
Traceback (most recent call last):
  File "/home/jychp/Repos/cartography/cartography/intel/analysis.py", line 34, in run
    GraphJob.run_from_json_file(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        path,
        ^^^^^
        neo4j_session,
        ^^^^^^^^^^^^^^
        {"UPDATE_TAG": config.update_tag},
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/jychp/Repos/cartography/cartography/graph/job.py", line 229, in run_from_json_file
    job.run(neo4j_session)
    ~~~~~~~^^^^^^^^^^^^^^^
  File "/home/jychp/Repos/cartography/cartography/graph/job.py", line 100, in run
    stm.run(neo4j_session)
    ~~~~~~~^^^^^^^^^^^^^^^
  File "/home/jychp/Repos/cartography/cartography/graph/statement.py", line 77, in run
    session.write_transaction(self._run_noniterative)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jychp/Repos/cartography/.venv/lib/python3.13/site-packages/neo4j/_meta.py", line 243, in inner
    return f(*args, **kwargs)
  File "/home/jychp/Repos/cartography/.venv/lib/python3.13/site-packages/neo4j/_sync/work/session.py", line 848, in write_transaction
    return self._run_transaction(
[...]

PR #1633, opened 2025-06-15
feat: Ontology - Phase 1 - Users & Devices

jychp:
This PR introduces the first iteration of the ontology  (see https://github.com/cartography-cncf/cartography/discussions/1579)!

It includes the following changes:

* Adds a semantic `UserAccount` label to all nodes representing user accounts (as an extra label).
* Introduces an abstract `User` node (replacing `Human`):

  * `User` nodes are automatically linked to any `UserAccount` node that has an email field.
  * Custom relationships can be added in specific cases.
  * Any new module defining a node with the `UserAccount` label and the appropriate field will automatically integrate with the ontology.
* Introduces an abstract `Device` node to model user-owned devices

**Structure:**

* The ontology is defined in a dedicated model: `cartography.models.ontology`.
* Node creation and relationship mapping are handled in a dedicated ingestion module: `cartography.intel.ontology`.

**Simplified algorithm:**

* For each module defined as a source of truth, existing nodes are fetched.
* Abstract nodes are generated according to a defined mapping.
* Abstract nodes are ingested into the graph.
* It's possible to define module-specific Cypher queries; in this PR, some modules include a Cypher statement to create the `(User)-[OWNS]->(Device)` relationship based on their data.

# Examples

Here is an example with Duo/Tailscale/SnipeIt before and after this PR:
Screenshot From 2025-06-17 22-55-40
Screenshot From 2025-06-17 22-55-18

You can also have only the Ontology by simply queryin `MATCH (o:Ontology) RETURN o`
Screenshot From 2025-06-18 15-05-31

# How to test
You can simply test this PR using the env demo
> uv run python3 -m demo

### DIFF
--- a/cartography/models/ontology/mapping/data/devices.py
+++ b/cartography/models/ontology/mapping/data/devices.py
@@ -7,7 +7,7 @@ bigfix_mapping = OntologyMapping(
     module_name="bigfix",
     nodes=[
         OntologyNodeMapping(
-            node_label="BigFixComputer",
+            node_label="BigfixComputer",
             fields=[
                 OntologyFieldMapping(
                     ontology_field="hostname", node_field="computername"
--- a/cartography/models/ontology/mapping/data/users.py
+++ b/cartography/models/ontology/mapping/data/users.py
@@ -131,7 +131,7 @@ lastpass_mapping = OntologyMapping(
     module_name="lastpass",
     nodes=[
         OntologyNodeMapping(
-            node_label="LastPassUser",
+            node_label="LastpassUser",
             fields=[
                 OntologyFieldMapping(ontology_field="email", node_field="email"),
                 OntologyFieldMapping(ontology_field="fullname", node_field="name"),
--- a/tests/unit/cartography/intel/ontology/test_ontology_mapping.py
+++ b/tests/unit/cartography/intel/ontology/test_ontology_mapping.py
@@ -1,4 +1,3 @@
-import warnings
 from typing import Type
 
 import cartography.models
@@ -35,23 +34,15 @@ def _get_model_by_node_label(node_label: str) -> Type[CartographyNodeSchema] | N
 def test_ontology_mapping_fields():
     # Verify that all ontology fields in the mapping exist as extra indexed fields
     # in the corresponding module's model.
-    for category, mappings in ONTOLOGY_MAPPING.items():
+    for _, mappings in ONTOLOGY_MAPPING.items():
         for module_name, mapping in mappings.items():
             for node in mapping.nodes:
                 # Load the model class for the module
                 model_class = _get_model_by_node_label(node.node_label)
-
-                # TODO: Migrate to assertion when gsuite is migrated to the new ontology models
-                # assert model_class is not None, (
-                #     f"Model class for node label '{node.node_label}' "
-                #     f"in module '{module_name}' not found."
-                # )
-                if model_class is None:
-                    warnings.warn(
-                        f"Model class for node label '{node.node_label}' in module '{module_name}' not found.",
-                        UserWarning,
-                    )
-                    continue
+                assert model_class is not None, (
+                    f"Model class for node label '{node.node_label}' "
+                    f"in module '{module_name}' not found."
+                )
 
                 # Check all ontology fields are in extra indexed fields
                 for mapping_field in node.fields:


cartography/data/ontology/devices.json
```diff
@@ -0,0 +1,93 @@
+{
```
achantavy: Should this be defined as a formal Python class instead so that we get typechecks and avoid potential typos in fields?

cartography/data/ontology/devices.json
```diff
@@ -0,0 +1,93 @@
...
+            },
+            "DuoPhone": {
+                "hostname": "name",
+                "os_version": "os_version",
+                "model": "model",
+                "platform": "platform"
+            }
+        },
+        "rels": [
+            {
```
achantavy: [thinking] I wonder if this could be represented as a list of RelSchema or MatchLinks.

If RelSchema, that won't work because some of the patterns can be complicated.

If Matchlink, that also won't be nice because Matchlinks are more imperative and not declarative.

In this case I think raw cypher (which _is_ declarative) is actually a good fit.

cartography/intel/gsuite/api.py
```diff
@@ -193,6 +193,7 @@ def load_gsuite_users(
         u.user_id = user.id,
         u.firstseen = $UpdateTag
         SET
+        u:UserAccount,
```
achantavy: In the data model itself, we could introduce a special field for "OntologyLabel". Would this help?

cartography/intel/ontology/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+
+@timeit
+def run(neo4j_session: neo4j.Session, config: Config) -> None:
+    common_job_parameters = {
+        "UPDATE_TAG": config.update_tag,
+    }
+
+    # Get source of truth from config
+    if config.ontology_users_source:
+        users_source_of_truth = config.ontology_users_source.split(",")
```
achantavy: [note to self] The terminology for these concepts should be documented

cartography/intel/ontology/users.py
```diff
@@ -0,0 +1,56 @@
...
+        update_tag,
+    )
+    link_ontology_nodes(neo4j_session, "users", update_tag)
+    cleanup(neo4j_session, common_job_parameters)
+
+
+@timeit
+def load_users(
+    neo4j_session: neo4j.Session,
+    data: List[Dict[str, Any]],
```
achantavy: Nit: use primitive `list` and `dict` types in new code

cartography/data/ontology/devices.json
```diff
@@ -0,0 +1,93 @@
+{
+
```
kunaals:
```suggestion
```

cartography/data/ontology/users.json
```diff
@@ -0,0 +1,129 @@
+{
+    "anthropic": {
+        "nodes": {
+            "AnthropicUser": {
+
```
kunaals:
```suggestion
```
lets generally get rid of all the extra blank lines in the json, perhaps we should add a json formatter to the linter?

cartography/models/tailscale/user.py
```diff
@@ -15,7 +16,8 @@ class TailscaleUserNodeProperties(CartographyNodeProperties):
     id: PropertyRef = PropertyRef("id")
     lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
     display_name: PropertyRef = PropertyRef("displayName")
-    login_name: PropertyRef = PropertyRef("loginName", extra_index=True)
+    login_name: PropertyRef = PropertyRef("loginName")
+    email: PropertyRef = PropertyRef("loginName", extra_index=True)
```
kunaals, line 20: Will we always want these ontology fields to have extra_index? Might be worth calling that out in the docs

cartography/models/ontology/device.py
```diff
@@ -0,0 +1,125 @@
...
+    platform: PropertyRef = PropertyRef("platform")
+    serial_number: PropertyRef = PropertyRef("serial_number", extra_index=True)
+
+
+@dataclass(frozen=True)
+class DeviceToNodeRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+# (:Device)-[:HAS_OBSERVATION]->(:DuoEndpoint)
```
kunaals: I think the name `HAS_OBSERVATION` is slightly confusing

I understand that the intent is that the source record of the `(:Device)` is the `(:ModuleDevice)`. I think the reason that it feels less intuitive than, say, `(:User)-[:HAS_ACCOUNT]-(:UserAccount)` is because there is no "observation" field or node anywhere. I think something like `[:OBSERVED_AS]` or even `[:SOURCE]` might be clearer. Open to ideas here though

cartography/models/ontology/device.py
```diff
@@ -0,0 +1,125 @@
...
+    target_node_label: str = "CrowdstrikeHost"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"hostname": PropertyRef("hostname")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "HAS_OBSERVATION"
+    properties: DeviceToNodeRelProperties = DeviceToNodeRelProperties()
+
+
+# (:Device)-[:RESOURCE]->(:BigfixComputer)
```
kunaals: For some reason uses different label `[:RESOURCE]`

cartography/intel/ontology/__init__.py
kunaals: Make sure to trim() here so that input like `"gsuite ,   entra"` doesn't break things. Ditto for `computers_source_of_truth`

cartography/intel/ontology/utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+import neo4j
+
+from cartography.graph.job import GraphJob
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+def load_ontology_mapping(module_name: str) -> dict:
```
achantavy: [nonblock] Consider changing the word "load" to a different verb since it has a different meaning in cartography

cartography/intel/ontology/utils.py
```diff
@@ -0,0 +1,118 @@
...
+    Args:
+        neo4j_session (neo4j.Session): The Neo4j session to use for querying the database.
+        source_of_truth (list[str]): A list of source of truth identifiers to filter the modules.
+        module_name (str): The name of the ontology module to use for the mapping (eg. users, devices, etc.).
+
+    Returns:
+        list[dict[str, Any]]: A list of dictionaries, each containing a node details formatted according to the ontology mapping.
+    """
+    results: list[dict[str, Any]] = []
+    modules_mapping = load_ontology_mapping(module_name)
```
achantavy: Echoing an earlier comment in one of the json files, it'd be nice to have this be a typed object

cartography/intel/ontology/utils.py
```diff
@@ -0,0 +1,118 @@
...
+        for node_label, fields in modules_mapping[source].get("nodes", {}).items():
+            if not isinstance(fields, dict):
+                logger.warning(
+                    "Node fields for '%s' in '%s' are not a dictionary.",
+                    node_label,
+                    source,
+                )
+                continue
+            query = f"MATCH (n:{node_label}) RETURN n"
+            for node in neo4j_session.run(query):
```
achantavy: Use `execute_read()` for transaction handling

cartography/models/airbyte/user.py
```diff
@@ -98,6 +99,7 @@ class AirbyteUserToWorkspaceMemberRel(CartographyRelSchema):
 @dataclass(frozen=True)
 class AirbyteUserSchema(CartographyNodeSchema):
     label: str = "AirbyteUser"
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["UserAccount"])
```
achantavy: Echoing earlier comment: we could make an extra field for 'ontology node label' or similar if it'd be useful.

cartography/cli.py
```diff
@@ -730,6 +730,26 @@ def _build_parser(self):
                 "Required if you are using the Trivy module. Ignored otherwise."
             ),
         )
+        parser.add_argument(
+            "--ontology-users-source",
```
achantavy, line 734: If experimental, make it `--experimental-ontology-users-source`

cartography/cli.py
```diff
@@ -730,6 +730,26 @@ def _build_parser(self):
...
+            type=str,
+            default=None,
+            help=(
+                "[EXPERIMENTAL!] Comma-separated list of sources of truth for user data in the ontology. "
```
achantavy: If experimental, make it `--experimental-ontology-devices-source`

demo/__main__.py
```diff
@@ -104,6 +107,11 @@ def main(force_flag: bool, analysis_job_directory: str) -> None:
         except Exception:
             logger.exception("Seed %s failed", seed_cls.__name__)
 
+    # Ontology
+    logger.info("Building ontology ...")
+    ontology.run(neo4j_session, config)
+
+    # TODO: Analysis: blocked due to https://github.com/cartography-cncf/cartography/issues/1591
```
achantavy: Fixed now? https://github.com/cartography-cncf/cartography/issues/1591

docs/root/modules/ontology/index.md
```diff
@@ -0,0 +1,68 @@
...
+- Validating the consistency of entities and relationships
+- Ensuring mapping between collected data and the defined semantic model
+
+The module provides functions to traverse, enrich, and leverage the ontology during data ingestion. It plays a key role in normalizing entities from heterogeneous sources.
+
+By default, nodes are created in the ontology based on data observed in various modules. For some node, such as User, Device, etc., you can specify "sources of truth" modules that will exclusively create those nodes. This allows for more controlled and accurate representation of certain entities.
+
+**Example: User Nodes and Source of Truth**
+
+If you set the `--ontology-users-source` parameter to `duo`, then a `User` node will be created for every account found in Duo. In contrast, for other integrations like Tailscale, only existing `User` nodes (those created by the source of truth) will be linked to Tailscale accounts. No new `User` nodes will be created from Tailscale data alone.
```
achantavy, line 34: Double-checking: is the word "source" used here consistent with how others use it in ontologies?

cartography/models/tailscale/user.py
jychp: I’ll add it to the ontology documentation, but I want to keep it simple and transparent from the module development perspective.
There’s also an `ensure_index` function in the query helper, so even if the index is missing, it will be created.
I’ll also try to add a unit test to verify that.

cartography/models/ontology/device.py
jychp: Fixed. bad copy pasta.

demo/__main__.py
jychp: Good catch, artifact from the main branch merge

cartography/intel/ontology/utils.py
jychp: Deprecated: this function will be deleted with the migration from JSON to Dataclass

cartography/data/ontology/devices.json
jychp: Deprecated, migrated to python dataclasses instead

cartography/data/ontology/users.json
jychp: Deprecated, migrated to python dataclasses instead

cartography/intel/gsuite/api.py
jychp: I do not think we need an other field, that will increase complexity for module developper.
I suggest instead adding an inline comment saying that this extra node is used by ontology.

cartography/models/ontology/mapping/specs.py
```diff
@@ -0,0 +1,28 @@
+from dataclasses import dataclass
+from dataclasses import field
```
achantavy, line 2: I prefer a directory structure where models / classes are in one folder, maybe called `spec/` and the instantiations of those are saved in a folder called `data/`.

jychp: Not sure of what you are thinking about but something like:
- /mapping/spec.py
- /mapping/data/users.py
- /mapping/data/devices.py

Was also questioning myself about putting this mapping thing under models/ instead of intel/

achantavy: Yeah I kinda think that the mapping should be under models and not intel. Up to you though

PR #1641, opened 2025-06-18
fix #1640: check_nodes is used without `assert` instruction in some tests 

### LINKED ISSUE #1640: check_nodes is used without `assert` instruction in some tests, resulting in no real test
*Description*:
In some test files `check_nodes` helper is used without `assert` leading to no test at all

- https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/aws/test_identitycenter.py
- https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/kandji/test_kandji.py
- https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/snipeit/test_snipeit_assets.py
- https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/snipeit/test_snipeit_users.py

jychp:
### Summary
Fixes #1640

---

While fixing the issue I discovered that the untested AWSSSOUser has an issue:
- test data stub refering to `ExternalIds` with is a list of dict (ref: https://github.com/cartography-cncf/cartography/blob/97678da15d5f8c3ff3cd496e14498b2959b893d0/tests/data/aws/identitycenter.py#L5)
- SSOUser model refering to `EternalId` (ref: https://github.com/cartography-cncf/cartography/blob/97678da15d5f8c3ff3cd496e14498b2959b893d0/cartography/models/aws/identitycenter/awsssouser.py#L20)
- by the way the link between SSOUser and Okta `UserAccount` should not work

I do not use it in production so I do not know if there is an issue with our test data or if the module do not work at all.
Any help is welcome :)

### DIFF
--- a/tests/integration/cartography/intel/aws/test_identitycenter.py
+++ b/tests/integration/cartography/intel/aws/test_identitycenter.py
@@ -23,7 +23,9 @@ def test_load_sso_users(neo4j_session):
     )
 
     # Use check_nodes to verify that the SSO users are correctly loaded
-    check_nodes(neo4j_session, "AWSSSOUser", ["id", "external_id"])
+    assert check_nodes(neo4j_session, "AWSSSOUser", ["id", "external_id"]) == {
+        ("aaaaaaaa-a0d1-aaac-5af0-59c813ec7671", "")
+    }
 
 
 def test_load_identity_center_instances(neo4j_session):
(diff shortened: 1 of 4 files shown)


achantavy:
lol this is a silly bug, thank you for catching it

achantavy (approved):

PR #1642, opened 2025-06-19
chore: pin dependencies for build CI

jychp:
### Summary

Switch CI build system to [uv](https://github.com/astral-sh/uv).

### Context

This PR replaces the legacy `pypa build` action with `uv build`, offering both performance and security improvements:

* Faster build times
* Safer builds by using only pinned dependencies
* Avoids unintentional resolution or fetching of unpinned packages

This change was motivated by a security alert.

### Related issues or links

* https://github.com/cartography-cncf/cartography/security/code-scanning/51

### Changes proposed

* **Switch to `uv build`**: Replaces legacy `python -m build` with `uv build` in our build workflow
* **Hardens build reproducibility**: Ensures only explicitly pinned dependencies are used at build time

## How to test this PR

Build the package locally using:

```bash
uv build
```

You should obtain the same `.tar.gz` and `.whl` artifacts in the `dist/` directory, faster and with reduced dependency resolution.


.github/workflows/publish-to-ghcr-and-pypi.yml
```diff
@@ -28,10 +28,7 @@ jobs:
           python-version: "3.10"
       - name: Install pypa/build
         run: >-
-          python -m
-          pip install
-          build[uv]
-          --user
+          python -m pip install --require-hashes -r build-requirements.txt --user
```
achantavy: dumb question but how does build-requirements.txt get updated? is this a common convention?

jychp: dependabot supports `*requirements.txt`, so it's kind of `commonn convention`

but uv now support dynamic version without bug, so I just pushed a more straight forward way to do it ;)

serge-wq (approved):

PR #1647, opened 2025-06-21
Add CodeBuild Project

d-aggarwal:
Add CodeBuild Project as resource in AWS Module 

Referenced CloudGoat Scenario - [codebuild_secrets](https://github.com/RhinoSecurityLabs/cloudgoat/tree/master/cloudgoat/scenarios/aws/codebuild_secrets)

### Related issues or links

### Checklist

- [] Include console log trace showing what happened before and after your changes.

Screenshot 2025-06-21 105830


cartography/intel/aws/codebuild.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def transform_codebuild_projects(
+    projects: List[Dict[str, Any]], region: str
+) -> List[Dict[str, Any]]:
+    """
+    Transform CodeBuild project data for ingestion into Neo4j.
+
+    - Only include environment variables of type 'PLAINTEXT'.
+    - Other types (e.g., 'PARAMETER_STORE', 'SECRETS_MANAGER') are skipped to avoid leaking secrets.
```
jychp: Maybe instead of skipping the variables we can add them but with a `name=<REDACTED>` template

cartography/models/aws/codebuild/project.py
```diff
@@ -0,0 +1,46 @@
...
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class CodeBuildProjectNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("arn")
+    arn: PropertyRef = PropertyRef("arn", extra_index=True)
+    created: PropertyRef = PropertyRef("created")
+    environment_variables: PropertyRef = PropertyRef("environmentVariables")
```
jychp, line 19: Is the region field missing ?

cartography/intel/aws/codebuild.py
```diff
@@ -0,0 +1,135 @@
...
+    paginator = client.get_paginator("list_projects")
+    project_names = []
+    for page in paginator.paginate():
+        project_names.extend(page.get("projects", []))
+    return project_names
+
+
+@timeit
+@aws_handle_regions
+def get_codebuild_projects(
```
jychp: You should merge these two functions and retrieve the project details directly inside the `for page in paginator.paginate()` loop.

With the current logic, if there are a lot of projects, you end up trying to fetch details for all of them at once, which can lead to timeouts or other errors.

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,191 @@
...
+def transform_efs_file_systems(
+    fileSystems: List[Dict[str, Any]], region: str
+) -> List[Dict[str, Any]]:
+    """
+    Transform SNS topic data for ingestion
+    """
+    transformed_file_systems = []
+    for file_system in fileSystems:
+        transformed_file_system = {
+            "FileSystemId": file_system["FileSystemId"],
```
jychp: It looks like 90% of the `transformed_file_system` dict is identical to the input `file_system`.

Accessing and copying each key individually significantly increases complexity and computation time.

You should either modify/add only the fields you need in place, or make a shallow copy of the original dict if you really need to avoid mutating it (though I don't think that's necessary in this case).

cartography/intel/aws/efs.py
```diff
@@ -0,0 +1,191 @@
...
+            ),
+        }
+        transformed_file_systems.append(transformed_file_system)
+
+    return transformed_file_systems
+
+
+@timeit
+@aws_handle_regions
+def get_efs_file_system_ids(
```
jychp: I don't really understand the purpose of this function.

It looks like it's just extracting a list of IDs — so why are you passing the `boto_context` and `region`?

This could be simplified directly at the call site with something like:
`xxx = [fs['FileSystemId'] for fs in fileSystems]`

cartography/models/aws/efs/file_system.py
```diff
@@ -0,0 +1,59 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class EfsFileSystemNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("FileSystemId")
+    arn: PropertyRef = PropertyRef("FileSystemArn", extra_index=True)
```
jychp: is the region field missing ?

cartography/models/aws/efs/mount_target.py
```diff
@@ -0,0 +1,78 @@
...
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class EfsMountTargetNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("MountTargetId")
+    arn: PropertyRef = PropertyRef("MountTargetId", extra_index=True)
+    fileSystem_id: PropertyRef = PropertyRef("FileSystemId")
```
jychp: is the region field missing ?

cartography/models/aws/efs/mount_target.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@dataclass(frozen=True)
+class EfsMountTargetToEfsFileSystemRel(CartographyRelSchema):
+    target_node_label: str = "EfsFileSystem"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("FileSystemId")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "ATTACHED_TO"
```
jychp: Personally, I prefer an inward relationship with the `MOUNTS` label, but that's definitely opinionated.

jychp (changes requested):
Thanks for your contribution, some minor changes and we will be able to merge it.

Did you try the module with real data ?

jychp:
@d-aggarwal by the way you could rename this PR and close https://github.com/cartography-cncf/cartography/pull/1628/files as you seems to have include the changes in that one ;)

cartography/intel/aws/efs.py
d-aggarwal: Hi, thanks for the review! I explicitly accessed each key because the input file_system contains extra fields that we don’t want in the transformed output. Let me know if you'd prefer I use a shallow copy and remove the unwanted keys instead.

jychp: This was more a suggestion than a real requirement, if it make sense for you to do this manual mapping, just keep it :)

d-aggarwal:
I’ve addressed the updates, and the EFS fixes are pushed in [https://github.com/cartography-cncf/cartography/pull/1628/files](https://github.com/cartography-cncf/cartography/pull/1628/files) to better separate out the resources.
This one will be rebased once the EFS PR is merged.

d-aggarwal:
I have added region in codebuild doc schema as well,
also rebased this PR for clean merge.

tests/data/aws/codebuild.py
```diff
@@ -0,0 +1,71 @@
+GET_PROJECTS = [
+    {
+        "name": "frontend-build",
+        "arn": "arn:aws:codebuild:eu-west-1:123456789012:project/frontend-build",
+        "description": "Builds front-end assets",
+        "source": {
+            "type": "GITHUB",
+            "location": "https://github.com/example/frontend",
```
jychp, line 8: Could be interesting to add the source / source_type into the model.
With a analysis job you will be able to join CoreBuild node to GitHubRepository

jychp (changes requested):
You should add these property, that will allow interesting analysis jobs.

wokaokf:
My name is xingaonai. The article you wrote is very interesting and taught me a lot. Thank you!

d-aggarwal:
> You should add these property, that will allow interesting analysis jobs.

Updated :)

docs/root/modules/aws/schema.md
```diff
@@ -799,6 +799,26 @@ Representation of an AWS [CloudWatch Log Group](https://docs.aws.amazon.com/Amaz
...
+
+| Field | Description |
+|-------|-------------|
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| id | The ARN of the CodeBuild Project |
+| arn | The Amazon Resource Name (ARN) of the CodeBuild Project |
+| region | The region of the codebuild project |
+| created | The creation time of the CodeBuild Project |
+| environment_variables | A set of environment variables to make available to builds for this build project |
```
achantavy: Add in more detail here since the approach is that we take the variables and construct a string out of it. Include 1 or 2 examples.

achantavy (approved):
Left a comment on making the documentation a bit more thorough since the code makes an opinonated stance on env vars. No blockers though, merge when ready. Thank you!

jychp (approved):

Issue #1650, opened 2025-06-24
Support for Microsoft Entra Applications as Nodes


achantavy:
### **Feature request template**

**Title**: Support for Microsoft Entra Applications as Nodes

**Description**:  
> We should add support for Microsoft Entra applications to be ingested, represented, and queried as  nodes in the graph.  
>  
**Relevant Links**:  
> - [Microsoft Docs: Application entity](https://learn.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0)  
> - [Microsoft Graph: List applications](https://learn.microsoft.com/en-us/graph/api/application-list?view=graph-rest-1.0&tabs=http)

achantavy:
Implemented in https://github.com/cartography-cncf/cartography/pull/1652

achantavy:
Implemented in https://github.com/cartography-cncf/cartography/pull/1652

Issue #1589, opened 2025-05-27
Support relationship property mapping from iterable dicts in one-to-many scenarios


jychp:
### Problem

The recent addition of the `one_to_many` flag in relationships has improved the usability of the data model for request ingestion. However, it doesn't fully address a specific use case:

**Connecting an entity A to multiple entities B *while also setting a property on the relationship***.

#### Example

In a raw request, PagerDuty creates a relationship like:

```
(u:PagerDutyUser)-[r:MEMBER_OF]->(t:PagerDutyTeam)
```

At the same time, it assigns a value to `r.role` using a variable.

Currently, this behavior isn't possible with the data model.

---

### Limitation

Right now, the target of a `PropertyRef` with `one_to_many=True` must be a list of strings or ints. But what we need is support for a list of **dicts**.

#### Desired Use Case

```python
@dataclass(frozen=True)
class PagerDutyTeamToUserProperties(CartographyRelProperties):
    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
    role: PropertyRef = PropertyRef("member.role")

@dataclass(frozen=True)
# (:PagerDutyUser)-[:MEMBER_OF]->(:PagerDutyTeam)
class PagerDutyTeamToUserRel(CartographyRelSchema):
    target_node_label: str = "PagerDutyUser"
    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
        {
            "id": PropertyRef(
                "member.id", 
                one_to_many=True, 
                iterate_dict=("members", "member")
            )
        }
    )
    direction: LinkDirection = LinkDirection.INWARD
    rel_label: str = "MEMBER_OF"
    properties: PagerDutyTeamToUserProperties = PagerDutyTeamToUserProperties()
```

The syntax probably needs refinement, but the core idea is to introduce a flag or attribute that tells the data model:

> “For this relationship, iterate over a list of dicts, and expose each dict element as a variable (e.g., `member`).”

This would give us the flexibility needed to map both nodes and relationship properties in one pass.

---

### Technical Considerations

Supporting this feature would likely require handling node matchers differently depending on whether a `PropertyRef` uses this option:

* **Without the option:** Regular `MATCH` clause.
* **With the option:** The `MATCH` clause needs to be nested within an `UNWIND`.

jychp:
cc @achantavy

jychp:
Linked to https://github.com/cartography-cncf/cartography/pull/1606

[...]

PR #1652, opened 2025-06-25
feat: Support for Microsoft Entra Applications as Nodes (Issue #1650)

shyammukund:
This PR...
Adds support for Entra Applications
Implements both user-app and group-app relationships
Comprehensive testing with unit + integration tests
Updated schema documentation

### Related issues or links

### Checklist
 **Passing Unit Tests and Integration Tests**
image

**Cartography Log**
image

 **Screenshot of Graph showing an Applications relationships with its Users and Groups**
image


jychp:
@shyammukund I deleted your screenshot because your secret was leaking.

Please consider revoking this secret ASAP.

cartography/models/entra/application.py
```diff
@@ -0,0 +1,150 @@
...
+@dataclass(frozen=True)
+class EntraAppRoleAssignmentToUserRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class EntraAppRoleAssignmentToUserRel(CartographyRelSchema):
+    target_node_label: str = "EntraUser"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"display_name": PropertyRef("principal_display_name")},
```
jychp: The display name can be modified by users and doesn't seem reliable enough to be used for a join. Could we use something else instead? Maybe the principal name?

Also, none of these fields are indexed, so the query might be very slow. We should add the `extra_index=True` flag to the selected property on both the `User` and `Group` nodes.

cartography/models/entra/application.py
```diff
@@ -0,0 +1,150 @@
...
+@dataclass(frozen=True)
+class EntraAppRoleAssignmentToGroupRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class EntraAppRoleAssignmentToGroupRel(CartographyRelSchema):
+    target_node_label: str = "EntraGroup"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"display_name": PropertyRef("principal_display_name")},
```
jychp: Same comment as for the `User` relationship.

cartography/models/entra/application.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "HAS_APP_ROLE"
```
jychp: For a first iteration, this is sufficient, but I believe they can have multiple roles.

Feel free to comment on [https://github.com/cartography-cncf/cartography/issues/1589](https://github.com/cartography-cncf/cartography/issues/1589) to help us find the best approach.

cartography/models/entra/application.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: EntraApplicationToTenantRelProperties = (
+        EntraApplicationToTenantRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class EntraApplicationSchema(CartographyNodeSchema):
```
jychp, line 42: I think you should split this file into two, one for each `NodeSchema`. I didn’t realize there were two nodes on my first quick read-through.

cartography/intel/entra/applications.py
```diff
@@ -0,0 +1,331 @@
...
+
+    :param client: GraphServiceClient
+    :return: List of Entra application data
+    """
+    applications = []
+
+    # Get all applications with pagination
+    request_configuration = client.applications.ApplicationsRequestBuilderGetRequestConfiguration(
+        query_parameters=client.applications.ApplicationsRequestBuilderGetQueryParameters(
+            top=999
```
jychp: We should make this configurable at the very least through a `GLOBAL` in this file.

We should also log a warning when this limit is reached to inform the user that the data will be incomplete.

cartography/intel/entra/applications.py
```diff
@@ -0,0 +1,331 @@
...
+                                    if assignment.app_role_id
+                                    else None
+                                ),
+                                "created_date_time": assignment.created_date_time,
+                                "principal_display_name": assignment.principal_display_name,
+                                "principal_type": assignment.principal_type,
+                                "resource_display_name": assignment.resource_display_name,
+                            }
+                            assignments.append(assignment_data)
+                except Exception as e:
```
jychp: Using a broad exception is not a good practice you should narrow down the exception you're catching.

tests/integration/cartography/intel/entra/test_applications.py
```diff
@@ -0,0 +1,163 @@
...
+    return_value=MOCK_ENTRA_APPLICATIONS_DICT,
+)
+@pytest.mark.asyncio
+async def test_sync_entra_applications(mock_get, mock_get_assignments, neo4j_session):
+    """
+    Ensure that applications actually get loaded and connected to tenant,
+    and both user-app and group-app relationships exist
+    """
+    # Setup - Clean up any existing applications and relationships
+    neo4j_session.run(
```
jychp: Cleaning up the graph is done automaticaly (see: [https://github.com/cartography-cncf/cartography/blob/master/tests/integration/conftest.py](https://github.com/cartography-cncf/cartography/blob/master/tests/integration/conftest.py)).

tests/integration/cartography/intel/entra/test_applications.py
```diff
@@ -0,0 +1,163 @@
...
+        MATCH (app:EntraApplication) DETACH DELETE app
+        """
+    )
+    neo4j_session.run(
+        """
+        MATCH (assignment:EntraAppRoleAssignment) DETACH DELETE assignment
+        """
+    )
+
+    neo4j_session.run(
```
jychp: That's OK, but I prefer using an `_ensure_local_neo4j_has_test_users` method defined in `test_users.py`.
As the module grows, this helps avoid a lot of code duplication and makes it easier to update test data if needed.

Example: [https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/tailscale/test\_users.py](https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/tailscale/test_users.py)

tests/unit/cartography/intel/entra/test_applications.py
jychp, line 1: I didn’t review this file—I'm not sure these tests are really necessary since everything is already covered by the integration tests and the code doesn’t implement complex logic.

(I know this isn’t the most conventional point of view, but it’s mine 🙂)

cartography/intel/entra/applications.py
```diff
@@ -0,0 +1,331 @@
...
+
+    :param neo4j_session: Neo4j session
+    :param tenant_id: Entra tenant ID
+    :param client_id: Azure application client ID
+    :param client_secret: Azure application client secret
+    :param update_tag: Update tag for tracking data freshness
+    :param common_job_parameters: Common job parameters containing UPDATE_TAG and TENANT_ID
+    """
+    # Create credentials and client
+    credential = ClientSecretCredential(
```
jychp, line 336: Improvement (non-blocking):

Instead of making a new authentication request in each submodule, maybe we could handle it in the `__init__.py` and pass the authenticated client as an attribute.
This would require some refactoring and test adjustments.

Maybe not something to address in this PR, but worth keeping in mind for the next ones.

cartography/intel/entra/applications.py
```diff
@@ -0,0 +1,331 @@
...
+        query_parameters=client.applications.ApplicationsRequestBuilderGetQueryParameters(
+            top=999
+        )
+    )
+    page = await client.applications.get(request_configuration=request_configuration)
+
+    while page:
+        if page.value:
+            for app in page.value:
+                app_data = {
```
jychp: This feels a bit redundant with the `transform` function.

You should return a list of objects and handle this logic within the `transform` function instead.

cartography/intel/entra/applications.py
```diff
@@ -0,0 +1,331 @@
...
+    """
+    Gets app role assignments for both users and groups using the Microsoft Graph API.
+
+    :param client: GraphServiceClient
+    :return: List of app role assignment data
+    """
+    assignments = []
+
+    # Get all users and their app role assignments
+    users_page = await client.users.get()
```
jychp: Fetching all users just to filter on assignments feels a bit overkill. Can we:

* Fetch users or groups in a single request by providing the role or app?
  **OR**
* Add filtering conditions directly in the request?

Also, maybe this logic should live in the users/groups collection submodule instead, to avoid pulling the same data twice.

jychp (changes requested):
Thanks a lot for your PR.

Overall, the contribution is very solid, please find a few comments and questions below.

shyammukund:
Thanks for the initial feedback @jychp! Will make the changes

cartography/models/entra/application.py
shyammukund: didn't change anything with this yet in the latest commit but am doing a little more digging to see if there is a better approach.

cartography/intel/entra/applications.py
shyammukund: Hey so for this i found that the microsoft sdk has a servicePrinciples endpoint that returns all the assignment types in a single call instead of having to query each user/group individually. With this change do you still think it would be better for this logic to be moved to users/groups because it no longer depends on users or groups to fetch data? Let me know what you think.

cartography/intel/entra/applications.py
```diff
@@ -0,0 +1,367 @@
...
+        query_parameters=client.applications.ApplicationsRequestBuilderGetQueryParameters(
+            top=APPLICATIONS_PAGE_SIZE
+        )
+    )
+    page = await client.applications.get(request_configuration=request_configuration)
+
+    while page:
+        if page.value:
+            # Log a warning if we're hitting the page size limit which might indicate incomplete data
+            if len(page.value) == APPLICATIONS_PAGE_SIZE:
```
kunaals: This feels unnecessary because we actually do pagination here? So even if someone has 2000 applications we will retrieve all of them?

cartography/intel/entra/applications.py
```diff
@@ -0,0 +1,367 @@
...
+                logger.warning(
+                    f"Retrieved {len(page.value)} applications (maximum page size). "
+                    f"There may be more applications that require additional API calls. "
+                    f"Consider adjusting APPLICATIONS_PAGE_SIZE if you expect more applications."
+                )
+
+            # Add raw application objects to the list
```
kunaals:
```suggestion
```
superfluous

cartography/models/entra/app_role_assignment.py
```diff
@@ -0,0 +1,114 @@
...
+from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+# App Role Assignment Schema
```
kunaals:
```suggestion
```

docs/root/modules/entra/schema.md
```diff
@@ -146,3 +152,51 @@ Representation of an Entra [Group](https://learn.microsoft.com/en-us/graph/api/g
     ```cypher
     (:EntraGroup)-[:MEMBER_OF]->(:EntraGroup)
     ```
+
+- Entra groups can have app role assignments to applications
+
+    ```cypher
+    (:EntraGroup)-[:HAS_APP_ROLE]->(:EntraApplication)
+    ```
```
kunaals:
```suggestion
```
duplicative - the relationships is mentioned below

cartography/models/entra/app_role_assignment.py
```diff
@@ -0,0 +1,114 @@
...
+@dataclass(frozen=True)
+class EntraAppRoleAssignmentToApplicationRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class EntraAppRoleAssignmentToApplicationRel(CartographyRelSchema):
+    target_node_label: str = "EntraApplication"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"display_name": PropertyRef("resource_display_name")},
```
kunaals: I think matching on display name here may be a little brittle - would be better to match on like app_id or the like

docs/root/modules/entra/schema.md
```diff
@@ -146,3 +152,51 @@ Representation of an Entra [Group](https://learn.microsoft.com/en-us/graph/api/g
...
+
+### EntraApplication
```
kunaals, line 156: It would be nice to have some example queries in here for common use cases

cartography/models/entra/app_role_assignment.py
kunaals: you might need to add like a `resource_id to `EntraAppRoleAssignmentNodeProperties` to match on

cartography/intel/entra/applications.py
jychp: I think your change make the job.
The pain point was pulling same data multiple times and you fix it :+1:

cartography/models/entra/application.py
achantavy: > The display name can be modified by users and doesn't seem reliable enough to be used for a join.

agree on the displayname not being reliable for a join

achantavy: > Also, none of these fields are indexed, so the query might be very slow. We should add the extra_index=True flag to the selected property on both the User and Group nodes.

If a property shows up in a target node matcher, we create indexes for it: https://github.com/cartography-cncf/cartography/blob/a00a353cca71aaa36df8ddd276bd8c71ca76abcc/cartography/graph/querybuilder.py#L529-L536

Issue #1025, opened 2022-11-09
AWS Inspector sync issue


T0mPh:
*Title*: *AWS Inspector sync issue*

*Description*:
> An unhandled exception is raised when cartography attempts to sync AWS Inspector in region ap-southeast-3

*To Reproduce*:
> I simply ran the following command
`cartography --neo4j-uri bolt://localhost:7687 --neo4j-password-prompt --neo4j-user neo4j`

*Logs*:
<pre>
INFO:cartography.intel.aws.inspector:Syncing AWS Inspector findings for account *redacted* and region ap-southeast-3
ERROR:cartography.sync:Unhandled exception during sync stage 'aws'
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/urllib3/connection.py", line 174, in _new_conn
    conn = connection.create_connection(
  File "/usr/local/lib/python3.10/dist-packages/urllib3/util/connection.py", line 72, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/usr/lib/python3.10/socket.py", line 955, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/botocore/httpsession.py", line 455, in send
    urllib_response = conn.urlopen(
  File "/usr/local/lib/python3.10/dist-packages/urllib3/connectionpool.py", line 787, in urlopen
    retries = retries.increment(
  File "/usr/local/lib/python3.10/dist-packages/urllib3/util/retry.py", line 525, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/usr/local/lib/python3.10/dist-packages/urllib3/packages/six.py", line 770, in reraise
    raise value
  File "/usr/local/lib/python3.10/dist-packages/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.10/dist-packages/urllib3/connectionpool.py", line 386, in _make_request
    self._validate_conn(conn)
  File "/usr/local/lib/python3.10/dist-packages/urllib3/connectionpool.py", line 1042, in _validate_conn
    conn.connect()
  File "/usr/local/lib/python3.10/dist-packages/urllib3/connection.py", line 358, in connect
[...]

Issue #988, opened 2022-09-16
AWS Inspector findings timeout


at4z:
*Title*: *AWS Inspector findings timeout*

*Description*:
> Cartography is crashing when attempting to retrieve Inspector finding from ap-northeast-3

*To Reproduce*:
> cartography --experimental-neo4j-4x-support --neo4j-uri bolt://127.0.0.1:7687  --aws-sync-all-profiles

*Logs*:
> INFO:cartography.intel.aws.inspector:Syncing AWS Inspector findings for account XXX and region ap-northeast-3
INFO:backoff:Backing off inner_function(...) for 1.0s (botocore.exceptions.ClientError: An error occurred (InternalServerErrorException) when calling the ListFindings operation (reached max retries: 4): Endpoint request timed out)
WARNING:cartography.util:Backing off 1.0 seconds after 1 tries. Calling function <function aws_handle_regions.<locals>.inner_function at 0x106b0e710>
INFO:backoff:Backing off inner_function(...) for 0.2s (botocore.exceptions.ClientError: An error occurred (InternalServerErrorException) when calling the ListFindings operation (reached max retries: 4): Endpoint request timed out)
WARNING:cartography.util:Backing off 0.2 seconds after 2 tries. Calling function <function aws_handle_regions.<locals>.inner_function at 0x106b0e710>
INFO:backoff:Backing off inner_function(...) for 0.2s (botocore.exceptions.ClientError: An error occurred (InternalServerErrorException) when calling the ListFindings operation (reached max retries: 4): Endpoint request timed out)
WARNING:cartography.util:Backing off 0.2 seconds after 3 tries. Calling function <function aws_handle_regions.<locals>.inner_function at 0x106b0e710>
INFO:backoff:Backing off inner_function(...) for 6.4s (botocore.exceptions.ClientError: An error occurred (InternalServerErrorException) when calling the ListFindings operation (reached max retries: 4): Endpoint request timed out)
WARNING:cartography.util:Backing off 6.4 seconds after 4 tries. Calling function <function aws_handle_regions.<locals>.inner_function at 0x106b0e710>
ERROR:backoff:Giving up inner_function(...) after 5 tries (botocore.exceptions.ClientError: An error occurred (InternalServerErrorException) when calling the ListFindings operation (reached max retries: 4): Endpoint request timed out)

*Please complete the following information:*:
>  - Cartography 0.65.0

*Additional context*:
[...]

PR #1654, opened 2025-06-26
Refactor AWS Inspector: reduce memory and add parallel ingestion

heryxpc:
### Summary
There is a memory issue with AWS Inspector module. The pagination only returns results until findings are retrieved from one account and all those objects are added to one single list. Considering there can be millions of findings, this can consume all the memory of the job running it.
This PR refactors this module to:
- Provide a maximum number of pages to `aws_paginate`
- Transform `aws_paginate` and `batch` into [generators](https://wiki.python.org/moin/Generators) to get results in batches of iterators.
- Breaks inspector findings ingestion in batches of 1000 elements.
- Adds aysnc processing to ingest multiple accounts in parallel

This processed ~110,000 findings using a max of ~350 MiB

```
INFO:cartography.intel.aws.inspector:Loading 1000 findings from account XXXXXXX
INFO:cartography.intel.aws.inspector:Loading 88 packages
INFO:cartography.intel.aws.inspector:Loading 766 findings from account XXXXXXX
INFO:cartography.intel.aws.inspector:Loading 86 packages

Filename: /Users/eryxp/src/cartography/cartography/intel/aws/inspector.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
   276    289.1 MiB    289.1 MiB           1   @profile
   277                                         def _sync_findings_for_account(
   278                                             neo4j_session: neo4j.Session,
   279                                             boto3_session: boto3.session.Session,
   280                                             region: str,
   281                                             account_id: str,
   282                                             update_tag: int,
   283                                             current_aws_account_id: str,
   284                                         ) -> None:
   285                                             """
   286                                             Syncs the findings for a given account in a given region.
   287                                             """
   288    289.1 MiB      0.0 MiB           1       findings = get_inspector_findings(boto3_session, region, account_id)
   289    289.1 MiB      0.0 MiB           1       print(type(findings))
   290    289.1 MiB      0.0 MiB           1       if not findings:
   291                                                 logger.info(f"No findings to sync for account {account_id} in region {region}")
   292                                                 return
   293    354.9 MiB  -2599.7 MiB         109       for f_batch in findings:
   294    355.0 MiB  -2588.2 MiB         108           finding_data, package_data = transform_inspector_findings(f_batch)
   295    355.0 MiB  -2592.9 MiB         108           logger.info(f"Loading {len(finding_data)} findings from account {account_id}")
   296    355.0 MiB  -5150.9 MiB         216           load_inspector_findings(
   297    355.0 MiB  -2592.9 MiB         108               neo4j_session,
   298    355.0 MiB  -2592.9 MiB         108               finding_data,
   299    355.0 MiB  -2592.9 MiB         108               region,
   300    355.0 MiB  -2592.9 MiB         108               update_tag,
   301    355.0 MiB  -2592.9 MiB         108               current_aws_account_id,
   302                                                 )
   303    355.0 MiB  -2588.1 MiB         108           logger.info(f"Loading {len(package_data)} packages")
   304    355.0 MiB  -5176.1 MiB         216           load_inspector_packages(
   305    355.0 MiB  -2588.1 MiB         108               neo4j_session,
   306    355.0 MiB  -2588.1 MiB         108               package_data,
   307    355.0 MiB  -2588.1 MiB         108               region,
   308    355.0 MiB  -2588.1 MiB         108               update_tag,
   309    355.0 MiB  -2588.1 MiB         108               current_aws_account_id,
   310                                                 )
```

Screenshot 2025-06-25 at 10 31 24 p m

### Related issues or links
https://github.com/cartography-cncf/cartography/issues/1025
https://github.com/cartography-cncf/cartography/issues/988

### Checklist

### DIFF
--- a/cartography/client/core/tx.py
+++ b/cartography/client/core/tx.py
@@ -222,7 +222,7 @@ def load_graph_data(
     :param kwargs: Allows additional keyword args to be supplied to the Neo4j query.
     :return: None
     """
-    for data_batch in list(batch(dict_list, size=10000)):
+    for data_batch in batch(dict_list, size=10000):
         neo4j_session.write_transaction(
             write_list_of_dicts_tx,
             query,
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -182,7 +182,7 @@ def load_ecr_repository_images(
     logger.info(
         f"Loading {len(repo_images_list)} ECR repository images in {region} into graph.",
     )
-    for repo_image_batch in list(batch(repo_images_list, size=10000)):
+    for repo_image_batch in batch(repo_images_list, size=10000):
         neo4j_session.write_transaction(
             _load_ecr_repo_img_tx,
             repo_image_batch,
--- a/cartography/intel/aws/inspector.py
+++ b/cartography/intel/aws/inspector.py
@@ -21,7 +21,7 @@ from cartography.util import to_synchronous
 
 logger = logging.getLogger(__name__)
 
-# As of 7/22/24, Inspector is only available in the below regions. We will need to update this hardcoded list here over
+# As of 7/1/25, Inspector is only available in the below regions. We will need to update this hardcoded list here over
 # time. :\ https://docs.aws.amazon.com/general/latest/gr/inspector2.html
 AWS_INSPECTOR_REGIONS = {
     "us-east-2",
@@ -47,7 +47,11 @@ AWS_INSPECTOR_REGIONS = {
     "eu-central-2",
     "me-south-1",
     "sa-east-1",
+    "us-gov-east-1",
+    "us-gov-west-1",
 }
+
+BATCH_SIZE = 1000
 
 
 @aws_handle_regions
@@ -72,15 +76,14 @@ def get_inspector_findings(
     account_id: str,
 ) -> Iterator[List[Dict[str, Any]]]:
     """
-    Query inspector2.list_findings by filtering the request, otherwise the request could tiemout.
+    Query inspector2.list_findings by filtering the request, otherwise the request could timeout.
     First, we filter by account_id. And since there may be millions of CLOSED findings that may never go away,
     only fetch those in ACTIVE or SUPPRESSED statuses.
     Run the query in batches of 1000 findings and return an iterator to fetch the results.
     """
-    batch_size = 1000
     client = session.client("inspector2", region_name=region)
     logger.info(
-        f"Getting findings in batches of {batch_size} for account {account_id} in region {region}"
+        f"Getting findings in batches of {BATCH_SIZE} for account {account_id} in region {region}"
     )
     aws_args: Dict[str, Any] = {
         "filterCriteria": {
@@ -99,7 +102,7 @@ def get_inspector_findings(
         }
     }
     findings_batches = batch(
-        aws_paginate(client, "list_findings", "findings", None, **aws_args), batch_size
+        aws_paginate(client, "list_findings", "findings", None, **aws_args), BATCH_SIZE
     )
     yield from findings_batches
 
--- a/cartography/intel/aws/resourcegroupstaggingapi.py
+++ b/cartography/intel/aws/resourcegroupstaggingapi.py
@@ -235,7 +235,7 @@ def load_tags(
     if len(tag_data) == 0:
         # If there is no data to load, save some time.
         return
-    for tag_data_batch in list(batch(tag_data, size=100)):
+    for tag_data_batch in batch(tag_data, size=100):
         neo4j_session.write_transaction(
             _load_tags_tx,
             tag_data=tag_data_batch,
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -229,7 +229,7 @@ def aws_paginate(
                 f"""aws_paginate: Key "{object_name}" is not present, check if this is a typo.
 If not, then the AWS datatype somehow does not have this key.""",
             )
-        if max_pages and i >= max_pages:
+        if max_pages is not None and i >= max_pages:
             logger.warning(f"Reached max batch size of {max_pages} pages")
             break
 
@@ -359,7 +359,7 @@ def camel_to_snake(name: str) -> str:
 
 def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> Iterable[List[Any]]:
     """
-    Takes an Iterable of items and returns an Iterator of lists of the same items,
+    Takes an Iterable of items and returns a Generator of lists of the same items,
      batched into chunks of the provided `size`.
 
     Use:
@@ -367,10 +367,7 @@ def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> Iterable[List[Any]
     batch(x, size=3) -> Iterator yielding [1, 2, 3], [4, 5, 6], [7, 8]
     """
     it = iter(items)
-    while True:
-        chunk = list(islice(it, size))
-        if not chunk:
-            return
+    while chunk := list(islice(it, size)):
         yield chunk
 
 
--- a/tests/unit/cartography/test_util.py
+++ b/tests/unit/cartography/test_util.py
@@ -105,7 +105,6 @@ def test_batch(mocker):
     ]
     # Act
     actual = list(batch(x, 5))
-    print(actual)
     # Assert
     assert actual == expected
     # Also check for empty input


cartography/util.py
```diff
@@ -204,30 +204,39 @@ def timed(*args, **kwargs):  # type: ignore
...
         else:
             logger.warning(
                 f"""aws_paginate: Key "{object_name}" is not present, check if this is a typo.
 If not, then the AWS datatype somehow does not have this key.""",
             )
-    return items
+        if i >= max_pages:
+            logger.warning(f"Reached max batch size of {max_pages} pages")
+            break
+    return items, next_token
```
ramonpetgrave64: One of the reasons I wrote this helper function was to automatically handle pagination. I think it might be a good idea to put a limit on the pagination, but there is one potential problem:

when you fetch 1k pages, and then want to fetch the **next** 1k pages, without starting over from the beginning.

I think a better way would be to convert this function to a generator, and we might also need to convert the `batch()` function to a generator, so you could invoke like `batch(aws_paginate(...))`, to fetch all items in all pages, but yield items in batches, and continually fetch pages as new batches of items are needed.

```
L = [
    list(range(5*n, 5*(n+1)))
    for n in range(3)
]

def paginate(max_pages = None):
    for i, p in enumerate(L):
        if max_pages is not None and i >= max_pages:
            return
        for n in p:
            yield n

print(L, '\n')
print(list(paginate()), '\n')
print(list(paginate(1)), '\n')
print(list(paginate(2)), '\n')
print(list(paginate(3)), '\n')
print(list(paginate(5)), '\n')
```

```
[[0, 1, 2, 3, 4], [5, 6, 7, 8, 9], [10, 11, 12, 13, 14]] 

[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14] 

[0, 1, 2, 3, 4] 

[0, 1, 2, 3, 4, 5, 6, 7, 8, 9] 

[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14] 

[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14] 
```

cartography/intel/aws/inspector.py
```diff
@@ -295,4 +318,47 @@ def sync(
...
+                update_tag,
+                current_aws_account_id,
+            )
+
+        to_synchronous(
+            *[
+                async_ingest_findings_for_account(account_id)
+                for account_id in member_accounts
+            ]
+        )
```
ramonpetgrave64, line 349: I'm happy `to_synchronous()` and `to_asynchronous()` are going to good use!

cartography/intel/aws/inspector.py
```diff
@@ -295,4 +318,47 @@ def sync(
...
+    for region in inspector_regions:
+        logger.info(
+            f"Syncing AWS Inspector findings delegated to account {current_aws_account_id} and region {region}",
+        )
+        member_accounts = get_member_accounts(boto3_session, region)
+        # the current host account may not be considered a "member", but we still fetch its findings
+        member_accounts.append(current_aws_account_id)
+
+        async def async_ingest_findings_for_account(account_id: str) -> None:
+            await to_asynchronous(
```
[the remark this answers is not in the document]
ramonpetgrave64: Can't find 4.4.4 docs anymore, but at least v5 docs say the session is thread-safe. Either way, I think it should be fine because we are actually using asyncio Tasks, not multi threading. This means that all these async Tasks are still within the same thread, and there shouldn't be any simultaneous writes.

https://neo4j.com/docs/python-manual/current/connect/?utm_source=GSearch&utm_medium=PaidSearch&utm_campaign=Evergreen&utm_content=AMS-Search-SEMCE-DSA-None-SEM-SEM-NonABM&utm_term=&utm_adgroup=DSA&gad_source=1&gad_campaignid=20973570601&gbraid=0AAAAADk9OYqnRW5r4K2o_8Lc5u7zq845q&gclid=CjwKCAjw3_PCBhA2EiwAkH_j4iqHrZzh9c9_UCl5qEYNJCcAyLfM7aQ0ZHH2E1G0TRgrwKw4PP5fNRoCxVsQAvD_BwE#_connect_to_the_database

ramonpetgrave64 (changes requested):

cartography/util.py
kunaals: +1 to the generator approach

heryxpc: Thank you for the feedback and proposed solution @ramonpetgrave64
Just to wrap my mind around it, the idea is to build `aws_paginate` as a generator and make a generator to call it.
Like
```
def aws_paginate(client, object_name, max_pages, **kwargs) -> list:
    paginator = client.get_paginator()
    for i, page in enumerate(paginator.paginate(**kwargs), start=1):
        if max_pages and i >= max_pages:
            break
        for items in page[object_name]:
            yield items # returns all items from one single page
```
Called by
```
def batch_findings() -> list:
    client = boto3.session.client()
    aws_args = {"arg1": "value1", "arg2": "value2"}
    max_pages = 100
    for batch in enumerate(aws_paginate(client, "findings", max_pages, **aws_args)):
        yield batch # returns one single page of findings
```
And the sync be like
```
def sync_findings():
    for batch in batch_findings():
        findings = transform_findings(batch)
        load_findings(findings)
    cleanup()
```

ramonpetgrave64: Sorry I think my example wasn't complete. I meant to just show how to make a generator. 

My intention was for `aws_paginate()` to yield a single item at a time, not yield all items from one single page.

The original `batch()` function would call `aws_paginate()` to receive those individual items. But instead of `batch()` collecting all those items at once, it will collect a batch of items, yield that batch, and repeat. It's meant to return `List[List]`, so we change it to return `Iterable[List]`.

lmk if you'd like a more complete example.

heryxpc: Hi Ramon, I followed the pattern you described and I believe I have a working solution, would you mind taking a look?

cartography/client/core/tx.py
```diff
@@ -222,7 +222,7 @@ def load_graph_data(
     :param kwargs: Allows additional keyword args to be supplied to the Neo4j query.
     :return: None
     """
-    for data_batch in batch(dict_list, size=10000):
+    for data_batch in list(batch(dict_list, size=10000)):
```
ramonpetgrave64: I don't think you need to convert this back into a list.

cartography/intel/aws/ecr.py
```diff
@@ -182,7 +182,7 @@ def load_ecr_repository_images(
     logger.info(
         f"Loading {len(repo_images_list)} ECR repository images in {region} into graph.",
     )
-    for repo_image_batch in batch(repo_images_list, size=10000):
+    for repo_image_batch in list(batch(repo_images_list, size=10000)):
```
ramonpetgrave64: same: I don't think you need to convert this back into a list.

cartography/util.py
```diff
@@ -208,26 +210,28 @@ def aws_paginate(
...
-            items.extend(page[object_name])
+            items = page[object_name]
+            yield from items
         else:
             logger.warning(
                 f"""aws_paginate: Key "{object_name}" is not present, check if this is a typo.
 If not, then the AWS datatype somehow does not have this key.""",
             )
-    return items
+        if max_pages and i >= max_pages:
```
ramonpetgrave64: non-blocker: 

```suggestion
        if max_pages is not None and i >= max_pages:
```

So that `aws_paginate(..., max_pages=0)` could mean that we want 0 pages, and then move this check earlier in the loop.

cartography/util.py
```diff
@@ -353,17 +357,21 @@ def camel_to_snake(name: str) -> str:
...
+    batch(x, size=3) -> Iterator yielding [1, 2, 3], [4, 5, 6], [7, 8]
     """
-    items = list(items)
-    return [items[i : i + size] for i in range(0, len(items), size)]
+    it = iter(items)
+    while True:
+        chunk = list(islice(it, size))
+        if not chunk:
+            return
+        yield chunk
```
ramonpetgrave64:
```suggestion
    while(chunk := list(islice(it, size))):
        yield chunk
```

tests/unit/cartography/test_util.py
```diff
@@ -104,11 +104,31 @@ def test_batch(mocker):
         [10, 11],
     ]
     # Act
-    actual = batch(x, 5)
+    actual = list(batch(x, 5))
+    print(actual)
```
ramonpetgrave64:
```suggestion
```

cartography/util.py
```diff
@@ -353,17 +357,21 @@ def camel_to_snake(name: str) -> str:
     return re.sub(r"(?<!^)(?=[A-Z])", "_", name).lower()
 
 
-def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> List[List]:
+def batch(items: Iterable, size: int = DEFAULT_BATCH_SIZE) -> Iterable[List[Any]]:
     """
-    Takes an Iterable of items and returns a list of lists of the same items,
+    Takes an Iterable of items and returns an Iterator of lists of the same items,
```
ramonpetgrave64:
```suggestion
    Takes an Iterable of items and returns a Generator of lists of the same items,
```

cartography/intel/aws/resourcegroupstaggingapi.py
```diff
@@ -235,7 +235,7 @@ def load_tags(
     if len(tag_data) == 0:
         # If there is no data to load, save some time.
         return
-    for tag_data_batch in batch(tag_data, size=100):
+    for tag_data_batch in list(batch(tag_data, size=100)):
```
ramonpetgrave64: same: I don't think you need to convert this back into a list.

ramonpetgrave64 (approved):
Thanks for implementing the generators. 

A few small things I would do differently, but it looks great!

cartography/intel/aws/inspector.py
```diff
@@ -46,51 +50,58 @@
...
 def get_inspector_findings(
     session: boto3.session.Session,
     region: str,
-    current_aws_account_id: str,
-) -> List[Dict[str, Any]]:
+    account_id: str,
+) -> Iterator[List[Dict[str, Any]]]:
     """
-    We must list_findings by filtering the request, otherwise the request could tiemout.
+    Query inspector2.list_findings by filtering the request, otherwise the request could tiemout.
```
kunaals:
```suggestion
    Query inspector2.list_findings by filtering the request, otherwise the request could timeout.
```

cartography/intel/aws/inspector.py
```diff
@@ -46,51 +50,58 @@
...
     First, we filter by account_id. And since there may be millions of CLOSED findings that may never go away,
-    we will only fetch those in ACTIVE or SUPPRESSED statuses.
-    list_members will get us all the accounts that
-    have delegated access to the account specified by current_aws_account_id.
+    only fetch those in ACTIVE or SUPPRESSED statuses.
+    Run the query in batches of 1000 findings and return an iterator to fetch the results.
     """
+    batch_size = 1000
```
kunaals: Could we make this a module level constant?

cartography/client/core/tx.py
kunaals: +1, batch should be able to handle generators, let's avoid the unnecessary conversion that might negate the other performance gains

cartography/util.py
heryxpc: The reason I didn't move it up was to avoid the case where `max_pages == i` and allow to return page `i` results before closing the loop. Does that makes sense?

PR #1657, opened 2025-06-28
Add automatic data formatting and sanitization for intel modules

jychp:
## Motivation

This PR addresses three key pain points in developing Cartography intel modules:

1. **Simplify module creation**: Many intel modules currently require extensive `transform` functions to clean and format data from APIs. This automatic sanitization mechanism eliminates much of this boilerplate code, making it transparent and declarative through schema definitions.

2. **Enhanced security**: Sensitive data that shouldn't be stored in the database will never even be sent to Neo4j. The sanitizer automatically filters out any fields not explicitly defined in node schemas, preventing accidental data leakage.

3. **Automatic formatting**: Common data formatting tasks like parsing dates, converting types, and handling empty strings/collections are now handled automatically, reducing manual data transformation code and improving consistency.

## Summary

This PR introduces automatic data formatting and sanitization capabilities for Cartography intel modules, improving data integrity and reducing boilerplate code when ingesting data from external APIs.

### Core Features

1. **Auto-formatting for PropertyRef** (`cartography/models/core/common.py`)
   - Added `auto_format` parameter to `PropertyRef` constructor
   - Supports automatic conversion to: `str`, `int`, `float`, `datetime`, `dict`, `list`
   - Handles edge cases (empty strings/collections become `None`)
   - Graceful error handling with fallback to string representation

2. **Data Sanitization Module** (`cartography/graph/sanitizer.py`)
   - New `data_dict_cleanup()` function for filtering and formatting data
   - Automatically removes fields not defined in node schemas
   - Recursively processes nested dictionaries
   - Applies auto-formatting rules from PropertyRef definitions
   - Extracts all required fields from node schemas including relationships

3. **Comprehensive Test Coverage** (`tests/unit/cartography/graph/test_querydatasanytizer.py`)
   - Unit tests for auto-formatting functionality
   - Tests for data sanitization with complex nested structures
   - Edge case handling validation

4. **Documentation Updates** (`docs/root/dev/writing-intel-modules.md`)
   - Added detailed explanation of auto-formatting feature
   - Usage examples and best practices
   - Data sanitization workflow documentation

## Benefits

- **Reduced boilerplate**: Transform functions can often be eliminated entirely
- **Data security**: Only schema-defined fields are processed, preventing unwanted data from entering the graph
- **Type safety**: Automatic type conversion reduces manual transformation code
- **Consistency**: Standardized handling of empty values and data formatting across modules
- **Error resilience**: Failed conversions don't break syncs, with comprehensive logging

## Example Usage

Before (manual transform required):
```python
def transform_data(raw_data: List[Dict]) -> List[Dict]:
    transformed = []
    for item in raw_data:
        cleaned = {
            'id': item['id'],
            'name': str(item.get('name', '')).strip() or None,
            'port': int(item.get('config', {}).get('port', 0)),
            'created_at': datetime.fromisoformat(item['metadata']['timestamp']),
            # Remove sensitive fields manually...
        }
        transformed.append(cleaned)
    return transformed
```
## Example with Tailscale Device

**Before (unparsed date and empty strings)**
Screenshot From 2025-06-28 14-28-48

**After**
Screenshot From 2025-06-28 14-31-57

### DIFF
--- a/cartography/graph/sanitizer.py
+++ b/cartography/graph/sanitizer.py
@@ -1,0 +1,117 @@
+from typing import Any
+from datetime import datetime
+from dateutil.parser import parse as parse_date
+from dataclasses import asdict
+import logging
+
+from cartography.models.core.common import PropertyRef
+from cartography.models.core.nodes import CartographyNodeSchema
+
+logger = logging.getLogger(__name__)
+
+
+def data_dict_cleanup(node_schema: CartographyNodeSchema, data: dict[str, Any]) -> dict[str, Any]:
+    """ This function cleans up a data dictionary based on the provided node schema.
+
+    This function will remove any keys that are not defined in the node schema,
+    and will also auto-format the values according to the `auto_format` specified in the
+    `PropertyRef` of the node schema.
+
+    :param node_schema: The schema defining the properties and structure of the node.
+    :type node_schema: CartographyNodeSchema
+    :param data: The data dictionary to be cleaned up.
+    :type data: dict[str, Any]
+    :return: A cleaned-up dictionary containing only the properties defined in the node schema,
+             with values auto-formatted according to the schema's `PropertyRef` definitions.
+    :rtype: dict[str, Any]
+    """
+    keys = _node_schema_to_property_refs(node_schema)
+    return _recursive_cleanup(data, keys)
+
+
+def _recursive_cleanup(data: dict, keys: dict[str, PropertyRef]) -> dict:
+    """ This function recursively cleans up a data dictionary based on the provided keys. """
+    cleaned_data = {}
+    sub_dicts_keys = {}
+
+    for key, p_ref in keys.items():
+        if p_ref.set_in_kwargs:
+            continue  # Skip keys that are set in kwargs, not in the data dict
+        # Check if the key is a simple key (not nested)
+        if "." not in key:
+            if key not in data:
+                continue
+            cleaned_data[key] = _auto_format_field(p_ref, data[key])
+        else:
+            # If it's a nested key, split it into prefix and suffix
+            prefix, suffix = key.split(".", 1)
+            if prefix not in data:
+                continue
+            sub_dicts_keys.setdefault(prefix, {})[suffix] = p_ref
+    # Now we can handle the sub-dictionaries
+    for prefix, suffixes in sub_dicts_keys.items():
+        sub_result = _recursive_cleanup(data[prefix], suffixes)
+        if len(sub_result) > 0:
+            cleaned_data[prefix] = sub_result
+    return cleaned_data
+
+
+def _node_schema_to_property_refs(node_schema: CartographyNodeSchema) -> dict[str, PropertyRef]:
+    """ This function extracts property references from a CartographyNodeSchema.
+    It collects all properties defined in the node schema, including those in sub-resource relationships
+    and other relationships, and returns them as a dictionary where the keys are property names
+    and the values are PropertyRef instances."""
+    results: dict[str, PropertyRef] = {}
+
+    for p_ref in asdict(node_schema.properties).values():
+        results[p_ref.name] = p_ref
+
+    if node_schema.sub_resource_relationship:
+        for p_ref in asdict(node_schema.sub_resource_relationship.properties).values():
+            results[p_ref.name] = p_ref
+        for p_ref in asdict(node_schema.sub_resource_relationship.target_node_matcher).values():
+            results[p_ref.name] = p_ref
+
+    if node_schema.other_relationships:
+        for rel in node_schema.other_relationships.rels:
+            for p_ref in asdict(rel.properties).values():
+                results[p_ref.name] = p_ref
+            for p_ref in asdict(rel.target_node_matcher).values():
+                results[p_ref.name] = p_ref
+
+    return results
+
+
+
+def _auto_format_field(p_ref: PropertyRef, value: Any) -> Any:
+    """ This function auto-formats a field based on the PropertyRef's auto_format type. """
+    try:
+        if value is None:
+            return None
+        if p_ref.auto_format is None:
+            return value
+        if p_ref.auto_format == str:
+            value = str(value)
+            if len(value) == 0:
+                return None
+            return value
+        if p_ref.auto_format == int:
+            return int(value)
+        if p_ref.auto_format == float:
+            return float(value)
+        if p_ref.auto_format == dict or p_ref.auto_format == list:
+            if len(value) == 0:
+                return None
+            return value
+        if p_ref.auto_format == datetime:
+            if isinstance(value, datetime):
+                return value
+            if isinstance(value, int) or isinstance(value, float):
+                return datetime.fromtimestamp(value)
+            return parse_date(value)
+    # Handling broad exceptions is generally discouraged, but here we log the error
+    # and return a string representation of the value to avoid breaking the data flow.
+    except Exception as e:
+        logger.warning(f"Error formatting field '{p_ref}': {e}. Value: {value}")
+        # If any error occurs, we fallback to string representation
+        return str(value)
--- a/cartography/graph/sanitizer.py
+++ b/cartography/graph/sanitizer.py
@@ -1,8 +1,9 @@
-from typing import Any
-from datetime import datetime
-from dateutil.parser import parse as parse_date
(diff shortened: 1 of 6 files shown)


tests/unit/cartography/graph/test_querydatasanytizer.py
```diff
@@ -0,0 +1,169 @@
+import math
```
kunaals: Typo in the filename, should be `test_querydatasanitizer`

docs/root/dev/writing-intel-modules.md
```diff
@@ -40,6 +40,11 @@ to make it easier to ingest to the graph. `transform` functions are sometimes om
 
 We have some best practices on handling transforms:
 
+```{hint}
+Before implementing a transform method, check if that can not be done with the `auto_format` option of the PropertyRef.
```
kunaals:
```suggestion
Try using the `auto_format` option of the PropertyRef instead of writing a transform method!
```

kunaals: I want to push people to use this feature :)

cartography/graph/sanitizer.py
```diff
@@ -0,0 +1,135 @@
...
+            if isinstance(value, int) or isinstance(value, float):
+                return datetime.fromtimestamp(value)
+            return parse_date(value)
+        if p_ref.auto_format == bool:
+            if isinstance(value, str):
+                if value.lower() in ("true", "1", "yes"):
+                    return True
+                if value.lower() in ("false", "0", "no"):
+                    return False
+                return value
```
kunaals, line 282: Thoughts on logging a warning here if we can't convert the value to a boolean? All other paths within this if statement return a boolean whereas this path silently returns a string. I see our default is to return a string but I think it may be nice to warn because if a string is being returned there is likely some unexpected output coming from the source API

kunaals (commented):
Looks pretty good to me, I think the multi-nested situation could use some eyes from others, I'm not sure what the best way to handle that is

cartography/graph/sanitizer.py
jychp: Good catch.

kunaals (approved):
LGTM!

cartography/graph/sanitizer.py
```diff
@@ -0,0 +1,293 @@
+"""
```
achantavy, line 1: I think this file should be renamed to 'formatter' instead of 'sanitizer' since its main function is to format data, not necessarily to clean it

cartography/graph/sanitizer.py
```diff
@@ -0,0 +1,293 @@
...
+    if rel_schema.source_node_matcher:
+        for p_ref in asdict(rel_schema.source_node_matcher).values():
+            if p_ref.set_in_kwargs:
+                continue
+            results[p_ref.name] = p_ref
+
+    return results
+
+
+def _auto_format_field(p_ref: PropertyRef, value: Any) -> Any:
```
achantavy, line 203: Is there a way to be stricter than `Any` with the typehints here? could do `type1 | type2 | typeN`. It's not correct right now that it accepts anything

docs/root/dev/writing-intel-modules.md
```diff
@@ -339,6 +344,52 @@ Here's how to represent this in the Cartography data model:
 
 Now we can use the same steps described above in this doc to finish data ingestion.
 
+#### Auto format
```
achantavy, line 347: Update agents.md too

cartography/models/tailscale/device.py
```diff
@@ -16,31 +17,35 @@ class TailscaleDeviceNodeProperties(CartographyNodeProperties):
...
     id: PropertyRef = PropertyRef("nodeId")
     lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
-    name: PropertyRef = PropertyRef("name")
-    hostname: PropertyRef = PropertyRef("hostname")
-    client_version: PropertyRef = PropertyRef("clientVersion")
-    update_available: PropertyRef = PropertyRef("updateAvailable")
-    os: PropertyRef = PropertyRef("os")
-    created: PropertyRef = PropertyRef("created")
-    last_seen: PropertyRef = PropertyRef("lastSeen")
+    name: PropertyRef = PropertyRef("name", auto_format=str)
```
achantavy, line 20: What's the intended behavior if some fields are auto_format and others are not?

cartography/graph/sanitizer.py
```diff
@@ -0,0 +1,293 @@
...
+from dateutil.parser import parse as parse_date
+
+from cartography.models.core.common import PropertyRef
+from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.relationships import CartographyRelSchema
+
+logger = logging.getLogger(__name__)
+
+
+def data_dict_cleanup(
```
achantavy, line 30: cleanup is used in the project to describe keeping things fresh

this should probably be 'format' related or normalization related

achantavy (commented):
Left some tactical comments related to 'if we decide to adopt this, then these will need to be addressed'.

Zooming out though, I'm somewhat concerned about how generalizable this is.

Many of the data returned from APIs will be in ugly-looking shapes that require a lot more work than the AWS APIs and similar (even some AWS APIs require a significant transform step to get them to be lists of dicts like how cartography likes).

Writing modules is already pretty hard, and I don't want to overwhelm a new module author with choices. For example, I wouldn't want them to start off with auto_format and then get stuck because it can't handle a weird nested structure, and then give up. I want to avoid situations where we're confused about what is happening at the transform step and at an autoformat step.

The above may be a weak specific example, and yeah the transforms add a lot of code, but one advantage of keeping the transforms as they are is that it's straightforward to understand that data from the API is this shape, data to the graph should be another, and I should do something to adjust it.

achantavy:
Additionally we could also add a link to this to explain python types as handled by the neo4j driver https://neo4j.com/docs/python-manual/current/data-types/

jychp:
@achantavy I see, sorry for not being clear earlier. This feature is meant to replace the transform stage **only when** transform is just field selection/formatting, but it can also be used alongside it.

The formatting is applied to the **final dict** passed to `load()`, so it’s fully compatible with any transform that happens before. Supported flows:

* transform only
* transform + auto-format
* auto-format only

Note: auto-format can target only a subset of fields.

Think of this feature as:

* a sanitizer that filters the dict sent to Neo4j by removing data not referenced by a `PropertyRef`
* a best-effort normalizer to coerce values into the expected formats

The goal isn’t to replace `transform()`, but to avoid making it mandatory when the transform step is only trivial formatting (e.g., stripping empty strings, date normalization, etc.).

kunaals:
I've been thinking about this more - it could be interesting for us to start using dataclasses as a mixin for our node schemas and then doing transforms using the @dataclass_transform decorator. https://medium.com/@23saini/simplify-data-transformations-with-the-dataclass-transform-decorator-13803d71f68a

Cartography mostly passes around dicts, but this isn't great because they lack type safety and fields can easily be missing. Using dataclasses throughout could make writing intelmodules simpler and give us strong typing from the get to the load

as an example, Shyam's recent PR where the transforms are simple: https://github.com/cartography-cncf/cartography/pull/1825/files?diff=split#diff-55692e7174b08993665d58564b659fbe8d4539ef593304a856cf0651635776a1R160

One could start building an intelmodule by defining the NodeSchema with a dataclass mixin (not actually sure how feasible this is) 

(aside: dataclasses do seem to look very similar to NodeSchemas, example below)
```
@dataclass
class KubernetesUser:
    id: str
    name: str
    cluster_name: str
    aws_role_arn: str
```

Instead of a transform function, you'd have transforms within the KubernetesUser class itself, example:
```
@dataclass_transform
@dataclass
class KubernetesUser(NodeSchema):
    name: str
    age: int
    email: str

    def transform_email(self):
        self.email = self.email.lower()
```
You'd define this object during a get and initialize everything as typed objects from the get-go
`kube_user = KubernetesUser(f"{cluster_name}/{username}", username, cluster_name, role_arn)`

You can easily output a dataclass back to JSON when needed, ex:
`kube_user_dict = asdict(kube_user)`

Just spitballing here, not sure if this is even feasible. I think it would give us type safety and make sanitization easier while making intelmodules even more approachable

Edit: this would also ensure that we're doing transforms in place which could potentially cut memory consumption in half

Issue #1660, opened 2025-06-30
Feature Request: Support for SentinelOne CVE data


serge-wq:
*Title*: *Support for SentinelOne CVE data*

*Description*:
Ingest CVEs detected on endpoints by SentinelOne agents.

Proposed initial schema:

```
(:S1Account)-[:RISK]->(:S1CVE)
(:S1Account)-[:RESOURCE]->(:S1ApplicationVersion)
(:S1Account)-[:RESOURCE]->(:S1Application)
(:S1Account)-[:RESOURCE]->(:S1Agent)
---
(:S1CVE)-[:AFFECTS]->(:S1ApplicationVersion)
(:S1Application)-[:VERSION]->(:S1ApplicationVersion)
(:S1Agent)-[:HAS_INSTALLED]->(:S1ApplicationVersion)
```

Questions we are going to be able to answer:
- What CVEs are detected on which endpoints?
- What applications versions are affected by which CVEs?
- What endpoints are affected by which CVEs on which applications?

PR #1661, opened 2025-06-30
Ingest SentinelOne account nodes 

serge-wq:
### Summary

First PR for the SentinelOne intelmodule. This PR covers the ingestion of the S1 account nodes which will serve as root nodes for the resources of the assets and CVEs. It is possible to provide a set of account IDs to ingest otherwise the module will ingest every account accessible to the user that owns the API token. 

I'll update the intelmodule and the corresponding schema docs once all the intended resources are implemented. 

Logs from running this locally:

Screenshot 2025-07-01 at 9 55 23 a m

API token was placed in the env variable `SENTINELONE_API_TOKEN`

### Related issues or links

### Checklist

### DIFF
--- a/cartography/intel/sentinelone/account.py
+++ b/cartography/intel/sentinelone/account.py
@@ -4,7 +4,6 @@ from typing import Any
 import neo4j
 
 from cartography.client.core.tx import load
-from cartography.graph.job import GraphJob
 from cartography.intel.sentinelone.utils import call_sentinelone_api
 from cartography.models.sentinelone.account import S1AccountSchema
 from cartography.util import timeit
@@ -103,22 +102,6 @@ def load_accounts(
     logger.info(f"Loaded {len(accounts_data)} SentinelOne account nodes")
 
 
-def cleanup_accounts(
-    neo4j_session: neo4j.Session, common_job_parameters: dict[str, Any]
-) -> None:
-    """
-    Remove stale SentinelOne account data
-    :param neo4j_session: Neo4j session
-    :param common_job_parameters: Job parameters for cleanup
-    """
-    logger.debug("Running SentinelOne account cleanup job")
-
-    # Use GraphJob to clean up stale account data
-    GraphJob.from_node_schema(S1AccountSchema(), common_job_parameters).run(
-        neo4j_session
-    )
-
-
 @timeit
 def sync_accounts(
     neo4j_session: neo4j.Session,
@@ -146,9 +129,6 @@ def sync_accounts(
 
     # 3. LOAD - Ingest to Neo4j using data model
     load_accounts(neo4j_session, transformed_accounts, update_tag)
-
-    # 4. CLEANUP - Remove stale data
-    cleanup_accounts(neo4j_session, common_job_parameters)
 
     synced_account_ids = [account["id"] for account in transformed_accounts]
     logger.info(f"Synced {len(synced_account_ids)} SentinelOne accounts")
--- a/cartography/sync.py
+++ b/cartography/sync.py
@@ -81,8 +81,9 @@ TOP_LEVEL_MODULES = OrderedDict(
         "jamf": cartography.intel.jamf.start_jamf_ingestion,
         "pagerduty": cartography.intel.pagerduty.start_pagerduty_ingestion,
         "trivy": cartography.intel.trivy.start_trivy_ingestion,
-        "analysis": cartography.intel.analysis.run,
         "sentinelone": cartography.intel.sentinelone.start_sentinelone_ingestion,
+        # Analysis should be the last stage
+        "analysis": cartography.intel.analysis.run,
     }
 )
 


cartography/intel/sentinelone/account.py
```diff
@@ -0,0 +1,145 @@
...
+    :param neo4j_session: Neo4j session
+    :param accounts_data: List of account data to process
+    :param update_tag: Update tag
+    """
+    if not accounts_data:
+        logger.warning("No account data provided to ensure_account_nodes")
+        return
+
+    query = """
+    UNWIND $accounts as account
```
achantavy: Use the data model :D

Copy the lastpass module as an example https://github.com/cartography-cncf/cartography/blob/1924f6b51cb1cb865f6c0751510758139b379134/cartography/intel/lastpass/users.py#L67

achantavy (changes requested):
Thanks for the PR, exciting stuff! Left a comment on using the data model.

Also don't forget to update README.md and schema.md.

If you're using an agent, `@` it to the AGENTS.md file so it can see the instructions

cartography/intel/sentinelone/account.py
```diff
@@ -0,0 +1,145 @@
...
+    SET a.name = account.name,
+        a.account_type = account.account_type,
+        a.active_agents = account.active_agents,
+        a.created_at = account.created_at,
+        a.expiration = account.expiration,
+        a.number_of_sites = account.number_of_sites,
+        a.state = account.state,
+        a.lastupdated = $update_tag
+    """
+    neo4j_session.run(
```
heryxpc: Did you try using the [load](https://github.com/cartography-cncf/cartography/blob/master/cartography/client/core/tx.py#L258) function instead of a the Cypher query?

cartography/intel/sentinelone/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+    :param config: A cartography.config object
+    :return: None
+    """
+    if not config.sentinelone_api_token or not config.sentinelone_api_url:
+        logger.info("SentinelOne API configuration not found - skipping this module.")
+        return
+
+    common_job_parameters = {
+        "UPDATE_TAG": config.update_tag,
+    }
```
heryxpc, line 29: `common_job_parameters` is never used.

cartography/intel/sentinelone/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+    # Sync SentinelOne account data (needs to be done first to establish the account node)
+    account_id = sync_accounts(
+        neo4j_session,
+        config.sentinelone_api_url,
+        config.sentinelone_api_token,
+        config.update_tag,
+    )
+
+    # Add account_id to common job parameters
+    common_job_parameters["ACCOUNT_ID"] = account_id
```
heryxpc: Same as above, it's not used.

cartography/intel/sentinelone/account.py
```diff
@@ -0,0 +1,145 @@
...
+            "number_of_sites": account.get("numberOfSites"),
+            "state": account.get("state", ""),
+        }
+        transformed_accounts_data.append(transformed_account)
+
+    return transformed_accounts_data
+
+
+@timeit
+def _fetch_accounts(
```
heryxpc: For naming consistency with other modules, I would use `_get_accounts` instead.

cartography/intel/sentinelone/account.py
```diff
@@ -0,0 +1,145 @@
...
+        endpoint="web/api/v2.1/accounts",
+        api_token=api_token,
+    )
+
+    accounts_data = response.get("data", [])
+
+    # Filter accounts by ID if specified
+    if account_ids:
+        accounts_data = [
+            account for account in accounts_data if account.get("id") in account_ids
```
heryxpc, line 39: This will throw an error if `account_ids` is not passed.

cartography/intel/sentinelone/account.py
serge-wq: Done :)

cartography/intel/sentinelone/account.py
serge-wq: Added data model.

cartography/intel/sentinelone/__init__.py
serge-wq: Added its use for the sync function.

cartography/sync.py
```diff
@@ -81,6 +82,7 @@
         "pagerduty": cartography.intel.pagerduty.start_pagerduty_ingestion,
         "trivy": cartography.intel.trivy.start_trivy_ingestion,
         "analysis": cartography.intel.analysis.run,
+        "sentinelone": cartography.intel.sentinelone.start_sentinelone_ingestion,
```
achantavy, line 84: Let analysis be the last stage.

Actually can you do me a big favor and add in a comment here too:

```python
# Analysis should be the last stage
```

heryxpc (approved):

achantavy (approved):
aww yee

PR #1663, opened 2025-06-30
Skip source merge option

kunaals:
### Summary
MATCH instead of MERGE source nodes when creating relationships via `RelSchema`

Unblocks #1662

### DIFF
--- a/cartography/graph/querybuilder.py
+++ b/cartography/graph/querybuilder.py
@@ -426,6 +426,7 @@ def filter_selected_relationships(
 def build_ingestion_query(
     node_schema: CartographyNodeSchema,
     selected_relationships: Optional[Set[CartographyRelSchema]] = None,
+    skip_source_merge: bool = False,
 ) -> str:
     """
     Generates a Neo4j query from the given CartographyNodeSchema to ingest the specified nodes and relationships so that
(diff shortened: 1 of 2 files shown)


jychp:
Great !

This issue bring a solution to https://github.com/cartography-cncf/cartography/discussions/1499
And bring a workaround for https://github.com/cartography-cncf/cartography/issues/1589 (that issues is blocking for at least 2 modules refactoring)

@heryxpc I think this PR will solve some of your problems

cartography/graph/querybuilder.py
```diff
@@ -446,16 +450,29 @@ def build_ingestion_query(
...
-                $set_node_properties_statement
-            $attach_relationships_statement
-        """,
-    )
+    if skip_source_merge:
+        # Only attach relationships to existing nodes – don't create new ones.
+        query_template = Template(
+            """
+            UNWIND $DictList AS item
+                MATCH (i:$node_label{id: $dict_id_field})
```
jychp, line 458: Nitpicking:
We could have a template with a $match_statement variable to avoid duplication of the global structure of the query.

jychp (approved):
LGTM

kunaals:
Ah I think we're going to supersede this with a slightly different approach where you specify the source_node type on the RelSchema @achantavy

achantavy:
closing out in favor of another approach

PR #1664, opened 2025-07-01
Add department and manager ingestion for Entra users

kunaals:
## Summary
- fetch department and job title using `$select` in Entra user API
- pull manager IDs via new `get_user_manager_id` helper
- handle manager lookup when transforming users
- update integration tests for manager IDs

## Testing
- `make test_unit`
- `make test_integration` *(fails: couldn't connect to localhost:7687)*

------
https://chatgpt.com/codex/tasks/task_b_685c656af1788323a1465934c91806fa


cartography/intel/entra/users.py
```diff
@@ -49,7 +97,18 @@ async def get_users(client: GraphServiceClient) -> list[User]:
 
 
 @timeit
-def transform_users(users: list[User]) -> list[dict[str, Any]]:
+async def get_user_manager_id(client: GraphServiceClient, user_id: str) -> str | None:
```
jychp: Not sure what you're trying to achieve with the manager logic.

Based on your test data, it looks like you're already retrieving the manager ID when fetching the user:  
`manager=DirectoryObject(id="11dca63b-cb03-4e53-bb75-fa8060285550")` — which makes sense.

Since you're already ingesting all users, the manager object will be part of the dataset. So why issue an additional API request to retrieve the manager details? Especially since you only end up storing the manager ID.

This extra call seems unnecessary. I don’t think you need any additional query here.

jychp (changes requested):
I think we can simplify and optimize the flow on managers.

cartography/intel/entra/users.py
kunaals: good call out, changed this to pull data from the original call via the SDK

jychp: Just delete this unused function and we can merge it :)

jychp (changes requested):
LGTM
Just some dead code to remove.

cartography/intel/entra/users.py
kunaals: Thanks for catching this! I also just realized that I forgot to update the EntraUser schema documentation - will push now

Issue #1348, opened 2024-08-28
[FEATURE REQUEST] Ingest Github dependency information using GraphQL API


heryxpc:
*Title*: *Ingest dependencies information as `Dependency` nodes using GH GraphQL API*

*Description*:
Currently, there is custom code to capture PythonLibrary dependencies https://github.com/lyft/cartography/blob/master/cartography/intel/github/repos.py#L523.
This is not generic (only supports Python setup.cfg and requirements.txt files), not covering other ecosystem's dependencies.

If the dependency graph is enabled (see https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/configuring-the-dependency-graph), Github automatically parses the project dependency lock files to generate a comprehensive graph.

This information is available via Github's GraphQL API, like:
```
repository(owner: "lyft", name: "cartography") {
    dependencyGraphManifests(first: 10) {
      nodes {
        dependencies(first: 10) {
          nodes {
            packageName
            requirements
          }
        }
      }
    }
  }
```

The idea is building a generic module that can ingest a node named `Dependency` during Github ingestion to surface (_at least_) direct dependencies.
This would be useful for software composition analysis to identify supply chain risks in any ecosystem supported by Github.

[optional *Relevant Links*:]
https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-supply-chain-security
https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph

achantavy:
This was implemented in #1670!

Issue #13, opened 2019-03-11
Schema: EC2SecurityGroup `id`s might not be unique between accounts


achantavy:
We might want to make the id a concatenation of `{AWS account id} + {EC2 security group ID}`. 

Right now we are only using `{Group ID}`.

sachafaust:
easy fix. The cleanup flow will do the job or cleaning up at the end.

Issue #12, opened 2019-03-10
Sync crash: botocore NoSuchEntity exception


achantavy:
Crash on sync:
```
botocore.errorfactory.NoSuchEntityException: An error occurred (NoSuchEntity) when calling the ListAccessKeys operation: The user with name _REDACTED_ cannot be found.
```

The user existed in the graph at the time it was sync'd last (https://github.com/lyft/cartography/blob/master/cartography/intel/aws/iam.py#L392)

but the user did not exist by the next sync when we tried to pull access key data:
https://github.com/lyft/cartography/blob/master/cartography/intel/aws/iam.py#L395

We might want to add an exception handler here to simply skip over the lost record.

ecdavis:
It looks like this happened due to the user being deleted after users were synced but before access keys were synced.

PR #1670, opened 2025-07-02
feat: Support for ingesting GitHub Dependencies and Manifests using GraphQL API (Issue #1348)

shyammukund:
- Replace Python-specific dependency parsing with generic dependency ingestion
- Use GitHub GraphQL API dependencyGraphManifests field to support all ecosystems
- Added DependencyGraphManifest nodes to GitHub dependency tracking, enabling relationships (GitHubRepository)-[:HAS_MANIFEST]->(DependencyGraphManifest)-[:HAS_DEP]->(Dependency) 
- Added relationship (GitHubRepository)-[:REQUIRES]->(Dependency) relationships.

- https://github.com/cartography-cncf/cartography/issues/1348

### Checklist

-  **Cartography Log**

INFO:cartography.sync:Starting sync stage 'github'
INFO:cartography.intel.github.users:Syncing GitHub users
INFO:cartography.intel.github.users:Retrieving users from GitHub organization subimagesec
INFO:cartography.intel.github.users:Retrieving enterprise owners from GitHub organization subimagesec
INFO:cartography.intel.github.users:Loading 1 GitHub organization to the graph
INFO:cartography.intel.github.users:Loading 4 GitHub users to the graph
INFO:cartography.intel.github.users:Loading 0 GitHub users to the graph
INFO:cartography.intel.github.users:Cleaning up GitHub users
INFO:cartography.graph.statement:Completed GitHubUser statement #1
INFO:cartography.graph.statement:Completed GitHubUser statement #2
INFO:cartography.graph.job:Finished job GitHubUser
INFO:cartography.graph.statement:Completed GitHubUser statement #1
INFO:cartography.graph.job:Finished job GitHubUser
INFO:cartography.intel.github.repos:Syncing GitHub repos
INFO:cartography.intel.github.repos:Retrieving repo collaborators for affiliation "DIRECT" on org "subimagesec".
INFO:cartography.intel.github.repos:Retrieving repo collaborators for affiliation "OUTSIDE" on org "subimagesec".
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #1
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #2
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #3
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #4
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #5
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #6
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #7
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #8
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #9
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #10
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #11
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #12
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #13
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #14
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #15
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #16
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #17
INFO:cartography.graph.statement:Completed github_repos_cleanup statement #18
INFO:cartography.graph.job:Finished job github_repos_cleanup
INFO:cartography.intel.github.teams:Loading 0 GitHub team-repos to the graph
INFO:cartography.graph.statement:Completed GitHubTeam statement #1
INFO:cartography.graph.statement:Completed GitHubTeam statement #2
INFO:cartography.graph.statement:Completed GitHubTeam statement #3
INFO:cartography.graph.statement:Completed GitHubTeam statement #4
INFO:cartography.graph.statement:Completed GitHubTeam statement #5
INFO:cartography.graph.statement:Completed GitHubTeam statement #6
INFO:cartography.graph.statement:Completed GitHubTeam statement #7
INFO:cartography.graph.statement:Completed GitHubTeam statement #8
INFO:cartography.graph.statement:Completed GitHubTeam statement #9
INFO:cartography.graph.statement:Completed GitHubTeam statement #10
INFO:cartography.graph.job:Finished job GitHubTeam
INFO:cartography.sync:Finishing sync stage 'github'

- **Neo4j Graph showing Repo->Dependencies, Manifest->Dependencies and Repo->Manifest**

image


tests/unit/cartography/intel/github/test_repos.py
```diff
@@ -45,3 +53,288 @@ def test_get_team_users_github_returns_none(
...
+    load_github_dependencies(mock_session, update_tag, dependencies)
+
+    # Assert
+    assert mock_session.run.call_count == 1
+    call_args = mock_session.run.call_args
+    assert call_args[1]["Dependencies"] == []
+    assert call_args[1]["UpdateTag"] == update_tag
+
+
+def test_load_github_dependencies_query_structure():
```
achantavy: I would remove this test. We want to test for outcomes affected by this PR and not implementation. Testing for what query was generated is an outcome not controlled by this PR

tests/unit/cartography/intel/github/test_repos.py
```diff
@@ -45,3 +53,288 @@ def test_get_team_users_github_returns_none(
...
+    assert "UNWIND $Dependencies AS dep" in call_args[0][0]
+    assert "MERGE (lib:Dependency{id: dep.id})" in call_args[0][0]
+    assert "MERGE (repo)-[r:REQUIRES]->(lib)" in call_args[0][0]
+
+    # Check the parameters
+    assert call_args[1]["Dependencies"] == dependencies
+    assert call_args[1]["UpdateTag"] == update_tag
+
+
+def test_load_github_dependencies_empty_list():
```
achantavy: Not a necessary test. Empty args passed to the load() function is already tested in the load function's own tests. 

The tests you add for your module should be relevant only for your new functionality, not existing functionality

tests/unit/cartography/intel/github/test_repos.py
```diff
@@ -45,3 +53,288 @@ def test_get_team_users_github_returns_none(
...
+            "package_manager": "PIP",
+            "manifest_path": "/requirements.txt",
+            "repo_url": "https://github.com/test/repo",
+        },
+    ]
+
+    # Act
+    load_github_dependencies(mock_session, update_tag, dependencies)
+
+    # Assert
```
achantavy: Test for the outcomes and not the implementation. Use `check_rels()` and `check_nodes()` to ensure that the nodes are created as expected. Don't test for brittle things like params passed to functions. These aren't really valuable.

We want to know if we run the `sync()` function, will the graph be created as we expect.

tests/unit/cartography/intel/github/test_repos.py
```diff
@@ -45,3 +53,288 @@ def test_get_team_users_github_returns_none(
...
+
+    # Check package with missing blobPath
+    test_dep = next(
+        dep for dep in output_list if dep["original_name"] == "test-package"
+    )
+    assert test_dep["manifest_path"] == ""
+    assert test_dep["ecosystem"] == "unknown"
+
+
+def test_transform_dependency_graph_no_dependencies():
```
achantavy: Not a very useful test for similar reasons as my other below comments, I wouldn't include this test

cartography/intel/github/repos.py
```diff
@@ -785,6 +917,73 @@ def load_collaborators(
...
+) -> None:
+    """
+    Ingest GitHub dependency data into Neo4j
+    :param neo4j_session: Neo4J session object for server communication
+    :param update_tag: Timestamp used to determine data freshness
+    :param dependencies: List of dependency objects from GitHub's dependency graph
+    :return: Nothing
+    """
+    query = """
+    UNWIND $Dependencies AS dep
```
achantavy: need to use data model.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -386,3 +386,121 @@ def test_python_library_in_multiple_requirements_files(neo4j_session):
...
+    Ensure that repositories are connected to dependencies from GitHub's dependency graph.
+    """
+    _ensure_local_neo4j_has_test_data(neo4j_session)
+
+    # Test NPM dependency with exact version
+    query = """
+    MATCH (repo:GitHubRepository{name:'cartography'})-[r:REQUIRES]->(dep:Dependency{name:'react'})
+    RETURN dep.id, dep.version, dep.ecosystem, r.requirements, r.manifest_path
+    """
+    nodes = neo4j_session.run(query)
```
achantavy: Need to use `check_rels()` and `check_nodes()` instead of this pattern. Refer to your previous Entra PRs. Use the new patterns

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -386,3 +386,121 @@ def test_python_library_in_multiple_requirements_files(neo4j_session):
...
+    """
+    _ensure_local_neo4j_has_test_data(neo4j_session)
+
+    # Test NPM dependency with complex version (no pinned version)
+    query = """
+    MATCH (repo:GitHubRepository{name:'cartography'})-[r:REQUIRES]->(dep:Dependency{name:'lodash'})
+    RETURN dep.id, dep.version, dep.ecosystem, r.requirements
+    """
+    nodes = neo4j_session.run(query)
+    actual_nodes = {
```
achantavy: `check_rels()`/`check_nodes()`

cartography/data/jobs/cleanup/github_repos_cleanup.json
```diff
@@ -19,6 +19,11 @@
     "iterative": true,
     "iterationsize": 100
   },
+  {
```
achantavy: Now that Dependencies are a data model, we don't need to have a manual cleanup job anymore. We can just do 

```python
    GraphJob.from_node_schema(GitHubDependencySchema(), common_job_parameters).run(
        neo4j_session,
    )
```

cartography/data/jobs/cleanup/github_repos_cleanup.json
```diff
@@ -39,6 +44,11 @@
...
```
achantavy: Can remove this too

cartography/intel/github/repos.py
```diff
@@ -93,6 +96,18 @@
                             text
                         }
                     }
+                    dependencyGraphManifests(first: 10) {
```
achantavy: I think 10 is a reasonable limit for a single repo, though for a monorepo with many different services in it, 10 might be too few. (for reference, a dependency graph manifest object is created for every requirements.txt/package.json/etc file in a repo) @hanzo @heryxpc what do you think? For large monorepos we might want pagination.

cartography/intel/github/repos.py
```diff
@@ -93,6 +96,18 @@
...
+                        nodes {
+                            blobPath
+                            dependencies(first: 100) {
```
achantavy, line 104: I wonder if this is a safe assumption or if we need to paginate. @hanzo  @heryxpc I imagine you've seen requirements.txt files with 100+ deps?

Double pagination is not possible using the graphql, so we'd have to pick just one list of items to iterate over.

cartography/intel/github/repos.py
```diff
@@ -785,6 +929,69 @@ def load_collaborators(
         )
 
 
+@timeit
+def load_python_requirements(
```
achantavy, line 1002: Nit: the diff in this PR confused me. I thought that this was a whole new function so I was confused about why we were adding in new manually written queries. If possible, avoid moving existing functions around too much so that it doesn't mess up the diffs and makes them easier to review

cartography/models/github/dependencies.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "REQUIRES"
+    properties: GitHubDependencyToRepositoryRelProperties = (
+        GitHubDependencyToRepositoryRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class GitHubDependencySchema(CartographyNodeSchema):
```
achantavy, line 66: Please update the schema.md docs as well

achantavy (changes requested):
Few bigger picture questions, maybe @hanzo or @heryxpc can chime in.

1. How sure are we that the GitHub dependency API serves as a full-on replacement for the older manual setup.cfg parsing?
2. If so, then we should eventually remove all previous setup.cfg parsing, requirements.txt parsing, etc. 

Currently this PR doesn't remove existing functionality, it just adds in GitHub's dep graph. Since @hanzo and @heryxpc are relying on the current functionality in produciton, definitely want to get your thoughts

achantavy:
Hey @aneeshusa :D 

Wondering if we can get your input. Remember how we built manual setup.cfg + requirements.txt parsing using the GitHub graphql? I think we've talked about this before, but do you know if GitHub's https://docs.github.com/en/graphql/reference/objects#dependencygraphmanifest serves as a full replacement for that one-off code we wrote?

More context here: https://github.com/cartography-cncf/cartography/issues/1348

Thank you!

cartography/intel/github/repos.py
```diff
@@ -785,6 +929,89 @@ def load_collaborators(
...
+    repo_urls: List[str],
+) -> None:
+    """
+    Delete GitHub dependencies and their relationships from the graph if they were not updated in the last sync.
+    :param neo4j_session: Neo4j session
+    :param common_job_parameters: Common job parameters containing UPDATE_TAG
+    :param repo_urls: List of repository URLs to clean up dependencies for
+    """
+    # Run cleanup for each repository separately
+    for repo_url in repo_urls:
```
achantavy, line 1008: Once the scoped_cleanup stuff is set correctly, you can simplify this by removing the loop and instead just calling

```
    GraphJob.from_node_schema(GitHubDepdencySchema(), common_job_parameters).run(
        neo4j_session
    )
```

this way you don't need to include `repo_url` as a param to the cleanup job anymore.

cartography/intel/github/repos.py
```diff
@@ -533,6 +557,126 @@ def _transform_setup_cfg_requirements(
...
+            # Extract version from requirements string if available
+            pinned_version = _extract_version_from_requirements(requirements)
+
+            # Create ecosystem-specific canonical name
+            canonical_name = _canonicalize_dependency_name(
+                package_name, package_manager
+            )
+
+            # Create unique ID for the dependency
+            dependency_id = (
```
achantavy, line 663: potential name collision between deps in different repos, and even between deps in the same repo but in different dep graph manifests. 

for id, consider concatenating more fields, and then including as separate properties all of those fields.

heryxpc:
> Few bigger picture questions, maybe @hanzo or @heryxpc can chime in.
> 
> 1. How sure are we that the GitHub dependency API serves as a full-on replacement for the older manual setup.cfg parsing?

It won't work for setup.cfg as this is not yet supported by Github https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/dependency-graph-supported-package-ecosystems#supported-package-ecosystems

> 2. If we are sure, then we should eventually remove all previous setup.cfg parsing, requirements.txt parsing, etc.
Don't do it yet.

Not yet.

> 
> Currently this PR doesn't remove existing functionality, it just adds in GitHub's dep graph. Since @hanzo and @heryxpc are relying on the current functionality in produciton, definitely want to get your thoughts

That's nice!, we can see each step (using the Dep Graph or the custom parsing) as enrichment/completeness of code dependencies for a given `GitHubRepository`.

Even if it's redundant with other dependencies information like [SemgrepDepdency](https://github.com/cartography-cncf/cartography/blob/master/docs/root/modules/semgrep/schema.md#semgrepdependency) there is value I believe to have a complete view from multiple sources.

Maybe this is a new case for ontologies, considering both nodes represent a code dependency and hence have:
- package manager
- package name
- version
The latter is a comment for future improvements, I don't think it's necessary to consider it now.

cartography/intel/github/repos.py
heryxpc: Github provides [203 as magic number](https://github.blog/enterprise-software/secure-software-development/secure-at-every-step-how-githubs-dependency-graph-is-generated/) for average total dependencies (I assume this includes transitive ones).
If I had to pick one value, I would go with 200.

cartography/intel/github/repos.py
heryxpc: This is a larger concern in Java ecosystem where projects have hierarchies of modules that hereby from parent modules. I would double it to `20` to consider more cases.

cartography/intel/github/repos.py
heryxpc: @shyammukund did you test if this returns transitive dependencies as well?
If it does then pagination is definitely needed as any "regular" production ready code might return hundreds/thousands of dependencies between direct and transitive.
If it doesn't return transitive dependencies, this code will only cover direct dependencies and might be good to clarify that in Cartography's documentation.

hanzo: From my testing on Go repos, this API call does return transitive dependencies as well. Sadly the API response does not mark dependencies as transitive vs direct, which is one of the reasons we opted to use [Semgrep](https://github.com/cartography-cncf/cartography/pull/1368) instead of the Github dependency graph.

shyammukund: @heryxpc Yes it does mark the transitive dependencies as well, however not marked as @hanzo said. I'll update it to fetch  20 manifest per pg and the first 200 dependencies in the manifest

aneeshusa:
Everyone else beat me to answering your question - looks like GitHub API is still missing the features that caused us to write our own for Python:
* `setup.cfg` support - though note most shops have moved to `uv`/`pyproject.toml` now, GitHub lists poetry support for pyproject.toml and pyproject.toml is a standard so it might work for uv users as well but you should test.
* showing direct vs transitive dependencies

Using the GraphQL API in general for languages it works well for seems reasonable though!

docs/root/modules/github/schema.md
```diff
@@ -285,6 +286,29 @@ Representation of a single Programming Language [language object](https://develo
     ```
 
 
+### Dependency
+
+Represents a software dependency from GitHub's dependency graph manifests.
```
achantavy: Include link to documentation to the underlying github object: https://docs.github.com/en/graphql/reference/objects#dependencygraphdependency

cartography/intel/github/repos.py
```diff
@@ -93,6 +96,18 @@
                             text
                         }
                     }
+                    dependencyGraphManifests(first: 20) {
```
achantavy, line 101: [non-block & no action required] For future reference if we need to paginate this, we should 

1. get the list of github repos from the graph (save on github API quota)
1. for each repo, do a separate graphql query to paginate the dependency graph manifests, returning their blobPaths. store that in the graph. 
1. for each dep graph manifest, paginate the dependencies.

cartography/intel/github/repos.py
shyammukund: Coming back to this I changed the dependencies back to 100 for now because when testing kept getting error that GraphQL API  has a max limit of 100 items per page. I'm guessing I need to add some pagination logic here to get around that.

cartography/models/github/manifests.py
```diff
@@ -0,0 +1,49 @@
...
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class DependencyGraphManifestNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id")
+    blob_path: PropertyRef = PropertyRef("blob_path", extra_index=True)
+    filename: PropertyRef = PropertyRef("filename", extra_index=True)
```
achantavy: extra_index not necessary here

cartography/models/github/manifests.py
```diff
@@ -0,0 +1,49 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class DependencyGraphManifestNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id")
+    blob_path: PropertyRef = PropertyRef("blob_path", extra_index=True)
```
achantavy: extra_index not necessary here because it's not really an ID that we expect to be queried frequently

cartography/models/github/dependencies.py
```diff
@@ -0,0 +1,74 @@
...
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class GitHubDependencyNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id")
+    name: PropertyRef = PropertyRef("name", extra_index=True)
+    original_name: PropertyRef = PropertyRef("original_name")
+    version: PropertyRef = PropertyRef("version")
+    ecosystem: PropertyRef = PropertyRef("ecosystem", extra_index=True)
```
achantavy: extra_index not necessary

cartography/intel/github/repos.py
```diff
@@ -533,6 +567,185 @@ def _transform_setup_cfg_requirements(
...
+
+        # Count dependencies in this manifest
+        dependencies = manifest.get("dependencies", {})
+        dependencies_count = len(dependencies.get("nodes", []) if dependencies else [])
+
+        # Create unique manifest ID by combining repo URL and blob path
+        manifest_id = f"{repo_url}#{blob_path}"
+
+        # Extract filename from blob path
+        filename = blob_path.split("/")[-1] if blob_path else ""
```
achantavy: Let's have the default set to `None` so that we don't have neo4j write a property to the graph set to the empty string (instead it will not write the property to the graph in the first place, which is what we want)

cartography/intel/github/repos.py
```diff
@@ -897,3 +1222,19 @@ def sync(
     repo_data = transform(repos_json, direct_collabs, outside_collabs)
     load(neo4j_session, common_job_parameters, repo_data)
     run_cleanup_job("github_repos_cleanup.json", neo4j_session, common_job_parameters)
+
+    # Collect repository URLs that have dependencies for cleanup
+    repo_urls_with_dependencies = list(
+        {dep["repo_url"] for dep in repo_data["dependencies"]}
+    )
+    cleanup_github_dependencies(
```
achantavy, line 1229: [block] oh, the correct order here should be 

1. cleanup deps
2. cleanup manifests
3. cleanup repos

because we're doing something unusual and making the repos the sub resource, and because repos _can_ be cleaned up themselves, we don't want to delete a stale repo before deleting a stale dep under that repo because that would make the stale dep become an orphan node and we'd be unable to clean that up

achantavy (approved):

PR #1671, opened 2025-07-03
feat: Add MatchLinks

achantavy:
Implements MatchLinks: a way to connect existing nodes in the graph together using Cartography's ORM.

- Automatic index creation for involved nodes _and_ relationships
- Automatic cleanup jobs scoped to the current sub resource (because sub resources are now stored as a property on matchlinks)
- Abstracts away cypher and graph query language
- Enables dev to more easily specify relationship properties using the ORM

Review the docs [here](https://github.com/cartography-cncf/cartography/blob/8958970ca3bf4cf8067f5b810ef768e1dc387274/docs/root/dev/matchlinks.md).

This would unblock the [cloudtrail management log spec](https://docs.google.com/document/d/1GvOeLB6zC1l0-PPn7wG-etybrP0bwlheKarxkJ_5_bs/edit?tab=t.0#heading=h.ylxiob8yqk4q).

### Checklist

- [see doc] Include a screenshot showing what the graph looked like before and after your changes.
- [see tests and doc] Include console log trace showing what happened before and after your changes.


cartography/graph/statement.py
```diff
@@ -56,7 +56,7 @@ def __init__(
 
         self.parent_job_name = parent_job_name if parent_job_name else None
         self.parent_job_sequence_num = (
-            parent_job_sequence_num if parent_job_sequence_num else None
+            parent_job_sequence_num if parent_job_sequence_num else 1
```
jychp, line 59: Not sure what is this modification for

cartography/graph/querybuilder.py
```diff
@@ -548,3 +549,136 @@ def build_create_index_queries(node_schema: CartographyNodeSchema) -> List[str]:
...
+        result.append(
+            index_template.safe_substitute(
+                NodeLabel=rel_schema.target_node_label,
+                NodeAttribute=target_key,
+            ),
+        )
+
+    # Create a composite index for the relationship between the source and target nodes.
+    # https://neo4j.com/docs/cypher-manual/4.3/indexes-for-search-performance/#administration-indexes-create-a-composite-index-for-relationships
+    rel_index_template = Template(
```
jychp, line 593: Oh nice, I did not know that was possible !

docs/root/dev/matchlinks.md
```diff
@@ -0,0 +1,201 @@
...
+1. Define a `CartographyRelProperties` object with some additional fields:
+    ```python
+    @dataclass(frozen=True)
+    class S3AccessRelProps(CartographyRelProperties):
+        # <Mandatory fields for MatchLinks>
+        lastupdated: PropertyRef = PropertyRef("UPDATE_TAG", set_in_kwargs=True)
+
+        # Cartography syncs objects account-by-account (or "sub-resource"-by-"sub-resource")
+        # We store the sub-resource label and id on the relationship itself so that we can
+        # clean up stale relationships without deleting relationships defined in other accounts.
```
jychp, line 57: It took me a while to understand the purpose of `_sub_resource_`.

I believe the name is misleading: the values provided here refer to some sort of parent of the relationship, not a true sub-resource. I understand this is tied to the broader `sub_resource` mechanism, but it still feels confusing to me.

More broadly, I think this highlights the limitations of our current approach, where nodes are always linked to a `Tenant` node. Making an indexed `tenant_id` (or `context_id`, or any relevant name) a required property similar to `lastupdated` on both Nodes and Relationships would likely simplify things significantly.

May be we can consider a change of paradigm and started to implement it on this PR and then progressively expends it to the whole code base ?

cartography/models/core/relationships.py
```diff
@@ -139,6 +162,22 @@ def direction(self) -> LinkDirection:
         """
         pass
 
+    @property
+    def source_node_label(self) -> str | None:
```
jychp, line 171: We should update the documentation and code comments to clarify that the `one_to_many` flag will be ignored in this context, due to the way the matching is implemented.

I think that’s a totally acceptable tradeoff to keep the codebase simple.

docs/root/dev/matchlinks.md
achantavy: It still is a sub resource. 

Here's the problem that this PR is trying to solve: we need a way to clean up stale relationships created via this new mechanism, and we need a way to scope the cleanups to the current sub resource otherwise all relationships outside of it will be cleaned up. (For globally-scoped relationships, I believe the analysis job pattern is a better fit.)

With nodes, we have a connection to the sub resource node, but with rels that doesn't make sense, so this current approach adds in a couple properties to the relationship itself. I have them prefixed with `_` to indicate that these are really meant for internal processing.

>Making an indexed tenant_id (or context_id, or any relevant name) a required property similar to lastupdated on both Nodes and Relationships would likely simplify things significantly.
> May be we can consider a change of paradigm and started to implement it on this PR and then progressively expends it to the whole code base ?

That's an interesting idea but would be a huge change to cartography. I'm not immediately against it but would need to balance that change against the other things we want to get done.

cartography/models/core/relationships.py
achantavy: Mm good point. Only some of the PropertyRef features are supported when using a MatchLink.

cartography/graph/statement.py
achantavy: Ah, this fixes a logging bug where it'll say something like "completed statement #None" because a parent_idx_seq_num wasn't passed in as it is an optional field. This makes it default to 1 instead of None.

cartography/graph/querybuilder.py
achantavy: Yeah, fairly new change in Neo4j. We haven't needed it even at scale, but if we're matching on multiple rel properties to clean them up in this way, I figured we should have some auto indexing.

jychp (approved):
LGTM 

Still think             `_sub_resource_label="AWSAccount",` is a bit confusing (as AWSAccount is more the parent resource of the current sub resource label) but I can live with it ;)

achantavy:
> Still think `_sub_resource_label="AWSAccount",` is a bit confusing (as AWSAccount is more the parent resource of the current sub resource) but I can live with it ;)

The word "sub resource" is a bit confusing haha. I've been using it to refer to the tenant or the parent, so in this case the sub resource is the AWS account. It was a term that was coined in the Slack channel a few years back.

jychp:
Let's move forward for now and we will discuss naming things with a beer because it will take so much time :)

kunaals (approved):
regurgitating from slack: i thought that we could simplify things by eliminating the "MatchLinks" concept. instead, we'd frame this change as "the way to create a relationship without creating any nodes" which would involve the CartographyRelSchema, and nothing else. after sleeping on it, i don't think it's blocker. it's deep enough in the data model that most contributors will never see it, and giving it a name will make it easier for those of us in the weeds to reference it. no blockers

Issue #1665, opened 2025-07-02
Add `User-Agent` header to all API requests


jychp:
**Description**:

Cartography currently makes HTTP requests without setting a `User-Agent` header. For better observability and integration with vendor APIs, all outbound requests from intel modules should include a standardized `User-Agent`.

**Proposed format**:

```
User-Agent: Cartography/0.x.y
```

Where `0.x.y` corresponds to the current version from `cartography.__version__`. If unavailable (e.g., in development), default to `Cartography/dev`.

**Benefits**:

* Easier debugging and tracing of requests in API logs
* Improves compatibility with vendors that expect or validate user-agents
* Aligns with standard HTTP client practices

**Suggested implementation**:

* Introduce a small utility (e.g., `build_session()`) in `cartography.util` to initialize a `requests.Session` with the appropriate `User-Agent`
* Refactor existing intel modules to use this session instead of directly instantiating `requests.Session()`

akhilvaitlaa7:
Hi, I’d like to take this up.

jychp:
@akhilvaitlaa7 feel free to work on it :)

wokaokf:
My name is [xingaonai](https://www.xgncrusher.com/). The article you wrote is very interesting and taught me a lot. Thank you!

Abhigaba:
Is this issue resolved ? , If not then I would like to try it.

jychp:
Hi @Abhigaba, there already is a PR for that issue: https://github.com/cartography-cncf/cartography/pull/1676

Omswastik-11:
Hi @jychp !! Is this issue is still open ?

PR #1676, opened 2025-07-06
Replaced direct requests.session() instantiation with build_session()…

devc007:
… in all relevant intel modules.

Replaced direct requests.Session/session() instantiation with build_session() in all relevant intel modules.
Ensures all outbound HTTP requests use a standardized User-Agent with the Cartography version.
Affects Anthropic, OpenAI, Tailscale, CVE, and Kandji integrations.

- https://github.com/cartography-cncf/cartography/issues/1665

### DIFF
--- a/cartography/intel/anthropic/__init__.py
+++ b/cartography/intel/anthropic/__init__.py
@@ -1,13 +1,12 @@
 import logging
 
 import neo4j
-import requests
 
 import cartography.intel.anthropic.apikeys
 import cartography.intel.anthropic.users
 import cartography.intel.anthropic.workspaces
 from cartography.config import Config
-from cartography.util import timeit
+from cartography.util import timeit, build_session
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


achantavy:
Hey @devc007, just wondering, what benefit do you see from making this change?

devc007:
@achantavy This issue was opened by @jychp, who identified that adding a standardized User-Agent header with the Cartography version would provide significant benefits for observability and vendor API integration.

Along with this I think here are some more benifit this could provide
1. Request tracing: API providers can easily identify and filter Cartography requests in their logs by version. 
2. Faster troubleshooting: When API calls fail, both our team and vendors can quickly isolate Cartography-specific issues.
3. Version-specific analytics: Track how different Cartography versions interact with various APIs.

The proposed implementation using a centralized build_session() utility in cartography.util looks solid - it ensures consistency across all intel modules while keeping the change minimal and non-breaking.
Please let me know if you think the implementation approach should be adjusted or if there are other considerations we should account for.

devc007:
Hey @achantavy i am hoping that this PR is good if not let me know, what need to update.

jychp:
@devc007, would you have time to revisit this PR and fix the errors flagged by the CI?
Totally fine if you can’t, just let me know so we can plan accordingly.

devc007:
@jychp Thanks for asking! Right now, I’m focusing on kgateway for LFX, so you can carry on with it.

jychp:
Close in favor of #1839

PR #1680, opened 2025-07-08
feat: SecurityGroupRule-ID is read, instead of generated

balint-tukacs-tc2:
### Summary
This PR updates the EC2 security group rules implementation to use AWS's native `SecurityGroupRuleId` as the unique identifier.
Changes:
   - Updated schema documentation for **IpRule** nodes
   - Clarified **IpPermissionInbound** vs **IpPermissionEgress** labels
   - Fixed relationship documentation
   - Added examples and improved field descriptions

### Checklist

Screenshot 2025-07-08 at 9 50 00


achantavy:
Hey @balint-tukacs-tc2, thanks for taking a stab at this. I think there are some larger issues with the EC2 security group PR that prevent us from merging this one immediately. Specifically, we need to update it to use the new data model first before changing other parts of the code. Can you please file this instead as an issue and when we do the refactor we'll address the issue?

achantavy:
@balint-tukacs-tc2 - I made the updates the the EC2 security group. I'm supportive of the change to use the built-in ruleid instead of generating it. Let's do this:

- ruleid: set this to the built-in rule id from aws
- arn: set this to the arn determined by AWS
- id: set this to the arn determined by AWS

And then change this https://github.com/cartography-cncf/cartography/blob/4b1617d6838e57bcecead7da4e86c393ef48af5a/cartography/intel/aws/resourcegroupstaggingapi.py#L116-L120

to 

```python
    "ec2:security-group": {
        "label": "EC2SecurityGroup",
        "property": "id",
    },
```

achantavy:
Going to close this out. If you're still able to, would appreciate you making those changes i outlined above in a separate PR. Thank you!

balint-tukacs-tc2:
Hi @achantavy,

Thanks for your feedback and inputs! Will look into it, after my vacation.

PR #1681, opened 2025-07-08
SentinelOne: Ingest agents

serge-wq:
### Summary
Adding the agent (endpoint) ingestion for S1 intelmodule which includes:

- Data models
- Sync job
- Updated docs with current state of the whole S1 intelmodule
- Created two different files for utility functions and API access (`api.py`, `utils.py`):
  - `utils.py` currently has a function that is not used with the current models, but will be used when application ingestion is added. We can discuss what it does in following PRs. 
- Nit fix: ordered the "Supported platforms" section of the README.md alphabetically. The list is starting to get long and having an order could make it easier to read

Screenshot 2025-07-08 at 3 20 02 p m

Screenshot 2025-07-08 at 3 50 06 p m

### Related issues or links

### Checklist


cartography/intel/sentinelone/agent.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    Get agent data from SentinelOne API
+    :param api_url: The SentinelOne API URL
+    :param api_token: The SentinelOne API token
+    :param account_id: The SentinelOne account ID
+    :return: Raw agent data from API
+    """
+    logger.info(f"Retrieving SentinelOne agent data for account {account_id}")
+
+    agents = get_paginated_results(
```
heryxpc, line 26: This is just from the recent hiccup of handling massive amount of paginated results (~2M).
Would you consider necessary to add batch processing to get the results?

The [batch](https://github.com/cartography-cncf/cartography/blob/9a0a3f0d383ef9218cb4881620c55669afbef961/cartography/util.py#L360) function could handle sets of calls if `get_paginated_results` yields results 1 by 1.
This could reduce the memory and allow to process new pages only when required (like during `transform_agents`) with minimum code changes.

docs/root/modules/sentinelone/schema.md
```diff
@@ -0,0 +1,54 @@
...
+Represents a SentinelOne agent installed on an endpoint device.
+
+| Field | Description |
+|-------|-------------|
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | The unique identifier for the SentinelOne agent |
+| **uuid** | The UUID of the agent |
+| **computer_name** | The name of the computer where the agent is installed |
+| **serial_number** | The serial number of the endpoint device |
```
heryxpc, line 39: Do you have any sample data on how this should look like?

heryxpc (approved):
LGTM, left a few comments just to consider ingestion of large scale deployments (e.g. millions of devices).

cartography/intel/sentinelone/agent.py
serge-wq: good callout, let me look into this.

cartography/intel/sentinelone/agent.py
```diff
@@ -0,0 +1,143 @@
...
+    Load SentinelOne agent data into Neo4j
+    :param neo4j_session: Neo4j session
+    :param data: The transformed agent data
+    :param account_id: The SentinelOne account ID
+    :param update_tag: Update tag for tracking data freshness
+    :return: None
+    """
+    if not data:
+        logger.warning(f"No agent data provided for account {account_id}")
+        return
```
achantavy: Not a necessary check, the lower level load function handles this case

cartography/intel/sentinelone/__init__.py
```diff
@@ -24,19 +25,33 @@ def start_sentinelone_ingestion(neo4j_session: neo4j.Session, config: Config) ->
...
-        config.sentinelone_api_token,
-        config.update_tag,
         common_job_parameters,
+        config.sentinelone_account_ids,
     )
 
+    # Sync agents for each account
+    for account_id in synced_account_ids:
+        # Add account-specific parameter
+        common_job_parameters["ACCOUNT_ID"] = account_id
```
achantavy: Nit: might be good to call this a S1_ACCOUNT_ID so there's no risk of collision in future

cartography/intel/sentinelone/utils.py
```diff
@@ -1,120 +1,9 @@
...
-        if current_page >= pagination.get("totalPages", 1):
-            break
-
-        current_page += 1
-
-    return total_items
+def get_application_id(name: str, vendor: str) -> str:
+    name_normalized = name.strip().lower().replace(" ", "_")
+    vendor_normalized = vendor.strip().lower().replace(" ", "_")
+    name_normalized = re.sub(r"[^\w]", "", name_normalized)
```
achantavy, line 7: Are the regexes necessary?

achantavy (approved):
no blockers, just add in the S1_ACCOUNT_ID instead of ACCOUNT_ID to avoid collision in future

PR #1041, opened 2022-12-08
Adds Lambda Support to AWS Inspector and fixes relationship bugs

csanders-git:

PR #897, opened 2022-08-17
Add AWS Inspector2

nathan-v:
Add AWS Inspector v2 support

PR #1682, opened 2025-07-10
fix: AWSInspector finding to package rel and missed properties

heryxpc:
### Summary
One finding can have multiple affected packages from boto3 response to [list_findings](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/inspector2/client/list_findings.html). This is modeled in the graph with the nodes/relationship `(:AWSInspectorFinding)-[:HAS]->(:AWSInspectorPackage)`. One package node was created with a custom id result from concatenating package `name`, `arch`, `version`, `release` and `epoch`.

Considering multiple findings could affect the same package based on the custom id, there were missing packages to findings, as it was only the last package with a unique `id` and `findingarn` being ingested to the graph. This left many findings with no `:HAS` relationship to a package.

- Modified the package id string to match AWS inspector format ([using RPM naming conventions](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/)).
- Relocated `fixedinversion` attribute and added `remediation`, `filepath`, `sourcelayerhash` and `sourcelambdalayerarn` to `:HAS` relationship.
- Handled the many to many relationship with [MatchLinks](https://github.com/cartography-cncf/cartography/blob/master/docs/root/dev/matchlinks.md)

Before:
Screenshot 2025-07-09 at 4 59 09 p m
Screenshot 2025-07-10 at 4 05 56 p m
Screenshot 2025-07-10 at 4 06 04 p m

After:
Screenshot 2025-07-09 at 4 59 25 p m
Screenshot 2025-07-10 at 4 05 39 p m
Screenshot 2025-07-10 at 4 05 46 p m

### Related issues or links
- https://github.com/cartography-cncf/cartography/pull/897
- https://github.com/cartography-cncf/cartography/pull/1041
- https://github.com/cartography-cncf/cartography/pull/1671

### Checklist


cartography/intel/aws/inspector.py
```diff
@@ -163,55 +167,45 @@ def transform_inspector_findings(
...
+        if epoch:
+            epoch = f"{epoch}:"
+        version = package["version"]  # Mandatory field
+        release = package.get("release", "")
+        if release:
+            release = f"-{release}"
+        arch = package.get("arch", "")
+        if arch:
+            arch = f".{arch}"
+        id = f"{name}|{epoch}{version}{release}{arch}"
```
heryxpc, line 207: Pattern in AWS Inspector UI and `fixVersion` attrinbute is `epoch:version-release.arch`, example: `1:2.06-61.amzn2023.0.16`

cartography/intel/aws/inspector.py
```diff
@@ -244,20 +238,35 @@ def load_inspector_findings(
 def load_inspector_packages(
     neo4j_session: neo4j.Session,
     packages: List[Dict[str, Any]],
-    region: str,
```
heryxpc, line 247: Adding region here adds no value, since the same package-version will probably be installed in multiple instances/containers/lambdas from different regions.
The only important piece is the `AWS_ACCOUNT` where the package was located to delete the package when the account is no longer synced.

cartography/intel/aws/inspector.py
```diff
@@ -244,20 +238,35 @@ def load_inspector_findings(
...
+    aws_update_tag: int,
+    current_aws_account_id: str,
+) -> None:
+    load_matchlinks(
+        neo4j_session,
+        InspectorFindingToPackageMatchLink(),
+        finding_to_package_map,
+        lastupdated=aws_update_tag,
+        _sub_resource_label="AWSAccount",
+        _sub_resource_id=current_aws_account_id,
```
heryxpc, line 266: Curious here if the `_sub_resource_*` should be the `AWSAcccount` or the `AWSInspectorFinding` related to the package.

cartography/intel/aws/inspector.py
```diff
@@ -270,6 +279,14 @@ def cleanup(
     GraphJob.from_node_schema(AWSInspectorPackageSchema(), common_job_parameters).run(
         neo4j_session,
     )
+    GraphJob.from_matchlink(
+        InspectorFindingToPackageMatchLink(),
+        "AWSAccount",
+        common_job_parameters["ACCOUNT_ID"],
+        common_job_parameters["UPDATE_TAG"],
```
heryxpc, line 286: Same comment as above.

cartography/models/aws/inspector/findings.py
```diff
@@ -135,6 +137,40 @@ class InspectorFindingToECRImageRel(CartographyRelSchema):
     )
 
 
+@dataclass(frozen=True)
+class InspectorFindingToPackageRelRelProperties(CartographyRelProperties):
```
heryxpc, line 141: Adding here the `(:AWSInspectorFinding)-[:HAS]->(:AWSInspectorPackage)` relationship and it's properties.

cartography/models/aws/inspector/findings.py
```diff
@@ -146,6 +182,7 @@ class AWSInspectorFindingSchema(CartographyNodeSchema):
     other_relationships: OtherRelationships = OtherRelationships(
         [
             InspectorFindingToEC2InstanceRel(),
+            # TODO: Fix ECRRepository and ECRImage relationships
```
heryxpc, line 185: This can be ignored, I'm not 💯  this was failing due to my local sync missing data.

serge-wq (approved):
LGTM

achantavy:
As discussed, ideally we'd have per-relationship attributes on a one-to-many situation, but we don't, so the matchlinks will accomplish this for now

PR #1683, opened 2025-07-10
Adds AWS Eventbridge intel module

AdityaPandeyCN:
### Summary
Added AWS Eventbridge Module, this is still work in progress

### Checklist


AdityaPandeyCN:
@achantavy  Can you have a look if time permits?

cartography/intel/aws/eventbridge.py
```diff
@@ -0,0 +1,210 @@
...
+
+    rules: list[dict[str, Any]] = []
+    for page in client.get_paginator("list_rules").paginate():
+        rules.extend(page.get("Rules", []))
+
+    targets_by_rule: dict[str, list[dict[str, Any]]] = {}
+    for rule in rules:
+        rule_name = rule["Name"]
+        targets: list[dict[str, Any]] = []
+        for page in client.get_paginator("list_targets_by_rule").paginate(
```
d-aggarwal, line 33: It’s recommended to model targets as a separate EventBridge Target node rather than embedding _target_arns.
If this PR is intended to focus only on EventBridge rule ingestion, it’s suggested to skip target logic for now and handle it separately.

cartography/intel/aws/eventbridge.py
```diff
@@ -0,0 +1,210 @@
...
+                        rule_arn=rule_arn,
+                        target_arn=target_arn,
+                        target_id=target["id"],
+                        target_role_arn=target.get("role_arn"),
+                        update_tag=update_tag,
+                    )
+                    break
+
+
+def _create_role_relationships(
```
d-aggarwal, line 146: It would be better to define relationships in the schema instead of creating them directly via Neo4j queries in the sync.

d-aggarwal (changes requested):
Thanks for the PR! Couple of suggestions for better alignment with the newer Cartography model.

PR #1684, opened 2025-07-12
Feat: AWS GuardDuty Ingestion

shyammukund:
### Summary
This PR implements
GuardDuty Intel Module
GuardDuty Schema
Documents GuardDuty Schema
Unit and Integration Tests for this feature

### Checklist

- [ ] Graph Showing Rel Between AWSRole, the finding, and which EC2 Instance it Affects
image

- [ ]  console log trace 
Screenshot 2025-07-11 at 5 03 23 PM


cartography/intel/aws/guardduty.py
```diff
@@ -0,0 +1,215 @@
...
+    if not detector_ids:
+        logger.info(f"No GuardDuty detectors found in region {region}")
+        return []
+
+    logger.info(f"Found {len(detector_ids)} GuardDuty detectors in region {region}")
+    return detector_ids
+
+
+@timeit
+def get_findings(
```
kunaals, line 80: Not sure if this makes sense, but might be good to filter out closed findings? Not sure if the API returns them though - worth checking

cartography/models/aws/guardduty/findings.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "AFFECTS"
+    properties: GuardDutyFindingToEC2InstanceRelRelProperties = (
+        GuardDutyFindingToEC2InstanceRelRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class GuardDutyFindingToS3BucketRelRelProperties(CartographyRelProperties):
```
kunaals, line 72: In a followup PR let's be sure to add rels for other finding types: https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_finding-types-active.html

kunaals: Let's make an issue for that after this PR is merged

cartography/intel/aws/guardduty.py
kunaals, line 80: If an org has a ton of findings, might be good to filter by severity. Perhaps we can add a `--aws-guardduty-severity-threshold` as a config option and filter on that?

shyammukund: now filters out archived findings.

shyammukund: Added optional filtering by severity level in CLI. For instance if user specifices --aws-security-threshold HIGH program will filter out findings that arent HIGH are CRITICAL severity

tests/integration/cartography/intel/aws/test_guardduty.py
```diff
@@ -0,0 +1,172 @@
...
+    mock_get_findings.assert_called()
+
+    # Verify the call included the severity_threshold parameter (as positional argument)
+    call_args = mock_get_findings.call_args
+    assert (
+        len(call_args.args) >= 4
+    ), f"Expected at least 4 positional args, got {len(call_args.args)}"
+    assert (
+        call_args.args[3] == "HIGH"
+    ), f"Expected severity_threshold='HIGH' as 4th arg, got {call_args.args}"
```
achantavy: Avoid call_args tests because they're brittle

docs/root/modules/aws/schema.md
```diff
@@ -139,6 +140,52 @@ Representation of AWS [IAM Groups](https://docs.aws.amazon.com/IAM/latest/APIRef
...
+    ```
+
+- GuardDuty findings may affect S3 Buckets
+    ```cypher
+    (:GuardDutyFinding)-[:AFFECTS]->(:S3Bucket)
+    ```
+
+- GuardDuty findings are labeled as security risks
+    ```cypher
+    (:GuardDutyFinding)-[]->(:Risk)
```
achantavy: [block] this indicates that there's a relationship which is not correct. I'd remove and instead annotate the title as

```
### GuardDutyFinding::Risk
```

at the heading

achantavy (changes requested):
Minor changes to docs

other than that, LGTM!

achantavy:
oh also remember to update readme.md

achantavy (approved):
lgtm

PR #1685, opened 2025-07-15
Feat: Added SAML Events to CloudTrail mgmt events feature

shyammukund:
[- https://github.com/cartography-cncf/cartography/issues/...](https://github.com/cartography-cncf/cartography/pull/1662)

### Checklist

- [ ] Graph Showing Both AssumedRole and AssumedRoleWithSAML Rel
image


cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
     return list(aggregated.values())
 
 
+@timeit
+def transform_saml_role_events_to_role_assumptions(
+    events: List[Dict[str, Any]],
+    region: str,
+    current_aws_account_id: str,
```
achantavy: [block] this parameter is not used in this function; remove

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
     return list(aggregated.values())
 
 
+@timeit
+def transform_saml_role_events_to_role_assumptions(
+    events: List[Dict[str, Any]],
+    region: str,
```
achantavy: [block] This param doesn't add much to this function except for the log message. I would remove the log message and remove the param

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
...
+            )
+            continue
+
+        event_time = event.get("EventTime")
+
+        key = (source_principal, destination_principal)
+
+        if key in aggregated:
+            aggregated[key]["times_used"] += 1
+            aggregated[key]["saml_count"] += 1  # All events are AssumeRoleWithSAML
```
achantavy: No reason to use an extra field if we're only tracking saml, right?

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
...
+                if existing_last is None or event_time > existing_last:
+                    aggregated[key]["last_used"] = event_time
+        else:
+            aggregated[key] = {
+                "source_principal_arn": source_principal,
+                "destination_principal_arn": destination_principal,
+                "times_used": 1,
+                "first_seen_in_time_window": event_time,
+                "last_used": event_time,
+                "saml_count": 1,
```
achantavy: Same here about only needing one field. Use just `times_used`.

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
...
+            }
+
+    return list(aggregated.values())
+
+
+@timeit
+def transform_web_identity_role_events_to_role_assumptions(
+    events: List[Dict[str, Any]],
+    region: str,
+    current_aws_account_id: str,
```
achantavy: [block] these params aren't needed

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
     return list(aggregated.values())
 
 
+@timeit
+def transform_saml_role_events_to_role_assumptions(
+    events: List[Dict[str, Any]],
+    region: str,
+    current_aws_account_id: str,
+) -> List[Dict[str, Any]]:
```
achantavy, line 273: Since we are defining the returned data type here, it'd be helpful to give information on what the data shape is for future developers. There are a couple ways to do this: we can use a namedtuple so that the field names are explicit (instead of needing to guess it with dict keys that can be easily mis-typed). Or we can simply write that in the doc string that this function returns a list of dicts that have these keys.

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
...
+
+    return list(aggregated.values())
+
+
+@timeit
+def transform_web_identity_role_events_to_role_assumptions(
+    events: List[Dict[str, Any]],
+    region: str,
+    current_aws_account_id: str,
+) -> List[Dict[str, Any]]:
```
achantavy, line 345: Same comment about making the datashape easy to follow by either using namedtuple or documenting the key names

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -158,6 +278,194 @@ def transform_assume_role_events_to_role_assumptions(
...
+                    github_aggregated[key] = {
+                        "source_repo_fullname": github_repo,
+                        "destination_principal_arn": destination_principal,
+                        "times_used": 1,
+                        "first_seen_in_time_window": event_time,
+                        "last_used": event_time,
+                        "web_identity_count": 1,
+                        "users": {
+                            github_repo
+                        },  # during aggregation using set to avoid duplicates
```
achantavy: Same comment about only including keys that are strictly necessary for the load function. The entire object will be passed to Neo4j, and at large scale it's wasteful to include a field if it won't be written to the graph

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -341,6 +733,184 @@ def sync_assume_role_events(
...
+    """
+    Sync CloudTrail SAML management events to create ASSUMED_ROLE_WITH_SAML relationships.
+
+    This function orchestrates the complete process:
+    1. Fetch CloudTrail SAML management events region by region
+    2. Transform events into role assumption records per region
+    3. Load role assumption relationships into Neo4j for each region
+
+    The resulting graph contains direct relationships like:
+    (AWSRole)-[:ASSUMED_ROLE_WITH_SAML {times_used, first_seen_in_time_window, last_used, lastused}]->(AWSRole)
```
achantavy: [block] why last_used and lastused? Should pick one

cartography/models/aws/cloudtrail/management_events.py
```diff
@@ -62,3 +59,143 @@ class AssumedRoleMatchLink(CartographyRelSchema):
...
+    times_used: PropertyRef = PropertyRef("times_used")
+    first_seen_in_time_window: PropertyRef = PropertyRef("first_seen_in_time_window")
+
+    # Web Identity-specific event tracking
+    web_identity_count: PropertyRef = PropertyRef("web_identity_count")
+
+    # Individual user tracking within provider-level aggregation
+    web_identity_users: PropertyRef = PropertyRef(
+        "web_identity_users"
+    )  # List of unique users
```
achantavy: Do we need this? This list can get huge

cartography/models/aws/cloudtrail/management_events.py
```diff
@@ -62,3 +59,143 @@ class AssumedRoleMatchLink(CartographyRelSchema):
...
+    # CloudTrail-specific relationship properties
+    last_used: PropertyRef = PropertyRef("last_used")
+    times_used: PropertyRef = PropertyRef("times_used")
+    first_seen_in_time_window: PropertyRef = PropertyRef("first_seen_in_time_window")
+
+    # Web Identity-specific event tracking
+    web_identity_count: PropertyRef = PropertyRef("web_identity_count")
+
+    # Individual user tracking within provider-level aggregation
+    web_identity_users: PropertyRef = PropertyRef(
```
achantavy: What's the difference between this and unique_user_count? Should pick one if they are the same

achantavy:
Oh, also update the PR title to include the assumerolewithwebidentity stuff

cartography/intel/aws/cloudtrail_management_events.py
shyammukund: fixed so docstring includes the keys that are contained in the return list of dicts

cartography/intel/aws/cloudtrail_management_events.py
shyammukund: removed users because we no longer are tracking the list of users and removed web count as times used tells us the same thing

cartography/intel/aws/cloudtrail_management_events.py
shyammukund: typo in docstring. should be lastupdated which refers to when cartography last updated this info (used in load_matchlinks) and last_used is when something in AWS happened. fixed to be accurate.

cartography/models/aws/cloudtrail/management_events.py
shyammukund: now that i removed the list of users it tells us the same thing as times used. deleted so that all 3 types of events only show times used.

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -149,15 +262,172 @@ def transform_assume_role_events_to_role_assumptions(
...
+                        "destination_principal_arn": destination_principal,
+                        "times_used": 1,
+                        "first_seen_in_time_window": event_time,
+                        "last_used": event_time,
+                    }
+            else:
+                # Skip non-GitHub events for now
+                logger.debug(
+                    f"Skipping non-GitHub WebIdentity event from provider '{identity_provider}'. Event: {event.get('EventId', 'unknown')}"
+                )
```
achantavy: remove this

cartography/intel/aws/cloudtrail_management_events.py
```diff
@@ -149,15 +262,172 @@ def transform_assume_role_events_to_role_assumptions(
...
+                continue
+        else:
+            logger.debug(
```
achantavy: remove this

cartography/models/aws/cloudtrail/management_events.py
```diff
@@ -62,3 +56,129 @@ class AssumedRoleMatchLink(CartographyRelSchema):
...
+    _sub_resource_id: PropertyRef = PropertyRef("_sub_resource_id", set_in_kwargs=True)
+
+    # CloudTrail-specific relationship properties
+    last_used: PropertyRef = PropertyRef("last_used")
+    times_used: PropertyRef = PropertyRef("times_used")
+    first_seen_in_time_window: PropertyRef = PropertyRef("first_seen_in_time_window")
+
+
+@dataclass(frozen=True)
+class AssumeRoleWithWebIdentityMatchLink(CartographyRelSchema):
```
achantavy: Does this conflict with the GitHubRepository matchlinks?

achantavy (changes requested):
Also update the schema docs. We're going from AWSSSOUser to AWSRole, as well as GitHubRepository to AWSRole

tests/integration/cartography/intel/aws/test_cloudtrail_management_events.py
```diff
@@ -35,43 +37,70 @@
...
+    """Clean up CloudTrail test data between tests."""
+    neo4j_session.run("MATCH ()-[r:ASSUMED_ROLE]-() DELETE r")
+    neo4j_session.run("MATCH ()-[r:ASSUMED_ROLE_WITH_SAML]-() DELETE r")
+    neo4j_session.run("MATCH ()-[r:ASSUMED_ROLE_WITH_WEB_IDENTITY]-() DELETE r")
+    neo4j_session.run("MATCH (n:AWSPrincipal) DETACH DELETE n")
+    neo4j_session.run("MATCH (n:AWSUser) DETACH DELETE n")
+    neo4j_session.run("MATCH (n:AWSRole) DETACH DELETE n")
+    neo4j_session.run("MATCH (n:AWSSSOUser) DETACH DELETE n")
+    neo4j_session.run("MATCH (n:GitHubRepository) DETACH DELETE n")
+    neo4j_session.run("MATCH (n:AWSAccount) DETACH DELETE n")
```
achantavy: Can just simplify to `match (n) detach delete n`

cartography/models/aws/cloudtrail/management_events.py
shyammukund: no because the githubrepo matchlinks use the GitHubRepoAssumeRoleWithWebIdentityMatchLink.

tests/integration/cartography/intel/aws/test_cloudtrail_management_events.py
```diff
@@ -383,11 +411,482 @@ def test_cloudtrail_management_events_handles_cross_account_relationships(
...
+@patch.object(
+    cartography.intel.aws.cloudtrail_management_events,
+    "get_saml_role_events",
+    return_value=[],
+)
+@patch.object(
+    cartography.intel.aws.cloudtrail_management_events,
+    "get_web_identity_role_events",
+    return_value=[
+        {
```
achantavy: It's cleaner to move this to its own file in tests/data/

tests/integration/cartography/intel/aws/test_cloudtrail_management_events.py
```diff
@@ -383,11 +411,482 @@ def test_cloudtrail_management_events_handles_cross_account_relationships(
...
+    """
+    # Arrange
+    _cleanup_cloudtrail_test_data(neo4j_session)
+    create_test_account(
+        neo4j_session, INTEGRATION_TEST_BASIC_ACCOUNT_ID, TEST_UPDATE_TAG
+    )
+    _ensure_local_neo4j_has_basic_test_data(neo4j_session)
+
+    # Create AWS roles that GitHub repos will assume
+    github_target_roles = [
```
achantavy: It's cleaner to move this to its own file in tests/data/

tests/integration/cartography/intel/aws/test_cloudtrail_management_events.py
```diff
@@ -383,11 +411,482 @@ def test_cloudtrail_management_events_handles_cross_account_relationships(
...
+)
+@patch.object(
+    cartography.intel.aws.cloudtrail_management_events,
+    "get_saml_role_events",
+    return_value=[],
+)
+@patch.object(
+    cartography.intel.aws.cloudtrail_management_events,
+    "get_web_identity_role_events",
+    return_value=[
```
achantavy: Ditto on tests/data

cartography/models/aws/cloudtrail/management_events.py
shyammukund: which uses the same matchlink properties as the AssumeRoleWithWebIdentityMatchLink

shyammukund: Removed general AssumeRoleWithWebIdentityMatchLink because it isn't being used.

PR #1690, opened 2025-07-16
feat: SentinelOne Applications Ingestion

serge-wq:
### Summary

Adding application ingestion as part of the SentinelOne Intelmodule implementation. 

Screenshot 2025-07-15 at 11 13 43 p m

### Related issues or links

### Checklist

### DIFF
--- a/cartography/intel/sentinelone/application.py
+++ b/cartography/intel/sentinelone/application.py
@@ -18,15 +18,16 @@ logger = logging.getLogger(__name__)
 
 
 @timeit
-def get_application_data(common_job_parameters: dict[str, Any]) -> list[dict[str, Any]]:
+def get_application_data(
+    account_id: str, api_url: str, api_token: str
+) -> list[dict[str, Any]]:
     """
     Get application data from SentinelOne API
-    :param common_job_parameters: Common job parameters containing API_URL, API_TOKEN, etc.
+    :param account_id: SentinelOne account ID
+    :param api_url: SentinelOne API URL
+    :param api_token: SentinelOne API token
     :return: A list of application data dictionaries
     """
-    account_id = str(common_job_parameters.get("ACCOUNT_ID"))
-    api_url = str(common_job_parameters.get("API_URL"))
-    api_token = str(common_job_parameters.get("API_TOKEN"))
     logger.info(f"Retrieving SentinelOne application data for account {account_id}")
     applications = get_paginated_results(
         api_url=api_url,
@@ -44,17 +45,20 @@ def get_application_data(common_job_parameters: dict[str, Any]) -> list[dict[str
 
 @timeit
 def get_application_installs(
-    app_inventory: list[dict[str, Any]], common_job_parameters: dict[str, Any]
+    app_inventory: list[dict[str, Any]], account_id: str, api_url: str, api_token: str
 ) -> list[dict[str, Any]]:
     """
     Get application installs from SentinelOne API
-    :param common_job_parameters: Common job parameters containing API_URL, API_TOKEN, etc.
+    :param app_inventory: List of applications to get installs for
+    :param account_id: SentinelOne account ID
+    :param api_url: SentinelOne API URL
+    :param api_token: SentinelOne API token
     :return: A list of application installs data dictionaries
     """
     logger.info(
         f"Retrieving SentinelOne application installs for "
         f"{len(app_inventory)} applications in account "
-        f"{common_job_parameters.get('ACCOUNT_ID')}",
+        f"{account_id}",
     )
 
     application_installs = []
@@ -63,21 +67,14 @@ def get_application_installs(
             f"Retrieving SentinelOne installs for {app.get('applicationName')} "
             f"{app.get('applicationVendor')} ({i + 1}/{len(app_inventory)})",
         )
-        if "applicationVendor" not in app or "applicationName" not in app:
-            logger.error(
-                "Application object does not contain applicationVendor or applicationName",
-                app,
-            )
-            continue
-
-        name = app.get("applicationName")
-        vendor = app.get("applicationVendor")
+        name = app["applicationName"]
+        vendor = app["applicationVendor"]
         app_installs = get_paginated_results(
-            api_url=str(common_job_parameters.get("API_URL")),
+            api_url=api_url,
             endpoint="/web/api/v2.1/application-management/inventory/endpoints",
-            api_token=str(common_job_parameters.get("API_TOKEN")),
+            api_token=api_token,
             params={
-                "accountIds": str(common_job_parameters.get("ACCOUNT_ID")),
+                "accountIds": account_id,
                 "limit": 1000,
                 "applicationName": name,
                 "applicationVendor": vendor,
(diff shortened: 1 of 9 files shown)


cartography/intel/sentinelone/application.py
```diff
@@ -0,0 +1,250 @@
...
+        f"Retrieving SentinelOne application installs for "
+        f"{len(app_inventory)} applications in account "
+        f"{account_id}",
+    )
+
+    application_installs = []
+    for i, app in enumerate(app_inventory):
+        logger.info(
+            f"Retrieving SentinelOne installs for {app.get('applicationName')} "
+            f"{app.get('applicationVendor')} ({i + 1}/{len(app_inventory)})",
```
heryxpc, line 68: Is this just the progress counter?

cartography/intel/sentinelone/application.py
```diff
@@ -0,0 +1,250 @@
...
+) -> list[dict[str, Any]]:
+    """
+    Transform SentinelOne application data for loading into Neo4j
+    :param applications_list: Raw application data from the API
+    :return: Transformed application data
+    """
+    transformed_data = []
+    for app in applications_list:
+        vendor = app["applicationVendor"].strip()
+        name = app["applicationName"].strip()
```
heryxpc, line 105: Why stripping here and not from the `get` function?

cartography/models/sentinelone/application_version.py
```diff
@@ -0,0 +1,96 @@
...
+    properties: S1ApplicationVersionToAccountRelProperties = (
+        S1ApplicationVersionToAccountRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class S1AgentToApplicationVersionRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+    installed_dt: PropertyRef = PropertyRef("installed_dt")
+    installation_path: PropertyRef = PropertyRef("installation_path")
```
heryxpc: Just for consistency with other attributes, I would rename these to `installeddatetime` and `installationpath`.

cartography/intel/sentinelone/utils.py
```diff
@@ -7,3 +7,17 @@ def get_application_id(name: str, vendor: str) -> str:
...
+    """
+    Generate a unique identifier for an application version
+    :param name: Application name
+    :param vendor: Application vendor
+    :param version: Application version
+    :return: Unique identifier for the application version
+    """
+    app_id = get_application_id(name, vendor)
+    version_normalized = version.strip().lower().replace(" ", "_")
+    version_normalized = re.sub(r"[^\w.-]", "", version_normalized)
```
heryxpc: Why removing `-` from the version?

heryxpc (approved):
Looks good, left a few minor comments, and please add a screenshot of the new nodes and relationships.

cartography/intel/sentinelone/application.py
serge-wq: yes

cartography/intel/sentinelone/application.py
```diff
@@ -0,0 +1,250 @@
...
+        # for consistency with the application data
+        for app_install in app_installs:
+            app_install["applicationVendor"] = vendor
+            app_install["applicationName"] = name
+        application_installs.extend(app_installs)
+
+    return application_installs
+
+
+@timeit
```
achantavy: Nit: timeit not needed for a transform

cartography/intel/sentinelone/application.py
achantavy: It's fine to let the `get` be as dumb as possible because during the integ tests the `get` is patched out so any transform logic that occurs there will not be applied to the test

cartography/intel/sentinelone/application.py
```diff
@@ -0,0 +1,250 @@
...
+            "id": app_id,
+            "name": name,
+            "vendor": vendor,
+        }
+        transformed_data.append(transformed_app)
+
+    return transformed_data
+
+
+@timeit
```
achantavy: nit: timeit not needed for transform

cartography/models/sentinelone/application.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: S1ApplicationToAccountRelProperties = (
+        S1ApplicationToAccountRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class S1ApplicationSchema(CartographyNodeSchema):
```
achantavy, line 41: No association with the host that it is installed on? or will that come later :)

tests/unit/cartography/intel/sentinelone/test_application.py
```diff
@@ -0,0 +1,428 @@
...
+        get_application_installs(
+            invalid_app_data, TEST_ACCOUNT_ID, TEST_API_URL, TEST_API_TOKEN
+        )
+
+    # Verify no API calls were made due to early failure
+    mock_get_paginated_results.assert_not_called()
+
+
+@patch("cartography.intel.sentinelone.application.get_paginated_results")
+def test_get_application_installs_preserves_vendor_name(mock_get_paginated_results):
```
achantavy: Nit: I don't think we need this test since the majority of the functionality is patched out, but if you feel it's useful to exercise the code path we can keep it

tests/unit/cartography/intel/sentinelone/test_application.py
```diff
@@ -0,0 +1,428 @@
...
+
+    # Call function
+    result = get_application_data(TEST_ACCOUNT_ID, TEST_API_URL, TEST_API_TOKEN)
+
+    # Verify return value
+    assert result == []
+
+
+@patch("cartography.intel.sentinelone.application.get_paginated_results")
+def test_get_application_installs(mock_get_paginated_results):
```
achantavy: Nit: I don't think we need this test since the majority of the functionality is patched out, but if you feel it's useful to exercise the code path we can keep it

achantavy (approved):
LGTM. just nits, no blockers, thank you!

cartography/models/sentinelone/application.py
serge-wq: The association happens at the `S1ApplicationVersion` level. Do you think there should be also a link at this level?

The relationships currently look like:
```
(S1Agent)-[HAS-INSTALLED]->(S1ApplicationVersion)<-[VERSION]-(S1Application)
```

achantavy: Oh ok, I'm good with either way. This is basically how s1 models their api?

cartography/models/sentinelone/application_version.py
serge-wq: Updated.

cartography/intel/sentinelone/utils.py
serge-wq: I think I'll probably remove the regexes and see how de IDs look like when testing.

cartography/models/sentinelone/application.py
serge-wq: Yes. So their API has several ways to get application information and each has different data 😬 . This is the way that made more sense to me to connect them.

Issue #1687, opened 2025-07-15
EC2Subnet “husks” created


achantavy:
**Title**: EC2Subnet "husks" created with only `subnetid` due to partial syncs and inconsistent property naming

**Description**:  
In Cartography `0.107.0rc1`, we're seeing the creation of incomplete `EC2Subnet` nodes (referred to as “husks”) that contain only the `subnetid` field and are missing the full set of expected subnet attributes. This happens when other syncs (e.g., EC2 instance or NetworkInterface) reference a subnet but do not merge it with the full subnet data from the `describe-subnets` sync. The result is that some nodes only contain `subnetid`, while others only contain `subnet_id`.

This leads to an incorrect and fragmented graph where some `EC2Subnet` nodes are not fully enriched, potentially breaking queries and causing inconsistent graph state.

**To Reproduce**:
Run the following Cypher queries on a graph built with Cartography `0.107.0rc1`:

```cypher
// Shows which syncs are creating EC2Subnet nodes with subnetid but no subnet_id
MATCH path = (s:EC2Subnet)--(r)
WHERE s.subnet_id IS NOT NULL AND s.subnetid IS NULL
RETURN DISTINCT labels(r);
```

returns

```
["AWSAccount"]
["NetworkInterface"]
```

This shows that the networkinterface sync creates these “husks”.

```cypher
// Shows which syncs are creating EC2Subnet nodes with subnet_id but no subnetid
MATCH path = (s:EC2Subnet)--(r)
WHERE s.subnetid IS NOT NULL AND s.subnet_id IS NULL
RETURN DISTINCT labels(r);
```

Returns

```
["AWSAccount"]
["AWSVpc"]
["EC2NetworkAcl"]
["LoadBalancerV2"]
["EC2RouteTableAssociation"]
["AWSTag"]
["Tag"]
["AutoScalingGroup"]
["DBSubnetGroup"]
```

This indicates inconsistent property usage across sync modules.

## Logs:
N/A — no stack trace, but evident from graph inspection.

## Screenshots:
N/A — can provide if needed.

## Please complete the following information:
	•	Cartography release version or commit hash: 0.107.0rc1

Additional context:
This is likely caused by inconsistent use of subnetid vs subnet_id across sync modules. The networkinterface sync appears to create minimal EC2Subnet nodes with subnetid, which aren’t merged with richer nodes from the EC2 subnet sync using subnet_id. This also suggests a missing merge clause or standardized property key in those syncs.

PR #1691, opened 2025-07-16
fix: subnetid vs subnet_id inconsistency (#1687)

achantavy:
Temp non-breaking change for #1687. In some cases, we refer to EC2 subnet ids with `subnetid` and `subnet_id` in others.

I first created a test to detect the inconsistency by running all possible module tests that create a node with a `subnetid` or a `subnet_id`. Before any code changes, it fails:

Before:
Screenshot 2025-07-16 at 4 19 31 PM

To fix, I made `subnetid` and `subnet_id` present on all nodes. In future we can standardize on one but this way we at least don't break any callers.

After:
Screenshot 2025-07-16 at 4 19 53 PM

- https://github.com/cartography-cncf/cartography/issues/1687

### Checklist

### DIFF
--- a/cartography/intel/aws/ec2/subnets.py
+++ b/cartography/intel/aws/ec2/subnets.py
@@ -45,15 +45,22 @@ def load_subnets(
 
     ingest_subnets = """
     UNWIND $subnets as subnet
-    MERGE (snet:EC2Subnet{subnetid: subnet.SubnetId})
+    MERGE (snet:EC2Subnet{id: subnet.SubnetId})
     ON CREATE SET snet.firstseen = timestamp()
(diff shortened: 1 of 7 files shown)


kunaals (approved):
Looks great, took me a second to understand what the test is doing but after a second read it makes sense. I think there is a discussion to be had about "husks" and objects that are created/represented in multiple modules

achantavy:
payin back that tech debt :)

the concept of a "husk" is like here's a subnet-as-known-by-an-ec2-instance vs here's a subnet as known by describe-ec2-subnets -- the first one is going to be incomplete, and the second one is going to be a lot more thorough, but it's still worth drawing the first one because we're letting each data type express what it knows about the environment. then when describe-ec2-subnets runs, we enrich the "husks". things just get problematic when these join keys are inconsistent, especially with hand-written queries.

Issue #67, opened 2019-05-06
Cartography should ingest security group ingress rules allowing traffic from other security groups.


ecdavis:

flyinbutrs:
Is this being actively worked on? If not, I might take a stab at it... very interested in having this in my graph.

achantavy:
@flyinbutrs Somehow we missed your comment here, sorry :(.

This is not being actively being worked on and we would definitely welcome contributions here!

andrewmackett:
> Is this being actively worked on? If not, I might take a stab at it... very interested in having this in my graph.

@flyinbutrs did you make any progress on this? I'd be really interested to see this added too.

flyinbutrs:
I have not worked on this @andrewmackett , my time hasn't been very free unfortunately.

achantavy:
This issue is old enough to have started first grade, but! I have a PR here: https://github.com/cartography-cncf/cartography/pull/1696

achantavy:
Added in #1696!

PR #1696, opened 2025-07-17
feat: Add ALLOWS_TRAFFIC_FROM between EC2 security groups (#67)

achantavy:
Addresses https://github.com/cartography-cncf/cartography/issues/67.

Maps AWS security groups allowing traffic from each other, including self referential ones. If sg A ALLOWS_TRAFFIC_FROM sg B, that means that (at least from a sg rules perspective) all assets in sg B will be able to reach sg A.

Screenshot 2025-07-17 at 3 05 54 PM

### Checklist

### DIFF
--- a/cartography/intel/aws/ec2/security_groups.py
+++ b/cartography/intel/aws/ec2/security_groups.py
@@ -57,14 +57,14 @@ def transform_ec2_security_group_data(
     ranges: List[Dict[str, Any]] = []
 
     for group in data:
-        groups.append(
-            {
-                "GroupId": group["GroupId"],
-                "GroupName": group.get("GroupName"),
-                "Description": group.get("Description"),
-                "VpcId": group.get("VpcId"),
-            },
-        )
+        group_record = {
+            "GroupId": group["GroupId"],
+            "GroupName": group.get("GroupName"),
+            "Description": group.get("Description"),
+            "VpcId": group.get("VpcId"),
+        }
+        # Collect referenced security groups for relationship loading
+        source_group_ids: set[str] = set()
 
         for rule_type, target in (
             ("IpPermissions", inbound_rules),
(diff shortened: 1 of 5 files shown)


tests/data/aws/ec2/security_groups.py
```diff
@@ -215,4 +215,57 @@
...
+                "IpProtocol": "tcp",
+                "IpRanges": [
+                    {
+                        "CidrIp": "10.0.0.0/8",
+                    },
+                ],
+                "Ipv6Ranges": [],
+                "PrefixListIds": [],
+                "ToPort": 22,
+                "UserIdGroupPairs": [
```
achantavy, line 233: Added this as a test sample for cross-sg

kunaals (approved):
Lgtm, do we want to support destination groups for egress as well in this PR? Or do that in a separate one?

Also I like the fact that we call the rel ALLOWS TRAFFIC FROM, it clarifies that sg A allows traffic from sg B, but doesn't necessarily say that the two groups can communicate. After the dest groups are in here, it would be nice to have a rel that says 2 SGs can communicate and on which ports. Much easier than reasoning through all the IpPerm nodes. Might be best as analysis job? Open to feedback on this idea - just spitballing here

achantavy:
Do you know what data field represents the destination group for egress? I don't know the concepts here very well. This PR goes over https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_UserIdGroupPair.html, which is taken from https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IpPermission.html (which was taken from https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SecurityGroup.html)

For now this PR is mostly just getting a base representation for allowing traffic from one group to another. Other enhancements can build off this guy

cartography/intel/aws/ec2/security_groups.py
```diff
@@ -88,6 +88,13 @@ def transform_ec2_security_group_data(
                 )
                 for ip_range in rule.get("IpRanges", []):
                     ranges.append({"RangeId": ip_range["CidrIp"], "RuleId": rule_id})
+                for pair in rule.get("UserIdGroupPairs", []):
+                    sg_id = pair.get("GroupId")
```
shyammukund, line 92: if theres no group id wont this fail? maybe worth it to log a skip warning

shyammukund: I guess you could say that for anything that uses .get() though

achantavy: it's an optional field. if there's no group id then sg_id will be None. no need to log because there will be tons of security groups that don't have this field

kunaals:
Shoot I missed this for loop:
```
        for rule_type, target in (
            ("IpPermissions", inbound_rules),
            ("IpPermissionsEgress", egress_rules),
        ):
```

In the case of IpPermissions, the security group is the source. In the case of IpPermissionsEgress, it's the destination. That's a slight oversight in this PR. I need to play around with this a bit more to have a complete understanding

PR #1697, opened 2025-07-18
Refactor RDSInstance to new data model

achantavy:
## Summary
- add RDSCluster and RDSSnapshot data models
- load RDS clusters and snapshots with the declarative loader
- update RDS ingestion functions to use new schemas

## Testing
- `pre-commit run --files cartography/intel/aws/rds.py cartography/models/aws/rds/cluster.py cartography/models/aws/rds/snapshot.py cartography/models/aws/rds/instance.py`
- `pytest tests/integration/cartography/intel/aws/test_rds.py -k test_load_rds_instances_basic -q` *(fails: ModuleNotFoundError: No module named 'backoff')*

------
https://chatgpt.com/codex/tasks/task_b_68787cc16a2c832f9dd387f7c6ac5667


cartography/models/aws/rds/subnet_group.py
```diff
@@ -0,0 +1,101 @@
...
+
+    label: str = "DBSubnetGroup"
+    properties: DBSubnetGroupNodeProperties = DBSubnetGroupNodeProperties()
+    sub_resource_relationship: DBSubnetGroupToAWSAccountRel = (
+        DBSubnetGroupToAWSAccountRel()
+    )
+    other_relationships: OtherRelationships = OtherRelationships(
+        [
+            DBSubnetGroupToRDSInstanceRel(),
+            DBSubnetGroupToEC2SubnetRel(),
```
achantavy, line 99: Previously we [created EC2Subnet "husks"](https://github.com/cartography-cncf/cartography/blob/6d876d6ef558caa055d39e7908a1c33bedf2acfd/cartography/intel/aws/rds.py#L431) where we `MERGE`'d in new `EC2Subnet` nodes because the RDS API told us their `id`s.

Now, by defining a relationship object this way, we are doing `MATCH` on existing subnets.

In order to preserve the previous behavior, we would have to define a new `EC2SubnetRDSInstanceSchema` object to represent a "EC2Subnet has known by an RDSInstance". Given that the RDS API does not give us any more information about an EC2 subnet other than its ID, this is a lot of complexity for not much gain.

I'm going to write a separate PR to document this right now, but I think the rule of thumb is to 
- if a given data type A refers to another data type B and offers fields about B that B does it have itself (like with an [EBSVolumeInstance](https://github.com/cartography-cncf/cartography/blob/6d876d6ef558caa055d39e7908a1c33bedf2acfd/cartography/models/aws/ec2/volumes.py#L107-L117)), then we should define a composite node. It would be called an "`BASchema`" to denote that it's a "`B`" object as known by an "`A`" object (and in effect perform a `merge`.
- Otherwise if data type A only refers to data type B by an id, then there is no need to define a composite node. Instead we can just define a relationship schema here (and in effect perform a `match` instead of a `merge`.

So, this isn't strictly a refactor because this is a functional change, but I think it's worth doing.

tests/integration/cartography/intel/aws/test_rds.py
```diff
@@ -1,16 +1,63 @@
...
+            MERGE (subnet:EC2Subnet{subnetid: $subnet_id})
+            ON CREATE SET subnet.firstseen = timestamp()
+            SET subnet.lastupdated = $update_tag
+            """,
+            subnet_id=subnet_id,
+            update_tag=TEST_UPDATE_TAG,
+        )
+
+
 def test_load_rds_clusters_basic(neo4j_session):
```
achantavy, line 52: I made sure that the existing tests keep working. Just needed to add a transform

tests/integration/cartography/intel/aws/test_rds.py
```diff
@@ -143,3 +206,177 @@ def test_load_rds_snapshots_basic(neo4j_session):
...
+    cartography.intel.aws.rds,
+    "get_rds_instance_data",
+    return_value=DESCRIBE_DBINSTANCES_RESPONSE["DBInstances"],
+)
+@patch.object(
+    cartography.intel.aws.rds,
+    "get_rds_cluster_data",
+    return_value=DESCRIBE_DBCLUSTERS_RESPONSE["DBClusters"],
+)
+def test_sync_rds_comprehensive(
```
achantavy, line 226: This test covers the entire sync function instead of just the load with the previous basic tests

cartography/models/aws/rds/cluster.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    scaling_configuration_info_max_capacity: PropertyRef = PropertyRef(
+        "ScalingConfigurationInfoMaxCapacity"
+    )
+    scaling_configuration_info_auto_pause: PropertyRef = PropertyRef(
+        "ScalingConfigurationInfoAutoPause"
+    )
+    deletion_protection: PropertyRef = PropertyRef("DeletionProtection")
+    region: PropertyRef = PropertyRef("Region", set_in_kwargs=True)
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
```
shyammukund, line 64: does the graph need all these properties. If not maybe better to take out whats not necessary to have less data the graph needs to load.

cartography/models/aws/rds/instance.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+    backup_retention_period: PropertyRef = PropertyRef("BackupRetentionPeriod")
+    endpoint_address: PropertyRef = PropertyRef("EndpointAddress")
+    endpoint_hostedzoneid: PropertyRef = PropertyRef("EndpointHostedZoneId")
+    endpoint_port: PropertyRef = PropertyRef("EndpointPort")
+    iam_database_authentication_enabled: PropertyRef = PropertyRef(
+        "IAMDatabaseAuthenticationEnabled"
+    )
+    auto_minor_version_upgrade: PropertyRef = PropertyRef("AutoMinorVersionUpgrade")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
```
shyammukund, line 60: same thing about the properties

cartography/models/aws/rds/cluster.py
achantavy: yeah, the RDS instance has [lots of fields](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBInstance.html). Existing code has this https://github.com/cartography-cncf/cartography/blob/6d876d6ef558caa055d39e7908a1c33bedf2acfd/cartography/intel/aws/rds.py#L55-L90. This PR keeps that existing functionality without removing fields

cartography/models/aws/rds/subnet_group.py
achantavy: Here's that PR https://github.com/cartography-cncf/cartography/pull/1698

cartography/models/aws/rds/cluster.py
```diff
@@ -0,0 +1,89 @@
...
+    db_cluster_resource_id: PropertyRef = PropertyRef("DbClusterResourceId")
+    clone_group_id: PropertyRef = PropertyRef("CloneGroupId")
+    cluster_create_time: PropertyRef = PropertyRef("ClusterCreateTime")
+    earliest_backtrack_time: PropertyRef = PropertyRef("EarliestBacktrackTime")
+    backtrack_window: PropertyRef = PropertyRef("BacktrackWindow")
+    backtrack_consumed_change_records: PropertyRef = PropertyRef(
+        "BacktrackConsumedChangeRecords"
+    )
+    capacity: PropertyRef = PropertyRef("Capacity")
+    engine_mode: PropertyRef = PropertyRef("EngineMode")
```
kunaals: nit: move engine_mode under engine_version to keep similar fields close to each other

cartography/models/aws/rds/snapshot.py
```diff
@@ -0,0 +1,108 @@
...
+    snapshot_create_time: PropertyRef = PropertyRef("SnapshotCreateTime")
+    engine: PropertyRef = PropertyRef("Engine")
+    allocated_storage: PropertyRef = PropertyRef("AllocatedStorage")
+    status: PropertyRef = PropertyRef("Status")
+    port: PropertyRef = PropertyRef("Port")
+    availability_zone: PropertyRef = PropertyRef("AvailabilityZone")
+    vpc_id: PropertyRef = PropertyRef("VpcId")
+    instance_create_time: PropertyRef = PropertyRef("InstanceCreateTime")
+    master_username: PropertyRef = PropertyRef("MasterUsername")
+    engine_version: PropertyRef = PropertyRef("EngineVersion")
```
kunaals: nit again: would be good to put this line with engine_version on line 24, underneath engine

kunaals (approved):
Looks great. Left couple nits

Issue #1668, opened 2025-07-02
migrate modules to the `ORM` data model


jychp:
*Description*:                                                                                                                                                                                   
  We're progressively migrating Cartography modules to the modern declarative data model using `CartographyNodeSchema` and `CartographyRelSchema`. This improves consistency, makes relationships  
  explicit, and enables scoped cleanup.                                                                                                                                                            
                                                                                                                                                                                                   
  If you're looking for a good first contribution, pick an existing module and help convert it.                                                                                                    
                                                                                                                                                                                                   
  *Checklist*                                                                                                                                                                                                                                                                                                                                                                     
  - [x] Jamf                                                                                                                                                                                       
  - [x] OCI                                                                                                                                                                                        
  - [ ] GitHub (in progress by @jychp)                                                                                                                                                             
  - [ ] Okta (see https://github.com/cartography-cncf/cartography/pull/1585)                                                                                                                       
  - [x] PagerDuty (see https://github.com/cartography-cncf/cartography/pull/1606)                                                                                                                                                                                                                 
                                                                                                                                                                                                   
  *Checklist for AWS*                                                                                                                                                                                                                                                                                                                    
  - [x] elasticsearch: ESDomain                                                                                                                                                                                                                                                                                               
  - [x] resourcegroupstaggingapi: AWSTag                                                                                                                                                           
  - [x] s3: S3Acl / S3Bucket / S3PolicyStatement                                                                                                                                                                                                                                                                                                       
  - [x] elastic_ip_addresses: ElasticIPAddress                                                                                                                                                     
  - [x] network_interfaces: NetworkInterface (models exist, intel not refactored)                                                                                                                  
  - [x] internet_gateways: AWSInternetGateway                                                                                                                                                      
  - [x] load_balancer_v2s: ELBV2Listener                                                                                                                                                           
  - [x] reserved_instances: EC2ReservedInstance

heryxpc:
Does the `AWSAccount` node needs migration as well?

jychp:
@heryxpc yes
The cleanup thing will not be used, but for consistency (and auto indexing) it's better.

And I'm trying to add more unit tests on the model itself, so definitely yes :)

PR #1700, opened 2025-07-19
chore: Refactor ECR to data model (#1668)

achantavy:
Refactors ECR to the new data model

#1668

### Checklist

image

Screenshot 2025-07-18 at 10 20 39 PM

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -57,6 +57,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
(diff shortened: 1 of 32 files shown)


achantavy:
@heryxpc @serge-wq -- FYI this PR breaks `cartography.intel.aws.ecr.load_ecr_repository_images()` as we now have to add in the account id as a param.

I don't think this should break anything with your workflow but just double check if you have a shim somewhere for ECR that calls `load_*()` instead of `sync_()`. The `sync()` func will be unaffected.

serge-wq (approved):
Looks good. Thanks for the heads up, will make sure to test before integrating.

Issue #1669, opened 2025-07-02
route53: AWSDNSRecord / AWSDNSZone / NameServer


heryxpc:

achantavy:
hey @heryxpc what did you have in mind here?

achantavy:
oh i see, this was specifically for the migration task

achantavy:
Fixed in https://github.com/cartography-cncf/cartography/pull/1703

Issue #701, opened 2021-10-06
DNSRecord/DNSZone/Nameserver are not tied to AWSAccount


marco-lancini:
The AWS resources of [DNSRecord](https://github.com/lyft/cartography/blob/master/docs/schema/aws.md#dnsrecord), [DNSZone](https://github.com/lyft/cartography/blob/master/docs/schema/aws.md#dnszone), and [Nameserver](https://github.com/lyft/cartography/blob/master/docs/schema/aws.md#nameserver) are not tied to an AWSAccount (i.e., the relationship `(AWSAccount)-[RESOURCE]->(x)` is missing).

This deviates from the standard convention (since all other resources can be accessed by an `AWSAccount`), and it prevents from associating these resources to the account they reside in

achantavy:
I agree this should be fixed. Our new data model has a convention where all nodes will tie back to their sub resource, so once we refactor it all then this issue will be fixed.

achantavy:
DNS ones addressed in https://github.com/cartography-cncf/cartography/pull/1703.

Still waiting for the AWSPolicy one to be addressed in https://github.com/cartography-cncf/cartography/pull/1710

PR #1703, opened 2025-07-20
chore: Refactor route53 to data model (#701). fix: incorrect NS-to-A record connections

achantavy:
- Refactors route53 to data model, keeping most existing tests intact and passing.
- Adds additional tests to cover the full sync.
- [Fixes a bug](https://github.com/cartography-cncf/cartography/pull/1703/commits/f0830b3964a68fb9def87af89e8b325968664fba) where we incorrectly drew edges from NS records to A records.

### Methodology
- Before starting the refactor at all, I first created a test to cover the sync() function
- Then I made sure the sync() test passed
- Then I continued on, making sure that the new test and all the old ones worked, making minimal changes to the old interface as possible

#1668 
#1669 
#701

### Checklist

image

image


cartography/models/aws/route53/dnsrecord.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@dataclass(frozen=True)
+class AWSDNSRecordToIpRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class AWSDNSRecordToIpRel(CartographyRelSchema):
+    target_node_label: str = "Ip"
```
achantavy, line 185: This node does not exist in the graph yet but this will be a great addition

tests/integration/cartography/intel/aws/test_route53.py
```diff
@@ -1,5 +1,6 @@
 import cartography.intel.aws.route53
```
achantavy, line 1: This is the existing integ test file. Most tests still work without needing to change interface.

tests/integration/cartography/intel/aws/test_route53_sync.py
```diff
@@ -0,0 +1,371 @@
+from unittest.mock import MagicMock
```
achantavy, line 1: this is a new set of tests that cover the whole sync() function

cartography/intel/aws/route53.py
```diff
@@ -295,27 +114,28 @@ def transform_record_set(record_set: Dict, zone_id: str, name: str) -> Optional[
         else:
             # this is a real A record
             # loop and add each value (IP address) to a comma separated string
-            # don't forget to trim that trailing comma!
-            # TODO can this be replaced with a string join?
-            value = ""
-            for a_value in record_set["ResourceRecords"]:
-                value = value + a_value["Value"] + ","
```
achantavy, line 302: Yeah so this doesn't make any sense but I will leave that for a follow-up PR since I've already gone a little bit past a true refactor.

The `value` field is used in analysis jobs to help determine what DNS records point to. Oftentimes this value is queried for without any parsing, so if this happens to be a comma separated list, those queries will always fail.

The fix will be to just remove the `value` field altogether and then draw from the node to the IP addresses/other records like a graph

shyammukund (approved):

PR #1706, opened 2025-07-21
chore: Refactor AWS Secrets Manager Module

shyammukund:
- Gets rid of legacy raw querying in intel module
- Creates SecretsManagerSecretSchema data model
- Implements transform function to convert api data to schema format
- Replaces manual cleanup 

### Checklist

Screenshot 2025-07-21 at 3 51 32 PM


cartography/models/aws/secretsmanager/__init__.py
```diff
@@ -0,0 +1,9 @@
+from cartography.models.aws.secretsmanager.secret import SecretsManagerSecretSchema
+from cartography.models.aws.secretsmanager.secret_version import (
+    SecretsManagerSecretVersionSchema,
+)
+
+__all__ = [
+    "SecretsManagerSecretSchema",
+    "SecretsManagerSecretVersionSchema",
+]
```
achantavy: Remove everything from this file, it's not needed. idk why cursor generates it, i get the same thing too when i use it

achantavy (approved):
Looks great, thank you! Just remove the __all__ thing and good to go

cartography/models/aws/secretsmanager/__init__.py
shyammukund: Do we still need to keep the empty file? I noticed a lot of other models have that empty init.py file

achantavy: Yeah it's a python convention: https://docs.python.org/3/reference/import.html#regular-packages

I don't quite understand it tbh but without it then imports fail

PR #1707, opened 2025-07-22
doc: add refererence (autodoc from docstring)

jychp:


cartography/graph/statement.py
```diff
@@ -30,13 +39,73 @@ def default(self, obj):
 
 # TODO move this cartography.util after we move util.run_*_job to cartography.graph.job.
 def get_job_shortname(file_path: Union[Path, str]) -> str:
+    """
+    Extract the short name from a file path by removing the path and extension.
```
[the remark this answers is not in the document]
jychp: @achantavy I think it's an actual bug isn't it ?

This function will indeed only remove the ext, and not the path

cartography/cli.py
```diff
@@ -758,10 +781,19 @@ def _build_parser(self):
...
+        Main entrypoint for the command line interface.
 
-        :type argv: string
-        :param argv: The parameters supplied to the command line program.
+        This method parses command line arguments, configures logging and various
+        service connections, validates input parameters, and executes the cartography
+        sync operation with the provided configuration.
+
+        Args:
+            argv: The command line arguments to parse. Should be a list of strings
```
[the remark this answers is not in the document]
jychp: It's a serialized list of strings, it's OK.

cartography/graph/querybuilder.py
```diff
@@ -321,25 +321,51 @@ def _build_node_properties_statement(
...
-        ```
-        i.id = item.Id,
-        i.node_prop_1 = item.Prop1,
-        i.node_prop_2 = $Prop2
-        ```
-    where `i` is a reference to the Neo4j node.
-    :param node_property_map: Mapping of node attribute names as str to PropertyRef objects
-    :param extra_node_labels: Optional ExtraNodeLabels object to set on the node as string
-    :return: The resulting Neo4j SET clause to set the given attributes on the node
+    Returns:
```
kunaals: duplicate `Returns:` and `Examples:` sections - this block and the one below repeat what's already above. remove lines 350-370

cartography/graph/querybuilder.py
```diff
@@ -321,25 +321,51 @@ def _build_node_properties_statement(
...
         set_clause: str = _build_node_properties_statement(node_property_map)
 
-    the returned set_clause will be
-        ```
-        i.id = item.Id,
-        i.node_prop_1 = item.Prop1,
-        i.node_prop_2 = $Prop2
-        ```
-    where `i` is a reference to the Neo4j node.
-    :param node_property_map: Mapping of node attribute names as str to PropertyRef objects
```
kunaals, line 340: same issue here - duplicate `Returns:` and `Examples:` sections. remove the repeated block

cartography/client/core/tx.py
```diff
@@ -367,23 +367,31 @@ def read_list_of_values_tx(
...
+            Not supported: ``MATCH (a:TestNode) RETURN a.name, a.age, a.x``
+        **kwargs: Additional keyword arguments passed to ``tx.run()``.
+
+    Returns:
+        List[Union[str, int]]: A list of string or integer values from the query results.
 
-        values = neo4j_session.execute_read(read_list_of_values_tx, query)
+    Examples:
+        >>> query = "MATCH (a:TestNode) RETURN a.name ORDER BY a.name"
+        >>> values = neo4j_session.read_transaction(read_list_of_values_tx, query)
```
kunaals: the old docstring used `execute_read()` which is the neo4j v5 api. `read_transaction()` is deprecated - should keep using `execute_read` in examples

docs/conf.py
```diff
@@ -15,18 +15,13 @@
 import sys
 from datetime import datetime
 
-sys.path.insert(0, os.path.abspath("../.."))
+sys.path.insert(0, os.path.abspath(".."))
```
kunaals: this breaks autodoc. `build.sh` copies conf.py to `generated/rst/`, so `os.path.abspath("..")` resolves to `generated/` not the repo root.

either revert to `"../.."` or use `__file__` for robustness:
```python
sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```

kunaals (commented):
thanks for this!

**duplicate docstring sections**
- `_build_node_properties_statement()` and `_build_rel_properties_statement()` both have duplicate `Returns:` and `Examples:` blocks - looks like a copy-paste artifact

**deprecated neo4j api in examples**
- the docstring examples use `read_transaction()` which is deprecated in neo4j v5. the old docstrings correctly used `execute_read()` - this change regresses to the deprecated api

**sys.path will break autodoc**
- changing from `"../.."` to `".."` breaks the path resolution. `build.sh` copies `conf.py` to `generated/rst/`, so from there `".."` resolves to `generated/` not the repo root. autodoc will fail with `ModuleNotFoundError: cartography` on clean checkouts

PR #1709, opened 2025-07-23
chore: refactor AWS Key Management System Module

shyammukund:
### Checklist

-  Graph showing rels 

Screenshot 2025-07-22 at 6 00 19 PM


cartography/intel/aws/kms.py
```diff
@@ -120,138 +125,133 @@ def get_grants(key: Dict, client: botocore.client.BaseClient) -> List[Any]:
...
-    )
+        # Convert datetime fields to epoch timestamps
+        transformed["CreationDate"] = dict_date_to_epoch(alias, "CreationDate")
+        transformed["LastUpdatedDate"] = dict_date_to_epoch(alias, "LastUpdatedDate")
+
+        transformed_data.append(transformed)
+    return transformed_data
 
 
 @timeit
```
achantavy: No need a timeit on a transform because it doesn't go over the network or do any other IO

cartography/intel/aws/kms.py
```diff
@@ -120,138 +125,133 @@ def get_grants(key: Dict, client: botocore.client.BaseClient) -> List[Any]:
...
```
achantavy, line 125: No need the timeit on a transform

cartography/intel/aws/kms.py
```diff
@@ -120,138 +125,133 @@ def get_grants(key: Dict, client: botocore.client.BaseClient) -> List[Any]:
...
+
+        if parsed_policy is not None:
+            # Policy exists and is internet accessible
+            policy_data[key_id] = {
+                "anonymous_access": parsed_policy["internet_accessible"],
+                "anonymous_actions": parsed_policy["accessible_actions"],
+            }
+        else:
+            # Set defaults for keys without policies or non-internet-accessible policies
+            policy_data[key_id] = {"anonymous_access": False, "anonymous_actions": []}
```
achantavy: In this else clause, `parsed_policy` is None. What does this mean? Does it mean that we could not determine the permission, or does it mean that access to the key is locked down?

If the first one, we should instead set anonymous_access to None to indicate that we don't know the answer. It'd be incorrect to assume that a resource is inaccessible by default

cartography/intel/aws/kms.py
```diff
@@ -120,138 +125,133 @@ def get_grants(key: Dict, client: botocore.client.BaseClient) -> List[Any]:
...
 
 
 @timeit
-def _load_kms_key_grants(
-    neo4j_session: neo4j.Session,
-    grants_list: List[Dict],
-    update_tag: int,
-) -> None:
+def transform_kms_keys(
+    keys: List[Dict], policy_data: Optional[Dict[str, Dict]] = None
```
achantavy: Why is `policy_data` optional? We get that value from `transform_kms_key_policies()`, and that never returns `None`, so it shouldn't need to be optional.

Having that be mandatory will simplify the code too because we won't need to have the check below for `if policy_data`.

cartography/intel/aws/kms.py
```diff
@@ -120,138 +125,133 @@ def get_grants(key: Dict, client: botocore.client.BaseClient) -> List[Any]:
...
-    MATCH (kmskey:KMSKey{id: grant.KeyId})
-    MERGE (g)-[r:APPLIED_ON]->(kmskey)
-    ON CREATE SET r.firstseen = timestamp()
-    SET r.lastupdated = $UpdateTag
+    Transform AWS KMS Keys to match the data model.
+    Converts datetime fields to epoch timestamps for consistency.
+    Optionally includes policy analysis properties.
     """
+    transformed_data = []
+    for key in keys:
```
achantavy, line 150: Nit: the word key is confusing since we're parsing over a dictionary with key-value pairs and the underlying data is KMSkeys. I don't have a great solution but just wanted to point this out. Non-blocking

cartography/intel/aws/kms.py
```diff
@@ -120,138 +125,133 @@ def get_grants(key: Dict, client: botocore.client.BaseClient) -> List[Any]:
...
+        transformed["DeletionDate"] = dict_date_to_epoch(key, "DeletionDate")
+        transformed["ValidTo"] = dict_date_to_epoch(key, "ValidTo")
+
+        # Add policy analysis if provided
+        if policy_data and key["KeyId"] in policy_data:
+            transformed.update(policy_data[key["KeyId"]])
+        else:
+            # Set defaults for policy analysis properties
+            transformed["anonymous_access"] = False
+            transformed["anonymous_actions"] = []
```
achantavy: Here I'd prefer to be explicit.

```python
perms = key["KeyId"]  # using a direct access here because if the data doesn't have a keyId then that's a big big problem and we should fail
transformed["anonymous_access"] = policy_data[perms].get("anonymous_access")  # using a `.get()` here because (now with my other suggestion) it's possible for anonymous_access to be None. This indicates that we don't know what the permission is.
transformed["anonymous_actions"] = policy_data[perms].get("anonymous_actions")  # same thing here. Instead of empty list, we have a None to indicate that we don't know. 
```

The end result if we cannot determine anonymous_actions for a given KMS key  is that the `KMSKey` node  will not have that field.

cartography/intel/aws/kms.py
```diff
@@ -319,50 +319,45 @@ def parse_policy(key: str, policy: Policy) -> Optional[Dict[Any, Any]]:
...
-    run_cleanup_job("aws_import_kms_cleanup.json", neo4j_session, common_job_parameters)
+    """
+    Run KMS cleanup using schema-based GraphJobs for all node types.
+    """
+    logger.debug("Running KMS cleanup using GraphJob for all node types")
+
+    # Clean up grants first (they depend on keys)
+    GraphJob.from_node_schema(KMSGrantSchema(), common_job_parameters).run(
+        neo4j_session
+    )
```
achantavy, line 287: Does this work without failures? The KMSGrant doesn't have `scoped_cleanup` ([ref](https://github.com/cartography-cncf/cartography/blob/2f49032775da1eddc7c25ec046f7d098f1cc16cb/cartography/models/core/nodes.py#L108)) set to False so without a sub-resource it's not supposed to actually clean up KMSGrants.

I do think it makes sense to introduce a link to the AWS Account from the KMS grant

achantavy (changes requested):

cartography/intel/aws/kms.py
shyammukund: No explicit failure but looking back at it i think it is likely not cleaning up the grants that are in the AWS Account being synced. I think it might be cleaning up all Grants in the database globally or not at all. Seeing if i can figure out if its doing anything as it is currently What would the next step be here to make sure KMSGrant can use the schema based cleanup?

cartography/intel/aws/kms.py
shyammukund: Here, Since I made it so there is always policy_data has a KeyID it should be fine to just update the dict with the KeyID since it will always exist right

cartography/intel/aws/kms.py
shyammukund: done

cartography/intel/aws/kms.py
shyammukund: yea that else block was hitting when its inaccessible AND when the key was locked down. added an elif to differentiate between when policy exists but not accessible and the else is now when the policy cant be determined.

achantavy: Let's clean this up and make it easier to understand. I realize that the code you're working off of isn't the most understandable to begin with.

Right now we're starting an if-else check based on `if parsed_policy is not None`, and then it moves on to `elif policy is not None` -- this is hard to follow because I would have expected the `elif` to say something about `parsed_policy`. I'd recommend reworking this.

First, reading over the original `parse_policy()` and `load_kms_key_details()`, it looks like we were ignoring the returned result of `parse_policy()` if it was `None`. In the past, this worked only because it was only used to determine whether to send a piece of data to the graph, but we should be more explicit here for readability.

So, I think we should change `parse_policy` to be like

```python
def parse_policy(key: str, policy: Policy) -> dict[str, Any]:
    policy = Policy(json.loads(policy['Policy']))
    inet_actions = policy.internet_accessible_actions()    

    return {
        'kms_key': key,
        'internet_accessible': policy.is_internet_accessible(),
        'accessible_actions': list(inet_actions) if inet_actions else None,
    }
```

Then, you don't have to do any None checking in `transform_kms_key_policies()` and then you can simplify that to just

```python
def transform_kms_key_policies(
    policy_alias_grants_data: list[tuple],
) -> dict[str, dict[str, Any]]:
    policy_data = {}
    for key_id, policy, *_ in policy_alias_grants_data:
        parsed_policy = parse_policy(key_id, policy)
        policy_data[key_id] = parsed_policy
    return policy_data
```

tests/data/aws/kms.py
```diff
@@ -68,7 +68,7 @@
 
 DESCRIBE_GRANTS = [
     {
-        "KeyId": "9a1ad414-6e3b-47ce-8366-6b8f26ba467d",
+        "KeyId": "arn:aws:kms:eu-west-1:000000000000:key/9a1ad414-6e3b-47ce-8366-6b8f26ba467d",
```
achantavy, line 71: Does the API return a data shape that looks like this?

If not, I'm concerned that the tests will not reflect real conditions

achantavy (approved):

PR #1710, opened 2025-07-23
chore: Refactor AWS IAM

achantavy:
Refactors AWS IAM

### Checklist


cartography/models/aws/iam/access_key.py
```diff
@@ -0,0 +1,103 @@
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: AccountAccessKeyToAWSAccountRelProperties = (
+        AccountAccessKeyToAWSAccountRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class AccountAccessKeySchema(CartographyNodeSchema):
+    label: str = "AccountAccessKey"
```
kunaals, line 93: I know this is in draft so I wont leave a full review yet, but I do think the term "account access key" may be a bit misleading. My understanding is that these are all just "access keys." However, if the access key is owned by root, then it effectively gives access to the entire account, which I guess could be considered an "account access key." I think most of these objects should just be access keys: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html

This might feel nit picky but I think vocab is important here given the slew of auth mechanism and key types in AWS

kunaals: Somewhat related to this TODO from intel/aws/iam.py https://github.com/cartography-cncf/cartography/blob/ba7e3767d23652c3b8096389c5d43896ea3a7608/cartography/intel/aws/iam.py#L561

cartography/models/aws/iam/access_key.py
```diff
@@ -0,0 +1,103 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class AccountAccessKeyNodeProperties(CartographyNodeProperties):
```
GeorgeGithiri5, line 15: It would be great to add comment on the classes

GeorgeGithiri5 (dismissed):

uv.lock
jychp, line 1: I think we should update dependencies in a separate PR.

cartography/models/aws/iam/group_membership.py
jychp, line 1: Why did you choose to define this relation in a separate file ? This is not the common pattern.

cartography/models/aws/iam/account_role.py
```diff
@@ -0,0 +1,24 @@
...
+class AWSAccountAWSRoleNodeProperties(CartographyNodeProperties):
+    # Required unique identifier
+    id: PropertyRef = PropertyRef("id")
+
+    # Automatic fields (set by cartography)
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class AWSAccountAWSRoleSchema(CartographyNodeSchema):
```
jychp, line 18: The name of this object is misleading.

Is it really necessary to define a separate `AWSAccount` object here, distinct from the one already defined (or expected to be defined) for the entire AWS module ?

Isn’t it risky to define a parallel object for something as critical as `AWSAccount`? Running cleanup against this object, with a potentially confusing name, could inadvertently delete core data.

If the concern is that this schema only handles a subset of fields compared to the main `AWSAccount` node, a name like `AWSAccountPartialSchema` would make this distinction clearer and reduce the risk of accidental misuse.

cartography/intel/aws/iam.py
```diff
@@ -280,223 +341,255 @@ def get_account_access_key_data(
...
-    ON CREATE SET unode:AWSPrincipal, unode.userid = $USERID, unode.firstseen = timestamp(),
-    unode.createdate = $CREATE_DATE
-    SET unode.name = $USERNAME, unode.path = $PATH, unode.passwordlastused = $PASSWORD_LASTUSED,
-    unode.lastupdated = $aws_update_tag
-    WITH unode
-    MATCH (aa:AWSAccount{id: $AWS_ACCOUNT_ID})
-    MERGE (aa)-[r:RESOURCE]->(unode)
-    ON CREATE SET r.firstseen = timestamp()
-    SET r.lastupdated = $aws_update_tag
+def get_group_memberships(
```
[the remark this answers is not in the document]
jychp: Not sure it's necessary here as the users should be agnostic of the region.

(this was an answer to CUBIC comment on missing aws_handle_region decorator)

cartography/intel/aws/iam.py
```diff
@@ -280,223 +341,255 @@ def get_account_access_key_data(
...
+    principal_arn: str,
+) -> Dict:
+    get_policy_query = """
+    MATCH
+    (principal:AWSPrincipal{arn:$Arn})-[:POLICY]->
+    (policy:AWSPolicy)-[:STATEMENT]->
+    (statements:AWSPolicyStatement)
+    RETURN
+    DISTINCT policy.id AS policy_id,
+    COLLECT(DISTINCT statements) AS statements
```
jychp, line 385: Why do we need COLLECT() with a subquery here and not a simple DISTINCT ?

cartography/intel/aws/iam.py
```diff
@@ -280,223 +341,255 @@ def get_account_access_key_data(
...
+def transform_users(users: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    user_data = []
+    for user in users:
+        user_record = {
+            "arn": user["Arn"],
+            "userid": user["UserId"],
+            "name": user["UserName"],
+            "path": user["Path"],
+            "createdate": str(user["CreateDate"]),
+            "passwordlastused": str(user.get("PasswordLastUsed", "")),
```
jychp, line 405: Are you sure you want an empty string here and not a None value ?

cartography/intel/aws/iam.py
```diff
@@ -280,223 +341,255 @@ def get_account_access_key_data(
...
+) -> list[dict[str, Any]]:
+    access_key_data = []
+    for user_arn, access_keys in user_access_keys.items():
+        for access_key in access_keys:
+            if access_key.get("AccessKeyId"):
+                access_key_record = {
+                    "accesskeyid": access_key["AccessKeyId"],
+                    "createdate": str(access_key["CreateDate"]),
+                    "status": access_key["Status"],
+                    "lastuseddate": str(access_key.get("LastUsedDate", "")),
```
jychp, line 441: why are you serializing the date here ? having datetime in the DB seems better than its string representation

cartography/intel/aws/iam.py
```diff
@@ -280,223 +341,255 @@ def get_account_access_key_data(
...
+                    "lastusedservice": access_key.get("LastUsedService", ""),
```
jychp, line 442: I think we prefer a None value instead of an empty string, that will prevent storing empty fields

cartography/intel/aws/iam.py
```diff
@@ -280,223 +341,255 @@ def get_account_access_key_data(
...
+                else:
+                    # This should not happen but who knows.
+                    logger.warning(f"Unknown principal type: {principal_type}")
+
+        role_record = {
+            "arn": role["Arn"],
+            "roleid": role["RoleId"],
+            "name": role["RoleName"],
+            "path": role["Path"],
+            "createdate": str(role["CreateDate"]),
```
jychp, line 519: Why are you serializing the date here ?

cartography/intel/aws/iam.py
```diff
@@ -514,81 +607,46 @@ def sync_assumerole_relationships(
...
-                    CreateDate=str(key["CreateDate"]),
-                    Status=key["Status"],
-                    LastUsedDate=key["LastUsedDate"],
-                    LastUsedService=key["LastUsedService"],
-                    LastUsedRegion=key["LastUsedRegion"],
-                    aws_update_tag=aws_update_tag,
-                )
+    GraphJob.from_matchlink(
+        STSAssumeRoleAllowMatchLink(),
+        "AWSAccount",  # _sub_resource_label
```
jychp: here you should use kwargs instead or args + comment ;)

cartography/intel/aws/iam.py
```diff
@@ -597,304 +655,460 @@ def ensure_list(obj: Any) -> List[Any]:
...
-        SET r.lastupdated = $aws_update_tag
-        """
-    neo4j_session.run(
-        ingest_policy_statement,
-        PolicyId=policy_id,
-        PolicyName=policy_name,
-        Statements=statements,
-        aws_update_tag=aws_update_tag,
-    ).consume()
+    for policy_id, statements in policy_statements.items():
```
jychp, line 816: Can we add the transform to add the policy_id in the data dict instead of kwargs ?
That will allow to remove that loop.

cartography/models/aws/iam/access_key.py
jychp: Do not have a strong opinion. but at least we should have a common extra label "AWSAccessKey" on both "AWSUserAccessKey" and "AWSAccountAccessKey?

cartography/intel/aws/iam.py
achantavy: No idea, I was keeping the existing code from before and just moving it to a better order. We can fix later on.

cartography/intel/aws/iam.py
achantavy: Trying to keep this as a refactor even if it's wrong. I'd like to change this in a follow-up

cartography/intel/aws/iam.py
achantavy: Agree, will follow up. For now keeping it as close to a true refactor as possible (even though I'm breaking other interfaces )

cartography/intel/aws/iam.py
achantavy: Agree, same reason as elsewhere though

cartography/intel/aws/iam.py
achantavy: Same reason

cartography/intel/aws/iam.py
achantavy: Good point, though this is a quirk of the data model:

I've made it so that the sub resource relationship of an AWSPolicyStatement is an AWSPolicy, not an AWSAccount. I did this to try minimize schema changes, though I do see that having statements be subresources of policies which are subresources of accounts feels weird. That said, having the statement be scoped to an account feels not correct either.

Anyway (getting to the point), the data model has a constraint that the sub resource id must be passed in via kwargs, and we can't do that if we pass that into the data dict.

I'm open to changing, but I did try to keep the schema the same.

cartography/models/aws/iam/access_key.py
achantavy: Agree, let's add these changes in a follow-up.

Currently the "AccountAccessKey" is supposed to represent https://docs.aws.amazon.com/IAM/latest/APIReference/API_AccessKey.html, but calling that an account access key isn't accurate, that should be an "IAMAccessKey"

cartography/models/aws/iam/account_role.py
achantavy: It's an "AWSAccount as known by an AWSRole". See "Composite Node Pattern": https://github.com/cartography-cncf/cartography/blob/master/AGENTS.md#multiple-intel-modules-modifying-the-same-node-type. From this convention it's understood that this has a subset of the fields compared to the main AWSAccount node.

The idea is we are discovering accounts that IAM knows about via trust documents. These are accounts that may be considered outside of our estate.

tests/integration/cartography/client/aws/test_iam.py
```diff
@@ -21,27 +24,38 @@ def _ensure_test_data(neo4j_session):
...
+    load(
         neo4j_session,
-        "arn:aws:iam::000000000000:group/example-group-0/example-group-0/inline_policy/group_inline_policy",
-        "group_inline_policy",
-        "inline",
-        "arn:aws:iam::000000000000:group/example-group-0",
-        TEST_UPDATE_TAG,
+        AWSInlinePolicySchema(),
+        [
+            {
```
jychp: Nit: for readability, we should define the object beforehand with an explicit name.

tests/integration/cartography/intel/aws/iam/test_iam.py
```diff
@@ -84,65 +108,15 @@ def _get_principal_role_nodes(neo4j_session):
...
         neo4j_session,
-        data,
+        tests.data.aws.iam.LIST_ROLES,
         TEST_ACCOUNT_ID,
         TEST_UPDATE_TAG,
     )
 
-    # Get TRUSTS_AWS_PRINCIPAL relationships from Neo4j.
+    # Assert: Get TRUSTS_AWS_PRINCIPAL relationships from Neo4j.
     result = neo4j_session.run(
```
jychp: Nit: rework with check_rels

tests/integration/cartography/intel/aws/iam/test_iam.py
```diff
@@ -172,22 +146,32 @@ def test_load_roles_creates_trust_relationships(neo4j_session):
...
+    load(
         neo4j_session,
-        "arn:aws:iam::000000000000:group/example-group-0/example-group-0/inline_policy/group_inline_policy",
-        "group_inline_policy",
-        "inline",
-        "arn:aws:iam::000000000000:group/example-group-0",
-        TEST_UPDATE_TAG,
+        AWSInlinePolicySchema(),
+        [
+            {
```
jychp: Nit: for readability, we should define the object beforehand with an explicit name.

tests/integration/cartography/intel/aws/iam/test_iam_sync.py
```diff
@@ -0,0 +1,335 @@
...
+    mock_get_group_managed_policy_data,
+    mock_get_role_list_data,
+    mock_get_role_policy_data,
+    mock_get_role_managed_policy_data,
+    neo4j_session,
+):
+    """Test IAM sync end-to-end"""
+    # Arrange
+    boto3_session = MagicMock()
+    neo4j_session.run("MATCH (n) DETACH DELETE n")
```
jychp: Not needed.
https://github.com/cartography-cncf/cartography/blob/938e8850a97f7a594488183a9523cb170b386baf/tests/integration/conftest.py#L17

jychp (approved):
LGTM.
Some comments to address in a follow-up PR.

tests/integration/cartography/intel/aws/iam/test_iam.py
achantavy: Fixing. Wanted to keep the tests the same to show that this doesn't break anything but now that it's shown that, I'll clean it up

PR #1718, opened 2025-07-25
chore: Refactor AWS IdentityCenter Role Assignments

shyammukund:
### Checklist

- [ ] Graph shows AWSRole-[ALLOWED_BY]->(AWSSSOUser) relationship
Screenshot 2025-07-24 at 5 39 07 PM


cartography/intel/aws/identitycenter.py
```diff
@@ -220,31 +225,103 @@ def get_role_assignments(
...
+    results = neo4j_session.run(get_sso_roles_query)
+
+    mapping = {}
+    for record in results:
+        role_arn = record["arn"]
+        # Extract account and permission set name from ARN
+        account_id = role_arn.split(":")[4]
+        role_name = role_arn.split("/")[-1]  # AWSReservedSSO_Name_Suffix
+
+        if role_name.startswith("AWSReservedSSO_"):
```
jychp: **Nitpick**: It took me several passes to understand this part of the code.
The confusion comes from the fact that it performs multiple things at once:

* filtering based on the prefix
* extracting the role name

I would have rather:

* first checked the prefix: `if not startswith: continue`
* then extracted the role name

**Note**: If this function is meant to be reused, a unit test would be a nice addition.

jychp (approved):
Just a nitpicking comment.

By the way, I found the `ALLOWED_BY` relationship between `SSOUser` and `Role` a bit misleading — I believe `INHERITED_FROM` would be more appropriate (cc @achantavy).
I understand this would be part of a broader refactoring, so this is just a suggestion for future consideration, not a requirement for this PR.

cartography/intel/aws/identitycenter.py
```diff
@@ -220,31 +225,103 @@ def get_role_assignments(
...
+def get_sso_roles_from_db(neo4j_session: neo4j.Session) -> Dict[str, str]:
+    """
+    Get SSO role ARNs from database (already synced by IAM module).
+    Returns mapping: {account_id:permission_set_name -> exact_role_arn}
+    """
+    get_sso_roles_query = """
+    MATCH (role:AWSRole)
+    WHERE role.arn CONTAINS '/aws-reserved/sso.amazonaws.com/'
+    RETURN role.arn as arn
+    """
```
achantavy: This is not scoped enough - as written it will return all AWSRoles across all accounts. I'll write a more thorough recommendation in a bit

cartography/models/aws/identitycenter/awspermissionset.py
```diff
@@ -88,5 +132,6 @@ class AWSPermissionSetSchema(CartographyNodeSchema):
         [
             PermissionSetToInstanceRel(),
             PermissionSetToAWSRoleRel(),
+            RoleAssignmentAllowedByMatchLink(),
```
achantavy: Remove this: the Matchlink shouldn't be included as part of the NodeSchema because then the cleanupbuilder code won't know how to handle it properly. This is an interesting quirk I hadn't thought of so we should have an error raise here

cartography/intel/aws/identitycenter.py
```diff
@@ -327,9 +407,19 @@ def sync_identity_center_instances(
                 instance_arn,
                 region,
             )
+
+            # Get SSO role mapping from database (synced by IAM module)
```
achantavy: Following up on my previous comment.

So at this point in the code we have all the role_assignments. This data structure looks like

```python
[{
  "UserId": "asdf", "PermissionSetArn": "arn",
},...]
```

We have `UserId` + `PermissionSetArn`, but we need `UserId` + `RoleArn`, so we can query for the role. In fact, we can get the role without even needing to do string parsing because the association from the role exists already!


I would do it like this
```python
from cartography.client.core.tx import read_list_of_dicts_tx


def get_permset_roles(
    neo4j_session: neo4j.Session,
    role_assignments: list[dict[str, Any]],
) -> list[dict[str, Any]]:
    permset_ids = list({ra["PermissionSetArn"] for ra in role_assignments})

    query = """
    MATCH (role:AWSRole)<-[:ASSIGNED_TO_ROLE]-(permset:AWSPermissionSet)
    WHERE permset.arn IN $PermSetIds
    RETURN permset.arn AS PermissionSetArn, role.arn AS RoleArn
    """
    permset_to_role = neo4j_session.execute_read(
        read_list_of_dicts_tx,
        query,
        PermSetIds=permset_ids,
    )

    permset_to_role_map = {
        entry["PermissionSetArn"]: entry["RoleArn"]
        for entry in permset_to_role
    }

    return [
        {
            **assignment,
            "RoleArn": permset_to_role_map.get(assignment["PermissionSetArn"]),
        }
        for assignment in role_assignments
    ]
```


and then you  would call it like this: 

```python
permset_roles = get_permset_roles(neo4j, role_assignments)
load_role_assignments(
    neo4j_session,
    permset_roles,
    current_aws_account_id,
    update_tag,
)
```


This way we can remove `transform_role_assignments` altogether and not need to do string matching

shyammukund: this makes a lot of sense. Instead of guessing the role ARN from the naming pattersn we can just ask the graph what roles are actually assigned to these permission sets.

cartography/intel/aws/identitycenter.py
shyammukund: Yeah the parsing logic got pretty gross looking. I went ahead and replaced this with alex's suggestion to use a get function to fetch permission set roles. this replaced the get_sso_roles_from_db and eliminated the need for the transform function. Would appreciate your feedback on this change whenever you get the chance!

Issue #1723, opened 2025-07-26
GCP cleanup jobs delete objects in other projects


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

GCP cleanup jobs are not scoped to the project

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run the GCP sync on multiple projects. The cleanup job of the currently running sync will delete assets where the update tag is stale in _all_ projects

This bug has existed for quite some time.

achantavy:
Repro:
```python


VPC_RESPONSE_2 = {
    "id": "projects/project-def/global/networks",
    "items": [
        {
            "autoCreateSubnetworks": True,
            "creationTimestamp": "2018-05-10T17:33:18.968-07:00",
            "description": "Default network for the project",
            "id": "234567",
            "kind": "compute#network",
            "name": "default2",
            "routingConfig": {
                "routingMode": "REGIONAL",
            },
            "selfLink": "https://www.googleapis.com/compute/v1/projects/project-def/global/networks/default2",
            "subnetworks": [
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/europe-west2/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/asia-east2/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/asia-east1/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/us-east1/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/northamerica-northeast1/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/europe-west1/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/europe-west3/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/asia-south1/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/europe-west4/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/asia-southeast1/subnetworks/default2",
                "https://www.googleapis.com/compute/v1/projects/project-def/regions/us-west2/subnetworks/default2",
[...]

PR #1724, opened 2025-07-27
chore: Refactor GCP VPC to new model, fix: cleanup by project

achantavy:
- Refactors GCP VPCs to the data model.
- Performs cleanup project by project 

Related to #1723

### Checklist

Before:
image

After:
image

### DIFF
--- a/cartography/intel/gcp/__init__.py
+++ b/cartography/intel/gcp/__init__.py
@@ -391,6 +391,7 @@ def _sync_multiple_projects(
     # Compute data sync
     for project in projects:
         project_id = project["projectId"]
+        common_job_parameters["PROJECT_ID"] = project_id
         logger.info("Syncing GCP project %s for Compute.", project_id)
         _sync_single_project_compute(
             neo4j_session,
@@ -399,10 +400,12 @@ def _sync_multiple_projects(
             gcp_update_tag,
             common_job_parameters,
         )
+        del common_job_parameters["PROJECT_ID"]
 
     # Storage data sync
     for project in projects:
         project_id = project["projectId"]
+        common_job_parameters["PROJECT_ID"] = project_id
         logger.info("Syncing GCP project %s for Storage", project_id)
         _sync_single_project_storage(
             neo4j_session,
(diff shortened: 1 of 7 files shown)


cartography/intel/gcp/compute.py
```diff
@@ -1267,8 +1254,7 @@ def sync_gcp_vpcs(
     """
     vpc_res = get_gcp_vpcs(project_id, compute)
     vpcs = transform_gcp_vpcs(vpc_res)
-    load_gcp_vpcs(neo4j_session, vpcs, gcp_update_tag)
-    # TODO scope the cleanup to the current project - https://github.com/cartography-cncf/cartography/issues/381
```
achantavy, line 1271: Whew finally

cartography/intel/gcp/__init__.py
```diff
@@ -391,6 +391,7 @@ def _sync_multiple_projects(
     # Compute data sync
     for project in projects:
         project_id = project["projectId"]
+        common_job_parameters["PROJECT_ID"] = project_id
```
jychp, line 394: Remark (for myself):
We should add a rule in Cubic to ensure that the common parameters dict is not altered during runs, it could lead to some unwanted behavior.

I personally prefer using a "local copy" rather than adding and deleting a key, but it's only cosmetic consideration.

cartography/models/gcp/compute/vpc.py
```diff
@@ -0,0 +1,50 @@
...
+    auto_create_subnetworks: PropertyRef = PropertyRef("auto_create_subnetworks")
+    routing_config_routing_mode: PropertyRef = PropertyRef(
+        "routing_config_routing_mode"
+    )
+    description: PropertyRef = PropertyRef("description")
+
+
+@dataclass(frozen=True)
+class GCPVpcToProjectRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("LASTUPDATED", set_in_kwargs=True)
```
jychp, line 30: As we discuss in Slack I really like to have `set_in_kwargs` in the upper case.

jychp (approved):
LGTM

cartography/intel/gcp/__init__.py
achantavy: Ah ok, will fix in future. AWS sync will need that too. common_job_parameters has these problems imo

it was originally designed to just hold params for the cleanup and analysis jobs, but because it's a convenient place to hold global state, things like the permission rels and the cloudtrail assumerole syncs now use it as a global config.

We should rethink this eventually but I don't have a good answer right now

Issue #1715, opened 2025-07-24
Crash in load_lambda_function()


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

The lambda sync crashes and doesn't retry. We should refactor it to the new data model so that it gets managed transactions with retries.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run the lambda sync and have a flaky environment

*Logs*:
> If applicable, copy and paste your console log with the failing stack trace.

```
  File "/app/.venv/lib/python3.13/site-packages/cartography/intel/aws/lambda_function.py", line 82, in load_lambda_functions
  File "/app/.venv/lib/python3.13/site-packages/cartography/intel/aws/lambda_function.py", line 293, in sync_lambda_functions
```

*Please complete the following information:*:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]

0.108.0rc2

achantavy:
Fixed in #1731

PR #1731, opened 2025-07-28
chore: Refactor AWS Lambda Module 

shyammukund:
- Refactored AWS Lambda module to use data model for loading Lambda functions, Aliases, EventSourceMapping, and Layers
- Replaced all manual cleanup with schema based cleanup
- Updated Integration tests to use check_rels and check_nodes now that we have the data model
- Updated rels in schema.md

- https://github.com/cartography-cncf/cartography/issues/1715

### Checklist

Screenshot 2025-07-28 at 4 15 02 PM


cartography/models/aws/lambda_function/alias.py
```diff
@@ -0,0 +1,77 @@
...
+    label: str = "AWSLambdaFunctionAlias"
+    properties: AWSLambdaFunctionAliasNodeProperties = (
+        AWSLambdaFunctionAliasNodeProperties()
+    )
+    sub_resource_relationship: AWSLambdaFunctionAliasToAWSAccountRel = (
+        AWSLambdaFunctionAliasToAWSAccountRel()
+    )
+
+
+# Matchlink relationship: AWSLambda --[:KNOWN_AS]--> AWSLambdaFunctionAlias
```
achantavy: Matchlink is usually reserved for analytical jobs that connect existing nodes together like with the cloudtrail connections. If we have nodes A and B and want to connect them with a matchlink, we have to do
- call API A, load API A
-  call API B, load API B
- read graph for A and B
- write matchlink for A and B

Since function aliases and lambdas would be known by the same API objects, I don't think we need to incur the performance hit to use a Matchlink here; is it just possible to use a standard schema object?

This way the steps would be 
- call Lambda API, load Lambdas
- call Function Alias API, load Function aliases

which would significantly save on the round trip times.

Matchlinks are only for those cases where we really have no way around it.

tests/integration/cartography/intel/aws/test_lambda.py
```diff
@@ -179,131 +239,194 @@ def test_load_lambda_function_aliases_relationships(neo4j_session):
...
+        transformed_data,
         TEST_REGION,
         TEST_ACCOUNT_ID,
         TEST_UPDATE_TAG,
     )
 
     esm = tests.data.aws.lambda_function.LIST_EVENT_SOURCE_MAPPINGS
 
-    cartography.intel.aws.lambda_function._load_lambda_event_source_mappings(
+    cartography.intel.aws.lambda_function.load_lambda_event_source_mappings(
```
achantavy, line 556: [ no action required ] Nit: avoid renaming the existing function so that it's clearer to show that your refactor PR doesn't break existing tests

tests/integration/cartography/intel/aws/test_lambda.py
```diff
@@ -1,5 +1,7 @@
 import cartography.intel.aws.lambda_function
 import tests.data.aws.lambda_function
+from tests.integration.util import check_nodes
+from tests.integration.util import check_rels
```
achantavy, line 10: Can you add a test that covers the whole sync function where the `get` is mocked out like with other refactors?

docs/root/modules/aws/schema.md
```diff
@@ -454,16 +467,21 @@ Representation of an [AWSLambdaLayer](https://docs.aws.amazon.com/lambda/latest/
...
 | **id** | The arn of the lambda function layer|
+| arn | The arn of the lambda function layer|
 | codesize | The size of the layer archive in bytes.|
 | signingprofileversionarn | The Amazon Resource Name (ARN) for a signing profile version.|
 | signingjobarn | The Amazon Resource Name (ARN) of a signing job. |
 
 #### Relationships
 
-- AWSLambda functions has AWS Lambda Layers.
+- AWSLambdaLayer are sub-resources in an AWS Account.
```
achantavy:
```suggestion
- AWSLambdaLayer belong to AWS Accounts
```

docs/root/modules/aws/schema.md
```diff
@@ -439,10 +448,14 @@ Representation of an [AWSLambdaEventSourceMapping](https://docs.aws.amazon.com/l
 
 #### Relationships
 
-- AWSLambda functions may have the resource AWSLambdaEventSourceMapping.
+- AWSLambdaEventSourceMapping are sub-resources in an AWS Account.
```
achantavy:
```suggestion
- AWSLambdaEventSourceMapping belong to AWS Accounts.
```

docs/root/modules/aws/schema.md
```diff
@@ -401,18 +405,23 @@ Representation of an [AWSLambdaFunctionAlias](https://docs.aws.amazon.com/lambda
...
+| arn | The arn of the lambda function alias|
+| aliasname |  The name of the lambda function alias |
 | functionversion | The function version that the alias invokes.|
 | revisionid |  A unique identifier that changes when you update the alias. |
 | description |  The description of the alias. |
 
 #### Relationships
 
-- AWSLambda functions may also have aliases.
+- AWSLambdaFunctionAlias are sub-resources in an AWS Account.
```
achantavy:
```suggestion
- AWSLambdaFunctionAlias belong to AWS Accounts.
```

achantavy (changes requested):
Minor minor comments. Also remember to remove lines from indexes.cypher related to the lambda node labels

cartography/intel/aws/lambda_function.py
```diff
@@ -117,170 +127,95 @@ def get_event_source_mappings(
...
+
+    # Clean up parent Lambda nodes last
+    GraphJob.from_node_schema(AWSLambdaSchema(), common_job_parameters).run(
+        neo4j_session
     )
 
 
 @timeit
-def sync_lambda_functions(
+def sync(
```
achantavy, line 292: This function has gotten long. Is there a way clean it up, perhaps by breaking it up into separate sync functions?

Like to have `sync()` call
- sync function aliases
- sync event source mappings
- sync lambda layers

and then each of those call their own respective get+transform+loads.

Example: https://github.com/cartography-cncf/cartography/blob/b68f2e372ba12a474a490a90eae33d08c87e883b/cartography/intel/aws/rds.py#L502-L533

Example: https://github.com/cartography-cncf/cartography/blob/3077dbb55ffa601b72d8dcd4263cfe10af6cc711/cartography/intel/aws/iam.py#L1277-L1326

PR #1737, opened 2025-07-29
Add disk ingestion for Trivy

kunaals:
## Summary
- support ingesting Trivy results stored on disk via `--trivy-results-dir`
- update Trivy ingestion logic for local files
- document new option
- add integration test

## Testing
- `make test_lint`
- `make test_unit`
- `make test_integration` *(fails: Couldn't connect to localhost:7687)*

------
https://chatgpt.com/codex/tasks/task_b_68890a90ce0c8323b13e10d88dc61893


cartography/intel/trivy/__init__.py
```diff
@@ -124,29 +126,73 @@ def sync_trivy_aws_ecr_from_s3(
...
+    if not json_files:
+        logger.error(
+            f"Trivy sync was configured, but no json files were found in {results_dir}."
+        )
+        raise ValueError("No Trivy json results found on disk")
+
+    logger.info(f"Processing {len(json_files)} local Trivy result files")
+
+    for file_path in json_files:
+        artifact_name, image_digest = sync_single_image_from_file(
```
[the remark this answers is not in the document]
kunaals: This isn't a big deal because folks should only look at vulns attached to an image they care about - the relations from vulns to images should be deleted on new syncs

cartography/intel/trivy/scanner.py
```diff
@@ -177,6 +178,18 @@ def get_json_files_in_s3(
...
 
+@timeit
+def get_json_files_in_dir(results_dir: str) -> set[str]:
+    """Return set of JSON file paths under a directory."""
+    results = set()
+    for root, _dirs, files in os.walk(results_dir):
+        for filename in files:
+            if filename.endswith(".json"):
+                results.add(os.path.join(root, filename))
+    logger.info(f"Found {len(results)} json files in {results_dir}")
```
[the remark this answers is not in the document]
kunaals: Exposing a file path is not an issue in internal logs

cartography/intel/trivy/__init__.py
```diff
@@ -124,29 +126,73 @@ def sync_trivy_aws_ecr_from_s3(
...
+def sync_trivy_aws_ecr_from_dir(
+    neo4j_session: Session,
+    results_dir: str,
+    update_tag: int,
+    common_job_parameters: dict[str, Any],
+) -> None:
+    """Sync Trivy scan results from local files for AWS ECR images."""
+    logger.info(f"Using Trivy scan results from {results_dir}")
+
+    images_in_graph: set[str] = get_scan_targets(neo4j_session)
```
achantavy: Nit: clearer to name these as `image_uris`

cartography/intel/trivy/scanner.py
```diff
@@ -361,3 +374,61 @@ def sync_single_image_from_s3(
...
+@timeit
+def read_scan_results_from_file(
+    file_path: str,
+) -> tuple[str | None, list[dict], str | None]:
+    """Read and parse Trivy scan results from a local file."""
+    logger.debug(f"Reading scan results from file: {file_path}")
+    with open(file_path, encoding="utf-8") as f:
+        trivy_data = json.load(f)
+
+    artifact_name = trivy_data.get("ArtifactName")
```
achantavy: Ah ok, so this from trivy is basically an ECR image URI: https://github.com/cartography-cncf/cartography/blob/b68f2e372ba12a474a490a90eae33d08c87e883b/tests/data/trivy/trivy_sample.py#L4.

I think we should do a `trivy_data["ArtifactName"]` instead of a `.get()` and raise exc if this is None because `ArtifactName` feels like a super important field and if it's missing then that sounds like something has gone very wrong and we shouldn't continue.

It'd also simplify the code to make `artifact_name` not be an optional field anymore.

cartography/intel/trivy/scanner.py
```diff
@@ -361,3 +374,61 @@ def sync_single_image_from_s3(
...
+
+    if "Results" in trivy_data and trivy_data["Results"]:
+        results = trivy_data["Results"]
+    else:
+        stat_handler.incr("image_scan_no_results_count")
+        logger.warning(
+            f"File scan data did not contain a `Results` key for {file_path}; continuing."
+        )
+        results = []
```
achantavy: When Trivy has 0 findings, IIRC the expected behavior is for the `Results` key to exist but have an empty list. If the `Results` key is not present at all, I think that's an anomaly where we should raise an exc to stop the graph from continuing with the cleanup jobs.

cartography/intel/trivy/scanner.py
```diff
@@ -361,3 +374,61 @@ def sync_single_image_from_s3(
...
+
+    image_digest = None
+    if "Metadata" in trivy_data and trivy_data["Metadata"]:
+        repo_digests = trivy_data["Metadata"].get("RepoDigests", [])
+        if repo_digests:
+            repo_digest = repo_digests[0]
+            if "@" in repo_digest:
+                image_digest = repo_digest.split("@")[1]
```
achantavy: If having an image digest is critical to other things later on in the workstream, then let's make it not optional and have it raise exc here if not present.

Otherwise if not critical, we can leave it it as it is.

achantavy (changes requested):
Left comments on whether we should return None and raise exception if some key trivy fields are missing, lmk what you think

cartography/intel/trivy/scanner.py
```diff
@@ -245,58 +342,6 @@ def load_scan_fixes(
     )
 
 
-@timeit
-def read_scan_results_from_s3(
```
kunaals, line 249: this is the logic that i moved into `_parse_trivy_data`

tests/data/aws/ecr.py
```diff
@@ -6,21 +6,21 @@
             "repositoryArn": "arn:aws:ecr:us-east-1:000000000000:repository/example-repository",
             "registryId": "000000000000",
             "repositoryName": "example-repository",
-            "repositoryUri": "000000000000.dkr.ecr.us-east-1/example-repository",
```
kunaals, line 9: This is wrong, ECR repos contain `amazonaws.com` in them

kunaals: It was messing up the Trivy integration test because we create ECR assets

tests/data/trivy/trivy_sample.py
```diff
@@ -22,7 +22,7 @@
             "000000000000.dkr.ecr.us-east-1.amazonaws.com/test-repository:1234567890"
         ],
         "RepoDigests": [
-            "000000000000.dkr.ecr.us-east-1.amazonaws.com/test-repository@sha256:88016fd13f6aa39b702c1714662"
```
kunaals, line 25: We make this the same as the repo digest SHAs from ECR

tests/integration/cartography/intel/trivy/test_helpers.py
```diff
@@ -0,0 +1,255 @@
+from neo4j import Session
```
kunaals, line 1: same as the tests that we had in tests/integration/cartography/intel/trivy/test_trivy_aws_ecr.py

cartography/intel/trivy/scanner.py
kunaals: done

cartography/intel/trivy/__init__.py
kunaals: done

cartography/intel/trivy/scanner.py
kunaals: done

cartography/intel/trivy/scanner.py
kunaals: done, digests are critical

cartography/intel/trivy/__init__.py
```diff
@@ -124,29 +127,86 @@ def sync_trivy_aws_ecr_from_s3(
...
+
+    logger.info(f"Processing {len(json_files)} local Trivy result files")
+
+    for file_path in json_files:
+        # First, check if the image exists in the graph before syncing
+        try:
+            # Peek at the artifact name without processing the file
+            with open(file_path, encoding="utf-8") as f:
+                trivy_data = json.load(f)
+                artifact_name = trivy_data.get("ArtifactName")
```
achantavy, line 157: [nonblock] could be worth logging the file name if artifactname ends up be falsy

PR #1750, opened 2025-08-01
feat(sentinelone): Add CVE ingestion to SentinelOne intelmodule

serge-wq:
### Summary

Some small refactors and the addition of CVE ingestion to the SentinelOne intelmodule.

Screenshot 2025-08-01 at 9 20 05 a m

Screenshot 2025-08-01 at 9 35 08 a m

### Related issues or links

### Checklist

### DIFF
--- a/docs/root/modules/sentinelone/schema.md
+++ b/docs/root/modules/sentinelone/schema.md
@@ -159,16 +159,10 @@ Represents a Common Vulnerability and Exposure (CVE) discovered in applications 
 | lastupdated | Timestamp of the last time the node was updated |
 | **id** | The unique identifier for the S1CVE record. Format: `S1\|{cve_id}`. |
 | **cve_id** | The CVE identifier (e.g., CVE-2023-12345) |
-| application_version_id | The ID of the affected application version |
 | base_score | The CVSS base score for the vulnerability |
 | cvss_version | The version of CVSS used for scoring |
-| days_detected | Number of days since the CVE was first detected |
-| detection_date | ISO 8601 timestamp of when the CVE was first detected |
-| last_scan_date | ISO 8601 timestamp of the last scan that detected this CVE |
-| last_scan_result | Result of the last scan for this CVE |
 | published_date | ISO 8601 timestamp of when the CVE was published |
 | severity | The severity level of the vulnerability (e.g., Critical, High, Medium, Low) |
-| status | Current status of the CVE (e.g., Active, Resolved) |
 
 #### Relationships
 
@@ -183,3 +177,13 @@ Represents a Common Vulnerability and Exposure (CVE) discovered in applications 
     ```
     (S1CVE)-[AFFECTS]->(S1ApplicationVersion)
     ```
+
+    The AFFECTS relationship includes additional properties specific to each CVE finding:
+
+    | Property | Description |
+    |----------|-------------|
+    | days_detected | Number of days since the CVE was first detected in this application version |
+    | detection_date | ISO 8601 timestamp of when the CVE was first detected |
+    | last_scan_date | ISO 8601 timestamp of the last scan that detected this CVE |
+    | last_scan_result | Result of the last scan for this CVE |
+    | status | Current status of the CVE finding (e.g., Active, Resolved) |
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ You can learn more about the story behind Cartography in our [presentation at BS
 - [Oracle Cloud Infrastructure](https://cartography-cncf.github.io/cartography/modules/oci/index.html) - IAM
 - [PagerDuty](https://cartography-cncf.github.io/cartography/modules/pagerduty/index.html) - Users, teams, services, schedules, escalation policies, integrations, vendors
 - [Scaleway](https://cartography-cncf.github.io/cartography/modules/scaleway/index.html) - Projects, IAM, Local Storage, Instances
-- [SentinelOne](https://cartography-cncf.github.io/cartography/modules/sentinelone/index.html) - Accounts, Agents
+- [SentinelOne](https://cartography-cncf.github.io/cartography/modules/sentinelone/index.html) - Accounts, Agents, Applications, Application Versions, CVEs
 - [Slack](https://cartography-cncf.github.io/cartography/modules/slack/index.html) - Teams, Users, UserGroups, Channels
 - [SnipeIT](https://cartography-cncf.github.io/cartography/modules/snipeit/index.html) - Users, Assets
 - [Tailscale](https://cartography-cncf.github.io/cartography/modules/tailscale/index.html) - Tailnet, Users, Devices, Groups, Tags, PostureIntegrations


tests/integration/cartography/intel/sentinelone/test_cve.py
```diff
@@ -0,0 +1,212 @@
...
+    "get_paginated_results",
+)
+def test_sync_cves(mock_get_paginated_results, neo4j_session):
+    """
+    Test that CVE sync works properly by syncing CVEs and verifying nodes and relationships
+    including relationships between S1CVE and S1ApplicationVersion
+    """
+    # Mock the API call to return test data
+    mock_get_paginated_results.return_value = CVES_DATA
+
```
achantavy, line 33:
```suggestion

# Arrange
```

tests/integration/cartography/intel/sentinelone/test_cve.py
```diff
@@ -0,0 +1,212 @@
...
+    # Create prerequisite S1ApplicationVersion nodes for the relationships
+    for app_version_id in EXPECTED_APP_VERSION_IDS.values():
+        neo4j_session.run(
+            "CREATE (av:S1ApplicationVersion {id: $app_version_id, lastupdated: $update_tag})",
+            app_version_id=app_version_id,
+            update_tag=TEST_UPDATE_TAG,
+        )
+
+    # Run the sync
```
achantavy:
```suggestion
    # Act: Run the sync
```

tests/integration/cartography/intel/sentinelone/test_cve.py
```diff
@@ -0,0 +1,212 @@
...
+    cartography.intel.sentinelone.cve.sync(
+        neo4j_session,
+        TEST_COMMON_JOB_PARAMETERS,
+    )
+
+    # Verify that the correct CVE nodes were created
```
achantavy, line 57:
```suggestion
    # Assert:
    # Verify that the correct CVE nodes were created
```

tests/data/sentinelone/cve.py
```diff
@@ -0,0 +1,110 @@
+CVE_ID_1 = "cve-test-123"
+CVE_ID_2 = "cve-test-456"
+CVE_ID_3 = "cve-test-789"
```
achantavy, line 3: just so i understand, how is this different from the `cveId`?

cartography/models/sentinelone/cve.py
```diff
@@ -0,0 +1,72 @@
...
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("application_version_id")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "AFFECTS"
+    properties: S1AffectsRelProperties = S1AffectsRelProperties()
+
+
+@dataclass(frozen=True)
+class S1CVESchema(CartographyNodeSchema):
```
achantavy, line 66: I think we should add in `ExtraNodeLabels` for `Risk` and _maybe_ `CVE` as well. What do you think?

cartography/intel/sentinelone/cve.py
```diff
@@ -0,0 +1,125 @@
...
+    GET -> TRANSFORM -> LOAD -> CLEANUP
+    """
+    logger.info("Syncing SentinelOne CVE data")
+
+    api_url = str(common_job_parameters.get("API_URL"))
+    api_token = str(common_job_parameters.get("API_TOKEN"))
+    account_id = str(common_job_parameters.get("S1_ACCOUNT_ID"))
+    update_tag = int(common_job_parameters.get("UPDATE_TAG", 0))
+
+    # 1. GET - Fetch data from API
```
achantavy: Nit: no need to add in the extra `#` comments. The code itself here is already self-documenting

achantavy (changes requested):
Looks good and straightforward to me! just left feedback on adding additional labels to S1CVE -- lmk your thoughts

cartography/models/sentinelone/cve.py
serge-wq: Definitely. Good callout.

tests/data/sentinelone/cve.py
serge-wq: Not very different. We have both fields in other CVE nodes. Do you think there could be conflicts? Maybe I can do something like:
```
id="S1|CVE-2025-123"
cve_id="CVE-2025-123"
```
just to keep consistency

cartography/intel/sentinelone/cve.py
serge-wq: Removed.

cartography/intel/sentinelone/api.py
```diff
@@ -37,35 +37,54 @@ def call_sentinelone_api(
...
-
-        response = requests.request(
-            method=method,
-            url=full_url,
-            headers=headers,
-            params=params,
-            json=data,
-            timeout=_TIMEOUT,
-        )
+    logger.debug(
```
heryxpc: nit: I don't see much value for debugging the HTTP call and there aren't many examples of it in other 'get' operations.

cartography/intel/sentinelone/cve.py
```diff
@@ -0,0 +1,118 @@
...
+def transform(cves_list: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """
+    Transform SentinelOne CVE data for loading into Neo4j
+    """
+    transformed_cves = []
+    for cve in cves_list:
+        app_version_id = get_application_version_id(
+            cve.get("applicationName", "unknown"),
+            cve.get("applicationVendor", "unknown"),
+            cve.get("applicationVersion", "unknown"),
```
heryxpc, line 42: What would happen in the worst case scenario (e.g. all are `"unknown"`?, Wouldn't that cause an invalid version ID or nodes with missing information being treated as the same application-version?

cartography/intel/sentinelone/cve.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+        transformed_cve = {
+            # Required fields - let them fail if missing
+            "id": cve["id"],
+            "cve_id": cve["cveId"],
+            # Optional fields - use .get() with None default
+            "application_version_id": app_version_id,
```
heryxpc, line 49: Is `app_version_id` optional as well hence why the `"unknown"` values are accepted?

cartography/intel/sentinelone/cve.py
```diff
@@ -0,0 +1,118 @@
...
+) -> None:
+    """
+    Sync SentinelOne CVEs following the standard pattern:
+    GET -> TRANSFORM -> LOAD -> CLEANUP
+    """
+    logger.info("Syncing SentinelOne CVE data")
+
+    api_url = str(common_job_parameters.get("API_URL"))
+    api_token = str(common_job_parameters.get("API_TOKEN"))
+    account_id = str(common_job_parameters.get("S1_ACCOUNT_ID"))
```
heryxpc: I don't think the string encoding is needed.

cartography/intel/sentinelone/cve.py
```diff
@@ -0,0 +1,118 @@
...
+    update_tag = int(common_job_parameters.get("UPDATE_TAG", 0))
```
heryxpc: This is a dangerous operation, setting the `UPDATE_TAG` to 0 breaks the assumption that all intel modules must have an update tag.
```suggestion

```

cartography/models/sentinelone/cve.py
heryxpc, line 1: Just my 2 🪙 , It would be nice to connect Sentinel CVE's with [NVD CVE nodes](https://github.com/cartography-cncf/cartography/blob/master/cartography/models/cve/cve.py).

docs/root/modules/sentinelone/schema.md
```diff
@@ -136,3 +142,44 @@ Represents a specific version of an application installed on SentinelOne agents.
...
+
+### S1CVE
+
+Represents a Common Vulnerability and Exposure (CVE) discovered in applications managed by SentinelOne.
+
+| Field | Description |
+|-------|-------------|
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | The unique identifier for the S1CVE record |
```
heryxpc: What is the shape of `S1CVE.id`?, and, can it be assumed that it's unique among all `risks` between all applications from any version?

cartography/intel/sentinelone/api.py
serge-wq: Removed the log call.

cartography/intel/sentinelone/cve.py
serge-wq: If we have any combination of "unknown" values here, it would only affect the links with the `S1ApplicationVersion` nodes. Worst case scenario we have `S1CVE` nodes that are not connected to any `S1ApplicationVersion`, which we can see why the API call is not returning this information in a case by case scenario.

cartography/intel/sentinelone/cve.py
serge-wq: I do the `str` cast to satisfy type hints requirements for function parameters run below.

cartography/models/sentinelone/cve.py
serge-wq: Agreed, let me work on this on a following PR

cartography/intel/sentinelone/cve.py
```diff
@@ -0,0 +1,123 @@
...
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get(api_url: str, api_token: str, account_id: str) -> list[dict[str, Any]]:
+    """
+    Fetch CVE data from SentinelOne API
+    Should be simple and raise exceptions on failure
```
achantavy: Bit of AI slop haha

docs/root/modules/sentinelone/schema.md
achantavy: Could be good to document in the code here that the convention is `S1|{CVE_ID}`.

I won't block here though

achantavy:
Oh oh would also be good to update README.md with the additional type supported by S1

docs/root/modules/sentinelone/schema.md
serge-wq: Added.

cartography/intel/sentinelone/cve.py
serge-wq: Removed.

PR #1756, opened 2025-08-02
closing the driver session explicitely

dikshant182004:
I have added the with key word where it was not present at the time of making the session and expliitely closed the session in contest.py

I was facing this warning at the time of testing this .\tests\integration\cartography\intel\okta\test_awssaml.py

DeprecationWarning: Relying on Driver's destructor to close the session is deprecated. Please make sure to close the session. Use it as a context (`with` statement) or make sure to call `.close()` explicitly. Future versions of the driver will not close drivers automatically.

### Checklist

### DIFF
--- a/demo/__main__.py
+++ b/demo/__main__.py
@@ -30,14 +30,13 @@ def main(force_flag: bool) -> None:
         )
     else:
         neo4j_driver = neo4j.GraphDatabase.driver(NEO4J_URL)
-    neo4j_session = neo4j_driver.session()
-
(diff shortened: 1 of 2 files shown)


achantavy:
Discussed on this thread: https://cloud-native.slack.com/archives/C080M2LRLDA/p1754138768398709

> rather than have the driver closed one-by-one like that, i'd rather make a strategy to do that properly for the whole project. this would probably be a large refactor

> until then i would recommend just filing this as an issue if it doesnt exist already and then ignoring the warning

Issue #1205, opened 2023-07-11
AWS S3: get bucket logging


achantavy:
*Description*:
> Describe your idea.  Please be detailed.  If a feature request, please
describe the desired behavior, what scenario it enables, and how it
would be used.

Cartography should retrieve and store the [bucket logging status](https://docs.aws.amazon.com/AmazonS3/latest/API/API_BucketLoggingStatus.html) of an s3 bucket so that we know if the feature is turned on. Unfortunately this needs an entirely separate API call to fetch.

[optional *Relevant Links*:]
> Any extra documentation required to understand the issue.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_bucket_logging.html

PR #1759, opened 2025-08-02
feat: add bucket logging api to aws s3 (#1205)

SachdevJai:
Added S3 bucket logging status support to retrieve and store bucket logging configs 
[reference](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_bucket_logging.html)

added a function to load these configs to neo4j
added a cleanup job to remove stale data

- https://github.com/cartography-cncf/cartography/issues/1205

### Checklist

image


SachdevJai:
@achantavy 

Hi, I tried updating the AWS S3 bucket schema using the API mentioned in the issue referenced

I'm not sure how I'm supposed to write tests for this and would appreciate if you could guide me 

Thank you

cartography/intel/aws/s3.py
```diff
@@ -479,6 +506,33 @@ def _load_bucket_ownership_controls(
...
+    """
+    # Load basic logging status
+    ingest_bucket_logging = """
+    UNWIND $bucket_logging_configs AS bucket_logging
+    MATCH (bucket:S3Bucket{name: bucket_logging.bucket})
+    SET bucket.logging_enabled = bucket_logging.logging_enabled,
+        bucket.logging_target_bucket = bucket_logging.target_bucket,
+        bucket.logging_target_prefix = bucket_logging.target_prefix,
+        bucket.logging_target_grants = bucket_logging.target_grants,
+        bucket.logging_target_object_key_format = bucket_logging.target_object_key_format,
```
achantavy:
```suggestion
```

No need to ingest these fields to the graph. We can wait for a future PR to incorporate these

cartography/intel/aws/s3.py
```diff
@@ -851,6 +911,67 @@ def parse_bucket_ownership_controls(
     }
 
 
+@timeit
```
achantavy:
```suggestion
```
Not necessary because this doesn't involve any network calls

cartography/intel/aws/s3.py
```diff
@@ -851,6 +911,67 @@ def parse_bucket_ownership_controls(
...
+
+    return {
+        "bucket": bucket,
+        "logging_enabled": True,
+        "target_bucket": logging_config.get("TargetBucket"),
+        "target_prefix": logging_config.get("TargetPrefix"),
+        "target_grants": (
+            json.dumps(logging_config.get("TargetGrants"))
+            if logging_config.get("TargetGrants")
+            else None
```
achantavy:
```suggestion
```
Not a fan of handling this as a string. This data isn't important to us anyway so it's ok to omit for now.

cartography/intel/aws/s3.py
```diff
@@ -851,6 +911,67 @@ def parse_bucket_ownership_controls(
...
+        ),
+        "target_object_key_format": (
+            json.dumps(logging_config.get("TargetObjectKeyFormat"))
+            if logging_config.get("TargetObjectKeyFormat")
+            else None
+        ),
```
achantavy:
```suggestion
```
No need to include this for now

tests/integration/cartography/intel/aws/test_s3.py
```diff
@@ -255,3 +255,86 @@ def test_s3_sns_relationship(neo4j_session):
     ) == {
         ("bucket-1", "arn:aws:sns:us-east-1:123456789012:test-topic"),
     }
+
+
+def test_load_s3_bucket_logging(neo4j_session):
+    """
+    Ensure that expected bucket gets loaded with their bucket logging fields.
```
achantavy, line 262: Nit: Please annotate with `# Arrange` + `# Act` + `# Assert` to make this clearer which parts are the things under test

achantavy (changes requested):
Thanks for taking this on and matching the existing patterns of this module. This is the only situation where not using the data model is ok: it's a small change that fits existing patterns and when the other patterns get refactored, this will be straightfoward to follow.

I left comments on removing fields that are not needed. We mostly only need the logging bucket name.

SachdevJai:
> I left comments on removing fields that are not needed. We mostly only need the logging bucket name.

Thank you for the suggestions
I have pushed new changes to resolve the issues 
Kindly take a look again :D

Issue #1255, opened 2023-09-25
[Feature request] Create a DNS_POINTS_TO relationship between DNSRecord and ElasticIPAddress 


0xp4blo:
**Title**:
Introduce (DNSRecord)-[DNS_POINTS_TO]->(ElasticIPAddress) relationship

**Description**:
Loading DNS details for Route53 records should create DNS_POINTS_TO relationship with ElasticIPAddress nodes.

**Relevant links**:
* [*`cartography/interl/aws/route53.py::link_aws_resources`*](https://github.com/lyft/cartography/blob/5c1aa119dad88bb2e934aee5bdcc77275f358ac7/cartography/intel/aws/route53.py#L18) is currently in charge of creating the following DNS_POINTS_TO relationships:

```
(AWSDNSRecord)-[DNS_POINTS_TO]-(AWSDNSRecord)
(AWSDNSRecord)-[DNS_POINTS_TO]-(LoadBalancer)
(AWSDNSRecord)-[DNS_POINTS_TO]-(LoadBalancerV2)
(AWSDNSRecord)-[DNS_POINTS_TO]-(EC2Instance)
```

heryxpc:
👍

PR #1761, opened 2025-08-03
feat: add DNS_POINTS_TO relationship from AWSDNSRecord to ElasticIPAddress (#1255)

SachdevJai:
### Summary

This PR adds a new `DNS_POINTS_TO` relationship from `AWSDNSRecord` nodes to `ElasticIPAddress` nodes, enabling cartography to automatically link DNS A records to their corresponding Elastic IP addresses when the DNS record's value matches the Elastic IP's public IP address.

### Related issues or links

https://github.com/cartography-cncf/cartography/issues/1255

### Checklist

image


tests/integration/cartography/intel/aws/test_route53.py
```diff
@@ -214,3 +225,25 @@ def test_cleanup_dnspointsto_relationships(neo4j_session):
...
+    _ensure_local_neo4j_has_test_route53_records(neo4j_session)
+
+    # Verify that the expected DNS record points to the expected ElasticIP
+    result = neo4j_session.run(
+        """
+        MATCH (n:AWSDNSRecord{id:"/hostedzone/HOSTED_ZONE/hello.what.example.com/A"})
+        -[:DNS_POINTS_TO]->(e:ElasticIPAddress{id:"192.168.1.1"})
+        return n.name, e.public_ip
+        """,
+    )
```
d-aggarwal, line 268: Please consider using the existing check_nodes and check_rels util functions instead of raw Cypher queries for verifying relationships and node in integration tests.

d-aggarwal (changes requested):
Thank you for your contribution!

tests/integration/cartography/intel/aws/test_route53.py
SachdevJai: Hi, I realised that there's multiple other tests which use raw queries for verifying data during testing, 

shall i refactor those as well or open a new issue for that?

SachdevJai: I have updated the test for this specific relation but lots of other tests in the same file and possibly more use the query approach

achantavy: Avoid changing existing tests. With a codebase that has old patterns and new patterns, the old tests help ensure that we don't introduce regressions.

SachdevJai: Adding new test data was causing some previous tests to fail due to addition of more relations

Should I update the expected values in those tests so they pass along with the new test data for this relationship?

achantavy: Since you're creating a new test, you could have the test graph start with a clean slate by running `match (n) detach delete n` as a first step. Then if I understand correctly you won't need to update previous tests.

SachdevJai: I see
Thank you I'll try this

SachdevJai: updated

SachdevJai:
Resolved the integration test failing
Should i rebase to squash all commits into one for a cleaner history?

achantavy:
No strong opinion on rebase or squash

achantavy (approved):
LGTM, thank you!!

Issue #1754, opened 2025-08-01
Circular relationships between DNSZone nodes


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

Somehow have introduced circular relationships between DNSZone nodes.

### Observed behavior:
```cypher
match p1 = (s:DNSZone)-[:SUBZONE]->(s1:DNSZone) 
MATCH p2 = (a:AWSAccount)--(s)
MATCH p3 = (a2:AWSAccount)--(s1)
where (s1)-[:SUBZONE]->(s)
return *
```

returns 
<img width="1098" height="678" alt="Image" src="https://github.com/user-attachments/assets/0ce07a7c-4bfb-4961-9ac8-f9da41b9f62f" />

### Expected behavior:

I think it should return nothing.

### Discussion:
Although the code has this definition: 
https://github.com/cartography-cncf/cartography/blob/3923bd89c3c71f6f30ecdda56b6e0d7a81807393/cartography/intel/aws/route53.py#L389-L392

We query for it with:

```
    MATCH (:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(z:AWSDNSZone)
        <-[:MEMBER_OF_DNS_ZONE]-(record:DNSRecord{type:"NS"})
        -[:DNS_POINTS_TO]->(ns:NameServer)<-[:NAMESERVER]-(z2:AWSDNSZone)
    WHERE record.name=z2.name AND NOT z=z2
    RETURN z.id as zone_id, z2.id as subzone_id
```

and the definition in the code is:
```
    A DNS zone B is a sub zone of A if:
    1. DNS zone A has an NS record that points to a nameserver
    2. That nameserver is associated with DNS zone B
    3. The NS record's name matches the name of DNS zone B
```
We should verify if that query is correct, and we should verify that our creation of the matchlink is correct.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run the route53 sync.

*Please complete the following information:*:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]

0.110.0rc1

janithashri:
The original query correctly identifies delegation through NS records, but it doesn't enforce the hierarchical nature of DNS. This could allow circular or invalid subzone relationships if misconfigured.

[...]

PR #1763, opened 2025-08-03
fix(aws/route53): Prevent circular SUBZONE relationshipsFix/only one file

janithashri:
### Summary

This PR resolves a bug where the Route53 intel module could incorrectly create circular `[:SUBZONE]` relationships between `AWSDNSZone` nodes.

The issue was caused by the subzone linking query correctly identifying DNS delegation but failing to enforce the hierarchical nature of domain names. This is fixed by adding a `WHERE` condition to the Cypher query in `link_sub_zones`. The new condition, `z2.name ENDS WITH '.' + z.name`, ensures that a zone is only identified as a subzone if its name is a valid subdomain of the parent, logically preventing cycles.

### Related issues or links

Fixes https://github.com/cartography-cncf/cartography/issues/1754

### Checklist

Bug Query:

Screenshot (412)

Valid Subzones:

Screenshot (414)

Screenshot (423)


achantavy:
Thanks for submitting, looking forward to trying this out! Will try to get a response back to you in 2 days, please ping back if you haven't heard back by then :)

cartography/intel/aws/route53.py
```diff
@@ -396,9 +396,11 @@ def link_sub_zones(
     """
     query = """
     MATCH (:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(z:AWSDNSZone)
-        <-[:MEMBER_OF_DNS_ZONE]-(record:DNSRecord{type:"NS"})
-        -[:DNS_POINTS_TO]->(ns:NameServer)<-[:NAMESERVER]-(z2:AWSDNSZone)
-    WHERE record.name=z2.name AND NOT z=z2
+    <-[:MEMBER_OF_DNS_ZONE]-(record:DNSRecord{type:"NS"})
+    -[:DNS_POINTS_TO]->(ns:NameServer)<-[:NAMESERVER]-(z2:AWSDNSZone)
+    WHERE record.name = z2.name AND
+    z2.name ENDS WITH '.' + z.name AND
```
achantavy: I think this is the fix, thank you very much. Previously the query was just connecting zone1 and zone2 if they were sharing the same name servers.

tests/data/aws/route53.py
```diff
@@ -150,3 +150,53 @@
         ],
     ),
 ]
+
+# This data simulates a parent zone delegating authority to a subzone.
+GET_ZONES_WITH_SUBZONE = [
```
achantavy, line 180: Looks like the tests are not done. Once this are finished let me know. I've tested it manually and I think it looks good but would like to have integ test coverage.

achantavy (changes requested):

tests/integration/cartography/intel/aws/test_route53.py
```diff
@@ -268,3 +268,53 @@ def test_load_dnspointsto_elasticip_relationships(neo4j_session):
...
+        neo4j_session, TEST_UPDATE_TAG, TEST_AWS_ACCOUNTID,
+    )
+
+    # Assert: Verify the graph state is correct after the run.
+    # 1. Test that the CORRECT subzone relationship WAS created.
+    result = neo4j_session.run(
+        """
+        MATCH (parent:AWSDNSZone{name:$ParentName})-[r:SUBZONE]->(child:AWSDNSZone{name:$ChildName})
+        RETURN count(r) AS rel_count
+        """,
```
d-aggarwal: Please consider using check_nodes and check_rel instead using raw queries in integration test.
For reference - you can look into GCP VPC module, that's refactored.

tests/integration/cartography/intel/aws/test_route53.py
```diff
@@ -268,3 +268,53 @@ def test_load_dnspointsto_elasticip_relationships(neo4j_session):
...
+        ChildName='sub.example.com',
+    )
+    assert result.single()['rel_count'] == 1
+
+    # 2. Test that NO incorrect relationships were created with the unrelated zone.
+    result = neo4j_session.run(
+        """
+        MATCH (z1:AWSDNSZone)-[r:SUBZONE]-(z2:AWSDNSZone{name:$UnrelatedName})
+        RETURN count(r) AS rel_count
+        """,
```
d-aggarwal: here as well

d-aggarwal (dismissed):
Thank you for your contribution.

tests/integration/cartography/intel/aws/test_route53.py
janithashri: Thanks for the feedback, @d-aggarwal! That makes perfect sense. I'll refactor the test assertions to use check_rels.

My initial thought was to keep the style consistent with the other existing tests in this file since it was a bug, but I agree that updating the new test to the modern standard is a good improvement for the codebase.

I'll push up the change shortly.

achantavy:
Hi @janithashri, linter and tests are failing. Please resolve, I really want to get this merged in. Also please address @jychp's comment about potentially committing a credential - I'd recommend just rotating it and not bothering to edit git history (e.g. rebase) because you'll likely end up breaking something

janithashri:
Thank you for the security notice, @jychp @achantavy . This was a mistake on my part. The credential was a test password and is not in use anywhere else, but to follow best security practices, I have revoked and rotated it.

janithashri:
I've pushed a new commit that resolves the failing test and the linter issues. All checks should be passing now. Thank you for your patience.

tests/integration/cartography/intel/aws/test_route53_sync.py
```diff
@@ -358,14 +339,16 @@ def test_sync_route53_sub_zones(mock_get_zones, neo4j_session):
     )
 
     # Assert - Sub-zone relationship should be created
-    assert check_rels(
```
achantavy, line 361: This is going to sound overly nitpicky, but this diff would be a lot cleaner if it just did 

```python
    assert check_rels(
        neo4j_session,
        "AWSDNSZone",
        "zoneid",
        "AWSDNSZone",
        "zoneid",
        "SUBZONE",
        rel_direction_right=False,
    ) == {
        ("/hostedzone/SUB_ZONE", "/hostedzone/HOSTED_ZONE"),
    }, "Sub-zone relationship should be created"
```

The "actual_rels" thing is something Cursor does and it's easy to get into the trap of accepting LLM generated code without reviewing it.

With the change above, the only difference made is just in the ids of the zones, keeping everything else the same so the diff is easy to follow.

Sorry if this sounds nitpicky. This comment is intended to be constructive and helpful.

tests/integration/cartography/intel/aws/test_route53_sync.py
```diff
@@ -317,7 +318,7 @@ def test_sync_route53_cleanup(mock_get_zones, neo4j_session):
 @patch.object(
     cartography.intel.aws.route53,
     "get_zones",
-    return_value=GET_ZONES_SAMPLE_RESPONSE,
```
achantavy, line 320: [question] I'm confused about why we need to change the test this much. Was the data or the setup incorrect to begin with?

janithashri: the old test was buggy..it used two zones with the same name and incorrectly expected them to be linked
It used two zones with the same name (example.com) and was only passing because of the original bug in the application code

achantavy:
lemme test something real quick, dont make any code changes

PR #1767, opened 2025-08-04
chore: Refactor AWS VPC

Daksh1603:
Refactor AWS VPC to the new data model.

Additional Changes Summary:
- New Data Model Schemas added for AWSVpc and AWSCidrBlock (IPv4 and IPv6)
- Changes to Intel module for AWS/ec2/vpc (transform function, clean graph job from schema and load from schema). 
- Updated Tests ( test for sync function was already present up to date ): added assert blocks for ipv4 and ipv6 cidr blocks.
- Minor Schema.md changes (added fields).

- https://github.com/cartography-cncf/cartography/issues/1668

### Checklist

image

 image


achantavy:
Thank you for the PR! need 2 days to review, please ping back if you don't hear back by then

achantavy:
Thank you for the PR! need 2 days to review, please ping back if you don't hear back by then

cartography/intel/aws/ec2/vpc.py
```diff
@@ -126,76 +119,99 @@ def load_ec2_vpcs(
...
+            update_tag,
+        )
+        load_ipv6_cidr_blocks(
+            neo4j_session,
+            ipv6_cidr_blocks,
+            region,
+            current_aws_account_id,
+            update_tag,
+        )
+    cleanup(neo4j_session, common_job_parameters)
```
achantavy, line 217: Make sure to also delete https://github.com/cartography-cncf/cartography/blob/master/cartography/data/jobs/cleanup/aws_import_vpc_cleanup.json and also delete these lines: https://github.com/cartography-cncf/cartography/blob/b95e5de9f57f1b3b80a511dff6a941ad206eabd6/cartography/data/indexes.cypher#L54-L55.

cartography/models/aws/ec2/vpc.py
```diff
@@ -0,0 +1,46 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class VPCNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("VpcId")
+    vpc_id: PropertyRef = PropertyRef("VpcId", extra_index=True)
```
achantavy: This needs to be `vpcid` and not `vpc_id` because of https://github.com/cartography-cncf/cartography/blob/b95e5de9f57f1b3b80a511dff6a941ad206eabd6/cartography/models/aws/ec2/network_acls.py#L34 and https://github.com/cartography-cncf/cartography/blob/b95e5de9f57f1b3b80a511dff6a941ad206eabd6/cartography/models/aws/ec2/security_groups.py#L51.



```suggestion
    vpcid: PropertyRef = PropertyRef("VpcId", extra_index=True)
```

In my testing, heres the graph after running cartography 0.110.0rc1 twice: 

Screenshot 2025-08-05 at 9 28 31 PM



and here's the graph after running this PR twice:


Screenshot 2025-08-05 at 9 32 30 PM


As seen above, we are missing the associations to ec2 security groups and to network acls.

tests/integration/cartography/intel/aws/ec2/test_vpc.py
```diff
@@ -85,4 +88,42 @@ def test_sync_vpc(mock_get_vpcs, neo4j_session):
...
+        ["id", "cidr_block", "association_id", "block_state"],
+    ) == {
+        (
+            "vpc-0a1b2|2001:db8::/32",
+            "2001:db8::/32",
+            "vpc-ipv6-cidr-assoc-0a1b2",
+            "associated",
+        ),
+    }
+
```
achantavy, line 267: Can we please add in some check_rels() tests for VPCs to subnets, VPCs to cidr blocks, VPCs to network acls, VPCs to security groups, etc?

achantavy (changes requested):
Thank you very much for getting started on this refactor! 

I found a bug, see the explanation and the screenshots. 

Also if you can please add in tests to check the vpc rels, that would be a huge help.

Thank you!

Daksh1603:
Hello, was working on a different issue until now, just visited this pr. 

I'll make the respective changes :D

cartography/intel/aws/ec2/vpc.py
Daksh1603: Got it!

cartography/models/aws/ec2/vpc.py
Daksh1603: Got it! this was critical. 

I believe it should have popped up in one of the other unit test that check for relations like network acls. But these unit test, even though sync VPC, dont quite assert their relations with it. 

I feel like I'll address this in a different pr, don't want to put this pr all over the place :D

tests/integration/cartography/intel/aws/ec2/test_vpc.py
Daksh1603: Have added all the relationships defined in the schema for aws module for VPC and CIDR Blocks to the integration test. Except the following relationships : 

- (AWSInternetGateway)-[ATTACHED_TO]->(AWSVpc)
- (AWSVpc)-[RESOURCE]->(AWSTransitGatewayAttachment {resource_type: 'vpc'})
- (AWSVpc)-[TAGGED]->(AWSTag)
- (:ECSTask)-[:NETWORK_INTERFACE]->(:NetworkInterface)

Have added basic comments that may not be needed for the long run, but so that this same long integration test code becomes easier to review for changes.

Will remove once you have reviewed.

cartography/models/aws/ec2/vpc_cidr.py
```diff
@@ -0,0 +1,90 @@
...
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class AWSIPv4CidrBlockNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("Id")
+    vpcid: PropertyRef = PropertyRef("VpcId")  # Extra index ?
```
achantavy:
```suggestion
    vpcid: PropertyRef = PropertyRef("VpcId")
```
Good question, not needed. The data model ensures that all properties used to look up a node have indexes: https://github.com/cartography-cncf/cartography/blob/b44e71a673b7c40d11171e22fe8d904ecc3c7198/cartography/graph/querybuilder.py#L530, and in this case `vpcid` is definitely used for other nodes matching on it to draw relationships.

`extra_indexes` is only needed for when we think a property will be queried for by a script or other human user

cartography/intel/aws/ec2/vpc.py
```diff
@@ -126,76 +119,99 @@ def load_ec2_vpcs(
...
+def cleanup(
+    neo4j_session: neo4j.Session, common_job_parameters: dict[str, Any]
+) -> None:
+    GraphJob.from_node_schema(AWSVpcSchema(), common_job_parameters).run(neo4j_session)
+    GraphJob.from_node_schema(AWSIPv4CidrBlockSchema(), common_job_parameters).run(
+        neo4j_session
+    )
+    GraphJob.from_node_schema(AWSIPv6CidrBlockSchema(), common_job_parameters).run(
+        neo4j_session
+    )
```
achantavy: Let's reverse the order of these cleanups to this:

```python

    GraphJob.from_node_schema(AWSIPv6CidrBlockSchema(), common_job_parameters).run(
        neo4j_session
    )
    GraphJob.from_node_schema(AWSIPv4CidrBlockSchema(), common_job_parameters).run(
        neo4j_session
    )
    GraphJob.from_node_schema(AWSVpcSchema(), common_job_parameters).run(neo4j_session)
```

We normally don't need to worry about order here because nodes are usually attached to sub resources, but CIDR blocks don't have a sub resource so therefore they aren't connected to AWS accounts.

It also doesn't make sense to make their cleanups unscoped because it is possible for a block defined in one AWS account to be referenced in another and we don't want to break that behavior for another account.

cartography/models/aws/ec2/vpc_cidr.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "BLOCK_ASSOCIATION"
+    properties: AWSIPv6CidrBlockToAWSVpcRelProperties = (
+        AWSIPv6CidrBlockToAWSVpcRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class AWSIPv6CidrBlockSchema(CartographyNodeSchema):
```
achantavy, line 90: Same thing about having a comment

cartography/models/aws/ec2/vpc_cidr.py
```diff
@@ -0,0 +1,90 @@
...
+        {"id": PropertyRef("VpcId")}
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "BLOCK_ASSOCIATION"
+    properties: AWSIPv4CidrBlockToAWSVpcRelProperties = (
+        AWSIPv4CidrBlockToAWSVpcRelProperties()
+    )
+
+
+@dataclass(frozen=True)
```
achantavy, line 44: Could be worth adding a comment saying 

there is no sub resource relationship here because a CIDR block can be associated with more than one account and it doesn't make sense to scope it to one. additionally, scoped_cleanup (https://github.com/cartography-cncf/cartography/blob/072e79419dfbcd2e899a42f3a740816d65659964/cartography/models/core/nodes.py#L108) does not make sense to be overriden here.

achantavy (changes requested):
Just need to reverse the order of the cleanup jobs but after that I think we're good to go!

cartography/intel/aws/ec2/vpc.py
Daksh1603: Got it, will make the respective changes!

cartography/models/aws/ec2/vpc_cidr.py
Daksh1603: Thanks for clarifying <3

Daksh1603:
Have made the required changes, I think this should be good to go now. 

Do let me know if any other thing needs to be changed :D

Issue #1769, opened 2025-08-04
Crash in AWS KMS sync - NoneType object is not subscriptable


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

KMS key sync crashes:

```
[2025-08-04T21:36:44Z] INFO cartography.graph.statement: Completed GET_SECRET:SecretsManagerSecret statement #1
[2025-08-04T21:36:44Z] INFO cartography.graph.statement: Completed aws_ec2_iaminstanceprofile statement #1
[2025-08-04T21:36:44Z] INFO cartography.graph.statement: Completed aws_ec2_iaminstanceprofile statement #2
[2025-08-04T21:36:44Z] INFO cartography.graph.job: Finished job aws_ec2_iaminstanceprofile
[2025-08-04T21:36:44Z] INFO cartography.graph.statement: Completed aws_lambda_ecr statement #1
[2025-08-04T21:36:44Z] INFO cartography.graph.statement: Completed aws_lambda_ecr statement #2
[2025-08-04T21:36:44Z] INFO cartography.graph.job: Finished job aws_lambda_ecr
[2025-08-04T21:36:44Z] ERROR sync.1754341200423-d3142a1f: Error during Cartography sync: Traceback (most recent call last):
  File "/app/app/sync.py", line 686, in run_sync
    future.result()
    ~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/app/app/sync.py", line 280, in wrapper
    return fn(*args, **kwargs)
  File "/app/app/sync.py", line 664, in sync_account
    _sync_one_account(
    ~~~~~~~~~~~~~~~~~^
        neo4j_session=neo4j_session,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
        aws_requested_syncs=aws_resource_functions,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
[...]

PR #1770, opened 2025-08-04
fix: Keys with AccessDenied are logged and skipped preventing crash

shyammukund:
- Added null check in transform_kms_key_policies to handle none
- Logging included at get_policy() and get_grants() level 
- unit test testing mock AccessDenied scenarios

- https://github.com/cartography-cncf/cartography/issues/1769

### Checklist

- [x] Include console log trace showing what happened after your changes.
Screenshot 2025-08-04 at 4 02 25 PM

### DIFF
--- a/cartography/intel/aws/kms.py
+++ b/cartography/intel/aws/kms.py
@@ -76,11 +76,10 @@ def get_policy(key: Dict, client: botocore.client.BaseClient) -> Any:
     try:
         policy = client.get_key_policy(KeyId=key["KeyId"], PolicyName="default")
     except ClientError as e:
-        policy = None
         if e.response["Error"]["Code"] == "AccessDeniedException":
-            logger.warning(
-                f"kms:get_key_policy on key id {key['KeyId']} failed with AccessDeniedException; continuing sync.",
-                exc_info=True,
+            policy = None
+            logger.info(
+                f"Skipping KMS key {key['KeyId']} policy due to AccessDenied; will use default policy values"
             )
         else:
             raise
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/kms.py
```diff
@@ -76,11 +76,10 @@ def get_policy(key: Dict, client: botocore.client.BaseClient) -> Any:
...
         policy = client.get_key_policy(KeyId=key["KeyId"], PolicyName="default")
     except ClientError as e:
-        policy = None
         if e.response["Error"]["Code"] == "AccessDeniedException":
-            logger.warning(
-                f"kms:get_key_policy on key id {key['KeyId']} failed with AccessDeniedException; continuing sync.",
-                exc_info=True,
+            policy = None
+            logger.info(
+                f"Skipping KMS key {key['KeyId']} policy due to AccessDenied; will use default policy values"
```
achantavy: I'm flexible here but I don't think we should change the logged message here. If we don't have access, I'd like it to be easy to trace and not have it be buried as an INFO log.

cartography/intel/aws/kms.py
```diff
@@ -113,9 +112,8 @@ def get_grants(key: Dict, client: botocore.client.BaseClient) -> List[Any]:
             grants.extend(page["Grants"])
     except ClientError as e:
         if e.response["Error"]["Code"] == "AccessDeniedException":
-            logger.warning(
-                f'kms:list_grants on key_id {key["KeyId"]} failed with AccessDeniedException; continuing sync.',
-                exc_info=True,
+            logger.info(
```
achantavy: Same here, I don't think we should change the log.

cartography/intel/aws/kms.py
```diff
@@ -187,6 +185,16 @@ def transform_kms_key_policies(
...
 
     for key_id, policy, *_ in policy_alias_grants_data:
+        # Skip keys with null policy since we can't parse it
+        if policy is None:
+            # Add default empty policy data for keys with null policies
+            policy_data[key_id] = {
+                "kms_key": key_id,
+                "anonymous_access": False,
+                "anonymous_actions": [],
+            }
```
achantavy: Double check, but I think we should do this instead:

```suggestion
            policy_data[key_id] = {}
```

As currently written, this says that we know that this key has anonymous_access = False, but that's actually not true -- we just didn't have access to evaluate that. Instead, we should leave those values as None to indicate that we couldn't determine it.

achantavy (changes requested):

Issue #1727, opened 2025-07-28
Kubernetes RBAC


kunaals:
**Kubernetes RBAC*

*Description*:
The Kubernetes module we have so far is great (thanks @ishaanverma !)

Security teams that I've spoken to often have a hard time with Kube RBAC. It's hard to reason about ClusterRoleBindings and RoleBindings especially when you have authZ rules defined in a combination of CSP-native APIs and various configmaps. Classic example is when companies are using EKS Access Entries and the aws-auth configmap within EKS. I think it would be really helpful to have a complete authZ picture across EKS, then we can expand to GKE and AKS

*Relevant Links*:
https://docs.aws.amazon.com/eks/latest/userguide/access-entries.html
https://securitylabs.datadoghq.com/articles/eks-cluster-access-management-deep-dive/#granting-access-to-an-eks-cluster-through-managed-access-policies

PR #1776, opened 2025-08-05
feature: Kubernetes RBAC Ingestion

shyammukund:
-https://github.com/cartography-cncf/cartography/issues/1727

### Checklist

- [x] Include a screenshot showing what the graph looks like after your changes.
image


cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,439 @@
...
+                    "creation_timestamp": get_epoch(rb.metadata.creation_timestamp),
+                    "resource_version": rb.metadata.resource_version,
+                    "role_name": rb.role_ref.name,
+                    "role_kind": rb.role_ref.kind,
+                    "subject_name": subject.name,
+                    "subject_namespace": subject.namespace,
+                    "subject_service_account_id": f"{subject.namespace}/{subject.name}",
+                    "role_id": f"{rb.metadata.namespace}/{rb.role_ref.name}",
+                }
+            )
```
kunaals:
```suggestion
        result.append(
            {
                "id": f"{rb.metadata.namespace}/{rb.metadata.name}",  # different namespaces can have role bindings with same name
                "name": rb.metadata.name,
                "namespace": rb.metadata.namespace,
                "uid": rb.metadata.uid,
                "creation_timestamp": get_epoch(rb.metadata.creation_timestamp),
                "resource_version": rb.metadata.resource_version,
                "role_name": rb.role_ref.name,
                "role_kind": rb.role_ref.kind,
                "service_account_ids": [f"{subject.namespace}/{subject.name}" for subject in service_account_subjects],
                "role_id": f"{rb.metadata.namespace}/{rb.role_ref.name}",
            }
        )
```
two comments here:
1. ID should just be namespace and rolebinding name, that does not relate to subjects
2. this would create a role binding for each subject in the role binding definition. instead, we want one rolebinding with a list of the service accounts that are part of it. this allows a one to many relationship - the role binding can point to many service accounts

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,439 @@
...
+            if rule.resources:
+                for resource in rule.resources:
+                    if resource not in all_resources:
+                        all_resources.append(resource)
+
+            # Extend verbs
+            if rule.verbs:
+                for verb in rule.verbs:
+                    if verb not in all_verbs:
+                        all_verbs.append(verb)
```
kunaals:
```suggestion
            all_api_groups.update(
                {"core" if api_group == "" else api_group for api_group in rule.api_groups or []}
            )
            all_resources.update(rule.resources or [])
            all_verbs.update(rule.verbs or [])
```
Using set comprehensions is more idiomatic in python and less verbose. The "or" catches the case where rule.resources or rules.verbs is None. 

This change applies to other transforms as well

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,439 @@
...
+    """
+    Transform Kubernetes ClusterRoles into a list of dictionaries.
+    Flattens rules into separate api_groups, resources, and verbs lists.
+    """
+    result = []
+    for cluster_role in cluster_roles:
+        # Flatten all rules into combined lists
+        all_api_groups = []
+        all_resources = []
+        all_verbs = []
```
kunaals:
```suggestion
        all_api_groups = set()
        all_resources = set()
        all_verbs = set()
```
I noticed the "if <blah> not in list, append" behavior below so I think you want sets here to prevent duplicates

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,439 @@
...
+                "id": cluster_role.metadata.name,
+                "name": cluster_role.metadata.name,
+                "uid": cluster_role.metadata.uid,
+                "creation_timestamp": get_epoch(
+                    cluster_role.metadata.creation_timestamp
+                ),
+                "resource_version": cluster_role.metadata.resource_version,
+                "api_groups": all_api_groups,
+                "resources": all_resources,
+                "verbs": all_verbs,
```
kunaals:
```suggestion
                "api_groups": list(all_api_groups),
                "resources": list(all_resources),
                "verbs": list(all_verbs),
```
Im not sure whether this is required for the neo4j driver, perhaps it supports sets too? But if it doesn't then you'll have to cast to list

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,439 @@
...
+    result = []
+    for crb in cluster_role_bindings:
+        # Only process ServiceAccount subjects
+        service_account_subjects = [
+            subject
+            for subject in (crb.subjects or [])
+            if subject.kind == "ServiceAccount"
+        ]
+
+        for subject in service_account_subjects:
```
kunaals: same feedback here as transform_role_bindings

cartography/models/kubernetes/clusterroles.py
```diff
@@ -0,0 +1,52 @@
...
+from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class KubernetesClusterRoleNodeProperties(CartographyNodeProperties):
```
kunaals, line 14: I notice that the KubernetesRole object has a cluster_name field, but the KubernetesClusterRole node does not. For both, one can figure out what cluster they're associated with easily using the RESOURCE to cluster relationship, but I think field consistency is important (currently, users have to rely on tracing the RESOURCE edge for KubernetesRole but for KubernetesClusterRole it's as easy as adding a field to the return in the cypher query). Let's choose one approach for consistency - Ishaan was leaning towards have cluster name on everything, in the AWS module though we have the user find out the account ID via the RESOURCE to AWSAccount edge

kunaals (changes requested):
Needs a little sandpapering. The only big question in here is whether we should have cluster_name / cluster_id on every kube node like Ishaan suggested in the spec https://docs.google.com/document/d/17ZGJ9TGSaRinWQn9j1gI5jkbcV7EQ40NvdE_-UU8N74/edit?tab=t.0

My vote would be to follow the convention of AWS and just rely on the fact that every kube node will connect back to it's cluster via the RESOURCE edge, so we dont need cluster_name or cluster_id on each node.

shyammukund:
@kunaals this is updated to reflect your feedback now. 

Couple considerations I thought of while working on the changes. 

KubeNamespace —CONTAINS—> KubeClusterRoleBinding
	- Since this isn’t semantically correct I removed this relationship. I feel like we can see which service accounts in a namespace are affected by a ClusterRoleBinding via the ServiceAccount ←— SUBJECT—- ClusterRoleBinding. 

ClusterName as a property on a Role/ClusterRole node
-I agree w Kunaals opinion that cluster name should not be a property on either of these nodes. Since we have the Cluster—-RESOURCE—> KubeRole/KubeClusterRole we don’t need to have the cluster name as a property.
How do we feel about keeping the cluster_name as a property on subjects though. Because technically we have Cluster —--RESOURCE—> ServiceAccount as well, so maybe we should remove that property on these nodes as well.

kunaals:
@shyammukund regarding

> How do we feel about keeping the cluster_name as a property on subjects though. Because technically we have Cluster —--RESOURCE—> ServiceAccount as well, so maybe we should remove that property on these nodes as well.

Agreed let's remove there too - stick to DRY principles here

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,454 @@
...
+                    "core" if api_group == "" else api_group
+                    for api_group in rule.api_groups or []
+                }
+            )
+            all_resources.update(rule.resources or [])
+            all_verbs.update(rule.verbs or [])
+
+        result.append(
+            {
+                "id": f"{role.metadata.namespace}/{role.metadata.name}",
```
kunaals: If I have 2 clusters in my environment and both have a namespace called "prod" and both of those namespaces have a role called "engineering", would that mess things up? I think we'll have some ID collisions here. I think maybe the id should be the uid so that this is truly unique.

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,454 @@
...
+        ]
+        group_subjects = [
+            subject for subject in (rb.subjects or []) if subject.kind == "Group"
+        ]
+
+        # Only create a RoleBinding node if it has at least one subject
+        if rb.subjects:
+            result.append(
+                {
+                    "id": f"{rb.metadata.namespace}/{rb.metadata.name}",
```
kunaals: Same comment here - i think ID should be uid

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,454 @@
...
+                    "core" if api_group == "" else api_group
+                    for api_group in rule.api_groups or []
+                }
+            )
+            all_resources.update(rule.resources or [])
+            all_verbs.update(rule.verbs or [])
+
+        result.append(
+            {
+                "id": cluster_role.metadata.name,
```
kunaals: same here - UID

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,454 @@
...
+        ]
+        group_subjects = [
+            subject for subject in (crb.subjects or []) if subject.kind == "Group"
+        ]
+
+        # Only create a ClusterRoleBinding node if it has at least one subject
+        if crb.subjects:
+            result.append(
+                {
+                    "id": f"{crb.metadata.name}",  # ClusterRoleBinding is cluster-scoped, so just use name
```
kunaals: same here - UID

cartography/intel/kubernetes/util.py
```diff
@@ -62,6 +63,21 @@ def __init__(
         super().__init__(api_client=api_client)
 
 
+class K8RbacApiClient(RbacAuthorizationV1Api):
```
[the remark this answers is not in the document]
kunaals: I would take a look at this and see if we can easily de-duplicate

tests/integration/cartography/intel/kubernetes/test_rbac.py
```diff
@@ -0,0 +1,683 @@
...
+from tests.integration.util import check_rels
+
+TEST_UPDATE_TAG = 123456789
+TEST_COMMON_JOB_PARAMETERS = {
+    "UPDATE_TAG": TEST_UPDATE_TAG,
+}
+
+
+@pytest.fixture
+def _create_test_cluster(neo4j_session):
```
achantavy: What's this for and can we delete it?

tests/integration/cartography/intel/kubernetes/test_rbac.py
```diff
@@ -0,0 +1,683 @@
...
+    cluster_2_roles_data = transform_roles(KUBERNETES_CLUSTER_2_ROLES_RAW)
+    cluster_2_rb_data = transform_role_bindings(
+        KUBERNETES_CLUSTER_2_ROLE_BINDINGS_RAW, KUBERNETES_CLUSTER_NAMES[1]
+    )
+    cluster_2_cr_data = transform_cluster_roles(KUBERNETES_CLUSTER_2_CLUSTER_ROLES_RAW)
+    cluster_2_crb_data = transform_cluster_role_bindings(
+        KUBERNETES_CLUSTER_2_CLUSTER_ROLE_BINDINGS_RAW, KUBERNETES_CLUSTER_NAMES[1]
+    )
+
+    # Act: Load transformed data
```
achantavy: Let's just test the `sync_*()` functions and not the `load_*()` ones by patching out the gets. Follow the convention from the other tests.

tests/integration/cartography/intel/kubernetes/test_rbac.py
```diff
@@ -0,0 +1,683 @@
...
+    cluster_1_rb_data = transform_role_bindings(
+        KUBERNETES_CLUSTER_1_ROLE_BINDINGS_RAW, KUBERNETES_CLUSTER_NAMES[0]
+    )
+    cluster_1_cr_data = transform_cluster_roles(KUBERNETES_CLUSTER_1_CLUSTER_ROLES_RAW)
+    cluster_1_crb_data = transform_cluster_role_bindings(
+        KUBERNETES_CLUSTER_1_CLUSTER_ROLE_BINDINGS_RAW, KUBERNETES_CLUSTER_NAMES[0]
+    )
+
+    # Act: Load RBAC resources for cluster 1
+    load_service_accounts(
```
achantavy: Same here on sync vs load

tests/integration/cartography/intel/kubernetes/test_rbac.py
```diff
@@ -0,0 +1,683 @@
...
+            "id",
+            "KubernetesClusterRole",
+            "id",
+            "ROLE_REF",
+        )
+        == expected_crb_to_cr_rels
+    )
+
+
+def test_rbac_cleanup(neo4j_session, _create_test_cluster):
```
achantavy: I wouldn't worry about testing the cleanup and I would remove. This part is thoroughly handled by the data model itself and it's a burden to expect all modules to test cleanup.

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,454 @@
...
+from cartography.models.kubernetes.clusterroles import KubernetesClusterRoleSchema
+from cartography.models.kubernetes.rolebindings import KubernetesRoleBindingSchema
+from cartography.models.kubernetes.roles import KubernetesRoleSchema
+from cartography.models.kubernetes.serviceaccounts import KubernetesServiceAccountSchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+def get_service_accounts(k8s_client: K8sClient) -> List[V1ServiceAccount]:
```
achantavy, line 31: Add timeit decorators for all get funcs

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,454 @@
...
+    service_accounts: List[V1ServiceAccount],
+) -> List[Dict[str, Any]]:
+    """
+    Transform Kubernetes ServiceAccounts into a list of dictionaries.
+    """
+    result = []
+    for sa in service_accounts:
+        result.append(
+            {
+                "id": f"{sa.metadata.namespace}/{sa.metadata.name}",
```
achantavy: Double checking: are we confident that this will not cause a service account id collision for a cartography installation with potentially many K8s clusters and many K8s namespaces?

cartography/intel/kubernetes/rbac.py
achantavy: Sharing for posterity: from discussion offline, looks like we're going with `cluster/namespace/name`

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,454 @@
...
+    cleanup_job.run(session)
+
+    cleanup_job = GraphJob.from_node_schema(
+        KubernetesClusterRoleBindingSchema(), common_job_parameters
+    )
+    cleanup_job.run(session)
+
+
+@timeit
+def sync_kubernetes_rbac(
```
achantavy, line 391: Nit: It's slightly cleaner to do

```
+-- sync_kubernetes_rbac
|    +-- _sync_service_accounts
|    |    +-- get...
|    |    +-- transform...
|    |    +-- load..
|    +-- _sync_roles
|    +-- _sync_role_bindings
|    +-- _sync_cluster_roles
|    +-- _sync_cluster_role_bindings
```

cartography/intel/kubernetes/util.py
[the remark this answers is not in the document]
achantavy: I looked at it a bit, I think it's fine as is; it's copying an existing pattern in the code: https://github.com/cartography-cncf/cartography/blob/fc87a9cabf10584b2b5bbb446cd4c914493afae0/cartography/intel/kubernetes/util.py. Our K8s client classes are inheriting from different base classes. The boilerplate is the same but it all is for different purposes.

achantavy (dismissed):

cartography/intel/kubernetes/rbac.py
```diff
@@ -0,0 +1,464 @@
...
+
+    return k8s_paginate(k8s_client.rbac.list_cluster_role_binding)
+
+
+def transform_service_accounts(
+    service_accounts: List[V1ServiceAccount], cluster_name: str
+) -> List[Dict[str, Any]]:
+    """
+    Transform Kubernetes ServiceAccounts into a list of dictionaries.
+    Uses cluster-scoped IDs to prevent collisions across multiple clusters.
```
kunaals, line 65: Thanks for this comment :)

cartography/models/kubernetes/clusterroles.py
kunaals: addressed

docs/root/modules/kubernetes/schema.md
```diff
@@ -154,3 +163,175 @@ Representation of a [Kubernetes Secret.](https://kubernetes.io/docs/concepts/con
     ```
     (:KubernetesNamespace)-[:CONTAINS]->(:KubernetesSecret)
     ```
+
+### KubernetesServiceAccount
+Representation of a [Kubernetes ServiceAccount.](https://kubernetes.io/docs/concepts/security/service-accounts/)
+
+| Field | Description |
+|-------|-------------|
+| id | Identifier for the ServiceAccount derived from namespace and name (e.g. `default/my-service-account`) |
```
kunaals: update to mention cluster, applies to other new Kube objects documented in this PR

kunaals (changes requested):
last change after this its ready to go

kunaals (approved):

Issue #1643, opened 2025-06-19
AWS permission relationships: correctly support AWS roles being able to remote into instances via SSM


achantavy:
*Description*:
> Describe your idea.  Please be detailed.  If a feature request, please
describe the desired behavior, what scenario it enables, and how it
would be used.

With AWS SSM, role A can remotely access EC2 instance X under the following conditions:

(1) if role A has the ssm:StartSession permission on that instance, and 
(2) if X is managed by SSM

(1) can be determined in cartography by looking at the policy statements
(2) is determined by looking if there exists an edge from X to an `SSMInstanceInformation` node. 

There currently isn't a way to use the existing permission_relationships module to define condition (2). It would be helpful to map out these situations. We could use an analysis job, but it'd be nice to use the existing permission relationships construct.

Issue #1639, opened 2025-06-18
AWS permission relationships: S3 GetObject does not work if `resource` refers to path under bucket


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

Say that the permission_relationships.yaml file is 

```yaml
- target_label: S3Bucket
  permissions:
  - S3:GetObject
  relationship_name: CAN_READ
```

and say that we have the following user-policy:

```
user: MyUser

Policy:
resource:
["arn:aws:s3:::my-bucket/*"]

action:
["s3:GetObject"]
```

Expectation:
We expect that there should be an edge drawn from user to the bucket like 
```cypher
(:AWSUser{name:'MyUser'})-[:CAN_READ]->(:S3Bucket{name:'my-bucket'})
```

Actual:
This does not happen because of the `/*` in the policy statement and because of the way that the string matching occurs in the permission relationship evaluation.
GetObject can refer to the "/" and the "/*" underneath a given bucket.

Not sure on the best way to solve. Special casing this just for GetObject feels like it could get unwieldy. Still, worth writing down.

>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]
0.105.0

PR #1785, opened 2025-08-06
Feat: Conditional target relations and custom resource arn schema for permission_mapping module [WIP]

Daksh1603:
In order to address these issues and extend functionality of the permission relationships, I have devised some solution fragments that can help for this.

2 OPTIONAL but new properties for permission relationships (YAML) : 

- conditional_target_relations: Basically a list of node relations to ensure that only map a relation when the target also has there relations protruding out. ( ISSUE : https://github.com/cartography-cncf/cartography/issues/1643 ).
Other example use case: Map relation from principal to resource that have DLQ.
- resource_arn_schema : By default, the cypher queries map relations and fetch resources based on arn property of the target resource. However, not every target resource node has arn. I stumbled on a  error when trying to create a relation for my ec2 instances. 

image

In order to define a custom target resource arn join to resemble your own pricinipal policies arn: added this field. This potentially addresses  https://github.com/cartography-cncf/cartography/issues/1639 ( have added a test unit and tried out on my end). The documentation for this has been updated in docs/root/modules/aws/permissions-mapping.md .

Additional Changes: 
- Added logger warning and tackled permission_relationships failing for nodes without arn property(Mentioned Error). Did not change arn property of ec2 instance since did not want this pr to go all over the place.
-  Additional regex pattern to devise property dict (yaml) -> arn string -> then back to property dict(yaml).
- These changes are only added functionality and previous layout, structuring of permissions,functionality and tests are entirely preserved.

Possibly to do:
- Update extract_properties_from_arn() regex pattern, or implement a better node property pass (will have to make changes to existing unit tests and functions a bit)
- Wildcards and SingleCharacters chars (* or ?) are escaped and dont act as wildcards -> for resource_arn_schema. 

- Addresses https://github.com/cartography-cncf/cartography/issues/1639
- Addresses https://github.com/cartography-cncf/cartography/issues/1643

### Checklist

Before:
image
(Tweaked neo4j queries in permission_relationships.py for temporarily  trying out on ec2 instances.)

After: 
image

image

### DIFF
--- a/cartography/data/permission_relationships.yaml
+++ b/cartography/data/permission_relationships.yaml
@@ -27,15 +27,3 @@
   - redshift:GetClusterCredentials
   - redshift:JoinGroup
   relationship_name: CAN_ADMINISTER
-
-
-# Map principals that can start SSM sessions on EC2 instances.
-# The resource_arn_schema field defines how to construct the resource ARN from node properties.
-# Properties in {{}} will be substituted with actual node property values.
-- target_label: EC2Instance
-  permissions:
-  - ssm:StartSession
-  relationship_name: CAN_SSM_ACCESS
-#  conditional_target_relations:
-#   - HAS_INFORMATION
-  resource_arn_schema: "arn:aws:ec2:{{region}}:*:instance/{{instanceid}}"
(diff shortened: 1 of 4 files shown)


Daksh1603:
Closing this since, complexity at user end and uninteded/unnoticed relations increases with schema implementation. I don't trust regex fully yet and would put it to minimum use.

PR #1787, opened 2025-08-06
feat: EntraUser sign on to AWS Identity Center (#1779)

### LINKED ISSUE #1779: Sign in with Entra ID to AWS
## Description
Support linking Microsoft Entra ID service principals to AWS Identity Center instances based on SAML federation.

The servicePrincipal object contains redirect URIs that include the AWS Identity Center instance ID (`d-xxxxxxx`). 

We should parse these URIs to extract the instance ID and create a `FEDERATES_TO` relationship from the service principal to a new `AWSIdentityCenterInstance` node. This enables mapping identity flows from Entra to AWS and supports environments with multiple federated Identity Center instances.

### Step 1

An instance of an entra application deployed in a tenant is called a service principal.

```
(:EntraApplication)-[:SPN]->(:ServicePrincipal)
```

### Step 2

The spn will have one or more redirect URIs, and one of those may map to an AWS identity center instance (after parsing out the d-**** string from the URI). If this matches up, we'd be able to make this association:

```
(:ServicePrincipal)-[:FEDERATES_TO]->(:AWSIdentityCenterInstance)
```

### Step 3

This would enable us to link Entra ID to AWS identity federation in cartography. Specifically if we have this path,

```(e:EntraUser)--(:EntraApplication)-[:SPN]->(:ServicePrincipal)-[:FEDERATES_TO]->(:AWSIdentityCenterInstance)--(a:AWSSSOUser)
```

where `e.user_principal_name = u.user_name`, then we can draw

```
(e:EntraUser)-[:CAN_SSO_TO]->(a:AWSSSOUser)
```

**Relevant Links**:  
https://docs.aws.amazon.com/singlesignon/latest/userguide/idp-microsoft-entra.html  
https://learn.microsoft.com/en-us/graph/api/resources/serviceprincipal

achantavy:
short_1

Screenshot 2025-08-06 at 3 47 12 PM

- https://github.com/cartography-cncf/cartography/issues/1779

### Checklist

image


tests/unit/cartography/intel/entra/test_applications.py
```diff
@@ -45,119 +32,3 @@ def test_transform_app_role_assignments():
     assert assignment3["principal_display_name"] == "Finance Team"
     assert assignment3["resource_display_name"] == "Finance Tracker"
     assert assignment3["principal_type"] == "Group"
-
-
-@patch("cartography.intel.entra.applications.load")
```
achantavy: removing these brittle and not very useful unit tests

cartography/intel/entra/applications.py
```diff
@@ -235,10 +245,48 @@ def transform_app_role_assignments(
     return result
 
 
+def transform_service_principals(
+    service_principals: list[ServicePrincipal],
```
achantavy: @shyammukund this is what I was talking about at lunch: use `list[SpecificType]` wherever possible (like here with the Entra SDK because it's available to us. This way we can do things like `spn.login_url` instead of `getattr(spn, 'login_url', None)`

cartography/intel/entra/applications.py
```diff
@@ -1,19 +1,27 @@
 import logging
+import re
 from typing import Any
-from typing import Dict
```
jychp: <3

cartography/intel/entra/applications.py
```diff
@@ -313,14 +377,100 @@ def cleanup_app_role_assignments(
     ).run(neo4j_session)
 
 
+@timeit
+def cleanup_service_principals(
```
jychp: nitpicking: I think all the cleanup logic can be in a single function as they have the same purpose and same parameters

cartography/intel/entra/applications.py
```diff
@@ -347,20 +497,14 @@ async def sync_entra_applications(
...
+    # Applications, app roles, and service principals
+    await sync_entra_apps_roles_spns(neo4j_session, client, update_tag, tenant_id)
 
-    # Load applications and assignments
-    load_applications(neo4j_session, transformed_applications, update_tag, tenant_id)
-    load_app_role_assignments(
-        neo4j_session, transformed_assignments, update_tag, tenant_id
-    )
+    # Attach sign-on relationships
+    sync_entra_to_aws_identity_center(neo4j_session, update_tag, tenant_id)
```
jychp: I don't really like that multiple `sync_*` functions have different behaviors:

* `sync_entra_applications`: performs cleanup and orchestrates other syncs
* `sync_entra_apps_roles_spns` and `sync_entra_to_aws_identity_center`: don't perform cleanup

For consistency, we should either:

* move the cleanup logic into each sub-function
* or (my preferred option) make the sub-functions private (`_sync_*`) to highlight the difference in responsibility

achantavy:
I've got this working on my local setup, going to test it e2e soon and then merge when that works

achantavy:
Looks good, going to merge this in.

It's also possible to link to additional federated apps via the entra service principal's login url. The code can be extended there through the federation folder

cartography/intel/entra/applications.py
achantavy: did the first option

cartography/intel/entra/applications.py
achantavy: done

achantavy:
>I’m a bit uncomfortable having AWS-specific logic inside an Entra module.

Created a federation/ folder to help keep these cross-cuts more organized

PR #1791, opened 2025-08-07
refactor(gcp/storage): align module with modern schema-based patterns

janithashri:
### Summary
This PR refactors the GCP Storage intel module from a legacy, manual Cypher query pattern to the modern, schema-based data model.

This change improves the module's maintainability, robustness, and consistency with the project's current standards. It replaces all manual Cypher queries with the declarative load() function and introduces a clean get -> transform -> load -> cleanup sync pattern. It also resolves a potential data integrity bug in the old code that could create duplicate nodes.

### Related issues or links
This refactoring is part of the ongoing effort to modernize the codebase and align all modules with the best practices outlined in the developer guide.

### Checklist

[x] Update/add unit or integration tests.

[x] (Please add your screenshot here)

[ x] (Please add your console log trace here)

[ ] Update the [schema](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) and [readme](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md).

[ ] Use the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).

Screenshot:
Screenshot (438)

Screenshot (437)

Screenshot (436)

### DIFF
--- a/cartography/models/gcp/storage.py
+++ b/cartography/models/gcp/storage.py
@@ -1,0 +1,115 @@
+import logging
+from dataclasses import dataclass
+
+from cartography.models.core.common import PropertyRef
+from cartography.models.core.nodes import CartographyNodeProperties, CartographyNodeSchema, ExtraNodeLabels
+from cartography.models.core.relationships import (
+    CartographyRelProperties, CartographyRelSchema, LinkDirection,
+    make_target_node_matcher, TargetNodeMatcher, OtherRelationships,
+)
+
+logger = logging.getLogger(__name__)
+
+
+# ### Node Properties ###
+@dataclass(frozen=True)
+class GCPStorageBucketProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id", extra_index=True)
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+    project_number: PropertyRef = PropertyRef("project_number", set_in_kwargs=True)
+    kind: PropertyRef = PropertyRef("kind")
+    location: PropertyRef = PropertyRef("location")
+    location_type: PropertyRef = PropertyRef("location_type")
+    meta_generation: PropertyRef = PropertyRef("meta_generation")
+    storage_class: PropertyRef = PropertyRef("storage_class")
+    time_created: PropertyRef = PropertyRef("time_created")
+    updated: PropertyRef = PropertyRef("updated")
+    self_link: PropertyRef = PropertyRef("self_link")
+    etag: PropertyRef = PropertyRef("etag")
+    owner_entity: PropertyRef = PropertyRef("owner_entity")
+    owner_entity_id: PropertyRef = PropertyRef("owner_entity_id")
+    iam_config_bucket_policy_only: PropertyRef = PropertyRef("iam_config_bucket_policy_only")
+    versioning_enabled: PropertyRef = PropertyRef("versioning_enabled")
+    retention_period: PropertyRef = PropertyRef("retention_period")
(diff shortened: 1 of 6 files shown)


cartography/models/gcp/storage.py
```diff
@@ -0,0 +1,115 @@
...
+        {"id": PropertyRef("project_number", set_in_kwargs=True)},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: GCPBucketLabelToGCPProjectRelProperties = GCPBucketLabelToGCPProjectRelProperties()
+
+
+# ### Main Schema ###
+@dataclass(frozen=True)
+class GCPBucketLabelSchema(CartographyNodeSchema):
```
jychp: Please split the two node schema in two separate files

cartography/models/gcp/storage.py
```diff
@@ -0,0 +1,115 @@
...
+    make_target_node_matcher, TargetNodeMatcher, OtherRelationships,
+)
+
+logger = logging.getLogger(__name__)
+
+
+# ### Node Properties ###
+@dataclass(frozen=True)
+class GCPStorageBucketProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id", extra_index=True)
```
jychp: id do not need extra_index flag

cartography/models/gcp/storage.py
```diff
@@ -0,0 +1,115 @@
...
+    kind: PropertyRef = PropertyRef("kind")
+    location: PropertyRef = PropertyRef("location")
+    location_type: PropertyRef = PropertyRef("location_type")
+    meta_generation: PropertyRef = PropertyRef("meta_generation")
+    storage_class: PropertyRef = PropertyRef("storage_class")
+    time_created: PropertyRef = PropertyRef("time_created")
+    updated: PropertyRef = PropertyRef("updated")
+    self_link: PropertyRef = PropertyRef("self_link")
+    etag: PropertyRef = PropertyRef("etag")
+    owner_entity: PropertyRef = PropertyRef("owner_entity")
```
jychp: It seems you added fields that were not initially present. Could you explain why, and what these fields are for?

cartography/models/gcp/storage.py
```diff
@@ -0,0 +1,115 @@
...
+logger = logging.getLogger(__name__)
+
+
+# ### Node Properties ###
+@dataclass(frozen=True)
+class GCPStorageBucketProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id", extra_index=True)
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+    project_number: PropertyRef = PropertyRef("project_number", set_in_kwargs=True)
+    kind: PropertyRef = PropertyRef("kind")
```
jychp: The `bucket_id` seems to be missing, any reason why you do not keep it ?

janithashri: I removed the bucket_id property because it was redundant with the node's primary id property.

The original manual query was setting both id and bucket_id to the same value, so as part of the refactor, I cleaned it up to use only the standard id field as the unique identifier.

Happy to add it back if needed

cartography/models/gcp/storage.py
janithashri: I double-checked, it turns out the new GCPStorageBucketProperties schema is just formally defining the same set of properties that the old transform function was already creating.

janithashri:
Hi @jychp, I've addressed all the requested changes, including splitting the model files. All checks are now passing, and this should be ready for another review.

PR #1794, opened 2025-08-08
feat: add `_module_name` and `_module_version` metadata to graph entities

jychp:
### Summary

This PR introduces automatic injection of traceability metadata into all Neo4j nodes and relationships created by Cartography. Two new properties are now added by default:

* `_module_name`: The source intel module that created the entity (e.g., `cartography:aws`, `cartography:github`).
* `_module_version`: The Cartography package version used during ingestion.

⚠️ WARNING: This feature rely on the ORM and new data model, this metadata will only be present on already migrated data (cf: https://github.com/cartography-cncf/cartography/issues/1668)

### Context

This change improves graph observability and debugging by allowing maintainers and users to:
* Identify the exact intel module that created or last updated a given node/relationship.
* Track the Cartography version used during data ingestion, facilitating reproducibility and troubleshooting.

This feature also gracefully handles external/custom schemas by tagging them as `unknown:<module_path>` and issuing a warning.

_Note: this feature has been requested to allow re-usability of our neo4j database in other tools._

### Changes proposed

* **Metadata injection utilities:**

  * `_get_cartography_version()`: Retrieves the installed Cartography package version using `importlib.metadata.version()`, with a fallback to `"dev"` for development environments or missing package metadata.
  * `_get_module_from_schema()`: Extracts the intel module name from the schema path and formats it as `cartography:<module>`.

    * For non-Cartography modules, sets `_module_name` to `unknown:<module_path>` and logs a warning.
* **Integration in data loading pipeline** to ensure `_module_name` and `_module_version` are automatically added to all ingested nodes and relationships.

### How to test this PR

1. Run any Cartography ingestion (or use demo environment `uv run python3 -m demo`)
2. In Neo4j Browser, query:

   ```cypher
   MATCH (n) RETURN DISTINCT n._module_name, n._module_version LIMIT 10;
   MATCH ()-[r]->() RETURN DISTINCT r._module_name, r._module_version LIMIT 10;
   ```
3. Verify that:
   * All entities have `_module_name` and `_module_version` set.
   * Values are correct for known modules.

### Screenshots

Screenshot From 2025-08-08 13-41-39

### Reference

https://cloud-native.slack.com/archives/C080M2LRLDA/p1754562787066309


Siryu6:
Thanks for reactivity on this feature 🙏

tests/unit/cartography/graph/test_matchlink.py
```diff
@@ -18,15 +20,19 @@ def test_build_matchlink_query():
     Test that build_matchlink_query() generates valid Cypher queries.
     """
     rel_schema = PrincipalToS3BucketPermissionRel()
+    module_version = _get_cartography_version()
```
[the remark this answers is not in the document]
achantavy: I _think_ I agree here. Here's what I think can help:

1. can we have a separate unit test just for `_get_cartography_version()`? (I see that tests for `_get_module_from_schema()` already exist)


<continued in next comment>

tests/unit/cartography/graph/test_matchlink.py
```diff
@@ -18,15 +20,19 @@ def test_build_matchlink_query():
...
+    expected = f"""
         UNWIND $DictList as item
-            MATCH (from:AWSPrincipal{principal_arn: item.principal_arn})
-            MATCH (to:S3Bucket{name: item.BucketName})
+            MATCH (from:AWSPrincipal{{principal_arn: item.principal_arn}})
+            MATCH (to:S3Bucket{{name: item.BucketName}})
             MERGE (from)-[r:CAN_ACCESS]->(to)
             ON CREATE SET r.firstseen = timestamp()
             SET
+                r._module_name = "{module_name}",
```
achantavy: 2. instead of changing `expected` to a `f""` string, can we just hardcode in the module name like

```
r._module_name = "cartography:scaleway",
```

tests/unit/cartography/graph/test_matchlink.py
```diff
@@ -18,15 +20,19 @@ def test_build_matchlink_query():
...
+                r._module_version = "{module_version}",
```
achantavy: I guess for module_version this is less important since that changes and adding a patch feels a bit silly. Eh I dont have a strong opinion here

tests/unit/cartography/graph/test_querybuilder_metadata.py
```diff
@@ -0,0 +1,26 @@
...
+from cartography.models.aws.identitycenter.awspermissionset import (
+    RoleAssignmentAllowedByMatchLink,
+)
+from cartography.models.tailscale.user import TailscaleUserToTailnetRel
+from tests.data.graph.querybuilder.sample_models.simple_node import SimpleNodeSchema
+
+
+def test_querybuilder_metadata_module_name():
+    # Regular Node Schema
+    assert _get_module_from_schema(CloudTrailTrailSchema()) == "cartography:aws"
```
achantavy, line 12: @Siryu6 How much granularity is required? Do we want to go down to the resource e.g. https://github.com/cartography-cncf/cartography/blob/b44e71a673b7c40d11171e22fe8d904ecc3c7198/cartography/intel/aws/resources.py#L63 ?

cartography/graph/querybuilder.py
```diff
@@ -312,6 +320,8 @@ def _build_attach_additional_links_statement(
             node_var=node_var,
             rel_var=rel_var,
             RelMerge=rel_merge,
+            module_name=_get_module_from_schema(link),
```
achantavy, line 323: Idea: another option could be to create a public property function on the node/rel schema https://github.com/cartography-cncf/cartography/blob/b44e71a673b7c40d11171e22fe8d904ecc3c7198/cartography/models/core/nodes.py#L52 called `module_name()`. This way you could reference it as `DataType.module_name`. Might be overthinking it though.

tests/unit/cartography/graph/test_matchlink.py
jychp: Added patch, that make sense and do not add too much complexity.

cartography/graph/querybuilder.py
jychp: This would require:

* Duplicating the logic in both the `Node` and `Rel` classes (since they don’t inherit from the same base class), or modifying the current inheritance structure
* Implementing our own `@property` for classes, as Python does not provide one by default, or instantiating an object each time

While I find the approach elegant, I suggest postponing it until we actually need it — by then, we will likely have additional requirements to account for.

jychp:
I tried to write a meaningful unit test, but since the version is dynamic we’d have to patch `importlib.metadata.version` to return a known string.
Given that the only purpose of `_get_cartography_version` is to call that function and return `"dev"` on error (and I don’t have a clean way to simulate that error), I don’t think the test adds value.

@achantavy if you see a solid way to test this, I’m all ears.

PR #1796, opened 2025-08-10
feat: Add ApiGatewayDeployment and Lambda Permission resource in AWS

d-aggarwal:
### Related issues or links

### Checklist


cartography/intel/aws/apigateway.py
```diff
@@ -40,6 +46,40 @@ def get_apigateway_rest_apis(
     return apis
 
 
+@timeit
+@aws_handle_regions
+def get_rest_api_ids(
```
jychp, line 49: I don’t think you need a function here, it’s just a simple list generation, and having a function makes the code less readable.

Even if you do want to keep it as a function, the `@timeit` decorator isn’t needed (execution time is negligible), and you also don’t need the region provided by the `@aws_handle_region` decorator.

cartography/models/aws/lambda_function/permission.py
```diff
@@ -0,0 +1,71 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class AWSLambdaPermissionNodeProperties(CartographyNodeProperties):
```
jychp: Do we really want a separate, independent object here?
It seems to:
* have a strict 1:1 relationship with a Lambda function
* lack its own ID or ARN, which suggests it’s not an independent object in AWS

Can we instead attach this information directly to the Lambda function node?

jychp (changes requested):
Open discussion about having a dedicated node for LambdaFunction permissions

cartography/intel/aws/apigateway.py
d-aggarwal: Thank you for your review. I have removed both the decorator, instead of passing full data of api's  that would create extra complexity, created this function to pass only ids.

cartography/models/aws/lambda_function/permission.py
d-aggarwal: I have attached the respective info directly to function node.

d-aggarwal:
Please Consider, i haven't tested it with AWS Account , since mine has crossed the free tier limit.

achantavy (changes requested):
@d-aggarwal can we split this to 2 PRs -- one for lambda and one for apigatewaydeployment? The size of this PR is already small which is very good, but it would be cleaner and easier to manage if each PR is constrained to one source type (or at least _related_ resource types), even if those PRs are bigger.  lambdas and apigateways don't have anything to do with each other so it's some what odd to have in one pr

d-aggarwal:
> @d-aggarwal can we split this to 2 PRs -- one for lambda and one for apigatewaydeployment? The size of this PR is already small which is very good, but it would be cleaner and easier to manage if each PR is constrained to one source type (or at least _related_ resource types), even if those PRs are bigger. lambdas and apigateways don't have anything to do with each other so it's some what odd to have in one pr

Thank you for your review, I have created two separated PR's for both resources.

PR #1797, opened 2025-08-10
feat: Add GlueJob as resource in AWS Module

d-aggarwal:
### Related issues or links

### Checklist

### DIFF
--- a/cartography/models/aws/glue/job.py
+++ b/cartography/models/aws/glue/job.py
@@ -1,0 +1,73 @@
+from dataclasses import dataclass
+
+from cartography.models.core.common import PropertyRef
+from cartography.models.core.nodes import CartographyNodeProperties
+from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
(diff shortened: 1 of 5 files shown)


cartography/models/aws/glue/job.py
```diff
@@ -0,0 +1,69 @@
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class GlueJobToGlueConnectionRel(CartographyRelSchema):
+    target_node_label: str = "GlueConnection"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("Connections", one_to_many=True)},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
```
jychp: Isn’t this actually a job that uses a connection?
IMHO, the relationship should be oriented in the other direction.

jychp (changes requested):
LGTM but I think we should change the relationship direction

cartography/models/aws/glue/job.py
d-aggarwal: You're right, will update the relationship

PR #1798, opened 2025-08-11
test: add sync test for GCP iam in integration

wahid18-maqs:
I added a new `test_sync()` in `test_iam.py` that:
Creates the test project.
Loads roles and service accounts using the the tests
Asserts that the expected nodes and relationships are created

### Checklist


wahid18-maqs:
@jych Thanks for the feedback earlier.
I’ve updated the test so it now directly calls `sync()` and runs the full integration flow fetching roles and service accounts from our static test data, loading them into Neo4j, and performing cleanup. This covers both nodes and relationships in one pass, exercising the real code paths inside `sync()` rather than re-implementing them.

tests/integration/cartography/intel/gcp/test_iam.py
```diff
@@ -96,3 +96,119 @@ def test_load_gcp_service_accounts(neo4j_session):
         )
         == expected_rels
     )
+
+def test_sync_gcp_iam(neo4j_session):
```
jychp, line 101: Please update the existing test instead of defining a new one to avoid duplicated tests.

tests/integration/cartography/intel/gcp/test_iam.py
```diff
@@ -96,3 +96,119 @@ def test_load_gcp_service_accounts(neo4j_session):
...
+        predefined_roles = [
+            role for role in tests.data.gcp.iam.LIST_ROLES_RESPONSE["roles"]
+            if role["name"].startswith("roles/")
+        ]
+        return {"roles": predefined_roles}
+
+    def list_next(*args, **kwargs):
+        return None
+
+    # List methods that wrap the execute functions
```
jychp: In Cartography, we use a simpler pattern: we patch the module’s get function.

This allows us to keep tests straightforward and maintainable while preserving an acceptable level of coverage.

jychp (changes requested):
I think we can simplify this while keeping the tradeoff acceptable.

tests/integration/cartography/intel/gcp/test_iam.py
achantavy: To make the diff cleaner, it might be better to create a new test for the sync() and then go back later and remove the old tests that cover only the loads(). I'm ok with a new test.

tests/integration/cartography/intel/gcp/test_iam.py
```diff
@@ -96,3 +96,119 @@ def test_load_gcp_service_accounts(neo4j_session):
...
+
+    def predefined_roles_execute():
+        predefined_roles = [
+            role for role in tests.data.gcp.iam.LIST_ROLES_RESPONSE["roles"]
+            if role["name"].startswith("roles/")
+        ]
+        return {"roles": predefined_roles}
+
+    def list_next(*args, **kwargs):
+        return None
```
achantavy: Why is this PR defining functions inside a test function? This is way overly complicated and feels like it was generated by an LLM and not reviewed. Like Jeremy said, patch out the get()s.

Please make sure that the code matches other existing code in the project and fits the arrange+act+assert pattern shown elsewhere. Refer your LLM to read the AGENTS.md and you will get much better results.

achantavy (changes requested):

wahid18-maqs:
Thanks, I'll address all the points you raised and push the changes soon.

wahid18-maqs:
- **Added:** Simple patches on the `get()` functions using existing test data  
- **Kept:** All the assertion logic testing graph outcomes  
- **Simplified:** `iam_client` parameter can be `None` since `get` functions are mocked

wahid18-maqs:
Hi @achantavy Sir, this time I have followed `AGENTS.md` guidelines for testing outcomes over implementation details. 
Apologies for the previous commit.!!

jychp:
Closed in favor of https://github.com/cartography-cncf/cartography/pull/2255

PR #1805, opened 2025-08-12
fix(github): GitHub Dependency Versions 

shyammukund:
- Removed repo_name property
- Replaced version property with requirements to avoid parsing
- Changed Dependency id to be canoncial_name|requirements instead of canonical_name|version
- Tests updated to reflect changes

### Checklist

- [x] New Dependency Node Properties
Screenshot 2025-08-12 at 2 46 52 PM

### DIFF
--- a/cartography/intel/aws/rds.py
+++ b/cartography/intel/aws/rds.py
@@ -9,6 +9,7 @@ import neo4j
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.models.aws.rds.cluster import RDSClusterSchema
+from cartography.models.aws.rds.event_subscription import RDSEventSubscriptionSchema
 from cartography.models.aws.rds.instance import RDSInstanceSchema
 from cartography.models.aws.rds.snapshot import RDSSnapshotSchema
 from cartography.models.aws.rds.subnet_group import DBSubnetGroupSchema
@@ -129,6 +130,38 @@ def load_rds_snapshots(
     load(
         neo4j_session,
         RDSSnapshotSchema(),
+        data,
+        lastupdated=aws_update_tag,
+        Region=region,
+        AWS_ID=current_aws_account_id,
+    )
+
+
+@timeit
+@aws_handle_regions
+def get_rds_event_subscription_data(
+    boto3_session: boto3.session.Session,
+    region: str,
+) -> List[Dict[str, Any]]:
+    client = boto3_session.client("rds", region_name=region)
+    paginator = client.get_paginator("describe_event_subscriptions")
+    subscriptions = []
+    for page in paginator.paginate():
+        subscriptions.extend(page["EventSubscriptionsList"])
+    return subscriptions
+
+
+@timeit
+def load_rds_event_subscriptions(
+    neo4j_session: neo4j.Session,
+    data: List[Dict],
+    region: str,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    load(
+        neo4j_session,
+        RDSEventSubscriptionSchema(),
         data,
         lastupdated=aws_update_tag,
         Region=region,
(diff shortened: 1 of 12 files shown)


cartography/intel/github/repos.py
```diff
@@ -658,11 +655,12 @@ def _transform_dependency_graph(
             # Create ecosystem identifier
             ecosystem = package_manager.lower() if package_manager else "unknown"
 
-            # Create simple dependency ID using canonical name and version
+            # Create simple dependency ID using canonical name and requirements
             # This allows the same dependency to be shared across multiple repos
+            requirements_for_id = requirements.strip() if requirements else ""
```
kunaals:
```suggestion
            requirements_for_id = (requirements or "").strip()
```
Super minor but this is slightly more pythonic using a ternary operator and doesn't repeat the word requirements in the same line

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -401,25 +401,25 @@ def test_sync_github_dependencies_end_to_end(neo4j_session):
...
     spring_core_id = "org.springframework:spring-core|5.3.21"
 
     # Assert - Test that new GitHub dependency graph nodes were created
     # Note: Database also contains legacy Python dependencies, so we check subset
     expected_github_dependency_nodes = {
         (react_id, "react", "18.2.0", "npm"),
-        (lodash_id, "lodash", None, "npm"),
-        (django_id, "django", "4.2.0", "pip"),
+        (lodash_id, "lodash", "^4.17.21", "npm"),
+        (django_id, "django", "==4.2.0", "pip"),
```
kunaals: For pip deps doesn't github return like "= 4.2.0"? Let's make sure the test data accurately represents what is returned from the API

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -401,25 +401,25 @@ def test_sync_github_dependencies_end_to_end(neo4j_session):
     # _ensure_local_neo4j_has_test_data has already called sync, now we test that the sync worked. Mock GitHub API data should
     # be transofrmed and in the Neo4j database.
 
-    # Create expected IDs with simple format: canonical_name|version
+    # Create expected IDs with format: canonical_name|requirements
     repo_url = "https://github.com/cartography-cncf/cartography"
     react_id = "react|18.2.0"
-    lodash_id = "lodash"
```
kunaals, line 407: I think the intent of having no version here was to test the case where requirements does not exist. Let's keep one of these examples without requirements to test this case

docs/root/modules/github/schema.md
```diff
@@ -326,13 +326,12 @@ Represents a software dependency from GitHub's dependency graph manifests. This
 |-------|-------------|
 | firstseen | Timestamp of when a sync job first discovered this node |
 | lastupdated | Timestamp of the last time the node was updated |
-| **id** | Simple identifier: `{canonical_name}|{version}` or `{canonical_name}` - same dependency shared across repos |
+| **id** | Identifier: `{canonical_name}|{requirements}` when a requirements string exists, otherwise `{canonical_name}` |
 | **name** | Canonical name of the dependency (ecosystem-specific normalization) |
 | **original_name** | Original name as specified in the manifest file |
-| **version** | Pinned version if specified, otherwise null |
+| **requirements** | Unparsed requirement string from the manifest (e.g., `"18.2.0"`, `"==4.2.0"`, `"^4.17.21"`, `"1.*.*"`) |
```
kunaals, line 332:
```suggestion
| **requirements** | Unparsed requirement string from the manifest (e.g., `"18.2.0"`, `"= 4.2.0"`, `"^4.17.21"`, `"1.*.*"`) |
```

kunaals (commented):
@achantavy @jychp this is a breaking change - we get rid of the `version` field on Dependency nodes. We now represent requirements exactly as they come back from the GitHub dependency graph in the `requirements` field on said node. 

The `version` field is still there on Dependency::PythonLibrary, but if someone has a query relying on just the Dependency node label, this breaks things for them.

Should we still include the `version` field but just copy the `requirements` value to it? Or potentially just stick to calling the field `version`, even though the GitHub API calls it requirements?

achantavy:
@kunaals 

I think we're ok to break.`Dependency` from a GitHub `DependencyGraphManifest` is new, so if we change the properties now there won't be many people (if at all) upset by the change.

I think the plan would be to 

- For `Dependency::PythonLibrary` keep `version` there unchanged
- For `Dependency` from dependencygraphmanifests, remove `version` and just use `requirements`.

kunaals (approved):
LGTM

PR #1809, opened 2025-08-13
test(github): improving test coverage for github module

SachdevJai:
refactored test_repos.py to follow the patch get and test_sync pattern for testing
added tests to check if repo properties are being transformed and loaded correctly
added missing tests for MAINTAINS and TRIAGE relationships for teams

### Checklist

```
  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
====================================================================== test session starts ======================================================================
platform linux -- Python 3.13.2, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/zai/lfx/cartography
configfile: setup.cfg
plugins: rerunfailures-15.1, anyio-4.9.0, mock-3.14.0, asyncio-0.26.0, cov-6.2.1
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 4 items                                                                                                                                               

tests/integration/cartography/intel/github/test_repos.py .                                                                                                [ 25%]
tests/integration/cartography/intel/github/test_teams.py .                                                                                                [ 50%]
tests/integration/cartography/intel/github/test_users.py ..                                                                                               [100%]

======================================================================= warnings summary ========================================================================
tests/integration/cartography/intel/github/test_repos.py: 4 warnings
tests/integration/cartography/intel/github/test_teams.py: 6 warnings
tests/integration/cartography/intel/github/test_users.py: 9 warnings
  /home/zai/lfx/cartography/cartography/client/core/tx.py:232: DeprecationWarning: write_transaction has been renamed to execute_write
    neo4j_session.write_transaction(

tests/integration/cartography/intel/github/test_repos.py: 23 warnings
tests/integration/cartography/intel/github/test_teams.py: 33 warnings
tests/integration/cartography/intel/github/test_users.py: 10 warnings
  /home/zai/lfx/cartography/cartography/graph/statement.py:132: DeprecationWarning: write_transaction has been renamed to execute_write
    summary: neo4j.ResultSummary = session.write_transaction(

tests/integration/cartography/intel/github/test_repos.py::test_sync_repos
tests/integration/cartography/intel/github/test_teams.py::test_sync_github_teams
tests/integration/cartography/intel/github/test_users.py::test_sync_with_cleanups
  /home/zai/lfx/cartography/.venv/lib/python3.13/site-packages/neo4j/_sync/driver.py:547: DeprecationWarning: Relying on Driver's destructor to close the session is deprecated. Please make sure to close the session. Use it as a context (`with` statement) or make sure to call `.close()` explicitly. Future versions of the driver will not close drivers automatically.
    _deprecation_warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================ 4 passed, 88 warnings in 2.53s =================================================================
```


tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -8,423 +10,398 @@
...
+        archived,
+        locked,
+        giturl,
+        url,
+        sshurl,
+        createdat,
+        updatedat,
+    ) = sample_repo
+
+    # Validate all properties have expected values
```
shyammukund: You can use check_nodes() here instead of having an assert to check that the every field on the node

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -8,423 +10,398 @@
...
+        "GitHubUser",
+        "id",
+        "OWNER",
+        rel_direction_right=True,
+    )
+    assert expected_repo_user_owner_rels.issubset(actual_repo_user_owner_rels)
 
 
-def test_repository_to_branches(neo4j_session):
+def _test_branch_relationships(neo4j_session):
```
shyammukund: You can get rid of the raw query in this test by using check_rels() in this test.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -8,423 +10,398 @@
...
+        "id",
+        "LANGUAGE",
+    )
+    assert expected_repo_language_rels.issubset(actual_repo_language_rels)
 
-def test_repository_to_collaborators(neo4j_session):
-    _ensure_local_neo4j_has_test_data(neo4j_session)
 
-    # Ensure outside collaborators are connected to the expected repos
+def _test_collaborator_relationships(neo4j_session):
```
shyammukund: same thing about using check_rels() here

shyammukund (changes requested):
Thanks for your contribution!  Just left some comments on using check_nodes() and check rels() consistently in the integration tests so that they follow established cartography patterns

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -8,423 +10,398 @@
...
+    )  # GitHubRepository -[:REQUIRES]-> PythonLibrary
+    _test_github_dependencies(
+        neo4j_session
+    )  # GitHubRepository -[:REQUIRES]-> Dependency
+    _test_dependency_manifests(
+        neo4j_session
+    )  # GitHubRepository -[:HAS_MANIFEST]-> DependencyGraphManifest -[:HAS_DEP]-> Dependency
+    _test_dependency_relationship_properties(
+        neo4j_session
+    )  # REQUIRES relationship properties
```
achantavy: Please follow https://github.com/cartography-cncf/cartography/blob/fd1fb47b06a1bc2f23382998814ff7de49cf64d1/tests/integration/cartography/intel/lastpass/test_users.py#L17 as an example. No need to introduce `_test()` functions because when a test fails, we need to go deeper by one level. It is ok for the test function to be long.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -8,423 +10,398 @@
...
+        ), "ID should not be None"
+        assert all(
+            prop is not None for prop in essential_props[1:7]
+        ), "Core text properties should not be None"
+        assert all(
+            isinstance(prop, bool) for prop in essential_props[8:12]
+        ), "Boolean flags should be boolean"
+        assert all(
+            prop is not None for prop in essential_props[12:17]
+        ), "URL and date properties should not be None"
```
achantavy: Not necessary. Integ tests in cartography mostly only need to test for nodes being attached to each other. All of this just makes the test unnecessarily long and hard to read.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -8,423 +10,398 @@
...
-def test_transform_and_load_repository_languages(neo4j_session):
+    assert private is True
+    assert disabled is False
+    assert archived is False
+    assert locked is True
+    # assert giturl == "git://github.com:simpsoncorp:sample_repo.git"  buggy url conversion
+    assert url == "https://github.com/simpsoncorp/sample_repo"
+    assert sshurl == "git@github.com:simpsoncorp/sample_repo.git"
+    assert createdat == "2011-02-15T18:40:15Z"
+    assert updatedat == "2020-01-02T20:10:09Z"
```
achantavy: Like Shyam said, you can use check_nodes() here if you really want to verify properties on a node, but checking each possible field isn't necessary.

tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -8,423 +10,398 @@
...
+
+
+def _test_owner_nodes_and_relationships(neo4j_session):
```
achantavy: As I mentioned earlier, let's not use these extra `_test()` functions because to trace a problem and read the test, we need to find what function called this function, and that's harder to do than just simply scrolling up and down in the same function. As currently written, I don't know what things happened to lead up to the set of assertions that are here.

achantavy (changes requested):
Thanks for taking a stab at this. To make the diff cleaner, `def test_sync_repos(mock_collabs, mock_get, neo4j_session):` should be at the bottom of the file.

Thank you for testing the `sync`, left comments below on not breaking things down to secondary `_test*()` functions to keep things simple. Write back if you have questions

achantavy:
Also, don't change existing tests. Just add in this as a new one to keep the diff clean. Since this is part of a refactor, making sure that previous tests still pass and easily seeing how the code changes is especially important.

achantavy:
Strategically for the refactor, here's the steps to make this clean:

1. add a test for the sync(). add it at the bottom of the file (or even in a separate file) to keep the diff clean. add in test data if necessary. at this point we now have old tests + a new test, and all should pass.
2. refactor the code to use the data model. while doing this refactor, it is possible that we may need to update or even delete the old tests. this is fine as long as the new test continues to pass.
3. we are done.

tests/integration/cartography/intel/github/test_repos.py
SachdevJai: its not necessary but it would help us catch minor bugs in transformation stages, one i found was the incorrect git url

SachdevJai:
I've updated the tests with all the suggestions in mind, Thank you for the support <3


I was wondering if it would be better to merge this into a github_refactor branch first? then proceed with the rest of the refactor? instead of master directly

shyammukund (changes requested):
Hey heres some considerations I think you should take into account before your next attempt at refactoring these tests. 

- Firstly, the data model repos.py has been modified slightly (replaced the version property with requirements). its a pretty minor change, but will affect a lot of the tests so i would firstly go pull the latest version of cartography into your branch before doing any more work on the tests.
- Nodes and relationships NEED to be tested using the check_nodes() and check_rels() functions. I know there are a lot of tests in the codebase that don't do this but we are in the process of refactoring those tests to use the modern way of writing integration tests. If you are having an LLM write the tests please make sure that you read through the code it is writing and make sure that it is following this format when asserting nodes and rels. I know you updated a lot of those in your latest commit, but there are still a few instances where it isn't used.
- Finally, the purpose of an integration test is to test the end to end sync of the module you are testing. By this i mean that the integration test should just check whether the intended nodes were created (using check_nodes()) and whether the intended relationship between those nodes was created (check_rels()). We don't really care about testing weird edge cases in the data or testing transform or load functions from the intelligence module (usually done in unit tests). I would recommend looking at the schema.md file for the github module and noting what are the nodes being created here and what are the relationships being created and asserting just those. Take a look at tests/integration/cartography/intel/aws/test_guardduty.py for a good example of what a simple integration test should look like. 

You've got the right idea in test_repos.py by having one test_sync_repos, but the contents of that function aren't necessarily testing what it should be. Please give this another shot and let me know if you have any other questions or if anything i said above didn't make sense!

SachdevJai:
> Hey heres some considerations I think you should take into account before your next attempt at refactoring these tests.
> 
> * Firstly, the data model repos.py has been modified slightly (replaced the version property with requirements). its a pretty minor change, but will affect a lot of the tests so i would firstly go pull the latest version of cartography into your branch before doing any more work on the tests.

I see, I'll do that 

> * Nodes and relationships NEED to be tested using the check_nodes() and check_rels() functions. I know there are a lot of tests in the codebase that don't do this but we are in the process of refactoring those tests to use the modern way of writing integration tests. If you are having an LLM write the tests please make sure that you read through the code it is writing and make sure that it is following this format when asserting nodes and rels. I know you updated a lot of those in your latest commit, but there are still a few instances where it isn't used.

The test where the cypher query is still there, has a where clause where manifest_path is not NULL, i couldnt figure out a way to filter for that since its a relationship property, please let me know how i can approach that. Also in the python pinned and unpinned dependencies, im just fetching all rels once using check_rels and asserting over those since its the same operation over and over again

> * Finally, the purpose of an integration test is to test the end to end sync of the module you are testing. By this i mean that the integration test should just check whether the intended nodes were created (using check_nodes()) and whether the intended relationship between those nodes was created (check_rels()). We don't really care about testing weird edge cases in the data or testing transform or load functions from the intelligence module (usually done in unit tests). I would recommend looking at the schema.md file for the github module and noting what are the nodes being created here and what are the relationships being created and asserting just those. Take a look at tests/integration/cartography/intel/aws/test_guardduty.py for a good example of what a simple integration test should look like.

> You've got the right idea in test_repos.py by having one test_sync_repos, but the contents of that function aren't necessarily testing what it should be. Please give this another shot and let me know if you have any other questions or if anything i said above didn't make sense!

Initially i was trying to just rewrite the existing tests in a more readable format keeping the same assertions, should i write entirely new tests for asserting the node creation and relationship creation regardless of what previous tests looked like? 

thank you for your suggestions

shyammukund:
@SachdevJai After you pull the latest version of cartography I would leave the existing tests for now and just write your new test_sync_repos function at the bottom of the file so we can easily tell what is old and what is new. Don't worry about making sure that everything that is tested before is in your test_sync function, but rather go through the data model and try to understand what relationships and nodes are being created and test those. 

To your point about trying to find a way to test the relationship properties i believe that check_rels() does this for you (@achantavy please correct me if I am wrong about this).

SachdevJai:
that makes sense @shyammukund 

from what i saw in the check_rels function it returns the attributes of the nodes we give to it if they are connected by the provided relationship

there is no way to filter using relationship properties but that might not be necessary since we're only verifying the existance of these relationships and nodes 

ill give this another shot

shyammukund:
@SachdevJai Create a new PR for this on your next commit and tag me. The diff in this got a little hard to read, so lets just start fresh. Appreciate your efforts on this!

Issue #1808, opened 2025-08-13
Incorrect URL conversion from ssh to git://


SachdevJai:
*Description*:
When converting SSH-style Git URLs to git:// URLs, ```_create_git_url_from_ssh_url()``` performs naive string replacements that produce malformed URLs.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

*Logs*:
> If applicable, copy and paste your console log with the failing stack trace.

*Screenshots*:
> If applicable, add screenshots to help explain your problem.

*Please complete the following information:*:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]
54741579ca518ac941b8a3d180318cccfc210136
>  - Python version: [e.g. 3.10.0]
3.11

*Additional context*:
> Add any other context about the problem here.

relevant files

https://git-scm.com/docs/git-pull#_git_urls

cartography/intel/github/repos.py

jychp:
From Slack:

```
Hi
I think the url conversion logic for ssh to git is incorrect in the gh repo module

def _create_git_url_from_ssh_url(ssh_url: str) -> str:
    """
    Return a git:// URL from the given ssh_url
    """
    return ssh_url.replace("/", ":").replace("git@", "git://")
this would return
git://github.com:cartography-cncf:cartography.git

git ls-remote git://github.com:cartography-cncf:cartography.git
fatal: no path specified; see 'git help pull' for valid url syntax
git://github.com/cartography-cncf/cartography.git
```

PR #1810, opened 2025-08-13
fix(github): Correct SSH to git URL conversion in _create_git_url_from_ssh_url #1808

Roniscend:
This PR fixes the malformed Git URL conversion in  create_git_url_from_ssh_url by replacing naive string replacement with proper SSH URL parsing using string operations. The function now correctly splits the SSH URL at `@` to separate the user from host/path, then splits the host and path at the first `:` and reconstructs them as a proper git:// URL format, converting `git@github.com:cartography-cncf/cartography.git` to `git://github.com/cartography-cncf/cartography.git` and fixing the "fatal: no path specified" errors that occurred with malformed URLs.


jychp:
@Roniscend looks promising, can you please:
- sign+rebase your commits
- launch test locally and fix errors (cf https://cartography-cncf.github.io/cartography/dev/developer-guide.html#run-the-full-test-suite)

Thank you for your contribution.

cartography/intel/github/repos.py
```diff
@@ -403,11 +404,21 @@ def _create_default_branch_id(repo_url: str, default_branch_ref_id: str) -> str:
...
 def _create_git_url_from_ssh_url(ssh_url: str) -> str:
     """
-    Return a git:// URL from the given ssh_url
+    Return a git:// URL from the given ssh_url.
+    Example:
+        git@github.com:cartography-cncf/cartography.git
+        -> git://github.com/cartography-cncf/cartography.git
     """
-    return ssh_url.replace("/", ":").replace("git@", "git://")
+    match = re.match(r'^git@([^:]+):(.+)$', ssh_url)
```
achantavy: Is there a way to do this without regex?

Roniscend: Yes there is but regex is better because it makes the expected URL pattern explicit and validates the format in one line, reducing logic branching. It also scales easily to handle edge cases like optional ports or variations without rewriting the parsing code.

achantavy: I feel like regex is overengineering it. Ports aren't expected in this string. The most complex I think it will be is deeper nested paths (like in a monorepo). 

I'd prefer to do something simple like

```python
def _create_git_url_from_ssh_url(ssh_url: str) -> str:
    """
    Convert SSH URL to git:// URL.
    Example:
        git@github.com:cartography-cncf/cartography.git
        -> git://github.com/cartography-cncf/cartography.git
    """
    # Remove the user part (e.g., "git@")
    _, host_and_path = ssh_url.split("@", 1)
    # Replace first ':' (separating host and repo) with '/'
    host, path = host_and_path.split(":", 1)
    return f"git://{host}/{path}"
```

Roniscend: I'll update the implementation to use string operations instead of regex as you suggested.

achantavy:
@Roniscend - can you please add a very short unit test for this function and clean up the PR description?

After that I'll be happy to merge!

Roniscend:
@achantavy Updated the PR description and added the unit test for `_create_git_url_from_ssh_url` as requested.

achantavy (approved):
Thank you!!

Roniscend:
@achantavy I noticed the pre-commit hooks identified some minor formatting issues (trailing whitespace and import ordering), so I ran the formatters locally and applied the fixes.

PR #1815, opened 2025-08-17
feat(aws): Attach LambdaPermission Properties to LambdaFunction in AWS Module

d-aggarwal:
### Related issues or links

### Checklist

image


cartography/intel/aws/lambda_function.py
```diff
@@ -48,6 +54,17 @@ def transform_lambda_functions(lambda_functions: List[Dict], region: str) -> Lis
...
+        function_arn = function_data["FunctionArn"]
+        if function_arn in permissions_by_arn:
+            permission_data = permissions_by_arn[function_arn]
+            transformed_function["AnonymousAccess"] = permission_data["AnonymousAccess"]
+            transformed_function["AnonymousActions"] = permission_data[
+                "AnonymousActions"
+            ]
+        else:
+            transformed_function["AnonymousAccess"] = None
+            transformed_function["AnonymousActions"] = None
```
achantavy: Run it through the integ tests, but I'm pretty sure we don't need the `else` clause here and can just leave that key undefined.

The idea is to avoid `if+else` wherever possible to reduce complexity

cartography/intel/aws/lambda_function.py
```diff
@@ -126,6 +143,53 @@ def get_event_source_mappings(
...
+            parsed_policy = parse_policy(function_arn, policy)
+            all_permissions[function_arn] = {
+                "AnonymousAccess": parsed_policy.get("AnonymousAccess"),
+                "AnonymousActions": parsed_policy.get("AnonymousActions"),
+            }
+        else:
+            all_permissions[function_arn] = {
+                "AnonymousAccess": None,
+                "AnonymousActions": None,
+            }
```
achantavy: We can avoid the if+else here by setting a default value and then overwriting it whereever possible

```suggestion
        all_permissions[function_arn] = {
            "AnonymousAccess": None,
            "AnonymousActions": None,
        }
        if policy:
            parsed_policy = parse_policy(function_arn, policy)
            all_permissions[function_arn] = {
                "AnonymousAccess": parsed_policy.get("AnonymousAccess"),
                "AnonymousActions": parsed_policy.get("AnonymousActions"),
            }
```

cartography/intel/aws/lambda_function.py
```diff
@@ -126,6 +143,53 @@ def get_event_source_mappings(
...
+
+    return all_permissions
+
+
+@timeit
```
achantavy:
```suggestion
```

achantavy:
Hi @d-aggarwal, are you able to follow-up here?

d-aggarwal:
> Hi @d-aggarwal, are you able to follow-up here?

I've updated the changes, also for functions with no policies attached, it was throwing warning and botocore exception, added try catch block for that.

Issue #1082, opened 2023-01-12
Add support for SES


JonZeolla:
**Feature request template**

*Title*: Add support for SES

*Description*:
Add support for SES as a node.

mohammedfirdouss:
@achantavy Can I work on this?. I am quite new to the community

jychp:
@mohammedfirdouss sure, feel free to work on it!

mohammedfirdouss:
Alright no problem @jychp

PR #1817, opened 2025-08-17
feat(aws): add support for AWS SES (Simple Email Service)

mohammedfirdouss:
## Add support for AWS SES (Simple Email Service)

Resolves: #1082

##  Overview
This pull request adds full support for **AWS Simple Email Service (SES)** in Cartography.  
The goal is to help security teams see and analyze their SES email setup as part of the security graph.

### Core Features
- **SES Identities**  
  - Collects email addresses and domains configured in SES  
  - Includes verification status and tokens  
  - DKIM configuration and status  
  - SNS topics for bounces, complaints, and deliveries  
  - Forwarding and header configuration settings  

- **SES Configuration Sets**  
  - Retrieves and maps configuration sets that define email sending policies  

### Graph Schema
- **New Node Types**
  - `SESIdentity` – represents a verified domain or email with metadata  
  - `SESConfigurationSet` – represents SES configuration sets  

- **Properties Captured**
  - Verification status and tokens  
  - DKIM settings and status  
  - SNS notification topics  
  - Forwarding and header settings  

- **Relationships**
  - `SESIdentity -[:RESOURCE]-> AWSAccount`  
  - `SESConfigurationSet -[:RESOURCE]-> AWSAccount`  

## Testing
- Verified data transformation functions  
- Checked identity type detection (Domain vs Email)  
- Tested DKIM setup variations  
- Validated SNS topic setup  
- Covered edge cases and error handling  

## Usage
- Works with Cypher queries and CLI like other AWS services  

## Implementation Details
- Follows patterns used by other AWS services (SNS, S3, etc.)  
- Collects data across all AWS regions  
- Handles API errors and missing permissions gracefully  
- Normalizes AWS responses into graph-friendly structures  
- Generates consistent ARN-like identifiers

### DIFF
--- a/cartography/intel/aws/apigateway.py
+++ b/cartography/intel/aws/apigateway.py
@@ -101,6 +101,7 @@ def get_rest_api_details(
 
 
 @timeit
+@aws_handle_regions
 def get_rest_api_stages(api: Dict, client: botocore.client.BaseClient) -> Any:
     """
     Gets the REST API Stage Resources.
@@ -142,6 +143,7 @@ def get_rest_api_client_certificate(
 
 
 @timeit
+@aws_handle_regions
 def get_rest_api_resources(api: Dict, client: botocore.client.BaseClient) -> List[Any]:
     """
     Gets the collection of Resource resources.
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -25,6 +25,7 @@ import backoff
 import boto3
 import botocore
 import neo4j
+from botocore.exceptions import EndpointConnectionError
 
 from cartography.graph.job import GraphJob
 from cartography.graph.statement import get_job_shortname
@@ -309,6 +310,13 @@ def aws_handle_regions(func: AWSGetFunc) -> AWSGetFunc:
                 return []
             else:
                 raise
+        except EndpointConnectionError:
+            logger.warning(
+                "Encountered an EndpointConnectionError. This means that the AWS "
+                "resource is not available in this region. Skipping.",
+                exc_info=True,
+            )
+            return []
 
     return cast(AWSGetFunc, inner_function)
 
--- a/tests/unit/cartography/intel/aws/test_apigateway.py
+++ b/tests/unit/cartography/intel/aws/test_apigateway.py
@@ -1,4 +1,10 @@
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from botocore.exceptions import ClientError
+
 import tests.data.aws.apigateway as test_data
+from cartography.intel.aws.apigateway import get_rest_api_resources
 from cartography.intel.aws.apigateway import parse_policy
 
 
@@ -15,3 +21,46 @@ def test_none_policy():
     res = parse_policy(None, None)
 
     assert (res) is None
+
+
+@patch("time.sleep")
+@patch("cartography.intel.aws.apigateway.logger")
+@patch("botocore.client.BaseClient.get_paginator")
+def test_get_rest_api_resources_retries_on_too_many_requests(
+    mock_get_paginator,
+    mock_logger,
+    mock_sleep,
+):
+    """
+    Test that get_rest_api_resources retries on TooManyRequestsException
+    and succeeds on the third attempt.
+    """
+    # Arrange
+    api = {"id": "test-api"}
+    client = MagicMock()
+    expected_resources = [
+        {"id": "resource1", "pathPart": "users"},
+        {"id": "resource2", "pathPart": "orders"},
+    ]
+    too_many_requests_error = ClientError(
+        error_response={
+            "Error": {"Code": "TooManyRequestsException", "Message": "Rate exceeded"}
+        },
+        operation_name="GetResources",
+    )
+    mock_paginator = MagicMock()
+    mock_paginator.paginate.side_effect = [
+        too_many_requests_error,  # First attempt fails
+        too_many_requests_error,  # Second attempt fails
+        [{"items": expected_resources}],  # Third attempt succeeds
+    ]
+    client.get_paginator.return_value = mock_paginator
+
+    # Act
+    result = get_rest_api_resources(api, client)
+
+    # Assert
+    assert result == expected_resources
+    assert mock_paginator.paginate.call_count == 3
+    mock_paginator.paginate.assert_called_with(restApiId="test-api")
+    client.get_paginator.assert_called_with("get_resources")
--- a/tests/unit/cartography/intel/aws/test_codebuild.py
+++ b/tests/unit/cartography/intel/aws/test_codebuild.py
@@ -1,0 +1,23 @@
+from unittest.mock import MagicMock
+
+from botocore.exceptions import EndpointConnectionError
+
+from cartography.intel.aws import codebuild
+
+
+def test_get_all_codebuild_projects_endpoint_connection_error():
+    """Ensure EndpointConnectionError is handled gracefully."""
+    # Arrange
+    boto3_session = MagicMock()
+    paginator = boto3_session.client.return_value.get_paginator.return_value
+    paginator.paginate.side_effect = EndpointConnectionError(
+        endpoint_url="https://codebuild.mx-central-1.amazonaws.com"
+    )
+
+    # Act
+    result = codebuild.get_all_codebuild_projects(boto3_session, "mx-central-1")
+
+    # Assert that we return nothing. In AWS region processing, endpoint connection
+    # failures are often about service availability in that region, not client/server
+    # network issues
+    assert result == []


mohammedfirdouss:
@jychp @achantavy I worked on this issue to add support for SES, can this be reviewed?
I have a sign off commit issue, I will work on that as well

cartography/models/aws/ses/identity.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@dataclass(frozen=True)
+class SESIdentityNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("IdentityArn")
+    arn: PropertyRef = PropertyRef("IdentityArn", extra_index=True)
+    identity: PropertyRef = PropertyRef("Identity", extra_index=True)
+    identity_type: PropertyRef = PropertyRef("IdentityType")
+    verification_status: PropertyRef = PropertyRef("VerificationStatus")
+    verification_token: PropertyRef = PropertyRef("VerificationToken")
```
jychp: I don't think we want to store sensitive tokens into Cartography.

cartography/models/aws/ses/identity.py
```diff
@@ -0,0 +1,55 @@
...
+    dkim_enabled: PropertyRef = PropertyRef("DkimEnabled")
+    dkim_verification_status: PropertyRef = PropertyRef("DkimVerificationStatus")
+    dkim_tokens: PropertyRef = PropertyRef("DkimTokens")
```
jychp: Do not store sensitive tokens into Cartography.

cartography/intel/aws/ses.py
```diff
@@ -0,0 +1,278 @@
...
+    try:
+        response = client.list_identities()
+        return response.get("Identities", [])
+    except Exception as e:
+        logger.warning(f"Error getting SES identities for region {region}: {e}")
+        return []
+
+
+@timeit
+def get_identity_attributes(
```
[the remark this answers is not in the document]
jychp: Please fix this one.

cartography/intel/aws/ses.py
```diff
@@ -0,0 +1,278 @@
...
+@aws_handle_regions
+def get_ses_identities(boto3_session: boto3.session.Session, region: str) -> List[Dict]:
+    """
+    Get all SES identities (email addresses and domains) for a region.
+    """
+    client = boto3_session.client("ses", region_name=region)
+    try:
+        response = client.list_identities()
+        return response.get("Identities", [])
+    except Exception as e:
```
jychp: For error handling, we prefer to fail loudly. It’s better not to catch the exception here and let it propagate.

cartography/intel/aws/ses.py
```diff
@@ -0,0 +1,278 @@
...
+        dkim_attrs = client.get_identity_dkim_attributes(Identities=identities)
+        
+        # Combine all attributes
+        for identity in identities:
+            identity_attributes[identity] = {
+                "verification": verification_attrs.get("VerificationAttributes", {}).get(identity, {}),
+                "notification": notification_attrs.get("NotificationAttributes", {}).get(identity, {}),
+                "dkim": dkim_attrs.get("DkimAttributes", {}).get(identity, {}),
+            }
+    except Exception as e:
```
jychp: For error handling, we prefer to fail loudly. It’s better not to catch the exception here and let it propagate.

cartography/intel/aws/ses.py
```diff
@@ -0,0 +1,278 @@
...
+        identity_type = "Domain" if "." in identity and "@" not in identity else "EmailAddress"
+        
+        # Create ARN-like identifier for the identity
+        identity_arn = f"arn:aws:ses:{region}:{account_id}:identity/{identity}"
+        
+        transformed_identity = {
+            "IdentityArn": identity_arn,
+            "Identity": identity,
+            "IdentityType": identity_type,
+            "VerificationStatus": verification_attrs.get("VerificationStatus", ""),
```
jychp: Keep `None` as the default value, not an empty string.
Empty strings are still stored as fields in Neo4j, while `None` values are not and we definitely don’t want empty strings being persisted.

cartography/intel/aws/ses.py
```diff
@@ -0,0 +1,278 @@
...
+    transformed_identities = []
+    
+    for identity in identities:
+        attrs = identity_attributes.get(identity, {})
+        verification_attrs = attrs.get("verification", {})
+        notification_attrs = attrs.get("notification", {})
+        dkim_attrs = attrs.get("dkim", {})
+        
+        # Determine identity type (email or domain)
+        identity_type = "Domain" if "." in identity and "@" not in identity else "EmailAddress"
```
jychp: The dot check isn’t relevant here both `email` and `domain` will always contain a dot. You can safely remove this check.

cartography/intel/aws/ses.py
```diff
@@ -0,0 +1,278 @@
...
+) -> List[Dict]:
+    """
+    Transform SES configuration set data for ingestion
+    """
+    transformed_sets = []
+    
+    for config_set in configuration_sets:
+        # Create ARN-like identifier for the configuration set
+        config_set_name = config_set.get("Name", "")
+        config_set_arn = f"arn:aws:ses:{region}:{account_id}:configuration-set/{config_set_name}"
```
jychp: I think using an "ARN-like" identifier can be misleading. We should choose an ID format that cannot be confused with a valid ARN.

cartography/models/aws/ses/configuration_set.py
```diff
@@ -0,0 +1,42 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class SESConfigurationSetNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("ConfigurationSetArn")
+    arn: PropertyRef = PropertyRef("ConfigurationSetArn", extra_index=True)
```
jychp: Do not store this field as "arn" if it's not a true ARN, this could be very misleading

cartography/intel/aws/ses.py
jychp, line 1: In Cartography, we follow a simpler pattern: patch the module’s `get()` and test the `sync()` (which implicitly exercises `transform()`).

This keeps tests straightforward and maintainable while preserving acceptable coverage.

For examples, see recent modules (e.g., Tailscale).

jychp (changes requested):
Thanks for your first contribution to the project. I’ve left several comments.

My main question: why sync the configuration set with only its name as the sole piece of data?
Could we link this node to `SESIdentity`?
Can we add additional fields to make this node meaningful on its own?

cartography/intel/aws/ses.py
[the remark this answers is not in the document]
mohammedfirdouss: Okay

mohammedfirdouss:
Okay @jychp I cleaned things up a bit, Apologies if this took a while, I took out the SES VerificationToken and DKIM fields so no sensitive tokens are stored.  
Switched optional fields to just use None instead of empty strings or False.  
Removed the try/except so it fails right away if something’s wrong.  
Also, I made identity detection simpler, just checking for '@'.  
Also swapped out the ARN-like IDs for internal carto: IDs and updated the config set schema to fit.

mohammedfirdouss:
> Thanks for your first contribution to the project. I’ve left several comments.
> 
> My main question: why sync the configuration set with only its name as the sole piece of data? Could we link this node to `SESIdentity`? Can we add additional fields to make this node meaningful on its own?

Hmm, okay so, The SES v1 ListConfigurationSets API only returns configuration set names. I synced just that minimal identifier at first to establish the node and account scoping. That way we avoided extra calls or unnecessary scope creep.

For the other question, yes we can. The SES v2 GetEmailIdentity includes the ConfigurationSetName for the default config set. If it’s present, we can model it like this:
```ruby
(:SESIdentity)-[:DEFAULT_CONFIGURATION_SET]->(:SESConfigurationSet)
```

I think we can add more details about the configuration set using SES v2

jychp:
> > Thanks for your first contribution to the project. I’ve left several comments.
> > My main question: why sync the configuration set with only its name as the sole piece of data? Could we link this node to `SESIdentity`? Can we add additional fields to make this node meaningful on its own?
> 
> Hmm, okay so, The SES v1 ListConfigurationSets API only returns configuration set names. I synced just that minimal identifier at first to establish the node and account scoping. That way we avoided extra calls or unnecessary scope creep.
> 
> For the other question, yes we can. The SES v2 GetEmailIdentity includes the ConfigurationSetName for the default config set. If it’s present, we can model it like this:
> 
> ```ruby
> (:SESIdentity)-[:DEFAULT_CONFIGURATION_SET]->(:SESConfigurationSet)
> ```
> 
> I think we can add more details about the configuration set using SES v2

For sure, we need to add (and test) the `DEFAULT_CONFIGURATION_SET` relationship.

I also think it would be worth looking at what SES v2 provides, because a node that only stores a name doesn’t add much analytical value.

mohammedfirdouss:
> > > Thanks for your first contribution to the project. I’ve left several comments.
> > > My main question: why sync the configuration set with only its name as the sole piece of data? Could we link this node to `SESIdentity`? Can we add additional fields to make this node meaningful on its own?
> > 
> > 
> > Hmm, okay so, The SES v1 ListConfigurationSets API only returns configuration set names. I synced just that minimal identifier at first to establish the node and account scoping. That way we avoided extra calls or unnecessary scope creep.
> > For the other question, yes we can. The SES v2 GetEmailIdentity includes the ConfigurationSetName for the default config set. If it’s present, we can model it like this:
> > ```ruby
> > (:SESIdentity)-[:DEFAULT_CONFIGURATION_SET]->(:SESConfigurationSet)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think we can add more details about the configuration set using SES v2
> 
> For sure, we need to add (and test) the `DEFAULT_CONFIGURATION_SET` relationship.
> 
> I also think it would be worth looking at what SES v2 provides, because a node that only stores a name doesn’t add much analytical value.

Makes sense, I’ll  see about adding the `DEFAULT_CONFIGURATION_SET` relationship and check SES v2 for extra fields so the node carries more value than just the name @jychp

PR #1819, opened 2025-08-17
feat(AWS): Add APIGateway Method and Integration as resource

d-aggarwal:
### Related issues or links

### Checklist

image

image


achantavy:
@d-aggarwal - getting a mypy error:

```
tests/unit/cartography/intel/aws/test_apigateway.py:7: error: Module "cartography.intel.aws.apigateway" has no attribute "get_rest_api_resources"  [attr-defined]
```

can you resolve? After that let's get this merged in :)

achantavy (dismissed):
mypy

d-aggarwal:
> @d-aggarwal - getting a mypy error:
> 
> ```
> tests/unit/cartography/intel/aws/test_apigateway.py:7: error: Module "cartography.intel.aws.apigateway" has no attribute "get_rest_api_resources"  [attr-defined]
> ```
> 
> can you resolve? After that let's get this merged in :)

i think the issue is coming  from [ this commit](https://github.com/cartography-cncf/cartography/commit/998f7e0309e5879f4cd63010289cf47c1bf75196),  I have changed the function name and update it to return method and integrations as well, because all 3 are calling the same api and  pass resource as a parameter.

and the commit is calling the previous version of function, i'm not sure how can i resolve this, lmk your thoughts about it.

jychp:
@d-aggarwal after you renamed the func, you updated the `integration` tests but not the `unit` tests: 

> tests/**unit**/cartography/intel/aws/test_apigateway.py:7: error: Module "cartography.intel.aws.apigateway" has no attribute "get_rest_api_resources"  [attr-defined]

d-aggarwal:
I've updated the changes :)

PR #1825, opened 2025-08-19
feature(kubernetes): K8s RBAC OIDC Integration for EKS Platform

shyammukund:
### Summary
- Ingestion of Kubernetes Users and Groups
- EKS integration with Kubernetes Module
- External OIDC Integration (Just Okta right now)
- AWS IAM Integration 

- enhancing K8s RBAC Functionality added here https://github.com/cartography-cncf/cartography/pull/1776

### Checklist

- [x] Graph 
image


cartography/intel/kubernetes/__init__.py
```diff
@@ -42,6 +45,25 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
...
                 session, client, config.update_tag, common_job_parameters
             )
+            # EKS identity provider sync (requires AWS regions)
+            if config.aws_regions:
+                boto3_session = boto3.Session()
+                regions = parse_and_validate_aws_regions(config.aws_regions)
+
+                # For EKS OIDC, we need to determine which region the cluster is in
+                # For now, sync with the first specified region
+                # NOTE: Should we be able to process clusters in multiple regions?
```
kunaals: Each EKS API server (the brain of the cluster) will only be in one region. Each client in get_k8s_clients should have the EKS cluster ARN which will contain the region of the cluster. In the kubeconfig yaml, we should look for "name" under "cluster" to get this ARN which contains the region

cartography/intel/kubernetes/__init__.py
```diff
@@ -42,6 +45,25 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
             sync_kubernetes_rbac(
                 session, client, config.update_tag, common_job_parameters
             )
+            # EKS identity provider sync (requires AWS regions)
+            if config.aws_regions:
```
kunaals: If someone runs the AWS and Okta syncs and _then_ runs the kubernetes sync with a separate invocation of the cartography CLI, they will not get these EKS goodies because aws_regions will be null. I think we should have an optional `k8s_managed_provider` config option that today only supports the options None or "eks".  This will give users more control and make it clear that the EKS bits are running

In the future we'll add gke and aks to this list

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+def parse_role_mappings(configmap: V1ConfigMap) -> List[Dict[str, Any]]:
+    """
+    Parse mapRoles from aws-auth ConfigMap.
+
+    :param configmap: V1ConfigMap containing aws-auth data
+    :return: List of role mapping dictionaries
+    """
+    map_roles_yaml = configmap.data["mapRoles"]
```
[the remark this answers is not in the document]
kunaals: We also need to support mapUsers and mapAccounts within the aws-auth configmap. This document sheds some light: https://github.com/kubernetes-sigs/aws-iam-authenticator?tab=readme-ov-file#full-configuration-format

mapUsers is just like mapRoles except it maps IAM users to Kubernetes user (AWS IAM roles vs AWS IAM users description here: https://docs.aws.amazon.com/IAM/latest/UserGuide/id.html

mapAccounts is a bit more complex. When an AWS account ID is specified in this list, **every single AWS IAM user and IAM role in that account can be a Kube user.** The Kube username is the ARN of that identity. It's helpful when you want any identity in your AWS account to be able to access kube _if_ it's in a role binding. Puts the full onus of authorization into kube RBAC.

Also, I think the aws-auth configmap can have any combination of mapRoles, mapUsers, and mapAccounts, so we should use get() functions in case of nulls

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,321 @@
...
+    role_mappings = yaml.safe_load(map_roles_yaml) or []
+
+    # Filter out templated entries because these are not real users
+    filtered_mappings = []
+    for mapping in role_mappings:
+        username = mapping.get("username", "")
```
kunaals: If username is empty, the AWS authenticator defaults to using the ARN as the username. Let's do the same

ex:
```
  - rolearn: arn:aws:iam::000000000000:role/KubernetesNode
    groups:
    - system:bootstrappers
    - aws:instances
```
username here should be arn:aws:iam::000000000000:role/KubernetesNode

kunaals: Same with mapUsers

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,321 @@
...
+
+    logger.info(
+        f"Transformed {len(users)} users and {len(groups)} groups with AWS role mappings"
+    )
+
+    return {"users": users, "groups": groups}
+
+
+@timeit
+def get_oidc_providers(
```
kunaals: Let's make this singular - AWS EKS only supports one external OIDC provider. Will require changes to transform and load functions as well

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,321 @@
...
+    client = boto3_session.client("eks", region_name=region)
+    oidc_providers = []
+
+    # Extract just the cluster name from ARN if needed
+    # ARN format: arn:aws:eks:region:account:cluster/cluster-name
+    if cluster_name.startswith("arn:aws:eks:"):
+        cluster_name = cluster_name.split("/")[-1]
+
+    # Get configured external identity provider configs
+    configs_response = client.list_identity_provider_configs(clusterName=cluster_name)
```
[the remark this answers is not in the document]
kunaals: Doesn't matter because there is only one - no need for pagination

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,321 @@
...
+    map_roles_yaml = configmap.data["mapRoles"]
+    role_mappings = yaml.safe_load(map_roles_yaml) or []
+
+    # Filter out templated entries because these are not real users
+    filtered_mappings = []
+    for mapping in role_mappings:
+        username = mapping.get("username", "")
+        if "{{" in username:
+            logger.debug(f"Skipping templated username: {username}")
+            continue
```
kunaals: Skipping templated usernames and groups could result in a lot of missed AWSRole to KubernetesUser and KubernetesGroup relationships. Its best we try to deal with this case

According to [this aws-auth template](https://github.com/kubernetes-sigs/aws-iam-authenticator?tab=readme-ov-file#full-configuration-format) there are three cases of templating for map roles:
1. {{AccountID}}
2. {{SessionName}}
3. {{SessionNameRaw}}
There are some others like {{AccessKeyID}} and {{EC2PrivateDNSName}} but lets leave that as a github issue for later.

Here's what we should do for each of them:
1. {{AccountID}}
Just replace with AWS account ID that is in the `rolearn`

2. {{SessionName}} and {{SessionNameRaw}}
Session names can be arbitrary (except with AWS SSO) so if an AWS role maps to a username or group with SessionName or SessionNameRaw, we should map that AWS role to **all Kube groups or kube users in rolebindings that _could_ match the format of the username**. Some regex likely required here.
Ex: if username is `subimage-scan{{SessionName}}` and there are Kubernetes users `subimage-scanfoo` and `subimage-scanbar` that AWS role should map to _both_  of these Kubernetes users.

SessionName vs SessionNameRaw: SessionName transliterates '@' to '-'. So if the entry in mapRoles looks like:
```
- "groups":
  - "system:authenticated"
  "rolearn": "arn:aws:iam::123456789111:role/SubImageScanRole"
  "username": "subimage-scan{{SessionNameRaw}}{{SessionNameRaw}}{{SessionName}}"
```
and my session name is "fo@o" then `kubectl auth whoami` will return the following
| ATTRIBUTE                                            | VALUE                                                                                   |
|------------------------------------------------------|-----------------------------------------------------------------------------------------|
| Username                                             | `subimage-scanfo@ofo@ofo-o`                                                               |
| UID                                                  | `aws-iam-authenticator:************:AROAS***********BU23`                                 |
| Groups                                               | [system:authenticated]                                                                  |
| Extra: accessKeyId                                   | [ASIA************N2]                                                                     |
| Extra: arn                                           | [arn:aws:sts::************:assumed-role/SubImageScanRole/fo@o]                           |
| Extra: canonicalArn                                  | [arn:aws:iam::************:role/SubImageScanRole]                                        |
| Extra: principalId                                   | [AROAS***********BU23]                                                                   |
| Extra: sessionName                                   | [fo@o]                                                                                  |
| Extra: sigs.k8s.io/aws-iam-authenticator/principalId | [AROAS***********BU23]                                                                   |


Example code snippet for the regex:
```
import re
from typing import Tuple, Pattern

SESSION_TOKEN = "{{SessionName}}"
SESSION_RAW_TOKEN = "{{SessionNameRaw}}"

# Rendered username pieces (post-template):
# - For {{SessionName}}: '@' would have been transliterated to '-', so no '@' shows up.
SESSION_RENDERED = r"[\w+=,.-]{2,64}"       # note: no '@'
# - For {{SessionNameRaw}}: raw session name, '@' allowed.
SESSION_RENDERED_RAW = r"[\w+=,.@-]{2,64}"

TOKEN_RE = re.compile(r"(\{\{SessionNameRaw\}\}|\{\{SessionName\}\})")

def template_to_regex(template: str) -> Tuple[str, Pattern]:
    """
    Convert an aws-iam-authenticator username template into a regex that matches
    all concrete usernames. Supports multiple {{SessionName}} and/or
    {{SessionNameRaw}} placeholders.

    If a token appears multiple times, all occurrences of that token type must be identical.
    (We can't enforce equality *between* SessionName and SessionNameRaw in pure regex.)
    """
    parts = TOKEN_RE.split(template)  # split on the raw template (NOT escaped)
    built = []
    seen = {"sess": False, "sessraw": False}

    for p in parts:
        if p == SESSION_TOKEN:
            if not seen["sess"]:
                built.append(f"(?P<sess>{SESSION_RENDERED})")
                seen["sess"] = True
            else:
                built.append(r"(?P=sess)")
        elif p == SESSION_RAW_TOKEN:
            if not seen["sessraw"]:
                built.append(f"(?P<sessraw>{SESSION_RENDERED_RAW})")
                seen["sessraw"] = True
            else:
                built.append(r"(?P=sessraw)")
        elif p:  # literal segment
            built.append(re.escape(p))

    regex_str = "^" + "".join(built) + "$"
    return regex_str, re.compile(regex_str)


# ---------- quick self-test ----------
if __name__ == "__main__":
    r1, c1 = template_to_regex("subimage-scan{{SessionName}}")
    print(r1)  # ^subimage\-scan(?P<sess>[\w+=,.-]{2,64})$
    print(bool(c1.match("subimage-scanfoo")))      # True
    print(bool(c1.match("subimage-scan")))         # False
    print(bool(c1.match("subimage-scanuser@x")))   # False

    r2, c2 = template_to_regex("subimage-{{SessionNameRaw}}-scan")
    print(r2)  # ^subimage\-(?P<sessraw>[\w+=,.@-]{2,64})\-scan$
    print(bool(c2.match("subimage-a@b-scan")))     # True

    r3, c3 = template_to_regex("x-{{SessionName}}-y-{{SessionName}}")
    print(bool(c3.match("x-abc-y-abc")))           # True
    print(bool(c3.match("x-abc-y-def")))           # False

    r4, c4 = template_to_regex("u-{{SessionName}}-v-{{SessionNameRaw}}")
    print(bool(c4.match("u-abc-v-abc")))           # True
    print(bool(c4.match("u-a-b-v-a@b")))           # True
```

kunaals: In addition to making a follow up GitHub issue for supporting {{AccessKeyID}} and {{EC2PrivateDNSName}} we should make an issue for the fact that when assuming roles using AWS Identity Center the session name is tied to your AWS Identity Center username

cartography/intel/kubernetes/__init__.py
```diff
@@ -38,6 +42,28 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
             common_job_parameters["CLUSTER_ID"] = cluster_info.get("id")
 
             sync_namespaces(session, client, config.update_tag, common_job_parameters)
+            sync_kubernetes_rbac(
+                session, client, config.update_tag, common_job_parameters
+            )
+            # EKS identity provider sync (requires AWS regions)
+            if config.aws_regions:
+                boto3_session = boto3.Session()
```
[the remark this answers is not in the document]
kunaals: I think this is legit given that we are making AWS API calls here

cartography/intel/kubernetes/eks.py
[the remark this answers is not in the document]
kunaals: I think this is legit given that we are making AWS API calls here

cartography/intel/kubernetes/eks.py
```diff
@@ -1,25 +1,276 @@
...
+
+    return processed_string
+
+
+def template_to_regex(template_string: str) -> str:
+    """
+    Convert a templated string into a regular expression pattern.
+
+    Supports:
+    - {{AccountID}}: Matches 12-digit AWS account IDs
```
kunaals: We should not have to regex match on account IDs because we'll statically put these into the template. We'll always know the account ID from the ARN of the role in either mapUsers or mapRoles or from the list of account IDs in mapAccounts

cartography/intel/kubernetes/eks.py
```diff
@@ -1,25 +1,276 @@
...
+    - {{SessionNameRaw}}: Matches session names (preserves special characters)
+    - {{SessionName}}: Matches session names (transliterated special characters)
+    """
+    if not template_string:
+        return ""
```
kunaals: lets raise an exception instead here, a ValueError

cartography/intel/kubernetes/eks.py
```diff
@@ -1,25 +1,276 @@
...
+
+    # Start with the template string
+    pattern = template_string
+
+    # Replace templates with regex capture groups
+    pattern = pattern.replace("{{AccountID}}", r"(\d{12})")
+    pattern = pattern.replace("{{SessionNameRaw}}", r"([^/\s]+)")
+    pattern = pattern.replace(
+        "{{SessionName}}", r"([^/\s]+)"
+    )  # Same pattern as SessionNameRaw
```
kunaals: > `# Same pattern as SessionNameRaw`

These sorts of comments are superfluous and do more harm than good

kunaals (commented):
i just took a look and this will need some more work. i'd recommend thinking about the procedural decomposition a bit more and diagramming the workflow. i can leave a more thorough review after UK dinner time today, this will need a couple hours from me to give a good review in the current state

in my head this should be broken down into like:
1. mapRoles and mapUsers

- replace AWS account id in group or username
- create regex matcher for group/username
- match on those groups/usernames from all existing kube subjects
- draw those relationships

2. mapAccounts

- find out what Kube subjects are valid AWS arns
- filter out those which are in mapRoles and mapUsers (this is why we do this second)
- extract AWS account ID from each and see if that account ID is in mapAccounts
- if so, draw connection between kube subject and AWS principal

cartography/intel/kubernetes/__init__.py
```diff
@@ -42,6 +45,25 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
...
+
+                sync_eks(
+                    session,
+                    client,
+                    boto3_session,
+                    region,
+                    config.update_tag,
+                    cluster_info.get("id", ""),
+                    client.name,
+                )
```
kunaals:
```suggestion
            # EKS identity provider sync
            boto3_session = boto3.Session()
            region = get_region_from_arn(client.core.name)
            sync_eks(
                session,
                client,
                boto3_session,
                region,
                config.update_tag,
                cluster_info.get("id", ""),
                client.name,
            )
```

kunaals: Add this method above ` def start_k8s_ingestion`

```
def get_region_from_arn(arn: str) -> str:
    """
    Given a Kubernetes cluster ARN like arn:aws:eks:us-east-1:205930638578:cluster/infra-test-eks
    return us-east-1
    """
    parts = arn.split(":")
    if len(parts) > 3:
        ValueError(f"Invalid ARN: {arn}")
    return parse_and_validate_aws_regions(parts[3])[0]
```

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,410 @@
...
+    :return: Dictionary with 'roles' and 'users' keys containing their respective mappings
+    """
+    result: Dict[str, List[Dict[str, Any]]] = {"roles": [], "users": []}
+
+    # Parse mapRoles
+    if "mapRoles" in configmap.data:
+        map_roles_yaml = configmap.data["mapRoles"]
+        role_mappings = yaml.safe_load(map_roles_yaml) or []
+
+        # Filter out templated entries because these are not real users
```
kunaals:
```suggestion
        # Filter out templated entries for now (insert GH issue about templated usernames and groups here)
```

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,410 @@
...
+        )
+    else:
+        logger.info("No mapRoles found in aws-auth ConfigMap")
+
+    # Parse mapUsers
+    if "mapUsers" in configmap.data:
+        map_users_yaml = configmap.data["mapUsers"]
+        user_mappings = yaml.safe_load(map_users_yaml) or []
+
+        # Filter out templated entries because these are not real users
```
kunaals:
```suggestion
```
dont repeat yourself

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,410 @@
...
+        logger.info(
+            f"Parsed {len(filtered_user_mappings)} user mappings from aws-auth ConfigMap"
+        )
+    else:
+        logger.info("No mapUsers found in aws-auth ConfigMap")
+
+    return result
+
+
+def extract_username_from_arn(arn: str) -> str:
```
kunaals: If there is an entry in the configmap like 
```
    - "groups":
      - "system:authenticated"
      "rolearn": "arn:aws:iam::123456789999:role/beepboop"
```
with _no username_ then it will have **no username** in kube. I checked the audit logs, there is no default username whatsoever. In that way - it should have no KubernetesUser either. As such I think this method is misleading. If there is no username field I think we should just point the role to **groups only**. So I think we should get rid of this method entirely and special case when there is no username. Just point the AWSRole or AWSUser to groups only. If there are groups and users both, we can point the AWSRole or AWSUser to the relevant KubernetesUser and KubernetesGroups

cartography/intel/kubernetes/rbac.py
```diff
@@ -263,6 +266,62 @@ def transform_cluster_role_bindings(
...
+    """
+    # Extract all users from rolebindings and clusterrolebindings
+    all_users = {
+        subject.name
+        for binding in chain(
+            role_bindings, cluster_role_bindings
+        )  # iterates through combined bindings and role bindings
+        for subject in (
+            binding.subjects or []
+        )  # iterating through each bindings subjects to get unique users
```
kunaals:
```suggestion
        )  # iterate through each binding's subjects to get unique users
```

cartography/intel/kubernetes/rbac.py
```diff
@@ -263,6 +266,62 @@ def transform_cluster_role_bindings(
...
+) -> List[Dict[str, Any]]:
+    """
+    Transform Kubernetes Users from RoleBindings and ClusterRoleBindings into a list of dicts.
+    """
+    # Extract all users from rolebindings and clusterrolebindings
+    all_users = {
+        subject.name
+        for binding in chain(
+            role_bindings, cluster_role_bindings
+        )  # iterates through combined bindings and role bindings
```
kunaals:
```suggestion
        )  # iterate through combined bindings and role bindings
```
better to use the imperative form for comments like this

cartography/intel/kubernetes/eks.py
kunaals: not to be confused with the behavior in mapAccounts which uses the canonicalArn of the user or role as the default username

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,410 @@
...
+    """
+    logger.info(f"Starting EKS identity provider sync for cluster {cluster_name}")
+
+    # 1. Sync AWS IAM mappings (aws-auth ConfigMap)
+    logger.info("Syncing AWS IAM mappings from aws-auth ConfigMap")
+    configmap = get_aws_auth_configmap(k8s_client)
+    auth_mappings = parse_aws_auth_map(configmap)
+
+    # Transform and load both role and user mappings
+    if auth_mappings["roles"] or auth_mappings["users"]:
```
kunaals: If roles doesn't exist this may result in a KeyNotFound and fail prematurely


```suggestion
    if auth_mappings.get("roles") or auth_mappings.get("users"):
```

cartography/intel/kubernetes/eks.py
```diff
@@ -0,0 +1,410 @@
...
+            transformed_data["users"],
+            transformed_data["groups"],
+            update_tag,
+            cluster_id,
+            cluster_name,
+        )
+        logger.info(
+            f"Successfully synced {len(auth_mappings['roles'])} AWS IAM role mappings "
+            f"and {len(auth_mappings['users'])} AWS IAM user mappings"
+        )
```
kunaals:
```suggestion
        logger.info(
            f"Successfully synced {len(auth_mappings.get('roles'))} AWS IAM role mappings "
            f"and {len(auth_mappings.get('users'))} AWS IAM user mappings"
        )
```

kunaals: ditto

cartography/models/kubernetes/groups.py
```diff
@@ -0,0 +1,109 @@
...
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class KubernetesGroupNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id")
+    name: PropertyRef = PropertyRef("name")
+    cluster_name: PropertyRef = PropertyRef("cluster_name")
+    aws_role_arn: PropertyRef = PropertyRef("aws_role_arn")
```
kunaals: This seems wrong given that multiple AWSRoles and AWSUsers can be associated with a given KubernetesGroup? 

example pseudocode aws-authmap

mapRoles:
 - arn: role/whatever
   - groups: foo
   username: bar
 - arn: role/beepboop
   - groups: foo
   username: baz

cartography/models/kubernetes/oidc.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@dataclass(frozen=True)
+class KubernetesOIDCProviderNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id")
+    issuer_url: PropertyRef = PropertyRef("issuer_url")
+    cluster_name: PropertyRef = PropertyRef("cluster_name")
+    k8s_platform: PropertyRef = PropertyRef(
+        "k8s_platform"
+    )  # right now we just have eks as a platform
```
kunaals:
```suggestion
    )
```
im sure we'd forget to remove this comment when we add AKS and GKE, no real value in having it

cartography/models/kubernetes/oidc.py
```diff
@@ -0,0 +1,56 @@
...
+    client_id: PropertyRef = PropertyRef("client_id")
+    status: PropertyRef = PropertyRef("status")
+    name: PropertyRef = PropertyRef("name")
+    arn: PropertyRef = PropertyRef(
+        "arn"
+    )  # AWSEKS-specific, will be None for other platforms
```
kunaals:
```suggestion
    aws_arn: PropertyRef = PropertyRef(
        "arn"
    ) 
```
lets make it explicit in the name of the field then, or lets just store the cluster_id here which maps directly to the cluster ID in Cartography to make this more generic. in KubernetesOIDCProviderToClusterRel instead of making a target_node_matcher like `{"id": PropertyRef("CLUSTER_ID", set_in_kwargs=True)}` we could instead have `{"id": PropertyRef("cluster_id")}`

kunaals: perhaps we dont even need this arn field, we could just remove it. when i go to the EKS console and click on the Okta OIDC provider they dont even surface it on the page

docs/root/modules/kubernetes/schema.md
```diff
@@ -335,3 +338,27 @@ Representation of a [Kubernetes ClusterRoleBinding.](https://kubernetes.io/docs/
     (:KubernetesClusterRoleBinding)-[:SUBJECT]->(:KubernetesServiceAccount)
     (:KubernetesClusterRoleBinding)-[:ROLE_REF]->(:KubernetesClusterRole)
     ```
+
+### KubernetesOIDCProvider
+Representation of an external OIDC identity provider infrastructure configuration for a Kubernetes cluster. This node contains the technical configuration details of how the cluster is set up to trust external identity systems (such as Auth0, Okta, Azure AD). The actual identity mapping between external users/groups and Kubernetes RBAC identities is handled by direct relationships from the external identity provider modules.
```
kunaals:
```suggestion
Representation of an external OIDC identity provider for a Kubernetes cluster. This node contains the configuration details of how the cluster is set up to trust external identity systems (such as Auth0, Okta, Entra). The ingestion of users/groups from the identity provider is handled by the respective identity provider Cartography module. Then the Kubernetes module creates relationships between those identities and KubernetesUsers and KubernetesGroups.
```

docs/root/modules/kubernetes/schema.md
```diff
@@ -335,3 +338,27 @@ Representation of a [Kubernetes ClusterRoleBinding.](https://kubernetes.io/docs/
...
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+
+#### Relationships
+- `KubernetesOIDCProvider` is trusted by a `KubernetesCluster`.
+    ```
+    (:KubernetesCluster)-[:TRUSTS]->(:KubernetesOIDCProvider)
+    ```
+
+Note: Identity mapping between external OIDC providers (Okta, Auth0, etc.) and Kubernetes users/groups is handled through direct relationships from the external identity provider nodes to Kubernetes nodes, not through the `KubernetesOIDCProvider` infrastructure node.
```
kunaals:
```suggestion
Note: Identity mapping between external OIDC providers (Okta, Auth0, etc.) and Kubernetes users/groups is handled through direct relationships from the external identity provider nodes to Kubernetes nodes, not through the `KubernetesOIDCProvider` metadata node.
```

tests/data/kubernetes/rbac.py
```diff
@@ -379,3 +423,55 @@
         "status_phase": "Active",
     },
 ]
+
+# Mock Okta Users for identity mapping tests
+# These map to existing Kubernetes users in the RBAC test data
+MOCK_OKTA_USERS = [
+    {
+        "id": "okta-user-1",
+        "email": "john.doe",  # Maps to existing K8s user "john.doe"
```
kunaals: `email` and `login` here should both be in the form of an email eg `john.doe@company.com`

tests/integration/cartography/intel/kubernetes/test_eks.py
```diff
@@ -0,0 +1,226 @@
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from kubernetes.client.models import V1ConfigMap
+
+from cartography.intel.aws.iam import load_roles
+from cartography.intel.aws.iam import load_users
+from cartography.intel.kubernetes.clusters import load_kubernetes_cluster
+from cartography.intel.kubernetes.eks import sync as eks_sync
```
kunaals: Let's call this sync_eks instead of eks_sync because that's what you call it in __init__.py under intel/kubernetes

tests/integration/cartography/intel/kubernetes/test_eks.py
```diff
@@ -0,0 +1,226 @@
...
+
+    # Assert: Verify AWS User to Kubernetes User relationships
+    expected_user_user_relationships = {
+        ("arn:aws:iam::123456789012:user/alice", "test-cluster/alice-user"),
+        ("arn:aws:iam::123456789012:user/bob", "test-cluster/bob-user"),
+        ("arn:aws:iam::123456789012:user/charlie", "test-cluster/charlie-user"),
+        (
+            "arn:aws:iam::123456789012:user/dana",
+            "test-cluster/dana",
+        ),  # Defaulted username
```
kunaals: Remember there are no default usernames in kube unless using mapAccounts so this part of the test might be moot now

tests/data/kubernetes/eks.py
```diff
@@ -0,0 +1,191 @@
...
+        "Arn": "arn:aws:iam::123456789012:user/service-account",
+        "UserName": "service-account",
+        "Path": "/",
+        "UserId": "AIDACKCEVSQ6C2EXAMPLE5",
+        "CreateDate": "2023-01-01T00:00:00Z",
+    },
+]
+
+# Mock OIDC provider data (raw AWS API responses)
+MOCK_OIDC_PROVIDERS = [
```
kunaals: In EKS you can only have 1 external OIDC provider so let's make this Okta only

tests/integration/cartography/intel/kubernetes/test_rbac.py
```diff
@@ -1,61 +1,84 @@
...
-import pytest
-
+from cartography.intel.kubernetes import rbac
 from cartography.intel.kubernetes.clusters import load_kubernetes_cluster
 from cartography.intel.kubernetes.namespaces import load_namespaces
 from cartography.intel.kubernetes.rbac import sync_kubernetes_rbac
+from cartography.intel.okta.groups import _load_okta_groups
+from cartography.intel.okta.organization import create_okta_organization
+
+# Import Okta loading functions
```
kunaals:
```suggestion
```

kunaals: remove superfluous comment

kunaals:
Follow up items:
- mapAccounts (https://github.com/cartography-cncf/cartography/issues/1857)
- the AWS role that created the cluster is added to mapUsers by default with the username admin. However, we filter our roles listed in mapUsers. we'll have to special case this.
- EKS Access Entries (https://github.com/cartography-cncf/cartography/issues/1855)
- Templated entries in the aws-auth configmap (https://github.com/cartography-cncf/cartography/issues/1854)

cartography/models/kubernetes/groups.py
shyammukund: A little confused on this comment. If we use your example the current transform logic would create 2 KubeGroups
KubeGroup: id="cluster/foo", aws_role_arn="role/whatever"
KubeGroup: id="cluster/foo", aws_role_arn="role/beepboop"

When we load that first entry 
- AWSRole(with arn "role/whatever")---MAPS_TO--->KubeGroup
When we load the second entry
- AWSRole(with arn"role/beepboop")---MAPS_TO--->KubeGroup

Even though the group has the same id the schema should make sure that two seperate MAP_TO rels are created and the MERGE functionality in schema loading will make it so two seperate nodes aren't created for the same group. Maybe i'm misunderstanding what you're saying

kunaals: This is not a rel though this is a node, I guess we'd keep overwriting the aws_role_arn in this case. But the AWS role ARN definitely does not belong as a property on the kube group itself. We should create the group and AWS roles and/or AWS users should point to the group

cartography/intel/kubernetes/eks.py
shyammukund: removed this function. So now an entry with a username will create the intended KubeUser + KubeGroup but an entry without a  username will now just create the associated KubeGroup instead of also creating a fake KubeUser. I left some logging in this function that tells us a little about the amount of entries that didn't have usernames but if you feel this is uncessary i can remove.

cartography/intel/kubernetes/eks.py
shyammukund: updated all the direct dict access in this file to use .get() to stay consistent.

cartography/models/kubernetes/oidc.py
shyammukund: Went ahead and removed this property of the node.

tests/data/kubernetes/eks.py
shyammukund: Called this MOCK_OKTA_PROVIDER since auth0 operates under okta but if you think this is still misleading ill change.

tests/data/kubernetes/eks.py
```diff
@@ -0,0 +1,191 @@
...
+MOCK_OKTA_PROVIDER = [
+    {
+        "identityProviderConfigName": "auth0-provider",
+        "issuerUrl": "https://company.auth0.com/",
+        "clientId": "abc123def456",
+        "usernamePrefix": "auth0:",
+        "groupsPrefix": "auth0:",
+        "status": "ACTIVE",
+        "identityProviderConfigArn": "arn:aws:eks:us-west-2:123456789012:identityproviderconfig/test-cluster/oidc/auth0-provider/12345",
+    },
```
kunaals:
```suggestion
```

tests/integration/cartography/intel/kubernetes/test_eks.py
```diff
@@ -0,0 +1,215 @@
...
+    assert expected_user_group_relationships.issubset(actual_user_group_relationships)
+
+    # Assert: Verify OIDC Provider nodes were created
+    expected_oidc_providers = {
+        (
+            f"{TEST_CLUSTER_NAME}/oidc/auth0-provider",
+            "https://company.auth0.com/",
+            "auth0-provider",
+            "eks",
+        ),
```
kunaals:
```suggestion
```

tests/integration/cartography/intel/kubernetes/test_eks.py
```diff
@@ -0,0 +1,215 @@
...
+    actual_oidc_providers = check_nodes(
+        neo4j_session,
+        "KubernetesOIDCProvider",
+        ["id", "issuer_url", "name", "k8s_platform"],
+    )
+    assert expected_oidc_providers.issubset(actual_oidc_providers)
+
+    # Assert: Verify Cluster TRUSTS OIDC Provider relationships
+    expected_cluster_relationships = {
+        (TEST_CLUSTER_ID, f"{TEST_CLUSTER_NAME}/oidc/auth0-provider"),
```
kunaals:
```suggestion
```

kunaals (changes requested):
one last nit - you forgot to actually delete the auth0 OIDC provider

Issue #1829, opened 2025-08-20
Build offical container for arm64


phmcder:
**Feature request template**

*Title*: *Build offical container for linux/arm64*

*Description*:
> I have been issued a corperate MacBook with a variety of security tools on it which prevents me from building a Arm64 compatible container.
Would it be possible to add the building of an linux/arm64 platform container so I am able to run Cartography from my laptop?

*Please complete the following information:*:
>  - Cartography release version: 0.109.0
>  - OS OSX
>  - Chipset M1 ARM

jychp:
Since we already use the `buildx` GitHub action, this should be straightforward.
I’ll take a look at it.

jychp:
@phmcder arn64 build is now available, let me know if everything work as expected

<img width="770" height="663" alt="Image" src="https://github.com/user-attachments/assets/eb1d61f6-d6af-4430-a575-05f878c988c8" />

phmcder:
@jychp thanks for doing that. Works a treat

PR #1830, opened 2025-08-20
chore: add build for `linux/arm64` docker image #1829

jychp:
### Summary
This PR reworks the CI workflow to support Docker image builds for multiple platforms (`amd64` and `arm64`).

### Context
Previously, the CI only produced Docker images for a single platform. Adding multi-platform support improves portability and ensures Cartography can run seamlessly across a wider range of environments.

### Linked issues
- #1829 

### Changes proposed
- **Multi-platform Docker builds:** Enable builds for `amd64` and `arm64`.
- No parallel matrix builds are used since there are no time constraints or multiple runners to optimize. This avoids the need to merge images before publishing.
- **Job separation:** The Python package build and Docker image build have been split into distinct jobs, allowing independent reruns when needed.

### DIFF
--- a/.github/workflows/publish-to-ghcr-and-pypi.yml
+++ b/.github/workflows/publish-to-ghcr-and-pypi.yml
@@ -9,16 +9,11 @@ on:
 permissions: {}
 
 jobs:
-  build-and-publish-to-pypi-and-ghcr:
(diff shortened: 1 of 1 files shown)


jychp:
There is no change and it's very common to have at least AMD and ARM.

The only thing that changes is the digest which must be adapted if you want to use ARM

PR #1839, opened 2025-08-24
chore: replaced direct requests.session() instantiation with build_session() 

rileydakota:
Continuation of work in #1676 

- #1665 

### Checklist

### DIFF
--- a/cartography/intel/cve/__init__.py
+++ b/cartography/intel/cve/__init__.py
@@ -9,15 +9,14 @@ from urllib3 import Retry
 from cartography.config import Config
 from cartography.intel.cve import feed
 from cartography.stats import get_stats_client
-from cartography.util import merge_module_sync_metadata
-from cartography.util import timeit
+from cartography.util import merge_module_sync_metadata, timeit, build_session
 
 logger = logging.getLogger(__name__)
 stat_handler = get_stats_client(__name__)
 
 
 def _retryable_session() -> Session:
-    session = Session()
+    session = build_session()
     retry_policy = Retry(
         total=8,
         connect=1,
--- a/cartography/intel/cve/__init__.py
+++ b/cartography/intel/cve/__init__.py
@@ -9,7 +9,9 @@ from urllib3 import Retry
 from cartography.config import Config
 from cartography.intel.cve import feed
 from cartography.stats import get_stats_client
-from cartography.util import merge_module_sync_metadata, timeit, build_session
+from cartography.util import build_session
+from cartography.util import merge_module_sync_metadata
+from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
 stat_handler = get_stats_client(__name__)
--- a/cartography/intel/kandji/devices.py
+++ b/cartography/intel/kandji/devices.py
@@ -10,7 +10,7 @@ from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.models.kandji.device import KandjiDeviceSchema
 from cartography.models.kandji.tenant import KandjiTenantSchema
-from cartography.util import timeit
+from cartography.util import timeit, build_session
 
 logger = logging.getLogger(__name__)
 _TIMEOUT = (60, 60)
@@ -33,7 +33,7 @@ def get(kandji_base_uri: str, kandji_token: str) -> List[Dict[str, Any]]:
     }
 
     devices: List[Dict[str, Any]] = []
-    session = Session()
+    session = build_session()
     while True:
         logger.debug("Kandji device offset: %s", offset)
 
--- a/cartography/intel/kandji/devices.py
+++ b/cartography/intel/kandji/devices.py
@@ -4,13 +4,13 @@ from typing import Dict
 from typing import List
 
 import neo4j
-from requests import Session
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.models.kandji.device import KandjiDeviceSchema
 from cartography.models.kandji.tenant import KandjiTenantSchema
-from cartography.util import timeit, build_session
+from cartography.util import build_session
+from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
 _TIMEOUT = (60, 60)
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -25,6 +25,7 @@ import backoff
 import boto3
 import botocore
 import neo4j
+import requests
 
 from cartography.graph.job import GraphJob
 from cartography.graph.statement import get_job_shortname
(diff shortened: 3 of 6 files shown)


cartography/intel/cve/__init__.py
```diff
@@ -17,7 +18,7 @@
 
 
 def _retryable_session() -> Session:
-    session = Session()
+    session = build_session()
     retry_policy = Retry(
```
jychp: Maybe we could put this retry policy in our generic builder (and make it configurable).

cartography/intel/kandji/devices.py
```diff
@@ -33,7 +33,7 @@ def get(kandji_base_uri: str, kandji_token: str) -> List[Dict[str, Any]]:
     }
 
     devices: List[Dict[str, Any]] = []
-    session = Session()
+    session = build_session()
```
jychp, line 32: Be careful the line 42 with headers=headers will override the user_agent.
You should remove this parameter and explicitly do a session.headers.update() instead

cartography/util.py
```diff
@@ -473,3 +474,19 @@ async def another_async_func(my_arg2):
     results = to_synchronous(future_1, future_2)
     """
     return asyncio.get_event_loop().run_until_complete(asyncio.gather(*awaitables))
+
+
+try:
```
jychp: We recently added a more reliable function for that (for node metadata):
https://github.com/cartography-cncf/cartography/blob/c7acdee1df558280c048d9eb408be7056cc682de/cartography/graph/querybuilder.py#L705

You should move this function to cartography/util and use it

cartography/util.py
```diff
@@ -473,3 +474,19 @@ async def another_async_func(my_arg2):
...
+    cartography_version = "unknown"
+
+
+def build_session() -> requests.Session:
+    """
+    Create a requests.Session with a custom User-Agent header that includes the Cartography version.
+    """
+    session = requests.Session()
+    user_agent = f"Cartography/{cartography_version}"
+    session.headers.update({"User-Agent": user_agent})
```
jychp, line 514: Idea (that could be done in a separate PR if wanted):

In every module we define _TIMEOUT(60, 60) and add it to every requests (but sometime it could be omitted).

We can add a timeout parameter to our builder (with (60, 60) as default value) and use a custom adapter to define the timeout at the session level:

```
TIMEOUT = (5, 5)

class MyHTTPAdapter(requests.adapters.HTTPAdapter):
    def __init__(self, timeout=None, *args, **kwargs):
        self.timeout = timeout
        super(MyHTTPAdapter, self).__init__(*args, **kwargs)

    def send(self, *args, **kwargs):
        kwargs['timeout'] = self.timeout
        return super(MyHTTPAdapter, self).send(*args, **kwargs)


session = requests.Session()
session.mount("http://", MyHTTPAdapter(TIMEOUT))
session.mount("https://", MyHTTPAdapter(TIMEOUT))
```

tests/unit/cartography/intel/kandji/test_devices.py
```diff
@@ -65,7 +65,7 @@ def test_get_devices_single_page(
     assert mock_session.return_value.get.call_count == 2
 
 
-@patch("cartography.intel.kandji.devices.Session")
+@patch("cartography.intel.kandji.devices.build_session")
 def test_get_devices_with_pagination(
```
jychp, line 69: We should add a test where we build (not send) the query and inspect to check if the header is present.

jychp (changes requested):
Thanks for continuing working on this PR

Two changes, 1 test suggested, 2 ideas that could be done in a separate PR

cartography/util.py
rileydakota: @jychp I set this up locally, but it appears to create a circular dependency when included in `util.py`. I do have it working when included in a `version.py`, and then importing from there, but not sure I love having a single file for a version function. Almost feels like it would be better to break util up into smaller sub modules, but that would be alot of work. Thoughts?

jychp: Oh yeah, now you mentioned it I remember having this issue when I implemented this metadata ...

I think for now you can just import the existing function (to avoid code duplication) and add a `TODO:` comment to find a better solution later, this will require some refactoring and this is out of this PR scope IMHO.

cartography/intel/kandji/devices.py
rileydakota: Moved it to using `session.headers.update()` instead of the `headers` argument. Will push after we figure out the best path for the `util` circular dep below.

cartography/util.py
rileydakota: Any issues with the Adapter class living in `util.py`? If not, happy to add it in this PR.

rileydakota: @jychp added it in the latest commit, let me know your thoughts here

tests/unit/cartography/intel/kandji/test_devices.py
rileydakota: @jychp Basically just use `build_session` to construct the Session and inspect in the test?

rileydakota: @jychp I added some test cases in in `test_util` that specifically test the `build_session` function. Take a look when you get a minute 🙏

rileydakota:
@jychp ready for another look when you have time 🙏

cartography/util.py
```diff
@@ -481,3 +485,45 @@ async def another_async_func(my_arg2):
...
+    # Single adapter handles both concerns
+    adapter = RetryTimeoutAdapter(
+        timeout=timeout if timeout else None, max_retries=retry_policy or Retry(total=0)
+    )
+
+    session.mount("https://", adapter)
+    session.mount("http://", adapter)
+
+    if retry_policy:
+        logger.info(f"Configured session with retry policy: {retry_policy}")
```
jychp, line 525: Non blocking: I think this should be a DEBUG level log.

cartography/util.py
```diff
@@ -481,3 +485,45 @@ async def another_async_func(my_arg2):
...
+    """
+    Create a requests.Session with a custom User-Agent header that includes the Cartography version.
+    """
+    session = requests.Session()
+    user_agent = f"Cartography/{_get_cartography_version()}"
+    session.headers.update({"User-Agent": user_agent})
+
+    # Single adapter handles both concerns
+    adapter = RetryTimeoutAdapter(
+        timeout=timeout if timeout else None, max_retries=retry_policy or Retry(total=0)
```
jychp, line 518: We define a default value for timeout, this should be always defined (and if manually set to None, then you do not need to force it, so the condition is not necessary).

Do we need to define Retry(total=0), can we just use None ?

jychp (approved):
LGTM
Some minor and none blocking comments

Issue #1843, opened 2025-08-26
AWS APIGatewayRestAPI sync potentially crashing on rate limit


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

See log below.

I would expect that we would handle this through either the `aws_handle_regions` decorator or some backoff decoator but it appears that the code is failing instead. Or maybe the retries have already been exhausted.

*To Reproduce*:
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run API Gateway  REST API sync on certain env

*Logs*:
> If applicable, copy and paste your console log with the failing stack trace.

```
[2025-08-25T09:34:37Z] INFO cartography.graph.job: Finished job aws_lambda_ecr
[2025-08-25T09:34:37Z] ERROR sync.1756103243919-11332670: Error during Cartography sync: Traceback (most recent call last):
  File "/app/app/sync.py", line 690, in run_sync
    future.result()
    ~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/app/app/sync.py", line 280, in wrapper
    return fn(*args, **kwargs)
  File "/app/app/sync.py", line 668, in sync_account
    _sync_one_account(
    ~~~~~~~~~~~~~~~~~^
        neo4j_session=neo4j_session,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
        aws_requested_syncs=aws_resource_functions,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/app/.venv/lib/python3.13/site-packages/cartography/intel/aws/__init__.py", line 79, in _sync_one_account
    RESOURCE_FUNCTIONS[func_name](**sync_args)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
[...]

PR #1846, opened 2025-08-26
fix: #1843: Handle APIGateway rate limit errors

achantavy:
Avoids crashing when API Gateway `get_rest_api_resources` encounters a TooManyRequestsException by decorating it with `aws_handle_regions`, which serves as a general catch-all exc handler for AWS `get` functions for retries and giving up.

Updates unit tests.

### DIFF
--- a/cartography/intel/aws/apigateway.py
+++ b/cartography/intel/aws/apigateway.py
@@ -101,6 +101,7 @@ def get_rest_api_details(
 
 
 @timeit
+@aws_handle_regions
 def get_rest_api_stages(api: Dict, client: botocore.client.BaseClient) -> Any:
     """
     Gets the REST API Stage Resources.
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/apigateway.py
```diff
@@ -142,6 +143,7 @@ def get_rest_api_client_certificate(
 
 
 @timeit
+@aws_handle_regions
```
achantavy, line 146: this is the main fix

cartography/intel/aws/apigateway.py
```diff
@@ -101,6 +101,7 @@ def get_rest_api_details(
...
```
achantavy, line 104: might as well add this decorator here too since it's the same pattern of problem

Issue #381, opened 2020-07-31
GCP cleanup jobs should be scoped to the current project


achantavy:
*Description*:
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

As an example, if we follow the flow of a gcp.compute.instances sync, https://github.com/lyft/cartography/blob/4ae9d32f4b43f8fe93db7c7f7583fb098b105784/cartography/intel/gcp/compute.py#L902-L905

notice that the cleanup job is not scoped to a specific project:
https://github.com/lyft/cartography/blob/4ae9d32f4b43f8fe93db7c7f7583fb098b105784/cartography/data/jobs/cleanup/gcp_compute_instance_cleanup.json#L10

which means that every instance will get cleaned up on every cleanup call regardless of the project it belongs to. This problem also applies to GKE and the other GCP resources. This can result in premature cleanups and can cause a minor performance hit since we're querying for all projects every single time.

*Please complete the following information:*:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]

4ae9d32f4b43f8fe93db7c7f7583fb098b105784

> Add any other context about the problem here.

The AWS sync solves this issue by scoping the cleanup to the current AWSAccount: https://github.com/lyft/cartography/blob/4ae9d32f4b43f8fe93db7c7f7583fb098b105784/cartography/data/jobs/cleanup/aws_import_ec2_instances_cleanup.json#L8

cc: @marco-lancini @ramonpetgrave64 for viz.

ramonpetgrave64:
What about doing the cleanup once per GCP Org?
And `account_specific_common_job_parameters = {**common_job_parameters, 'AWS_ID': account_id}` could be a better way of setting parameters in https://github.com/lyft/cartography/blob/00e0e98c6af3bd0489e59b0dbb6670ba85b71786/cartography/intel/aws/__init__.py#L89-L98

PR #1878, opened 2025-09-05
refactor: GCP storage bucket sync to new data model #381

achantavy:
Fixes part of #381

## Summary
- Refactor GCP Storage bucket ingestion to use new node schema and generic loader
- Add data model definitions for GCPBucket and GCPBucketLabel
- Update tests for new transform and loading workflow
- Ensure cleanup removes bucket labels before buckets to avoid dangling references

### DIFF
--- a/cartography/intel/gcp/storage.py
+++ b/cartography/intel/gcp/storage.py
@@ -62,7 +62,7 @@ def get_gcp_buckets(storage: Resource, project_id: str) -> Dict:
 
 
 @timeit
-def transform_gcp_buckets(bucket_res: Dict) -> Tuple[List[Dict], List[Dict]]:
+def transform_gcp_buckets_and_labels(bucket_res: Dict) -> Tuple[List[Dict], List[Dict]]:
     """
     Transform the GCP Storage Bucket response object for Neo4j ingestion.
 
@@ -190,7 +190,7 @@ def sync_gcp_buckets(
     """
     logger.info("Syncing Storage objects for project %s.", project_id)
     storage_res = get_gcp_buckets(storage, project_id)
-    buckets, bucket_labels = transform_gcp_buckets(storage_res)
+    buckets, bucket_labels = transform_gcp_buckets_and_labels(storage_res)
     load_gcp_buckets(neo4j_session, buckets, project_id, gcp_update_tag)
     load_gcp_bucket_labels(neo4j_session, bucket_labels, project_id, gcp_update_tag)
     cleanup_gcp_buckets(neo4j_session, common_job_parameters)
--- a/tests/integration/cartography/intel/gcp/test_storage.py
+++ b/tests/integration/cartography/intel/gcp/test_storage.py
@@ -14,8 +14,8 @@ def _ensure_local_neo4j_has_test_storage_bucket_data(neo4j_session):
     _create_test_project(neo4j_session)
 
     bucket_res = tests.data.gcp.storage.STORAGE_RESPONSE
-    buckets, bucket_labels = cartography.intel.gcp.storage.transform_gcp_buckets(
-        bucket_res
+    buckets, bucket_labels = (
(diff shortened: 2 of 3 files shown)


cartography/models/gcp/storage/bucket.py
```diff
@@ -0,0 +1,117 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class GCPBucketNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id", extra_index=True)
```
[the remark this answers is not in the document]
kunaals: Similar to AWS, a given GCP bucket name (which is the "id" here) is globally unique across all storage accounts across all of GCP: https://cloud.google.com/storage/docs/buckets

cartography/intel/gcp/storage.py
```diff
@@ -58,188 +62,98 @@ def get_gcp_buckets(storage: Resource, project_id: str) -> Dict:
...
-    """
-
-    query = """
-    MERGE(p:GCPProject{projectnumber:$ProjectNumber})
-    ON CREATE SET p.firstseen = timestamp()
-    SET p.lastupdated = $gcp_update_tag
-
-    MERGE(bucket:GCPBucket{id:$BucketId})
-    ON CREATE SET bucket.firstseen = timestamp(),
-    bucket.bucket_id = $BucketId
```
kunaals, line 138: We lose this `bucket_id` field in the data model. I think this is fine, it's redundant, but we should potentially call this out in the release notes in case someone is relying on this

cartography/intel/gcp/storage.py
```diff
@@ -58,188 +62,98 @@ def get_gcp_buckets(storage: Resource, project_id: str) -> Dict:
...
-    bucket.log_bucket = $LogBucket,
-    bucket.requester_pays = $RequesterPays,
-    bucket.default_kms_key_name = $DefaultKmsKeyName
-
-    MERGE (p)-[r:RESOURCE]->(bucket)
-    ON CREATE SET r.firstseen = timestamp()
-    SET r.lastupdated = $gcp_update_tag
-    """
-    for bucket in buckets:
-        neo4j_session.run(
```
kunaals, line 162: There are 3 fields on the `bucket` dict that are not persisted to the graph:
- etag: https://cloud.google.com/storage/docs/metadata#etags
- updated: https://cloud.google.com/storage/docs/metadata#modification-time
- default_event_based_hold: https://cloud.google.com/storage/docs/object-holds

etag seems kinda useless, though updated and the object holds seem useful. This data model migration doesn't include these fields which is fine, but in a fast-follow PR is might be nice to add these features that we used to "get" but not actually include in the graph

cartography/intel/gcp/storage.py
```diff
@@ -58,188 +62,98 @@ def get_gcp_buckets(storage: Resource, project_id: str) -> Dict:
 
 
 @timeit
-def transform_gcp_buckets(bucket_res: Dict) -> List[Dict]:
+def transform_gcp_buckets(bucket_res: Dict) -> Tuple[List[Dict], List[Dict]]:
```
kunaals: Given that we transform buckets and bucket labels here and load them separately, let's call this `transform_gcp_buckets_and_labels` to make things a bit clearer? Optionally we could separate this into two methods but I dont think that's required until this module uses generators

kunaals (approved):
left a few nits, nothing blocking. also this seems similar to https://github.com/cartography-cncf/cartography/pull/1791

cartography/intel/gcp/storage.py
achantavy: Good point, will fix. No reason to break interface on something this small

PR #1879, opened 2025-09-05
Add integration tests for GCP project sync

achantavy:
## Summary
- add integration tests covering GCP project sync behavior, including parent linkage, parentless projects, and cleanup

## Testing
- `pytest tests/integration/cartography/intel/gcp/test_crm.py::test_sync_gcp_projects` *(fails: Couldn't connect to localhost:7687)*

------
https://chatgpt.com/codex/tasks/task_b_68ba75a11db8832f8254e4d337944463

### DIFF
--- a/tests/integration/cartography/intel/gcp/test_crm.py
+++ b/tests/integration/cartography/intel/gcp/test_crm.py
@@ -1,7 +1,11 @@
+from unittest.mock import patch
+
 import cartography.intel.gcp.crm
 import tests.data.gcp.crm
+from tests.integration.util import check_nodes, check_rels
 
 TEST_UPDATE_TAG = 123456789
+COMMON_JOB_PARAMS = {"UPDATE_TAG": TEST_UPDATE_TAG}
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/cartography/intel/gcp/test_crm.py
```diff
@@ -156,3 +165,153 @@ def test_load_gcp_projects_without_parent(neo4j_session):
...
+    assert actual_nodes == {
+        (
+            "this-project-has-a-parent-232323",
+            "folders/1414",
+            "organizations/1337",
+        ),
+    }
+
+
+def test_sync_gcp_projects_without_parent(neo4j_session) -> None:
```
kunaals, line 239: There are 2 more cases that I think of for these tests which are nested GCP folders (project -> folder -> folder -> org) and projects that are _not_ in a folder and are associated directly with the org (project -> org). Note that the latter is not the same as no parent which is an "unmanaged" GCP project

kunaals: just pushed the changes

PR #1904, opened 2025-09-13
feat(github): add repo committers

achantavy:
Adds support for repo committers in the last 30 days.

### Checklist

### DIFF
--- a/cartography/intel/github/__init__.py
+++ b/cartography/intel/github/__init__.py
@@ -4,6 +4,7 @@ import logging
 
 import neo4j
 
+import cartography.intel.github.commits
 import cartography.intel.github.repos
 import cartography.intel.github.teams
 import cartography.intel.github.users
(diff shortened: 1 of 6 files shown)


cartography/intel/github/__init__.py
```diff
@@ -1,18 +1,45 @@
...
+    :param organization: GitHub organization name
+    :return: List of repository names
+    """
+    org_url = f"https://github.com/{organization}"
+    query = """
+    MATCH (org:GitHubOrganization {id: $org_url})<-[:OWNER]-(repo:GitHubRepository)
+    RETURN repo.name
+    ORDER BY repo.name
+    """
+    return cast(
```
jychp, line 33: I love to see this kind of hard typing :)

cartography/intel/github/commits.py
```diff
@@ -0,0 +1,418 @@
...
+            cost
+            remaining
+            resetAt
+        }
+    }
+"""
+
+
+@timeit
+def get_repo_commits_l30d(
```
jychp: non blocking: instead of having a hard coded "l30d" we should have that as a parameter with default value to allow modifications/dynamic configuration in the future

achantavy:
> I’m really happy to see all the possibilities that MatchLink opens up.

Yeah on one hand, it might have fewer round trips to pull the commit history and then put that into the GitHubUser property schema itself, but pulling the commit history can be very memory intensive so it doesn't make sense to include both users and commits resident in memory at the same time. Plus the commit history isn't a core property of the GitHubUser, so it makes a bit more sense for it to be a separate thing like this.

PR #1905, opened 2025-09-13
Track root MFA via IAM account summary

achantavy:
## Summary
- collect IAM account summary to track whether the account's root user has MFA enabled
- document new `root_mfa_enabled` field on `AWSAccount`
- add test coverage for account summary loading

## Testing
- `pytest tests/integration/cartography/intel/aws/iam/test_iam.py::test_load_account_summary -q` *(fails: Couldn't connect to localhost:7687)*

------
https://chatgpt.com/codex/tasks/task_b_68c594a11674832fa496f731e1edf8c2

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -1019,6 +1019,49 @@ def sync_user_access_keys(
 
 
 @timeit
+def get_account_summary(boto3_session: boto3.session.Session) -> Dict:
+    client = boto3_session.client("iam")
+    return client.get_account_summary()
+
+
+@timeit
+def load_account_summary(
+    neo4j_session: neo4j.Session,
+    summary: Dict,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    query = (
+        "MERGE (aa:AWSAccount{id: $AWS_ACCOUNT_ID}) "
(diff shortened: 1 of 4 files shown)


cartography/intel/aws/iam.py
```diff
@@ -1018,6 +1018,49 @@ def sync_user_access_keys(
...
+
+
+@timeit
+def load_account_summary(
+    neo4j_session: neo4j.Session,
+    summary: Dict,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    query = (
```
jychp, line 1034: Maybe add a comment here to explain why this update has to be done in a separate call that doesn’t follow our usual pattern or align with the ORM schema.

jychp (approved):
LGTM
Maybe we can add a comment to explain why we handle this property differently.

cartography/intel/aws/iam.py
achantavy: Going to hold back on merging this: the `merge` is very not correct, it should be a `match`

achantavy: but yes will include an explanation

jychp: Yeah, I was questioning whether to use MERGE or MATCH. I concluded this should be acceptable since if this code is reached, the Account will already have been created anyway.

achantavy:
Codex butchered this. This doesn't meet our bar; closing

PR #1914, opened 2025-09-15
Ksikka/gcp projects folders subresource

kunaals:

### Checklist


cartography/intel/gcp/crm/folders.py
```diff
@@ -3,30 +3,51 @@
...
+    :return: List of folder dicts with 'name' field containing full resource names (e.g., "folders/123456")
     """
+    results: List[Dict] = []
+    client = resourcemanager_v3.FoldersClient()
     try:
-        req = crm_v2.folders().search(body={})
-        res = req.execute()
-        return res.get("folders", [])
-    except HttpError as e:
+        # BFS over folders starting at the org root
```
achantavy, line 27: okok leetcode superstar ok

cartography/intel/gcp/crm/folders.py
```diff
@@ -3,30 +3,51 @@
...
+                        "name": folder.name,
+                        "parent": parent,
+                        "displayName": folder.display_name,
+                        "lifecycleState": folder.state.name,
+                    }
+                )
+                if folder.name:
+                    queue.append(folder.name)
+        return results
+    except Exception as e:
```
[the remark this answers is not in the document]
achantavy: I'd err toward not catching the exception and failing loudly as well

cartography/intel/gcp/crm/folders.py
```diff
@@ -37,68 +58,55 @@ def load_gcp_folders(
...
     data: List[Dict],
     gcp_update_tag: int,
+    org_resource_name: str,
 ) -> None:
     """
-    Ingest the GCP folders to Neo4j.
+    Load GCP folders into the graph.
+    :param org_resource_name: Full organization resource name (e.g., "organizations/123456789012")
     """
+    # Transform data in place: add parent_org or parent_folder fields based on parent type
```
achantavy, line 67: Nit: not going to block on this, but it would be more consistent with other modules to create a separate transform function that operates on the data as a copy to keep a sort of functional style. I do like having transform funcs where it's easy later on to see what are the key names and values of each dict in the list. Again, not a blocker though.

cartography/intel/gcp/crm/folders.py
```diff
@@ -37,68 +58,55 @@ def load_gcp_folders(
...
     gcp_update_tag: int,
     common_job_parameters: Dict,
-) -> None:
+    org_resource_name: str,
+    defer_cleanup: bool = False,
+) -> List[Dict]:
     """
     Get GCP folder data using the CRM v2 resource object, load the data to Neo4j, and clean up stale nodes.
+    :param org_resource_name: Full organization resource name (e.g., "organizations/123456789012")
+    :param defer_cleanup: If True, skip the cleanup job. Used for hierarchical cleanup scenarios.
```
achantavy, line 102: Maybe add a bit more color here to this description. When do we want this to be set true, when do we want this to be set false.

cartography/intel/gcp/crm/orgs.py
```diff
@@ -3,30 +3,38 @@
...
+        for org in client.search_organizations():
+            orgs.append(
+                {
+                    "name": org.name,
+                    "displayName": org.display_name,
+                    "lifecycleState": org.state.name,
+                }
+            )
+        return orgs
+    except Exception as e:
```
achantavy, line 35: I wonder if this module is mature enough to get rid of the try+except clause so that we can crash loudly instead of burying and continuing (potentially with cleanup jobs deleting old data). Wdyt?

cartography/intel/gcp/crm/projects.py
```diff
@@ -1,144 +1,109 @@
...
+                results.append(
+                    {
+                        "projectId": getattr(proj, "project_id", None),
+                        "projectNumber": project_number,
+                        "name": getattr(proj, "display_name", None),
+                        "lifecycleState": proj.state.name,
+                        "parent": project_parent,
+                    }
+                )
+        except Exception as e:
```
achantavy, line 45: same thing about broad exceptions

cartography/intel/gcp/__init__.py
```diff
@@ -179,24 +172,81 @@ def start_gcp_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
...
+        # Store the full resource name for cleanup operations
+        common_job_parameters["ORG_RESOURCE_NAME"] = org_resource_name
+
+        # Sync folders under org
+        folders = sync_gcp_folders(
+            neo4j_session,
+            config.update_tag,
+            common_job_parameters,
+            org_resource_name,
+            defer_cleanup=True,
```
achantavy, line 213: How about we make `sync_gcp_folders()` and `sync_gcp_projects()`not do cleanup at all? This way we don't need to add in the extra parameter and keep things to one codepath.

If the intent is to give users who import cartography as a library a choice, it seems like there's a decent amount of logic happening here in `_sync_project_resources` that they'd be missing out on, so they probably shouldn't be forking the functionality anyway.

cartography/intel/gcp/__init__.py
```diff
@@ -179,24 +172,81 @@ def start_gcp_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
...
+        logger.debug(f"Running cleanup for projects and folders in {org_resource_name}")
+        GraphJob.from_node_schema(GCPProjectSchema(), common_job_parameters).run(
+            neo4j_session
+        )
+        GraphJob.from_node_schema(GCPFolderSchema(), common_job_parameters).run(
+            neo4j_session
+        )
+
+        # Save org cleanup job for later
+        org_cleanup_jobs.append((GCPOrganizationSchema, dict(common_job_parameters)))
```
achantavy, line 241: oh wow, multi-org

achantavy (approved):
Before merging, I think you can delete
- https://github.com/cartography-cncf/cartography/blob/master/cartography/data/jobs/cleanup/gcp_crm_folder_cleanup.json
- https://github.com/cartography-cncf/cartography/blob/master/cartography/data/jobs/cleanup/gcp_crm_organization_cleanup.json
- https://github.com/cartography-cncf/cartography/blob/master/cartography/data/jobs/cleanup/gcp_crm_project_cleanup.json


Left some non-blocking thoughts on exception handling as well as on removing the `defer` flag altogether, lemme know what you think.

cartography/intel/gcp/crm/orgs.py
kunaals: I think so - better to fail loudly and surgically retry or fix bugs

cartography/intel/gcp/__init__.py
kunaals: haha yeah technically we supported multi org before as well

cartography/intel/gcp/__init__.py
kunaals: Good call  - lets not do cleanup at all in those methods

cartography/intel/gcp/crm/projects.py
kunaals: will remove

cartography/intel/gcp/crm/folders.py
kunaals: will remove defer_cleanup

cartography/intel/gcp/crm/folders.py
kunaals: added

PR #1922, opened 2025-09-18
fix(github): isolate contributor lists per repository

byarr:
### Summary

GitHub collaborators accumulate between repos because we share a list of collaborators between runs of fetch repos.

This PR stops sharing data between repos and adds a test for the issue.

- https://cloud-native.slack.com/archives/C080M2LRLDA/p1758092875954949

### Checklist


tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -390,6 +394,79 @@ def test_python_library_in_multiple_requirements_files(neo4j_session):
...
+        ]
+        return PaginatedGraphqlData(nodes=nodes, edges=edges)
+
+    mock_repo_collaborators.side_effect = collaborators_side_effect
+
+    cartography.intel.github.repos.sync(
+        neo4j_session, TEST_JOB_PARAMS, "some_key", "http://localhost", "simpsoncorp"
+    )
+
+    # transform takes 3 args
```
achantavy: Instead of testing the call_args, can we test outcomes instead? 

Specifically, this test should let us reproduce the bug you observed. We should be able to take the cartography master branch, and apply just this test without any other changes, run it, and then see that contributors are attached to things they should not be attached to.

achantavy (changes requested):
Thanks for noticing this and taking the initiative to fix it! Left a comment about updating the test so that it reproduces the issue you saw. Let me know if you have questions

PR #1923, opened 2025-09-18
(feat) Add GitHub organization emails to Users

byarr:
### Summary

Added organization verified domain emails to github user node.

For our organisation, many users use their existing personal github accounts (and usually hide their email). However, as they use SAML for SSO we do have access to their verified domain email address.

Points for discussion/improvement:
- list/single value - When I did this originally, i took the first email address as a single property. This is "cleaner" but doesnt accurate reflect github. Not sure when a single "login" would have multiple emails.
- Pick a representation for empty/missing - i.e. make sure we use None for [] or the other way around?
- missing email - i was also tempted to when the User doesnt have a visible email, but does have an organisation verified one, use that?
- In an ontology world we would probably do something different?

### Checklist


achantavy:
> Points for discussion/improvement:

> list/single value - When I did this originally, i took the first email address as a single property. This is "cleaner" but doesnt accurate reflect github. Not sure when a single "login" would have multiple emails.
> Pick a representation for empty/missing - i.e. make sure we use None for [] or the other way around?
> missing email - i was also tempted to when the User doesnt have a visible email, but does have an organisation verified one, use that?
> In an ontology world we would probably do something different?


I think we could put the org emails as a list on the node. If there are no emails, then we wouldn't set the property at all. The data model _should_ handle writing the data for us, but yes, we would have some GitHubUser nodes where their org_emails property will either be non-existent, an empty list, or a list with 1 or more items. I think this is fine because most node properties in cartography behave like this already (for better or worse).

We could also use the `one_to_many` flag ([ref1](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#one-to-many-relationships), [ref2](https://github.com/cartography-cncf/cartography/blob/master/AGENTS.md#define-one-to-many-relationship)) to define an association between GitHub users and potentially multiple other EntraUsers (since lots of orgs use both GitHub and Entra). Could also do it between GitHub user and multiple other OktaUser. This can all come later though.

But yeah, these are the directions I'd go with. If you'd like I can carry on and take a stab here, but otherwise let me know if this makes sense to you and I'll let you finish up :)

byarr:
> I think we could put the org emails as a list on the node. If there are no emails, then we wouldn't set the property at all. The data model _should_ handle writing the data for us, but yes, we would have some GitHubUser nodes where their org_emails property will either be non-existent, an empty list, or a list with 1 or more items. I think this is fine because most node properties in cartography behave like this already (for better or worse).

Ok, in which case I think this PR does that already..

PR #1928, opened 2025-09-18
chore(gcp): Migrate GKE to data model

kunaals:
### Summary
Migrate GKE to data model

- https://github.com/cartography-cncf/cartography/issues/1723

### Checklist


cartography/models/gcp/gke.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+
+@dataclass(frozen=True)
+class GCPGKEClusterSchema(CartographyNodeSchema):
+    label: str = "GKECluster"
+    properties: GCPGKEClusterNodeProperties = GCPGKEClusterNodeProperties()
+    sub_resource_relationship: GCPGKEClusterToProjectRel = GCPGKEClusterToProjectRel()
+    # Placeholders for future relationships (e.g., to networks, subnets, regions)
+    other_relationships: OtherRelationships = OtherRelationships([])
```
achantavy, line 77: Nit: I think the data model handles this correctly but the more correct way would be to leave this as None or to omit altogether

achantavy (approved):
Also delete https://github.com/cartography-cncf/cartography/blob/master/cartography/data/jobs/cleanup/gcp_gke_cluster_cleanup.json before merging

cartography/models/gcp/gke.py
kunaals: ai slop my bad

PR #1931, opened 2025-09-19
refactor: use transactional writes in AWS, GSuite, Okta loaders

kunaals:
## Summary
- last week, we introduced a change that introduced transactions to the GitHub loader and eliminated a bunch of transient errors: https://github.com/cartography-cncf/cartography/pull/1927
- this PR brings that same goodness to AWS, GSuite, and Okta, the last users of non-transactional loads in the graph. we add a reusable `run_write_query` helper that wraps `neo4j.Session.execute_write`

## Testing
- unit / integration tests

------
https://chatgpt.com/codex/tasks/task_b_68cdbeb6b28c8323bc1827d64e81072f

### DIFF
--- a/cartography/client/core/tx.py
+++ b/cartography/client/core/tx.py
@@ -19,7 +19,9 @@ from cartography.util import batch
 logger = logging.getLogger(__name__)
 
 
-def run_write_query(neo4j_session: neo4j.Session, query: str, **parameters: Any) -> None:
+def run_write_query(
+    neo4j_session: neo4j.Session, query: str, **parameters: Any
+) -> None:
     """Execute a write query inside a managed transaction."""
 
     def _run_query_tx(tx: neo4j.Transaction) -> None:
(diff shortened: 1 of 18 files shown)


cartography/client/core/tx.py
```diff
@@ -19,6 +19,17 @@
 logger = logging.getLogger(__name__)
 
 
+def run_write_query(
+    neo4j_session: neo4j.Session, query: str, **parameters: Any
+) -> None:
+    """Execute a write query inside a managed transaction."""
+
+    def _run_query_tx(tx: neo4j.Transaction) -> None:
+        tx.run(query, **parameters).consume()
```
achantavy, line 28: I'm not 100% if an explicit consume is necessary anymore. Will double check. Not a blocker though since it's easy to remove in a follow-up.

achantavy (approved):

PR #1935, opened 2025-09-22
Add AWS Identity Center group ingestion

kunaals:
## Summary
- add an AWSSSOGroup schema and loader to model Identity Center groups
- extend the Identity Center sync job to ingest groups and their role assignments
- document the new group node type and expand integration fixtures/tests

## Testing
- pytest tests/integration/cartography/intel/aws/test_identitycenter.py *(fails: neo4j.exceptions.ServiceUnavailable - no Neo4j instance listening on localhost:7687)*

------
https://chatgpt.com/codex/tasks/task_b_68d16e433e648323aab3cc5915e45f89


cartography/models/aws/identitycenter/awsssouser.py
```diff
@@ -14,7 +14,7 @@
 
 @dataclass(frozen=True)
 class SSOUserProperties(CartographyNodeProperties):
-    id: PropertyRef = PropertyRef("UserId", extra_index=True)
+    id: PropertyRef = PropertyRef("UserId")
```
kunaals, line 17: id has an index by defalt

cartography/models/aws/identitycenter/awssogroup.py
```diff
@@ -0,0 +1,70 @@
...
+    display_name: PropertyRef = PropertyRef("DisplayName")
+    description: PropertyRef = PropertyRef("Description")
+    identity_store_id: PropertyRef = PropertyRef("IdentityStoreId")
+    external_id: PropertyRef = PropertyRef("ExternalId", extra_index=True)
+    region: PropertyRef = PropertyRef("Region")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class AWSSSOGroupToAWSAccountRelRelProperties(CartographyRelProperties):
```
achantavy: Nit: "RelRel"

cartography/intel/aws/identitycenter.py
```diff
@@ -351,7 +507,31 @@ def sync_identity_center_instances(
             )
 
             users = get_sso_users(boto3_session, identity_store_id, region)
-            transformed_users = transform_sso_users(users)
+            groups = get_sso_groups(boto3_session, identity_store_id, region)
```
achantavy, line 519: right now we're interleaving the get+transform+load calls between groups and users. let's do it in this order:

1. get+transform+load for groups first (since users define an `OtherRelationships` to groups, we should load the groups first)
2. and then do get+transform+load for users

this way nothing is interleaved

tests/integration/cartography/intel/aws/test_identitycenter.py
```diff
@@ -1,9 +1,15 @@
 import tests.data.aws.identitycenter
```
achantavy, line 1: All AWSSOGroup rels are covered by automated tests except for

```
(1)    (AWSSSOGroup)<-[ALLOWED_BY]-(AWSRole)
(2)    (AWSSSOGroup)<-[RESOURCE]-(AWSAccount)
```

You've already provided screenshot proof of it working so I'm not worried here but it'd be good to have an automated test.

I also realize there isn't a `sync_*()` test to begin with making this annoying. As a quick and dirty (but useful) way to address this, we could add in `MERGE (:AWSRole)` in the `test_load_group` test and check that the role got associated. Of these rels tests I'm more interested in (1) than (2) since the sub_resource stuff mostly takes care of (2)

achantavy (changes requested):

cartography/models/aws/identitycenter/awssogroup.py
kunaals: thanks good catch

tests/integration/cartography/intel/aws/test_identitycenter.py
kunaals: good call will add

cartography/intel/aws/identitycenter.py
kunaals: done

PR #1938, opened 2025-09-23
feat: cartography-rules

achantavy:
## Summary

This PR introduces a new rules framework for Cartography that enables security teams to evaluate their infrastructure against common security frameworks like MITRE ATT&CK.

### Text mode:
rules-text-output

### JSON mode:
rules-json-output

### Why

Cartography gives a good map of the environment but users often don’t know what to look for first. In this project we will provide a way for Cartography users to quickly find security issues across multiple platforms and providers. We also want to enable Cartography community members to view and contribute their own rulesets.

### Approach
- Start with well-known best practices frameworks like MITRE ATT&CK.
- Present a text summary to human users, and allow for a json output to let machines read the results.

### Data model
- A **Fact** is a fact about your technical environment e.g. your internet-open EC2 instances, your internet-open storage accounts. It consists of a `cypher_query` that returns that information in table form and `cypher_visual_query` that can be used to return a representation of data in a visual form to the graph UI.
- A **Requirement** is a one or more facts put together toward a security goal. For example MITRE ATT&CK has a requirement called ["Exploit Public Facing Application"](https://attack.mitre.org/techniques/T1190/). We would represent this by having one or more Facts related to public-facing assets.
- A **Framework** like MITRE ATT&CK consists of one or more Requirements.

A Fact is embedded in a single Requirement. This is because we want to avoid the situation where updating a single Fact is impossible because it is referenced by dozens of Requirements that need that Fact to audit slightly different things.

### Choices

We explain this in more detail in the rules README, but to quickly summarize:

- For each framework technique, we ask: "What does an attacker actually need to execute this?"
- Each framework is written to cover multiple clouds/SaaS. Attackers don't necessarily separate attack opportunities by what cloud you're using, so we shouldn't either.
- Our Facts only return items that we believe can present real, immediate risk to an environment
- We do not include any expectation of thresholds because each organization has very different requirements.

### User experience

```bash
# See what's available
cartography-rules list

# Run specific framework  
cartography-rules run mitre-attack

# Run all frameworks
cartography-rules run all

# json output
cartography-rules run all --output json
```

### Things to come later

- Risk acceptance workflow
- More rules packs
- A community-curated rules-pack containing items that showcase Cartography’s strengths like lateral movement in identity and networking

### Data-flow summary

**Input**: Framework contains Requirements, each with multiple Facts that define Cypher queries and metadata.

**Execution**: For each Fact, the system runs cypher_query against Neo4j and flattens the nested objects into FactResults containing the findings plus extracted fields (fact_id, requirement_id, etc.).

**Output**: FactResults are aggregated into a FrameworkResult with summary statistics (total_findings, total_facts) and converted to JSON or formatted text output.

### Checklist

- [x] Include a screenshot

### DIFF
--- a/cartography/rules/spec/result.py
+++ b/cartography/rules/spec/result.py
@@ -34,7 +34,7 @@ class RequirementResult:
 @dataclass
 class FrameworkResult:
     """
-    The formal object output by `--output json` from the `cartography-runrules` CLI.
+    The formal object output by `--output json` from the `cartography-rules` CLI.
     """
 
     framework_id: str


cartography/rules/data/facts/aws/s3/buckets.py
```diff
@@ -0,0 +1,36 @@
...
+    id="aws_s3_public",
+    name="Internet-Accessible S3 Storage Attack Surface",
+    description=("AWS S3 buckets accessible from the internet"),
+    cypher_query="""
+    MATCH (b:S3Bucket)
+    WHERE b.anonymous_access = true
+       OR (b.anonymous_actions IS NOT NULL AND size(b.anonymous_actions) > 0)
+       OR EXISTS {
+         MATCH (b)-[:POLICY_STATEMENT]->(stmt:S3PolicyStatement)
+         WHERE stmt.effect = 'Allow'
```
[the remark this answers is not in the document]
achantavy: we will support this eventually in cartography

cartography/rules/data/facts/aws/ec2/instances.py
```diff
@@ -0,0 +1,23 @@
...
+    ),
+    cypher_query="""
+    MATCH (a:AWSAccount)--(ec2:EC2Instance)-[:MEMBER_OF_EC2_SECURITY_GROUP]->(sg:EC2SecurityGroup)<-[:MEMBER_OF_EC2_SECURITY_GROUP]-(rule:IpPermissionInbound)
+    MATCH (rule)<-[:MEMBER_OF_IP_RULE]-(ip:IpRange{range:'0.0.0.0/0'})
+    WHERE rule.fromport IN [22, 3389, 3306, 5432, 6379, 9200, 27017]
+    RETURN a.name AS account, ec2.instanceid AS instance, rule.fromport AS port, sg.groupid AS sg order by account, instance, port, sg
+    """,
+    cypher_visual_query="""
+    MATCH p=(a:AWSAccount)--(ec2:EC2Instance)-[:MEMBER_OF_EC2_SECURITY_GROUP]->(sg:EC2SecurityGroup)<-[:MEMBER_OF_EC2_SECURITY_GROUP]-(rule:IpPermissionInbound)
+    MATCH p2=(rule)<-[:MEMBER_OF_IP_RULE]-(ip:IpRange{range:'0.0.0.0/0'})
```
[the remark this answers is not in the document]
achantavy: can update later. most envs don't have ipv6

cartography/rules/data/facts/aws/ec2/instances.py
```diff
@@ -0,0 +1,23 @@
...
+
+aws_ec2_instance_internet_exposed = Fact(
+    id="aws_ec2_instance_internet_exposed",
+    name="Internet-Exposed EC2 Instances on Common Management Ports",
+    description=(
+        "EC2 instances exposed to the internet on ports 22, 3389, 3306, 5432, 6379, 9200, 27017"
+    ),
+    cypher_query="""
+    MATCH (a:AWSAccount)--(ec2:EC2Instance)-[:MEMBER_OF_EC2_SECURITY_GROUP]->(sg:EC2SecurityGroup)<-[:MEMBER_OF_EC2_SECURITY_GROUP]-(rule:IpPermissionInbound)
+    MATCH (rule)<-[:MEMBER_OF_IP_RULE]-(ip:IpRange{range:'0.0.0.0/0'})
```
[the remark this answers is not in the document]
achantavy: can update later. most envs don't have ipv6

docs/root/usage/rules.md
```diff
@@ -0,0 +1,79 @@
...
+- If a query returns no results, you’ve successfully eliminated obvious attack opportunities for that technique.
+- If it does return findings, you now have a clear list of potential targets for attackers.
+
+
+## Why do it this way
+
+Security teams already face too many competing priorities. Our goal is to highlight only the facts that represent meaningful risks.
+
+For example:
+An EC2 security group that allows inbound internet traffic is only relevant if there are compute instances actually attached to it. Otherwise, it’s noise.
```
kunaals: Do we have a stronger example we could put here?

docs/root/usage/rules.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Setup
+
+```bash
+export NEO4J_URI=bolt://localhost:7687 # or your Neo4j URI
+export NEO4J_USER=neo4j # or your username
+export NEO4J_DATABASE=neo4j
+
+# Store the Neo4j password in an environment variable. You can name this anything you want.
+```bash
```
kunaals: Either have closing backtickets under NEO4J_DATABASE or remove this new ```bash

cartography/rules/data/frameworks/mitre_attack/requirements.py
```diff
@@ -0,0 +1,36 @@
+"""
```
kunaals: I think that most frameworks will require custom queries in their associated Facts. To avoid a combinatorial explosion of very similar-looking Facts in the community_facts dir, it would be good to include Framework specific Facts within the files that include the requirement. This will make requirements files (oof Python collision) quite long. As such, it would be good to put each Requirement in it's own file. The __init__ for the given framework can be untouched.

cartography/rules/runners.py
```diff
@@ -0,0 +1,264 @@
...
+    total_facts: int,
+    total_findings: int,
+):
+    """Format and output the results of framework execution."""
+    if output_format == "json":
+        if len(framework_names) == 1:
+            print(json.dumps(asdict(all_results[0]), indent=2))
+        else:
+            combined_output = [asdict(result) for result in all_results]
+            print(json.dumps(combined_output, indent=2))
```
kunaals:
```suggestion
        combined_output = [asdict(result) for result in all_results]
        print(json.dumps(combined_output, indent=2))
```
If someone is using the json output it's likely they're using it for automation. Keeping things array makes the contract more stable:

```
  # With array (consistent):
  cartography-runrules mitre-attack --output json | jq '.[0].total_findings'
  cartography-runrules all --output json | jq '.[].total_findings | add'

  # With mixed format (inconsistent):
  cartography-runrules mitre-attack --output json | jq '.total_findings'     # object
  cartography-runrules all --output json | jq '.[].total_findings | add'     # array
```

cartography/rules/spec/model.py
```diff
@@ -0,0 +1,77 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any
+
+
+class Provider(str, Enum):
```
kunaals: I think for the sake of consistency, we should call this "Module" given that's what we've called various cartography integrations since inception

cartography/rules/spec/result.py
```diff
@@ -0,0 +1,30 @@
+# Execution result classes
```
kunaals, line 1: I think it would be better to call this file "result.py"

cartography/rules/runners.py
```diff
@@ -0,0 +1,264 @@
...
+    if output_format == "text":
+        print(f"Connecting to Neo4j at {uri}...")
+    driver = GraphDatabase.driver(uri, auth=(neo4j_user, neo4j_password))
+
+    try:
+        # Test connection
+        with driver.session(database=neo4j_database) as session:
+            result = session.run("RETURN 1 as test")
+            if result.single()["test"] != 1:
+                raise Exception("Connection test failed")
```
kunaals: There's a utility for this: `driver.verify_connectivity()`
https://neo4j.com/docs/python-manual/current/connect/

cartography/rules/runners.py
```diff
@@ -0,0 +1,264 @@
...
+        # Generate and display clickable Neo4j Browser URL
+        browser_url = _generate_neo4j_browser_url(neo4j_uri, fact.cypher_visual_query)
+        print(
+            f"  \033[36m{'Neo4j Query:':<12}\033[0m \033]8;;{browser_url}\033\\Click to run visual query\033]8;;\033\\"
+        )
+
+    with driver.session(database=database) as session:
+        result = session.run(fact.cypher_query)
+        findings = [dict(record) for record in result]
+        finding_count = len(findings)
```
kunaals, line 58: Could we use `Driver.execute_query()` for everything in here the rules module? Would be good to have transactions on everything

cartography/rules/spec/result.py
```diff
@@ -0,0 +1,30 @@
...
+class FactResult:
+    fact_id: str
+    fact_name: str
+    fact_description: str
+    fact_provider: str
+    requirement_id: str
+    finding_count: int = 0
+    findings: list[dict[str, Any]] | None = None
+    requirement_url: str | None = None
+
```
kunaals, line 32: Should we also have a Requirements Result? This way in the output we can cleanly go from Framework -> Requirement -> Fact and if a Requirement within a Framework is all green we can just minimize those results and put a big green check mark or something. Keeps the data flow consistent throughout

kunaals: Would require some slight changes to the runners I suppose

cartography/rules/cli.py
```diff
@@ -0,0 +1,166 @@
+"""
```
kunaals, line 1: It could be nice to use Typer (which is built on Click and Rich) to make this CLI more friendly. In the Claude Code era I think pretty CLIs are becoming a requirement 😄 https://typer.tiangolo.com/

I think this is not as important for the standard and drift detect CLIs because those are more config and data heavy, this CLI actually has a presentation layer

cartography/rules/formatters.py
```diff
@@ -0,0 +1,42 @@
+"""
+Output formatting utilities for Cartography rules.
+"""
+
+import re
+from urllib.parse import quote
+
+
+def _generate_neo4j_browser_url(neo4j_uri: str, cypher_query: str) -> str:
```
kunaals, line 9: There is also an ssc scheme for bolt and neo4j: https://neo4j.com/docs/python-manual/current/connect-advanced/

cartography/rules/spec/model.py
```diff
@@ -0,0 +1,77 @@
...
+
+    OKTA = "Okta"
+    """Okta identity and access management"""
+
+    CROSS_CLOUD = "CROSS_CLOUD"
+    """Multi-cloud or provider-agnostic rules"""
+
+
+@dataclass(frozen=True)
+class Fact:
```
kunaals, line 29: I like the Requirement `attributes` tag - I think Fact should also have an extensible field. Perhaps we could add `attributes` or `tags` to Fact as well?

cartography/rules/spec/model.py
```diff
@@ -0,0 +1,77 @@
...
+    Notes:
+    - `attributes` is reserved for metadata such as tags, categories, or references.
+    - Do NOT put evaluation logic, thresholds, or org-specific preferences here.
+    """
+
+    id: str
+    name: str
+    description: str
+    facts: tuple[Fact, ...]
+    attributes: dict[str, Any] | None = None
```
kunaals, line 64: I like the example you give in the Requirements in the facts directory, could we leave a lightweight comment here with examples of attributes, like 

```
  attributes = {
      "tactic": "initial_access",
      "technique_id": "T1190",
      "random_field": "whatever"
  }
```

cartography/rules/cli.py
```diff
@@ -0,0 +1,166 @@
...
+            nargs="?",
+            choices=available_frameworks,
+            help='Security framework to execute (or "all" to execute all frameworks)',
+        )
+
+        parser.add_argument(
+            "--list", action="store_true", help="List available frameworks and exit"
+        )
+
+        parser.add_argument(
```
kunaals: It would be nice to eventually store these common arguments between CLIs in one place (like uri, username, password, etc). If the CLI args or the env vars got out of sync it would hinder usability. For example Neo4j tends to use the word "username" and as such Langgraph and others use the env var `NEO4J_USERNAME` whereas we use `NEO4J_USER`, it would be unfortunate for us to somehow use both USER and USERNAME across different CLIs

docs/root/usage/rules.md
```diff
@@ -0,0 +1,79 @@
+# Cartography Rules
+
+You can use the `cartography-runrules` command to evaluate your environment against security frameworks.
+
+## What are rules in Cartography?
+
+Cartography ingests data from your environment and stores it in a knowledge graph. With `cartography-runrules`, we provide a set of pre-defined queries that let you evaluate your environment against common security frameworks, along with community-curated rules.
```
kunaals:
```suggestion
With `cartography-runrules`, we provide a set of pre-defined queries that let you evaluate your environment against common security frameworks, along with community-curated rules.
```

docs/root/usage/rules.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+## Our opinionated approach
+
+These rules are designed from an attacker’s perspective. For each framework technique, we ask: "What does an attacker actually need to execute this technique?"
+
+Our queries surface opportunities that enable attacks at all stages of the kill chain—initial access, lateral movement, data exfiltration, and beyond.
```
kunaals:
```suggestion
Our queries surface opportunities that enable attacks at all stages of the kill chain: initial access, lateral movement, data exfiltration, and beyond.
```

docs/root/usage/rules.md
```diff
@@ -0,0 +1,79 @@
...
+set +o history # avoid storing the password in the shell history; can also use something like 1password CLI.
+export NEO4J_PASSWORD=password
+set -o history # turn shell history back on
+```
+
+## Usage
+
+See available frameworks and facts
+```bash
+cartography-runrules list
```
kunaals: I think ideally this would be composable. Something like

`cartography-runrules list` -> every framework, requirement, fact
`cartography-runrules list mitre-attack` -> every requirement, fact under mitre
`cartography-runrules list mitre-attack <requirement>` every fact under mitre-attack <requirement>

What do you think?

docs/root/usage/rules.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+Run a specific framework
+```bash
+cartography-runrules mitre-attack
```
kunaals: Ditto here, would be good to be able to specify requirement after the framework. Something like 
`cartography-runrules mitre-attack <requiement>`

cartography/rules/spec/model.py
achantavy: I'm thinking we leave off the attribute from the fact for now because I want to avoid duplicate data between Requirements and Facts. Right now I think it makes more sense for Requirements than Facts to have attributes. We can always add them to Facts later on.

achantavy:
> Oh one more thing, we should call the `facts` dir `community_facts` and store Facts associated with a Framework within the Requirement file itself. `community_facts` will be Facts that arent necessarily associated with a Framework.
> 
> We'll need to update the CLI to run particular community facts

Maybe we can create a framework just called "cartography-community" or something and put those there. Can decide in follow-up PR where we add more rules.

docs/root/usage/rules.md
achantavy: Overhauled the doc so it flows better

cartography/rules/cli.py
achantavy: Yeah, will do once we move the main CLI over to typer

cartography/rules/data/frameworks/mitre_attack/requirements/t1190_exploit_public_facing_application/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+MITRE ATT&CK Framework
+Security framework based on MITRE ATT&CK tactics and techniques
+"""
+
+from cartography.rules.spec.model import Fact
+from cartography.rules.spec.model import Module
+from cartography.rules.spec.model import Requirement
+
+# AWS
+_aws_ec2_instance_internet_exposed = Fact(
```
achantavy, line 11: using `_` as a prefix convention here to indicate that these queries are not meant to be imported into other files

kunaals (approved):

PR #1941, opened 2025-09-24
feat(entra): Add manager reporting relationship for Entra users

achantavy:
- add a REPORTS_TO relationship on Entra users that links each user to their manager via manager_id
- document the new relationship in the Entra schema docs
- extend the Entra user integration test to verify the reporting edge


docs/root/modules/entra/schema.md
```diff
@@ -73,7 +73,7 @@ Representation of an Entra [User](https://learn.microsoft.com/en-us/graph/api/us
 - Entra users can have a manager
 
     ```cypher
-    (:EntraUser)-[:MANAGES]->(:EntraUser)
```
achantavy, line 76: This never existed before and probably was AI slop

Searching for "MANAGES" gets me: 
Image

Issue #1736, opened 2025-07-29
feat(azure): improve Azure coverage (Application Gateway)


achantavy:
*Description*:
> Describe your idea.  Please be detailed.  If a feature request, please
describe the desired behavior, what scenario it enables, and how it
would be used.

Cartography should add Azure support for the following modules:

* [ ] Networking
  * [x] Virtual networks
  * [x] Subnets
  * [x] Network Security Groups
  * [x] Load Balancers
  * [ ] Application gateways
* [x] Container services
  * [x] Azure Kubernetes Service
  * [x] Container Instances
* [x] Serverless
  * [x] Azure Functions
  * [x] App Service
  * [x] Logic Apps
  * [x] Event Grid
* [x] Key Vault
* [x] Monitoring
  * [x] Azure Monitor
  * [x] Azure Security Center
* [x] Resource groups
* [x] Storage and data
  * [x] Data Lake
  * [x] Synapse
  * [x] Data Factory
  * [x] Event Hub

----- Edit @jychp -----
### Old related PR

> You can take inspiration from the following PRs (outdated but still useful). Feel free to link your PR to the older ones to help close any that are no longer needed.

- https://github.com/cartography-cncf/cartography/pull/1076
- https://github.com/cartography-cncf/cartography/pull/1075
- https://github.com/cartography-cncf/cartography/pull/1074
- https://github.com/cartography-cncf/cartography/pull/836
- https://github.com/cartography-cncf/cartography/pull/764

Ytemiloluwa:
[...]

PR #1947, opened 2025-09-27
feat(azure): add support for Event Grid Topics

janithashri:
### Summary

This pull request introduces a new intel module to ingest Azure Event Grid Topics. This change adds a new `:AzureEventGridTopic` node and connects it to the existing `:AzureSubscription` node with a `:RESOURCE` relationship.

This implements the Event Grid Topics. Ingestion for Event Subscriptions will be handled in a follow-up pull request.

### Related issues or links

- Addresses part of #1736

### Checklist

Screenshot 2025-09-27 200649

Screenshot 2025-09-27 201427

### DIFF
--- a/cartography/intel/azure/event_grid.py
+++ b/cartography/intel/azure/event_grid.py
@@ -2,8 +2,6 @@ import logging
 from typing import Any
 
 import neo4j
-from azure.core.exceptions import ClientAuthenticationError
-from azure.core.exceptions import HttpResponseError
 from azure.mgmt.eventgrid import EventGridManagementClient
 
 from cartography.client.core.tx import load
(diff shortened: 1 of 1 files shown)


cartography/intel/azure/event_grid.py
```diff
@@ -0,0 +1,103 @@
...
+
+@timeit
+def get_event_grid_topics(credentials: Credentials, subscription_id: str) -> list[dict]:
+    """
+    Get a list of Event Grid Topics from the given Azure subscription.
+    """
+    try:
+        client = EventGridManagementClient(credentials.credential, subscription_id)
+        return [topic.as_dict() for topic in client.topics.list_by_subscription()]
+    except (ClientAuthenticationError, HttpResponseError) as e:
```
achantavy: Won't block on this since our treatment of error handling in azure is mostly like this, but let's have a follow-up PR fix the error handling strategy:

- if the error is due to a flaky API (a transient problem), then we log a warning and continue. In some cases we can choose to retry.
- If the error is unlikely to be resolved by a retry, we allow the code to fail loudly by not handling the exception at all.

cartography/intel/azure/event_grid.py
```diff
@@ -0,0 +1,103 @@
...
+        logger.warning(
+            f"Failed to get Event Grid topics for subscription {subscription_id}: {str(e)}"
+        )
+        return []
+
+
+@timeit
```
achantavy: No need timeit for a transform

d-aggarwal:
@janithashri are you able to follow up here?

PR #1968, opened 2025-10-04
feat(azure): add support for Data Lake file systems

janithashri:
### Summary

This pull request introduces a new intel module to ingest Azure Data Lake Gen2 file systems. This change adds a new `:AzureDataLakeFileSystem` node and connects it to the existing `:AzureStorageAccount` node with a `:CONTAINS` relationship.

The intel logic first identifies storage accounts with hierarchical namespace enabled and then ingests their blob containers as file systems. The implementation follows the project's modern, schema-based pattern and includes integration tests and schema documentation.

### Related issues or links

- Addresses part of #1736

### Checklist

Screenshot 2025-10-04 130436

### DIFF
--- a/cartography/intel/azure/data_lake.py
+++ b/cartography/intel/azure/data_lake.py
@@ -1,0 +1,106 @@
+import logging
+from typing import Any
+
+import neo4j
+from azure.core.exceptions import ClientAuthenticationError, HttpResponseError
+from azure.mgmt.storage import StorageManagementClient
+
+from cartography.client.core.tx import load
+from cartography.graph.job import GraphJob
+from cartography.models.azure.data_lake_filesystem import AzureDataLakeFileSystemSchema
+from cartography.util import timeit
+
+from .util.credentials import Credentials
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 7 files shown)


cartography/intel/azure/data_lake.py
```diff
@@ -0,0 +1,125 @@
...
+        logger.warning(f"Failed to get Storage Accounts for Data Lake sync: {str(e)}")
+        return []
+
+
+@timeit
+def get_filesystems_for_account(
+    client: StorageManagementClient,
+    account: dict,
+) -> list[dict]:
+    resource_group_name = _get_resource_group_from_id(account["id"])
```
achantavy, line 44: [question] If there is no resource group name, then there is no filesystem?

In general, let's try to be thoughtful with the error handling (though with other Azure modules prior to you working on this it was handled inconsistently). I commented on this in the other PR but let's try to make it so that transient errors are warned on and skipped over and permanent ones (like if the network is hard down or if the user forgot to configure their credential) fail loudly.

When in doubt, code for the happy path and don't prematurely optimize by adding a bunch of conditionals and nested logic for code paths and situations that we haven't seen. Keep it flat and simple as much as possible

PR #1974, opened 2025-10-06
fix(github, core): Allow overriding load() batch size #1958

achantavy:

### Checklist

### DIFF
--- a/cartography/client/core/tx.py
+++ b/cartography/client/core/tx.py
@@ -14,6 +14,7 @@ from cartography.graph.querybuilder import build_ingestion_query
 from cartography.graph.querybuilder import build_matchlink_query
 from cartography.models.core.nodes import CartographyNodeSchema
 from cartography.models.core.relationships import CartographyRelSchema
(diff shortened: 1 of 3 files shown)


cartography/client/core/tx.py
```diff
@@ -236,11 +241,17 @@ def load_graph_data(
...
     :param dict_list: The data to load to the graph represented as a list of dicts.
+    :param batch_size: Number of rows to include in each message when loading data. If the data in a single message is
+    too large, Neo4j will throw a ConnectionResetError - see issue #1957. Relationship-heavy data should use smaller
+    batch sizes.
     :param kwargs: Allows additional keyword args to be supplied to the Neo4j query.
     :return: None
     """
-    for data_batch in batch(dict_list, size=10000):
-        neo4j_session.write_transaction(
+    if batch_size <= 0:
```
achantavy, line 250: Since we're exposing batch_size as a parameter to the main `load()` interface, I figure it's worth adding a guardrail

achantavy:
this is actually not related, the error thrown is related to the index creation

PR #1980, opened 2025-10-07
feat(rules): start MITRE T1098. add auto-tab for technique

achantavy:
### Summary

- Added support for some MITRE ATT&CK T1098 Account Manipulation rules in AWS
- Added a new `target_assets` field to the Requirement data model to clearly document what specific Cartography entities each security requirement analyzes
- Updates Typer so that we show a description of each technique as we tab through

https://github.com/user-attachments/assets/de72ff0c-b20f-47e4-9b26-f780f367280c

### Checklist

- [ ] Update the schema and readme.

- [ ] Use the NodeSchema data model.

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -30,7 +30,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@05b42c624433fc40578a4040d5cf5e36ddca8cde # v2.4.2
+        uses: ossf/scorecard-action@4eaacf0543bb3f2c246792bd56e8cdeffafb205a # v2.4.3
         with:
           results_file: results.sarif
           results_format: sarif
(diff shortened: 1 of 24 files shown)


cartography/rules/spec/model.py
```diff
@@ -58,9 +58,21 @@ class Requirement:
...
     description: str
+    """A brief description of the requirement."""
+    target_assets: str
+    """
+    A short description of the assets that this requirement is related to. E.g. "Cloud
+    identities that can manipulate other identities". This field is used as
+    documentation: `description` tells us information about the requirement;
+    `target_assets` tells us what specific objects in cartography we will search for to
+    find Facts related to the requirement.
+    """
```
achantavy, line 73: I made this field so that we can more easily stay focused when updating rules.

The description of a requirement is often very high level and we need to translate it to the specific asset being audited. Having a separate "target assets" field helps the rule author stay focused and align reviewers on what things should be queried for.

cartography/rules/data/frameworks/mitre_attack/requirements/t1098_account_manipulation/__init__.py
```diff
@@ -0,0 +1,317 @@
...
+            [p IN ['iam:Create','iam:Attach','iam:Put','iam:Update','iam:Add'] |
+                p] AS patterns
+        WITH a, principal, principal_type, stmt,
+            // Filter on the desired IAM actions
+            [action IN stmt.action
+                WHERE ANY(prefix IN patterns WHERE action STARTS WITH prefix)
+                OR action = 'iam:*'
+                OR action = '*'
+            ] AS matched_actions
+        // Return only statement actions that we matched on
```
achantavy, line 33: In general, the visual queries are much simpler because they don't need to do any text processing. Here, we make sure to return only the actions of the statement that we matched on instead of all the actions of the statement. The visual query doesn't need that logic.

Issue #1732, opened 2025-07-28
Implement GSuite group owners


achantavy:
Implement GSuite group owners

*Description*:
> Describe your idea.  Please be detailed.  If a feature request, please
describe the desired behavior, what scenario it enables, and how it
would be used.

Add `(:GSuiteUser)-[:OWNER_OF]->(:GSuiteGroup)` as a path in the graph.

I'm not sure what GSuite metadata object contains this though.

dikshant182004:
hi @achantavy ,i would like to work on this issue .If there is no existing pr related to this can u plz assign this to me .

PR #1986, opened 2025-10-11
tests: add integ coverage for googleworkspace(gsuite)

achantavy:
Getting ready for a data model refactor on gsuite/google workspace so that we can extend it more easily

Unblock #1732 

### Checklist


kunaals:
We should rename this to Google workspace, it's been a while now. Perhaps we can maintain backwards compat on the node labels for a little while to not break people?

PR #1989, opened 2025-10-13
feat(trivy): Support digest URIs in trivy

kunaals:
### Summary
This PR enables the Trivy intel module to ingest scan results for digest-qualified image URIs (e.g., `repo.example.com/app@sha256:abc123...`) while keeping backward compatibility with tag-based scans

- Introduced `_prepare_trivy_data()` helper that resolves tag URIs and digest URIs by inspecting `ArtifactName`, `Metadata.RepoTags`, and `Metadata.RepoDigests` fields
- Added `_get_scan_targets_and_aliases()` to build a mapping of digest-qualified URIs to tag URIs from ECR images in the graph
- Updated both S3 and local file ingestion paths to process all available scan files and map them to the correct ECR image nodes
- Removed filename-based matching logic (`_get_intersection`) in favor of content-based resolution
- Added test coverage for digest-based file ingestion

**Why this matters:**
- Supports multi-architecture images where each platform-specific digest needs separate vulnerability scanning
- Allows Trivy scanner to use digest references (immutable) instead of tags (mutable) for more reliable scanning
- Enables ingestion of scans for both amd64 and arm64 digests of the same image

### Related issues or links

- https://github.com/cartography-cncf/cartography/issues/1990

### Checklist

### DIFF
--- a/cartography/intel/aws/ec2/elastic_ip_addresses.py
+++ b/cartography/intel/aws/ec2/elastic_ip_addresses.py
@@ -6,6 +6,7 @@ import boto3
 import neo4j
 from botocore.exceptions import ClientError
 
+from cartography.client.core.tx import run_write_query
 from cartography.util import aws_handle_regions
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
@@ -83,7 +84,8 @@ def load_elastic_ip_addresses(
         SET r.lastupdated = $update_tag
     """
 
-    neo4j_session.run(
+    run_write_query(
+        neo4j_session,
         ingest_addresses,
         elastic_ip_addresses=elastic_ip_addresses,
         Region=region,
--- a/cartography/intel/aws/ec2/internet_gateways.py
+++ b/cartography/intel/aws/ec2/internet_gateways.py
@@ -5,6 +5,7 @@ from typing import List
 import boto3
 import neo4j
 
+from cartography.client.core.tx import run_write_query
 from cartography.util import aws_handle_regions
 from cartography.util import run_cleanup_job
 from cartography.util import timeit
@@ -63,13 +64,14 @@ def load_internet_gateways(
         SET r.lastupdated = $aws_update_tag
     """
 
-    neo4j_session.run(
+    run_write_query(
+        neo4j_session,
         query,
         internet_gateways=internet_gateways,
         region=region,
         aws_account_id=current_aws_account_id,
         aws_update_tag=update_tag,
-    ).consume()
+    )
 
 
 @timeit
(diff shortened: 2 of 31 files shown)


cartography/intel/trivy/__init__.py
```diff
@@ -110,18 +149,33 @@ def sync_trivy_aws_ecr_from_s3(
...
             "a folder, it MUST end with a trailing slash '/'. "
         )
-        logger.error(f"JSON files in S3: {json_files}")
         raise ValueError("No ECR images with S3 json scan results found.")
 
-    logger.info(f"Processing {len(intersection)} ECR images with S3 scan results")
-    for image_uri, s3_object_key in intersection:
-        sync_single_image_from_s3(
+    logger.info(f"Processing {len(json_files)} Trivy result files from S3")
+    s3_client = boto3_session.client("s3")
```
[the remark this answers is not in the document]
kunaals: I don't think this applies here because we're pulling S3 objects with fully qualified URLs that include the exact region of the bucket they are coming from

tests/unit/cartography/intel/trivy/test_scanner.py
```diff
@@ -489,3 +424,57 @@ def test_sync_trivy_aws_ecr_from_s3_no_matches(
...
+        "Body": body
+    }
+
+    sync_trivy_aws_ecr_from_s3(
+        neo4j_session=MagicMock(),
+        trivy_s3_bucket="test-bucket",
+        trivy_s3_prefix="trivy-scans/",
+        update_tag=123,
+        common_job_parameters={},
+        boto3_session=mock_boto_session.return_value,
```
achantavy, line 475: [nonblock] Just curious, does the test still pass if we put supply `mock_boto_session` here instead of `mock_boto_session.return_value`?

cartography/intel/trivy/__init__.py
```diff
@@ -11,62 +11,104 @@
...
 
-        if image_uri in image_uris:
-            intersection.append((image_uri, s3_object_key))
+    if isinstance(metadata, dict):
+        for key in ("RepoTags", "RepoDigests"):
+            values = metadata.get(key)
+            if isinstance(values, list):
+                candidates.extend(str(v).strip() for v in values if v)
+            elif isinstance(values, str):
+                candidates.append(values.strip())
```
achantavy: Is it possible to avoid `isinstance` and explicitly get each value?

E.g. can we do 

```
# handle repotags: add that to the candidates list

# now handle repodigests: add those to the candidates list
```

kunaals: good call, updated. these should always be lists according to https://trivy.dev/v0.67/docs/configuration/reporting/#json

tests/unit/cartography/intel/trivy/test_scanner.py
kunaals: That would break things because it wouldn’t have the paginator chain or the canned pages that we set on .return_value

PR #2000, opened 2025-10-14
fix(*): Ensure ingestion queries use managed transactions

kunaals:
## Summary
- problem: at scale we see nodes that dont exist somehow when running neo4j_session.run()
- replace direct neo4j_session.run writes across ingestion modules with run_write_query so writes occur inside managed transactions
- execute read-side queries through execute_read helpers instead of autocommit sessions
- add the necessary transaction helper imports throughout the updated modules

------
https://chatgpt.com/codex/tasks/task_b_68eddce1256483238adf68e653e64742

### DIFF
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -27,7 +27,6 @@ import botocore
 import neo4j
 from botocore.exceptions import EndpointConnectionError
 
-from cartography.client.core.tx import run_write_query
 from cartography.graph.job import GraphJob
 from cartography.graph.statement import get_job_shortname
 from cartography.stats import get_stats_client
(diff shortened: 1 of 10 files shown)


cartography/util.py
```diff
@@ -153,6 +153,9 @@ def merge_module_sync_metadata(
     :param synced_type: The sub-module's type
     :param update_tag: Timestamp used to determine data freshness
     """
+    # Import here to avoid circular import with cartography.client.core.tx
+    from cartography.client.core.tx import run_write_query
```
achantavy, line 157: Is there a way to avoid the import down here?

kunaals: Unfortunately not :(

PR #2005, opened 2025-10-15
fix(ecr): ECR image layers batching

kunaals:
### Summary
Make batch_size configurable in `load`, `load_graph_data`, and `load_matchlinks` for modules with larger amounts of data that could exceed Neo4j's transaction memory limits. Adjust ECR image layers to be 1000 instead of 10000 due to this error in an environment with 100k images.

```
neo4j.exceptions.TransientError: {code: Neo.TransientError.General.MemoryPoolOutOfMemoryError} {message: The allocation of an extra 663.2 MiB would use more than the limit 824.0 MiB. Currently using 731.5 MiB. dbms.memory.transaction.total.max threshold reached}
```

### Checklist

~~- [ ] Update/add unit or integration tests.~~ intentionally didn't change tests to show that this shows
~~- [ ] Include a screenshot showing what the graph looked like before and after your changes.~~ intentionally looks the same 
~~- [ ] Include console log trace showing what happened before and after your changes.~~ not required - tests capture

### DIFF
--- a/cartography/client/core/tx.py
+++ b/cartography/client/core/tx.py
@@ -262,6 +262,8 @@ def load_graph_data(
     :param kwargs: Allows additional keyword args to be supplied to the Neo4j query.
     :return: None
     """
+    if batch_size <= 0:
+        raise ValueError(f"batch_size must be greater than 0, got {batch_size}")
     for data_batch in batch(dict_list, size=batch_size):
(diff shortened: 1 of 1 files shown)


achantavy (approved):
lgtm, though it'd be good to disallow a batch size <= 0 in the lower level functions. stamping for now to unblock

PR #2006, opened 2025-10-15
fix(aws): Rely on aws_handle_regions for SCP deny handling

kunaals:
## Summary
- handle SCP denies in the aws_handle_regions decorator

Inspired by this error message:
```
botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the ListFindings operation: User: arn:aws:sts::123456789111:assumed-role/beepboop/botocore-session-1760504409 is not authorized to perform: inspector2:ListFindings on resource: arn:aws:inspector2:us-east-1:123456789111:/findings/list with an explicit deny in a service control policy
```

## Testing
- Added basic unit test

------
https://chatgpt.com/codex/tasks/task_b_68ef416cd9f88323acbc0e8813c45dea


cartography/intel/aws/inspector.py
```diff
@@ -82,6 +83,10 @@ def get_inspector_findings(
     only fetch those in ACTIVE or SUPPRESSED statuses.
     Run the query in batches of 1000 findings and return an iterator to fetch the results.
     """
+    # Note: We can't use @aws_handle_regions decorator here because this function returns a generator.
```
achantavy, line 88: Making sure I understand: so right now, aws_handle_regions is not doing anything in aws inspector?

cartography/intel/aws/inspector.py
```diff
@@ -300,38 +305,63 @@ def _sync_findings_for_account(
...
+            )
+            load_inspector_finding_to_package_match_links(
+                neo4j_session,
+                finding_to_package_map,
+                update_tag,
+                current_aws_account_id,
+            )
+    except botocore.exceptions.ClientError as e:
+        error_code = e.response.get("Error", {}).get("Code")
+        if error_code in {"AccessDenied", "AccessDeniedException"}:
```
achantavy: It'd be good to cover the same error codes that aws_handle_regions does: https://github.com/cartography-cncf/cartography/blob/eccddcecf9a6e6a058c16663fa1dd22c279c6cee/cartography/util.py#L273C5-L283

cartography/intel/aws/inspector.py
kunaals: Correct

PR #2022, opened 2025-10-27
feat: Spacelift Integration 

shyammukund:
### Checklist

image

### DIFF
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -347,7 +347,6 @@ def aws_handle_regions(func: AWSGetFunc) -> AWSGetFunc:
             logger.warning(
                 "Encountered an EndpointConnectionError. This means that the AWS "
                 "resource is not available in this region. Skipping.",
-                exc_info=True,
             )
             return []
 


cartography/intel/spacelift/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+
+@timeit
+def start_spacelift_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
+    """
+    Perform ingestion of Spacelift data.
+    :param neo4j_session: Neo4j session for database interface
+    :param config: A cartography.config object
+    :return: None
+    """
+    if not config.spacelift_api_token or not config.spacelift_api_endpoint:
```
kunaals, line 29: I believe we had talked about asking the caller to provide the key_id and the key_secret instead of the API key token itself. After thinking about it a bit, i think it's okay to have users exchange the id and secret at the /apikeytoken endpoint and just put in the token here. We can link folks to https://docs.spacelift.io/integrations/api#spacelift-api-key to generate the thing. Or we could accept the ID and secret which would help usability slightly. Up to you here. If you found it kinda dumb and painful that you had to exchange the ID and secret to get another token, then lets abstract that away from users here

cartography/intel/spacelift/runs.py
```diff
@@ -0,0 +1,460 @@
...
+                    }
+                )
+                logger.info(
+                    f"Mapped instance {instance_id} to updater run {updater_run_id}"
+                )
+
+    logger.info(
+        f"Built run-to-instances map with {len(run_to_instances)} runs affecting EC2 instances"
+    )
+    logger.info(f"Run-to-instances map: {run_to_instances}")
```
kunaals:
```suggestion
```
Let's remove this - bit verbose for prod. I think the log we have about length directly above is good enough

cartography/intel/spacelift/runs.py
```diff
@@ -0,0 +1,460 @@
...
+        vendor = entity.get("vendor", {})
+        if vendor.get("__typename") != "EntityVendorTerraform":
+            continue
+
+        terraform_data = vendor.get("terraform", {})
+        if terraform_data.get("__typename") != "TerraformResource":
+            continue
+
+        values_json = terraform_data.get("values")
+        values = json.loads(values_json)
```
kunaals: Let's ensure that values_json is valid json so that it doesn't break the json.loads(). A try catch here would be helpful. Perhaps we could let values_json be as-is, but then add a bit of defense like this:
```python
try:
    values = json.loads(values_json)
    instance_id = values["id"]
except (json.JSONDecodeError, KeyError) as e:
    logger.warning(f"Failed to parse entity values: {e}")
    continue
```

cartography/intel/spacelift/account.py
```diff
@@ -0,0 +1,96 @@
...
+        raise ValueError(
+            f"Could not extract account ID from API endpoint: {api_endpoint}"
+        )
+
+    logger.info(f"Extracted account ID: {account_id}")
+
+    return {
+        "id": account_id,
+        "name": account_id,
+    }
```
kunaals: We could just return account_id here as a string

cartography/intel/spacelift/account.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def transform_account(account_data: dict) -> dict:
+    return {
+        "id": account_data["id"],
+        "account_id": account_data["id"],
+        "name": account_data.get("name"),
+    }
```
kunaals: Let's inline this in the sync() method, its not worth creating the separate method

kunaals: Esp because id, account_id, and name are all effectively just account_data["id"]

cartography/intel/spacelift/account.py
```diff
@@ -0,0 +1,96 @@
...
+
+    load_account(
+        neo4j_session, transformed_account, common_job_parameters["UPDATE_TAG"]
+    )
+
+    cleanup_account(neo4j_session, common_job_parameters)
+
+    account_id = transformed_account["account_id"]
+    logger.info(f"Synced Spacelift account: {account_id}")
+    return account_id
```
kunaals:
```suggestion
    account_id = get_account(api_endpoint)
    transformed_account = {... dict with fields id, account_id, name all equal to account_id var from above line}
    load_account(
        neo4j_session, transformed_account, common_job_parameters["UPDATE_TAG"]
    )
    cleanup_account(neo4j_session, common_job_parameters)
    logger.info(f"Synced Spacelift account: {account_id}")
    return account_id
```

kunaals: No need for the spaces between lines

cartography/models/spacelift/commit.py
```diff
@@ -0,0 +1,103 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class GitCommitNodeProperties(CartographyNodeProperties):
```
kunaals: Having a module called GitCommit that is not actually from GitHub or GitLab is misleading. We should call this SpaceliftGitCommit or something similar and then potentially give it an overlapping label with actual GitCommits one day if we start ingesting commits from an actual source control system

cartography/models/spacelift/commit.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Properties for a Git Commit node.
```
kunaals:
```suggestion
    Properties for a Spacelift GitCommit node.
```

cartography/models/spacelift/run.py
```diff
@@ -0,0 +1,185 @@
...
+    GENERATES relationship from a Run to its parent Stack.
+    (:SpaceliftRun)<-[:GENERATES]-(:SpaceliftStack)
+    """
+
+    target_node_label: str = "SpaceliftStack"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("stack_id")},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "GENERATES"
```
kunaals: I think we talked about this in the spec but I believe this should be in the past tense because the stack GENERATED the runs in the graph

cartography/models/spacelift/run.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "TRIGGERED"
+    properties: SpaceliftRunToUserRelProperties = SpaceliftRunToUserRelProperties()
+
+
+@dataclass(frozen=True)
+class SpaceliftRunToWorkerRelProperties(CartographyRelProperties):
+    """
+    Properties for the EXECUTES relationship between a Worker and a Run.
```
kunaals: Similar to GENERATED and TRIGGERED I do think that instead of EXECUTES and AFFECTS it's probably more correct to say EXECUTED and AFFECTED because this is all in the past and runs are all one-off

cartography/models/spacelift/space.py
```diff
@@ -0,0 +1,142 @@
+"""
+Spacelift Space data model.
+
+A SpaceliftSpace represents an organizational container within Spacelift.
+Spaces can contain stacks, policies, contexts, modules, and worker pools.
```
kunaals:
```suggestion
"""
A SpaceliftSpace represents an organizational container within Spacelift.
Spaces can contain stacks, policies, contexts, modules, and worker pools.
```

kunaals: superfluous

kunaals: i'd remove from the top of all the Spacelift models

docs/root/modules/spacelift/schema.md
```diff
@@ -0,0 +1,318 @@
+## Spacelift Schema
+
+```mermaid
```
kunaals, line 3: Thanks for this mermaid diagram - this is great!

tests/data/spacelift/spacelift_data.py
```diff
@@ -0,0 +1,261 @@
...
+                    },
+                ],
+            },
+        ],
+    },
+}
+
+# EC2 instance data - mock existing EC2 instances in Neo4j
+# These would typically be created by the AWS EC2 sync module
+EC2_INSTANCES_DATA = [
```
kunaals, line 248: Could we use the EC2 mock data that we already have in https://github.com/cartography-cncf/cartography/blob/master/tests/data/aws/ec2/instances.py? And perhaps actually run the EC2 sync like we do in the EC2 tests already? https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py#L22-L38

tests/integration/cartography/intel/spacelift/test_spacelift.py
```diff
@@ -0,0 +1,281 @@
...
+            SET i.id = $instance_id,
+                i.region = $region,
+                i.instancetype = $instance_type,
+                i.state = $state
+            """,
+            instance_id=instance["InstanceId"],
+            region=instance["Region"],
+            instance_type=instance["InstanceType"],
+            state=instance["State"],
+        )
```
kunaals: Ah like this is where we could just use this and run the actual EC2 sync to make the tests more robust: https://github.com/cartography-cncf/cartography/blob/master/tests/integration/cartography/intel/aws/ec2/test_ec2_instances.py#L22-L38

In case someone changes the EC2 sync in a way that breaks this in the future

cartography/intel/spacelift/account.py
```diff
@@ -0,0 +1,82 @@
...
+    hostname = parsed.hostname or ""
+    # Extract subdomain (everything before .app.spacelift.io)
+    account_id = hostname.split(".")[0] if hostname else ""
+
+    if not account_id:
+        raise ValueError(
+            f"Could not extract account ID from API endpoint: {api_endpoint}"
+        )
+
+    logger.info(f"Extracted account ID: {account_id}")
```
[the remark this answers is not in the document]
achantavy: This isn't necessarily sensitive but I do agree that we shouldn't log here because it's overly verbose

cartography/intel/spacelift/account.py
```diff
@@ -0,0 +1,82 @@
...
+from cartography.models.spacelift.spaceliftaccount import SpaceliftAccountSchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_account(api_endpoint: str) -> str:
+
+    logger.info("Extracting Spacelift account information from API endpoint")
```
achantavy: Not necessary to log this

cartography/intel/spacelift/workerpools.py
```diff
@@ -0,0 +1,135 @@
...
+    worker_pools_data: list[dict[str, Any]],
+    update_tag: int,
+    account_id: str,
+) -> None:
+    """
+    Load Spacelift worker pools data into Neo4j using the data model.
+    """
+    if not worker_pools_data:
+        logger.warning("No worker pools data provided to load_worker_pools")
+        return
```
achantavy: Not necessary because `load()` handles this; remove

docs/root/modules/spacelift/schema.md
achantavy: +1 these are so helpful

tests/integration/cartography/intel/spacelift/test_spacelift.py
```diff
@@ -0,0 +1,304 @@
...
+    assert actual_ec2_nodes is not None
+    assert expected_ec2_nodes == actual_ec2_nodes
+
+    common_job_parameters = {
+        "UPDATE_TAG": TEST_UPDATE_TAG,
+        "SPACELIFT_ACCOUNT_ID": TEST_ACCOUNT_ID,
+        "account_id": TEST_ACCOUNT_ID,
+    }
+
+    # Sync all Spacelift resources in the correct order
```
achantavy:
```suggestion
    # Act: Sync all Spacelift resources in the correct order
```

tests/integration/cartography/intel/spacelift/test_spacelift.py
```diff
@@ -0,0 +1,304 @@
...
+):
+    """
+    End-to-end integration test for Spacelift module.
+    Tests syncing of all Spacelift resources and their relationships,
+    including Run-[:AFFECTED]->EC2Instance relationships.
+
+    This test uses the real AWS EC2 sync to populate EC2 instances,
+    making it more robust than manually creating EC2 nodes.
+    """
+    # Mock all API calls using the mock data file
```
achantavy:
```suggestion
    # Arrange
    # Mock all API calls using the mock data file
```

tests/integration/cartography/intel/spacelift/test_spacelift.py
```diff
@@ -0,0 +1,304 @@
...
+    )
+    sync_runs(
+        neo4j_session,
+        spacelift_session,
+        TEST_API_ENDPOINT,
+        TEST_ACCOUNT_ID,
+        common_job_parameters,
+    )
+
+    # Check that SpaceliftAccount nodes were created
```
achantavy, line 165:
```suggestion
    # Assert
    # Check that SpaceliftAccount nodes were created
```

Issue #2027, opened 2025-10-27
Unable to load NODE


kunaals:
### Expected behavior
When running certain large syncs, in this case the EC2 security group sync, we see an error from Neo4j `Unable to load NODE 4:.....`

### Actual behavior
We would expect the sync to run successfully.

### To Reproduce
Seems somewhat random but generally occurs on large environments and in places where we run things concurrently.

### Logs
```
[2025-10-27T08:53:30Z] ERROR sync.1761548400144-e0d50e9f: Error during Cartography sync: Traceback (most recent call last):
  File "/app/app/sync/__init__.py", line 1350, in run_sync
    aws_provider.run(
    ~~~~~~~~~~~~~~~~^
        sync_logger=sync_logger,
        ^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
        stop_event=stop_event,
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/app/app/sync/providers/aws.py", line 286, in run
    result = future.result()
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/app/app/sync/__init__.py", line 448, in wrapper
    return fn(*args, **kwargs)
  File "/app/app/sync/providers/aws.py", line 247, in sync_account
    _sync_one_account(
    ~~~~~~~~~~~~~~~~~^
        neo4j_session=neo4j_session,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
        sync_logger=sync_logger,
        ^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/app/app/sync/providers/aws.py", line 94, in _sync_one_account
    RESOURCE_FUNCTIONS[func_name](**sync_args)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/cartography/util.py", line 225, in timed
    return method(*args, **kwargs)
[...]

PR #2028, opened 2025-10-27
fix(tx): Retry certain ClientErrors

kunaals:
### Summary
Add selective retry logic for `Neo.ClientError.Statement.EntityNotFound` errors that occur during concurrent write operations.

**Problem:**
Cartography syncs frequently encounter `EntityNotFound` errors when running concurrent provider syncs (AWS + GCP + Okta, etc.), especially under page cache pressure in Neo4j Aura. These errors occur because Neo4j's query execution pipeline reads and binds entity references before acquiring write locks. When multiple threads execute concurrent MERGE/DELETE operations, one thread can delete an entity that another thread has already referenced, resulting in `EntityNotFound`.

**Why Retry is Safe:**
Neo4j maintainers explicitly state in [neo4j/neo4j#6823](https://github.com/neo4j/neo4j/issues/6823) that `EntityNotFound` during concurrent operations is expected behavior and should be retried. However, the Neo4j Python driver classifies it as a non-retryable `ClientError` because most ClientErrors (syntax, auth, invalid params) are permanent failures.

**Solution:**
Implemented selective retry logic that:
- Retries ONLY `Neo.ClientError.Statement.EntityNotFound` (up to 5 times with exponential backoff)
- Fails immediately on all other ClientErrors (syntax errors, auth failures, etc.)
- Uses custom logging to provide context about concurrent write operations

**Changes Made:**
1. Added helper functions to identify EntityNotFound errors
2. Retrofitted existing functions with retry logic:
3. Updated all remaining locations that used `execute_write()` directly to use `execute_write_with_retry()`

**Impact:**
- All Cartography write operations now automatically retry EntityNotFound errors
- Reduces sync failures from transient concurrent write conflicts
- Particularly beneficial for users running large syncs or multiple concurrent providers
- No changes required to existing intel modules

### Related issues or links
- https://github.com/neo4j/neo4j/issues/6823 - Neo4j maintainers confirm EntityNotFound should be retried during concurrent operations
- https://github.com/cartography-cncf/cartography/issues/2027 related to page cache eviction issues commonly seen in Neo4j Aura under memory pressure

### Checklist

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -872,7 +872,27 @@ class CLI:
             default="SPACELIFT_API_TOKEN",
             help=(
                 "The name of an environment variable containing the Spacelift API token. "
-                "Required if you are using the Spacelift intel module. Ignored otherwise."
+                "Alternative to using API key ID/secret. Ignored if API key credentials are provided."
+            ),
+        )
+        parser.add_argument(
+            "--spacelift-api-key-id-env-var",
+            type=str,
+            default="SPACELIFT_API_KEY_ID",
+            help=(
+                "The name of an environment variable containing the Spacelift API key ID. "
+                "Use with --spacelift-api-key-secret-env-var for automatic token exchange. "
+                "Alternative to providing a pre-generated token."
+            ),
+        )
+        parser.add_argument(
+            "--spacelift-api-key-secret-env-var",
+            type=str,
+            default="SPACELIFT_API_KEY_SECRET",
+            help=(
+                "The name of an environment variable containing the Spacelift API key secret. "
+                "Use with --spacelift-api-key-id-env-var for automatic token exchange. "
+                "Alternative to providing a pre-generated token."
             ),
         )
         parser.add_argument(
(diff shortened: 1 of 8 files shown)


cartography/client/core/tx.py
```diff
@@ -20,6 +24,158 @@
...
+                    f"Successfully recovered from EntityNotFound error after {entity_attempts} "
+                    f"{'retry' if entity_attempts == 1 else 'retries'}. Function: {target}"
+                )
+            return result
+        except _NETWORK_EXCEPTIONS as exc:
+            if network_attempts >= _MAX_NETWORK_RETRIES - 1:
+                raise
+            network_attempts += 1
+            wait = next(network_wait)
+            if wait is None:
```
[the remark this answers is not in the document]
kunaals: This is wrong. We saw this edge case in production.

cartography/client/core/tx.py
```diff
@@ -242,7 +456,38 @@ def write_list_of_dicts_tx(
...
+
+    :param neo4j_session: The Neo4j session
+    :param query: The Neo4j write query to run
+    :param data_batch: A batch of data to write
+    :param kwargs: Additional keyword args for the query
+    :return: None
+    """
+
+    def _operation() -> None:
+        neo4j_session.write_transaction(
```
achantavy: Use execute_write()

cartography/client/core/tx.py
```diff
@@ -20,6 +24,158 @@
...
+            backoff_handler(
+                {
+                    "exception": exc,
+                    "target": target,
+                    "tries": network_attempts,
+                    "wait": wait,
+                }
+            )
+            time.sleep(wait)
+            del exc
```
achantavy: Might not be necessary, I think Python should clear out exception references at the end of the except block

cartography/client/core/tx.py
```diff
@@ -20,6 +24,158 @@
...
+            _entity_not_found_backoff_handler(
+                {
+                    "exception": exc,
+                    "target": target,
+                    "tries": entity_attempts,
+                    "wait": wait,
+                }
+            )
+            time.sleep(wait)
+            del exc
```
achantavy: Same here

cartography/client/core/tx.py
```diff
@@ -242,7 +456,38 @@ def write_list_of_dicts_tx(
...
+    concurrent operations.
+
+    :param neo4j_session: The Neo4j session
+    :param query: The Neo4j write query to run
+    :param data_batch: A batch of data to write
+    :param kwargs: Additional keyword args for the query
+    :return: None
+    """
+
+    def _operation() -> None:
```
achantavy: Non-blocking style nit: is there a way to do this without defining the function inline like this?

kunaals: Yeah we could use a partial instead https://docs.python.org/3/library/functools.html#functools.partial

PR #2036, opened 2025-10-30
feat(ontology): enable Ontology users for AWS and Okta

jychp:
### Summary

Adds ontology support for **AWS** and **Okta** user entities, improving cross-source identity mapping and unification.

### Context

This update extends ontology coverage to key identity providers and introduces a mechanism to prevent unwanted node creation when no definitive source is available.

### Changes proposed

* **Ontology integration:** Added ontology mapping support for AWSUser and OktaUser nodes.
* **Quality safeguard:** Introduced a new flag `eligible_for_source` in `OntologyMappingNode` to prevent low-confidence or incomplete node creation when no source of truth is configured.

### How to test this PR

1. Run ontology sync with AWS and Okta enabled.
2. Verify that ontology nodes are correctly linked for AWS/Okta users.
3. Confirm that no orphan ontology nodes are created when no eligible source is configured.

### Checklist

* [x] Unit and ontology integration tests updated.
* [x] Verified graph correctness in local/real environment.

### DIFF
--- a/tests/unit/cartography/intel/ontology/test_ontology_mapping.py
+++ b/tests/unit/cartography/intel/ontology/test_ontology_mapping.py
@@ -12,7 +12,9 @@ MODELS = list(load_models(cartography.models))
 # Unfortunately, some nodes are not yet migrated to the new data model system.
 # We need to ignore them in this test for now as we are not able to load their model class.
 # This is a temporary workaround until all models are migrated.
-OLD_FORMAT_NODES = ['OktaUser', ]
+OLD_FORMAT_NODES = [
+    "OktaUser",
+]
 
 
 def test_ontology_mapping_modules():


docs/root/modules/ontology/index.md
```diff
@@ -68,6 +68,28 @@ This prevents creating ontology nodes that cannot be properly identified or matc
...
+- Required fields are not available or reliable enough for primary node creation
+
+**Example:**
+```python
+OntologyNodeMapping(
+    node_label="AWSUser",
+    eligible_for_source=False,  # Cannot create User ontology nodes
+    fields=[
+        OntologyFieldMapping(ontology_field="username", node_field="name")
+    ],
```
achantavy, line 87: side note, at some point we should find a way to make aws user eligible

cartography/intel/ontology/utils.py
```diff
@@ -56,7 +63,7 @@ def get_source_nodes_from_graph(
                 for field in node.fields:
                     value = node_data.get(field.node_field)
                     # Skip nodes missing required fields
-                    if field.required and value is None:
+                    if field.required and value in (None, ""):
```
achantavy: Is this an equivalent statement?

```python
if field.required and not value:
```

basically checking if `value` is Falsy?

tests/unit/cartography/intel/ontology/test_ontology_mapping.py
```diff
@@ -38,6 +38,9 @@ def test_ontology_mapping_fields():
     for _, mappings in ONTOLOGY_MAPPING.items():
         for module_name, mapping in mappings.items():
             for node in mapping.nodes:
+                # TODO: Remove that uggly exception once all models are migrated to the new data model system
+                if node.node_label in ("OktaUser",):
```
achantavy: Should we keep these exceptions in a defined constant at the top of this file?

achantavy (commented):
left comments on small details

achantavy (approved):

PR #2042, opened 2025-11-03
feat: Add EC2 ownership tracking to Spacelift module

shyammukund:
- (https://github.com/cartography-cncf/cartography/pull/2022)

### Checklist

image


cartography/intel/spacelift/ec2_ownership.py
```diff
@@ -0,0 +1,194 @@
...
+from cartography.util import aws_handle_regions
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+# Regex pattern to match EC2 instance IDs
+INSTANCE_ID_PATTERN = re.compile(r"\b(i-[0-9a-f]{8,17})\b")
+
+
+@aws_handle_regions
```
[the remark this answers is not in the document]
shyammukund: lol cubic told me to add this but now doesn't like it. lemme know whether i should remove this decorator

tests/integration/cartography/intel/spacelift/test_spacelift.py
```diff
@@ -239,23 +239,6 @@ def test_spacelift_end_to_end(
     assert actual_run_nodes is not None
     assert expected_run_nodes == actual_run_nodes
 
-    # Check that Run-[:AFFECTED]->EC2Instance relationships were created
```
kunaals, line 242: Hm can we keep the old AFFECTED rels too? So if the EC2 instance IDs are readable from the Spacelift API, we still ingest them and create the rels

docs/root/modules/spacelift/schema.md
```diff
@@ -276,14 +277,25 @@ Representation of a job that can touch infrastructure. It is the execution insta
...
-    (SpaceliftRun)-[AFFECTED{action}]->(EC2Instance)
+    (SpaceliftRun)<-[COMMITTED]-(SpaceliftGitCommit)
     ```
 
-    The `action` property indicates the action taken on the instance (e.g., "create", "update", "delete").
-    This relationship uses the `one_to_many` pattern to connect a single run to multiple EC2 instances it manages.
+- SpaceliftRuns can affect EC2 Instances (optional, requires EC2 ownership configuration):
+
+    ```
+    (SpaceliftRun)-[AFFECTED{event_time, event_name, aws_account, aws_region}]->(EC2Instance)
```
kunaals: nit: remove the field names from the rel because they're already beneath. and we dont want people confused thinking that these are part of the rel name 
```suggestion
    (SpaceliftRun)-[AFFECTED]->(EC2Instance)
```

cartography/intel/spacelift/__init__.py
```diff
@@ -95,6 +97,24 @@ def start_spacelift_ingestion(neo4j_session: neo4j.Session, config: Config) -> N
         common_job_parameters,
     )
 
+    # Sync EC2 ownership relationships from CloudTrail data (optional)
+    if (
+        hasattr(config, "spacelift_ec2_ownership_aws_profile")
+        and hasattr(config, "spacelift_ec2_ownership_s3_bucket")
+        and hasattr(config, "spacelift_ec2_ownership_s3_key")
+    ):
```
kunaals:
```suggestion
    if all(hasattr(config, attr) for attr in [
        "spacelift_ec2_ownership_aws_profile",
        "spacelift_ec2_ownership_s3_bucket",
        "spacelift_ec2_ownership_s3_key",
    ]):
```
slightly more pythonic to use all()

kunaals: If someone passes in the s3_bucket and s3_key but not the profile, it might be nice to just use the default profile (i.e. `aws_session = boto3.Session()`). you could remove the spacelift_ec2_ownership_aws_profile attr check from here and wrap the below aws_session. will leave a separate comment for that

cartography/intel/spacelift/__init__.py
```diff
@@ -95,6 +97,24 @@ def start_spacelift_ingestion(neo4j_session: neo4j.Session, config: Config) -> N
...
+        aws_session = boto3.Session(
+            profile_name=config.spacelift_ec2_ownership_aws_profile
+        )
```
kunaals:
```suggestion
        if hasattr(config, "spacelift_ec2_ownership_aws_profile"):
            aws_session = boto3.Session(
                profile_name=config.spacelift_ec2_ownership_aws_profile
            )
        else:
            aws_session = boto3.Session()
```

cartography/intel/spacelift/ec2_ownership.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+    logger.info(f"Fetching EC2 ownership data from s3://{bucket_name}/{object_key}")
+
+    # Create S3 client from the boto3 session
+    s3_client = aws_session.client("s3")
+
+    # Get the object from S3
+    response = s3_client.get_object(Bucket=bucket_name, Key=object_key)
+
+    # Read the object body (returns a StreamingBody)
```
kunaals:
```suggestion
```

cartography/intel/spacelift/ec2_ownership.py
```diff
@@ -0,0 +1,194 @@
...
+) -> list[dict[str, Any]]:
+    """
+    Fetch EC2 ownership data from S3 bucket containing Athena query results.
+    """
+    logger.info(f"Fetching EC2 ownership data from s3://{bucket_name}/{object_key}")
+
+    # Create S3 client from the boto3 session
+    s3_client = aws_session.client("s3")
+
+    # Get the object from S3
```
kunaals:
```suggestion
```

cartography/intel/spacelift/ec2_ownership.py
```diff
@@ -0,0 +1,194 @@
...
+    response = s3_client.get_object(Bucket=bucket_name, Key=object_key)
+
+    # Read the object body (returns a StreamingBody)
+    object_body = response["Body"].read()
+
+    # Decode bytes to string and parse JSON
+    json_content = json.loads(object_body.decode("utf-8"))
```
kunaals, line 38: `json_content = json.load(io.TextIOWrapper(response["Body"], encoding="utf-8"))`
Something like this may be more efficient for large files, just something to think about if you run into sluggishness, non blocker

kunaals (commented):
Left some small suggestions - looks quite close to me

cartography/intel/spacelift/ec2_ownership.py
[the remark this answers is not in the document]
kunaals: I think let's keep it, it guards against important AWS error codes

cartography/cli.py
```diff
@@ -866,6 +866,34 @@ def _build_parser(self):
                 "Required if you are using the Spacelift intel module. Ignored otherwise."
             ),
         )
+        parser.add_argument(
+            "--spacelift-ec2-ownership-aws-profile",
+            type=str,
+            default=None,
```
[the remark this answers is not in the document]
kunaals: I think it's onto something here @shyammukund. either you can check whether the attribute exists AND is none, or just argparse.SUPPRESS

tests/integration/cartography/intel/spacelift/test_spacelift.py
shyammukund: @kunaals will there by any issues with having 2 different rels both called AFFECTED?

kunaals: It makes querying a little more confusing but the relationships have should have unique IDs if setup correctly

PR #2046, opened 2025-11-05
feat(spacelift): CloudTrailEvent node type 

kunaals:
### Summary
Refactors Spacelift EC2 ownership tracking to use proper CloudTrailEvent nodes instead of MatchLink relationships. Fixes a bug where multiple CloudTrail events from the same run affecting the same EC2 instance would overwrite each other. Each CloudTrail event is now preserved as its own node using CloudTrail's native eventid as the identifier.

### Related issues or links

- Builds on #2042 (original EC2 ownership tracking feature)
- Related to #2022 & #2045

### Checklist

### DIFF
--- a/cartography/intel/azure/__init__.py
+++ b/cartography/intel/azure/__init__.py
@@ -14,7 +14,9 @@ from . import cosmosdb
 from . import data_lake
 from . import functions
 from . import logic_apps
+from . import monitor
 from . import resource_groups
+from . import security_center
 from . import sql
 from . import storage
 from . import subscription
@@ -96,6 +98,20 @@ def _sync_one_subscription(
         common_job_parameters,
     )
     data_lake.sync(
+        neo4j_session,
+        credentials,
+        subscription_id,
+        update_tag,
+        common_job_parameters,
+    )
+    monitor.sync(
+        neo4j_session,
+        credentials,
+        subscription_id,
+        update_tag,
+        common_job_parameters,
+    )
+    security_center.sync(
         neo4j_session,
         credentials,
         subscription_id,
--- a/cartography/intel/azure/monitor.py
+++ b/cartography/intel/azure/monitor.py
@@ -1,0 +1,105 @@
+import logging
+from typing import Any
+
+import neo4j
+from azure.core.exceptions import HttpResponseError
+from azure.mgmt.monitor import MonitorManagementClient
+
+from cartography.client.core.tx import load
+from cartography.graph.job import GraphJob
+from cartography.models.azure.monitor import AzureMonitorMetricAlertSchema
+from cartography.util import timeit
+
+from .util.credentials import Credentials
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_metric_alerts(client: MonitorManagementClient) -> list[dict]:
+    """
+    Get a list of Metric Alerts from the given Azure subscription.
+    """
+    try:
+        return [
+            alert.as_dict() for alert in client.metric_alerts.list_by_subscription()
+        ]
+    except HttpResponseError:
+        logger.warning(
+            "Failed to get Azure Monitor Metric Alerts due to a transient error.",
+            exc_info=True,
+        )
+        return []
+
+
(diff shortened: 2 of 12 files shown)


cartography/models/spacelift/cloudtrailevent.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "AFFECTED"
+    properties: CloudTrailEventToEC2InstanceRelProperties = (
+        CloudTrailEventToEC2InstanceRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class CloudTrailEventSchema(CartographyNodeSchema):
```
achantavy: Let's rename to CloudtrailSpaceliftEvent because it's not a general cloudtrail event and we don't want other people importing and using this schema as one.

cartography/intel/spacelift/ec2_ownership.py
```diff
@@ -235,18 +257,25 @@ def sync_ec2_ownership(
...
+    if events:
+        load_cloudtrail_events(neo4j_session, events, update_tag, account_id)
     else:
-        logger.warning("No EC2 ownership mappings found - no relationships created")
+        logger.warning("No CloudTrail events found - no nodes created")
+
+    common_job_parameters = {
+        "UPDATE_TAG": update_tag,
+        "SPACELIFT_ACCOUNT_ID": account_id,
+        "account_id": account_id,
```
achantavy: If in future we support spacelift things with other providers, the name 'account_id' could be confusing so maybe we should rename this to 'aws_account_id'

achantavy (changes requested):
Couple of comments on names.

I agree with this fix, just asking one more non-blocking question on the PR description for documentation sake:

> Fixes a bug where multiple CloudTrail events from the same run affecting the same EC2 instance would overwrite each other.

Could we make more fields on the matchlink relationship so that they're unique? Like could we add in the action type (describe, run, start, etc) and the person who triggered it to the matchlink rel so that they don't overwrite each other? Or is it a limitation (or bug) of the matchlink implementation that if we have `(node1)-[matchlink]-(node2)` then  there must be strictly 1 matchlink?

cartography/models/spacelift/cloudtrailevent.py
```diff
@@ -0,0 +1,116 @@
...
+
+    target_node_label: str = "EC2Instance"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "instanceid": PropertyRef("instance_ids", one_to_many=True),
+        }
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "AFFECTED"
+    properties: CloudTrailEventToEC2InstanceRelProperties = (
```
shyammukund: what if we put the event name as a property on the rel here as well.

kunaals:
@achantavy @shyammukund thanks for the review

Alex regarding your question about "is it a limitation (or bug) of the matchlink implementation that if we have (node1)-[matchlink]-(node2) then there must be strictly 1 matchlink?" **Yes,** inside matchlinks we do a MERGE on the links, ie `(node1)-[rel_name]-(node2)` and it overwrites existing AFFECTS rels between that particular run and the EC2 instance

cartography/intel/spacelift/ec2_ownership.py
shyammukund: I realized that the account_id and SPACELIFT_ACCOUNT_ID are the same thing. ill remove  one of these

tests/integration/cartography/intel/spacelift/test_spacelift.py
```diff
@@ -119,7 +119,6 @@ def test_spacelift_end_to_end(
 
     common_job_parameters = {
         "UPDATE_TAG": TEST_UPDATE_TAG,
-        "SPACELIFT_ACCOUNT_ID": TEST_ACCOUNT_ID,
```
achantavy, line 122: Why remove this?

achantavy (approved):

Issue #1658, opened 2025-06-28
Feature Request: Add Support for Amazon Neptune (and other openCypher-compatible graph databases)


triggan:
*Title*: Add Support for Amazon Neptune (and other openCypher-compatible graph databases)

*Description*:
> I would like to propose adding support for Amazon Neptune as an alternative graph database backend for Cartography, while maintaining full backward compatibility with existing Neo4j deployments. This enhancement would enable users to leverage AWS-managed graph database services without requiring code changes to their existing Cartography workflows.

*Background*:
>   Amazon Neptune supports the Neo4j Bolt protocol and openCypher query language, making it  compatible with Cartography's existing graph operations. Many organizations already using AWS infrastructure would benefit from being able to use Neptune as their graph database backend, taking advantage of AWS's managed service benefits (automated backups, scaling, security, etc.) while maintaining their existing Cartography investment.

## Proposed Enhancement

The enhancement would introduce database backend selection through CLI parameters while preserving all existing functionality:
  1. Database Type Selection: Add a `--graph-database-type` parameter to choose between `neo4j` (default) and `neptune` (or adding additional names for any future graph database support)
  2. Unified Connection Parameters: Introduce generic `--graph-*` parameters alongside existing `--neo4j-*` parameters for backward compatibility
  3. Neptune-Specific Authentication: Neptune would require AWS IAM Authentication when used against Neptune clusters with IAM Authentication enabled.  Most of this integration is already available in the AWS modules within the project. Some additional support may require use of `botocore` directly.
  4. Transparent Operation: Leverage Neptune's Neo4j Bolt driver compatibility to minimize code changes.
  
### Existing Neo4j usage (unchanged)
```
  cartography --neo4j-uri bolt://localhost:7687
```
  ### New Neptune usage
```
  cartography --graph-database-type neptune \
    --graph-uri bolt://neptune-cluster.amazonaws.com:8182
```

  ## Implementation Considerations
  ### Areas Requiring Changes
  1. CLI Parameter Structure (cartography/cli.py)
    - Add database type selection parameter.  An alternative approach would be to parse the URI and look for the use of `neptune.amazonaws.com` and infer the use of Neptune.  Though this approach may not be conducive for support of other graph databases.
    - Add unified graph connection parameters.
    - Maintain backward compatibility with existing --neo4j-* parameters.
  2. Connection Management (cartography/sync.py)
    - Implement database driver factory pattern
[...]

PR #2047, opened 2025-11-05
Export to file instead of neo4j

vdeturckheim:
### Summary
This PR introduces the ability to export a scan result to ndjson files rather than neo4j.
The goal is to allow a database-agnostic output allowing users to write the result anywhere they want.

At this point, I am mostly asking whether this is something you'd be interested to land. If that's the case, I'll make sure all intel modules are compatible with this output format.

### Related issues or links

- https://github.com/cartography-cncf/cartography/issues/1658 
  - This would provide a DIY way of supporting any db inckuding Neptune.

### Checklist

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -77,14 +77,6 @@ class CLI:
             ),
         )
         parser.add_argument(
-            "--export-format",
-            type=str,
-            default="ndjson",
-            help=(
-                "Export format. Currently only 'ndjson' is supported."
-            ),
-        )
-        parser.add_argument(
             "--no-neo4j-write",
             action="store_true",
             help=(
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -65,6 +65,33 @@ class CLI:
             action="store_true",
             help="Restrict cartography logging to warnings and errors only.",
         )
+        # Optional export-only sink (off by default)
+        parser.add_argument(
+            "--export-file",
+            type=str,
+            default=None,
+            help=(
+                "Write a gzipped NDJSON export of graph nodes and edges to this path. "
+                "When set, cartography will tee writes to this file while continuing to write to Neo4j unless "
+                "--no-neo4j-write is also provided."
+            ),
+        )
+        parser.add_argument(
+            "--export-format",
+            type=str,
+            default="ndjson",
+            help=(
+                "Export format. Currently only 'ndjson' is supported."
+            ),
+        )
+        parser.add_argument(
+            "--no-neo4j-write",
+            action="store_true",
+            help=(
+                "Do not write to Neo4j. Useful when generating an export file for offline processing. "
+                "Requires --export-file."
+            ),
+        )
         parser.add_argument(
             "--neo4j-uri",
             type=str,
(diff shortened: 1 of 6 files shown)


cartography/sinks/file_export.py
```diff
@@ -0,0 +1,156 @@
...
+        with self._lock:
+            if self._closed:
+                return
+            try:
+                self._fh.close()
+            finally:
+                self._closed = True
+
+
+class FileExportSink:
```
achantavy, line 40: As discussed on Slack, I'm planning on having 2 export formats: 

1. this one, which is a general graph format that writes node + edge data to a file. 
2. another format that I have not implemented yet, which will be a cartography-specific JSON-like format. I'm thinking of naming it "cartography-offline-record" or something. The idea is we make json files that look like this:

```
  {
    "record_type": "offline",
    "schema_id": "cartography.models.aws.guardduty.findings.GuardDutyFindingSchema",
    "data": [{"id": "finding-123", "title": "...", "resource_id": "i-123"}],
    "kwargs": {"lastupdated": 1700000000, "Region": "us-east-1", "AWS_ID": "123456789012"}
  }
```

and then we have a `cartography-import` cli command that will take those `offline` records, run them against the `schema_id`, and then write them to Neo4j. This way we can support moving data between environments where having permanent trust relationships set up is not feasible.

So, with that in mind, we should have a more specific name for your format than `FileExportSink`. How's `GeneralGraphSink` or `NodeEdgeSink` or something else?

cartography/sinks/file_export.py
```diff
@@ -0,0 +1,156 @@
+import atexit
+import gzip
+import io
+import json
+import os
+import threading
+from dataclasses import asdict
+from typing import Any, Dict, Iterable, List, Optional
```
achantavy, line 8: Non-blocking nit: use the Python 3.9 built-in typehints instead of importing `Dict`, `List`, `Optional`.

cartography/sinks/file_export.py
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        *,
+        uid: Any,
+        labels: List[str],
+        props: Dict[str, Any],
+        update_tag: Any,
+        sub_resource_label: Optional[str] = None,
+        sub_resource_id: Optional[Any] = None,
+    ) -> None:
+        rec: Dict[str, Any] = {
```
achantavy, line 70: Could we add documentation on the shape of this data for this format?

cartography/sinks/file_export.py
```diff
@@ -0,0 +1,156 @@
...
+        rel_type: str,
+        update_tag: Any,
+        props: Optional[Dict[str, Any]] = None,
+        # Fallback matching description if uid is unknown
+        from_match: Optional[Dict[str, Any]] = None,
+        to_match: Optional[Dict[str, Any]] = None,
+        sub_resource_label: Optional[str] = None,
+        sub_resource_id: Optional[Any] = None,
+    ) -> None:
+        rec: Dict[str, Any] = {
```
achantavy, line 98: Same thing on having documentation for this format.

cartography/cli.py
```diff
@@ -65,6 +65,25 @@ def _build_parser(self):
             action="store_true",
             help="Restrict cartography logging to warnings and errors only.",
         )
+        # Optional export-only sink (off by default)
+        parser.add_argument(
+            "--export-file",
```
achantavy, line 70: Let's have a more specific name since (from my other comment) I think we will have at least this format and one more export format.

achantavy (changes requested):
Thanks for the discussion [on Slack](https://cloud-native.slack.com/archives/C080M2LRLDA/p1762367604839939)! Left some thoughts about documentation and export formats. Before I continue with my changes I would like to get yours merged in so that we minimize merge conflicts. Great PR

PR #2049, opened 2025-11-06
refactor: rules engine

jychp:
### Summary

Refactor the rules engine from a 4-level hierarchy (**Framework → Requirement → Finding → Fact**) to a lean 2-level model (**Rule → Fact**). This removes framework/requirement abstractions, flattens the file layout, simplifies execution, and streamlines the CLI.

### Context

Originally, the open source Cartography project introduced a **rules framework** to organize graph queries ("facts") under compliance and security standards. Each **Framework** (e.g., MITRE ATT&CK) contained multiple **Requirements** (e.g., T1234), and each requirement included one or more **Facts** — Cypher queries identifying relevant assets.

The intent was to make it simple for contributors to add new security or compliance checks in a structured way. In practice, adoption stalled. Adding a single query required not only knowing what to measure and how to express it in Cypher, but also understanding the hierarchical **Framework → Requirement → Finding** model and some GRC context. That extra layer made contribution too complex for most engineers, slowing community growth and innovation.

### Changes proposed

* **Model and Types**
  * Removed `Framework` and `Requirement` models.
  * Simplified the data model to just `Rule` and `Fact`.
  * Added `Fact.maturity` for lifecycle visibility and kept `Rule.version`, `tags`, and `references` for traceability.

* **Output Models**
  * Introduced Pydantic-based `Finding` models for structured, typed output.
  * Each `Rule` declares an `output_model` class that extends `Finding`.
  * The output model defines the expected fields from `cypher_query` results.
  * Automatic handling of:
    * Source field population (module name)
    * Extra fields storage
    * Type coercion (numbers to strings, lists to comma-separated strings, etc.)

* **Execution**
  * Replaced `run_frameworks()` traversal with a flat `run_rules()` flow — rules directly execute their facts without intermediate layers.
  * Results are parsed through the Pydantic output model for validation and structure.

* **Filesystem layout**
  * Removed nested `frameworks/.../requirements/...` directories.
  * Added a flat `cartography/rules/data/rules/` structure — each rule is a single Python file (e.g., `mfa_missing.py`, `object_storage_public.py`, etc.).

## Reviewer notes

* Focus on the model changes in `spec/model.py`, especially the dual query structure and `Finding` integration.
* Check CLI behavior in `cli.py` and traversal in `runners.py`.
* Review output model definitions and their field mappings.
* Sanity-check that each new rule file declares `facts`, `output_model`, and meaningful `tags`.

## Why this matters

* **Simpler mental model:** Two levels, not four.
* **Framework-agnostic:** Same rule can map to MITRE, CIS, or product-specific tags.
* **Faster/cleaner execution:** Shallower traversal, fewer joins.
* **Easier to extend:** Add a single file per rule; facts remain composable.
* **Type-safe output:** Pydantic models ensure consistent, validated output structure.


cartography/rules/spec/model.py
```diff
@@ -47,55 +82,93 @@ class Fact:
...
+                    parsed_output[key] = value
+            try:
+                # Try to parse normally
+                result.append(self.output_model(**parsed_output))
+            except ValidationError as e:
+                # Handle validation errors
+                logger.warning(
+                    "Validation error parsing rule output for rule %s: %s",
+                    self.id,
+                    type(e).__name__,
```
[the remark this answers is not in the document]
jychp: This allow to avoid logging PII

cartography/rules/spec/model.py
```diff
@@ -47,55 +82,93 @@ class Fact:
...
-    requirement_url: str | None = None
-    """A URL reference to the requirement in the framework, e.g. https://attack.mitre.org/techniques/T1098/"""
+    # Config to coerce numbers to strings during instantiation
+    model_config = ConfigDict(coerce_numbers_to_str=True)
+
+    # Coerce o strings
+    @no_type_check
+    @model_validator(mode="before")
+    @classmethod
+    def coerce_to_string(cls, data: Any) -> Any:
```
achantavy, line 100: Is it possible to remove this function? I thought the point of pydantic is so that we don't need to do any weird typechecking like this

cartography/rules/spec/model.py
```diff
@@ -47,55 +82,93 @@ class Fact:
...
+            parsed_output: dict[str, Any] = {"extra": {}, "source": fact.module.value}
+            for key, value in result_item.items():
+                if key not in self.output_model.model_fields and value is not None:
+                    parsed_output["extra"][key] = value
+                else:
+                    parsed_output[key] = value
+            try:
+                # Try to parse normally
+                result.append(self.output_model(**parsed_output))
+            except ValidationError as e:
```
achantavy: Somewhat concerned about swallowing errors. Would it be advisable to loudly fail here?

cartography/models/core/model.py
```diff
@@ -0,0 +1,75 @@
...
+    CartographyRelSchema,
+    CartographyNodeProperties,
+    CartographyRelProperties,
+)
+
+NODES: dict[str, Type[CartographyNodeSchema]] = {}
+RELATIONSHIPS: dict[str, Type[CartographyRelSchema]] = {}
+
+
+def load_models(module, module_name: str | None = None) -> Generator[
```
achantavy: I think you might have mentioned this, but this func is for ontology and not for the rules engine, right?

cartography/models/core/model.py
```diff
@@ -0,0 +1,75 @@
...
+            sub_module_name = module_name
+        for v in sub_module.__dict__.values():
+            if not inspect.isclass(v):
+                continue
+            if v in _MODEL_CLASSES:
+                continue
+            if issubclass(v, _MODEL_CLASSES):
+                yield (sub_module_name, v)
+
+        if hasattr(sub_module, "__path__"):
```
achantavy: Is it possible to avoid these hasattr and runtime type checks? If not, np

cartography/rules/formatters.py
```diff
@@ -44,3 +52,57 @@ def _generate_neo4j_browser_url(neo4j_uri: str, cypher_query: str) -> str:
 
     # Construct the Neo4j Browser URL with pre-populated query
     return f"{browser_uri}browser/?cmd=edit&arg={encoded_query}"
+
+
+def to_serializable(obj):
```
achantavy, line 57: I'm not an expert here but runtime type checks seem scary. I was thinking that Pydantic would help us keep everything statically defined so we wouldn't need to do our own logic and do different things in different situations based on the type -- to me it feels dangerous

cartography/rules/spec/model.py
jychp: This function is only there as a safety net when the field type is `string`. Unfortunately, in a lot of cases the data is still too heterogeneous (dates can be a non-parsed string, a `datetime`, a timestamp stored as an `int`, or even a timestamp stored as a `string`).

It only applies when the expected type is `string`. Any field defined as `int` in the model will still come out as a proper `int`.

I tried removing this function and everything immediately started failing on type validation. So for now I think it’s a necessary evil.

cartography/rules/spec/model.py
jychp: I’m generally in favor of surfacing parsing errors without making them blocking, but that’s a pretty opinionated take.

I can totally let the error bubble up here if you prefer, your call.

cartography/models/core/model.py
jychp: This function was originally used for unit tests. I had moved it into utils during my previous attempt at auto-typing the return value based on a regex in the query.

It’s not used at runtime, so I’ll move it to test package to make that clearer.

cartography/models/core/model.py
jychp: This function is not used at runtime, will move it for clarity.

cartography/rules/formatters.py
jychp: This function is only used to convert the module’s models into JSON when calling `JSON.dumps()`. Those models are plain dataclasses; only the results are Pydantic classes.

I could’ve done a manual field-by-field transformation instead, but it would’ve been heavy, tedious, and not very maintainable.

I agree that runtime type checking can be risky, but for a simple serialization step used only for display, I think it’s safe.

PR #2050, opened 2025-11-09
feat(azure): Add Azure Tags for storage accounts

janithashri:
### Summary

This pull request introduces a standardized, shared model for ingesting Azure Tags across all resources. It implements this new pattern for **Azure Storage Accounts** as the first use case.

- **New Shared Node:** `:AzureTag`
- **New Relationship:** `(:AzureStorageAccount)-[:TAGGED]->(:AzureTag)`

### Description

This PR establishes the architectural foundation for Azure tagging in Cartography. Instead of creating specific tag nodes for every resource type, it uses a **shared node pattern**:
* All tags are ingested as generic `:AzureTag` nodes.
* The node ID is deterministic (`key:value`), ensuring that multiple resources with the same tag (e.g., `env:prod`) connect to the **same** tag node in the graph.
* This allows for powerful, simple global queries (e.g., "Find *all* resources where `env=prod`").

**Changes included:**
1.  Defined the global `AzureTagSchema`.
2.  Added a shared `transform_tags()` utility for use by all Azure modules.
3.  Updated the **Storage** module to ingest tags using this new pattern.
4.  Added comprehensive integration tests verifying the shared node behavior.

### Related issues or links

- Addresses part of #1736

### Checklist

Screenshot 2025-11-09 164735
Screenshot 2025-11-09 170239

- [x] Use the NodeSchema [data model](https://cartography-cncf-dev.readthedocs.io/en/latest/dev/writing-intel-modules.html#defining-a-node).
- [x] Confirm that the linter actually passes.

### DIFF
--- a/tests/integration/cartography/intel/azure/test_storage.py
+++ b/tests/integration/cartography/intel/azure/test_storage.py
@@ -43,9 +43,18 @@ def test_load_storage_account_data(neo4j_session):
 
 def test_load_storage_tags(neo4j_session):
     """
-    Test that tags are correctly loaded and linked to storage accounts.
+    Test that tags are correctly loaded and linked to storage accounts and subscription.
     """
-    # 1. Arrange: Load the storage accounts first so they exist to be tagged
+    # 1. Arrange: Create subscription and load storage accounts
+    neo4j_session.run(
+        """
+        MERGE (as:AzureSubscription{id: $subscription_id})
+        ON CREATE SET as.firstseen = timestamp()
+        SET as.lastupdated = $update_tag
+        """,
+        subscription_id=TEST_SUBSCRIPTION_ID,
+        update_tag=TEST_UPDATE_TAG,
+    )
     storage.load_storage_account_data(
         neo4j_session,
         TEST_SUBSCRIPTION_ID,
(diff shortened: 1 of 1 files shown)


cartography/models/azure/tags/storage_tag.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@dataclass(frozen=True)
+class AzureStorageTagsSchema(AzureTagSchema):
+    """
+    Inherits standard tag properties from AzureTagSchema and adds the
+    specific relationship to AzureStorageAccount.
+    """
+
+    other_relationships: OtherRelationships = OtherRelationships(
```
achantavy, line 36: Let's reverse this. Instead of defining the relationship on the tag to the storage account, let's define the relationship on the storage account to the tag. This way, as the code grows, the tag won't need to have multiple relationships defined on it, making the ingestion query inefficient

achantavy (changes requested):

cartography/models/azure/tags/storage_tag.py
```diff
@@ -0,0 +1,40 @@
...
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("resource_id")},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "TAGGED"
+    properties: StorageAccountToTagRelProperties = StorageAccountToTagRelProperties()
+
+
+@dataclass(frozen=True)
+class AzureStorageTagsSchema(AzureTagSchema):
```
achantavy: Let's avoid inheritance

cartography/models/azure/tags/tag.py
```diff
@@ -0,0 +1,20 @@
...
+class AzureTagProperties(CartographyNodeProperties):
+    # The ID is a string: "{key}:{value}"
+    id: PropertyRef = PropertyRef("id")
+    key: PropertyRef = PropertyRef("key")
+    value: PropertyRef = PropertyRef("value")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class AzureTagSchema(CartographyNodeSchema):
```
achantavy: As discussed in our sync, we should

- create a sub resource relationship to the subscription that this tag applies to
- include the subscription id in the tag itself to avoid collisions, so the `id` of the AzureTag will have the form `{sub_id}:{key}:{value}` or similar. 

We need this because we sync in the azure tags one subscription at a time, and when we clean them up we don't want to delete tags that apply to a subscription that is not currently being synced.

cartography/intel/azure/storage.py
achantavy, line 1058: Put the cleanup for Azure storage tags in here.

achantavy (changes requested):

achantavy (approved):

PR #2066, opened 2025-11-20
feat(ontology): add ComputeInstance and Container

jychp:
### Summary

This PR extends Cartography’s ontology by introducing three new global semantic labels: `APIKey`, `ComputeInstance`, and `Container`. These unified labels standardize how these resource types are modeled across cloud providers and infrastructure platforms.

### Context

Cartography currently models API keys, compute instances, and containers differently depending on the source module. This PR adds global semantic labels to align these resources under consistent ontology categories, enabling cross-provider queries and simplifying integrations.

### Related issues or links

None.

### Changes proposed

* Add the global semantic label `APIKey` and apply it to:

  * Anthropic (`AnthropicApiKey`)
  * OpenAI (`OpenAIApiKey`, `OpenAIAdminApiKey`)
  * Scaleway (`ScalewayApiKey`)
* Add the global semantic label `ComputeInstance` and apply it to:

  * AWS (`EC2Instance`)
  * DigitalOcean (`DigitalOceanDroplet`)
  * Scaleway (`ScalewayInstance`)
* Add the global semantic label `Container` and apply it to:

  * AWS (`ECSContainer`)
  * Azure (`AzureContainerInstance`)
  * Kubernetes (`KubernetesContainer`)
* Add new ontology mapping files under `cartography/models/ontology/mapping/data/`.
* Update model schemas to include `ExtraNodeLabels`.
* Update ontology documentation at `docs/root/modules/ontology/schema.md`.

### DIFF
--- a/cartography/models/ontology/mapping/data/containers.py
+++ b/cartography/models/ontology/mapping/data/containers.py
@@ -43,9 +43,7 @@ kubernetes_mapping = OntologyMapping(
                 OntologyFieldMapping(
                     ontology_field="image_digest", node_field="status_image_sha"
                 ),
-                OntologyFieldMapping(
-                    ontology_field="state", node_field="status_state"
-                ),
+                OntologyFieldMapping(ontology_field="state", node_field="status_state"),
                 # cpu: Not exposed as a direct field in KubernetesContainer node
                 # memory: Not exposed as a direct field in KubernetesContainer node
                 # region: Not exposed as a direct field in KubernetesContainer node


cartography/models/ontology/mapping/data/containers.py
```diff
@@ -0,0 +1,92 @@
...
+                OntologyFieldMapping(
+                    ontology_field="image_digest", node_field="status_image_sha"
+                ),
+                OntologyFieldMapping(
+                    ontology_field="status", node_field="status_state"
+                ),
+                # cpu: Not exposed as a direct field in KubernetesContainer node
+                # memory: Not exposed as a direct field in KubernetesContainer node
+                OntologyFieldMapping(
+                    ontology_field="region", node_field="cluster_name"
```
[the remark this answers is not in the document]
jychp: No, ontology mapping refer to node field, not the key in the data dict.
This is checked by unit tests anyway.

tests/unit/cartography/intel/ontology/test_ontology_mapping.py
```diff
@@ -65,19 +66,23 @@ def test_ontology_mapping_fields():
...
-                    assert model_property is not None, (
+                    found = False
+                    for model_class in model_classes:
+                        model_property = getattr(
+                            model_class.properties, mapping_field.node_field, None
+                        )
+                        if model_property is not None:
+                            found = True
+                            break
+                    assert found, (
```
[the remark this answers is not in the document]
jychp: The break statement only impact the first loop level, so all extra_fields a still tested

cartography/models/ontology/mapping/data/containers.py
kunaals: I think this is a mistake? Is the clustername really the region?

cartography/models/ontology/mapping/data/containers.py
```diff
@@ -0,0 +1,92 @@
+from cartography.models.ontology.mapping.specs import OntologyFieldMapping
```
kunaals, line 1: Do we want to include cluster / pod on here as well? 
Generally I see container management platforms structured like:

1. Kube: `cluster -> namespace -> pod -> container`
2. ECS: `cluster -> service (optional) -> task -> container`
3. Azure Container Instance: `container group -> container`
(not including Google Cloud Run because that's more like a lambda)

Overall this feels like: Cluster -> WorkloadGroup -> Workload -> Container
Azure Container Instance would only be Workload and Container

Should we have these as fields on Containers? Or should we create these as ontology nodes later and just create relationships then?

cartography/models/ontology/mapping/__init__.py
```diff
@@ -25,6 +29,8 @@
 # They are leveraged directly by the load functions of each module at ingestion time
 SEMANTIC_LABELS_MAPPING: dict[str, dict[str, OntologyMapping]] = {
     "useraccounts": USERACCOUNTS_ONTOLOGY_MAPPING,
+    "apikeys": APIKEYS_ONTOLOGY_MAPPING,
+    "computeinstance": COMPUTE_INSTANCE_ONTOLOGY_MAPPING,
```
kunaals, line 36: We need to add containers here, right?

docs/root/modules/ontology/schema.md
```diff
@@ -78,13 +76,40 @@ A user often has one or many user accounts.
...
+
+| Field | Description |
+|-------|-------------|
+| _ont_email | User's email address (often used as primary identifier). |
+| _ont_username | User's login name or username. |
+| _ont_fullname | User's full name. |
+| _ont_firstname | User's first name. |
+| _ont_lastname | User's last name. |
+| _ont_has_mfa | Whether multi-factor authentication is enabled for this account. |
+| _ont_inactive | Whether the account is inactive, disabled, suspended, or locked. |
```
kunaals, line 106: Would it be better to say status here or state?

cartography/models/ontology/mapping/data/containers.py
jychp: I think we want that on ontology. Will do that on next PR (will try to make more atomic PR)

cartography/models/ontology/mapping/__init__.py
jychp: Thx good catch

docs/root/modules/ontology/schema.md
jychp: This have been change in the mean time to `active` after a discussion with Alex

cartography/models/ontology/mapping/data/containers.py
```diff
@@ -0,0 +1,90 @@
...
+                OntologyFieldMapping(ontology_field="name", node_field="name"),
+                OntologyFieldMapping(ontology_field="image", node_field="image"),
+                OntologyFieldMapping(
+                    ontology_field="image_digest", node_field="status_image_sha"
+                ),
+                OntologyFieldMapping(
+                    ontology_field="status", node_field="status_state"
+                ),
+                # cpu: Not exposed as a direct field in KubernetesContainer node
+                # memory: Not exposed as a direct field in KubernetesContainer node
```
kunaals, line 48: Huh this would be super useful to have. Unrelated to this PR, let's add these feels to KubernetesContainers

docs/root/modules/ontology/schema.md
```diff
@@ -163,3 +164,48 @@ API keys are used across different cloud providers and SaaS platforms for authen
...
+It generalizes concepts like EC2 Instances, DigitalOcean Droplets, and Scaleway Instances.
+
+| Field | Description |
+|-------|-------------|
+| _ont_id | The unique identifier for the instance. |
+| _ont_name | The name of the instance. |
+| _ont_region | The region or zone where the instance is located. |
+| _ont_public_ip_address | The public IP address of the instance. |
+| _ont_private_ip_address | The private IP address of the instance. |
+| _ont_state | The current state of the instance (e.g., running, stopped). |
```
kunaals, line 186: We use the word "state" here but "status" for containers. Both seem to have similar values (running, suspended, terminated). I think it would be better to align on one word. Also, it would be nice if we standardized the states across compute instances and containers across all clouds. This can be done in a later PR though

kunaals (commented):
One small question state vs status

kunaals (approved):
approving to unblock, scope offline and decided to go with state over status

cartography/models/ontology/mapping/data/containers.py
kunaals: Added here: https://github.com/cartography-cncf/cartography/pull/2095

PR #2076, opened 2025-11-21
feat: ontology for inactive users + fix github mapping + inactive rule

jychp:
### Summary

Fix GitHub user extra label, propagate the `inactive` status to users, and tighten graph rules around inactive / unlinked user accounts. 

### Context

Some GitHub user nodes were missing the extra semantic label because only one of the two `NodeSchema` definitions declared it.
In parallel, the `inactive` status was only present on `UserAccount` nodes, which made it harder to reason about inactive users at the `User` level and to write consistent rules around them.
This PR aligns the schemas and rules so inactive users and unlinked user accounts are handled more consistently across the graph.

### Changes proposed

* **Fix missing extra label on GitHub users:**
  Ensure both GitHub user `NodeSchema` definitions declare the same extra label so GitHub user nodes are consistently tagged.
* **Propagate `inactive` from `UserAccount` to `User`:**
  When loading user data, copy the `inactive` flag from the `UserAccount` node onto the corresponding `User` node.
* **Filter on `inactive` in “unlinked user account” rule:**
  Update the unlinked user account rule to only consider active accounts (ignore accounts marked as inactive).
* **Add a rule for inactive users:**
  Introduce a new rule that surfaces users marked as inactive so they can be monitored / cleaned up explicitly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Base image
-FROM python:3.10.19-slim@sha256:975a1e200a16719060d391eea4ac66ee067d709cc22a32f4ca4737731eae36c0 AS base
+FROM python:3.10.19-slim@sha256:c299e10e0070171113f9a1f109dd05e7e634fa94589b056e0e87bb22b2b382a2 AS base
 # Default to ''. Overridden with a specific version specifier e.g. '==0.98.0' by build args or from GitHub actions.
 ARG VERSION_SPECIFIER
 # the UID and GID to run cartography as
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -4,7 +4,7 @@
 # - This dockerfile will get called with .cache as a volume mount.
 # - The current working directory on the host building this container
 #   is the cartography source tree from github.
-FROM python:3.10.19-slim@sha256:975a1e200a16719060d391eea4ac66ee067d709cc22a32f4ca4737731eae36c0
+FROM python:3.10.19-slim@sha256:c299e10e0070171113f9a1f109dd05e7e634fa94589b056e0e87bb22b2b382a2
 
 # The UID and GID to run cartography as.
 # This needs to match the gid and uid on the host.


cartography/rules/data/rules/inactive_user_active_accounts.py
```diff
@@ -0,0 +1,47 @@
...
+from cartography.rules.spec.model import Rule
+
+# Facts
+_inactive_user_active_accounts_ontology = Fact(
+    id="inactive-user-active-accounts-ontology",
+    name="Active accounts linked to inactive users",
+    description="Finds user accounts that remain active despite being linked to inactive user identities.",
+    cypher_query="""
+    MATCH (u:User)-[:HAS_ACCOUNT]-(a:UserAccount)
+    WHERE COALESCE(u.inactive, False) = True
```
achantavy: This where clause makes my head spin a little bit but I get it. could be good to have a comment but it's fine

docs/root/modules/ontology/schema.md
```diff
@@ -67,6 +67,7 @@ A user often has one or many user accounts.
 | fullname | User's full name. |
 | firstname | User's first name. |
 | lastname | User's last name. |
+| inactive | Boolean indicating if the user is inactive (e.g. disabled in the IDP). |
```
achantavy: I feel like we should normalize instead to the postive form in all cases

jychp:
> I think we should normalize the ontology labels to positive form in all cases, what do you think?

I went with the negative form because this field can legitimately be `null` in some cases (for example with Airbyte or OpenAI). When `inactive` is `null`, the account should still be treated as active so the default falsy behavior works exactly the way we want.

If we used an `active` flag instead, then `active = null` would still evaluate as “active,” which feels counterintuitive and makes the logic harder to reason about.

In this PR I made the rule explicit with a `COALESCE`, but for day-to-day usage of the Pydantic class in Python, it’s cleaner and closer to natural falsy semantics to keep the negative form.

docs/root/modules/ontology/schema.md
jychp: I went with the negative form because this field can legitimately be `null` in some cases (for example with Airbyte or OpenAI). When `inactive` is `null`, the account should still be treated as active so the default falsy behavior works exactly the way we want.

If we used an `active` flag instead, then `active = null` would still evaluate as “active,” which feels counterintuitive and makes the logic harder to reason about.

In this PR I made the rule explicit with a `COALESCE`, but for day-to-day usage of the Pydantic class in Python, it’s cleaner and closer to natural falsy semantics to keep the negative form.

achantavy:
What do you think of these semantics?

Active: True --> user is active 
Active: False --> user is not active
Active: NULL or undefined --> we don't know whether the user is active

Let me know if I missed a usecase or if there's an angle I'm not considering.

The ontology is meant to normalize all the specific types, so I really do think we should normalize the property names.

AGENTS.md
```diff
@@ -1032,304 +992,192 @@ OntologyNodeMapping(
...
-)
+from cartography.rules.data.rules.my_rule import my_rule
+
+RULES = {
+    # ... existing rules
+    my_rule.id: my_rule,
+}
 ```
 
-### Troubleshooting Ontology Integration
```
achantavy, line 1280: All of these things are ok to delete?

jychp: Yes, not so usefull and the AGENTS.md size was to big and can not be read at once by agents.

Issue #2078, opened 2025-11-21
azure: crash in data_factory syncs


achantavy:
### Expected behavior
> Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

data factory sync shouldn't fail

### Actual behavior
> Describe what is actually happening.

Crash

### To Reproduce
> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

Run cartography 0.121.0

### Logs
> If applicable, copy and paste your console log with the failing stack trace.

```
cannot access local variable 'factory_id' where it is not associated with a value
```

### Please complete the following information:
>  - Cartography release version or commit hash [e.g. 0.12.0 or 95e8e11913e2a44a4d4682506d8364a638ceac69]
0.121.0

achantavy:
Fixed in #2079

PR #2079, opened 2025-11-21
fix(azure): #2078 factory_id bug

achantavy:
Adds test coverage and fixes the problem

Created tests to repro the issue.

Before fix:
Screenshot 2025-11-21 at 2 14 56 PM

After fix:
image

### DIFF
--- a/cartography/intel/azure/data_factory_dataset.py
+++ b/cartography/intel/azure/data_factory_dataset.py
@@ -119,9 +119,7 @@ def sync_data_factory_datasets(
         all_transformed_datasets.extend(transformed_datasets)
         datasets_by_factory[factory_id] = transformed_datasets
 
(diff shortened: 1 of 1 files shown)


cartography/intel/azure/data_factory_dataset.py
```diff
@@ -121,9 +119,7 @@ def sync_data_factory_datasets(
         all_transformed_datasets.extend(transformed_datasets)
         datasets_by_factory[factory_id] = transformed_datasets
 
-    load_datasets(
-        neo4j_session, all_transformed_datasets, subscription_id, factory_id, update_tag
-    )
+    load_datasets(neo4j_session, all_transformed_datasets, subscription_id, update_tag)
```
achantavy, line 122: This is the main fix

PR #2100, opened 2025-11-26
fix(ci): Check uv lockfile up to date with pyproject.toml

kunaals:
### Summary
We have no CI checks in place to ensure that the uv.lock file actually represents what's in the pyproject.toml dependencies. This change ensures they are in sync 

### Related issues or links

- https://docs.astral.sh/uv/concepts/projects/sync/

This is what the linter says when they're out of sync
Screenshot 2025-11-26 at 3 22 01 PM


achantavy:
Oh oh maybe we could do this in pre-commit to catch it before CI also

```
repos:
  - repo: https://github.com/astral-sh/uv-pre-commit
    # uv version.
    rev: 0.9.13
    hooks:
      - id: uv-lock
```

https://docs.astral.sh/uv/guides/integration/pre-commit/

jychp:
@kunaals nice call, but as @alex said we should also update the pre-config file

achantavy:
I played around with this earlier with the pre-commit hooks and it blocked one of my commits when I thought my uv.lock was up to date. I'm a bit confused on how this works. Let's do a ride-along/shadow before merging if we can haha

kunaals:
Yeah happy to do an explainer and yes ca update precommit

kunaals:
@achantavy 

> I played around with this earlier with the pre-commit hooks and it blocked one of my commits when I thought my uv.lock was up to date. I'm a bit confused on how this works. Let's do a ride-along/shadow before merging if we can haha

The hook runs uv lock which may regenerate the lockfile with slight differences (version ordering, metadata) especially if your local uv version differs from the hook's version. When that happens, it blocks because there are unstaged changes. The fix is just `git add uv.lock` and commit again. I think this PR is ready to merge

PR #2108, opened 2025-12-04
feat(ontology): add ThirdPartyApp ontology node

jychp:
### Summary

This PR introduces the new `ThirdPartyApp` semantic label into the ontology. It provides a unified representation of external applications authorized or used across multiple identity providers. The implementation includes support for Okta, Entra ID, Google Workspace, and Keycloak.

### Context

Different identity providers expose third-party or integrated applications differently. Until now, ontology nodes describing them were fragmented across providers with inconsistent properties.
This PR consolidates them under a single normalized node type, improving graph consistency, enabling cross-provider queries, and simplifying enrichment in downstream features such as attack-path analysis or SaaS visibility.

### Changes proposed

* Added new ontology node: `ThirdPartyApp`.
* Added ingestion support across four identity providers:
  * Okta
  * Entra ID
  * Google Workspace
  * Keycloak


cartography/models/ontology/mapping/data/thirdpartyapps.py
```diff
@@ -0,0 +1,141 @@
...
+                    required=True,
+                ),
+                OntologyFieldMapping(
+                    ontology_field="name",
+                    node_field="display_name",
+                    required=True,
+                ),
+                # enabled: Not available - Entra applications don't have an enabled field in current schema
+                # native_app: Not available - Application type not currently ingested
+                # protocol: Not available - All Entra applications use OAuth 2.0/OpenID Connect
```
kunaals: Should we have a hardcoded default here? Is that possible?

cartography/models/ontology/mapping/data/thirdpartyapps.py
```diff
@@ -0,0 +1,141 @@
...
+                    node_field="app_id",
+                    required=True,
+                ),
+                OntologyFieldMapping(
+                    ontology_field="name",
+                    node_field="display_name",
+                    required=True,
+                ),
+                # enabled: Not available - Entra applications don't have an enabled field in current schema
+                # native_app: Not available - Application type not currently ingested
```
kunaals, line 91: Worth making an issue to add this field to Entra?

cartography/graph/querybuilder.py
```diff
@@ -85,6 +85,56 @@ def _build_ontology_field_statement_equal_boolean(
...
+        )
+        return None
+
+    # Format the value appropriately for Cypher
+    if isinstance(extra_value, str):
+        formatted_value = f'"{_escape_cypher_string(extra_value)}"'
+    elif isinstance(extra_value, bool):
+        formatted_value = str(extra_value).lower()
+    elif extra_value is None:
+        formatted_value = "null"
```
kunaals: I think this codepath is unreachable because of the `if extra_value is None` on line 114?

cartography/models/ontology/mapping/data/users.py
```diff
@@ -43,6 +43,17 @@
             query="MATCH (u:User)-[:HAS_ACCOUNT]->(:UserAccount)-[:OWNS|HAS]->(k:APIKey) MERGE (u)-[r:OWNS]->(k) ON CREATE SET r.firstseen = timestamp() SET r.lastupdated = $UPDATE_TAG",
             iterative=False,
         ),
+        OntologyRelMapping(
+            __comment__="Link User to ThirdPartyApp",
+            query="MATCH (u:User)-[:HAS_ACCOUNT]->(:UserAccount)-[or:AUTHORIZED|APPLICATION]->(a:ThirdPartyApp) MERGE (u)-[r:AUTHORIZED]->(a) ON CREATE SET r.firstseen = timestamp() SET r.lastupdated = $UPDATE_TAG, r.scopes = coalesce(or.scopes, [])",
```
kunaals: `[or:AUTHORIZED|APPLICATION]`
I'm scared to name this rel "or" given that OR is a reserved word in cypher, maybe we can use something else just to be super super safe?

cartography/models/ontology/mapping/data/thirdpartyapps.py
jychp: Yes ... that's why I added "static_value" special handling, just forget to use it ...

jychp: (in fact I added the field, but forget to remove the comment ...)

cartography/models/ontology/mapping/data/thirdpartyapps.py
jychp: Not sure their is really that info in Entra API

PR #2113, opened 2025-12-05
feat(kube): Add kube service to load balancer relationship

kunaals:
### Summary
It would be cool if cartography could **find all paths from external LB to pods**. Something like:
```cypher
MATCH path = (lb:LoadBalancerV2)-[:*]->(pod:KubernetesPod)
RETURN path
```

This PR addressed one piece of the puzzle: it maps KubernetesService nodes with type load balancer to their respective AWS load balancer (ALB or NLB) using hostnames.

More generally:
1. https://github.com/cartography-cncf/cartography/pull/2064 captures the Kubernetes-side relationship: Ingress -> Service -> Pod
1. Our change captures the cross-cloud relationship: Service -> AWS LoadBalancerV2
1. Next step: There should likely exist KubernetesIngress -> LoadBalancerV2 for Ingress-created ALBs

### Related issues or links

- https://github.com/cartography-cncf/cartography/pull/2064
The above PR adds KubernetesIngress nodes which are at the L7 layer and create ALBs. KubernetesIngress nodes target KubernetesService nodes. 

### Checklist


heryxpc (approved):
LGTM

PR #2130, opened 2025-12-09
fix(aws): ECR Sync Crash for Single-Platform Images

chuck-duplocloud:
### Summary

Fix ECR sync crash when encountering single-platform Docker images.

Cartography's ECR module raises a `ValueError` ("No manifest list found for digest...") when syncing single-platform images. This occurs because `_get_platform_specific_digests()` uses a restrictive `acceptedMediaTypes` filter in `batch_get_image`, and when the image is actually single-platform (not a manifest list), the API returns empty results—causing the function to crash.

Single-platform images are common when using `docker push` directly (typical in CI/CD pipelines) rather than `docker buildx` for multi-arch builds. These are perfectly valid Docker images.

**The fix:** Instead of raising an error when `batch_get_image` returns no images, return empty results so the caller treats it as a regular single-platform image:

```python
if not response.get("images"):
    # Image is not actually a manifest list despite the media type hint
    # Return empty results so caller treats it as a regular image
    logger.debug(
        f"Digest {manifest_list_digest} in repository {repository_name} "
        f"is not a manifest list despite media type hint"
    )
    return [], set()
```

### Related issues or links

- Affects Cartography version 0.122.0 (and likely earlier versions)
- AWS ECR's `describe_images` and `batch_get_image` APIs can be inconsistent about what constitutes a "manifest list"
- Single-platform images occur when using `docker push` directly, older Docker versions, or not explicitly creating multi-arch builds

### Checklist

**Before (crash):**

```
ValueError: No manifest list found for digest sha256:914758fa1c15b12c7dfa8cab15eb53b7bbb5143386911da492b00c73c49eef6f in repository <repo-name>
```

**After (graceful handling):**

```
DEBUG: Digest sha256:914758fa... in repository <repo-name> is not a manifest list despite media type hint
```

*(ECR sync completes successfully, treating the image as a regular single-platform image)*

- [x] N/A - This change does not modify the schema; it fixes error handling logic.

- [x] N/A - This is a bug fix, not a new module.

---

**File changed:** `cartography/intel/aws/ecr.py`  
**Function:** `_get_platform_specific_digests()` (around line 60)

### DIFF
--- a/cartography/intel/gcp/permission_relationships.py
+++ b/cartography/intel/gcp/permission_relationships.py
@@ -253,7 +253,9 @@ def get_resource_ids(
 
 def parse_permission_relationships_file(file_path: str) -> list[dict[str, Any]]:
     try:
-        if not os.path.isabs(file_path):
+        if os.path.isabs(file_path):
+            resolved_file_path = file_path
+        else:
             resolved_file_path = os.path.join(os.getcwd(), file_path)
         with open(resolved_file_path) as f:
             relationship_mapping = yaml.load(f, Loader=yaml.FullLoader)


kunaals:
Thanks for the fix! I'll take a look at this by EOD

kunaals (approved):
Looks good to me. I just merged in master and fixed the lint checks

PR #2134, opened 2025-12-10
fix(gcp): Handle absolute paths in permission_relationships file parsing

kunaals:
### Summary
Fix `UnboundLocalError` in `parse_permission_relationships_file()` when an absolute path is provided.

The function had a bug where `resolved_file_path` was only assigned when the input path was relative (inside the `if not os.path.isabs(file_path):` block). When an absolute path was provided (e.g., from `importlib.resources.files()`), the variable was never set, causing:

```
UnboundLocalError: cannot access local variable 'resolved_file_path' where it is not associated with a value
```

This fix ensures `resolved_file_path` is always assigned regardless of whether the input path is absolute or relative.

### Related issues or links

N/A - Bug discovered while integrating cartography's GCP permission_relationships module.

### Checklist

**Before:**
```python
# When passing an absolute path like '/path/to/gcp_permission_relationships.yaml'
parse_permission_relationships_file('/path/to/file.yaml')
# Raises: UnboundLocalError: cannot access local variable 'resolved_file_path'
```

**After:**
```python
# Now works correctly with absolute paths
parse_permission_relationships_file('/path/to/file.yaml')
# Returns: list of permission relationship mappings
```

### DIFF
--- a/cartography/intel/gcp/permission_relationships.py
+++ b/cartography/intel/gcp/permission_relationships.py
@@ -253,7 +253,9 @@ def get_resource_ids(
 
 def parse_permission_relationships_file(file_path: str) -> list[dict[str, Any]]:
     try:
-        if not os.path.isabs(file_path):
+        if os.path.isabs(file_path):
+            resolved_file_path = file_path
+        else:
             resolved_file_path = os.path.join(os.getcwd(), file_path)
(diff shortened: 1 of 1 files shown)


Daksh1603:
My bad for missing this and thanks for adding!

Issue #2127, opened 2025-12-08
AWS Identity Center ALLOWED_BY only draws association to one role


achantavy:
The AWS Identity Center sync's `(:AWSSSOUser)-[:ALLOWED_BY]-(:AWSRole)` relations are incomplete.

### Actual behavior
> Describe what is actually happening.
In the below example, all 5 users should have admin on all of those roles in all of those accounts pictured, but cartography is only drawing that association to just one of them. The query is 

```cypher
match p2 = (r:AWSRole)--(a:AWSAccount)
where r.name contains 'Administrator'

optional match p3 = (r)-[:ALLOWED_BY]-(sso:AWSSSOUser)

return *
```
<img width="1419" height="584" alt="Image" src="https://github.com/user-attachments/assets/f8e37efb-c2bc-49af-a16e-e9daa5ff0394" />

For a more precise query, 
```cypher
match p2 = (r:AWSRole)--(a:AWSAccount)
where r.name contains 'Administrator'

optional match p3 = (r)-[:ALLOWED_BY]-(sso:AWSSSOUser)

return r.name, sso.user_name
```

here the output is

```
r.name,sso.user_name
AWSReservedSSO_AdministratorAccess_c142268ac35e40cf,shyam
AWSReservedSSO_AdministratorAccess_c142268ac35e40cf,alex
AWSReservedSSO_AdministratorAccess_c142268ac35e40cf,kunaal
AWSReservedSSO_AdministratorAccess_c142268ac35e40cf,zeki
AWSReservedSSO_AdministratorAccess_c142268ac35e40cf,jeremy
AWSReservedSSO_AdministratorAccess_eeda64a425220007,null
AWSReservedSSO_AdministratorAccess_a2db4713f950decd,null
AWSReservedSSO_AdministratorAccess_1345b6218254e280,null
```

### Expected behavior
> Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

We expect the graph to look like this instead:
<img width="1419" height="584" alt="Image" src="https://github.com/user-attachments/assets/2ea3d371-6c8e-4477-af49-4caaf292a9da" />

The output here of the precise query should be
```
r.name,sso.user_name
AWSReservedSSO_AdministratorAccess_c142268ac35e40cf,shyam
AWSReservedSSO_AdministratorAccess_c142268ac35e40cf,alex
[...]

Issue #2120, opened 2025-12-08
AWS Identity Center Permission Set False Positives


achantavy:
## Expected behavior
When a user or group is assigned an AWS Identity Center Permission Set for specific accounts, Cartography should only show that the user has access to those specific accounts - not all accounts where the Permission Set is provisioned.

For example: If a Permission Set "AdministratorAccess" is provisioned to 91 accounts, and a user is assigned that Permission Set for only 3 specific accounts, queries should return only those 3 accounts, not all 91.

## Actual behavior
Cartography currently shows that users have access to ALL accounts where a Permission Set is provisioned, regardless of whether the user was actually assigned that Permission Set for those specific accounts.

This leads to false positives. For instance, a user with AdministratorAccess on 3 accounts is shown as having AdministratorAccess on all 91 accounts where that Permission Set exists.

### Problem 1: Loss of Account Scope in User-to-PermissionSet Relationships

```python
for assignment in user_role_assignments_raw:
    uid = assignment["UserId"]
    perm_set = assignment["PermissionSetArn"]
    user_permission_sets.setdefault(uid, []).append(perm_set)
    # AccountId is dropped here!
```

**What happens:**
1. AWS API `list_account_assignments_for_principal()` correctly returns the three-way relationship: `{UserId, PermissionSetArn, AccountId}`
2. The code aggregates this into `UserId -> [PermissionSetArn]`, **dropping the AccountId**
3. This mapping is used to create `(AWSSSOUser)-[HAS_PERMISSION_SET]->(AWSPermissionSet)` relationships
4. These relationships have no account scope, implying the user has the Permission Set on ALL provisioned accounts

```python
class AWSSSOUserToPermissionSetRel(CartographyRelSchema):
    target_node_label: str = "AWSPermissionSet"
    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
        {"arn": PropertyRef("AssignedPermissionSets", one_to_many=True)},
    )
    direction: LinkDirection = LinkDirection.OUTWARD
    rel_label: str = "HAS_PERMISSION_SET"
```

This creates a direct relationship without any account context.

### Problem 2: Incorrect Role ARN Enrichment - One-to-Many Mapping Loss

**The bug has two parts:**

#### Part 2a: Query doesn't filter by Account
```python
query = """
MATCH (role:AWSRole)<-[:ASSIGNED_TO_ROLE]-(permset:AWSPermissionSet)
WHERE permset.arn IN $PermSetIds
[...]

PR #2136, opened 2025-12-11
fix(aws): identity center data quality issues and doc gaps #2127, #2120:

achantavy:
#### Before

Screenshot 2025-12-11 at 11 44 18 AM

2 users have the AdministratorAccess permissionset but for another account. We incorrectly label them as having administrator for _all_ accounts.

#### After

Screenshot 2025-12-11 at 11 38 55 AM

We correctly handle the per-account mapping.

- Fixes https://github.com/cartography-cncf/cartography/issues/2127 where we incorrectly mapped AWS SSO principals to just one arbitrary role
- Fixes https://github.com/cartography-cncf/cartography/issues/2120 where if a user had a permissionset on one account, we treated it as if they had that permission on _all_ accounts. Also fixed incorrect schema documentation about how we represent AWS SSO permissionsets. Specifically, 
  - Calls out that `ALLOWED_BY` maps role to principal only if the principal is assigned to the associated permissionset _for the given account_.
  - Renames misleading function names that made the code hard to understand
- Fixes bug where AWSSSOUsers and groups did not have a region defined on them.

### Checklist

### DIFF
--- a/cartography/models/aws/identitycenter/awssogroup.py
+++ b/cartography/models/aws/identitycenter/awssogroup.py
@@ -12,7 +12,7 @@ from cartography.models.core.relationships import TargetNodeMatcher
 
 
 @dataclass(frozen=True)
-class SSOGroupProperties(CartographyNodeProperties):
+class AWSSSOGroupProperties(CartographyNodeProperties):
     id: PropertyRef = PropertyRef("GroupId")
     display_name: PropertyRef = PropertyRef("DisplayName")
     description: PropertyRef = PropertyRef("Description")
@@ -61,7 +61,7 @@ class AWSSSOGroupToPermissionSetRel(CartographyRelSchema):
 @dataclass(frozen=True)
 class AWSSSOGroupSchema(CartographyNodeSchema):
     label: str = "AWSSSOGroup"
-    properties: SSOGroupProperties = SSOGroupProperties()
+    properties: AWSSSOGroupProperties = AWSSSOGroupProperties()
     sub_resource_relationship: AWSSSOGroupToAWSAccountRel = AWSSSOGroupToAWSAccountRel()
     other_relationships: OtherRelationships = OtherRelationships(
         [
--- a/cartography/models/aws/identitycenter/awsssouser.py
+++ b/cartography/models/aws/identitycenter/awsssouser.py
@@ -13,7 +13,7 @@ from cartography.models.core.relationships import TargetNodeMatcher
 
 
 @dataclass(frozen=True)
-class SSOUserProperties(CartographyNodeProperties):
+class AWSSSOUserProperties(CartographyNodeProperties):
     id: PropertyRef = PropertyRef("UserId")
     user_name: PropertyRef = PropertyRef("UserName")
     identity_store_id: PropertyRef = PropertyRef("IdentityStoreId")
@@ -23,19 +23,21 @@ class SSOUserProperties(CartographyNodeProperties):
 
 
 @dataclass(frozen=True)
-class SSOUserToOktaUserRelRelProperties(CartographyRelProperties):
+class AWSSSOUserToOktaUserRelRelProperties(CartographyRelProperties):
     lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
 
 
 @dataclass(frozen=True)
-class SSOUserToOktaUserRel(CartographyRelSchema):
+class AWSSSOUserToOktaUserRel(CartographyRelSchema):
     target_node_label: str = "UserAccount"
     target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
         {"id": PropertyRef("ExternalId")},
     )
     direction: LinkDirection = LinkDirection.INWARD
     rel_label: str = "CAN_ASSUME_IDENTITY"
-    properties: SSOUserToOktaUserRelRelProperties = SSOUserToOktaUserRelRelProperties()
+    properties: AWSSSOUserToOktaUserRelRelProperties = (
+        AWSSSOUserToOktaUserRelRelProperties()
+    )
 
 
 @dataclass(frozen=True)
@@ -94,14 +96,14 @@ class AWSSSOUserToPermissionSetRel(CartographyRelSchema):
 @dataclass(frozen=True)
 class AWSSSOUserSchema(CartographyNodeSchema):
     label: str = "AWSSSOUser"
-    properties: SSOUserProperties = SSOUserProperties()
+    properties: AWSSSOUserProperties = AWSSSOUserProperties()
     extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(
         ["UserAccount"]
     )  # UserAccount label is used for ontology mapping
     sub_resource_relationship: AWSSSOUserToAWSAccountRel = AWSSSOUserToAWSAccountRel()
     other_relationships: OtherRelationships = OtherRelationships(
         [
-            SSOUserToOktaUserRel(),
+            AWSSSOUserToOktaUserRel(),
             AWSSSOUserToSSOGroupRel(),
             AWSSSOUserToPermissionSetRel(),
         ],


cartography/intel/aws/identitycenter.py
```diff
@@ -411,66 +443,108 @@ def get_user_group_memberships(
...
+    RETURN permset.arn AS PermissionSetArn, role.arn AS RoleArn, account.id AS AccountId
     """
     permset_to_role = neo4j_session.execute_read(
         read_list_of_dicts_tx,
         query,
         PermSetIds=permset_ids,
     )
-
-    # Create mapping from permission set ARN to role ARN
-    permset_to_role_map = {
```
achantavy, line 437: This is (one of) the main bug(s). A permset can map to more than one role. The way this is written, we map to just one!

cartography/intel/aws/identitycenter.py
```diff
@@ -411,66 +443,108 @@ def get_user_group_memberships(
...
+    } or,
+    a list of  {
+        "GroupId": str,
+        "PermissionSetArn": str,
+        "AccountId": str,
+        "RoleArn": str
+    }
+    """
+    # Get unique permission set ARNs from role assignments
+    permset_ids = list[str]({ra["PermissionSetArn"] for ra in principal_assignments})
```
kunaals:
```suggestion
    permset_ids = list({ra["PermissionSetArn"] for ra in principal_assignments})
```

cartography/intel/aws/identitycenter.py
```diff
@@ -411,66 +443,108 @@ def get_user_group_memberships(
...
-        enriched_assignments.append(
+
+@timeit
+def get_principal_roles(
+    neo4j_session: neo4j.Session,
+    principal_assignments: list[dict[str, Any]],
+) -> list[dict[str, Any]]:
+    """
+    :param neo4j_session: neo4j.Session
+    :param role_assignments: either a list of {
```
kunaals:
```suggestion
    :param principal_assignments: either a list of {
```

cartography/intel/aws/identitycenter.py
kunaals, line 313: in the AWSSSOUser model the `region` field does not have `set_in_kwargs=True` so i think this field will be dropped? unless its in the dict itself (which i dont think it is)

kunaals, line 340: same region comment here, needs `set_in_kwargs=True` in the SSOGroupProperties

docs/root/usage/samplequeries.md
```diff
@@ -174,3 +174,11 @@ MATCH (repo:GitHubRepository)-[edge:REQUIRES]->(dep:Dependency)
 RETURN repo.name, dep.name, edge.specifier, dep.version
 ```
 [test it locally](http://localhost:7474/browser/?preselectAuthMethod=NO_AUTH&db=neo4j&connectURL=bolt://neo4j:neo4j@localhost:7474&cmd=edit&arg=MATCH%20%28repo%3AGitHubRepository%29-%5Bedge%3AREQUIRES%5D-%3E%28dep%3ADependency%29%0ARETURN%20repo.name%2C%20dep.name%2C%20edge.specifier%2C%20dep.version)
+
+### What AWS accounts can a given AWS Identity Center user access? Via what permission sets and roles?
+```cypher
+MATCH (user:AWSSSOUser{user_name:"myuser"})-[:ALLOWED_BY]-(role:AWSRole)<-[:RESOURCE]-(account:AWSAccount)
```
kunaals:
```suggestion
MATCH (user:AWSSSOUser{user_name:"myuser"})<-[:ALLOWED_BY]-(role:AWSRole)<-[:RESOURCE]-(account:AWSAccount)
```

docs/root/usage/samplequeries.md
```diff
@@ -174,3 +174,11 @@ MATCH (repo:GitHubRepository)-[edge:REQUIRES]->(dep:Dependency)
 RETURN repo.name, dep.name, edge.specifier, dep.version
 ```
 [test it locally](http://localhost:7474/browser/?preselectAuthMethod=NO_AUTH&db=neo4j&connectURL=bolt://neo4j:neo4j@localhost:7474&cmd=edit&arg=MATCH%20%28repo%3AGitHubRepository%29-%5Bedge%3AREQUIRES%5D-%3E%28dep%3ADependency%29%0ARETURN%20repo.name%2C%20dep.name%2C%20edge.specifier%2C%20dep.version)
+
+### What AWS accounts can a given AWS Identity Center user access? Via what permission sets and roles?
```
kunaals: thanks for adding this sample query, let's add more if you're open to it? i feel like these are easy to write now after being immersed in the permissionset / role / ssouser world but harder to write later

ideas:
1. all AWS accounts and roles a given SSO user can access
```cypher
MATCH (u:AWSSSOUser {user_name: $user_name})<-[:ALLOWED_BY]-(r:AWSRole)<-[:RESOURCE]-(a:AWSAccount)
RETURN DISTINCT a.id AS account_id, r.arn AS role_arn
```
or using ontology
```cypher
MATCH (ua:UserAccount {id: $external_id})-[:CAN_ASSUME_IDENTITY]->(u:AWSSSOUser)
MATCH (u)<-[:ALLOWED_BY]-(r:AWSRole)<-[:RESOURCE]-(a:AWSAccount)
RETURN DISTINCT a.id AS account_id, r.arn AS role_arn
```

2. all users AND groups that can assume a given AWSRole via Identity Center
(i recognize that the list_account_assignments_for_principal API resolves this for users, but cartography users folks may want to explicitly list groups as well)
```cypher
// Users
MATCH (r:AWSRole {arn: $role_arn})-[:ALLOWED_BY]->(u:AWSSSOUser)
RETURN 'user' AS principal_type,  u.id AS principal_id,  u.user_name AS principal_name

UNION

// Groups
MATCH (r:AWSRole {arn: $role_arn})-[:ALLOWED_BY]->(g:AWSSSOGroup)
RETURN 'group' AS principal_type, g.id AS principal_id, g.display_name AS principal_name
```

3. permission sets provisioned to a given AWS account
“provisioned to an account” == permission sets that have created roles in that account:
```cypher
MATCH (a:AWSAccount {id: $account_id})-[:RESOURCE]->(r:AWSRole)
MATCH (r)<-[:ASSIGNED_TO_ROLE]-(ps:AWSPermissionSet)
RETURN DISTINCT ps.arn  AS permission_set_arn, ps.name AS permission_set_name
```
disclaimer: i have not tested these

cartography/intel/aws/identitycenter.py
```diff
@@ -411,66 +443,108 @@ def get_user_group_memberships(
...
+    for assignment in principal_assignments:
+        lookup_key = (assignment["PermissionSetArn"], assignment["AccountId"])
+        role_arn = permset_to_role.get(lookup_key)
+        principal_roles.append(
             {
                 **assignment,
                 "RoleArn": role_arn,
             }
         )
+    return principal_roles
```
kunaals, line 527:
```suggestion
    unmatched = 0

    for assignment in principal_assignments:
        lookup_key = (assignment["PermissionSetArn"], assignment["AccountId"])
        role_arn = permset_to_role.get(lookup_key)
        if role_arn is None:
            unmatched += 1
        principal_roles.append(
            {
                **assignment, 
                "RoleArn": role_arn
            }
        )

    if unmatched:
        logger.info("Identity Center: %d of %d principal assignments had no matching AWSRole; "
            "this usually means IAM roles for some permission sets/accounts "
            "haven't been ingested yet.",
            unmatched,
            len(principal_assignments),
        )

    return principal_roles
```
this might be overkill, but i think it would be helpful to tell users when we _cannot_ match an assignment to a role. im open to not including this

disclaimer: have not tested this

tests/integration/cartography/intel/aws/test_identitycenter.py
kunaals, line 622:
```suggestion
   MATCH (i:AWSIdentityCenter {id: $instance_arn})-[:HAS_PERMISSION_SET]->(ps:AWSPermissionSet)
   RETURN ps.id AS id
```

might also be worth adding an explicit assertion that there are no ALLOWED_BY edges from AWSRole to AWSSSOUser/AWSSSOGroup in this scenario

cartography/intel/aws/identitycenter.py
achantavy: Good catch, that's another current bug actually, I just noticed that AWSSSOUsers in the graph don't have a region set. I'll update the model.

kunaals (approved):
LGTM now, last nit is rename `SSOGroupProperties` class to `AWSSSOGroupProperties` in the group model

PR #2138, opened 2025-12-11
fix(core): Handle EquivalentSchemaRuleAlreadyExists in parallel sync

kunaals:
### Summary

When running multiple sync operations in parallel (e.g., GCP project syncs), a race condition in Neo4j can cause `EquivalentSchemaRuleAlreadyExists` errors during index creation. 

Even though we use `CREATE INDEX IF NOT EXISTS`, Neo4j has a known race condition where concurrent index creation can fail if another session creates the index between the existence check and the actual creation.

This change catches and ignores these errors in `_run_index_query_with_retry` since they indicate the index already exists, which is the desired state.

**Error example (redacted):**
```
neo4j.exceptions.ClientError: {code: Neo.ClientError.Schema.EquivalentSchemaRuleAlreadyExists} 
{message: An equivalent index already exists, 'Index( id=XXX, name='index_XXXXXXXX', 
type='RANGE', schema=(:GCPPolicyBinding {id}), indexProvider='range-1.0' )'.}
```

### Related issues or links

- Neo4j issue documenting race condition behavior: https://github.com/neo4j/neo4j/issues/6823

### Checklist

**Before:** Parallel GCP syncs fail with `EquivalentSchemaRuleAlreadyExists` error when multiple syncs try to create the same index simultaneously.

**After:** The error is caught and logged at debug level, allowing parallel syncs to continue successfully since the index already exists (desired state).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -21,5 +21,4 @@ If you are changing a node or relationship:
 
 If you are implementing a new intel module:
 - [ ] Use the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).
-
 - [ ] Confirm that the linter actually passes (submitting a PR where the linter fails shows reviewers that you did not test your code and will delay your review).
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -21,5 +21,4 @@ If you are changing a node or relationship:
 
 If you are implementing a new intel module:
 - [ ] Use the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).
-
```
kunaals, line 24: this unintentional newline has been bugging me lol

PR #2172, opened 2025-12-22
feat: Add GitLab and Workday intel modules

sachafaust:
### Summary

This PR adds two new intel modules to Cartography with production-grade quality:

1. **GitLab Module** - Syncs repositories, groups, and programming languages from any GitLab instance
2. **Workday Module** - Syncs employee and organization data from Workday HR system

Both modules have been **live tested at production scale** and include comprehensive documentation, testing, and zero limitations.

### Features

**GitLab Module:**
- Production-tested against a real GitLab instance with thousands of repositories
- Group (namespace) support with OWNER relationships
- Programming language detection with percentage tracking
- Complete language analysis for ALL repos (NO sampling or limits)
- Multi-instance support - URL-prefixed IDs prevent collisions across multiple GitLab instances
- Parallel execution - 10 concurrent workers for language detection
- 48 repository properties - visibility, URLs, stats, features, timestamps
- Shared ProgrammingLanguage nodes with GitHub module

**Workday Module:**
- Production-tested against a real Workday instance with enterprise-scale employee data
- Organization hierarchical structure support
- Complete manager relationship graphs (REPORTS_TO)
- Human label integration - Enables cross-module identity queries with Duo, Okta, etc.
- 17 employee properties - name, email, title, location, team, etc.
- Clean field mapping from Workday API (handles hyphens in field names)

### Related issues or links

N/A - New feature contribution

### Checklist

**Tests:**
- [x] Added 37 comprehensive tests (25 unit + 12 integration)
  - GitLab: 13 unit tests + 6 integration tests
  - Workday: 12 unit tests + 6 integration tests
- [x] All tests passing
- [x] Live production validation completed for both modules

**Schema documentation:**
- [x] Created comprehensive schema documentation:
  - docs/root/modules/gitlab/ (config.md, schema.md, index.md)
  - docs/root/modules/workday/ (config.md, schema.md, index.md)
- [x] Includes Mermaid diagrams, property tables, and sample Cypher queries
- [x] Added to docs/root/usage/schema.md index

**New intel module requirements:**
- [x] Uses NodeSchema data model (CartographyNodeSchema throughout)
- [x] Linter passes (isort, black, flake8, mypy, pyupgrade)
  - All 14 pre-commit hooks passing
  - Zero linting errors or warnings

### Implementation Details

**Code Quality:**
- Modern CartographyNodeSchema pattern (follows latest best practices)
- Complete type annotations (mypy validated)
- Comprehensive docstrings on all functions
[...]

PR #2173, opened 2025-12-22
feat(gitlab): GitLab orgs, groups, dependencies

shyammukund:
## Summary

This PR builds on top of #2172 and adds comprehensive GitLab support with significant architectural improvements based on reviewer feedback.

### Features Added
- **GitLabOrganization** - Top-level groups (the root of your GitLab hierarchy)
- **GitLabGroup** - Nested subgroups within organizations
- **GitLabProject** - Repositories (renamed from `GitLabRepository` to align with [GitLab's official terminology](https://docs.gitlab.com/api/projects/))
- **GitLabBranch** - Branches within projects
- **GitLabDependencyFile** - Manifest files (package.json, requirements.txt, etc.)
- **GitLabDependency** - Dependencies extracted from manifest files
- **Languages** - Stored as a JSON property on GitLabProject (e.g., `{"Python": 65.5, "JavaScript": 34.5}`)

#### 1. Replaced ThreadPoolExecutor with asyncio + httpx
Following the established Entra module pattern, language fetching now uses:
- `httpx.AsyncClient` with `asyncio.Semaphore` for rate limiting
- Single connection pool instead of 10 separate GitLab client instances
- Reference: `cartography/intel/entra/__init__.py`

#### 2. Replaced raw Cypher with standard schema
The original PR used raw Cypher for LANGUAGE relationships. This PR:
- Stores languages as a JSON property on `GitLabProject` instead of separate `GitLabLanguage` nodes
- Follows the GitHub module pattern (languages as property, not separate nodes)
- Eliminates the need for `gitlab_repos_cleanup.json`
- Single `load()` call handles everything

#### 3. Two-tier sub_resource_relationship hierarchy
Implemented proper cleanup scoping with cascade delete:

```
GitLabOrganization (root tenant)
    └── RESOURCE → GitLabGroup
    └── RESOURCE → GitLabProject
                       └── RESOURCE → GitLabBranch
                       └── RESOURCE → GitLabDependencyFile
                       └── REQUIRES → GitLabDependency
```

**Why cascade delete was needed:**

Without cascade delete, when a GitLab project is deleted from GitLab:
1. Project node gets cleaned up (scoped to org)
2. But branches/dependencies (scoped to project) are never cleaned because the project wasn't in the sync
3. Results in orphaned nodes with zero relationships

This is the same problem that exists in GCP (we've seen 740+ orphaned buckets in production in a large environment). The cascade delete solution:
```python
GraphJob.from_node_schema(GitLabProjectSchema(), params, cascade_delete=True)
```

This deletes all children with `RESOURCE` relationships before deleting the stale parent node.

#### 4. Renamed GitLabRepository → GitLabProject
Aligns with [GitLab's official API terminology](https://docs.gitlab.com/api/projects/) where "projects" is the correct term (not "repositories").

#### 5. GitLab-specific language storage (not shared ProgrammingLanguage)
Unlike the original PR which reused GitHub's `ProgrammingLanguage` node, this stores languages as a JSON property on each project. Benefits:
- Simpler schema (no separate node type)
- No cleanup complexity
- Follows GitHub module pattern
- Percentage data preserved: `{"Python": 65.5, "JavaScript": 34.5}`

Sample query to find all languages:
```cypher
MATCH (p:GitLabProject)
WHERE p.languages CONTAINS '"Python"'
RETURN p.name, p.languages
```

#### 6. Multi-instance support via web_url IDs
All node IDs use GitLab's `web_url` which inherently includes the instance URL, supporting multiple GitLab instances in the same Neo4j database without ID collisions:
- `https://gitlab.com/myorg/project`
- `https://gitlab.example.com/myorg/project`

### Configuration
```bash
cartography \
  --gitlab-url "https://gitlab.com" \
  --gitlab-organization-id 12345678 \
  --gitlab-token-env-var "GITLAB_TOKEN"
```

Required PAT scopes: `read_user`, `read_repository`, `read_api`

## Related issues or links

- Builds on #2172 (original GitLab PR)
- GitLab API docs: https://docs.gitlab.com/api/projects/

## Checklist

- [x] Update/add unit or integration tests
- [x] Update the [schema](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) documentation
- [x] Uses the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node)
- [x] Linter passes
Screenshot 2025-12-24 at 6 29 58 PM

### DIFF
--- a/cartography/intel/gitlab/dependencies.py
+++ b/cartography/intel/gitlab/dependencies.py
@@ -187,7 +187,7 @@ def _parse_cyclonedx_sbom(
     Format: https://cyclonedx.org/
 
     GitLab stores the source manifest/lock file in the SBOM metadata as
-    'gitlab:dependency_scanning:input_file'. All dependencies in this SBOM
+    'gitlab:dependency_scanning:input_file:path'. All dependencies in this SBOM
     came from that single file.
 
     :param sbom_data: Parsed JSON from gl-sbom-*.cdx.json file


docs/root/modules/gitlab/schema.md
```diff
@@ -0,0 +1,288 @@
+## GitLab Schema
```
kunaals, line 1: can we also have a config.md that explains how to setup gitlab?

cartography/data/jobs/cleanup/gitlab_post_ingestion_branches_cleanup.json
```diff
@@ -0,0 +1,8 @@
+{
```
kunaals: these custom cleanup jobs seem quite similar to the default ones, can we just use the default ones from the datamodel?

cartography/intel/gitlab/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+
+        # Sync projects (within organizations and groups)
+        cartography.intel.gitlab.projects.sync_gitlab_projects(
+            neo4j_session,
+            gitlab_url,
+            token,
+            config.update_tag,
+            common_job_parameters,
+        )
```
kunaals: lets remove the cleanup from each of these sync functions and do them at the bottom in reverse order (projects then groups)

cartography/intel/gitlab/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        return
+
+    auth_tokens = json.loads(base64.b64decode(config.gitlab_config).decode())
+    common_job_parameters = {
+        "UPDATE_TAG": config.update_tag,
+    }
+
+    # Sync phase: Sync all resources from all GitLab instances
+    for auth_data in auth_tokens["instances"]:
```
kunaals: let's only sync one gitlab org at a time for now. callers can always run multiple cartography syncs with different gitlab orgs to get all their orgs in the graph because we dont cleanup root level orgs

cartography/intel/gitlab/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+            common_job_parameters,
+        )
+
+        # Sync projects (within organizations and groups)
+        cartography.intel.gitlab.projects.sync_gitlab_projects(
+            neo4j_session,
+            gitlab_url,
+            token,
+            config.update_tag,
+            common_job_parameters,
```
kunaals: lets pass in the gitlab org_id via common job parameters because its needed in these syncs but we already have the org_id from the orgs sync. you can make the orgs sync return the org_id

cartography/intel/gitlab/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        "UPDATE_TAG": config.update_tag,
+    }
+
+    # Sync phase: Sync all resources from all GitLab instances
+    for auth_data in auth_tokens["instances"]:
+        gitlab_url = auth_data["url"]
+        token = auth_data["token"]
+
+        # Sync organizations (top-level groups)
+        cartography.intel.gitlab.organizations.sync_gitlab_organizations(
```
kunaals: potentially we could just use an org_id that the caller passes in and just ingest one org. then the args that the gitlab module requires would probably look something like:
```
--gitlab-org org_id
--gitlab-url url (default to https://gitlab.com)
--gitlab-token $TOKEN
```

cartography/intel/gitlab/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        "gitlab_post_ingestion_projects_cleanup.json",
+        neo4j_session,
+        common_job_parameters,
+    )
+
+    run_cleanup_job(
+        "gitlab_post_ingestion_groups_cleanup.json",
+        neo4j_session,
+        common_job_parameters,
+    )
```
kunaals: we can run normal cleanup for sub-resources of **projects** and then here we can just run the clean up for projects and then groups, and lets not run cleanup for orgs, leave that to the caller

cartography/models/gitlab/projects.py
```diff
@@ -0,0 +1,150 @@
...
+        "path_with_namespace", extra_index=True
+    )  # Full path
+    description: PropertyRef = PropertyRef("description")
+    visibility: PropertyRef = PropertyRef("visibility")  # private, internal, public
+    default_branch: PropertyRef = PropertyRef("default_branch")  # Default branch name
+    archived: PropertyRef = PropertyRef("archived")  # Is project archived
+    created_at: PropertyRef = PropertyRef("created_at")
+    last_activity_at: PropertyRef = PropertyRef("last_activity_at")
+    languages: PropertyRef = PropertyRef(
+        "languages", extra_index=True
```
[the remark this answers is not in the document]
kunaals: We're keeping it, we should add the ability to add TEXT indexes in the future after we migrate to Neo4j 5

kunaals:
@sachafaust mind taking a look at this PR? for some reason it wont let me add you as a reviewer. only breaking change in terms of datamodel is now we store languages used by a repo on the repo node itself instead of making a new node for just that field

kunaals:
i mistakenly merged this and revived the code in https://github.com/cartography-cncf/cartography/pull/2181

PR #2179, opened 2025-12-23
feat(ontology): add tenant label into ontology

jychp:
A tenant represents the top-level organizational boundary or billing entity within a cloud provider or SaaS platform.
Tenants serve as the root container for all resources, users, and configurations within a given service.
We add a Tenant semantic label to all nodes that have outward 'RESOURCE' relationships.


tests/unit/cartography/intel/ontology/test_ontology_mapping.py
```diff
@@ -21,6 +21,9 @@
 OLD_FORMAT_NODES = [
     "OktaUser",
     "OktaApplication",
+    "OktaOrganization",
+    "AWSAccount",
+    "EntraTenant",  # A bit specific, main label is AzureTenant
```
kunaals:
```suggestion
    "EntraTenant",  # main label is AzureTenant
```

cartography/models/ontology/mapping/data/tenants.py
```diff
@@ -0,0 +1,327 @@
...
+
+# DigitalOcean
+digitalocean_mapping = OntologyMapping(
+    module_name="digitalocean",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="DOAccount",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="uuid", required=True
```
kunaals, line 90: this is name to uuid, intentional?

cartography/models/ontology/mapping/data/tenants.py
```diff
@@ -0,0 +1,327 @@
...
+
+# Keycloak
+keycloak_mapping = OntologyMapping(
+    module_name="keycloak",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="KeycloakRealm",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="name", required=True
```
kunaals, line 204: is the use of name (realm) instead of displayname intentional here? im not too familiar with keycloak so im not sure if displayname is often null

cartography/models/ontology/mapping/data/tenants.py
```diff
@@ -0,0 +1,327 @@
...
+
+# GCP
+gcp_mapping = OntologyMapping(
+    module_name="gcp",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="GCPOrganization",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="orgname", required=True
```
kunaals: `orgname` is like organizations/123456789 whereas `displayname` is a human readable name, not sure which one you want

kunaals (approved):
approving to unblock, left some nits

cartography/models/ontology/mapping/data/tenants.py
jychp: there is no name, but an id and a uuid, so I use uuid as a name, not ideal but no other solution for now (DigitalOcean should be reworked at some point)

cartography/models/ontology/mapping/data/tenants.py
jychp: name is the slug and much more stable than the displayname

PR #2180, opened 2025-12-24
Fix Neo4j error in Semgrep SCA findings with unknown vulnerability identifiers

AdiLyft:
### Summary
Fixes Neo4j `CypherTypeError` when Semgrep returns vulnerability identifiers that don't match CVE or GHSA patterns.

**Problem:** 
- `_build_vuln_url()` returns `None` for unknown vulnerability formats
- Creates `ref_urls = [None]` which Neo4j rejects

**Solution:**
- Filter out `None` values: `ref_urls = [url] if url else []`
- Unknown vulnerabilities now get empty `ref_urls` instead of `[None]`

**Testing:** 
- Added test case for unknown vulnerability identifier format.

### Checklist


heryxpc:
Included at https://github.com/cartography-cncf/cartography/pull/2187

PR #2181, opened 2025-12-25
feat(graph): Add cascade_delete option to cleanup builder

kunaals:
## Summary

Adds a `cascade_delete` parameter to `build_cleanup_queries()` and `GraphJob.from_node_schema()` that, when enabled, also deletes child nodes that the stale parent points to via RESOURCE relationships.

This is useful for hierarchical data like GCP and GitLab organizations/groups where deleting a stale parent should also clean up its children.

### Query Behavior

In Cartography, RESOURCE relationships point from parent to child: `(Parent)-[:RESOURCE]->(Child)`

When `cascade_delete=True`, the cleanup query becomes:
```python
cascade_rel_label = node_schema.sub_resource_relationship.rel_label
if node_schema.sub_resource_relationship.direction == LinkDirection.INWARD:
    cascade_rel_clause = f"<-[:{cascade_rel_label}]-"
else:
    cascade_rel_clause = f"-[:{cascade_rel_label}]->"
delete_action_clauses = [
    f"""
WHERE n.lastupdated <> $UPDATE_TAG
WITH n LIMIT $LIMIT_SIZE
CALL {{
    WITH n
    OPTIONAL MATCH (n){cascade_rel_clause}(child)
    WITH child WHERE child IS NOT NULL AND child.lastupdated <> $UPDATE_TAG
    DETACH DELETE child
}}
DETACH DELETE n;
""",
]
```
We use a unit subquery to delete children so we don’t have to collect() a potentially huge child list, and to ensure the parent row is always preserved for deletion. An OPTIONAL MATCH + WHERE on the child can unintentionally drop the parent row or force list materialization when fan‑out is large. The subquery does the child deletes as a side effect and then reliably deletes the stale parent. 

When `cascade_delete=False` (default), the standard query is used:
```cypher
WHERE n.lastupdated <> $UPDATE_TAG
WITH n LIMIT $LIMIT_SIZE
DETACH DELETE n;
```

### Backward Compatibility
The default is `False` to maintain backward compatibility - existing behavior is unchanged.

## Test Plan

- [x] integration tests verify actual cascade behavior in Neo4j

## Use Case

This feature is needed for the GitLab module (PR #2182) which has hierarchical data:
- GitLab Organization → Groups → Projects

It will also be needed in the GCP module to fix a bug where project resource nodes get orphaned

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -409,7 +409,9 @@ def transform_users(users: list[dict[str, Any]]) -> list[dict[str, Any]]:
             "name": user["UserName"],
             "path": user["Path"],
             "createdate": str(user["CreateDate"]),
+            "createdate_dt": user["CreateDate"],
             "passwordlastused": str(user.get("PasswordLastUsed", "")),
+            "passwordlastused_dt": user.get("PasswordLastUsed"),
         }
         user_data.append(user_record)
 
@@ -427,6 +429,7 @@ def transform_groups(
             "name": group["GroupName"],
             "path": group["Path"],
             "createdate": str(group["CreateDate"]),
+            "createdate_dt": group["CreateDate"],
             "user_arns": group_memberships.get(group["Arn"], []),
         }
         group_data.append(group_record)
@@ -444,8 +447,10 @@ def transform_access_keys(
                 access_key_record = {
                     "accesskeyid": access_key["AccessKeyId"],
                     "createdate": str(access_key["CreateDate"]),
+                    "createdate_dt": access_key["CreateDate"],
                     "status": access_key["Status"],
                     "lastuseddate": str(access_key.get("LastUsedDate", "")),
+                    "lastuseddate_dt": access_key.get("LastUsedDate"),
                     "lastusedservice": access_key.get("LastUsedService", ""),
                     "lastusedregion": access_key.get("LastUsedRegion", ""),
                     "user_arn": user_arn,  # For the sub-resource relationship
@@ -524,6 +529,7 @@ def transform_role_trust_policies(
             "name": role["RoleName"],
             "path": role["Path"],
             "createdate": str(role["CreateDate"]),
+            "createdate_dt": role["CreateDate"],
             "trusted_aws_principals": list(trusted_aws_principals),
             "account_id": get_account_from_arn(role["Arn"]),
         }
--- a/cartography/models/aws/iam/access_key.py
+++ b/cartography/models/aws/iam/access_key.py
@@ -22,8 +22,10 @@ class AccountAccessKeyNodeProperties(CartographyNodeProperties):
 
     # Business fields from AWS IAM access keys
     createdate: PropertyRef = PropertyRef("createdate")
+    createdate_dt: PropertyRef = PropertyRef("createdate_dt")
     status: PropertyRef = PropertyRef("status")
     lastuseddate: PropertyRef = PropertyRef("lastuseddate")
+    lastuseddate_dt: PropertyRef = PropertyRef("lastuseddate_dt")
     lastusedservice: PropertyRef = PropertyRef("lastusedservice")
     lastusedregion: PropertyRef = PropertyRef("lastusedregion")
 
--- a/cartography/models/aws/iam/group.py
+++ b/cartography/models/aws/iam/group.py
@@ -27,6 +27,7 @@ class AWSGroupNodeProperties(CartographyNodeProperties):
     name: PropertyRef = PropertyRef("name")
     path: PropertyRef = PropertyRef("path")
     createdate: PropertyRef = PropertyRef("createdate")
+    createdate_dt: PropertyRef = PropertyRef("createdate_dt")
 
 
 @dataclass(frozen=True)
--- a/cartography/models/aws/iam/role.py
+++ b/cartography/models/aws/iam/role.py
@@ -26,6 +26,7 @@ class AWSRoleNodeProperties(CartographyNodeProperties):
     name: PropertyRef = PropertyRef("name")
     path: PropertyRef = PropertyRef("path")
     createdate: PropertyRef = PropertyRef("createdate")
+    createdate_dt: PropertyRef = PropertyRef("createdate_dt")
 
 
 @dataclass(frozen=True)
--- a/cartography/models/aws/iam/user.py
+++ b/cartography/models/aws/iam/user.py
@@ -25,7 +25,9 @@ class AWSUserNodeProperties(CartographyNodeProperties):
     name: PropertyRef = PropertyRef("name")
     path: PropertyRef = PropertyRef("path")
     createdate: PropertyRef = PropertyRef("createdate")
+    createdate_dt: PropertyRef = PropertyRef("createdate_dt")
     passwordlastused: PropertyRef = PropertyRef("passwordlastused")
+    passwordlastused_dt: PropertyRef = PropertyRef("passwordlastused_dt")
 
 
 @dataclass(frozen=True)
--- a/cartography/rules/data/rules/__init__.py
+++ b/cartography/rules/data/rules/__init__.py
@@ -1,3 +1,27 @@
+from cartography.rules.data.rules.cis_aws_iam import cis_1_12_unused_credentials
+from cartography.rules.data.rules.cis_aws_iam import cis_1_13_multiple_access_keys
+from cartography.rules.data.rules.cis_aws_iam import cis_1_14_access_key_not_rotated
+from cartography.rules.data.rules.cis_aws_iam import cis_1_15_user_direct_policies
+from cartography.rules.data.rules.cis_aws_iam import cis_1_18_expired_certificates
+from cartography.rules.data.rules.cis_aws_logging import cis_3_1_cloudtrail_multi_region
+from cartography.rules.data.rules.cis_aws_logging import (
+    cis_3_4_cloudtrail_log_validation,
+)
+from cartography.rules.data.rules.cis_aws_logging import cis_3_5_cloudtrail_cloudwatch
+from cartography.rules.data.rules.cis_aws_logging import cis_3_7_cloudtrail_encryption
+from cartography.rules.data.rules.cis_aws_networking import cis_5_1_unrestricted_ssh
+from cartography.rules.data.rules.cis_aws_networking import cis_5_2_unrestricted_rdp
+from cartography.rules.data.rules.cis_aws_networking import cis_5_4_default_sg_traffic
+from cartography.rules.data.rules.cis_aws_networking import unrestricted_all_ports
+from cartography.rules.data.rules.cis_aws_storage import cis_2_1_1_s3_versioning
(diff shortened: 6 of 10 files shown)


cartography/graph/cleanupbuilder.py
```diff
@@ -172,13 +181,27 @@ def _build_cleanup_node_and_rel_queries(
...
+    if cascade_delete:
+        # When cascade_delete is enabled, also delete stale children that have RESOURCE relationships from stale nodes.
+        # We check child.lastupdated to avoid deleting children that were re-parented to a new tenant in the current sync.
+        # In Cartography, RESOURCE relationships point from parent to child: (Parent)-[:RESOURCE]->(Child)
+        delete_action_clauses = [
+            """
+        WHERE n.lastupdated <> $UPDATE_TAG
+        WITH n LIMIT $LIMIT_SIZE
+        OPTIONAL MATCH (n)-[:RESOURCE]->(child)
+        WHERE child.lastupdated <> $UPDATE_TAG
```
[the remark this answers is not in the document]
kunaals: fixed good catch

cartography/graph/cleanupbuilder.py
```diff
@@ -172,13 +189,27 @@ def _build_cleanup_node_and_rel_queries(
...
-        """
+    if cascade_delete:
+        # When cascade_delete is enabled, also delete stale children that have RESOURCE relationships from stale nodes.
+        # We check child.lastupdated to avoid deleting children that were re-parented to a new tenant in the current sync.
+        # In Cartography, RESOURCE relationships point from parent to child: (Parent)-[:RESOURCE]->(Child)
+        delete_action_clauses = [
+            """
+        WHERE n.lastupdated <> $UPDATE_TAG
+        WITH n LIMIT $LIMIT_SIZE
+        OPTIONAL MATCH (n)-[:RESOURCE]->(child)
```
achantavy: The label here is not always `RESOURCE`. Different modules use different labels, e.g. the subresource rel between a Github repo and a dependency is `REQUIRES` ([src](https://github.com/cartography-cncf/cartography/blob/eeb2590d7cab2444b683037bc0f13d2fc2b35fcc/cartography/models/github/dependencies.py#L69)), and  Github repo to a depgraph manifest is `HAS_MANIFEST` ([src](https://github.com/cartography-cncf/cartography/blob/eeb2590d7cab2444b683037bc0f13d2fc2b35fcc/cartography/models/github/manifests.py#L48)).

To be safe, need to use `node_schema.sub_resource_relationship.rel_label` here.

cartography/graph/cleanupbuilder.py
```diff
@@ -172,13 +189,27 @@ def _build_cleanup_node_and_rel_queries(
...
+        WHERE child IS NULL OR child.lastupdated <> $UPDATE_TAG
```
achantavy: What is the `child IS NULL` check for?

docs/root/dev/writing-intel-modules.md
```diff
@@ -446,6 +446,53 @@ synced.
...
+```
+
+**When to use cascade_delete:**
+
+- Use `cascade_delete=True` when child nodes are meaningless without their parent (e.g., GitLab branches without their project)
+- Use `cascade_delete=False` (default) when children should persist independently or when another module manages their lifecycle
+
+**Important notes:**
+
+- Only affects direct children (one level deep via `RESOURCE` relationships). Grandchildren require cleaning up intermediate levels first.
```
achantavy, line 491: Right now the code does not handle grandchildren and there isn't a way to clean up the intermediate levels first, as the query will match on all stale nodes of the given label and look for children regardless of depth. This will leave behind orphaned chains.

For example, if we have a node called `Group` and if the graph looks like:

```cypher
(g1:Group)-[:RESOURCE]->(g2:Group)-[:RESOURCE]->(c1:Child)
```

then when we run 

```python
    GraphJob.from_node_schema(
        Group(),
        {"UPDATE_TAG": UPDATE_TAG, "sub_resource_id": ID},
        cascade_delete=True,
    ).run(neo4j_session)
```

we will run this query:

```cypher
WHERE n.lastupdated <> $UPDATE_TAG
WITH n LIMIT $LIMIT_SIZE
OPTIONAL MATCH (n)-[:RESOURCE]->(child)
WHERE child IS NULL OR child.lastupdated <> $UPDATE_TAG
DETACH DELETE child, n;
```

and there is no guarantee that it will hit `g1` or `g2` first. If `g1` gets hit first, we will delete `g1` and `g2` and leave `c2` orphaned.

kunaals: yeah this was intentional - i didn't want to handle the grandchildren case. we handle the cleanup layer by layer in the sync itself. however, we were missing a tool to actually delete children of a tenant node, this brings that in. but if someone wants to delete grandchildren they'll have to traverse a layer deeper themselves.

cartography/graph/cleanupbuilder.py
kunaals: when we are within the cascade_delete if statement there are 2 cases:
1. there is a tenant node like GitHubRepo which children that are stale
2. there is a tenant node like GitHubRepo with no children

if someone calls the cleanup job on that GitHubRepo node, we want to delete all it's subresources, then the tenant level node itself. the `child.lastupdated <> $UPDATE_TAG` captures the case where children are stale and `child IS NULL` captures the case where that particular GitHubRepo node happens to have no children

cartography/graph/cleanupbuilder.py
kunaals: Ooh good catch

docs/root/dev/writing-intel-modules.md
achantavy: Talked about this on Slack: 

There currently are no situations in cartography where a given node schema is a subresource of itself. This means that we won't run into the example I outlined above.

There are 2 conditions though:

1. As long as subresource relationships are between nodes of different labels, we will be ok.

2. The cleanup operations must happen in order. 


To make this a bit more concrete, say we have

```cypher
(o:Organization)-[:RESOURCE]->(g:Group)-[:RESOURCE]->(c:Child)
```

then if we do a cascade cleanup on `o`, then (assuming that `o` and `g` are stale) `o` and `g` will be deleted, leaving `c` orphaned. 

So to handle this, cleanups of subresources must happen in reverse hierarchical order, specifically from children to parents.

cartography/graph/cleanupbuilder.py
```diff
@@ -172,13 +192,28 @@ def _build_cleanup_node_and_rel_queries(
         )
 
     # The cleanup node query must always be before the cleanup rel query
-    delete_action_clauses = [
-        """
+    if cascade_delete:
+        # When cascade_delete is enabled, also delete stale children that have relationships from stale nodes
+        # matching the sub_resource_relationship rel_label. We check child.lastupdated to avoid deleting children
+        # that were re-parented to a new tenant in the current sync.
+        cascade_rel_label = node_schema.sub_resource_relationship.rel_label
```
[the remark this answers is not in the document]
achantavy: > P1: cascade_delete query can fail to delete stale parent nodes when the parent has only current children because the WHERE clause after OPTIONAL MATCH filters out the parent row. 

Good catch, but this is ok. I'll walk through a concrete example. Say we have

```cypher
(p:Parent)-[:RESOURCE]->(c1,c2,c3:Child)
```

Suppose also that `p.lastupdated` is stale, but all the c1, c2, c3 `lastupdated` values are current.  `p` will not be deleted because of the `WHERE [...] child.lastupdated <> $UPDATE_TAG` clause. This is expected and fine because we don't want to delete parents and create orphans.


> Also, cascade matching always uses -> and ignores sub_resource_relationship.direction, which can make cascade behavior incorrect when the relationship is INWARD.

This is a valid finding though. We do have to consider the relationship direction. We can do this by doing something like this: https://github.com/cartography-cncf/cartography/blob/62f9246b2fb732b28fe3305cb153b4998c1787f0/cartography/graph/querybuilder.py#L526-L537

Leaving out the direction would probably work for 99 out of 100 cases but it would break things in unpredictable hard to trace ways as the code grows. Neo4j performs slightly better with bounded directions too.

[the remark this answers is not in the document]
achantavy: Hmm.. on second thought, the convention is for sub resource rels to always point toward the resource. Let me double check if this is enforced in code or if it's just a convention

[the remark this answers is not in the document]
achantavy: Mm, not guaranteed. I think we should just handle the directionality.

tests/integration/cartography/graph/test_cleanup_cascade.py
```diff
@@ -0,0 +1,192 @@
...
+    # Parent deleted, child-2 (stale) deleted, but child-1 (re-parented) preserved
+    assert check_nodes(neo4j_session, "InterestingAsset", ["id"]) == set()
+    assert check_nodes(neo4j_session, "ChildNode", ["id"]) == {("child-1",)}
+
+
+def test_cascade_delete_removes_parent_with_only_current_children(neo4j_session):
+    """
+    Test that a stale parent is deleted even when all its children are current.
+    This guards against OPTIONAL MATCH filtering out the parent row.
+    """
```
achantavy: Yes, by design, our system deletes parents and leaves behind children if the cleanups are done out of order (therefore our guidance to module authors is to perform cleanups in reverse hierarchical order if there are parent-child sub resource relationships), but it feels weird to have a test to check for this undesirable condition.

Some options:
- we could delete this test altogether
- we could add a comment here explaining the correct order of cleanups

---

Oh, and a small nit on comment:

>     This guards against OPTIONAL MATCH filtering out the parent row.

Since the code no longer uses an `optional match`, this sentence is now confusing and requires someone reading the code to have been following this PR. I would remove this comment.

tests/integration/cartography/graph/test_cleanup_cascade.py
```diff
@@ -115,7 +115,7 @@ def _setup_parent_without_children(neo4j_session, lastupdated: int):
 def test_cascade_delete_works_for_childless_parents(neo4j_session):
     """
     Test cascade_delete=True still deletes parents that have no children.
-    Regression test: OPTIONAL MATCH with WHERE on null child must not filter out the row.
+    Regression test: OPTIONAL MATCH + subquery must not filter out the parent row.
```
achantavy: Nit: remove mention of the query implementation detail in the test comment. The implementation might change over time but the test will continue to check for behaviors

achantavy (approved):
Left some notes about deleting a test but looks good.

As follow-up, should make it super clear that

- avoid multi level sub resources wherever possible
- but if theres no way around it, need to use cascade delete
- and need to do the cascade delete from bottom up

phew

tests/integration/cartography/graph/test_cleanup_cascade.py
kunaals: test removed

tests/integration/cartography/graph/test_cleanup_cascade.py
kunaals: done

Issue #1886, opened 2025-09-08
AWS IAM list_server_certificates


achantavy:
## Description
> Describe your idea.  Please be detailed.  If a feature request, please
> describe the desired behavior, what scenario it enables, and how it
> would be used.

We should add support for `list_server_certificates` in AWS IAM.

## Motivation
> Why is this feature needed? What problem does it solve or opportunity does it unlock?

IAM has its own set of certificates and this auditing these would unlock compliance scenarios.

## Alternatives Considered
> List other approaches or ideas considered, and why they were not chosen.

N/A

## Relevant Links
> Any extra documentation required to understand the issue.

PR #2190, opened 2026-01-04
feat(aws): add IAM certificate sync support

Tushar240503:
Summary
This PR implements support for syncing AWS IAM Server Certificates into Cartography. This allows for auditing of server certificates, tracking their expiration dates, and visualizing their relationship within the security graph.

Key Implementation Details:

New Model: Defined
AWSServerCertificateSchema
using the modern CartographyNodeSchema system.
Efficient Fetching: Implemented "List-Only" fetching in
get_server_certificates
using the IAM paginator to avoid N+1 API call overhead.
Data Transformation: Added logic to convert AWS datetime objects to strings for Neo4j compatibility.
Sync Logic: Integrated into the main IAM
sync()
and
cleanup_iam()
flows to ensure data is updated and stale nodes are removed.

i also added tests for the same

#1886

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -1517,6 +1517,7 @@ def main(argv=None):
     logging.getLogger("neo4j").setLevel(logging.WARNING)
     logging.getLogger("azure.identity").setLevel(logging.WARNING)
     logging.getLogger("httpx").setLevel(logging.WARNING)
+    logging.getLogger("slack_sdk").setLevel(logging.WARNING)
     logging.getLogger("azure.core.pipeline.policies.http_logging_policy").setLevel(
         logging.WARNING
     )
--- a/cartography/data/jobs/analysis/aws_ecs_asset_exposure.json
+++ b/cartography/data/jobs/analysis/aws_ecs_asset_exposure.json
@@ -1,0 +1,16 @@
+{
+  "name": "AWS ECS internet exposure",
+  "statements": [
+    {
+      "__comment": "Clean up any previous exposed_internet attribute on ECSContainer",
+      "query": "MATCH (container:ECSContainer) WHERE container.exposed_internet IS NOT NULL WITH container LIMIT $LIMIT_SIZE REMOVE container.exposed_internet RETURN COUNT(*) AS TotalCompleted",
+      "iterative": true,
+      "iterationsize": 1000
+    },
+    {
+      "__comment": "Mark ECS containers as exposed to the internet when they are behind an internet-facing load balancer",
+      "query": "MATCH (lb:LoadBalancerV2 {exposed_internet: true})-[:EXPOSE]->(ip:EC2PrivateIp)<-[:PRIVATE_IP_ADDRESS]-(ni:NetworkInterface)<-[:NETWORK_INTERFACE]-(task:ECSTask)-[:HAS_CONTAINER]->(container:ECSContainer) SET container.exposed_internet = true",
+      "iterative": false
+    }
+  ]
+}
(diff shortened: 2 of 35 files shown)


cartography/intel/aws/iam.py
```diff
@@ -295,6 +296,16 @@ def get_role_list_data(boto3_session: boto3.Session) -> Dict:
     return {"Roles": roles}
 
 
+@timeit
+def get_server_certificates(boto3_session: boto3.Session) -> List[Dict]:
```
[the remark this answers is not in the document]
Tony-Stark20: This was done to maintain consistency with the other global IAM functions in the file.

cartography/intel/aws/iam.py
```diff
@@ -1105,6 +1116,57 @@ def sync_role_assumptions(
...
+    transformed_certs = []
+    for cert in certificates:
+        transformed_certs.append(
+            {
+                "ServerCertificateName": cert["ServerCertificateName"],
+                "ServerCertificateId": cert["ServerCertificateId"],
+                "Arn": cert["Arn"],
+                "Path": cert["Path"],
+                "Expiration": str(cert["Expiration"]),
+                "UploadDate": str(cert["UploadDate"]),
```
heryxpc: I'm curious, why not keeping the `datetime` format? IIRC these are automatically transformed to Python `datetime` objects and Neo4j driver transforms them to `ZONED DATETIME`.

cartography/models/aws/iam/server_certificate.py
```diff
@@ -0,0 +1,55 @@
...
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class AWSServerCertificateNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("Arn")
+    arn: PropertyRef = PropertyRef("Arn", extra_index=True)
+    server_certificate_id: PropertyRef = PropertyRef(
+        "ServerCertificateId", extra_index=True
```
heryxpc, line 18: Checking the documentation from AWS, `ServerCertificateId` might be a good candidate for a unique ID over the `arn`.

cartography/intel/aws/iam.py
Tony-Stark20: Thanks for the input! I noticed that in the existing AWS EKS ingestion code, datetime fields such as createdAt are converted to strings before being written to Neo4j, 
`transformed_dict["created_at"] = str(transformed_dict["createdAt"])`
So I followed the same pattern here to stay consistent with the current AWS modules.

If the preferred direction going forward is to keep native Python datetime objects so the Neo4j driver stores them as ZONED DATETIME, I’m happy to update this PR accordingly

cartography/models/aws/iam/server_certificate.py
```diff
@@ -0,0 +1,55 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class AWSServerCertificateNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("ServerCertificateId")
```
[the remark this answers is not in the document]
heryxpc: Server certificates are unique across all AWS accounts based on https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-prefixes
All have a prefix `ASCA`.

PR #2196, opened 2026-01-06
feat(rules): GCP CIS benchmarks

kunaals:
## Summary

This PR implements five CIS Google Cloud Platform Foundation Benchmark v4.0.0 compliance rules in Cartography's rules engine. These rules help organizations detect security misconfigurations in their GCP environment that violate CIS best practices.

The rules implemented are:
- **CIS GCP 3.1**: Detects projects with the default VPC network still present (should be deleted)
- **CIS GCP 3.6**: Flags firewall rules allowing SSH (port 22) from the internet (0.0.0.0/0 or ::/0)
- **CIS GCP 3.7**: Flags firewall rules allowing RDP (port 3389) from the internet (0.0.0.0/0 or ::/0)
- **CIS GCP 4.9**: Identifies compute instances with public IP addresses attached
- **CIS GCP 5.2**: Detects Cloud Storage buckets without uniform bucket-level access enabled

Each rule follows the AWS CIS rules patterns with:
- Dedicated output model per rule (e.g., `UnrestrictedSshOutput`, `BucketUniformAccessOutput`)
- Resource-specific field names (e.g., `firewall_id`, `vpc_name`, `bucket_name`)
- STRIDE threat model tags for categorization
- Stable maturity level and version 1.0.0

## Related issues or links

- [CIS Google Cloud Platform Foundation Benchmark v4.0.0](https://www.cisecurity.org/benchmark/google_cloud_computing_platform)

## Checklist

### DIFF
--- a/cartography/rules/data/rules/__init__.py
+++ b/cartography/rules/data/rules/__init__.py
@@ -27,6 +27,12 @@ from cartography.rules.data.rules.cis_aws_storage import cis_2_1_5_s3_access_log
 from cartography.rules.data.rules.cis_aws_storage import cis_2_1_6_s3_encryption
 from cartography.rules.data.rules.cis_aws_storage import cis_2_2_1_rds_encryption
 from cartography.rules.data.rules.cis_aws_storage import cis_2_3_1_ebs_encryption
+from cartography.rules.data.rules.cis_google_workspace import (
+    cis_4_1_1_1_admin_2sv_not_enforced,
+)
+from cartography.rules.data.rules.cis_google_workspace import (
+    cis_4_1_1_3_user_2sv_not_enforced,
+)
 from cartography.rules.data.rules.cloud_security_product_deactivated import (
     cloud_security_product_deactivated,
 )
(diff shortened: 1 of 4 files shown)


cartography/rules/data/rules/cis_4_0_gcp.py
```diff
@@ -0,0 +1,305 @@
...
+CIS_REFERENCES = [
+    RuleReference(
+        text="CIS Google Cloud Platform Foundation Benchmark v4.0.0",
+        url="https://www.cisecurity.org/benchmark/google_cloud_computing_platform",
+    ),
+]
+
+
+class GCPCISFinding(Finding):
+    control_id: str | None = None
```
jychp: When displaying as a list, the first nome null element is used as display name. You should use prioritize resource name and id.

cartography/rules/data/rules/cis_4_0_gcp.py
```diff
@@ -0,0 +1,305 @@
...
+    MATCH (fw)<-[:ALLOWED_BY]-(rule:GCPIpRule)<-[:MEMBER_OF_IP_RULE]-(range:IpRange)
+    WHERE range.range IN ['0.0.0.0/0', '::/0']
+      AND coalesce(fw.disabled, false) = false
+      AND (
+        (rule.protocol = 'tcp' AND rule.fromport <= 3389 AND rule.toport >= 3389)
+        OR rule.protocol = 'all'
+      )
+    RETURN
+        '3.7' AS control_id,
+        'Ensure That RDP Access Is Restricted From the Internet' AS control_title,
```
jychp: Do we really need to return that static metadata? That's generate a lot of overhead and payload size and I think we already have the information in the rule ?

jychp (changes requested):
One important comment on result format and one additional question on the query format.

cartography/rules/data/rules/cis_4_0_gcp.py
kunaals: fixed

cartography/rules/data/rules/cis_4_0_gcp.py
kunaals: fixed

PR #2197, opened 2026-01-06
feat(rules): Google Workspace CIS benchmarks

kunaals:
### Summary

Implements CIS Google Workspace Foundations Benchmark v1.4.0 compliance rules for 2-Step Verification enforcement, following the established AWS CIS rules pattern.

**Rules implemented:**
- **CIS 4.1.1.3**: Users without enforced 2-Step Verification
- **CIS 4.1.1.1**: Admin accounts without enforced 2-Step Verification (privileged access)

Each rule has its own output model with domain-specific fields (`user_id`, `primary_email`, `tenant_id`, etc.) matching the AWS CIS pattern.

**Note:** Requires user security fields ingestion (`admin.directory.user.security` scope). OAuth high-risk scope rule deferred to build on #2159 risk annotations.

### Related issues or links

- [CIS Google Workspace Foundations Benchmark v1.4.0](https://www.cisecurity.org/benchmark/google_workspace)
- #2159 (OAuth risk annotations - will enable OAuth rule)

### Checklist


cartography/rules/data/rules/google_workspace_cis.py
```diff
@@ -0,0 +1,228 @@
...
+from cartography.rules.spec.model import RuleReference
+
+
+class GoogleWorkspaceCISFinding(Finding):
+    """Shared finding model for Google Workspace CIS checks."""
+
+    cis_id: str | None = None
+    control_name: str | None = None
+    severity: str | None = None
+    target_label: str | None = None
```
jychp: When displayed as a row these fields are considered in order to choose the display name. You should prioritize label and email.

cartography/rules/data/rules/google_workspace_cis.py
```diff
@@ -0,0 +1,228 @@
...
+    ),
+    module=Module.GOOGLEWORKSPACE,
+    maturity=Maturity.EXPERIMENTAL,
+    cypher_query="""
+    MATCH (u:GoogleWorkspaceUser)
+    WHERE coalesce(u.is_enforced_in_2_sv, false) = false
+    RETURN
+        u.id AS target_id,
+        u.primary_email AS target_email,
+        'Control 1.1 - Users without enforced 2-Step Verification' AS control_name,
```
jychp: Do we really want to add this static informations ? For large numbers of findings it's a significative overhead. And I feels that we already have this information into the rule object.

cartography/rules/data/rules/google_workspace_cis.py
kunaals: fixed

cartography/rules/data/rules/google_workspace_cis.py
kunaals: fixed

PR #2198, opened 2026-01-06
feat(aws): ingest flattened billing, encryption, and stream properties

### LINKED ISSUE #363: Extend DynamoDB intel module
We should probably extend the DynamoDB intel module to gather more information about tables, as at the moment we only collect the table name.

The DescribeTable API call returns a set of useful information:
* https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DescribeTable.html

Tushar240503:
Summary
Enhanced the DynamoDB intel module to ingest detailed table metadata and improve queryability in Neo4j.

This change moves away from storing complex nested objects and instead flattens DynamoDB table metadata into first-class node properties (following the pattern used in the RDS module). As a result, important attributes such as billing mode, encryption (SSE), streams, archival, and restore information can now be queried directly in Cypher without parsing JSON strings.

Additionally, timestamp fields are now ingested as native Python datetime objects and mapped to Neo4j DateTime types, enabling more powerful time-based queries.

Related issues or links
https://github.com/cartography-cncf/cartography/issues/363
Fixes #363

Screenshot 2026-01-24 at 12 37 10 PM


tests/data/aws/dynamodb.py
```diff
@@ -134,7 +145,7 @@
                     },
                     {
                         "IndexSizeBytes": 33333333,
-                        "IndexName": "sample_index_2",
+                        "IndexName": "sample_index_3",
```
jychp, line 148: The index name was changed from `"sample_index_2"` to `"sample_index_3"` here. Could you confirm this change is intentional?

tests/integration/cartography/intel/aws/test_dynamodb.py
```diff
@@ -207,6 +421,112 @@ def test_load_dynamodb(mock_get_instances, neo4j_session):
...
         (TEST_ACCOUNT_ID, 1234),
     }
+
+
+@patch.object(
+    cartography.intel.aws.dynamodb,
+    "get_dynamodb_tables",
+    return_value=tests.data.aws.dynamodb.LIST_DYNAMODB_TABLES["Tables"],
+)
+def test_load_dynamodb_table_properties(mock_get_instances, neo4j_session):
```
jychp: Nice addition of focused tests! However, these tests (`test_load_dynamodb_table_properties` and `test_load_dynamodb_encryption_properties`) duplicate setup logic with the main test.

Consider if the new properties verification could be added to the existing `test_load_dynamodb` test instead, since they all run the same sync operation. Alternatively, you could use a pytest fixture to reduce setup duplication.

This is a minor suggestion - the current approach is also valid.

cartography/models/aws/dynamodb/tables.py
```diff
@@ -18,15 +18,49 @@ class DynamoDBTableNodeProperties(CartographyNodeProperties):
...
+    )
+
+    # Streams
+    latest_stream_arn: PropertyRef = PropertyRef("LatestStreamArn")
+    latest_stream_label: PropertyRef = PropertyRef("LatestStreamLabel")
+    stream_enabled: PropertyRef = PropertyRef("StreamEnabled")
+    stream_view_type: PropertyRef = PropertyRef("StreamViewType")
+
+    # Encryption
+    sse_status: PropertyRef = PropertyRef("SSEStatus")
```
jychp: Excellent work on the data model! The flattened properties are well-organized with clear grouping comments.

One optional enhancement: For frequently queried security-relevant fields, consider adding `extra_index=True` to support efficient security audit queries:

```python
sse_status: PropertyRef = PropertyRef("SSEStatus", extra_index=True)
```

This would enable efficient queries like:
```cypher
MATCH (t:DynamoDBTable) WHERE t.sse_status IS NULL OR t.sse_status <> 'ENABLED' RETURN t
```

cartography/intel/aws/dynamodb.py
```diff
@@ -40,42 +40,79 @@ def get_dynamodb_tables(
...
 
 @timeit
-def transform_dynamodb_tables(dynamodb_tables: List, region: str) -> Any:
-    ddb_table_data: List[Dict[str, Any]] = []
-    ddb_gsi_data: List[Dict[str, Any]] = []
-
-    for table in dynamodb_tables:
-        ddb_table_data.append(
+def transform_dynamodb_tables(
+    dynamodb_tables: List[Dict],
```
jychp: Consider updating to Python 3.9+ type hints style throughout this file. The project uses modern type hints per AGENTS.md:

```python
# Current (at top of file):
from typing import Dict
from typing import List

# Recommended:
# Remove Dict and List imports; use built-in types instead
```

Then update function signatures:
- `List[Dict]` → `list[dict]`
- `Dict[str, Any]` → `dict[str, Any]`
- `List[Dict[str, Any]]` → `list[dict[str, Any]]`

cartography/models/aws/dynamodb/tables.py
```diff
@@ -18,15 +18,49 @@ class DynamoDBTableNodeProperties(CartographyNodeProperties):
...
+    sse_type: PropertyRef = PropertyRef("SSEType")
+    sse_kms_key_arn: PropertyRef = PropertyRef("SSEKMSKeyArn")
```
jychp: Instead of a property can we make a relationship to a KMSKey entity ?

cartography/models/aws/dynamodb/tables.py
```diff
@@ -18,15 +18,49 @@ class DynamoDBTableNodeProperties(CartographyNodeProperties):
...
     )
 
+    # Billing
+    billing_mode: PropertyRef = PropertyRef("BillingMode")
+    last_update_to_pay_per_request_date_time: PropertyRef = PropertyRef(
+        "LastUpdateToPayPerRequestDateTime",
+    )
+
+    # Streams
+    latest_stream_arn: PropertyRef = PropertyRef("LatestStreamArn")
```
jychp: Instead of a property can we make a relationship to an entity ?

cartography/models/aws/dynamodb/tables.py
```diff
@@ -18,15 +18,49 @@ class DynamoDBTableNodeProperties(CartographyNodeProperties):
...
+    stream_enabled: PropertyRef = PropertyRef("StreamEnabled")
+    stream_view_type: PropertyRef = PropertyRef("StreamViewType")
+
+    # Encryption
+    sse_status: PropertyRef = PropertyRef("SSEStatus")
+    sse_type: PropertyRef = PropertyRef("SSEType")
+    sse_kms_key_arn: PropertyRef = PropertyRef("SSEKMSKeyArn")
+
+    # Archival
+    archival_backup_arn: PropertyRef = PropertyRef("ArchivalBackupArn")
```
jychp: Instead of a property can we make a relationship to an entity ?

cartography/models/aws/dynamodb/tables.py
```diff
@@ -18,15 +18,49 @@ class DynamoDBTableNodeProperties(CartographyNodeProperties):
...
+    archival_date_time: PropertyRef = PropertyRef("ArchivalDateTime")
+    archival_reason: PropertyRef = PropertyRef("ArchivalReason")
+
+    # Restore
+    restore_date_time: PropertyRef = PropertyRef("RestoreDateTime")
+    restore_in_progress: PropertyRef = PropertyRef("RestoreInProgress")
+    source_backup_arn: PropertyRef = PropertyRef("SourceBackupArn")
```
jychp: Instead of a property can we make a relationship to an entity ?

tests/data/aws/dynamodb.py
Tony-Stark20: Since the Index ARN is
`"arn:aws:dynamodb:us-east-1:table/sample-table/index/sample_3-index",`
I updated the index name accordingly from "IndexName": "sample_index_2" to "IndexName": "sample_index_3" so that it matches the ARN

kunaals:
@Tushar240503 mind updating the PR description to follow our template, and include a screenshot of the graph after you've run a sync (please redact any sensitive data!)

Tony-Stark20:
> @Tushar240503 mind updating the PR description to follow our template, and include a screenshot of the graph after you've run a sync (please redact any sensitive data!)

@kunaals Done

cartography/intel/aws/dynamodb.py
```diff
@@ -40,54 +44,170 @@ def get_dynamodb_tables(
...
+        if stream_arn:
+            stream_nodes.append(
+                {
+                    "Arn": stream_arn,
+                    "TableArn": table_arn,
+                    "StreamLabel": table_data.get("LatestStreamLabel"),
+                    "StreamEnabled": stream_spec.get("StreamEnabled"),
+                    "StreamViewType": stream_spec.get("StreamViewType"),
+                }
+            )
```
jychp, line 138: **Suggestion:** For tables that have a `LatestStreamArn` but no `StreamSpecification`, stream nodes will be created with `None` values for `StreamEnabled` and `StreamViewType` (as seen in the test data for `sample-table`).

Consider whether creating stream nodes in this case is desirable. The current behavior is technically correct, but you could add a condition to only create stream nodes when `StreamSpecification` exists:

```python
if stream_arn and stream_spec:
    stream_nodes.append(...)
```

This is not blocking - the current approach provides a consistent view of all streams.

cartography/models/aws/dynamodb/sse.py
```diff
@@ -0,0 +1,95 @@
...
+
+    target_node_label: str = "KMSKey"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"arn": PropertyRef("KMSMasterKeyArn")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "USES_KMS_KEY"
+    properties: DynamoDBSSEDescriptionToKMSKeyRelProperties = (
+        DynamoDBSSEDescriptionToKMSKeyRelProperties()
+    )
```
jychp, line 78: We recently introduce a OrderedDict and constraints in the aws __init__ file, can you have a look on it and add a constraint to be sure KMS is always synced before dynamodb tables ?

Tony-Stark20: Updated

cartography/intel/aws/dynamodb.py
Tony-Stark20: Updated

cartography/intel/aws/resources.py
```diff
@@ -75,6 +75,7 @@
         "iam": iam.sync,
         "iaminstanceprofiles": sync_iam_instance_profiles,
         "s3": s3.sync,
+        "kms": kms.sync,
         "dynamodb": dynamodb.sync,
```
jychp, line 81: Can you add a comment to explain we need kms before dynamodb (see lines after) to avoid ordering change in the future ?

jychp (changes requested):
One last chance and we will be able to ship it.

Thx!

PR #2200, opened 2026-01-07
feat(gcp): Support for GCP Cloud SQL 

shyammukund:
- This PR takes the work done in a previous PR by @janithashri and adds coverage for Backup Configurations in Cloud SQL. 

- https://github.com/cartography-cncf/cartography/pull/2038

### Checklist

image

### DIFF
--- a/cartography/intel/aws/ec2/tgw.py
+++ b/cartography/intel/aws/ec2/tgw.py
@@ -101,7 +101,7 @@ def load_transit_gateways(
     ingest_transit_gateway = """
     MERGE (ownerAccount:AWSAccount {id: $OwnerId})
     ON CREATE SET ownerAccount.firstseen = timestamp()
-    SET ownerAccount.lastupdated = $update_tag
+    SET ownerAccount.lastupdated = $update_tag, ownerAccount :Tenant
 
     MERGE (tgw:AWSTransitGateway {id: $ARN})
     ON CREATE SET tgw.firstseen = timestamp(), tgw.arn = $ARN
--- a/cartography/intel/aws/organizations.py
+++ b/cartography/intel/aws/organizations.py
@@ -110,7 +110,7 @@ def load_aws_accounts(
     query = """
     MERGE (aa:AWSAccount{id: $ACCOUNT_ID})
     ON CREATE SET aa.firstseen = timestamp()
-    SET aa.lastupdated = $aws_update_tag, aa.name = $ACCOUNT_NAME, aa.inscope=true
+    SET aa.lastupdated = $aws_update_tag, aa.name = $ACCOUNT_NAME, aa.inscope=true, aa :Tenant
     REMOVE aa.foreign
     """
     for account_name, account_id in aws_accounts.items():
--- a/cartography/intel/okta/organization.py
+++ b/cartography/intel/okta/organization.py
@@ -24,8 +24,8 @@ def create_okta_organization(
     """
     ingest = """
     MERGE (org:OktaOrganization{id: $ORG_NAME})
-    ON CREATE SET org.name = org.id, org.firstseen = timestamp()
-    SET org.lastupdated = $okta_update_tag
+    ON CREATE SET org.name = org.id, org.firstseen = timestamp(), org._ont_name = org.id
+    SET org.lastupdated = $okta_update_tag, org :Tenant
     """
 
     run_write_query(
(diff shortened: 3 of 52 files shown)


tests/data/gcp/cloud_sql.py
```diff
@@ -0,0 +1,67 @@
+MOCK_INSTANCES = {
+    "items": [
+        {
+            "selfLink": "projects/test-project/instances/carto-sql-test-instance",
```
kunaals: the `selfLink` format here doesn't match what the real cloud sql api returns. i verified against the actual api in project `test-cloudsql-project-483520` and the real format is a full url:

```
https://sqladmin.googleapis.com/sql/v1beta4/projects/test-cloudsql-project-483520/instances/free-trial-first-project
```

this mock uses a partial path which will cause all relationship matchers to fail in production since the instance id (derived from selfLink) won't match.

suggested fix:
```python
"selfLink": "https://sqladmin.googleapis.com/sql/v1beta4/projects/test-project/instances/carto-sql-test-instance",
```

tests/integration/cartography/intel/gcp/test_cloud_sql.py
```diff
@@ -0,0 +1,205 @@
...
+from tests.data.gcp.cloud_sql import MOCK_DATABASES
+from tests.data.gcp.cloud_sql import MOCK_INSTANCES
+from tests.data.gcp.cloud_sql import MOCK_USERS
+from tests.integration.util import check_nodes
+from tests.integration.util import check_rels
+
+TEST_UPDATE_TAG = 123456789
+TEST_PROJECT_ID = "test-project"
+TEST_INSTANCE_NAME = "carto-sql-test-instance"
+TEST_INSTANCE_ID = f"projects/{TEST_PROJECT_ID}/instances/{TEST_INSTANCE_NAME}"
```
kunaals: this needs to be updated to match the real selfLink format (full url). see comment on mock data file.

suggested fix:
```python
TEST_INSTANCE_ID = f"https://sqladmin.googleapis.com/sql/v1beta4/projects/{TEST_PROJECT_ID}/instances/{TEST_INSTANCE_NAME}"
```

tests/data/gcp/cloud_sql.py
```diff
@@ -0,0 +1,67 @@
...
+                    "enabled": True,
+                    "startTime": "03:00",
+                    "pointInTimeRecoveryEnabled": True,
+                    "transactionLogRetentionDays": 7,
+                    "backupRetentionSettings": {
+                        "retainedBackups": 30,
+                    },
+                },
+                "ipConfiguration": {
+                    "privateNetwork": "projects/test-project/global/networks/carto-sql-vpc",
```
kunaals: the official docs show `privateNetwork` with a leading slash: `/projects/myProject/global/networks/default`. i couldn't verify this against the real api since the test instance uses public ip, but there's a risk of format mismatch with `GCPVpc` nodes which use `projects/...` (no leading slash) as their id.

could you verify this against an instance with private ip configured? if the api does return with a leading slash, we'll need to normalize it in the transform.

cartography/intel/gcp/cloud_sql_instance.py
```diff
@@ -0,0 +1,140 @@
...
+                "backendType": inst.get("backendType"),
+                "service_account_email": inst.get("serviceAccountEmailAddress"),
+                "connectionName": inst.get("connectionName"),
+                "tier": settings.get("tier"),
+                "disk_size_gb": settings.get("dataDiskSizeGb"),
+                "disk_type": settings.get("dataDiskType"),
+                "availability_type": settings.get("availabilityType"),
+                "backup_enabled": backup_config.get("enabled"),
+                "require_ssl": ip_config.get("requireSsl"),
+                "network_id": ip_config.get("privateNetwork"),
```
kunaals: (related to above) consider adding defensive normalization here to handle potential leading slash inconsistency:

```python
network = ip_config.get("privateNetwork")
"network_id": network.lstrip("/") if network else None,
```

this ensures the relationship to `GCPVpc` nodes will match regardless of whether the api returns `/projects/...` or `projects/...`.

tests/integration/cartography/intel/gcp/test_cloud_sql.py
```diff
@@ -0,0 +1,205 @@
...
+TEST_VPC_ID = f"projects/{TEST_PROJECT_ID}/global/networks/carto-sql-vpc"
```
kunaals, line 18: if we confirm the privateNetwork format issue above, this may also need updating to match.

shyammukund: Made it so that we convert the Cloud SQL API format which has the slash to match the GCP VCP format which doesn't. We can keep the existing VPC id as the final id shouldn't have a slash.

kunaals (approved):

PR #2203, opened 2026-01-07
fix: AWS and GCP Schema Sidebar 

shyammukund:
This PR just modifies the docs so that GCP Vertex and CloudSQL appear under the GCP Schema instead of having their own tab on the side bar. Also makes sure AWS BedRock and Sagemaker appear under AWS Schema instead of having own tab on sidebar. 

Before: 
image

After:
image

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,5 +5,6 @@
 ```bash
 uv sync --group doc
 uv run sphinx-autobuild docs/root docs/generated/docs -c docs --port 8000
+```
 
 Then visit http://localhost:8000. Changes to files in `docs/root/` will automatically trigger a rebuild.


docs/README.md
```diff
@@ -0,0 +1,30 @@
...
+For local development with auto-reload:
+
+```bash
+uv sync --group doc
+cd docs
+# First sync files from root/ to generated/rst/
+rsync -av root/ conf.py generated/rst/
+# Start the auto-build server, watching root/ for changes
+uv run sphinx-autobuild generated/rst generated/docs --watch root/ --pre-build "rsync -av root/ conf.py generated/rst/" --port 8000
+```
```
kunaals: for me i didnt have to rsync or any of that, can we just do:

```suggestion
```bash
uv sync --group doc
uv run sphinx-autobuild docs/root docs/generated/docs -c docs --port 8000
```
```
or does that not work in your env?

PR #2204, opened 2026-01-08
fix(security): Add pagination safety limits to prevent infinite loops

kunaals:
### Summary
Prevents unbounded pagination from becoming a DoS vector when upstream APIs misbehave, limiting CPU/memory growth and graph pollution while keeping defaults high and configurable for large tenants. Adds consistent logging when limits are hit and fixes pagination edge cases (Snipe-IT URL mutation, Okta next-link handling) to avoid infinite loops.

- N/A

### Checklist

- [x] Update/add unit or integration tests. (Ran `uv run --frozen pytest -vvv --cov-report term-missing --cov=cartography tests/unit`)


jychp:
I’m not really sold on this approach. It’s simple, but still pretty limited:
	•	it relies on a setting shared across all providers
	•	it requires extra configuration

Could we imagine something like:
	•	keep a set of all object IDs we’ve already seen (memory overhead should be acceptable, and we could even cap the size to ~2x items_per_page)
	•	break out of the loop if all returned items have already been seen

Feels like this would fit nicely with a decorator like:
`@pagination_safeguard(id_field="customId", per_page: int | None = 100)`

Issue #2205, opened 2026-01-08
Python 3.10 reaches end-of-life in October 2026


kunaals:
Python 3.10 is scheduled to reach end-of-life (EOL) in **October 2026**. We uses Python 3.10 exclusively across:

- GitHub Actions CI workflows
- Docker images
- Dev env

References

- [Python EOL Schedule - endoflife.date](https://endoflife.date/python)
- [Status of Python versions - Python Dev Guide](https://devguide.python.org/versions/)

PR #2206, opened 2026-01-08
chore: add .python-version file to pin Python 3.10 for development

kunaals:
### Summary

Adds a `.python-version` file pinned to Python 3.10 to ensure all developers use the same Python version that matches CI and Docker.

**Changes:**
- Add `.python-version` file with `3.10`
- Remove `.python-version` from `.gitignore` so it can be tracked

This helps avoid issues like `pymsalruntime` not having wheels for Python 3.13, which causes `uv sync` to fail when developers use newer Python versions.

### Related issues or links

- https://github.com/cartography-cncf/cartography/issues/2205

### Checklist

**Before (Python 3.13):**
```
$ uv sync
Using CPython 3.13.2
Creating virtual environment at: .venv
Resolved 214 packages in 4ms
error: Distribution `pymsalruntime==0.17.1 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're using CPython 3.13 (`cp313`), but `pymsalruntime` (v0.17.1) only has wheels with the following Python ABI tags: `cp310`, `cp311`, `cp312`
```

**After (with `.python-version` pinning 3.10):**
```
$ uv sync
Using CPython 3.10
# Works successfully
```


achantavy (approved):
Dev can also do `uv python pin` but it's good to be nice out of the box here and avoid the rough edge in the dev exp

PR #2209, opened 2026-01-08
feat(gcp): Support for GCP CloudRun

shyammukund:
- Support for CloudRun including Services, Jobs, Executions, and Revisions

### Checklist

image

### DIFF
--- a/cartography/intel/gcp/cloudrun/execution.py
+++ b/cartography/intel/gcp/cloudrun/execution.py
@@ -32,7 +32,7 @@ def get_executions(
     try:
         # Determine which locations to query
         if location == "-":
-            # Discover locations by listing services (Cloud Run v2 API workaround)
+            # Discover all Cloud Run locations for this project
             locations = discover_cloud_run_locations(client, project_id)
         else:
             # Query specific location
--- a/cartography/intel/gcp/cloudrun/revision.py
+++ b/cartography/intel/gcp/cloudrun/revision.py
@@ -110,7 +110,7 @@ def transform_revisions(revisions_data: list[dict], project_id: str) -> list[dic
         # Get service account email (v2 API has serviceAccount at top level)
         service_account_email = revision.get("serviceAccount")
 
-        # Construct log URI - Cloud Console logs viewer URL
+        # Get log URI directly from API response
         log_uri = revision.get("logUri")
 
         transformed.append(
--- a/cartography/intel/gcp/cloudrun/util.py
+++ b/cartography/intel/gcp/cloudrun/util.py
@@ -5,6 +5,7 @@ Utility functions for GCP Cloud Run intel module.
 import logging
 
 from googleapiclient.discovery import Resource
+from googleapiclient.errors import HttpError
 
 logger = logging.getLogger(__name__)
 
@@ -18,7 +19,7 @@ def discover_cloud_run_locations(client: Resource, project_id: str) -> set[str]:
     Falls back to discovering via services list if the v1 API call fails.
     """
     try:
-        # Try using v1 API's locations.list() - this is more reliable
+        # Use v1 API's locations.list() to get all Cloud Run regions
         from cartography.intel.gcp.clients import build_client
         from cartography.intel.gcp.clients import get_gcp_credentials
 
@@ -56,13 +57,16 @@ def discover_cloud_run_locations(client: Resource, project_id: str) -> set[str]:
                 "v1 API returned no locations, falling back to service-based discovery"
             )
 
-    except Exception as e:
+    except HttpError as e:
+        # Only fall back for HTTP/API errors (e.g., API not enabled, 404, etc.)
+        # Auth errors (DefaultCredentialsError, RefreshError) will propagate
+        # since the fallback would also fail with the same auth issue
         logger.warning(
             f"Could not discover locations via v1 API: {e}. "
             f"Falling back to discovery via services list.",
         )
 
-    # Fallback: discover from services (current implementation)
+    # Fallback: discover locations by extracting them from service resource names
     logger.debug("Using service-based discovery for Cloud Run locations")
     services_parent = f"projects/{project_id}/locations/-"
     services_request = (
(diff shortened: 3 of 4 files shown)


cartography/models/gcp/cloudrun/service.py
```diff
@@ -0,0 +1,49 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class GCPCloudRunServiceProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id")
```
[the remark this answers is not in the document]
kunaals: id has a default index, no need for the extra index flag

cartography/intel/gcp/cloudrun/service.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Extract location and short name from the full resource name
+        name_match = re.match(
+            r"projects/[^/]+/locations/([^/]+)/services/([^/]+)",
+            full_name,
+        )
+        location = name_match.group(1) if name_match else None
+        short_name = name_match.group(2) if name_match else None
+
+        # Get latest ready revision - this is a short name, convert to full resource name
```
kunaals: this comment is incorrect - the v2 api returns `latestReadyRevision` as the full resource name already (e.g., `projects/x/locations/y/services/z/revisions/r`), not a short name. the code then wraps it again, creating a broken double-nested path like `projects/.../revisions/projects/.../revisions/...`

just use the value directly:
```python
latest_ready_revision = service.get("latestReadyRevision")
```

cartography/intel/gcp/cloudrun/revision.py
```diff
@@ -0,0 +1,187 @@
...
+        location = name_match.group(1) if name_match else None
+        service_short_name = name_match.group(2) if name_match else None
+        short_name = name_match.group(3) if name_match else None
+
+        # Construct the full service resource name
+        service_full_name = None
+        if location and service_short_name:
+            service_full_name = f"projects/{project_id}/locations/{location}/services/{service_short_name}"
+
+        # Get container image from template.containers[0].image
```
kunaals: v2 revision resources have `containers` and `serviceAccount` at the **top level**, not nested under `template`. this will always return None for both fields.

should be:
```python
containers = revision.get("containers", [])
if containers:
    container_image = containers[0].get("image")

service_account_email = revision.get("serviceAccount")
```

verified against actual api response from `run.googleapis.com/v2`

cartography/intel/gcp/cloudrun/revision.py
```diff
@@ -0,0 +1,187 @@
...
+        container_image = None
+        template = revision.get("template", {})
+        containers = template.get("containers", [])
+        if containers and len(containers) > 0:
+            container_image = containers[0].get("image")
+
+        # Get service account email from template.serviceAccount
+        service_account_email = template.get("serviceAccount")
+
+        # Construct log URI - Cloud Console logs viewer URL
```
kunaals: the api already returns `logUri` directly on the revision resource - no need to construct it manually (and the constructed url format might drift from what google uses).

just use:
```python
log_uri = revision.get("logUri")
```

cartography/intel/gcp/cloudrun/execution.py
```diff
@@ -0,0 +1,193 @@
...
+        short_name = name_match.group(3) if name_match else None
+
+        # Construct the full job resource name
+        job_full_name = None
+        if location and job_short_name:
+            job_full_name = (
+                f"projects/{project_id}/locations/{location}/jobs/{job_short_name}"
+            )
+
+        # Get status - completion condition
```
kunaals: `completionStatus` doesn't exist on the full Execution resource - it only appears in `ExecutionReference` (the embedded summary in Job responses). this will always be None.

you could either:
- remove this field since you already capture the count fields
- derive status from the conditions array
- leave as-is and document it's always None

cartography/intel/gcp/cloudrun/util.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+
+import logging
+
+from googleapiclient.discovery import Resource
+
+logger = logging.getLogger(__name__)
+
+
+def discover_cloud_run_locations(client: Resource, project_id: str) -> set[str]:
```
kunaals, line 13: this approach will miss cloud run jobs in regions that have no services. if someone deploys only jobs (no services) in a region, those jobs and their executions will be silently skipped.

the v1 api has a `locations.list` endpoint that works:
```
run.googleapis.com/v1/projects/{project}/locations
```

consider using that instead, or at minimum document this limitation.

tests/data/gcp/cloudrun.py
```diff
@@ -0,0 +1,70 @@
...
+        },
+    ],
+}
+
+MOCK_REVISIONS = {
+    "revisions": [
+        {
+            "name": "projects/test-project/locations/us-central1/services/test-service/revisions/test-service-00001-abc",
+            "service": "projects/test-project/locations/us-central1/services/test-service",
+            "template": {
```
kunaals: this mock structure doesn't match the real v2 api. revisions have `containers` and `serviceAccount` at the top level, not under `template`.

should be:
```python
{
    "name": "projects/.../revisions/test-service-00001-abc",
    "service": "test-service",
    "containers": [
        {"image": "gcr.io/test-project/test-image:latest"}
    ],
    "serviceAccount": "test-sa@test-project.iam.gserviceaccount.com",
    "logUri": "https://console.cloud.google.com/logs/..."
}
```

also note: `service` field in v2 is just the short name, not the full resource path

cartography/intel/gcp/cloudrun/util.py
kunaals: or if the newer api has something we could use that

shyammukund: I've updated the discover_cloud_run_locations function in util.py to:                                                                                             
                                                                                                                                                                          
  1. First try the v1 API's locations.list() - This will return all Cloud Run regions, regardless of whether they have services, jobs, or both                            
  2. Fall back to service-based discovery if the v1 API call fails for any reason        
  
  T

PR #2213, opened 2026-01-10
docs: Make uv the preferred installation method

### LINKED ISSUE #1940: Update install.md native install docs with uv instructions
We use uv now. Need to include that there instead of the normal pip stuff.

Might even want to consider making this option 1 because (at least for me) with uv installs taking so fast I barely ever use docker-compose anymore.

Side note: This is _not_ a good-first-issue.

Prachi194agrawal:
## Description
Updates the installation documentation to make `uv` the preferred and fastest installation method.

Resolves #1940

## Changes Made
- ✨ **New Option 1**: Native install with uv (marked as fastest method)
  - Added uv installation instructions
  - Documented both `uv sync` (using lockfile) and `uv pip install cartography`
  - All commands use `uv run` (e.g., `uv run cartography`, `uv run cartography-rules`)
  
- 🔄 **Reordered installation options** to prioritize uv:
  - Option 1: Native install with uv (fastest) ⭐ NEW
  - Option 2: Run docker-compose (previously Option 1)
  - Option 3: Manually run 2 containers (previously Option 2)
  - Option 4: Native install with pip (previously Option 3)

- 📝 **Updated Option 4** (pip method) to include a note recommending uv for better performance

- 🔗 Added cross-references between options for better navigation

## Why This Change?
As mentioned in #1940, the project now uses `uv` for package management. With uv's significantly faster installation times (often 10-100x faster than pip), it makes sense to recommend it as the primary installation method. For many developers, uv is now fast enough that it's preferred over docker-compose.

## Testing
- [x] Verified all markdown syntax is correct
- [x] Checked that all internal links and references work
- [x] Confirmed all code blocks use proper bash syntax highlighting
- [x] Ensured all uv commands follow best practices from uv documentation

## Screenshots
The documentation now clearly presents uv as Option 1 with comprehensive setup instructions, making it easy for new users to get started with the fastest installation method.

### DIFF
--- a/docs/root/install.md
+++ b/docs/root/install.md
@@ -2,9 +2,121 @@
 
 Time to set up a test machine to run Cartography.
 
-## Option 1: Run docker-compose (preferred)
+## Option 1: Native install with uv (fastest)
 
-This is the quickest way to get started: you will run a cartography cron job and a Neo4j docker container on your
+This is the fastest way to get started if you already have Python 3.10+ and Neo4j running. Using [uv](https://docs.astral.sh/uv/), a fast Python package installer, you can install and run Cartography natively on your machine.
+
+![yourowntestmachine.png](images/yourowntestmachine.png)
+
+1. **Ensure that you have Python 3.10 set up on your machine.**
+
+    Older or newer versions of Python may work but are not explicitly supported. You will probably have more luck with newer versions.
+
+1. **Install [uv](https://docs.astral.sh/uv/)** if you don't already have it:
+
+    ```bash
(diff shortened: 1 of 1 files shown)


jychp (changes requested):
Your PR introduces a link to a missing image and adds duplicate documentation (this section is about installation, not usage, which is covered below).

We’re fine with using AI, we actually encourage it, but please use it in a way that’s relevant to everyone and that adds real value. Otherwise, the PR works against you instead of showcasing your skills.

PR #2214, opened 2026-01-12
feat(googleworkspace): make device sync optional based on OAuth scopes

kunaals:
### Summary
Makes device sync optional for the Google Workspace module when using OAuth authentication.

**Problem:** The `cloud-identity.devices.readonly` scope requires Cloud Identity Premium or Google Workspace Enterprise. Users without these editions get `invalid_scope` errors when trying to use the googleworkspace module.

**Solution:**
1. Allow callers to specify a `scopes` array in the OAuth auth payload (falls back to `OAUTH_SCOPES` if not provided)
2. Skip device sync gracefully if the devices scope is not in the credentials, logging an info message instead of failing

This enables users to sync users, groups, and OAuth apps without needing Cloud Identity Premium.

### Related issues or links
- Users without Cloud Identity Premium cannot use the googleworkspace module due to hardcoded scopes

### Checklist

**Before:** Sync fails with `invalid_scope` error if refresh token doesn't include devices.readonly scope
**After:** Sync succeeds for users/groups/oauth_apps, logs "Skipping device sync - cloud-identity.devices.readonly scope not authorized"

### DIFF
--- a/docs/root/modules/googleworkspace/config.md
+++ b/docs/root/modules/googleworkspace/config.md
@@ -40,6 +40,30 @@ This module allows authentication from a service account or via OAuth tokens.
 1. Populate an environment variable of your choice with the contents of the base64 output from the previous step.
 1. Call the `cartography` CLI with `--googleworkspace-tokens-env-var YOUR_ENV_VAR_HERE` and `--googleworkspace-auth-method oauth`.
 
+##### Optional: Custom Scopes
+
+By default, cartography requests all supported scopes. If you need to use a subset of scopes (for example, if you don't have Cloud Identity Premium and cannot use the `cloud-identity.devices.readonly` scope), you can specify a custom `scopes` field in the OAuth JSON payload:
+
+```python
+import json
+import base64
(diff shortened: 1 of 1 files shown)


cartography/intel/googleworkspace/__init__.py
```diff
@@ -113,6 +113,10 @@ def start_googleworkspace_ingestion(
             str(base64.b64decode(config.googleworkspace_config).decode())
         )
         logger.info("Attempting to authenticate to Google Workspace using OAuth")
+        # Allow scopes to be specified in the auth payload, falling back to defaults.
+        # This enables callers to use a subset of scopes (e.g., without devices.readonly
+        # if they don't have Cloud Identity Premium).
+        oauth_scopes = auth_tokens.get("scopes", OAUTH_SCOPES)
```
jychp, line 119: You should add this to the module documentation as it is not a standard field in the JSON provided by Google.

jychp (approved):
LGTM

We should add extra doc to document that behavior

PR #2219, opened 2026-01-12
feat: add new intel module for WorkOS

jychp:
### Type of change

### Summary

New intel module for [WorkOS](https://workos.com) — syncs identity, access, and directory data into the graph.

**Resources ingested (11 node types):**

| Node | Label(s) | Source API |
|---|---|---|
| `WorkOSEnvironment` | `Tenant` | Local (root node from client ID) |
| `WorkOSOrganization` | — | `organizations.list_organizations` |
| `WorkOSUser` | `UserAccount` | `user_management.list_users` |
| `WorkOSRole` | — | `organizations.list_organization_roles` |
| `WorkOSOrganizationMembership` | — | `user_management.list_organization_memberships` |
| `WorkOSInvitation` | — | `user_management.list_invitations` |
| `WorkOSDirectory` | — | `directory_sync.list_directories` |
| `WorkOSDirectoryUser` | — | `directory_sync.list_users` |
| `WorkOSDirectoryGroup` | — | `directory_sync.list_groups` |
| `WorkOSOrganizationDomain` | — | `organization_domains.list_organization_domains` |
| `WorkOSAPIKey` | `APIKey` | `organizations.list_api_keys` |
| `WorkOSApplication` | `ThirdPartyApp` | `connect.list_applications` |

**Relationships mapped:**
- Environment → all resources (`:RESOURCE`)
- Organization → Roles (`:HAS`), API Keys (`:OWNS`)
- User → OrganizationMembership (`:MEMBER_OF`)
- OrganizationMembership → Organization (`:IN`), Role (`:WITH_ROLE`)
- Invitation → Organization (`:FOR_ORGANIZATION`), User (`:INVITES`, `:INVITED_BY`)
- Directory → Organization (`:BELONGS_TO`), DirectoryUser/Group (`:HAS`)
- DirectoryUser → Organization (`:BELONGS_TO`), DirectoryGroup (`:MEMBER_OF`)
- DirectoryGroup → Organization (`:BELONGS_TO`)
- OrganizationDomain → Organization (`:DOMAIN_OF`)
- Application → Organization (`:BELONGS_TO`)

**Ontology mappings:** `APIKey`, `Tenant`, `UserAccount` for cross-platform queries.

**Usage:**
```bash
cartography --neo4j-uri bolt://localhost:7687 \
  --workos-apikey-env-var WORKOS_API_KEY \
  --workos-client-id client_XXXXX \
  --selected-modules workos
```

### Related issues or links
- Requires `workos` Python SDK >= 5.44.0 (Connect applications API)

### Breaking changes
None.

### How was this tested?
- 12 integration tests covering all node types and relationships (run against local Neo4j)
- `make test_unit` and `make test_integration` pass
- CI green (linter, tests neo4j4, tests neo4j5)

#### General

#### Proof of functionality

### Notes for reviewers
- All WorkOS resources hang off a `WorkOSEnvironment` root node (keyed by client ID), following the same pattern as other multi-tenant modules (Okta, Entra, etc.)
- Per-item error handling is used when iterating over orgs/directories so one failing entity doesn't abort the full sync
- The module uses cursor-based pagination via a shared `paginated_list()` helper in `util.py`
- 63 new files, ~4900 lines added (models, intel, tests, docs, test data)


jychp:
carbon (2)

jychp:
Screenshot 2026-03-31 at 9 53 09 AM

cartography/intel/workos/api_keys.py
```diff
@@ -0,0 +1,148 @@
...
+    api_keys = []
+
+    for org_id in organization_ids:
+        try:
+            org_keys = paginated_list(
+                client.organizations.list_api_keys,
+                organization_id=org_id,
+            )
+            api_keys.extend(org_keys)
+        except Exception as e:
```
kunaals: same issue here: if one org api call fails and we continue, cleanup can delete that org's previously synced api keys. i'd rather fail the sync than silently drop data.

cartography/intel/workos/organization_domains.py
```diff
@@ -0,0 +1,132 @@
...
+    domains = []
+
+    for org_id in organization_ids:
+        try:
+            org_domains = paginated_list(
+                client.organizations.list_organization_domains,
+                organization_id=org_id,
+            )
+            domains.extend(org_domains)
+        except Exception as e:
```
kunaals: catching and continuing here means cleanup can treat a transient per-org failure like "this org has no domains" and delete previously synced data. i think this should raise instead of soft-skipping.

cartography/intel/workos/invitations.py
```diff
@@ -0,0 +1,122 @@
...
+    result = []
+
+    for invitation in invitations:
+        invitation_dict = {
+            "id": invitation.id,
+            "email": invitation.email,
+            "state": invitation.state,
+            "organization_id": invitation.organization_id,
+            "inviter_user_id": invitation.inviter_user_id,
+            "token": getattr(invitation, "token", None),
```
kunaals: do we want to persist `token` and `accept_invitation_url` at all? these look like invitation secrets, and storing them in neo4j feels riskier than the value we get from keeping them.

cartography/intel/workos/roles.py
```diff
@@ -0,0 +1,130 @@
...
+
+    for org_id, org_roles in roles_by_org.items():
+        for role in org_roles:
+            role_dict = {
+                "id": role.id,
+                "slug": role.slug,
+                "name": role.name,
+                "description": role.description,
+                "type": role.type,
+                "organization_id": org_id,
```
kunaals: this assigns `organization_id` to every returned role, but this endpoint can return environment roles too. those shared roles will end up looking org-scoped/nondeterministic in the graph.

cartography/intel/workos/directory_users.py
```diff
@@ -0,0 +1,156 @@
...
+            "idp_id": user.idp_id,
+            "directory_id": user.directory_id,
+            "organization_id": user.organization_id,
+            "first_name": user.first_name,
+            "last_name": user.last_name,
+            "email": user.email,
+            "group_ids": [g.id for g in user.groups],
+            "state": user.state,
+            "created_at": user.created_at,
+            "updated_at": user.updated_at,
```
kunaals, line 92: we only keep the singular `role` here and drop `roles[]`, and we never relate directory users to `WorkOSRole`. that seems like we're losing real rbac data during ingest.

kunaals (dismissed):
.

cartography/intel/workos/roles.py
```diff
@@ -0,0 +1,134 @@
...
+            if role.id in seen_ids:
+                continue
+            seen_ids.add(role.id)
+            role_dict = {
+                "id": role.id,
+                "slug": role.slug,
+                "name": role.name,
+                "description": role.description,
+                "type": role.type,
+                "organization_id": org_id if role.type == "organization" else None,
```
kunaals: tiny follow-up: the sdk type here is `OrganizationRole`, so i think this condition never matches and `organization_id` stays null for org-scoped roles.

cartography/intel/workos/directory_users.py
```diff
@@ -0,0 +1,148 @@
...
+            "custom_attributes": (
+                json.dumps(user.custom_attributes) if user.custom_attributes else None
+            ),
+            "raw_attributes": (
+                json.dumps(user.raw_attributes) if user.raw_attributes else None
+            ),
+            "role": user.role.slug if user.role else None,
```
kunaals: small follow-up here: we still only persist the singular 'role' and drop roles[], and there's still no link from directory users to WorkOSRole.

kunaals (approved):
approving to unblock, couple more nits

PR #2221, opened 2026-01-12
fix(rules): fix missing data in rules visual cypher query

jychp:
### Summary

This PR simply fix some cypher queries issues in rules


cartography/rules/data/rules/workload_identity_admin_capabilities.py
```diff
@@ -80,7 +80,7 @@
             OR action = 'iam:*'
             OR action = '*'
         )
-        WITH p, p1, p2, p3, ec2
+        WITH p, p1, p2, p3, a, ec2
```
achantavy, line 83: `a` is already included in `p`. Do we need to have this?

Issue #2107, opened 2025-12-04
feat(kubernetes): Add KubernetesPod -> KubernetesSecret relationships


jychp:
### Summary

Add explicit relationships in Cartography between `KubernetesPod` and `KubernetesSecret`.

The ingestion should cover:
* Secrets mounted as volumes (`spec.volumes[].secret`)
* Secrets injected via environment variables (`env` / `envFrom` using `secretKeyRef` / `secretRef`)
  Bonus scope:
* Secrets potentially readable through the pod’s ServiceAccount RBAC permissions

### Motivation
Cartography currently does not expose which pods consume which Kubernetes secrets.
Making these dependencies explicit provides clear security value:
* Enables attack-path analysis when a pod is compromised
* Highlights overly broad or sensitive secrets
* Helps identify stale or unused secrets
* Improves configuration auditing and posture reviews

### Proposed Solution
xtend the Kubernetes ingestion module to emit relationships between `KubernetesPod` and `KubernetesSecret`.

#### 1. Secret volumes

Inspect `spec.volumes[]` and, for each secret volume (`volume.secret.secretName`), create:
```
(KubernetesPod)-[:USES_SECRET_VOLUME]->(KubernetesSecret)
```

Matching based on `(namespace, name)`.

#### 2. Secrets injected via environment variables

For each container

* Parse `env[].valueFrom.secretKeyRef.name`
* Parse `envFrom[].secretRef.name`

Emit:

```
(KubernetesPod)-[:USES_SECRET_ENV]->(KubernetesSecret)
```

#### 3. Bonus: Secrets readable via RBAC

For each pod:

* Resolve `spec.serviceAccountName`
* Determine associated Roles/ClusterRoles through RoleBindings/ClusterRoleBindings
* Inspect RBAC rules:

  * `resources` includes `"secrets"`
  * `verbs` includes `"get"`, `"list"`, `"watch"`
  * Optionally narrow via `resourceNames`

Emit either:

```
(KubernetesPod)-[:CAN_READ_SECRET_VIA_RBAC]->(KubernetesSecret)
```

or, as a first pass if per-secret inference is complex:

```
(KubernetesPod)-[:HAS_SECRET_READ_RBAC]->(KubernetesNamespace)
```

### Additional Context

RBAC resolution is intentionally marked as bonus scope due to its complexity.
Volume-based and env-based secret usage alone would already provide significant value.

PR #2224, opened 2026-01-14
feat(k8s): link Kubernetes secret

Tushar240503:
Summary

This PR adds explicit relationships in Cartography between KubernetesPod and KubernetesSecret for the two most common secret consumption patterns:

Secrets mounted as volumes

Secrets injected via environment variables

These relationships make pod–secret dependencies first-class citizens in the graph and unlock meaningful security and posture analysis.

#2107

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,6 +19,32 @@ This guide teaches you how to write intel modules for Cartography using the mode
 - `cartography/graph/job.py` - Cleanup job utilities
 - `cartography/models/core/` - Base data model classes
 
+## 📝 Git and Pull Request Guidelines
+
+**Signing Commits**: All commits must be signed using the `-s` flag. This adds a `Signed-off-by` line to your commit message, certifying that you have the right to submit the code under the project's license.
+
+```bash
+# Sign a commit with a message
+git commit -s -m "feat(module): add new feature"
+```
+
+**Pull Request Descriptions**: When creating a pull request, use the template at `.github/pull_request_template.md`.
+
+Example PR creation:
+```bash
+gh pr create --title "feat(core): add BufferError retry handling" --body "$(cat <<'EOF'
+### Summary
+Add retry handling for BufferError to cartography's core Neo4j retry logic.
+
+### Related issues or links
+- https://github.com/cartography-cncf/cartography/issues/1234
+
+### Checklist
+- [x] Update/add unit or integration tests.
+EOF
+)"
+```
+
 ## 📋 Table of Contents
 
 1. @Quick Start: Copy an Existing Module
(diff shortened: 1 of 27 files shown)


cartography/models/kubernetes/pods.py
```diff
@@ -68,6 +68,45 @@ class KubernetesPodToKubernetesClusterRel(CartographyRelSchema):
...
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:KubernetesPod)-[:USES_SECRET_VOLUME]->(:KubernetesSecret)
+class KubernetesPodToSecretVolumeRel(CartographyRelSchema):
+    target_node_label: str = "KubernetesSecret"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "cluster_name": PropertyRef("CLUSTER_NAME", set_in_kwargs=True),
```
jychp: We should add the cluster_name into the composite_id

cartography/intel/kubernetes/pods.py
```diff
@@ -81,6 +81,55 @@ def _extract_pod_containers(pod: V1Pod) -> dict[str, Any]:
...
+        containers_to_scan.extend(pod.spec.ephemeral_containers)
+
+    for container in containers_to_scan:
+        # env[].valueFrom.secretKeyRef
+        if container.env:
+            for env in container.env:
+                if (
+                    env.value_from
+                    and env.value_from.secret_key_ref
+                    and namespace
```
jychp: can namespace can be None ?

if no, just remove the condition
if yes, you can exit early to avoid loops

cartography/intel/kubernetes/pods.py
```diff
@@ -81,6 +81,55 @@ def _extract_pod_containers(pod: V1Pod) -> dict[str, Any]:
     return containers
 
 
+def _extract_pod_secrets(pod: V1Pod) -> tuple[list[str], list[str]]:
+    """
+    Extract all secret names referenced by a pod.
+    Returns a tuple of (volume_secret_ids, env_secret_ids).
+    Each list contains unique secret IDs in the format: {namespace}/{secret_name}
+    """
+    volume_secrets = []
```
jychp: should be a set instead of a list and list(set()) at the end

cartography/intel/kubernetes/pods.py
Tony-Stark20: @jychp For Pods, namespace cannot be None. Pods are namespaced objects, and Kubernetes automatically assigns the default namespace if none is specified.
So for objects coming from the API server, pod.metadata.namespace will always be set.
The only case where it could be None is with manually constructed mocks or incomplete test objects.
This check was originally suggested by  the cubic-dev-ai bot, but given the above, I think we can safely remove it rather than exit early

PR #2225, opened 2026-01-14
fix(aws): Sync Failure in AWS Bedrock

shyammukund:
### Summary

**Bug:** The AWS bedrock sync for custom models fails loudly in regions where custom models are not supported (anywhere not us-east-1 or us-west-2). ListCustomModels API returns a Validation Exception: Unknown Operation causing the entire sync to fail. See below for full error. 

**Fix:** The get_custom_models function now catches this specific error and gracefully skipping and logging unsupported regions allowing the sync to continue.

### DIFF
--- a/cartography/intel/aws/bedrock/custom_models.py
+++ b/cartography/intel/aws/bedrock/custom_models.py
@@ -39,7 +39,7 @@ def get_custom_models(
     to retrieve full details (jobArn, jobName, trainingDataConfig, outputDataConfig).
     """
     if region not in CUSTOM_MODELS_SUPPORTED_REGIONS:
-        logger.info(
+        logger.debug(
             "Bedrock custom models not supported in region %s. Skipping.",
             region,
         )


cartography/intel/aws/bedrock/custom_models.py
```diff
@@ -44,8 +45,22 @@ def get_custom_models(
...
     paginator = client.get_paginator("list_custom_models")
     model_summaries = []
-    for page in paginator.paginate():
-        model_summaries.extend(page.get("modelSummaries", []))
+    try:
+        for page in paginator.paginate():
+            model_summaries.extend(page.get("modelSummaries", []))
+    except botocore.exceptions.ClientError as e:
+        # Custom models are only supported in us-east-1 and us-west-2.
+        # Other regions return ValidationException with "Unknown Operation".
```
jychp: Instead of catching error we should filter on region before making the call to save some API call for performance

PR #2230, opened 2026-01-15
feat(gcp): Support GCP Artifact Registry

shyammukund:
- This PR adds coverage for GCP Artifact Registry 
 

### Checklist

image

### DIFF
--- a/cartography/intel/gcp/artifact_registry/manifest.py
+++ b/cartography/intel/gcp/artifact_registry/manifest.py
@@ -282,7 +282,15 @@ def sync_artifact_registry_manifests(
     logger.info(f"Syncing Artifact Registry image manifests for project {project_id}.")
 
     # Get all manifests concurrently using async
-    all_manifests = asyncio.run(
+    # Use get_event_loop() + run_until_complete() to avoid tearing down loop
+    try:
+        loop = asyncio.get_event_loop()
+    except RuntimeError:
+        # No event loop in current thread, create one
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+
+    all_manifests = loop.run_until_complete(
         get_all_manifests_async(credentials, docker_artifacts_raw)
     )
 


cartography/intel/gcp/artifact_registry/repository.py
```diff
@@ -0,0 +1,169 @@
...
+                        previous_response=response,
+                    )
+                )
+        except (PermissionDenied, DefaultCredentialsError, RefreshError) as e:
+            logger.warning(
+                f"Failed to get Artifact Registry repositories for project {project_id} "
+                f"in location {location} due to permissions or auth error: {e}",
+            )
+            continue
+        except HttpError as e:
```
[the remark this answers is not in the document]
shyammukund: i ran into this error when testing on multiple different projects so i think we should keep catching this

kunaals (commented):
Lets have different node types for Containers, Language Packages, Helm Charts, and Generic Artifacts

cartography/intel/gcp/artifact_registry/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+    # Sync repositories
+    repositories_raw = sync_artifact_registry_repositories(
+        neo4j_session,
+        client,
+        project_id,
+        update_tag,
+        common_job_parameters,
+    )
+
+    if not repositories_raw:
```
kunaals: returning early here skips artifact/manifest cleanup. if a project's repos get deleted (or the api returns empty for some reason), we'd never clean up the orphaned nodes. should run cleanup even when empty.

cartography/intel/gcp/artifact_registry/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+    artifacts_raw = sync_artifact_registry_artifacts(
+        neo4j_session,
+        client,
+        repositories_raw,
+        project_id,
+        update_tag,
+        common_job_parameters,
+    )
+
+    if not artifacts_raw:
```
kunaals: same issue - if docker images get removed, manifest cleanup never runs. run the cleanup portion even when `artifacts_raw` is empty.

cartography/intel/gcp/artifact_registry/util.py
```diff
@@ -0,0 +1,78 @@
+import logging
+from typing import List
```
kunaals: nit: use `list[str]` instead of `List[str]` for consistency with rest of codebase

cartography/intel/gcp/artifact_registry/util.py
```diff
@@ -0,0 +1,78 @@
...
+    Gets all available Artifact Registry locations for a project.
+    Filters to regions that commonly support Artifact Registry to improve sync performance.
+    """
+    try:
+        req = client.projects().locations().list(name=f"projects/{project_id}")
+        res = req.execute()
+
+        # Filter to commonly-used regions to avoid excessive API calls
+        # Reference: https://cloud.google.com/artifact-registry/docs/repositories/repo-locations
+        supported_regions = {
```
kunaals: this hardcoded region allowlist will silently skip repos in newer/less common regions. can we just iterate all locations the api returns? if there's a perf reason for filtering, let's make it configurable or at least document why.

cartography/intel/gcp/artifact_registry/util.py
```diff
@@ -0,0 +1,78 @@
...
+            if location_id in supported_regions:
+                locations.append(location_id)
+
+        logger.info(
+            f"Found {len(locations)} supported Artifact Registry locations "
+            f"(filtered from {len(all_locations)} total) for project {project_id}"
+        )
+        return locations
+
+    except HttpError as e:
```
kunaals, line 31: returning `[]` on HttpError can mask failures - sync continues with empty data and cleanup wipes existing nodes. consider re-raising so the sync fails visibly.

cartography/models/gcp/artifact_registry/docker_image.py
```diff
@@ -0,0 +1,82 @@
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "CONTAINS"
+    properties: GCPArtifactRegistryDockerImageToRepositoryRelProperties = (
+        GCPArtifactRegistryDockerImageToRepositoryRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class GCPArtifactRegistryDockerImageSchema(CartographyNodeSchema):
+    label: str = "GCPArtifactRegistryDockerImage"
```
kunaals: rename to `GCPArtifactRegistryContainerImage` - "docker" is legacy terminology, gcp supports oci format in these repos too

cartography/models/gcp/artifact_registry/manifest.py
```diff
@@ -0,0 +1,80 @@
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "HAS_MANIFEST"
+    properties: GCPArtifactRegistryImageManifestToArtifactRelProperties = (
+        GCPArtifactRegistryImageManifestToArtifactRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class GCPArtifactRegistryImageManifestSchema(CartographyNodeSchema):
+    label: str = "GCPArtifactRegistryImageManifest"
```
kunaals: rename to `GCPArtifactRegistryPlatformImage` - "ImageManifest" is confusing since the parent (`ContainerImage`) is also a manifest. this node represents a platform-specific image within a multi-arch image index.

cartography/intel/gcp/artifact_registry/artifact.py
```diff
@@ -0,0 +1,815 @@
...
+    except (PermissionDenied, DefaultCredentialsError, RefreshError) as e:
+        logger.warning(
+            f"Failed to get Python packages for repository {repository_name} "
+            f"due to permissions or auth error: {e}",
+        )
+        return []
+
+
+@timeit
+def get_go_modules(client: Resource, repository_name: str) -> list[dict]:
```
kunaals: this will fail at runtime - the artifact registry api doesn't have a `list` method for goModules. per google docs, goModules only supports `upload`. same issue exists for `get_apt_artifacts` and `get_yum_artifacts` below (apt/yum only have `upload` and `import`, no `list`).

verified by hitting the api directly:
```
GET /v1/.../goModules → 404 Not Found
GET /v1/.../aptArtifacts → 404 Not Found
```

options:
1. remove go/apt/yum support entirely for now
2. use the generic `files` or `packages` api if one exists
3. document that these formats aren't supported yet

refs:
- https://cloud.google.com/artifact-registry/docs/reference/rest/v1/projects.locations.repositories.goModules
- https://cloud.google.com/artifact-registry/docs/reference/rest/v1/projects.locations.repositories.aptArtifacts

cartography/intel/gcp/artifact_registry/artifact.py
```diff
@@ -0,0 +1,815 @@
...
+        if not artifacts_raw:
+            continue
+
+        # Route to appropriate collection based on format
+        if repo_format == "DOCKER":
+            # Split Docker format artifacts by media type
+            # Helm charts are stored as OCI artifacts with helm in the media type
+            for artifact in artifacts_raw:
+                media_type = artifact.get("mediaType", "")
+                if HELM_MEDIA_TYPE_IDENTIFIER in media_type.lower():
```
kunaals: the dockerImages api returns an `artifactType` field which would be more reliable for detecting helm charts. `mediaType` is the manifest media type, whereas `artifactType` explicitly declares what the artifact is (e.g., `application/vnd.cncf.helm.config.v1+json`).

current approach could miss helm oci artifacts that use standard oci manifest media types but declare helm via artifactType.

```python
artifact_type = artifact.get("artifactType", "")
if HELM_MEDIA_TYPE_IDENTIFIER in artifact_type.lower() or HELM_MEDIA_TYPE_IDENTIFIER in media_type.lower():
```

cartography/intel/gcp/artifact_registry/manifest.py
```diff
@@ -0,0 +1,226 @@
...
+    :param docker_artifacts_raw: List of raw Docker artifact data from the API.
+    :param project_id: The GCP project ID.
+    :param update_tag: The update tag for this sync.
+    :param common_job_parameters: Common job parameters for cleanup.
+    """
+    logger.info(f"Syncing Artifact Registry image manifests for project {project_id}.")
+
+    all_manifests: list[dict] = []
+
+    for artifact in docker_artifacts_raw:
```
kunaals: this fetches manifests sequentially with blocking `requests.get()` calls - each one waits ~300ms for network round trip before starting the next. for a registry with 100 multi-arch images that's ~30s of wall clock time vs ~2s with concurrent fetches.

the codebase already has a pattern for this in `ecr_image_layers.py:503-758` using `httpx.AsyncClient` + `asyncio.Semaphore`:

```python
async def get_manifest_list_async(credentials, image_uri, http_client, semaphore):
    async with semaphore:
        response = await http_client.get(manifest_url, headers=headers, timeout=30.0)
        return image_uri, response.json().get("manifests", [])

async def fetch_all_manifests_async(credentials, artifacts, max_concurrent=50):
    semaphore = asyncio.Semaphore(max_concurrent)
    async with httpx.AsyncClient() as http_client:
        tasks = [asyncio.create_task(get_manifest_list_async(...)) for a in artifacts]
        for task in asyncio.as_completed(tasks):
            uri, manifests = await task
```

since this is pure http to the docker registry api (not googleapiclient), it's a good fit for full async rather than the threadpool wrapper approach.

docs/root/modules/gcp/schema.md
```diff
@@ -1434,3 +1434,229 @@ Representation of a GCP [Secret Manager Secret Version](https://cloud.google.com
...
+
+#### Overview
+
+Google Cloud Artifact Registry is a universal package manager for managing container images and language packages. Cartography ingests the following Artifact Registry resources with dedicated node types for each artifact category:
+
+```mermaid
+graph LR
+    Project[GCPProject]
+    Repository[GCPArtifactRegistryRepository]
+    DockerImage[GCPArtifactRegistryDockerImage]
```
kunaals: node labels in the diagram don't match the actual model classes:
- `GCPArtifactRegistryDockerImage` should be `GCPArtifactRegistryContainerImage`
- `GCPArtifactRegistryImageManifest` should be `GCPArtifactRegistryPlatformImage`

cartography/intel/gcp/artifact_registry/artifact.py
shyammukund: Decided to go ahead and remove support for go/apt/yum. This includes removing the generic artifact node because  apt and yum were the only artifacts that used that data model.

PR #2239, opened 2026-01-19
feat(aws): EKS Access Entry Mappings

### LINKED ISSUE #1855: Add Access Entries from EKS in Kubernetes Module
*Title*: *We are potentially missing coverage of many Kubernetes Users and Groups by not processing access entries from EKS. Right now we are just getting Users from the aws-auth-config-map and the RoleBindings and ClusterRole bindings. Processing Access Entries will allow us to get users and groups that aren't in the other 2 sources. *

*Description*:
- Access Entries live outside the cluster in EKS. Example of the json object that the AWS API returns for Access Entries
- ```  
  "accessEntry": {
    "clusterName": "eks-customer",
    "principalArn": "arn:aws:iam::111122223333:user/eks-admin-user",
    "kubernetesGroups": [],
    "accessEntryArn": "arn:aws:eks:us-west-2:111122223333:access-entry/eks-customer/user/111122223333/eks-admin-user/0acb1bc6-cb0a-ede6-11ae-a6506e3d36p0",
    "createdAt": "2025-04-14T22:45:48.097000-05:00",
    "modifiedAt": "2025-04-14T22:45:48.097000-05:00",
    "tags": {},
    "username": "arn:aws:iam::111122223333:user/eks-admin-user",
    "type": "STANDARD"
  }

[optional *Relevant Links*:]
- look at intel/kubernetes/rbac.py to see how we ingest users and groups from the role bindings and cluster role bindings 
- look at intel/kubernetes/eks.py to see how we ingest users and groups from the aws-auth-map

Tushar240503:
Summary

PR Description
Summary

Add support for EKS Access Entries as a source of Kubernetes identity mappings in Cartography.

EKS Access Entries are a newer EKS feature that provides an alternative to the aws-auth ConfigMap for granting IAM principals access to Kubernetes clusters. This PR ingests Access Entries via the EKS API and maps IAM Users and Roles to Kubernetes Users and Groups in the graph.

Changes

Fetch EKS Access Entries using ListAccessEntries and DescribeAccessEntry

Transform Access Entries into:

KubernetesUser nodes

KubernetesGroup nodes

Create MAPS_TO relationships from:

AWSUser → KubernetesUser

AWSRole → KubernetesUser

AWSUser → KubernetesGroup

AWSRole → KubernetesGroup

Integrate Access Entry ingestion into the existing EKS sync flow

Add  integration tests covering Access Entry behavior

Screenshot 2026-01-24 at 3 30 27 AM
Screenshot 2026-01-24 at 3 30 09 AM

Related issues or links
Fixes #1855


cartography/intel/kubernetes/eks.py
```diff
@@ -218,10 +216,56 @@ def get_oidc_provider(
...
+    for page in page_iterator:
+        for principal_arn in page.get("accessEntries", []):
+            try:
+                detail_response = client.describe_access_entry(
+                    clusterName=cluster_name, principalArn=principal_arn
+                )
+                access_entries.append(detail_response["accessEntry"])
+            except Exception as e:
+                logger.warning(
+                    f"Failed to describe access entry for principal {principal_arn}: {e}"
```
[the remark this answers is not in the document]
Tony-Stark20: In general, IAM principal ARNs are not classified as PII, since they identify roles or users within an AWS account rather than real individuals. From a pure PII-compliance standpoint, logging them is usually acceptable.

cartography/intel/kubernetes/eks.py
```diff
@@ -1,11 +1,10 @@
 import logging
 from typing import Any
-from typing import Dict
-from typing import List
```
Tushar240503: This was done for  updating to Python 3.9+ type hints

cartography/intel/kubernetes/eks.py
```diff
@@ -218,10 +217,61 @@ def get_oidc_provider(
...
+            try:
+                detail_response = client.describe_access_entry(
+                    clusterName=cluster_name, principalArn=principal_arn
+                )
+                access_entries.append(detail_response["accessEntry"])
+            except ClientError as e:
+                # If the access entry is not found, we can safely skip it.
+                if e.response["Error"]["Code"] == "ResourceNotFoundException":
+                    logger.warning(
+                        f"Access entry lookup failed for principal {principal_arn}: {e}"
```
Tushar240503: Only ResourceNotFoundException is safe to ignore here.
Ignoring AccessDeniedException could return partial data, causing cleanup to
mistakenly delete real EKS access entries that still exist.

cartography/intel/kubernetes/eks.py
```diff
@@ -254,9 +304,76 @@ def transform_oidc_provider(
...
+    """
+    Transform EKS Access Entries into KubernetesUser and KubernetesGroup data.
+
+    Access Entries map IAM principals (users or roles) to Kubernetes users and groups.
+    Each access entry has:
+    - principalArn: The IAM principal (user or role) ARN
+    - username: The Kubernetes username (defaults to principalArn if not specified)
+    - kubernetesGroups: List of Kubernetes groups the user belongs to
+
+    Returns a dictionary with 'users' and 'groups' keys containing transformed data.
```
jychp, line 319: Nice work on this feature! The transform function is well-structured.

One minor observation: the docstring mentions "defaults to principalArn if not specified" for username, but looking at the mock test data, the fourth entry has `username` explicitly set to the ARN value rather than being absent.

Could you clarify: does the AWS API always return a `username` field (potentially set to the ARN), or can it actually be `None`/missing? If the latter, you may want to add a test case for that scenario. If the former, consider updating the docstring to reflect that AWS sets it to the ARN rather than it being a default applied by this code.

Tony-Stark20: It always returns a username 
Link to the doc 
https://docs.aws.amazon.com/eks/latest/userguide/set-custom-username.html

cartography/intel/kubernetes/eks.py
```diff
@@ -254,9 +304,75 @@ def transform_oidc_provider(
...
+    """
+    all_users = []
+    all_groups = []
+
+    for entry in access_entries:
+        principal_arn = entry.get("principalArn")
+        username = entry.get("username")
```
jychp: The docstring states AWS always populates `username`, but `.get()` allows `None` values through silently—creating nodes with IDs like `cluster_name/None`.

Use `entry["username"]` to fail loudly if the field is unexpectedly missing

Tony-Stark20: Done

jychp (approved):
Thanks for that PR and your follow up on my comments.

Issue #2237, opened 2026-01-16
feat: Add GitLab Container Registry support


kunaals:
## Summary

This proposal adds **GitLab Container Registry** support to Cartography, enabling discovery and vulnerability scanning of container images hosted in GitLab.

We want Trivy findings to point to images from GitLab and Google artifact registries. Rather than creating an abstract `ContainerImage` layer via `extra_node_labels`, this proposal takes the simpler approach of:
1. Adding GitLab Container Registry as its own module with `GitLabContainerImage` nodes
2. Explicitly extending Trivy to support GitLab images (one relationship per registry)
3. Recommending **ontology mapping** as the proper long-term solution for cross-registry unification

---

## Why Not a ContainerImage Abstraction?

We considered adding `extra_node_labels: ["ContainerImage"]` to unify ECR, GitLab, GAR, etc. under a common label. However, this approach:

- Creates an implicit contract (must have `digest`, `type`, specific properties)
- Expects consumers (Trivy, K8s) to target the abstract label
- Requires all registries to conform to a shared interface
- **That's literally what ontology mapping is designed for**

The `extra_node_labels` pattern is meant for simple semantic grouping (like `:ECSContainer:Container`), not for building infrastructure that depends on matching against that label.

**If cross-registry unified querying is needed, use the ontology system** in `models/ontology/mapping/`. That provides:
- Explicit field normalization (`_ont_*` properties)
- Validation tooling
- A proper contract that's enforced and documented

---

### Design Principles

1. **Each registry is its own module** - GitLab syncs `GitLabContainerImage`, GAR syncs `GARContainerImage`, etc.
2. **Trivy explicitly supports each registry** - One relationship schema per registry it knows about
3. **No magic abstraction** - More verbose but completely transparent
4. **Ontology for unification** - If you need cross-registry queries, use ontology mapping

---

#### 1.1 Create GitLab Container Registry Models

**File:** `cartography/models/gitlab/container_registry.py`

```python
@dataclass(frozen=True)
class GitLabContainerImageNodeProperties(CartographyNodeProperties):
    id: PropertyRef = PropertyRef("id")
    digest: PropertyRef = PropertyRef("digest", extra_index=True)
    name: PropertyRef = PropertyRef("name")  # tag name
    path: PropertyRef = PropertyRef("path")
    location: PropertyRef = PropertyRef("location")  # full pull URL
    total_size: PropertyRef = PropertyRef("total_size")
[...]

PR #2254, opened 2026-01-21
feat(gitlab): Support GitLab Container Registries

shyammukund:
- addresses phase 1 of https://github.com/cartography-cncf/cartography/issues/2237

### Checklist

image


cartography/intel/gitlab/container_images.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+    response = fetch_registry_manifest(
+        gitlab_url,
+        registry_url,
+        repository_name,
+        reference,
+        token,
+        accept_header=MANIFEST_ACCEPT_HEADER,
+    )
+    response.raise_for_status()
```
kunaals, line 82: if a tag gets deleted between the list call and this manifest fetch, `raise_for_status()` will abort the entire sync. you could handle 404 as a skip:

```python
if response.status_code == 404:
    logger.warning(f"tag {reference} not found, may have been deleted")
    return None
response.raise_for_status()
```

then callers would need to handle the None return and continue to the next tag.

cartography/intel/gitlab/container_images.py
```diff
@@ -0,0 +1,294 @@
...
+) -> tuple[list[dict[str, Any]], list[dict[str, Any]]]:
+    """
+    Fetch container image manifests for all repositories via the Registry V2 API.
+
+    For each repository, fetches tags and then retrieves the manifest for each tag.
+    Returns raw manifest data for transformation, plus manifest lists for attestation discovery.
+    """
+    all_manifests: list[dict[str, Any]] = []
+    manifest_lists: list[dict[str, Any]] = []
+    seen_digests: set[str] = set()
```
kunaals: this global `seen_digests` set means if the same image digest exists in repo-a and repo-b, only repo-a gets scanned for cosign attestations. if repo-b has its own `.sig`/`.att` tags, theyd be missed.

options:
1. scope by repo: `seen_digests: dict[str, set[str]]` keyed by repository_name
2. keep global de-dup but accept that attestation discovery is best-effort for shared digests
3. remove de-dup entirely if attestation coverage matters more than avoiding duplicate image nodes

cartography/intel/gitlab/container_images.py
```diff
@@ -0,0 +1,294 @@
...
+
+    for manifest in raw_manifests:
+        media_type = manifest.get("mediaType")
+        is_manifest_list = media_type in MANIFEST_LIST_MEDIA_TYPES
+
+        # Extract child image digests for manifest lists
+        child_image_digests = None
+        if is_manifest_list:
+            manifests_array = manifest.get("manifests", [])
+            child_image_digests = [
```
kunaals, line 237: this includes attestation-manifest entries in `child_image_digests`, but those digests are filtered out earlier (lines 148-153) and never ingested as ContainerImage nodes. the CONTAINS_IMAGE relationship will try to link to non-existent digests.

filter them here too:

```python
child_image_digests = [
    m.get("digest") for m in manifests_array
    if m.get("digest") and m.get("annotations", {}).get("vnd.docker.reference.type") != "attestation-manifest"
]
```

cartography/intel/gitlab/container_repository_tags.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Fetch full details for each tag to get the digest
+    detailed_tags = []
+    for tag in tags:
+        tag_name = tag.get("name")
+        if not tag_name:
+            continue
+
+        detail_url = f"{gitlab_url}/api/v4/projects/{project_id}/registry/repositories/{repository_id}/tags/{tag_name}"
+        response = requests.get(
```
kunaals: this uses raw `requests.get()` without the retry logic from util.py. a single transient 429 or 5xx will fail silently (line 60-66 catches it but falls back to basic info without digest), which breaks REFERENCES relationships for those tags.

you could use `get_single()` from util.py which has retry + rate limit handling:

```python
from cartography.intel.gitlab.util import get_single

try:
    tag_detail = get_single(
        gitlab_url,
        token,
        f"/api/v4/projects/{project_id}/registry/repositories/{repository_id}/tags/{tag_name}",
    )
    detailed_tags.append(tag_detail)
except requests.exceptions.HTTPError:
    logger.warning(f"Failed to fetch details for tag {tag_name}")
    detailed_tags.append(tag)
```

cartography/intel/gitlab/container_images.py
```diff
@@ -0,0 +1,294 @@
...
+        registry_url, repository_name = _parse_repository_location(location)
+
+        # Fetch tags for this repository
+        tags = get_paginated(
+            gitlab_url,
+            token,
+            f"/api/v4/projects/{project_id}/registry/repositories/{repository_id}/tags",
+        )
+
+        for tag in tags:
```
kunaals, line 137: (perf, not blocking) this sequential loop will be slow for large registries - each tag requires a round-trip for manifest fetch, potentially another for config blob. for orgs with 100 repos x 10 tags = 1000+ sequential http calls.

if you want to optimize later, `httpx.AsyncClient` with `asyncio.gather()` and a semaphore (e.g., max 10 concurrent) would help significantly.

cartography/intel/gitlab/container_image_attestations.py
```diff
@@ -0,0 +1,250 @@
...
+
+        if not all([image_digest, registry_url, repository_name]):
+            continue
+
+        # Type narrowing after the guard check
+        image_digest = str(image_digest)
+        registry_url = str(registry_url)
+        repository_name = str(repository_name)
+
+        for suffix in ATTESTATION_SUFFIXES:
```
kunaals, line 70: (perf, not blocking) same sequential pattern here - for each image we probe 2 attestation tags (.sig and .att). no bulk API exists so this is the only way, but bounded concurrency would help at scale if needed.

cartography/intel/gitlab/container_images.py
shyammukund: scope by repository seems like the best option here. i think it solves the attestation discovery issue, deduplicates within a repo. only downside is that if you have the same image in multiple repos there will be multiple nodes for the same image in different repos. but maybe thats the correct way to map it because they would be seperate instances in different repos. 

@kunaals any thoughts?

shyammukund:
@kunaals The correctional issue have been addressed. I think it would be best that the async refactor be done in a fast follow on a new PR after this gets merged.

PR #2257, opened 2026-01-22
refactor(gcp): refactor compute in GCP

jychp:
## Summary

Completes the migration of the GCP Compute module to the Cartography data model, eliminating all raw Cypher queries from `cartography/intel/gcp/compute.py`.

Includes #2256.

### New model schemas

- `GCPFirewallTargetTagSchema` - for firewall target tags with `TARGET_TAG` relationship
- `GCPSubnetStubSchema` - for creating subnet stub nodes (preserves legacy MERGE behavior)

### Analysis job

- `gcp_compute_instance_vpc_analysis.json` - creates derived `MEMBER_OF_GCP_VPC` relationships

### Bug fixes

- Added missing `project_id` property to `GCPInstanceNodeProperties` and `GCPForwardingRuleNodeProperties`
- Fixed `_zones_to_regions` to use zone's `region` URL field instead of parsing zone names (supports AI zones like `us-south1-ai`)
- Added error handling in `get_gcp_subnets` and `get_gcp_regional_forwarding_rules` for invalid regions
- Moved cleanup calls outside region loops for efficiency

## Design Decision: Analysis Job for MEMBER_OF_GCP_VPC

The `MEMBER_OF_GCP_VPC` relationship is a derived relationship that traverses multiple hops (Instance → NIC → Subnet → VPC). Rather than keeping raw Cypher in the intel module, we moved this to an analysis job.

**Rationale:**

- Keeps `compute.py` clean with zero raw Cypher - all ingestion uses `load()` and all cleanup uses `GraphJob.from_node_schema()`
- Follows the existing pattern for derived relationships (similar to `aws_ec2_keypair_analysis.json`, `gsuite_human_link.json`)
- Acceptable tradeoff: while we aim to limit analysis jobs, this one is justified to maintain code consistency pending a future rework of derived relationship handling in the data model

## Test Plan

- [x] All 46 GCP integration tests pass
- [x] `test_sync_gcp_instances_with_vpc_relationship` updated to call analysis job


cartography/intel/gcp/compute.py
```diff
@@ -680,480 +824,258 @@ def load_gcp_subnets(
...
+    allowed_rules, denied_rules = _transform_firewall_ip_rules(fw_list)
+    if allowed_rules:
+        load(
+            neo4j_session,
+            GCPIpRuleAllowedSchema(),
+            allowed_rules,
+            lastupdated=gcp_update_tag,
+            PROJECT_ID=project_id,
+        )
+    if denied_rules:
```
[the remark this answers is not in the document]
kunaals: I think this is true, i dont see a cleanup job for GCPIpRuleDeniedSchema

cartography/data/jobs/analysis/gcp_compute_instance_vpc_analysis.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "GCP Instance to VPC derived relationship analysis",
```
kunaals, line 2: we should call this in a sync pipeline somewhere or document that it needs to be scheduled separately

cartography/models/gcp/compute/forwarding_rule.py
```diff
@@ -0,0 +1,136 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class GCPForwardingRuleNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("partial_uri")
```
kunaals, line 16: this partial_uri used to be field. for backwards compat, lets add this field back as a duplicate to not break callers
```suggestion
    id: PropertyRef = PropertyRef("partial_uri")
    partial_uri: PropertyRef = PropertyRef("partial_uri")
```

cartography/data/jobs/analysis/gcp_compute_instance_vpc_analysis.json
kunaals: we could also reintroduve the instance-to-vpc rel in intel/gcp/compute.py

cartography/intel/gcp/compute.py
```diff
@@ -582,57 +596,187 @@ def _parse_port_string_to_rule(
...
+    :param instances: List of transformed GCP instances
+    :return: List of network tag objects ready for ingestion
+    """
+    tags: list[dict] = []
+    seen_tags: set[str] = set()
+    for instance in instances:
+        for tag in instance.get("tags", {}).get("items", []):
+            for nic in instance.get("networkInterfaces", []):
+                tag_id = _create_gcp_network_tag_id(nic["vpc_partial_uri"], tag)
+                if tag_id not in seen_tags:
```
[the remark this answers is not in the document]
kunaals: this is correct and a serious issue, we should dedupe on (tag_id, instance_partial_uri)

kunaals (commented):
one more thing that codex found when i ran CLI locally: 

> [P3] Clean stale forwarding-rule VPC/Subnet links — /Users/kunaalsikka/src/cartography/cartography/intel/gcp/compute.py:1152-1153 cleanup_gcp_forwarding_rules now uses GCPForwardingRuleSchema, which has no subnet/VPC relationships, so stale RESOURCE edges to GCPSubnet/GCPVpc are never cleaned. If a forwarding rule changes between subnet-backed and VPC-backed, the old edge will persist and produce incorrect topology. Consider running cleanup on the schema variants with those relationships or re-adding explicit rel cleanup.

cartography/data/jobs/analysis/gcp_compute_instance_vpc_analysis.json
jychp: we can not use the native relationship due to the multi hop path needed to construct that relation

kunaals: Its okay to put it at the end of start_gcp_ingestion

kunaals (approved):
approving to unblock, lets just make sure to add the analysis job to the end of the start_gcp_ingestion function

PR #2263, opened 2026-01-22
feat(trivy): Add Trivy Support for GitLab Container Registries

shyammukund:
- This PR adds trivy support for GitLab Container registries, mirroring the trivy support for ECR Images

- https://github.com/cartography-cncf/cartography/issues/2237
- This PR is stacked on https://github.com/cartography-cncf/cartography/pull/2254

### Checklist

### DIFF
--- a/cartography/client/gitlab/container_images.py
+++ b/cartography/client/gitlab/container_images.py
@@ -37,7 +37,7 @@ def get_gitlab_container_tags(
     """
     query = """
     MATCH (tag:GitLabContainerRepositoryTag)
-    WHERE tag.location IS NOT NULL AND tag.digest IS NOT NULL
+    WHERE tag.location IS NOT NULL
     RETURN tag.location AS location, tag.digest AS digest
     """
     return neo4j_session.read_transaction(read_list_of_tuples_tx, query)


cartography/intel/trivy/__init__.py
```diff
@@ -48,6 +49,50 @@ def _get_scan_targets_and_aliases(
     return image_uris, digest_aliases
 
 
+def _get_gitlab_scan_targets_and_aliases(
+    neo4j_session: Session,
+) -> tuple[set[str], dict[str, str]]:
+    """
+    Return GitLab container image URIs and a mapping of digest-qualified URIs to base URIs.
+    """
```
kunaals, line 62: this function only builds digest aliases (`uri@digest`), not tag aliases. unlike ecr which adds tagged uris to `image_uris`, gitlab only adds base uris. the first matching check in `_prepare_trivy_data` will always fail for gitlab and fall through to digest alias lookup.

works fine when RepoDigests is present, but if trivy outputs RepoTags without RepoDigests (locally built images), gitlab scans silently skip. consider also querying `GitLabContainerRepositoryTag.location` for tag-based aliases.

cartography/models/trivy/findings.py
```diff
@@ -53,6 +53,17 @@ class TrivyFindingToImageRel(CartographyRelSchema):
     properties: TrivyFindingToImageRelProperties = TrivyFindingToImageRelProperties()
 
 
+@dataclass(frozen=True)
+class TrivyFindingToGitLabImageRel(CartographyRelSchema):
+    target_node_label: str = "GitLabContainerImage"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("ImageDigest")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
```
kunaals, line 70: multi-arch correctness concern: this assumes `ImageDigest` will match `GitLabContainerImage.id`. for multi-arch images where trivy reports a platform digest but the graph only has the manifest list, findings won't link. the behavior depends on how trivy was invoked and the docker version used.

either:
1. ensure child platform images are always ingested with their digests
2. add fallback matching logic
3. document the requirement that child images must exist

cartography/models/trivy/package.py
```diff
@@ -39,6 +39,17 @@ class TrivyPackageToImageRel(CartographyRelSchema):
     properties: TrivyPackageToImageRelProperties = TrivyPackageToImageRelProperties()
 
 
+@dataclass(frozen=True)
+class TrivyPackageToGitLabImageRel(CartographyRelSchema):
+    target_node_label: str = "GitLabContainerImage"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("ImageDigest")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
```
kunaals, line 51: same multi-arch concern as findings.py - packages may fail to link if digest doesn't match.

tests/data/trivy/trivy_gitlab_sample.py
```diff
@@ -0,0 +1,107 @@
...
+            "sha256:layer1111222333444555666777888999000aaabbbcccdddeeefff00011122",
+        ],
+        "RepoTags": ["registry.gitlab.example.com/myorg/awesome-project/app:latest"],
+        "RepoDigests": [
+            "registry.gitlab.example.com/myorg/awesome-project/app@sha256:aaa111222333444555666777888999000aaabbbcccdddeeefff000111222333"
+        ],
+        "ImageConfig": {
+            "architecture": "amd64",
+            "os": "linux",
+        },
```
kunaals, line 33: test data uses a single RepoDigests entry matching a single-arch image. doesn't exercise:
1. multi-entry RepoDigests (which entry gets selected?)
2. digest matching a manifest_list type image
3. platform-specific child image matching

consider adding test cases to verify the relationship logic handles these scenarios.

cartography/models/trivy/findings.py
shyammukund: so to make sure that child platform images are always ingested I added tracking and locking to show that a child platform image gets ingested and if it doesn't this is logged instead of failing silently. This should make it easier to determine why a finding did not link.

cartography/models/trivy/package.py
shyammukund: Since a digest alias is created from a child image the child image tracking from the prior comment should help ensure that digest are getting ingested as well.

cartography/intel/gitlab/container_images.py
```diff
@@ -172,31 +175,58 @@ def get_container_images(
...
+                                registry_url,
+                                repository_name,
+                                child_config["digest"],
+                                token,
+                            )
+                        except Exception as e:
+                            logger.warning(
+                                f"Failed to fetch config blob for child {child_digest[:16]}...: {e}. "
+                                f"Architecture metadata may be incomplete."
+                            )
```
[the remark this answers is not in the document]
kunaals: this is intentional because this metadata is optional

cartography/client/gitlab/container_images.py
```diff
@@ -0,0 +1,25 @@
...
+
+    :param neo4j_session: The neo4j session object.
+    :return: 2-tuples of (uri, digest) for each GitLab container image.
+    """
+    query = """
+    MATCH (img:GitLabContainerImage)
+    WHERE img.uri IS NOT NULL AND img.digest IS NOT NULL
+    RETURN img.uri AS uri, img.digest AS digest
+    """
+    return neo4j_session.read_transaction(read_list_of_tuples_tx, query)
```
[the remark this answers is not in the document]
kunaals: this is a codebase wide inconsistency so not going to fix in this PR

PR #2267, opened 2026-01-23
feat(trivy): Add Trivy Support for GCP Artifact Registry

shyammukund:
- This PR adds trivy support for GCP Artifact Registries, mirroring the trivy support for ECR Images

- Issue: https://github.com/cartography-cncf/cartography/issues/2237
- This PR is stacked on https://github.com/cartography-cncf/cartography/pull/2230

### Checklist


cartography/client/gcp/artifact_registry.py
```diff
@@ -0,0 +1,25 @@
...
+    neo4j_session: neo4j.Session,
+) -> Set[Tuple[str, str]]:
+    """
+    Queries the graph for all GCP Artifact Registry container images with their URIs and digests.
+
+    :param neo4j_session: The neo4j session object.
+    :return: 2-tuples of (uri, digest) for each GCP container image.
+    """
+    query = """
+    MATCH (img:GCPArtifactRegistryContainerImage)
```
kunaals: this only queries ContainerImage nodes which have manifest list digests. for multi-arch images, trivy scans one platform (linux/amd64) and reports that platform's digest, which lives on GCPArtifactRegistryPlatformImage. also you're ignoring the `tags` field entirely - if trivy reports with a tag uri, there's nothing to match. need to query both node types and include tags like ecr does.

cartography/intel/trivy/__init__.py
```diff
@@ -19,12 +20,12 @@
 stat_handler = get_stats_client("trivy.scanner")
 
 
-def _get_scan_targets_and_aliases(
+def _get_ecr_scan_targets_and_aliases(
     neo4j_session: Session,
     account_ids: list[str] | None = None,
 ) -> tuple[set[str], dict[str, str]]:
```
kunaals, line 28: this adds digest-qualified uris like `image@sha256:abc`, but trivy might report tag-based uris like `image:latest` in ArtifactName or RepoTags. those won't match because you're not querying or mapping tags.

cartography/intel/trivy/__init__.py
```diff
@@ -19,12 +20,12 @@
...
     """
-    Return tag URIs and a mapping of digest-qualified URIs to tag URIs.
+    Return ECR tag URIs and a mapping of digest-qualified URIs to tag URIs.
     """
```
kunaals, line 31: this creates a malformed uri. gcp's `uri` already includes the digest (e.g. `image@sha256:abc`), so this produces `image@sha256:abc@sha256:abc`. compare to ecr on lines 45-46 which strips the tag first with rsplit. even if fixed, this doesn't help with tag matching since gcp query doesn't return tags.

cartography/models/trivy/findings.py
```diff
@@ -53,6 +53,17 @@ class TrivyFindingToImageRel(CartographyRelSchema):
     properties: TrivyFindingToImageRelProperties = TrivyFindingToImageRelProperties()
 
 
+@dataclass(frozen=True)
+class TrivyFindingToGCPImageRel(CartographyRelSchema):
+    target_node_label: str = "GCPArtifactRegistryContainerImage"
```
kunaals, line 66: this matches on ContainerImage.digest which is the manifest list digest for multi-arch images. trivy reports platform-specific digests, so these relationships will silently fail to create for any multi-arch image. need to also link to GCPArtifactRegistryPlatformImage.

cartography/models/trivy/package.py
```diff
@@ -39,6 +39,17 @@ class TrivyPackageToImageRel(CartographyRelSchema):
     properties: TrivyPackageToImageRelProperties = TrivyPackageToImageRelProperties()
 
 
+@dataclass(frozen=True)
+class TrivyPackageToGCPImageRel(CartographyRelSchema):
+    target_node_label: str = "GCPArtifactRegistryContainerImage"
```
kunaals, line 47: same issue as findings.py - packages won't link to multi-arch gcp images because you're matching manifest list digest instead of platform-specific digest that trivy actually reports.

tests/data/trivy/trivy_gcp_sample.py
```diff
@@ -0,0 +1,110 @@
...
+# sha256:abc123
+TRIVY_GCP_SAMPLE = {
+    "SchemaVersion": 2,
+    "CreatedAt": "2025-05-17T13:51:07.592255-07:00",
+    "ArtifactName": "us-central1-docker.pkg.dev/test-project/docker-repo/my-app:latest",
+    "ArtifactType": "container_image",
+    "Metadata": {
+        "Size": 50000000,
+        "OS": {"Family": "debian", "Name": "12.8"},
+        "ImageID": "sha256:abc123",
```
kunaals: this uses the same digest (`sha256:abc123`) as the ContainerImage in artifact_registry.py test data. in real multi-arch scenarios these would be different digests and your relationships would fail silently. need different digests to actually test the matching logic.

tests/data/trivy/trivy_gcp_sample.py
```diff
@@ -0,0 +1,110 @@
...
+        "DiffIDs": [
+            "sha256:layer1111222333444555666777888999000aaabbbcccdddeeefff00011122",
+        ],
+        "RepoTags": [
+            "us-central1-docker.pkg.dev/test-project/docker-repo/my-app:latest",
+            "us-central1-docker.pkg.dev/test-project/docker-repo/my-app:v1.0.0",
+        ],
+        "RepoDigests": [
+            "us-central1-docker.pkg.dev/test-project/docker-repo/my-app@sha256:abc123"
```
kunaals: same digest again in RepoDigests. the test passes but masks the multi-arch bug where manifest list digest != platform-specific digest.

tests/integration/cartography/intel/trivy/test_trivy_gcp_disk.py
```diff
@@ -0,0 +1,133 @@
...
+    "builtins.open",
+    new_callable=mock_open,
+    read_data=json.dumps(TRIVY_GCP_SAMPLE),
+)
+@patch(
+    "cartography.intel.trivy.get_json_files_in_dir",
+    return_value={"/tmp/scan.json"},
+)
+@patch(
+    "cartography.intel.gcp.artifact_registry.manifest.get_all_manifests_async",
```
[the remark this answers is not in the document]
jychp: OK here, we patch the function from another module.

PR #2299, opened 2026-01-27
fix(gcp): Fix GCP Sync and cascade delete

### LINKED ISSUE #2298: GCP modules crash on 'API not enabled' errors instead of gracefully skipping
### Expected behavior

When a GCP API is not enabled on a project (or quota project), Cartography should log a warning and skip that module gracefully, allowing other modules to continue syncing. This is how `gcf.py` (Cloud Functions) currently works.

### Actual behavior

Cartography crashes with an unhandled `HttpError 403` and stops the entire GCP sync stage, preventing other modules from running.

### To Reproduce

1. Set `GOOGLE_CLOUD_QUOTA_PROJECT` to a project that does NOT have Cloud SQL Admin API enabled
2. Run Cartography with GCP module against an org that has a project with CloudSQL
3. Observe crash when CloudSQL sync attempts to run

### Logs

```
ERROR:cartography.sync:Unhandled exception during sync stage 'gcp'
  File ".../cartography/intel/gcp/cloud_sql_instance.py", line 27, in get_sql_instances
    response = request.execute()
  ...
googleapiclient.errors.HttpError: <HttpError 403 when requesting 
https://sqladmin.googleapis.com/sql/v1beta4/projects/<REDACTED>/instances?alt=json 
returned "Cloud SQL Admin API has not been used in project <REDACTED> before or it is disabled. 
Enable it by visiting https://console.developers.google.com/apis/api/sqladmin.googleapis.com/overview?project=<REDACTED> 
then retry.">
```

- Cartography release version: master (commit f4a4dabc)
- Python version: 3.10.16
- OS: macOS Darwin 25.2.0

### Additional context

**Root cause:** Some modules (like `cloud_sql_instance.py`) only catch `PermissionDenied` exceptions, but "API not enabled" errors come as `HttpError` with status 403.

**Existing pattern:** `gcf.py` handles this correctly by catching `HttpError` and checking for "API has not been used" in the message, then returning `None` to skip sync.

**Suggested fix:** 
1. Create a shared utility for detecting "API not enabled" errors
2. Audit and update affected modules to handle this gracefully
3. Return `None` to skip sync rather than crashing

**Modules to audit:**
- `cloud_sql_instance.py` (confirmed affected)
- `compute.py`, `gke.py`, `bigquery.py`, `dns.py`, `pubsub.py`, etc.

shyammukund:
### Type of change
- [x] Documentation update

### Summary

**Problem 1: GCP syncs crash when APIs aren't enabled on the quota project**

When syncing multiple GCP projects, Cartography would crash with "API has not been used in project X" errors. This happened because GCP checks API enablement against the *quota project* (where the service account lives), not the target project being scanned. Users had no guidance on which APIs to enable or where.

**Problem 2: One failed instance kills cleanup for all instances**

The `had_failures` pattern in multi-instance modules (Bigtable clusters, Cloud SQL databases, etc.) would skip cleanup for ALL resources if ANY single instance failed. This caused stale data to accumulate silently - if 1 of 100 Bigtable instances failed, we'd preserve stale data for all 100.

**Problem 3: Cloud SQL client missing credentials**

The Cloud SQL Admin API client was being built without passing credentials, causing it to fall back to ADC resolution instead of using the explicitly provided credentials.

**Problem 4: Child resource cleanup skipped when parent list is empty**

Bigtable backups and Vertex AI deployed models cleanup was skipped when their parent lists were empty (truthy check instead of `is not None`). This left stale nodes when all parent resources were deleted.

**Problem 5: Orphaned resources when GCP projects are deleted**

When a GCP project was deleted between syncs, its child resources (instances, buckets, etc.) would remain as orphans since resource cleanup is scoped to PROJECT_ID and we only sync existing projects.

**Problem 6: cascade_delete looked in wrong direction for children**

The `cascade_delete` feature in `cleanupbuilder.py` had inverted direction logic - it looked in the same direction as the `sub_resource_relationship` instead of the opposite direction. For INWARD sub_resource relationships (the standard pattern), this meant it looked for `(child)-[:REL]->(node)` instead of `(node)-[:REL]->(child)`, causing cascade_delete to fail silently for all modules using the standard RESOURCE relationship pattern.

### Changes

1. **Documentation**: Added "API Enablement Requirements" section explaining that APIs must be enabled on the host/quota project, with copy-paste `gcloud` commands
2. **Improved error detection**: `is_api_disabled_error()` now checks the `reason` field first (`accessNotConfigured`, `SERVICE_DISABLED`) before falling back to message patterns. Explicitly rejects IAM-related reasons (`forbidden`, `insufficientPermissions`) so permission issues still fail loudly.
3. **Removed `had_failures` pattern**: Multi-instance modules now always run cleanup - partial failures don't prevent cleaning up successfully synced data
4. **Fixed Cloud SQL credentials**: Pass credentials explicitly to `build_client()`
5. **Fixed child resource cleanup**: Bigtable backups and Vertex AI deployed models now always run cleanup when parent sync succeeds, even if parent list is empty
6. **Added cascade_delete**: GCPProject, GCPFolder, and GCPOrganization cleanup jobs now use `cascade_delete=True` to automatically delete orphaned child nodes when parents are deleted
7. **Fixed cascade_delete direction logic**: Inverted the relationship direction in `cleanupbuilder.py` when looking for children - INWARD sub_resource now looks OUTWARD for children, and vice versa

### Related issues or links

- Fixes #2298

### How was this tested?

**Live testing against real GCP APIs:**
- Compute API (disabled): Returns `reason: accessNotConfigured` → correctly detected
- Bigtable/Cloud Functions (disabled): Returns empty errors array with message pattern → correctly detected via fallback
- Storage API (enabled, permission denied): Returns `reason: forbidden` → correctly NOT identified as API disabled

**Unit tests:**
- Added 15 unit tests for `is_api_disabled_error()` covering all detection scenarios
- Updated cascade_delete unit tests to expect correct direction in generated queries

**Integration testing:**
Ran GCP sync locally against multiple projects with mixed API enablement:

```
INFO:cartography.intel.gcp:Syncing Cloud Run Services for project test-cloud-run-XXXXX.
DEBUG:cartography.graph.job:Starting job 'Cleanup GCPCloudRunService'.
INFO:cartography.graph.job:Finished job GCPCloudRunService

WARNING:cartography.intel.gcp.bigtable_instance:Could not retrieve Bigtable instances on project test-cloud-run-XXXXX due to permissions issues or API not enabled. Skipping sync to preserve existing data.

WARNING:cartography.intel.gcp.cloud_sql_instance:Could not retrieve Cloud SQL instances on project test-cloud-run-XXXXX due to permissions issues or API not enabled. Skipping sync to preserve existing data.
```

- Cloud Run (API enabled): Syncs and runs cleanup ✓
- Bigtable/Cloud SQL (API disabled): Logs warning, skips sync, preserves existing data ✓
- All unit tests pass
- Linting passes

**cascade_delete testing:**
- 5 unit tests verify correct query generation with inverted direction
- 4 integration tests verify cascade deletion works end-to-end
- 4 GCP-specific cascade delete tests verify real-world scenarios

#### General

#### Proof of functionality

### Notes for reviewers

The key insight is that GCP's quota project architecture means APIs must be enabled on the *host project* (where the service account was created), not necessarily on each target project. The documentation update makes this clear to users.

The cleanup behavior change is intentional: stale data accumulation is worse than losing data when access is lost. For single-resource modules we preserve data on API errors; for multi-instance modules we now clean up what we successfully synced.

**New in this update:**
- `is_api_disabled_error()` now uses the more reliable `reason` field from GCP error responses, with message pattern fallback for APIs that use different formats
- Fixed edge case where child resource cleanup (backups, deployed models) was skipped when parent list was empty
- Added `cascade_delete=True` to hierarchical cleanup to handle orphaned resources when projects/folders/orgs are deleted between syncs
- **Fixed cascade_delete direction bug**: The cascade logic was looking in the wrong direction for children. For standard INWARD RESOURCE relationships (parent points to child), we need to look OUTWARD from the node to find children. The previous code looked INWARD, which would never find children. This fix enables cascade_delete to work correctly for GCP, GitLab, and any other module using the standard relationship pattern.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,6 +22,7 @@ Detailed procedures are available in separate documents:
 | [Enriching the Ontology](docs/agents/enrich-ontology.md) | Adding ontology mappings for cross-module querying |
 | [Adding a New Node Type](docs/agents/add-node-type.md) | Advanced node schema properties and configurations |
 | [Adding a New Relationship](docs/agents/add-relationship.md) | Relationships, MatchLinks, and multi-module patterns |
+| [Adding Analysis Jobs](docs/agents/analysis-jobs.md) | Post-ingestion graph enrichment and cross-resource analysis |
 | [Creating Security Rules](docs/agents/create-rule.md) | Security rules, facts, and compliance conventions |
 | [Refactoring Legacy Code](docs/agents/refactor-legacy.md) | Converting legacy Cypher to modern data model |
 | [Troubleshooting](docs/agents/troubleshooting.md) | Common errors, debugging tips, and key files reference |
@@ -30,9 +31,10 @@ Detailed procedures are available in separate documents:
 
 **Key Cartography Concepts:**
 - **Intel Module**: Component that fetches data from external APIs and loads into Neo4j
-- **Sync Pattern**: `get()` -> `transform()` -> `load()` -> `cleanup()`
+- **Sync Pattern**: `get()` -> `transform()` -> `load()` -> `cleanup()` -> `analysis` (optional)
 - **Data Model**: Declarative schema using `CartographyNodeSchema` and `CartographyRelSchema`
 - **Update Tag**: Timestamp used for cleanup jobs to remove stale data
+- **Analysis Jobs**: Post-ingestion queries that enrich the graph (e.g., internet exposure, permission inheritance)
 
 **Critical Files to Know:**
 - `cartography/config.py` - Configuration object definitions
@@ -55,6 +57,9 @@ from cartography.models.core.relationships import (
 from cartography.client.core.tx import load, load_matchlinks
 from cartography.graph.job import GraphJob
 from cartography.util import timeit
+
+# For analysis jobs (optional)
+from cartography.util import run_analysis_job, run_scoped_analysis_job, run_analysis_and_ensure_deps
 
 logger = logging.getLogger(__name__)
 ```
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -343,6 +343,14 @@ def _perform_aws_analysis(
         neo4j_session,
     )
 
+    run_analysis_and_ensure_deps(
+        "aws_ecs_asset_exposure.json",
+        {"ecs", "ec2:load_balancer_v2"},
+        requested_syncs_as_set,
+        common_job_parameters,
+        neo4j_session,
+    )
+
 
 @timeit
 def start_aws_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
--- a/cartography/intel/aws/cloudtrail.py
+++ b/cartography/intel/aws/cloudtrail.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 @timeit
 @aws_handle_regions
 def get_cloudtrail_trails(
-    boto3_session: boto3.Session, region: str
+    boto3_session: boto3.Session, region: str, current_aws_account_id: str
 ) -> List[Dict[str, Any]]:
     client = boto3_session.client(
         "cloudtrail", region_name=region, config=get_botocore_config()
@@ -29,14 +29,32 @@ def get_cloudtrail_trails(
     trails = client.describe_trails()["trailList"]
     trails_filtered = []
     for trail in trails:
-        if trail.get("HomeRegion") == region:
-            selectors = client.get_event_selectors(TrailName=trail["TrailARN"])
-            trail["EventSelectors"] = selectors.get("EventSelectors", [])
-            trail["AdvancedEventSelectors"] = selectors.get(
-                "AdvancedEventSelectors",
-                [],
-            )
-            trails_filtered.append(trail)
+        # Filter by home region to avoid duplicates across regions
+        if trail.get("HomeRegion") != region:
+            continue
+
+        # Filter to only trails owned by this account.
+        # Organization trails from other accounts are visible via describe_trails()
+        # but should not be linked as RESOURCE of this account.
+        # ARN format: arn:aws:cloudtrail:{region}:{account_id}:trail/{name}
+        trail_arn = trail.get("TrailARN", "")
+        arn_parts = trail_arn.split(":")
+        if len(arn_parts) >= 5:
+            trail_account_id = arn_parts[4]
+            if trail_account_id != current_aws_account_id:
+                logger.debug(
+                    f"Skipping trail {trail_arn} - owned by account {trail_account_id}, "
+                    f"not current account {current_aws_account_id}",
+                )
+                continue
+
+        selectors = client.get_event_selectors(TrailName=trail["TrailARN"])
+        trail["EventSelectors"] = selectors.get("EventSelectors", [])
+        trail["AdvancedEventSelectors"] = selectors.get(
+            "AdvancedEventSelectors",
+            [],
+        )
+        trails_filtered.append(trail)
 
     return trails_filtered
 
@@ -105,7 +123,9 @@ def sync(
         logger.info(
             f"Syncing CloudTrail for region '{region}' in account '{current_aws_account_id}'.",
         )
-        trails_filtered = get_cloudtrail_trails(boto3_session, region)
+        trails_filtered = get_cloudtrail_trails(
+            boto3_session, region, current_aws_account_id
+        )
         trails = transform_cloudtrail_trails(trails_filtered, region)
 
         load_cloudtrail_trails(
--- a/cartography/intel/gsuite/__init__.py
+++ b/cartography/intel/gsuite/__init__.py
@@ -18,6 +18,7 @@ from googleapiclient.discovery import Resource
 from cartography.config import Config
 from cartography.intel.gsuite import groups
 from cartography.intel.gsuite import users
+from cartography.util import run_analysis_job
 from cartography.util import timeit
 
 OAUTH_SCOPES = [
@@ -168,3 +169,10 @@ def start_gsuite_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
             config.update_tag,
             scoped_job_parameters,
         )
+
+    # Link GSuite users to Human nodes if they exist
+    run_analysis_job(
+        "gsuite_human_link.json",
+        neo4j_session,
+        common_job_parameters,
+    )
--- a/cartography/intel/keycloak/__init__.py
+++ b/cartography/intel/keycloak/__init__.py
@@ -14,6 +14,7 @@ import cartography.intel.keycloak.roles
 import cartography.intel.keycloak.scopes
 import cartography.intel.keycloak.users
 from cartography.config import Config
+from cartography.util import run_analysis_job
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
@@ -151,3 +152,11 @@ def start_keycloak_ingestion(neo4j_session: neo4j.Session, config: Config) -> No
                     config.keycloak_url,
                     realm_scopped_job_parameters,
                 )
+
+        # Run inheritance analysis after all realms are synced
+        # This computes transitive group memberships, role assignments, and scope permissions
+        run_analysis_job(
+            "keycloak_inheritance.json",
+            neo4j_session,
+            common_job_parameters,
+        )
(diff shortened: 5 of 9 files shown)


cartography/intel/gcp/bigtable_app_profile.py
```diff
@@ -2,32 +2,52 @@
...
+    except HttpError as e:
+        if is_api_disabled_error(e):
+            logger.warning(
+                "Could not retrieve Bigtable app profiles for instance %s due to permissions "
+                "issues or API not enabled. Skipping.",
+                instance_id,
             )
-        )
-    return app_profiles
+            return None
```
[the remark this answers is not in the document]
shyammukund: The entire point of this PR is to gracefully handle "API not enabled" errors by returning None instead of crashing the entire sync. This is the established pattern from gcf.py (lines 
  50-65) which already works this way. We only return None for specific permission/API-disabled errors, and we do re-raise all other HttpError exceptions.

cartography/intel/gcp/bigtable_app_profile.py
```diff
@@ -88,9 +108,11 @@ def sync_bigtable_app_profiles(
     for inst in instances:
         instance_id = inst["name"]
         app_profiles_raw = get_bigtable_app_profiles(client, instance_id)
-        all_app_profiles_transformed.extend(
-            transform_app_profiles(app_profiles_raw, instance_id),
-        )
+        # Skip this instance if API is not enabled or access denied
+        if app_profiles_raw is not None:
```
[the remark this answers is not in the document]
shyammukund: we check if data is not None: before running cleanup. When we return None for API-disabled errors, we skip both load and cleanup, preserving existing data. This is 
  exactly the pattern from gcf.py lines 167-176

cartography/intel/gcp/bigtable_cluster.py
```diff
@@ -2,32 +2,55 @@
 
 import neo4j
 from googleapiclient.discovery import Resource
+from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
```
kunaals, line 8: this check handles the per-instance skip, but cleanup still runs unconditionally after the loop. if any `get_bigtable_clusters` call returns `None`, we'll still wipe all previously ingested cluster data for the project. might want to track a `skip_cleanup` flag when any call returns `None` and conditionally skip the cleanup.

cartography/intel/gcp/bigtable_table.py
```diff
@@ -2,32 +2,55 @@
 
 import neo4j
 from googleapiclient.discovery import Resource
+from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
```
kunaals, line 7: same issue - if `get_bigtable_tables` returns `None` for some instances due to permission errors, cleanup still runs and will delete tables we couldn't refresh.

cartography/intel/gcp/bigtable_app_profile.py
```diff
@@ -2,32 +2,55 @@
 
 import neo4j
 from googleapiclient.discovery import Resource
+from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.gcp.util import gcp_api_execute_with_retry
```
kunaals, line 9: cleanup is unconditional after this loop. if we can't fetch app profiles for some instances, this will delete previously ingested data we intended to preserve.

cartography/intel/gcp/bigtable_backup.py
```diff
@@ -2,33 +2,58 @@
 
 import neo4j
 from googleapiclient.discovery import Resource
+from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
```
kunaals, line 7: same pattern - `get_bigtable_backups` returning `None` doesn't prevent cleanup from running. previously ingested backups will get deleted even though we couldn't verify if they still exist.

cartography/intel/gcp/cloud_sql_database.py
```diff
@@ -2,9 +2,12 @@
 
 import neo4j
 from googleapiclient.discovery import Resource
+from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
```
kunaals, line 7: cleanup runs regardless of whether any `get_sql_databases` calls returned `None`. if a project has instances but lacks permission to list databases on some of them, this will delete all previously synced database records.

cartography/intel/gcp/cloud_sql_user.py
```diff
@@ -2,25 +2,48 @@
 
 import neo4j
 from googleapiclient.discovery import Resource
+from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
```
kunaals, line 7: same data-loss risk - unconditional cleanup after `get_sql_users` can return `None` on permission errors. consider tracking skipped instances and skipping cleanup if we couldn't refresh all data.

cartography/intel/gcp/artifact_registry/artifact.py
```diff
@@ -1,13 +1,13 @@
 import logging
 
 import neo4j
-from google.api_core.exceptions import PermissionDenied
```
kunaals, line 4: cleanup functions run unconditionally at the end of this function, but the `get_*` functions can return `None` on api disabled/permission errors. this means if we couldn't fetch artifacts for some repos, we'll still delete all existing artifact data for the project.

kunaals:
to add to my earlier comments - there's actually a more severe case here: the `return instances_raw or []` pattern in the parent syncs loses the "api disabled" signal.

for example:
1. `sync_bigtable_instances` can't fetch instances (api disabled), returns `[]`
2. `sync_bigtable_clusters` is called with `instances=[]`
3. the for loop doesn't iterate, nothing loads
4. cleanup runs unconditionally → all existing clusters get deleted

a cleaner fix might be to either:
- preserve `None` vs `[]` distinction in return types and have callers skip sub-resource syncs when parent returns `None`
- or in sub-resource syncs, skip cleanup if both `instances` is empty AND nothing was loaded (indicating we couldn't verify anything)

the partial failure case (some instances return `None`) could use a `had_failures` flag to skip cleanup when we couldn't refresh everything.

shyammukund:
- Preserved the none return value in parent sync s to allow callers to distinguish between 
         - None - API disabled couldn't check 
         - [] - API worked, but there aren't any instances to ingest
-  Now we do an explicit None check before we call the resource function to know whether to cleanup sub-resources or whether to skip 
- Added had_failures flag

tests/integration/cartography/intel/gitlab/test_cascade_delete.py
```diff
@@ -0,0 +1,309 @@
...
+        # Verify initial state
+        assert len(check_nodes(neo4j_session, "GitLabProject", ["id"])) == 1
+        assert len(check_nodes(neo4j_session, "GitLabBranch", ["id"])) == 2
+
+        # Run cleanup with cascade_delete=True
+        common_job_params = {
+            "UPDATE_TAG": TEST_UPDATE_TAG_V2,
+            "org_url": TEST_ORG_URL,
+        }
+        GraphJob.from_node_schema(
```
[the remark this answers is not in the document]
kunaals: The goal here is to simulate stale data without hitting GitLab at all, so calling the full sync would force more API/mocking plumbing. The existing pattern (GraphJob directly) is also repeated in the GCP cascade tests and keeps the focus on cascade logic, so this isn’t a violation of an “integration-test-via-sync” rule (no such rule exists elsewhere in the repo). The test is still integration-level because it uses a real Neo4j session and the actual cleanup job.

tests/integration/cartography/intel/gcp/test_cascade_delete.py
```diff
@@ -0,0 +1,370 @@
...
+        creds.quota_project_id = "test-quota-project"
+        return creds
+
+    @patch.object(
+        cartography.intel.gcp,
+        "get_gcp_credentials",
+    )
+    @patch.object(
+        cartography.intel.gcp,
+        "_sync_project_resources",
```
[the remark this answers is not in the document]
kunaals: The test is already hitting the real sync entrypoint via cartography.intel.gcp.start_gcp_ingestion(...), and it mocks only the API-layer helpers (get_gcp_organizations, get_gcp_folders, get_gcp_projects). _sync_project_resources drives internal resource syncing, which in a full run would try to call additional GCP APIs and populate many nodes; the test only cares about cascade delete, so it patches _sync_project_resources to skip that expensive path while still letting the rest of the sync (and the cascade cleanup job) run naturally. That keeps the integration test focused and deterministic. Theatre around “mock only external APIs” seems aimed at avoiding mocking Cartography internals when you can let them run, but here the internal start_gcp_ingestion already runs, and this single patch just silences resource syncing, so the integration coverage is still intact.

kunaals (approved):

PR #2302, opened 2026-01-28
feat(gitlab): Add GitLab User Coverage

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- Adds in coverage for GitLab Users into the GitLab module allowing visualization of which users are apartof which groups and which repos that they have commited to. 

### How was this tested?

- included integration test coverage and ran sync against gitlab enviornment

#### General

#### Proof of functionality

image


cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,411 @@
...
+    users_by_url = {record["web_url"]: record for record in user_records}
+
+    # Aggregate: (user_url, project_url) -> list of commit dates
+    activity: dict[tuple[str, str], list[str]] = {}
+
+    for project_url, commits in commits_by_project.items():
+        for commit in commits:
+            committed_date = commit.get("committed_date")
+            author_name = commit.get("author_name")
+
```
kunaals, line 128: matching commits to users by display name is pretty fragile - git author names are user-configured and can differ from gitlab profile names (e.g. "John Doe" in gitlab but "jdoe" in git config). also if two users have the same display name, one will silently overwrite the other in the users_by_name dict. could we try matching by author_email first when it's available, and fall back to name? or at least log a warning when we see duplicate names?

cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,411 @@
...
+    If a user is a member of multiple ancestor groups, only the highest access level is returned.
+    """
+    logger.debug(f"Fetching members for group ID {group_id}")
+    members = get_paginated(gitlab_url, token, f"/api/v4/groups/{group_id}/members/all")
+    logger.debug(f"Fetched {len(members)} members for group ID {group_id}")
+    return members
+
+
+def get_commits(gitlab_url: str, token: str, project_id: int) -> list[dict[str, Any]]:
+    """
```
kunaals, line 59: this fetches the entire commit history for every project on each sync run. for orgs with lots of projects or long-lived repos this is going to be really slow and could hit api rate limits. the gitlab commits api supports a `since` param - might be worth adding that to only pull recent commits (last 90 days or so), or making it configurable.

cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,411 @@
...
+    logger.info(
+        f"Transformed {len(user_memberships)} unique users: "
+        f"{users_without_groups} org-only, {total_memberships} group memberships"
+    )
+    return records
+
+
+@timeit
+def load_users(
+    neo4j_session: neo4j.Session,
```
kunaals, line 307: using the same GitLabUserSchema for both membership records and commit activity records is a bit awkward - the schema has both MEMBER_OF and COMMITTED_TO in OtherRelationships, but each load call only provides data for one of them. it works because cartography skips relationships when the PropertyRef value is missing, but it's not obvious from reading the code. might be cleaner to either split into two schemas, or add a comment explaining why this is intentional.

tests/integration/cartography/intel/gitlab/test_users.py
```diff
@@ -0,0 +1,154 @@
...
+        check_rels(
+            neo4j_session,
+            "GitLabUser",
+            "id",
+            "GitLabProject",
+            "id",
+            "COMMITTED_TO",
+        )
+        == expected_commits
+    )
```
kunaals, line 168: the test checks that the COMMITTED_TO relationship exists but doesn't verify the properties on it (commit_count, first_commit_date, last_commit_date). would be good to assert those values too so we catch regressions if the transform logic changes.

tests/integration/cartography/intel/gitlab/test_users.py
```diff
@@ -0,0 +1,154 @@
...
+    neo4j_session.run("MATCH (u:GitLabUser) DETACH DELETE u")
+
+    _create_test_organization(neo4j_session)
+    _create_test_group(neo4j_session)
+    _create_test_project(neo4j_session)
+
+    # Mock API responses
+    mock_get_organization.return_value = {"web_url": TEST_ORG_URL, "name": "myorg"}
+    mock_get_group_members.return_value = GET_GITLAB_GROUP_MEMBERS
+    mock_get_commits.return_value = GET_GITLAB_COMMITS
```
kunaals, line 86: the mock for get_group_members returns the same data for both the org-level call and the group-level call, so we're not really testing the distinction between org members vs group members. might want to use side_effect to return different data for each call, or split the org member fetch into its own function that can be mocked separately.

tests/data/gitlab/users.py
```diff
@@ -0,0 +1,49 @@
+"""Test data for GitLab users module."""
+
+TEST_ORG_URL = "https://gitlab.example.com/myorg"
+
+# Single user for simple testing
+GET_GITLAB_ORG_MEMBERS = [
```
kunaals, line 7: GET_GITLAB_ORG_MEMBERS is defined here but never used in the tests - the test relies on GET_GITLAB_GROUP_MEMBERS being returned for both org and group member fetches. either use this data or remove it to avoid confusion.

kunaals (commented):
the overall structure follows cartography patterns well. left a few comments on potential improvements - mainly around commit fetching performance and author matching reliability.

cartography/intel/gitlab/users.py
shyammukund: So this was intentional because the only thing we use the commit data for is to create the rel between users and the repos they commit to. Therefore it made sense to me to ingest commit activity with users.

cartography/intel/gitlab/users.py
shyammukund: Decided to match by email then fallback to author name and log a warning when we see duplicate names

cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,483 @@
...
+
+    logger.info(
+        f"Commit author matching: {email_matches} by email, "
+        f"{name_matches} by name, {no_matches} unmatched"
+    )
+
+    # Build records by taking existing user data and adding commit properties
+    records = []
+    for (user_url, project_url), dates in activity.items():
+        dates.sort()
```
kunaals: `dates.sort()` sorts strings - breaks when timezones differ (e.g. `2024-01-01T02:00:00-08:00` vs `2024-01-01T10:00:00Z`). see `github/commits.py:303-306` for the pattern: parse to datetime with `fromisoformat()`, use `min()`/`max()`

cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,483 @@
...
+
+    # Fetch members for all descendant groups
+    group_members_by_group: dict[str, list[dict[str, Any]]] = {}
+    for group in groups:
+        group_id = group.get("id")
+        group_url = group.get("web_url")
+        if not group_id or not group_url:
+            continue
+
+        members = get_group_members(gitlab_url, token, group_id)
```
kunaals: wrap in try/except, log warning, continue - one bad group (403/404) shouldn't kill the entire sync. same pattern as `github/commits.py:188-192`

cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,483 @@
...
+
+    # Fetch commits for all projects to build commit activity
+    commits_by_project: dict[str, list[dict[str, Any]]] = {}
+    for project in projects:
+        project_id = project.get("id")
+        project_url = project.get("web_url")
+        if not project_id or not project_url:
+            continue
+
+        commits = get_commits(gitlab_url, token, project_id, commits_since_days)
```
kunaals: same here for commit fetches - catch httperror, log, continue

cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,483 @@
...
+            # Check for duplicate names
+            if name in users_by_name and users_by_name[name] != web_url:
+                duplicate_names.add(name)
+            users_by_name[name] = web_url
+
+    # Warn about duplicate names (silent data loss risk)
+    if duplicate_names:
+        logger.warning(
+            f"Found {len(duplicate_names)} users with duplicate display names: "
+            f"{', '.join(sorted(duplicate_names))}. "
```
kunaals: logging actual names is pii - just log the count instead

cartography/intel/gitlab/users.py
```diff
@@ -0,0 +1,483 @@
...
+    total_memberships = len(records) - users_without_groups
+    logger.info(
+        f"Transformed {len(user_memberships)} unique users: "
+        f"{users_without_groups} org-only, {total_memberships} group memberships"
+    )
+    return records
+
+
+@timeit
+def load_users(
```
kunaals, line 306: nit: add brief comment explaining why the same schema is used for both membership and commit loads (cartography skips rels when propertyref is missing) - not obvious to future readers

PR #2306, opened 2026-01-29
test(gcp): add integration tests for empty parent cleanup behavior

kunaals:
## Summary
- Adds integration tests for the "empty parent cleanup" fix from PR #2299
- Tests that cleanup runs correctly even when parent resource lists are empty
- Covers Bigtable backups, clusters, and Vertex AI deployed models

**Note**: This test file was accidentally omitted from PR #2299 which implemented the fix being tested here.


tests/integration/cartography/intel/gcp/test_empty_parent_cleanup.py
```diff
@@ -0,0 +1,276 @@
...
+def _create_gcp_project(neo4j_session, project_id: str, update_tag: int):
+    """Create a GCPProject node for testing."""
+    neo4j_session.run(
+        "MERGE (p:GCPProject {id: $project_id}) SET p.lastupdated = $tag",
+        project_id=project_id,
+        tag=update_tag,
+    )
+
+
+class TestBigtableBackupCleanupWithEmptyClusters:
```
jychp: Classes are not needed here and don't follow the pattern of other tests files, can we remove it ?

jychp (changes requested):

tests/integration/cartography/intel/gcp/test_empty_parent_cleanup.py
kunaals: good catch. fixed this, and added caller-level coverage for the empty-parent regression path.

AGENTS.md
```diff
@@ -272,6 +272,11 @@ assert check_rels(
 ) == expected_rels
 ```
 
+### Integration Test Boundary
+
+- Integration tests may seed prerequisite graph state with Cypher, but should exercise real Cartography `sync_*`, `load()`, and cleanup functions whenever practical.
+- Prefer mocking only external boundaries such as API clients, service discovery, credentials, and network responses. Mock Cartography sync functions only for unit-style caller-wiring tests.
```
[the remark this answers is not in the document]
kunaals: good catch. tightened this so the integration-test guidance points at real sync flows, not load-level tests.

tests/integration/cartography/intel/gcp/test_empty_parent_cleanup.py
```diff
@@ -0,0 +1,318 @@
...
+
+    This covers the project-level orchestration path where truthy checks on
+    instances_raw would skip child cleanup when all instances were deleted.
+    """
+    neo4j_session.run("MATCH (n) DETACH DELETE n")
+    _create_gcp_project(neo4j_session, TEST_PROJECT_ID, TEST_UPDATE_TAG)
+    _create_stale_project_resource(
+        neo4j_session,
+        "GCPBigtableCluster",
+        "projects/test-project/instances/inst/clusters/old-cluster",
```
jychp, line 66: Nitpick: cluster id `projects/test-project/instances/inst/clusters/old-cluster` and the stale backup id at line 71 reference the same instance `inst` but unrelated cluster paths. Either make the backup live under `clusters/old-cluster/backups/...` so the fixture is internally consistent, or drop the contrived ids and use shorter values — the current pair reads as if they were related when they are not.

jychp (approved):
LGTM

tests/integration/cartography/intel/gcp/test_empty_parent_cleanup.py
kunaals: good catch. made the stale backup path use the same old-cluster fixture.

PR #2331, opened 2026-02-02
feat(metrics): add resource count metrics to load operations

jychp:
### Type of change

### Summary

This PR adds StatsD metrics for tracking nodes and relationships loaded during sync operations. This improves observability by providing counts of resources ingested per type.

**New metrics emitted:**
- `node.{label}.loaded` - Count of nodes loaded per node type (e.g., `node.awsaccount.loaded`)
- `relationship.{label}.loaded` - Count of relationships loaded per type (e.g., `relationship.resource.loaded`)

**Changes:**
- Added `stat_handler` to `cartography/client/core/tx.py`
- Added metric emission after `load()` and `load_matchlinks()` functions
- Added INFO-level logging showing the count and type of resources loaded

These metrics complement the existing Neo4j counters already emitted in `statement.py` (nodes_deleted, relationships_deleted, etc.) and provide visibility into the ingestion side of the sync process.

### How was this tested?
- Added 4 new unit tests:
  - `test_load_emits_metrics_and_logs`
  - `test_load_no_metrics_for_empty_data`
  - `test_load_matchlinks_emits_metrics_and_logs`
  - `test_load_matchlinks_no_metrics_for_empty_data`
- All existing unit tests pass (38 total)
- Integration tests pass and show the new logging:
  ```
  INFO:cartography.client.core.tx:Loaded 2 CAN_ACCESS relationships
  ```

#### General

#### Proof of functionality

### Notes for reviewers

This is a minimal, focused change that adds metrics to the core load functions. The metrics are only emitted when StatsD is enabled (`statsd_enabled=True`), so there's no performance impact when metrics are disabled.

### DIFF
--- a/cartography/rules/cli.py
+++ b/cartography/rules/cli.py
@@ -35,14 +35,30 @@ class OutputFormat(str, Enum):
 
 
 def complete_rules(incomplete: str) -> Generator[str, None, None]:
-    """Autocomplete rules names."""
+    """
+    Autocomplete rule names for CLI tab completion.
+
+    Args:
+        incomplete (str): The partial rule name typed by the user.
+
+    Yields:
+        str: Rule names that start with the incomplete string.
+    """
     for name in RULES.keys():
         if name.startswith(incomplete):
             yield name
 
 
 def complete_rules_with_all(incomplete: str) -> Generator[str, None, None]:
-    """Autocomplete rules names plus 'all'."""
+    """
+    Autocomplete rule names plus 'all' for CLI tab completion.
+
+    Args:
+        incomplete (str): The partial rule name typed by the user.
+
+    Yields:
+        str: Rule names (plus 'all') that start with the incomplete string.
+    """
     for name in builtins.list(RULES.keys()) + ["all"]:
         if name.startswith(incomplete):
             yield name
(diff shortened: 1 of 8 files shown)


tests/unit/cartography/client/core/test_tx.py
```diff
@@ -672,3 +672,148 @@ def test_buffer_error_with_none_wait_time(mock_logger, mock_sleep):
...
+    mock_node_schema.label = "TestNode"
+    mock_build_query.return_value = "UNWIND ..."
+    mock_build_cond_labels.return_value = []
+
+    test_data = [{"id": "1"}, {"id": "2"}, {"id": "3"}]
+
+    load(mock_session, mock_node_schema, test_data, lastupdated=12345)
+
+    # Verify metric was emitted with correct count
+    mock_stat_handler.incr.assert_called_once_with("node.testnode.loaded", 3)
```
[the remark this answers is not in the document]
jychp: Not for integration tests.

kunaals (approved):

cartography/client/core/tx.py
```diff
@@ -822,6 +824,11 @@ def load(
     for query in conditional_label_queries:
         run_write_query(neo4j_session, query, **kwargs)
 
+    # Emit metrics for loaded nodes
+    node_count = len(dict_list)
+    stat_handler.incr(f"node.{node_schema.label.lower()}.loaded", node_count)
```
heryxpc, line 829: This a broader question but wondering if two schemas can have the same label as they might exist in different intel modules/packages.
I don't think a change is needed here, but it came through my mind.

cartography/client/core/tx.py
```diff
@@ -882,3 +889,8 @@ def load_matchlinks(
     load_graph_data(
         neo4j_session, matchlink_query, dict_list, batch_size=batch_size, **kwargs
     )
+
+    # Emit metrics for loaded relationships
+    rel_count = len(dict_list)
+    stat_handler.incr(f"relationship.{rel_schema.rel_label.lower()}.loaded", rel_count)
```
heryxpc, line 895: For this I'm sure it will repeat a lot.
Any idea on how to label the relationship to distinguish among different relationships created?

cartography/client/core/tx.py
jychp: Should not, common label are used as extra_label only.

PR #2335, opened 2026-02-03
feat(rules): Add Framework object for structured compliance metadata

jychp:
### Type of change

### Summary

This PR introduces a `Framework` dataclass to link rules to compliance frameworks (CIS, NIST, etc.) with structured fields for filtering and querying.

**Key changes:**
- **Framework class**: New dataclass with `name`, `short_name`, `requirement`, `scope`, and `revision` fields. All fields are normalized to lowercase for case-insensitive comparison.
- **Rule.frameworks**: Rules can now declare compliance framework references via a `frameworks` tuple.
- **Rule.has_framework()**: Method to check if a rule matches given framework criteria.
- **CLI enhancements**:
  - `--framework` / `-f` filter option for `list` and `run` commands
  - New `frameworks` command to list all compliance frameworks
  - Auto-select all rules when `--framework` is provided without a rule argument
- **Runner functions**: `get_all_frameworks()` and `filter_rules_by_framework()` for programmatic filtering.
- **RuleResult**: Now exposes `rule_tags` and `rule_frameworks` for downstream consumers.
- **Migration**: All existing CIS rules migrated to use Framework objects.

**Example usage:**
```bash
# List all CIS rules
cartography-rules list --framework CIS

# Run only CIS AWS 5.0 rules
cartography-rules run --framework CIS:aws:5.0

# List available frameworks
cartography-rules frameworks
```

### How was this tested?

- `make test_unit` - 483 tests passed
- `make test_lint` - All checks passed
- Manual CLI testing with `--framework` filter

#### General

#### Proof of functionality

### Notes for reviewers

This 'tag' approach for linking to a Framework was chosen (rather than the old composition system where a rule only existed for one framework) to offer maximum flexibility and reusability.

This design follows the existing pattern in the codebase and enables downstream systems like SubImage to build rule collections based on compliance frameworks.


cartography/rules/cli.py
```diff
@@ -86,11 +87,99 @@ def complete_facts(
...
 # CLI Commands
 # ----------------------------
 
 
+@app.command(name="frameworks")  # type: ignore[misc]
+def frameworks_cmd() -> None:
+    """
+    List all compliance frameworks referenced by rules.
+
+    \b
```
kunaals, line 142: i think extra character got left behind

jychp: `\b` is the Typer/Click literal block marker.

PR #2346, opened 2026-02-05
chore: add Python 3.13 support with deprecation warning for previous versions

jychp:
### Type of change

- [ ] Documentation update
- [x] Other (please describe): Python 3.13 migration + dependency cleanup

### Summary

Prepares Cartography for Python 3.13 and cleans up `pyproject.toml` dependency declarations.

**Python 3.13 support:**
- Add deprecation warning in `cartography/__init__.py` (`DeprecationWarning` for library imports, `logger.warning` in CLI) for Python < 3.13
- Update CI workflows (`test_suite.yml`, `publish-to-ghcr-and-pypi.yml`) to use Python 3.13
- Update `Dockerfile` and `dev.Dockerfile` to use `python:3.13-slim` base image
- Update `.python-version` to 3.13
- Update `pyproject.toml` classifier to Python 3.13
- Update documentation (`install.md`, `developer-guide.md`)

**Dependency cleanup (`pyproject.toml`):**
- Remove 17 duplicate dependencies between `dependencies` and `dependency-groups.dev`
- Remove stale TODO comment (uv migration already done)
- Normalize PEP 508 syntax (remove parentheses, extra spaces)
- Add loose lower bounds (`>=x.y.0`) to all previously unpinned dependencies
- Bump lower bounds past major version gaps (`marshmallow` 3.0.0rc7→4.0.0, `aioboto3` 13→15, `dnspython` 1→2, `google-api-python-client` 1→2)
- Remove unnecessary upper bounds (`azure-mgmt-resource`, `azure-mgmt-sql`, `cloudflare`, `testcontainers`, `types-requests`)
- Replace exact pins (`==`) with loose lower bounds for dev deps (`black`, `pytest-cov`)
- Add comment documenting dependency versioning convention

### ⚠️ Note on msal prerelease

`azure-cli-core` still pins `msal==1.35.0b1` (a beta) as of v2.84.0. This is an upstream Azure packaging decision — stable `msal` 1.35.0 and 1.35.1 exist on PyPI, but `azure-cli-core` hasn't updated its pin yet. uv resolves this automatically since it's the only version satisfying the transitive constraint. No `--prerelease=allow` flag is needed on our side.

Refs:
- https://pypi.org/pypi/azure-cli-core/2.84.0/json
- https://pypi.org/project/msal/

### Related issues or links

- Refs #2205

### How was this tested?

- `uv lock --check` passes
- CI will validate Python 3.13 compatibility across linter, unit tests, integration tests, and Docker build

#### General

### Notes for reviewers

The previous dependency blocker (`msal` beta / `pymsalruntime` lacking Python 3.13 wheels) is partially resolved — `uv lock` resolves cleanly without `--prerelease=allow`, but `azure-cli-core` still transitively pins `msal==1.35.0b1`. This is an upstream issue with no workaround on our side.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -2439,7 +2439,9 @@ def main(argv=None):
 
     # Show Python version deprecation warning visibly to CLI users.
     # The library-level DeprecationWarning in __init__.py is hidden by default.
-    from cartography import _MIN_PYTHON, _MIN_PYTHON_STR
+    from cartography import _MIN_PYTHON
+    from cartography import _MIN_PYTHON_STR
+
     if sys.version_info < _MIN_PYTHON:
         logger.warning(
             "Cartography is tested on Python %s+ only. "


cartography/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+_MIN_PYTHON = (3, 13)
+_MIN_PYTHON_STR = ".".join(map(str, _MIN_PYTHON))
+
+if sys.version_info < _MIN_PYTHON:
+    warnings.warn(
+        f"Cartography is tested on Python {_MIN_PYTHON_STR}+ only. "
+        f"Backward compatibility with Python 3.10-3.12 is not guaranteed. "
+        f"Python 3.10 support will be removed in October 2026. "
+        "See: https://github.com/cartography-cncf/cartography/issues/2205",
+        FutureWarning,
```
kunaals: i like the intent here, but i think this should be `DeprecationWarning` rather than `FutureWarning`. python’s warning guidance treats `FutureWarning` as something meant to be shown to application end users, while `DeprecationWarning` is the library-facing category and is hidden by default for imported modules. that seems like a better fit for `import cartography`, especially since `FutureWarning` can turn into an exception under `-W error` / `PYTHONWARNINGS=error`.

refs:
https://docs.python.org/3/library/warnings.html
https://docs.python.org/3/whatsnew/3.7.html#pep-565-show-deprecationwarning-in-main

cartography/__init__.py
```diff
@@ -0,0 +1,15 @@
+import sys
+import warnings
+
+_MIN_PYTHON = (3, 13)
+_MIN_PYTHON_STR = ".".join(map(str, _MIN_PYTHON))
+
+if sys.version_info < _MIN_PYTHON:
+    warnings.warn(
```
kunaals, line 8: i think we should add a visible startup warning in the cli path as well. that keeps library imports low-noise but still tells cartography cli users that 3.13 is now the tested/default runtime

kunaals (approved):
looks good overall and i’m happy to approve this to unblock. one wording tweak for the pr description though: i don’t think we should say the azure/msal prerelease blocker is fully gone yet. `azure-cli-core` still depends on `msal==1.35.0b1` today, so the prerelease in `uv.lock` is still coming from upstream azure packaging rather than our resolver doing something odd
https://pypi.org/pypi/azure-cli-core/2.84.0/json
https://pypi.org/project/msal/

PR #2351, opened 2026-02-05
feat(gitlab): Adding Coverage for Gitlab Image Layers

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- This PR adds coverage for GitLab Image Layers extending the GitLab Container Registry coverage. 

### Related issues or links

- Fixes #

### How was this tested?

- Added to Integration tests
- Used Test GitLab Env to confirm functionality works 

#### General

#### Proof of functionality

image


kunaals (changes requested):
to preserve layer order, please align gitlab layer modeling with the ecr pattern. add ordered `layer_diff_ids` (or equivalent) onto the image node, and create `NEXT` relationships between layers plus `HEAD`/`TAIL` links back to the image. right now `HAS_LAYER` is orderless and `position` isn’t persisted, so ordering is lost. mirroring the ecr approach would make layer ordering queryable and consistent across providers.

cartography/intel/gitlab/container_images.py
```diff
@@ -16,6 +16,9 @@
 from cartography.intel.gitlab.util import fetch_registry_blob
 from cartography.intel.gitlab.util import fetch_registry_manifest
 from cartography.intel.gitlab.util import get_paginated
+from cartography.models.gitlab.container_image_layers import (
+    GitLabContainerImageLayerSchema,
+)
 from cartography.models.gitlab.container_images import GitLabContainerImageSchema
 from cartography.util import timeit
```
kunaals, line 23: layer_digests becomes [] for empty layers; consider normalizing to none to skip relationship matching, consistent with child_image_digests.

cartography/intel/gitlab/container_images.py
```diff
@@ -299,13 +310,56 @@ def transform_container_images(
...
                 "child_image_digests": child_image_digests,
+                "layer_digests": layer_digests,
             }
         )
 
     logger.info(f"Transformed {len(transformed)} container images")
     return transformed
 
 
+def transform_container_image_layers(
```
kunaals, line 328: image_digest is not used by the schema or rels; should be removed or wired into the model.

cartography/models/gitlab/container_images.py
```diff
@@ -84,6 +84,30 @@ class GitLabContainerImageContainsImageRel(CartographyRelSchema):
...
+class GitLabContainerImageToLayerRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class GitLabContainerImageToLayerRel(CartographyRelSchema):
+    """
+    Relationship from an image to its constituent layers.
+    Only applies to images with type="image" (not manifest lists).
+    The position property indicates layer order (0 = base layer).
```
kunaals: docstring mentions position on the rel, but the rel has no position property; update the doc or the model.

kunaals (changes requested):
few extra things

cartography/models/gitlab/container_images.py
kunaals: actually no need for this if we have the layer_diff_ids list on the image node and the next/head/tail pointers

cartography/models/gitlab/container_image_layers.py
```diff
@@ -0,0 +1,157 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class GitLabContainerImageLayerNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("digest")
```
kunaals: can we confirm the id choice here? this uses compressed layer `digest`, while ecr keys by uncompressed `diff_id`. if we want cross-provider layer dedupe by content, these should likely align.

cartography/models/gitlab/container_image_layers.py
```diff
@@ -0,0 +1,157 @@
...
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "TAIL"
+    properties: GitLabContainerImageLayerTailOfImageRelProperties = (
+        GitLabContainerImageLayerTailOfImageRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class GitLabContainerImageLayerSchema(CartographyNodeSchema):
```
kunaals, line 85: can we add the semantic `imagelayer` label here like ecr does? without it, cross-provider layer queries can miss gitlab layers.

cartography/intel/gitlab/container_images.py
```diff
@@ -299,13 +312,97 @@ def transform_container_images(
...
+        layers = manifest.get("layers", [])
+        config = manifest.get("_config", {})
+        diff_ids = config.get("rootfs", {}).get("diff_ids", [])
+
+        # Process each layer in the chain
+        for i, layer in enumerate(layers):
+            layer_digest = layer.get("digest")
+            if not layer_digest:
+                continue
+
```
kunaals, line 369: if `diff_ids` are missing we store `diff_id = none`. can we make the intended behavior explicit (skip, fallback, or keep) so this is a deliberate decision and not silent drift?

tests/integration/cartography/intel/gitlab/test_container_registry.py
```diff
@@ -381,3 +381,55 @@ def test_sync_container_registry(
...
+            "GitLabContainerImage",
+            "id",
+            "GitLabContainerImageLayer",
+            "digest",
+            "HAS_LAYER",
+        )
+        == expected_has_layer_rels
+    )
+
+    # Note: HEAD/TAIL/NEXT relationships verified in production sync
```
kunaals: right now this test never actually asserts `head`/`tail`/`next`. can we add one image with 2-3 layers and assert ordering links so regressions are caught?

docs/root/modules/gitlab/schema.md
```diff
@@ -469,6 +471,60 @@ Representation of a container image identified by its digest. Images are content
     (KubernetesContainer)-[HAS_IMAGE]->(GitLabContainerImage)
     ```
```
kunaals, line 478: small docs drift: we now write `layer_digests` on gitlabcontainerimage, but it is not listed in the image fields table above. can we document that field too?

PR #2370, opened 2026-02-09
feat(github): parse workflow YAML for actions, secrets, and permissions

jychp:
### Type of change

### Summary

Enhances the GitHub Actions workflow sync to parse workflow YAML content and extract security-relevant information:

- **Secrets usage**: Extracts all `${{ secrets.* }}` references from workflow files using regex on the raw YAML string, catching secrets in all locations (`env`, `with`, `run`, `if`, `outputs`, `strategy`, etc.). Supports both dot notation (`secrets.NAME`) and bracket notation (`secrets['NAME']`).
- **Third-party actions**: Creates `GitHubAction` nodes with metadata (owner, version, pinned status) and `USES_ACTION` relationships to workflows
- **Reusable workflow calls**: Detects `workflow_call` triggers and external workflow references
- **Permissions**: Parses all 10 permission scopes (contents, actions, packages, pull-requests, issues, deployments, statuses, checks, id-token, security-events)
- **Environment variables** and **trigger events**: Captured as workflow node properties

New files:
- `workflow_parser.py` — YAML parser with dataclasses (`ParsedWorkflow`, `ParsedAction`)
- `models/github/action.py` — `GitHubAction` node schema with org sub-resource and workflow relationship
- `tests/data/github/workflow_content.py` — 9 test workflow YAML fixtures
- `tests/unit/.../test_workflow_parser.py` — 29 unit tests for the parser

Modified files:
- `models/github/workflow.py` — Added parsed fields (trigger_events, permissions_*, env_vars, job_count, has_reusable_workflow_calls) and `REFERENCES_SECRET` relationship (one_to_many)
- `intel/github/actions.py` — Added content fetching, parsing, enrichment in sync loop; action node creation and loading
- `tests/integration/.../test_actions.py` — Added `test_sync_github_actions_workflow_parsing` integration test

### Breaking changes

None. All new fields are additive. Existing `GitHubWorkflow` nodes gain new properties but no existing properties or relationships are removed.

### How was this tested?

- **29 unit tests** covering action reference parsing (standard, pinned, local, docker, reusable), secret extraction (dot and bracket notation), permissions parsing, full workflow YAML parsing, and deduplication
- **9 integration tests** (including new `test_sync_github_actions_workflow_parsing`) verifying end-to-end sync with mocked API and workflow content, checking node properties, `USES_ACTION` relationships, and `REFERENCES_SECRET` relationships
- Tested in production against real organisation (see screenshot in comments)

#### General

#### Proof of functionality

### Notes for reviewers

- The workflow content is fetched via the GitHub Contents API (`GET /repos/{owner}/{repo}/contents/{path}`) for each workflow. Rate limiting is handled by the existing `call_github_rest_api` machinery.
- Secrets are matched via regex on the raw YAML string, supporting both `${{ secrets.NAME }}` and `${{ secrets['NAME'] }}` notations. This catches references in all locations regardless of YAML structure.
- YAML parse failures are logged without exception details to avoid leaking sensitive workflow content.
- Missing optional values (action `owner`, `version`) use `None` instead of empty strings.
- The `GitHubAction` node ID format is `{org}:{raw_uses}` (e.g., `simpsoncorp:actions/checkout@v4`), scoped to the organization.


jychp:
Screenshot 2026-02-09 at 5 50 56 PM

cartography/intel/github/actions.py
```diff
@@ -353,6 +358,140 @@ def transform_env_variables(
...
+    :param workflow_id: The workflow ID (for relationship)
+    :param organization: The organization name
+    :return: List of action data dicts
+    """
+    actions = deduplicate_actions(parsed.actions)
+    result = []
+
+    for action in actions:
+        # Create a unique ID for the action based on the raw uses string
+        # This ensures actions/checkout@v4 is different from actions/checkout@v3
```
kunaals: i think this id can collide for local actions across repos. for example, `./.github/actions/build` in two repos in the same org will map to the same `githubaction` node (`{org}:{raw_uses}`), which can create incorrect cross-repo `uses_action` edges. can we include repo context in the id for local actions?

cartography/intel/github/actions.py
```diff
@@ -353,6 +358,140 @@ def transform_env_variables(
...
+            f"https://github.com/{organization}/{repo_name}/actions/secrets/{secret_name}"
+        )
+        # Add org-level secret ID
+        secret_ids.append(
+            f"https://github.com/{organization}/actions/secrets/{secret_name}"
+        )
+
+    return {
+        **workflow,
+        "trigger_events": parsed.trigger_events if parsed.trigger_events else None,
```
kunaals, line 440: we parse global permissions (`permissions: read-all` / `write-all`) into `_global`, but this mapping only persists per-scope fields. so global-only workflows end up with all permission fields as null. should we store the global permission explicitly (or expand it) so we don’t lose that signal?

tests/unit/cartography/intel/github/test_workflow_parser.py
```diff
@@ -0,0 +1,305 @@
...
+        "pull-requests": "write",
+    }
+    result = parse_permissions(perms)
+    assert result["contents"] == "read"
+    assert result["pull_requests"] == "write"
+
+
+def test_parse_permissions_string():
+    result = parse_permissions("read-all")
+    assert result["_global"] == "read-all"
```
kunaals: can we add coverage for `permissions: read-all` / `write-all` and assert what lands on the workflow node after `enrich_workflow_with_parsed_content`? that would lock in expected behavior and prevent silent regressions here.

PR #2371, opened 2026-02-09
Fix syntax error in Cloudflare zone filtering

### LINKED ISSUE #2372: Cloudflare zone intel module does not filter by account
### Expected behavior

Expected: The Cloudflare intel module imports the Cloudflare Accounts, and then imports the zones for that account.

### Actual behavior

Actual: The module imports the Accounts, and then each account gets _every_ zone attached to it.

### To Reproduce

Required: A Cloudflare user with access to multiple Cloudflare Accounts, each account having their own zones.

Create a Cloudflare User API token (`https://dash.cloudflare.com/profile/api-tokens`), with Account Resources set to "Include all accounts".

Run the intel module, and notice that all zones are attached to all accounts.

>  - Python version: 3.10.19

>  - OS (feel free to omit this if you don't think it's relevant to your issue): OSX

Here is a pull request that fixes the issue:

https://github.com/cartography-cncf/cartography/pull/2371

kirkj-lightspeed:
### Type of change

### Summary

The Cloudflare client library is used incorrectly, which means that zones from all accounts are connected to every account. This happens if your API token is scoped across multiple Cloudflare accounts.

This fix changes the `client.zones.list` call, to use the correct filter syntax.

### Details

Cartography uses the cloudflare library, version 4.3.1: https://pypi.org/project/cloudflare/

The current syntax in zones.py passes the account_id as a simple parameter:

```
client.zones.list(account=account_id)]
```

However, the Cloudflare API for List Zones uses a different syntax, requiring a complex account object. This is different to the other API methods, such as List Members and List Roles:

https://developers.cloudflare.com/api/resources/zones/methods/list/

The correct syntax to pass this complex object to the client library is:

```
client.zones.list(account={"id": account_id})
```

### Related issues or links

- Fixes #2372

### How was this tested?

This was tested manually using a Personal API token attached to my user profile, which is granted access to multiple Cloudflare accounts.

I have not created a unit test for this fix, as I am not sure how to do that.

#### General

#### Proof of functionality

  - Example:
    ```
    ```

I haven't run the un-patched sync job to completion, as it runs for hours. However with this change the sync runs quickly:

```
INFO:cartography.graph.job:Finished job CloudflareDNSRecord
INFO:cartography.sync:Finishing sync stage 'cloudflare'
```


kunaals:
Thanks for this contribution @kirkj-lightspeed! Sorry for the delay, merging this in now

PR #2391, opened 2026-02-17
feat(AWS): Complete AWS Compute Internet Exposure Coverage

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- AWS already has coverage for 
     - (:AWSLoadBalancerV2)-[:EXPOSE]->(:EC2Instance)
     - EC2 Instance exposed_internet property (through analysis job)
     - ECS Container exposed_internet property (through analysis job)
     - (:AWSLoadBalancerV2)-[:MEMBER_OF_EC2_SECURITY_GROUP]->(:EC2SecurityGroup) 

- Expand coverage to have 
     - (:AWSLoadBalancerV2)-[:EXPOSES]->(:ECSContainer)
     - (:AWSLoadBalancerV2)-[:PROTECTS]->(:EC2NetworkAcl) 

### How was this tested?

image

#### General

### DIFF
--- a/cartography/data/jobs/scoped_analysis/aws_lb_container_exposure.json
+++ b/cartography/data/jobs/scoped_analysis/aws_lb_container_exposure.json
@@ -8,7 +8,7 @@
     },
     {
       "__comment": "Create: EXPOSE relationships for private containers behind internet-facing load balancers",
-      "query": "MATCH (aa:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(lb:AWSLoadBalancerV2 {scheme: 'internet-facing'})-[:EXPOSE]->(ip:EC2PrivateIp)<-[:PRIVATE_IP_ADDRESS]-(ni:NetworkInterface)<-[:NETWORK_INTERFACE]-(task:ECSTask)-[:HAS_CONTAINER]->(c:ECSContainer) WHERE ni.public_ip IS NULL MERGE (lb)-[r:EXPOSE]->(c) SET r.lastupdated = $UPDATE_TAG, r.exposure_type = 'via_lb_only'",
+      "query": "MATCH (aa:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(lb:AWSLoadBalancerV2 {scheme: 'internet-facing'})-[:EXPOSE]->(ip:EC2PrivateIp)<-[:PRIVATE_IP_ADDRESS]-(ni:NetworkInterface)<-[:NETWORK_INTERFACE]-(task:ECSTask)-[:HAS_CONTAINER]->(c:ECSContainer) WHERE ip.public_ip IS NULL MERGE (lb)-[r:EXPOSE]->(c) SET r.lastupdated = $UPDATE_TAG, r.exposure_type = 'via_lb_only'",
       "iterative": false
     }
   ]


cartography/data/jobs/analysis/aws_lb_nacl_direct.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "AWS LoadBalancer to NACL direct relationship",
+  "statements": [
+    {
+      "__comment": "Cleanup: Remove all previous PROTECTS relationships between NACLs and LBs",
+      "query": "MATCH (:EC2NetworkAcl)-[r:PROTECTS]->(:AWSLoadBalancerV2) DELETE r",
```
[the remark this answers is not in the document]
kunaals: this rel does not exist yet so no need

cartography/data/jobs/analysis/aws_lb_container_exposure.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "AWS LoadBalancer to ECS Container direct relationship",
+  "statements": [
+    {
+      "__comment": "Cleanup: Remove all previous EXPOSE relationships between LBs and containers",
+      "query": "MATCH (:AWSLoadBalancerV2)-[r:EXPOSE]->(:ECSContainer) DELETE r",
```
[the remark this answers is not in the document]
kunaals: this rel does not exist yet so no need

kunaals: im not an expert on analysis jobs but do both queries in this file need to be scoped to a tenant? Like should there be some resource type rels in here?

shyammukund: So I believe that global analysis jobs like these ones arne't scoped to the tenant like how its done in aws_lambda_ecr.json and aws_ecs_asset_exposure.json —   
  they just match on node labels with no {id: $AWS_ID} filter

kunaals: hm but this will delete all such rels across the whole graph, this isn't scoped to particular assets

cartography/intel/aws/__init__.py
```diff
@@ -141,6 +141,18 @@ def _sync_one_account(
         common_job_parameters,
     )
 
+    run_scoped_analysis_job(
```
kunaals: this runs unconditionally per account. can we gate this on requested syncs/deps so partial syncs don't create/clean lb rels from stale graph state?

cartography/intel/aws/__init__.py
```diff
@@ -141,6 +141,18 @@ def _sync_one_account(
...
+        "aws_lb_container_exposure.json",
+        neo4j_session,
+        common_job_parameters,
+    )
+
+    run_scoped_analysis_job(
```
kunaals: same concern here: should this be conditional on the relevant ec2/lb/nacl syncs being requested?

cartography/data/jobs/scoped_analysis/aws_lb_nacl_direct.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "AWS LoadBalancer to NACL direct relationship",
+  "statements": [
+    {
+      "__comment": "Cleanup: Remove stale PROTECTS relationships between NACLs and LBs for this account",
+      "query": "MATCH (aa:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(lb:AWSLoadBalancerV2)<-[r:PROTECTS]-(:EC2NetworkAcl) WHERE r.lastupdated <> $UPDATE_TAG DELETE r",
+      "iterative": true
+    },
+    {
+      "__comment": "Create: PROTECTS relationships via subnet traversal",
+      "query": "MATCH (aa:AWSAccount{id: $AWS_ID})-[:RESOURCE]->(lb:AWSLoadBalancerV2)-[:SUBNET]->(subnet:EC2Subnet)<-[:PART_OF_SUBNET]-(nacl:EC2NetworkAcl) MERGE (nacl)-[r:PROTECTS]->(lb) SET r.lastupdated = $UPDATE_TAG, r.via_subnet = subnet.subnetid",
```
kunaals: `via_subnet` can be lossy here. if a nacl-lb pair matches via multiple subnets, this single property gets overwritten nondeterministically. should this be aggregated to a list?

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -326,7 +326,7 @@ def test_sync_one_account_all_sync_functions(
     # Check that the boilerplate functions get called as expected. Brittle, but a good sanity check.
     assert mock_autodiscover.call_count == 0
     assert mock_cleanup.call_count == 0
-    assert mock_analysis.call_count == 1
+    assert mock_analysis.call_count == 3
```
kunaals, line 329: this only updates call count. can we add coverage that validates actual graph output for the new rels (and partial-sync behavior), not just invocation count?

cartography/data/jobs/scoped_analysis/aws_lb_nacl_direct.json
shyammukund: Leaning towards dropping this property. I think the PROTECTS rel already tells you that the NACL protects the Load Balancer. If you need to know which subnets you can traverse (nacl)-[:PART_OF_SUBNET]->(subnet)<-[:SUBNET]-(lb).

shyammukund: Also the analysis jobs don't depend on it so i think it makes sense to avoid the complexity of aggregating to a list.

tests/integration/cartography/intel/aws/test_init.py
shyammukund: noted: added test coverage to the LB integration test that mocks the traversal chain and then validates the analysis job creates the intended relationship

PR #2393, opened 2026-02-18
feat(eks,kubernetes): add cert parsing diagnostics and kube TLS posture metadata

kunaals:
### Type of change

- [x] Documentation update

### Summary

This change improves Cartography’s ability to answer a common operational question:

**“Is this cluster certificate setup healthy, and if not, why?”**

Before this, cert-related issues were hard to reason about from graph data alone. Missing or malformed certificate material could look similar to “no data,” and kubeconfig TLS posture was not visible on `KubernetesCluster` nodes.

With this PR:
- EKS clusters now carry explicit certificate parsing outcomes so operators can distinguish healthy, missing, and invalid cert states.
- Kubernetes clusters now carry kubeconfig TLS posture signals so reviewers can quickly identify risky setups (for example, skipped TLS verification or missing CA material).
- The graph now supports faster debugging, clearer evidence for reviews, and safer comparisons across environments without requiring deep log digging.

### Breaking changes

None.

### How was this tested?

Screenshot 2026-03-01 at 4 40 47 PM

#### General

#### Proof of functionality

### Notes for reviewers

Practical mental model:
- EKSCluster cert metadata = server/control-plane certificate diagnostics (AWS-side).
- KubernetesCluster kubeconfig TLS posture = client verification configuration diagnostics (client-side).

These can disagree, and that disagreement is useful for troubleshooting.

Recommended review lens:
- Does the resulting graph make cert/TLS state easy to interpret?
- Are failure modes explicit enough for operators to triage quickly?
- Do the docs and queries provide enough context for practical troubleshooting?

### DIFF
--- a/.github/workflows/publish-to-ghcr-and-pypi.yml
+++ b/.github/workflows/publish-to-ghcr-and-pypi.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Build a binary wheel and a source tarball
         run: uv build
       - name: Publish distribution 📦 to PyPI
-        if: startsWith(github.ref, 'refs/tags')
+        if: github.event_name == 'release'
         # --check-url checks if files already exist on PyPI before uploading.
         # If the version already exists, it skips without failing, allowing
         # downstream jobs (Docker build) to proceed on re-runs.
@@ -111,7 +111,7 @@ jobs:
           MAX_ATTEMPTS=30
           ATTEMPT=0
           while [ $ATTEMPT -lt $MAX_ATTEMPTS ]; do
-            if pip index versions cartography 2>&1 | grep -q "$VERSION"; then
+            if curl --fail --silent --show-error "https://pypi.org/pypi/cartography/$VERSION/json" > /dev/null; then
               echo "✅ cartography==$VERSION is now available on PyPI"
               exit 0
             fi
--- a/cartography/data/jobs/analysis/gcp_role_resource_edge_migration.json
+++ b/cartography/data/jobs/analysis/gcp_role_resource_edge_migration.json
@@ -1,0 +1,10 @@
+{
+  "name": "GCP role resource edge migration",
+  "statements": [
+    {
+      "__comment__": "DEPRECATED: temporary backward-compatibility cleanup; remove in v1.0.0 once legacy project->role edges are no longer expected.",
+      "query": "MATCH (:GCPProject)-[r:RESOURCE]->(role:GCPRole) WHERE role.scope IN ['GLOBAL', 'ORGANIZATION'] OR (role.name IS NOT NULL AND (role.name STARTS WITH 'roles/' OR role.name STARTS WITH 'organizations/')) DELETE r",
+      "iterative": false
+    }
+  ]
+}
--- a/cartography/intel/gcp/__init__.py
+++ b/cartography/intel/gcp/__init__.py
@@ -15,6 +15,10 @@ from googleapiclient.discovery import Resource
 from cartography.config import Config
 from cartography.graph.job import GraphJob
 from cartography.intel.gcp import artifact_registry
+from cartography.intel.gcp import bigquery_connection
+from cartography.intel.gcp import bigquery_dataset
+from cartography.intel.gcp import bigquery_routine
+from cartography.intel.gcp import bigquery_table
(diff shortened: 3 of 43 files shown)


cartography/intel/aws/eks.py
```diff
@@ -67,6 +75,130 @@ def _process_logging(cluster: Dict) -> bool:
...
+        else cert.not_valid_before
+    )
+    not_after_utc = (
+        cert.not_valid_after_utc
+        if hasattr(cert, "not_valid_after_utc")
+        else cert.not_valid_after
+    )
+    not_before_utc = _ensure_utc(not_before_utc)
+    not_after_utc = _ensure_utc(not_after_utc)
+    cert_metadata["certificate_authority_not_before"] = _format_datetime_iso(
```
jychp: Why are you using str instead of storing the datetime object ?

cartography/intel/aws/eks.py
```diff
@@ -67,6 +75,130 @@ def _process_logging(cluster: Dict) -> bool:
...
+        not_before_utc,
+    )
+    cert_metadata["certificate_authority_not_after"] = _format_datetime_iso(
```
jychp: Why are you using str instead of storing the datetime object ?

cartography/intel/aws/eks.py
```diff
@@ -1,10 +1,18 @@
...
+from datetime import timezone
 from typing import Any
 from typing import Dict
 from typing import List
 
 import boto3
 import neo4j
+from cryptography import x509
+from cryptography.hazmat.primitives import hashes
+from cryptography.x509.extensions import ExtensionNotFound
```
jychp, line 14: cryptography is imported directly here (and in the unit tests) but is not declared as an explicit dependency in pyproject.toml. It's only pulled in transitively via PyJWT[crypto]. Please add cryptography to avoid breakage if the transitive path changes.

cartography/intel/aws/eks.py
```diff
@@ -67,6 +75,130 @@ def _process_logging(cluster: Dict) -> bool:
...
+        "certificate_authority_sha256_fingerprint": None,
+        "certificate_authority_subject": None,
+        "certificate_authority_issuer": None,
+        "certificate_authority_not_before": None,
+        "certificate_authority_not_after": None,
+        "certificate_authority_subject_key_identifier": None,
+        "certificate_authority_authority_key_identifier": None,
+        "certificate_authority_expired": None,
+        "certificate_authority_days_until_expiry": None,
+    }
```
jychp: `certificate_authority_expired` and `certificate_authority_days_until_expiry` are computed at sync time and go stale between syncs.

Since `certificate_authority_not_after` is already persisted, consumers can compute freshness at query time. Consider either removing these derived fields in favor of a documented Cypher expression, or adding a note in the schema docs that they reflect sync-time state only.

cartography/intel/aws/eks.py
kunaals: good call — i switched these to store timezone-aware datetime values directly (utc) instead of strings, and updated tests accordingly in c4a3a00ca.

cartography/intel/aws/eks.py
kunaals: yep, agreed. this now stores native datetime values (utc) for not_after as well, rather than serializing to str.

cartography/intel/aws/eks.py
kunaals: great catch. i added `cryptography` as an explicit dependency in pyproject.toml and refreshed uv.lock so we are not relying on transitive install behavior.

cartography/intel/aws/eks.py
kunaals: agreed on staleness risk. i kept these as convenience snapshot fields but updated schema docs to explicitly mark them as sync-time values and added a query-time cypher example using certificate_authority_not_after + datetime().

kunaals: followed up on this and removed both derived fields entirely (`certificate_authority_expired` and `certificate_authority_days_until_expiry`) so we only persist source-of-truth cert data. this avoids stale sync-time state and keeps the model cleaner. pushed in fe10f228a.

jvstokes:
This is one hell of a good PR

PR #2414, opened 2026-02-23
feat(kubernetes): add gateway-api ingestion for Gateway/HTTPRoute paths

### LINKED ISSUE #2301: feat(kube): Implement gateway-api relationships
## Description
According to the latest news about [NGINX Ingress retirement](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/), it is scheduled for March 2026.
Add a relationship between `gateway` pods and `service` through `HTTPRoute`.

Permit users to see how the new Kubernetes feature `gateway-api` is implemented on their clusters.
This relationship would complete the path mapping from the gateway to the pods:

Gateway to HTTPRoute:
`MATCH path = (gateway:gateway.networking.k8s.io/v1)-[*]->(HTTPRoute:gateway.networking.k8s.io/v1)
RETURN path`

HTTPRoute to Service:
`MATCH path = (HTTPRoute:gateway.networking.k8s.io/v1)-[*]->(svc:KubernetesSVC)
RETURN path`

At this point, the Services are already connected to Pods (existing feat), and then show the last path to permit full visibility of `gateway-api` implementation.

Discuss if `GatewayClass` implementation is required to perform the feat.

## Motivation
This relationship is needed to:

- Implement the new Kubernetes feature `gateway-api` on Cartography
- Enable security analysis of gateway traffic paths
- Support queries that trace L7 traffic flow into clusters
- Provide visibility into which Gateway resources are associated with which HTTPRoute, Services and Pods 

## Alternatives Considered
No other alternatives, the existing features do not include `gateway-api` support.

## Relevant Links

- [Gateway-API](https://gateway-api.sigs.k8s.io/)
- [HTTPRoute](https://gateway-api.sigs.k8s.io/guides/http-routing/)

kunaals:
### Type of change

### Summary
- Added new Kubernetes gateway-api ingestion support for `Gateway` and `HTTPRoute` resources from `gateway.networking.k8s.io/v1`.
- Introduced new node schemas:
  - `KubernetesGateway`
  - `KubernetesHTTPRoute`
- Added relationships needed for traffic-path visibility:
  - `(:KubernetesGateway)-[:ROUTES]->(:KubernetesHTTPRoute)` via `spec.parentRefs`
  - `(:KubernetesHTTPRoute)-[:TARGETS]->(:KubernetesService)` via `spec.rules[].backendRefs`
- Wired the new sync (`sync_gateway_api`) into the Kubernetes ingestion pipeline.
- Extended `K8sClient` with `CustomObjectsApi` support and updated epoch parsing to handle RFC3339 timestamp strings returned by custom resources.
- Added integration test coverage and test data for load and sync flows.

### Related issues or links
- Fixes #2301

### Breaking changes
None.

### How was this tested?
- `ruff check cartography/intel/kubernetes/gateway_api.py cartography/intel/kubernetes/__init__.py cartography/intel/kubernetes/util.py cartography/models/kubernetes/gateway_api.py tests/data/kubernetes/gateway_api.py tests/integration/cartography/intel/kubernetes/test_gateway_api.py`
- `pytest -q tests/integration/cartography/intel/kubernetes/test_gateway_api.py` *(fails in this environment because Docker daemon is unavailable for testcontainers/Neo4j)*

#### General

#### Proof of functionality

### Notes for reviewers
- This change intentionally focuses on path mapping needed for gateway-api visibility (`Gateway -> HTTPRoute -> Service`), which then joins existing `Service -> Pod` relationships.
- `GatewayClass` entity ingestion was not added here because it is not required to build the requested traffic path relationships. It can be introduced later if maintainers want richer control-plane modeling.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_699cb0f46e008323b0eaefb0786db8fd)


cartography/models/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,191 @@
...
+
+@dataclass(frozen=True)
+class KubernetesHTTPRouteToKubernetesServiceRel(CartographyRelSchema):
+    target_node_label: str = "KubernetesService"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "cluster_name": PropertyRef("CLUSTER_NAME", set_in_kwargs=True),
+            "namespace": PropertyRef("backend_service_namespaces", one_to_many=True),
+            "name": PropertyRef("backend_service_names", one_to_many=True),
+        }
```
jychp, line 174: blocking: Both `KubernetesGatewayToHTTPRouteRel` and `KubernetesHTTPRouteToKubernetesServiceRel` use two `one_to_many=True` fields (namespace + name) in the same `TargetNodeMatcher`. The query builder expands these as independent `IN` checks (`WHERE t.namespace IN [...] AND t.name IN [...]`), creating a cartesian product instead of zipping paired values.

If an HTTPRoute targets `(ns-a, svc-1)` and `(ns-b, svc-2)`, this will also match `(ns-a, svc-2)` and `(ns-b, svc-1)` whenever those nodes exist — creating spurious cross-namespace relationships. The tests pass only because all test data uses the same namespace.

This is the first use of multiple `one_to_many` fields in a single matcher across the entire codebase. Consider using a composite key (e.g. `namespace/name`) as a single `one_to_many` field, and matching against a corresponding composite property on the target nodes.

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+        if continue_token:
+            kwargs["_continue"] = continue_token
+
+        response = client.custom.list_cluster_custom_object(
+            group=group,
+            version=version,
+            plural=plural,
+            limit=100,
+            **kwargs,
+        )
```
jychp: blocking: `_list_cluster_custom_objects()` does not catch `ApiException`. If the Gateway API CRDs are not installed on a cluster (common today), this returns a 404 that propagates through `sync_gateway_api()` and hits the `except Exception` in `start_k8s_ingestion()`, aborting the entire cluster sync — preventing `sync_ingress` and everything after it from running.

The existing `k8s_paginate()` in `util.py` handles this by catching `ApiException` and logging. Apply similar error handling here, or return an empty list on 404.

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+    routes: list[dict[str, Any]],
+) -> None:
+    route_parents: dict[tuple[str, str], set[tuple[str, str]]] = {}
+    for route in routes:
+        route_namespace = route["namespace"]
+        route_name = route["name"]
+        for gateway_namespace, gateway_name in zip(
+            route.get("parent_gateway_namespaces", []),
+            route.get("parent_gateway_names", []),
+            strict=False,
```
jychp: [suggestion] `strict=False` here silently ignores mismatched array lengths. Since `parent_gateway_namespaces` and `parent_gateway_names` are always constructed in parallel, consider `strict=True` to catch bugs early.

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+
+        continue_token = response.get("metadata", {}).get("continue")
+        if not continue_token:
+            break
+
+    logger.debug("Fetched %d %s resources", len(all_resources), resource_name)
+    return all_resources
+
+
+def get_gateways(client: K8sClient) -> list[dict[str, Any]]:
```
jychp, line 88: add `@timeit`

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+    return _list_cluster_custom_objects(
+        client,
+        group="gateway.networking.k8s.io",
+        version="v1",
+        plural="gateways",
+    )
+
+
+def get_http_routes(client: K8sClient) -> list[dict[str, Any]]:
```
jychp, line 98: add `@timeit`

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+
+    for gateway in gateways:
+        attached = sorted(
+            route_parents.get((gateway["namespace"], gateway["name"]), set())
+        )
+        gateway["attached_route_namespaces"] = [namespace for namespace, _ in attached]
+        gateway["attached_route_names"] = [name for _, name in attached]
+
+
+def load_gateways(
```
jychp, line 221: add `@timeit`

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+        neo4j_session,
+        KubernetesGatewaySchema(),
+        gateways,
+        lastupdated=update_tag,
+        CLUSTER_ID=cluster_id,
+        CLUSTER_NAME=cluster_name,
+    )
+
+
+def load_http_routes(
```
jychp, line 239: add `@timeit`

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+        neo4j_session,
+        KubernetesHTTPRouteSchema(),
+        routes,
+        lastupdated=update_tag,
+        CLUSTER_ID=cluster_id,
+        CLUSTER_NAME=cluster_name,
+    )
+
+
+def cleanup(session: neo4j.Session, common_job_parameters: dict[str, Any]) -> None:
```
jychp, line 257: add `@timeit`

jychp (changes requested):

cartography/intel/kubernetes/util.py
```diff
@@ -112,10 +129,14 @@ def get_k8s_clients(kubeconfig: str) -> list[K8sClient]:
     return clients
 
 
-def get_epoch(date: datetime | None) -> int | None:
-    if date:
-        return int(date.timestamp())
-    return None
+def get_epoch(date: datetime | str | None) -> int | None:
```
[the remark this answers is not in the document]
kunaals: good catch. reverted the util change so gateway-api keeps the raw timestamps instead of teaching get_epoch to parse strings.

tests/integration/cartography/intel/kubernetes/test_gateway_api.py
```diff
@@ -0,0 +1,160 @@
...
+        "KubernetesCluster",
+    ]:
+        neo4j_session.run(f"MATCH (n:{label}) DETACH DELETE n")
+
+
+def test_gateway_api_relationships(neo4j_session):
+    _create_test_cluster(neo4j_session)
+
+    try:
+        load_http_routes(
```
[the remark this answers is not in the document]
kunaals: keeping this split. the kubernetes integration suite already uses loader-level tests alongside a sync end-to-end test here, and i added a regression case for the namespace/name pairing bug.

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+def transform_gateways(gateways: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    transformed: list[dict[str, Any]] = []
+
+    for gateway in gateways:
+        metadata = gateway.get("metadata", {})
+        spec = gateway.get("spec", {})
+
+        transformed.append(
+            {
+                "uid": metadata.get("uid"),
```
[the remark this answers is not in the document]
kunaals: good catch. switched the required metadata fields over to bracket access so missing uid, name, or namespace fails loudly.

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+    resource_name = f"{group}/{version}/{plural}"
+    all_resources: list[dict[str, Any]] = []
+    continue_token: str | None = None
+
+    while True:
+        kwargs: dict[str, Any] = {}
+        if continue_token:
+            kwargs["_continue"] = continue_token
+
+        response = client.custom.list_cluster_custom_object(
```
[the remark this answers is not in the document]
kunaals: good catch. _list_cluster_custom_objects() now catches apiexception, logs, and returns cleanly when the crd is missing.

cartography/intel/kubernetes/gateway_api.py
```diff
@@ -0,0 +1,236 @@
...
+        routes,
+        lastupdated=update_tag,
+        CLUSTER_ID=cluster_id,
+        CLUSTER_NAME=cluster_name,
+    )
+
+
+def cleanup(session: neo4j.Session, common_job_parameters: dict[str, Any]) -> None:
+    logger.debug("Running cleanup job for Kubernetes gateway-api resources")
+    GraphJob.from_node_schema(KubernetesGatewaySchema(), common_job_parameters).run(
```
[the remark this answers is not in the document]
kunaals: good catch. cleanup now runs httproute before gateway.

cartography/models/kubernetes/gateway_api.py
kunaals: good catch. i switched both joins to a single qualified-name key and added a cross-namespace regression test so we do not create cartesian matches.

cartography/intel/kubernetes/gateway_api.py
kunaals: good catch. _list_cluster_custom_objects() now handles apiexception and returns cleanly when the gateway api crds are missing.

cartography/intel/kubernetes/gateway_api.py
kunaals: i removed the parallel namespace/name arrays entirely here. this now uses qualified-name lists instead, so there is no zip to get out of sync.

cartography/intel/kubernetes/gateway_api.py
kunaals: added @timeit.

cartography/intel/kubernetes/gateway_api.py
kunaals: added @timeit.

cartography/intel/kubernetes/gateway_api.py
kunaals: added @timeit.

cartography/intel/kubernetes/gateway_api.py
kunaals: added @timeit.

cartography/intel/kubernetes/gateway_api.py
kunaals: added @timeit.

jychp (approved):
LGTM

PR #2420, opened 2026-02-24
feat(ontology): add UserGroup semantic label for cross-platform group queries

jychp:
### Type of change

### Summary

Add the `UserGroup` semantic label to the ontology system, enabling unified querying of user groups across 15 different cloud providers and SaaS platforms.

**Ontology fields**: `name` (required), `description`, `email`

**Modules covered** (14 modules, 15 node types):
- **AWS**: `AWSGroup`, `AWSSSOGroup`
- **Identity Providers**: `EntraGroup`, `KeycloakGroup`, `OCIGroup`, `OktaGroup`, `ScalewayGroup`
- **Google**: `GoogleWorkspaceGroup`, `GSuiteGroup`
- **Collaboration**: `GitHubTeam`, `GitLabGroup`, `SlackGroup`, `PagerDutyTeam`
- **Other**: `DuoGroup`, `TailscaleGroup`

This allows cross-platform queries like:
```cypher
MATCH (g:UserGroup) RETURN g._ont_name, g._ont_source
```

### How was this tested?
- `make test_lint` passes (all pre-commit hooks green)
- `make test_unit` passes (612 tests, 0 failures)
- Ontology mapping import verified programmatically

#### General

#### Proof of functionality

### Notes for reviewers

- OktaGroup is a legacy module using raw Cypher. The `UserGroup` label and `_ont_*` properties are set directly in the Cypher query, following the same pattern as `OktaUser` with `UserAccount`.
- Nodes that already had `ExtraNodeLabels` (AWSGroup→AWSPrincipal, EntraGroup→EntraIdentity, GoogleWorkspaceGroup/GSuiteGroup→GCPPrincipal) now have `UserGroup` appended to their existing labels.


kunaals:
something a bit more meta: is the UserGroup label too broad for all the different types of groups we want to bring into ontology? especially if we want it to be the IAM posture “backbone” as specified?

the PR maps true IAM groups and also collaboration/namespace constructs into one label. i fear this may produce noisy results for security queries (for example: GitLab namespace groups, Slack usergroups, and PagerDuty teams might not be seen as equivalent IAM primitives in the eyes of some folks).

would it be better have like a `_ont_group_type` (ex: iam_access, directory, collaboration, namespace)? or would it make sense to split into multiple labels (AccessGroup, CollaborationGroup)?

i dont have a strong opinion here, just wanted to bring it up and discuss before we lay the concrete here

PR #2423, opened 2026-02-24
fix(aws) Add batching to avoid threads hanging on s3 and ecr GetDetails

heryxpc:
### Type of change

- [ ] Documentation update

### Summary
fix(aws) Add batching to avoid threads hanging when getting s3 and ecr details

### Related issues or links
With the increase in the scale of our infrastructure, we are seeing threads hanging in `s3` and `ecr` syncs.
The root cause for s3 is that there is no max limit to the connection pool to AWS API at https://github.com/cartography-cncf/cartography/blob/80a35b27e41d57c76d521d2bfda10d6e3d7edc35/cartography/intel/aws/s3.py#L1252 and all bucket details are being requested at once via https://github.com/cartography-cncf/cartography/blob/80a35b27e41d57c76d521d2bfda10d6e3d7edc35/cartography/intel/aws/s3.py#L169.
For ecr if a repository has a very large volume of images (think > 500k) querying all details could also deplete and hangout the running threads: https://github.com/cartography-cncf/cartography/blob/80a35b27e41d57c76d521d2bfda10d6e3d7edc35/cartography/intel/aws/ecr.py#L381

- Fixes #

### Breaking changes
No breaking change, but might slowdown aws syncs.

### How was this tested?
- All unit and integration tests pass.
- Running a local sync for each module.

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
Set the batch size based on the connection pools to AWS api.


cartography/intel/aws/ecr.py
```diff
@@ -378,7 +381,9 @@ async def async_get_images(repo: Dict[str, Any]) -> None:
 
     # Sort repositories by name to ensure consistent processing order
     sorted_repos = sorted(repositories, key=lambda x: x["repositoryName"])
-    to_synchronous(*[async_get_images(repo) for repo in sorted_repos])
+    for i in range(0, len(sorted_repos), REPO_BATCH_SIZE):
+        batch = sorted_repos[i : i + REPO_BATCH_SIZE]
+        to_synchronous(*[async_get_images(repo) for repo in batch])
```
kunaals, line 386: hm this code scares me a bit, maybe a larger refactor is required here? perhaps we can use aioboto3?

kunaals:
also take a look at this related PR that cleans up both ecr and s3! https://github.com/cartography-cncf/cartography/pull/2427#issuecomment-3955183807

heryxpc:
hey @kunaals https://github.com/cartography-cncf/cartography/pull/2427#issuecomment-3955183807 looks great, but I'm worried that it might bring synchronization issues since we are doing some monkey patching for asyncio and threads in general.
I will run first my version and then test yours when it's merged as well.

PR #2428, opened 2026-02-25
Adding ubuntu intel module and tests

AdiLyft:
### Type of change

- [ ] Documentation update

### Summary

Add a new Ubuntu CVE intel module that syncs vulnerability and security notice data from the Ubuntu Security API (`https://ubuntu.com/security/cves.json` and `/security/notices.json`).

The module implements a **two-flow sync mechanism** for both CVEs and Notices:
- **Full sync with offset-based crash resume** — pages through the entire API (20 items/page). After each page, the current offset is persisted to a `UbuntuSyncMetadata` node in Neo4j, so if the process crashes it resumes from the last committed offset rather than starting over.
- **Incremental sync with watermark** — once a full sync completes, subsequent runs only fetch items updated (CVEs) or published (Notices) after the stored watermark, sorted in reverse order and stopping as soon as already-seen data is reached.

#### Node types created:
- **UbuntuCVE** (also labeled **CVE**): vulnerability records with Ubuntu-specific priority, CVSS v3 scores, and mitigation info. Node `id` is prefixed with `USV|` (e.g., `USV|CVE-2024-1234`) to avoid collisions with NVD `CVE` nodes; the original CVE identifier is stored in `cve_id`.
- **UbuntuCVEFeed**: parent node representing the data source, used as the sub-resource for cleanup scoping
- **UbuntuSecurityNotice**: Ubuntu security advisories (USNs) that announce patches for CVEs
- **UbuntuSyncMetadata**: tracks sync state (`full_sync_complete`, `full_sync_offset`, watermark) per module

#### Relationships:
- `(UbuntuCVEFeed)-[:RESOURCE]->(UbuntuCVE)`
- `(UbuntuCVEFeed)-[:RESOURCE]->(UbuntuSecurityNotice)`
- `(UbuntuSecurityNotice)-[:ADDRESSES]->(UbuntuCVE)` (one-to-many via `cves_ids`)

#### Ingestion order:
Feed → CVEs → Notices (feed node is created first so that sub-resource relationships can be established)

#### CLI flags added:
- `--ubuntu-security-enabled` — boolean to enable the module
- `--ubuntu-security-api-url` — optional, defaults to `https://ubuntu.com`

### Notes for reviewers

- The Ubuntu Security API is public and requires no authentication, so no API keys or secrets are needed.
- The API enforces a max page size of 20 (`limit` must be between 1 and 20; requests above 20 return a 422). The module pages through all results at 20 per request.
- As of today, the API reports ~64,000 CVEs and ~10,300 notices, which means a full sync requires ~3,700 API calls. This is comparable to the existing NVD CVE module's behavior.
- The `UbuntuCVE` node gets an extra label `CVE` via `ExtraNodeLabels` so it can be queried alongside CVEs from other sources (e.g., NVD). To prevent ID collisions with NVD's `CVE` nodes, UbuntuCVE IDs are prefixed with `USV|` (e.g., `USV|CVE-2024-1234`). The original CVE identifier is preserved in the `cve_id` property (indexed) for easy lookup.
- Each module (CVEs, Notices) maintains its own `UbuntuSyncMetadata` node (`UbuntuCVE_sync_metadata` and `UbuntuNotice_sync_metadata`) to independently track sync progress.
- The feed node is synced via a dedicated `feed.py` module before CVEs and Notices, following the pattern: `feed → cves → notices`.

### Related issues or links

None

### Breaking changes

None

### How was this tested?

**Unit tests (6 tests, mocked API + real Neo4j):**

*CVE tests (3):*
- `test_sync_ubuntu_cves` — verifies CVE nodes are created with correct properties, `USV|` prefixed IDs, `cve_id` field, extra `CVE` label, and CVSS v3 fields from nested impact data
- `test_sync_metadata_full_sync` — verifies that after a full sync, metadata has `full_sync_complete=True`, offset reset to 0, and watermark set to the max `updated_at` across loaded CVEs
- `test_sync_incremental` — seeds metadata with `full_sync_complete=True` and an old watermark, verifies sync takes the incremental path (calls `_fetch_cves` with `since=` instead of `start_offset=`) and updates the watermark after all pages

*Notice tests (3):*
- `test_sync_ubuntu_notices` — verifies notice nodes are created with correct properties
- `test_sync_ubuntu_notices_cve_relationships` — verifies one-to-many `ADDRESSES` relationships between notices and CVEs (using `USV|` prefixed CVE IDs)
- `test_sync_ubuntu_notices_feed_relationship` — verifies `RESOURCE` relationships between the feed node and notices

**Live crash-resume test (local script with real Ubuntu API + real Neo4j):**

Tests both CVEs and Notices with `page_size=5` and simulated crashes:
- Run 1: Full sync fetches 15 items (3 pages), then simulated crash — offset saved at 15
- Run 2: Resume from offset 15, fetches 5 more items (1 page), then simulated crash — offset saved at 20
- Run 3: Resume from offset 20, fetches 5 more items, full sync completes — offset resets to 0, `full_sync_complete=True`, watermark set
- Run 4: Incremental sync — fetches only recently updated/published items using the watermark

Both modules passed all 4 runs with correct state transitions, no duplicate ingestion, and all RESOURCE/ADDRESSES relationships intact.
Screenshot 2026-03-04 at 4 17 23 PM

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### DIFF
--- a/cartography/intel/gcp/artifact_registry/util.py
+++ b/cartography/intel/gcp/artifact_registry/util.py
@@ -3,6 +3,10 @@ import logging
 from googleapiclient.discovery import Resource
 from googleapiclient.errors import HttpError
 
+from cartography.intel.gcp.util import gcp_api_execute_with_retry
+from cartography.intel.gcp.util import get_error_reason
+from cartography.intel.gcp.util import is_api_disabled_error
+from cartography.intel.gcp.util import is_billing_disabled_error
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
@@ -15,7 +19,7 @@ def get_artifact_registry_locations(client: Resource, project_id: str) -> list[s
     """
     try:
         req = client.projects().locations().list(name=f"projects/{project_id}")
-        res = req.execute()
+        res = gcp_api_execute_with_retry(req)
 
         locations = [
             location.get("locationId")
@@ -29,6 +33,28 @@ def get_artifact_registry_locations(client: Resource, project_id: str) -> list[s
         return locations
 
     except HttpError as e:
+        if is_billing_disabled_error(e):
+            logger.warning(
+                "Artifact Registry billing is disabled for project %s. Skipping Artifact Registry sync for this project. Full details: %s",
+                project_id,
+                e,
+            )
+            return []
+        if is_api_disabled_error(e):
+            logger.info(
+                "Artifact Registry API appears disabled for project %s. Skipping Artifact Registry sync for this project. Full details: %s",
+                project_id,
+                e,
+            )
+            return []
+        reason = get_error_reason(e)
+        if reason in {"forbidden", "insufficientPermissions", "IAM_PERMISSION_DENIED"}:
+            logger.warning(
+                "Missing permissions for Artifact Registry in project %s. Skipping Artifact Registry sync for this project. Full details: %s",
+                project_id,
+                e,
+            )
+            return []
         logger.error(
             f"Failed to get Artifact Registry locations for project {project_id}: {e}",
         )
--- a/cartography/intel/gcp/compute.py
+++ b/cartography/intel/gcp/compute.py
@@ -17,6 +17,7 @@ from cartography.intel.gcp.cloud_armor import sync_gcp_cloud_armor
 from cartography.intel.gcp.instancegroup import sync_gcp_instance_groups
 from cartography.intel.gcp.util import gcp_api_execute_with_retry
 from cartography.intel.gcp.util import get_error_reason
+from cartography.intel.gcp.util import is_permission_denied_error
 from cartography.intel.gcp.util import parse_compute_full_uri_to_partial_uri
 from cartography.models.gcp.compute.firewall import GCPFirewallSchema
 from cartography.models.gcp.compute.firewall_target_tag import (
@@ -85,7 +86,7 @@ def get_zones_in_project(
                 e,
             )
             return None
-        elif reason == "forbidden":
+        elif is_permission_denied_error(e):
             logger.info(
                 (
                     "Your GCP identity does not have the compute.zones.list permission for project %s; skipping "
--- a/cartography/intel/gcp/instancegroup.py
+++ b/cartography/intel/gcp/instancegroup.py
@@ -14,6 +14,8 @@ from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.intel.gcp.util import gcp_api_execute_with_retry
 from cartography.intel.gcp.util import get_error_reason
+from cartography.intel.gcp.util import is_api_disabled_error
+from cartography.intel.gcp.util import is_billing_disabled_error
 from cartography.intel.gcp.util import parse_compute_full_uri_to_partial_uri
 from cartography.models.gcp.compute.instance_group import GCPInstanceGroupSchema
 from cartography.util import timeit
@@ -65,6 +67,24 @@ def _get_instance_group_members(
             if reason in {"backendError", "rateLimitExceeded", "internalError"}:
                 logger.warning(
                     "Transient error listing members for instance group %s: %s; skipping.",
+                    instance_group_name,
+                    e,
+                )
+                return []
+            if reason in {
+                "forbidden",
+                "insufficientPermissions",
+                "IAM_PERMISSION_DENIED",
+            }:
+                logger.warning(
+                    "Missing permissions listing members for instance group %s: %s; skipping members.",
+                    instance_group_name,
+                    e,
+                )
+                return []
+            if is_api_disabled_error(e) or is_billing_disabled_error(e):
+                logger.info(
+                    "Compute API unavailable for listing members of instance group %s: %s; skipping members.",
                     instance_group_name,
                     e,
                 )
--- a/cartography/intel/gcp/secretsmanager.py
+++ b/cartography/intel/gcp/secretsmanager.py
@@ -12,6 +12,9 @@ from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.intel.gcp.util import gcp_api_execute_with_retry
 from cartography.intel.gcp.util import get_error_reason
+from cartography.intel.gcp.util import is_api_disabled_error
+from cartography.intel.gcp.util import is_billing_disabled_error
+from cartography.intel.gcp.util import is_permission_denied_error
 from cartography.models.gcp.secretsmanager.secret import GCPSecretManagerSecretSchema
 from cartography.models.gcp.secretsmanager.secret_version import (
     GCPSecretManagerSecretVersionSchema,
@@ -54,10 +57,32 @@ def get_secrets(secretmanager: Resource, project_id: str) -> List[Dict]:
                 e,
             )
             return []
-        elif reason == "forbidden":
+        elif is_permission_denied_error(e):
             logger.warning(
                 (
                     "You do not have secretmanager.secrets.list access to the project %s. "
+                    "Full details: %s"
+                ),
+                project_id,
+                e,
+            )
+            return []
+        elif is_billing_disabled_error(e):
+            logger.warning(
+                (
+                    "Billing is disabled for project %s. "
+                    "Skipping Secret Manager sync for this project. "
+                    "Full details: %s"
+                ),
+                project_id,
+                e,
+            )
+            return []
+        elif is_api_disabled_error(e):
+            logger.info(
+                (
+                    "Secret Manager API appears disabled for project %s. "
+                    "Skipping Secret Manager sync for this project. "
                     "Full details: %s"
                 ),
                 project_id,
@@ -104,11 +129,31 @@ def get_secret_versions(
                 e,
             )
             return []
-        elif reason == "forbidden":
+        elif is_permission_denied_error(e):
             logger.warning(
                 (
                     "You do not have secretmanager.versions.list access to the secret %s. "
                     "Full details: %s"
+                ),
+                secret_name,
+                e,
+            )
+            return []
+        elif is_billing_disabled_error(e):
+            logger.warning(
+                (
+                    "Billing is disabled while listing versions for secret %s. "
+                    "Skipping versions sync for this secret. Full details: %s"
+                ),
+                secret_name,
+                e,
+            )
+            return []
+        elif is_api_disabled_error(e):
+            logger.info(
+                (
+                    "Secret Manager API appears disabled while listing versions for secret %s. "
+                    "Skipping versions sync for this secret. Full details: %s"
                 ),
                 secret_name,
                 e,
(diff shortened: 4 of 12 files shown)


AdiLyft:
@heryxpc

cartography/cli.py
```diff
@@ -1246,6 +1248,27 @@ def run(
...
+                    "--ubuntu-security-enabled",
+                    help="Enable Ubuntu Security CVE and Notice ingestion.",
+                    rich_help_panel=PANEL_UBUNTU,
+                    hidden=PANEL_UBUNTU not in visible_panels,
+                ),
+            ] = False,
+            ubuntu_security_api_url: Annotated[
+                str | None,
+                typer.Option(
+                    "--ubuntu-security-api-url",
```
heryxpc, line 1265: This probably will never change and if it does, we might notice if the sync breaks.

cartography/intel/ubuntu/cves.py
```diff
@@ -0,0 +1,130 @@
...
+
+@timeit
+def transform(raw_cves: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    result: list[dict[str, Any]] = []
+    for cve in raw_cves:
+        impact = cve.get("impact") or {}
+        base_metric_v3 = impact.get("baseMetricV3") or {}
+        cvss_v3 = base_metric_v3.get("cvssV3") or {}
+        transformed = {
+            "id": cve["id"],
```
heryxpc: If CVE id is like `CVE-XXXX-XXX` then this code will override the NVD `:CVE` nodes as they use the same label and id format.
https://github.com/cartography-cncf/cartography/blob/300097a22fad1bb296df9a8bf5de28ce22dbe55d/tests/integration/cartography/intel/cve/test_feed.py#L46

I suggest using something more specific to the Ubuntu Feed or creating a new id like `USV|CVE-XXXX-XXX`. They follow a similar pattern for [Ubuntu Security Notices](https://ubuntu.com/security/notices/USN-7604-1) (USN)

cartography/intel/ubuntu/cves.py
```diff
@@ -0,0 +1,130 @@
...
+
+@timeit
+def sync(
+    neo4j_session: neo4j.Session,
+    api_url: str,
+    update_tag: int,
+    common_job_parameters: dict[str, Any],
+) -> None:
+    logger.info("Starting Ubuntu CVE sync")
+    load_feed(neo4j_session, api_url, update_tag)
```
heryxpc: `load_feed` might need it's own sync to match the expected pattern.

cartography/intel/ubuntu/cves.py
```diff
@@ -0,0 +1,130 @@
...
+    raw_cves = get(api_url)
+    transformed = transform(raw_cves)
+    load_cves(neo4j_session, transformed, update_tag)
```
heryxpc: Might add more complexity, but instead of querying all the CVEs on every sync, you could:
1. Query if previous sync metadata exists.
2. If exists, find the latest updated date (using  the API `updated_at` field that needs to be added, not the `lastupdated` autogenerated node attributed).
3. Call the api instead with `sort_by=updated` to fetch first the latest CVEs.
4. Get all the CVEs with date later or equal to `updated_at`.
5. Stop querying any more CVEs.

cartography/models/ubuntu/notices.py
```diff
@@ -0,0 +1,59 @@
...
+
+@dataclass(frozen=True)
+class UbuntuSecurityNoticeSchema(CartographyNodeSchema):
+    label: str = "UbuntuSecurityNotice"
+    properties: UbuntuSecurityNoticeNodeProperties = (
+        UbuntuSecurityNoticeNodeProperties()
+    )
+    sub_resource_relationship: None = None
+    scoped_cleanup: bool = False
+    other_relationships: OtherRelationships = OtherRelationships(
```
heryxpc, line 76: You also need a `sub_resource_relationship` if you aim to use automated clean up.

cartography/intel/ubuntu/cves.py
heryxpc: This also involves some complexity on the cleanup process, as the oldest `:UbuntuCVE` nodes won't have the same `lastupdated` value. You might need to update them manually or skip entirely the cleanup process.

AdiLyft: QQ, in order for this optimized approach to work, the very first run would perform the full ingestion and fetch all CVE and perform 3K + API calls with subsequent runs performing way less calls as they are only looking for new updates, correct? Or should we only go to a certain lookback timestamp and ignore any CVEs older than that?

AdiLyft:
@heryxpc 

The Ubuntu CVE/Notice sync now uses incremental sync. Here's how it works:

- On each sync, we create an `UbuntuSyncMetadata` node in Neo4j that stores the timestamp of the most recently seen CVE/notice.
- On the next run, we read that timestamp and only fetch items newer than it from the API. If nothing is new, we skip the load entirely.
- This brings daily API calls down from thousands to just 1-2 pages.

The open question is: **what should happen on the very first run when no `UbuntuSyncMetadata` node exists?**

- **Option A:** Fall back to some fixed date in the past `2026-01-01` — backfills start of this year, but the first sync takes a while because Ubuntu's `sort_by=updated` API is very slow (~10s per page) and we will make a lot of API calls.
- **Option B:** Fall back to "now" — first sync fetches nothing, only catches new CVEs going forward. No historical data.

Which do we want?

cartography/intel/ubuntu/cves.py
heryxpc: I suggest a full sync on first run and later partial updates based on updated dates.
If the first run can fail, a resume feature might be a nice-to-have to avoid restarting from the begining again.

AdiLyft: this is addressed.

cartography/models/ubuntu/notices.py
AdiLyft: added this in.

cartography/intel/ubuntu/cves.py
AdiLyft: Added this in.

cartography/intel/ubuntu/cves.py
AdiLyft: Added this in.

cartography/intel/ubuntu/notices.py
```diff
@@ -0,0 +1,266 @@
...
+        transformed = {
+            "id": notice["id"],
+            "title": notice.get("title"),
+            "summary": notice.get("summary"),
+            "description": notice.get("description"),
+            "published": notice.get("published"),
+            "notice_type": notice.get("type"),
+            "instructions": notice.get("instructions"),
+            "is_hidden": notice.get("is_hidden"),
+            "cves_ids": [f"USV|{cid}" for cid in notice.get("cves_ids", [])],
```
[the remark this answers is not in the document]
AdiLyft: Addressed.

cartography/models/ubuntu/cves.py
```diff
@@ -0,0 +1,64 @@
...
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class UbuntuCVENodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("id")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+    cve_id: PropertyRef = PropertyRef("cve_id", extra_index=True)
```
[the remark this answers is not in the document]
AdiLyft: Addressed.

cartography/intel/ubuntu/feed.py
```diff
@@ -0,0 +1,35 @@
...
+from cartography.models.ubuntu.feed import UbuntuCVEFeedSchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+FEED_ID = "ubuntu-security-cve-feed"
+
+
+@timeit
+def sync(
```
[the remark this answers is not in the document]
AdiLyft: Ignore - we want to persist the data

cartography/models/ubuntu/notices.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+
+@dataclass(frozen=True)
+class UbuntuSecurityNoticeSchema(CartographyNodeSchema):
+    label: str = "UbuntuSecurityNotice"
+    properties: UbuntuSecurityNoticeNodeProperties = (
+        UbuntuSecurityNoticeNodeProperties()
+    )
+    sub_resource_relationship: UbuntuNoticeToUbuntuCVEFeedRel = (
```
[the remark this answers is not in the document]
AdiLyft: Addressed.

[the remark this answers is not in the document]
AdiLyft: Ignored

docs/root/modules/ubuntu/schema.md
```diff
@@ -0,0 +1,79 @@
...
+| notice\_type | The type of notice (e.g., USN) |
+| instructions | Instructions for remediation |
+| is\_hidden | Whether the notice is hidden |
+
+#### Relationships
+
+- UbuntuSecurityNotice nodes are linked to their feed
+
+    ```
+    (UbuntuCVEFeed)-[:RESOURCE]->(UbuntuSecurityNotice)
```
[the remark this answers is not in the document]
AdiLyft: Ignored.

docs/root/modules/ubuntu/schema.md
```diff
@@ -0,0 +1,79 @@
...
+
+### UbuntuCVE
+
+Representation of a CVE from the [Ubuntu Security API](https://ubuntu.com/security/cves). This node also carries the `CVE` label.
+
+| Field | Description |
+|-------|-------------|
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | Unique identifier, prefixed with `USV\|` (e.g., `USV\|CVE-2024-1234`) |
```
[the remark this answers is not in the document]
AdiLyft: Addressed.

cartography/intel/ubuntu/cves.py
```diff
@@ -0,0 +1,283 @@
...
+def get_sync_metadata(neo4j_session: neo4j.Session) -> dict[str, Any]:
+    result = neo4j_session.run(
+        """
+        MATCH (s:UbuntuSyncMetadata {id: $sync_id})
+        RETURN s.full_sync_complete AS full_sync_complete,
+               s.full_sync_offset AS full_sync_offset,
+               s.last_updated_at AS last_updated_at
+        """,
+        sync_id=_SYNC_METADATA_ID,
+    ).single()
```
heryxpc: nit: use [session.execute_read](https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.Session.execute_read) or Cartography's `cartography.client.core.tx.read_single_dict_tx` to use the default transaction retry mechanism.

AdiLyft: Added this in.

AdiLyft:
Hi @kunaals @jychp, can you take a look at this PR? We would like to merge this change in.

PR #2434, opened 2026-02-27
feat(rules): add CIS Kubernetes Benchmark v1.12 framework

jychp:
### Type of change

### Summary

Implements 14 CIS Kubernetes Benchmark v1.12.0 rules covering the controls that can be evaluated using Kubernetes graph data currently ingested by Cartography:

**Section 5.1 – RBAC and Service Accounts (12 rules):**
- 5.1.1: cluster-admin role usage
- 5.1.2: roles granting access to secrets
- 5.1.3: wildcard use in Roles/ClusterRoles
- 5.1.4: roles granting pod creation
- 5.1.5: default service account bindings
- 5.1.7: system:masters group usage
- 5.1.8: bind/impersonate/escalate permissions
- 5.1.9: persistent volume creation access
- 5.1.10: node proxy sub-resource access
- 5.1.11: CSR approval sub-resource access
- 5.1.12: webhook configuration access
- 5.1.13: service account token creation access

**Section 5.4 – Secrets Management (1 rule):**
- 5.4.1: secrets used as environment variables

**Section 5.6 – General Policies (1 rule):**
- 5.6.4: pods in default namespace

All rules use the `Framework(short_name="CIS", scope="kubernetes", revision="1.12")` metadata, enabling CLI filtering via `cartography-rules run --framework cis:kubernetes:1.12`.

### Related issues or links

- Adds Kubernetes to the list of CIS Benchmark frameworks supported by Cartography alongside AWS, GCP, and Google Workspace

### How was this tested?

- 227 unit tests covering rule structure, framework metadata, fact metadata, rule registration, and rule ID conventions
- All Cypher queries verified against the Kubernetes schema documentation (`docs/root/modules/kubernetes/schema.md`) and the actual model definitions in `cartography/models/kubernetes/`
- Linter passes (`make test_lint`)

#### General

#### Proof of functionality

### Notes for reviewers

- Rules that check RBAC resources/verbs (5.1.2–5.1.4, 5.1.8–5.1.13) all exclude `system:` prefixed roles (`NOT cr.name STARTS WITH 'system:'`) to reduce false positives from built-in Kubernetes roles
- Rules that can apply to both cluster-scoped and namespace-scoped roles use multi-fact patterns (separate Facts for ClusterRole and Role)
- Rules for inherently cluster-scoped resources (5.1.10 node proxy, 5.1.11 CSR, 5.1.12 webhooks, 5.1.13 SA tokens) only check ClusterRoles
- Count queries count total assets (not filtered) per the Fact spec convention, consistent with all existing CIS rules

### DIFF
--- a/cartography/rules/data/rules/cis_kubernetes_rbac.py
+++ b/cartography/rules/data/rules/cis_kubernetes_rbac.py
@@ -35,8 +35,8 @@ CIS_REFERENCES = [
 class ClusterAdminUsageOutput(Finding):
     """Output model for cluster-admin role usage check."""
 
-    binding_id: str | None = None
     binding_name: str | None = None
+    binding_id: str | None = None
     subject_type: str | None = None
     subject_name: str | None = None
     cluster_name: str | None = None
@@ -443,11 +443,11 @@ cis_k8s_5_1_4_pod_create_access = Rule(
 class DefaultSaBindingsOutput(Finding):
     """Output model for default service account bindings check."""
 
+    binding_name: str | None = None
     binding_id: str | None = None
+    binding_type: str | None = None
     service_account_name: str | None = None
     namespace: str | None = None
-    binding_name: str | None = None
-    binding_type: str | None = None
     role_name: str | None = None
     cluster_name: str | None = None
 
@@ -561,8 +561,8 @@ cis_k8s_5_1_5_default_sa_bindings = Rule(
 class SystemMastersGroupOutput(Finding):
     """Output model for system:masters group usage check."""
 
-    binding_id: str | None = None
     binding_name: str | None = None
+    binding_id: str | None = None
     binding_type: str | None = None
     role_name: str | None = None
     cluster_name: str | None = None
--- a/cartography/rules/data/rules/cis_kubernetes_workloads.py
+++ b/cartography/rules/data/rules/cis_kubernetes_workloads.py
@@ -35,10 +35,10 @@ CIS_REFERENCES = [
 class SecretsInEnvVarsOutput(Finding):
     """Output model for secrets in environment variables check."""
 
-    pod_id: str | None = None
     pod_name: str | None = None
+    pod_id: str | None = None
     namespace: str | None = None
-    secret_names: str | None = None
+    secret_names: list[str] | None = None
     cluster_name: str | None = None
 
 


cartography/rules/data/rules/cis_kubernetes_workloads.py
```diff
@@ -0,0 +1,165 @@
...
+_k8s_secrets_in_env_vars = Fact(
+    id="k8s_secrets_in_env_vars",
+    name="Kubernetes pods using secrets via environment variables",
+    description=(
+        "Detects pods that reference secrets through environment variables. "
+        "Secrets as environment variables are more susceptible to accidental exposure "
+        "through logging, error messages, or child process inheritance. "
+        "Prefer mounting secrets as files instead."
+    ),
+    cypher_query="""
```
kunaals, line 54: this returns one row per `pod x secret` pair, but `asset_id_field` is `pod_name`, so we get duplicate findings for the same pod when it references multiple secrets. also in multi-cluster setups `pod_name` can collide across clusters/namespaces, which can undercount affected assets. can we either aggregate to one row per pod (with `collect(distinct secret.name)`) or switch to a stable asset id (`pod.id` or a cluster/namespace/name composite)?

cartography/rules/data/rules/cis_kubernetes_rbac.py
```diff
@@ -0,0 +1,1149 @@
...
+    WHERE sa.name = 'default'
+    RETURN COUNT(sa) AS count
+    """,
+    module=Module.KUBERNETES,
+    maturity=Maturity.EXPERIMENTAL,
+)
+
+cis_k8s_5_1_5_default_sa_bindings = Rule(
+    id="cis_k8s_5_1_5_default_sa_bindings",
+    name="CIS K8s 5.1.5: Default Service Accounts Actively Used",
```
kunaals: the rule name says ‘actively used’, but this logic checks bindings to default service accounts, not active pod usage of default service accounts. can we rename this to match current behavior (for example ‘default service account bindings’) or add pod->serviceaccount usage data so it truly reflects active use?

cartography/rules/data/rules/cis_kubernetes_workloads.py
```diff
@@ -0,0 +1,165 @@
...
+_k8s_pods_in_default_namespace = Fact(
+    id="k8s_pods_in_default_namespace",
+    name="Kubernetes pods running in default namespace",
+    description=(
+        "Detects pods running in the default namespace. "
+        "Resources should be segregated into namespaces to allow for "
+        "resource quota management, network policy enforcement, and "
+        "access control separation."
+    ),
+    cypher_query="""
```
kunaals, line 124: this only checks pods in `default`, while cis 5.6.4 is broader (‘default namespace should not be used’). should we either broaden this with additional facts for other namespaced resources (services/secrets/serviceaccounts/rolebindings/etc.) or explicitly scope the rule name/description to pods?

cartography/rules/data/rules/cis_kubernetes_rbac.py
```diff
@@ -0,0 +1,1149 @@
...
+
+_k8s_secret_access_clusterroles = Fact(
+    id="k8s_secret_access_clusterroles",
+    name="Kubernetes ClusterRoles granting access to secrets",
+    description=(
+        "Detects ClusterRoles that grant get, list, or watch access to secrets. "
+        "Access to secrets should be restricted to the smallest possible group "
+        "of users to reduce the risk of privilege escalation."
+    ),
+    cypher_query="""
```
kunaals, line 135: this and related rbac checks are role-definition based, so unbound roles can be flagged. would it be useful to include effective-permission context (for example `binding_count` via optional match on rolebindings/clusterrolebindings), or provide a bound-only variant to reduce false positives?

cartography/rules/data/rules/cis_kubernetes_rbac.py
```diff
@@ -0,0 +1,1149 @@
...
+_k8s_default_sa_cluster_role_bindings = Fact(
+    id="k8s_default_sa_cluster_role_bindings",
+    name="Kubernetes default service accounts with ClusterRoleBindings",
+    description=(
+        "Detects default service accounts that are bound to ClusterRoles "
+        "via ClusterRoleBindings. Default service accounts should not have "
+        "extra privileges to ensure proper auditability of workload permissions."
+    ),
+    cypher_query="""
+    MATCH (cluster:KubernetesCluster)-[:RESOURCE]->(crb:KubernetesClusterRoleBinding)
```
kunaals, line 464: small follow-up on dedup semantics: for cluster-admin/default-sa style findings, using stable ids (`crb.id`/`rb.id`) is safer than display name in aggregated multi-cluster environments, since names can collide.

kunaals (changes requested):
few changes to consider

PR #2435, opened 2026-03-02
feat(rules): add NIST AI RMF aligned AI governance checks

kunaals:
### Type of change

- [x] Documentation update

### Summary

This PR adds a NIST AI RMF-aligned rules pack to cover AI governance questions that are not addressed well by the existing rule sets.

The goal is to help teams answer practical governance and risk questions inside Cartography's existing rules workflow: what AI-connected third-party apps are in use, whether those apps have high-risk grants, whether privileged identities are authorizing them, whether deployed AI agents are visible in runtime inventory, whether that inventory has coverage gaps, and whether provider API keys are stale or lack clear ownership.

These checks are intended to improve visibility and prioritization for emerging AI usage patterns, not to make formal certification claims.

### Related issues or links

- none

### Breaking changes

None.

### How was this tested?

- `uv run pytest tests/unit/rules/test_nist_ai_rmf.py tests/unit/rules/test_cis_4_0_gcp.py tests/unit/rules/test_cis_google_workspace.py`
- `uv run cartography-rules list --framework NIST-AI-RMF`
- `NEO4J_PASSWORD=dummy uv run cartography-rules run nist_ai_third_party_app_inventory --no-experimental --neo4j-password-env-var NEO4J_PASSWORD`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

- All new facts remain `EXPERIMENTAL` because AI-provider coverage and AIBOM-driven inventory are still evolving.
- The AIBOM additions are there to close a visibility gap between identity-side AI usage and runtime-deployed AI systems.
- Review follow-ups in this PR are focused on keeping the resulting governance and compliance output trustworthy as coverage expands.


cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -0,0 +1,822 @@
...
+_openai_nist_ai_stale_or_unowned_api_keys = Fact(
+    id="openai_nist_ai_stale_or_unowned_api_keys",
+    name="OpenAI API keys stale/unused or lacking owner attribution",
+    description=(
+        "Finds OpenAI API keys that are stale/unused (90+ days) or lack clear "
+        "owner attribution."
+    ),
+    cypher_query="""
+    MATCH (org:OpenAIOrganization)-[:RESOURCE]->(k)
+    WHERE k:OpenAIApiKey OR k:OpenAIAdminApiKey
```
jychp, line 666: `MATCH (org:OpenAIOrganization)-[:RESOURCE]->(k)` only reaches `OpenAIAdminApiKey` nodes. Regular `OpenAIApiKey` nodes are ingested under `(:OpenAIProject)-[:RESOURCE]->(:OpenAIApiKey)`, not directly under the org. This query silently drops all project-scoped API keys.

Consider traversing through projects, e.g. `MATCH (org:OpenAIOrganization)-[:RESOURCE*1..2]->(k) WHERE k:OpenAIApiKey OR k:OpenAIAdminApiKey`.

cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -0,0 +1,822 @@
...
+    cypher_count_query="""
+    MATCH (source:AIBOMSource)
+    WHERE
+        coalesce(source.image_matched, false) = false
+        OR toLower(coalesce(source.source_status, 'completed')) <> 'completed'
+        OR (
+            source.analysis_status IS NOT NULL
+            AND toLower(source.analysis_status) <> 'completed'
+        )
+    RETURN COUNT(source) AS count
```
jychp, line 603: suggestion: This count query applies the gap WHERE filter, so it only counts failing assets. Per convention, `cypher_count_query` should count ALL assets of the evaluated type regardless of fact criteria. This makes Total == Failing and Passing always 0.

Should be just:
```cypher
MATCH (source:AIBOMSource)
RETURN COUNT(source) AS count
```

cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -0,0 +1,822 @@
...
+        size(model_names) AS model_count,
+        model_names,
+        size(tool_names) AS tool_count,
+        tool_names,
+        size(memory_names) AS memory_count,
+        memory_names,
+        size(prompt_names) AS prompt_count,
+        prompt_names,
+        count(DISTINCT embedding) AS embedding_count,
+        collect(DISTINCT embedding.name) AS embedding_names
```
jychp: suggestion: Embeddings are aggregated directly in RETURN instead of using an intermediate `WITH ... collect(...)` pass like the other component types (models, tools, memory, prompts). Consider making this consistent to avoid subtle grouping issues if the query evolves.

cartography/rules/data/rules/nist_ai_rmf.py
kunaals: good catch. fixed in d838b9d7.

cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -0,0 +1,566 @@
...
+            ANY(term IN {AI_ALLOWLIST_TERMS_CYPHER}
+                WHERE normalized_name CONTAINS term OR normalized_client_id CONTAINS term
+            )
+            OR normalized_name =~ '{AI_HEURISTIC_REGEX}'
+            OR normalized_client_id =~ '{AI_HEURISTIC_REGEX}'
+      )
+    RETURN *
+    """,
+    cypher_count_query="""
+    MATCH (u:GoogleWorkspaceUser)
```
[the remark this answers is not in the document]
kunaals: fixed earlier in dd4e95a.

cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -0,0 +1,574 @@
...
+            OR normalized_name =~ '{AI_HEURISTIC_REGEX}'
+            OR normalized_client_id =~ '{AI_HEURISTIC_REGEX}'
+      )
+    RETURN *
+    """,
+    cypher_count_query=f"""
+    MATCH (app:ThirdPartyApp)
+    WITH
+        toLower(coalesce(app._ont_name, app.display_name, app.display_text, app.name, '')) AS normalized_name,
+        toLower(coalesce(app._ont_client_id, app.client_id, app.app_id, app.id, '')) AS normalized_client_id
```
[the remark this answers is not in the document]
kunaals: fixed earlier in 6328403.

cartography/rules/data/rules/nist_ai_rmf.py
kunaals: good catch. fixed in d838b9d7.

cartography/rules/data/rules/nist_ai_rmf.py
kunaals: fixed in 706c2b76

kunaals:
@jychp this rule surfaced something odd: i’m seeing duplicate openai key findings where the same key id shows up both as a project-scoped key and as an admin key. i think you've mentioned this sort of issue before

i see it's coming from the openai intel module rather than just the query. openai’s admin api docs say [/organization/admin_api_keys](https://platform.openai.com/docs/api-reference/admin-api-keys) returns “a list of admin and project api key objects”, and we also separately ingest project keys from [/organization/projects/{project_id}/api_keys](https://platform.openai.com/docs/api-reference/project-api-keys/list). so it looks possible for the same project key to land in cartography twice under different labels, same ID

you’ve raised this kind of duplicate-identity issue before, so i wanted to sanity check with you: does the right fix here seem like filtering the intel-module admin sync down to real organization.admin_api_key objects, rather than trying to paper over it in this rule using like `DISTINCT k.id`?

if so, i’m thinking we should make that intel-module fix before merging this pr. i can do that if you think its best

kunaals:
above comment will be fixed by https://github.com/cartography-cncf/cartography/pull/2511

Issue #2389, opened 2026-02-16
feat(gcp): Add unified GCP Labels support across all resource types


kunaals:
# feat(gcp): Add unified GCP Labels support across all resource types

## Description

Add a centralized `GCPLabel` node type that ingests GCP labels across all supported resource types, similar to how `AWSTag` works for AWS resources.

Currently, cartography only ingests labels for GCP Storage Buckets (`GCPBucketLabel`), and the existing implementation has a broken ID format (`GCPBucket_{key}` — missing the value and bucket ID, causing collisions across buckets with the same label key). No other GCP resource type has label support despite labels being available on all major GCP resources.

**Proposed design:**
- Create a unified `GCPLabel` node (with extra label `Label`) using template-based Cypher queries (same proven pattern as `AWSTag`)
- ID format: `{resource_id}:{key}:{value}` — globally unique, fixes the bucket collision bug
- Relationship: `(resource)-[:LABELED]->(GCPLabel)`
- Extract labels from each resource's existing API response (no new API calls needed)
- Backward compatibility: apply `GCPBucketLabel` as a conditional node label on `GCPLabel` nodes where `resource_type = "GCPBucket"`

**GCP resources to support (Phase 1):**
- `GCPInstance` (Compute Engine)
- `GCPBucket` (Cloud Storage)
- `GKECluster` (Google Kubernetes Engine)
- `GCPBigtableInstance` (Cloud Bigtable)
- `GCPCloudSQLInstance` (Cloud SQL)
- `GCPDNSZone` (Cloud DNS)
- `GCPSecretManagerSecret` (Secret Manager)
- `GCPCloudRunService` (Cloud Run services)
- `GCPCloudRunJob` (Cloud Run jobs)

**Example queries enabled:**
```cypher
-- Find all resources with a specific label
MATCH (resource)-[:LABELED]->(l:GCPLabel {key: "env", value: "production"})
RETURN resource

-- Find all labels on a specific instance
MATCH (i:GCPInstance {name: "my-instance"})-[:LABELED]->(l:GCPLabel)
RETURN l.key, l.value

-- Cross-resource label search (all resources labeled "team=backend")
MATCH (resource)-[:LABELED]->(l:Label {key: "team", value: "backend"})
RETURN labels(resource), resource.id

-- Backward-compatible bucket label query
MATCH (l:GCPBucketLabel)
RETURN l.key, l.value
```

## Motivation

[...]

PR #2437, opened 2026-03-02
feat(gcp): Add GCP Labels Support for GCPInstances and GCPBuckets

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

This PR adds a GCP Label node type for GCPInstances an GCPBuckets. Once merged a seperate PR will be made to wire the rest of the resources to create labels for the rest of the GCP resources. 

### Related issues or links

- Fixes https://github.com/cartography-cncf/cartography/issues/2389

### How was this tested?

- Added asserts to GCPBuckets integration test (test_storage.py) to verify the pattern. 
- graph
image

#### General

#### Proof of functionality


cartography/intel/gcp/compute.py
```diff
@@ -1191,6 +1192,15 @@ def sync_gcp_instances(
     instance_list = transform_gcp_instances(instance_responses)
     load_gcp_instances(neo4j_session, instance_list, gcp_update_tag, project_id)
     cleanup_gcp_instances(neo4j_session, common_job_parameters)
+    # Sync unified GCPLabel nodes - use transformed list since it has partial_uri set
+    sync_labels(
```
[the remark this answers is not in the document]
shyammukund: The buckets we just loaded in load_gcp_buckets all have the current UPDATE_TAG, so cleanup_gcp_buckets won't delete them. It only deletes buckets from previous sync runs. And the raw_buckets list only contains buckets from the current API response — the same ones we just loaded. So the MATCH will always find them. So the order is safe. The cleanup only removes old data, and we're only creating labels for data we just loaded.

cartography/intel/gcp/labels.py
```diff
@@ -0,0 +1,378 @@
...
+            **kwargs,
+        )
+        total_deleted += deleted
+        if deleted == 0:
+            break
+    return total_deleted
+
+
+@timeit
+def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
```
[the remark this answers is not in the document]
shyammukund: It's fine because GraphJob.from_node_schema() generates cleanup queries based on the schema's sub_resource_relationship — it traces the path from the node back to its parent (e.g., GCPProject) to   
  scope the deletion. But our GCPLabelSchema has sub_resource_relationship: None because the nodes are created via template Cypher, not the declarative load() function. There's no schema-defined      
  relationship path for GraphJob to use.

cartography/intel/gcp/labels.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+    mapping = LABEL_RESOURCE_TYPE_MAPPINGS.get(resource_type)
+    if not mapping:
+        logger.warning("Unknown GCP resource type for labels: %s", resource_type)
+        return []
+
+    labels_field = mapping["labels_field"]
+    # id_field is the key in the resource dict to use as the resource ID.
+    # Falls back to "property" which is also used for the Cypher match.
+    id_field = mapping.get("id_field", mapping["property"])
```
kunaals: this id_field fallback to mapping["property"] looks unsafe for most mapped resource types. raw gcp payloads often use name/selflink while node match property is id, so labels get silently dropped. can we add explicit id_field per resource type and add tests for cloud sql/gke/bigtable/secret manager/cloud run?

cartography/intel/gcp/labels.py
```diff
@@ -0,0 +1,378 @@
...
+    return label_data
+
+
+def _load_labels_tx(
+    tx: neo4j.Transaction,
+    label_data: List[Dict],
+    resource_type: str,
+    project_id: str,
+    update_tag: int,
+) -> None:
```
kunaals: calling cleanup() inside each sync_labels() invocation is risky with the current global cleanup logic. this should probably move to a single orchestrated cleanup pass per project/update_tag.

cartography/intel/gcp/labels.py
```diff
@@ -0,0 +1,378 @@
...
+
+    Iterates over each supported resource type and removes labels and relationships
+    that were not updated in the current sync run. Also removes orphaned label nodes
+    that have no remaining relationships.
+    """
+    cleanup_batch_size = common_job_parameters.get(
+        "gcp_label_cleanup_batch",
+        DEFAULT_CLEANUP_BATCH_SIZE,
+    )
+    for _label, path in _RESOURCE_CLEANUP_PATHS.items():
```
kunaals: cleanup currently loops all resource types every time sync_labels runs. this causes cross-resource data loss on partial syncs (e.g. compute sync deleting bucket unified labels). can we scope cleanup to current resource_type or defer unified cleanup to one project-level pass after all label syncs complete?

cartography/intel/gcp/labels.py
```diff
@@ -0,0 +1,378 @@
...
+    update_tag: int,
+) -> None:
+    """
+    Transaction function that creates GCPLabel nodes and LABELED relationships
+    using a template-based Cypher query.
+
+    The template substitutes the resource's node label and ID property so that the same
+    query pattern works across all GCP resource types.
+    """
+    INGEST_LABEL_TEMPLATE = Template(
```
kunaals: this docstring mentions backward-compatible extra labels (like gcpbucketlabel), but this ingest path is raw cypher. schema-level conditional labels only run via load(), so this currently won’t apply those labels automatically.

cartography/models/gcp/labels/label.py
```diff
@@ -0,0 +1,43 @@
...
+    template-based queries because GCPLabel has dynamic LABELED relationships to many
+    different resource types (GCPInstance, GCPBucket, GKECluster, etc.). The cleanup is also
+    handled manually due to this dynamic nature.
+
+    The LABELED relationship goes FROM the resource TO the GCPLabel:
+    (resource)-[:LABELED]->(GCPLabel)
+    """
+
+    label: str = "GCPLabel"
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Label"])
```
kunaals, line 117: if we want backward compatibility, can we add a conditional node label here for bucket-backed labels (e.g. conditionalnodelabel with resource_type == "gcpbucket")?

docs/root/modules/gcp/schema.md
```diff
@@ -224,6 +224,29 @@ Representation of a GCP [Storage Bucket Label](https://cloud.google.com/storage/
     ```
 
 
+### Label::GCPLabel
+
+Representation of a GCP [Label](https://cloud.google.com/resource-manager/docs/labels-overview). GCP Labels can be applied to many resource types. This is a unified label node type similar to how `AWSTag` works for AWS resources.
+
```
kunaals, line 230: can we scope this wording to what this pr actually syncs today (compute + bucket)? right now the text reads like broad resource coverage, but those code paths aren’t wired yet.

tests/integration/cartography/intel/gcp/test_storage.py
```diff
@@ -170,3 +170,36 @@ def test_sync_gcp_buckets(mock_get_buckets, neo4j_session):
         ("project-abc", "GCPBucket_label_key_1"),
         ("project-abc", "GCPBucket_label_key_2"),
     }
+
+    # Assert - unified GCPLabel nodes also created by sync_labels
```
kunaals: nice to see unified label assertions added for storage. can we also add regression tests for (1) id_field mapping across non-bucket types and (2) partial-sync behavior so one resource sync doesn’t delete another resource type’s unified labels?

cartography/intel/gcp/labels.py
shyammukund: Decided to remove the mapping for the resources that haven't been wired into the sync yet for now. Will add back on follow up PR.

cartography/intel/gcp/labels.py
shyammukund: cleanup now takes a resource_type parameter and only cleans up labels for that specific resource type.

tests/integration/cartography/intel/gcp/test_storage.py
shyammukund: Added a unit test to make sure the id_field mapping works as intended. I believe that with the change of our cleanup scoped to the current resource type we shouldn't have to worry about  one resources sync deleting another resources labels.

PR #2438, opened 2026-03-02
fix(ci): unblock arm64 Docker publish and harden PyPI version check

kunaals:
### Type of change

### Summary
This PR resolves recent release pipeline failures and removes a risky archived dependency.

Why these changes:
1. `dockerfile` is archived upstream and had Linux arm64 wheel gaps, making release reliability fragile.
2. `pip index ... | grep` was brittle for prerelease version detection.

What changed:
- Replaced external Dockerfile parser usage with a lightweight in-repo parser in `cartography/intel/supply_chain.py`.
- Removed `dockerfile>=3.0.0` from dependencies and updated `uv.lock`.
- Added parser tests for multiline continuation and heredoc handling.
- In `.github/workflows/publish-to-ghcr-and-pypi.yml`, kept master’s trigger/guard behavior and switched PyPI version polling to the exact, version-specific JSON endpoint:
  - `if: github.event_name == 'release'`
  - `curl --fail --silent --show-error "https://pypi.org/pypi/cartography/$VERSION/json"`

Review follow-up (nits addressed):
- Inlined parser logic into `_parse_instructions` (removed stale helper naming).
- Removed stale exception-oriented parser docstring wording.
- Added a warning log for unterminated heredoc delimiters.
- Added a code comment documenting the known simplification for comment-only continuation lines.
- Added additional tests for quoted heredoc delimiters and 3-line continuations.

### Breaking changes
None.

### Helpful logs
GitHub Actions failure analysis (`gh run view --log-failed`):
- Docker arm64 failure (`0.131.0`):
  ```
  × Failed to build `dockerfile==3.4.0`
  $ GOPATH=/tmp/tmp... go get -d
  error: [Errno 2] No such file or directory: 'go'
  ```
- Prerelease PyPI polling failure (`0.130.1rc1`):
  ```
  Attempt 30/30: Package not yet available...
  ❌ Failed to verify cartography==0.130.1rc1 on PyPI
  ```

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
- Parser dependency usage was isolated to `supply_chain.py`, so replacement risk is contained and covered by targeted tests.
- Workflow behavior matches `master` for triggers/guards, with only the version-availability polling method hardened.

### DIFF
--- a/.github/workflows/publish-to-ghcr-and-pypi.yml
+++ b/.github/workflows/publish-to-ghcr-and-pypi.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Build a binary wheel and a source tarball
         run: uv build
       - name: Publish distribution 📦 to PyPI
-        if: startsWith(github.ref, 'refs/tags')
+        if: github.event_name == 'release'
         # --check-url checks if files already exist on PyPI before uploading.
         # If the version already exists, it skips without failing, allowing
         # downstream jobs (Docker build) to proceed on re-runs.


.github/workflows/publish-to-ghcr-and-pypi.yml
```diff
@@ -104,14 +100,15 @@ jobs:
...
+          VERSION="${VERSION#v}"
           echo "VERSION=$VERSION" >> $GITHUB_ENV
           echo "Waiting for cartography==$VERSION to be available on PyPI..."
 
           # Poll PyPI for up to 5 minutes
           MAX_ATTEMPTS=30
           ATTEMPT=0
           while [ $ATTEMPT -lt $MAX_ATTEMPTS ]; do
-            if pip index versions cartography 2>&1 | grep -q "$VERSION"; then
+            if curl --fail --silent --show-error "https://pypi.org/pypi/cartography/$VERSION/json" > /dev/null; then
```
achantavy, line 114: whats this for?

pyproject.toml
```diff
@@ -34,7 +34,6 @@ classifiers = [
 requires-python = ">=3.10"
 dependencies = [
     "backoff>=2.1.2",
-    "dockerfile>=3.0.0",
```
achantavy, line 37: pour one out for dockerfile 🫡

.github/workflows/publish-to-ghcr-and-pypi.yml
kunaals: pip index versions lists all available versions. for a prerelease like 0.130.1rc1, pip may not include it in the output without --pre, which is what caused this failure described. grep on that output is a substring match, 0.130 would match 0.130.0

curl .../cartography/$VERSION/json is an exact version lookup and it returns 200s or 404s, no ambiguity

this will ensure this kinda prerelease failure doesn't happen again

.github/workflows/publish-to-ghcr-and-pypi.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Build a binary wheel and a source tarball
         run: uv build
       - name: Publish distribution 📦 to PyPI
-        if: startsWith(github.ref, 'refs/tags')
+        if: github.event_name == 'release'
```
kunaals, line 31: switched PyPI publish guard to this to make intent explicit and prevent manual workflow_dispatch reruns from attempting to re-upload immutable artifacts

PR #2439, opened 2026-03-03
feat(sentry): add Sentry intel module

jychp:
### Type of change

### Summary

Add a new Cartography intel module for [Sentry](https://sentry.io), the error monitoring and performance tracking platform.

**Resources synced:**
- **SentryOrganization** — organizations (with `Tenant` semantic label)
- **SentryTeam** — teams (with `UserGroup` semantic label)
- **SentryUser** — organization members (with `UserAccount` semantic label)
- **SentryProject** — projects
- **SentryRelease** — releases
- **SentryAlertRule** — issue alert rules

**Graph relationships:**
```
SentryOrganization -[:RESOURCE]-> SentryTeam / SentryUser / SentryProject / SentryRelease
SentryProject -[:RESOURCE]-> SentryAlertRule
SentryUser -[:MEMBER_OF]-> SentryTeam
SentryProject -[:HAS_TEAM]-> SentryTeam
```

**Ontology mappings** registered for `Tenant`, `UserAccount`, and `UserGroup` semantic labels.

Supports both Sentry SaaS and self-hosted instances via `--sentry-host`.

### Related issues or links

- New module, no related issues.

### How was this tested?

- 6 integration tests covering all resource types (organizations, teams, members, projects, releases, alert rules)
- Tests verify node creation, semantic labels, and all relationship types (RESOURCE, MEMBER_OF, HAS_TEAM)
- Full lint suite passes (`make lint`)

#### General

#### Proof of functionality

### Notes for reviewers

- Authentication uses Bearer token via `--sentry-api-key-env-var` (reads from env var)
- Cursor-based pagination follows Sentry's Link header convention (`rel="next"; results="true/false"`)
- Release IDs use the `version` string (globally unique) rather than the integer `id` from the API (which may collide across orgs)
- Alert rules are synced per-project since the Sentry API scopes them to projects
- Member-to-team resolution uses `teamRoles` from the members API endpoint, mapping team slugs to IDs


jychp:
carbon (1)

jychp:
Screenshot 2026-03-14 at 10 52 21 AM

cartography/intel/sentry/releases.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+
+@timeit
+def transform(raw_releases: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    result: list[dict[str, Any]] = []
+    for release in raw_releases:
+        r = release.copy()
+        # Use version as id since release "id" is an integer that may collide across orgs
+        r["id"] = release["version"]
```
kunaals: am i reading this right that release ids are just the version string now? if `--sentry-org` is unset and we sync more than one org, wouldn’t the same version in two orgs collapse into one `SentryRelease` node?

cartography/intel/sentry/alertrules.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+
+def cleanup(
+    neo4j_session: neo4j.Session,
+    common_job_parameters: dict[str, Any],
+    project_id: str,
+) -> None:
+    # Alert rules are sub-resources of projects, so we need per-project cleanup
+    params = {**common_job_parameters, "PROJECT_ID": project_id}
```
kunaals: is there another cleanup path for alert rules that i’m missing? this cleanup only runs for project ids we still saw in the current sync, so if a project disappears from sentry it seems like its `SentryAlertRule` nodes would be left behind as orphans.

cartography/intel/sentry/members.py
```diff
@@ -0,0 +1,116 @@
...
+    result: list[dict[str, Any]] = []
+    for member in raw_members:
+        m = member.copy()
+        m["id"] = member["id"]
+        # Extract user details if the member has accepted
+        user = member.get("user") or {}
+        m["has2fa"] = user.get("has2fa")
+        # Owners are implicit admins of all teams
+        if member.get("orgRole") == "owner":
+            m["team_ids"] = all_team_ids
```
kunaals, line 84: maybe i’m misunderstanding the sentry role model, but should org owners really get `MEMBER_OF` and `ADMIN_OF` edges to every team here? this looks like it’s turning org-level ownership into explicit membership in all teams, which feels like it could overstate access in the graph.

PR #2443, opened 2026-03-03
feat(aibom): ingest AI BOM reports and link detections to canonical ECR images

kunaals:
### Type of change
- [x] Documentation update

### Summary
Cartography already models container image identity and vulnerability context, but it did not have first-class visibility into whether a container image contains AI-agent behavior or tool usage patterns.

This change adds an ingestion path for Cisco AIBOM JSON so teams can ask graph-native questions like:
- Which shipped images currently contain AI agents?
- Which tools are present in those agent-enabled images?
- Which workflow contexts are those detections associated with?

The design links detections to the most canonical ECR image available for a tag: manifest lists first, with fallback to a single concrete image when no manifest list exists. This avoids duplicate detections across per-architecture child manifests while still supporting single-platform images.

The module is ingestion-only (no scanner execution in Cartography), so organizations can keep scanning and policy controls in their existing CI/CD or artifact pipelines while still centralizing AI-component intelligence in the graph.

### Related issues or links
- N/A

### How was this tested?
- added tests
- ran locally

#### General

#### Proof of functionality

Screenshot 2026-03-10 at 12 15 56 AM

### Notes for reviewers
- This PR is intentionally ECR-first and keeps AIBOM detections attached to a single canonical image node per source image URI.
- Non-ECR registries and scanner execution orchestration can be added later without changing the stored AIBOM node model.

### Follow up issues:
1. https://github.com/cartography-cncf/cartography/issues/2476
2. https://github.com/cartography-cncf/cartography/issues/2477

### DIFF
--- a/cartography/intel/gcp/artifact_registry/artifact.py
+++ b/cartography/intel/gcp/artifact_registry/artifact.py
@@ -250,38 +250,42 @@ def get_go_modules(client: Resource, repository_name: str) -> list[dict]:
 @timeit
 def get_apt_artifacts(client: Resource, repository_name: str) -> list[dict]:
     """
-    Gets APT artifacts for a repository.
+    Gets APT package versions for a repository.
 
     :param client: The Artifact Registry API client.
     :param repository_name: The full repository resource name.
-    :return: List of APT artifact dicts from the API.
+    :return: List of APT package-version dicts from the API.
     """
     artifacts: list[dict] = []
     try:
-        request = (
-            client.projects()
-            .locations()
-            .repositories()
-            .aptArtifacts()
-            .list(parent=repository_name)
-        )
-        while request is not None:
-            response = request.execute()
-            artifacts.extend(response.get("aptArtifacts", []))
-            request = (
-                client.projects()
-                .locations()
-                .repositories()
-                .aptArtifacts()
-                .list_next(
-                    previous_request=request,
-                    previous_response=response,
+        packages_resource = client.projects().locations().repositories().packages()
+        packages_request = packages_resource.list(parent=repository_name)
+        while packages_request is not None:
+            packages_response = packages_request.execute()
+            for package in packages_response.get("packages", []):
+                package_name = (
+                    package.get("displayName")
+                    or package.get("name", "").split("/packages/")[-1]
                 )
+                versions_resource = packages_resource.versions()
+                versions_request = versions_resource.list(parent=package.get("name"))
+                while versions_request is not None:
+                    versions_response = versions_request.execute()
+                    for version in versions_response.get("versions", []):
+                        version["packageName"] = package_name
+                        artifacts.append(version)
+                    versions_request = versions_resource.list_next(
+                        previous_request=versions_request,
+                        previous_response=versions_response,
+                    )
+            packages_request = packages_resource.list_next(
+                previous_request=packages_request,
+                previous_response=packages_response,
             )
         return artifacts
     except (PermissionDenied, DefaultCredentialsError, RefreshError) as e:
         logger.warning(
-            f"Failed to get APT artifacts for repository {repository_name} "
+            f"Failed to get APT package versions for repository {repository_name} "
             f"due to permissions or auth error: {e}",
         )
         return []
@@ -290,38 +294,42 @@ def get_apt_artifacts(client: Resource, repository_name: str) -> list[dict]:
 @timeit
 def get_yum_artifacts(client: Resource, repository_name: str) -> list[dict]:
     """
-    Gets YUM artifacts for a repository.
+    Gets YUM package versions for a repository.
 
     :param client: The Artifact Registry API client.
     :param repository_name: The full repository resource name.
-    :return: List of YUM artifact dicts from the API.
+    :return: List of YUM package-version dicts from the API.
     """
     artifacts: list[dict] = []
     try:
-        request = (
-            client.projects()
-            .locations()
-            .repositories()
-            .yumArtifacts()
-            .list(parent=repository_name)
-        )
-        while request is not None:
-            response = request.execute()
-            artifacts.extend(response.get("yumArtifacts", []))
-            request = (
-                client.projects()
-                .locations()
-                .repositories()
-                .yumArtifacts()
-                .list_next(
-                    previous_request=request,
-                    previous_response=response,
+        packages_resource = client.projects().locations().repositories().packages()
+        packages_request = packages_resource.list(parent=repository_name)
+        while packages_request is not None:
+            packages_response = packages_request.execute()
+            for package in packages_response.get("packages", []):
+                package_name = (
+                    package.get("displayName")
+                    or package.get("name", "").split("/packages/")[-1]
                 )
+                versions_resource = packages_resource.versions()
+                versions_request = versions_resource.list(parent=package.get("name"))
+                while versions_request is not None:
+                    versions_response = versions_request.execute()
+                    for version in versions_response.get("versions", []):
+                        version["packageName"] = package_name
+                        artifacts.append(version)
+                    versions_request = versions_resource.list_next(
+                        previous_request=versions_request,
+                        previous_response=versions_response,
+                    )
+            packages_request = packages_resource.list_next(
+                previous_request=packages_request,
+                previous_response=packages_response,
             )
         return artifacts
     except (PermissionDenied, DefaultCredentialsError, RefreshError) as e:
         logger.warning(
-            f"Failed to get YUM artifacts for repository {repository_name} "
+            f"Failed to get YUM package versions for repository {repository_name} "
             f"due to permissions or auth error: {e}",
         )
         return []
--- a/tests/data/gcp/artifact_registry.py
+++ b/tests/data/gcp/artifact_registry.py
@@ -21,6 +21,24 @@ MOCK_REPOSITORIES = [
         "createTime": "2024-01-02T00:00:00Z",
         "updateTime": "2024-01-16T00:00:00Z",
     },
+    {
+        "name": "projects/test-project/locations/us-east1/repositories/apt-repo",
+        "format": "APT",
+        "mode": "STANDARD_REPOSITORY",
+        "description": "APT packages repository",
+        "sizeBytes": "256000",
+        "createTime": "2024-01-03T00:00:00Z",
+        "updateTime": "2024-01-17T00:00:00Z",
+    },
+    {
+        "name": "projects/test-project/locations/us-east1/repositories/yum-repo",
+        "format": "YUM",
+        "mode": "STANDARD_REPOSITORY",
+        "description": "YUM packages repository",
+        "sizeBytes": "384000",
+        "createTime": "2024-01-04T00:00:00Z",
+        "updateTime": "2024-01-18T00:00:00Z",
+    },
 ]
 
 # Manifest list data for multi-arch images (returned in imageManifests field)
@@ -80,6 +98,24 @@ MOCK_MAVEN_ARTIFACTS = [
         "version": "1.0.0",
         "createTime": "2024-01-05T00:00:00Z",
         "updateTime": "2024-01-05T00:00:00Z",
+    },
+]
+
+MOCK_APT_ARTIFACTS = [
+    {
+        "name": "projects/test-project/locations/us-east1/repositories/apt-repo/packages/curl/versions/7.88.1",
+        "packageName": "curl",
+        "createTime": "2024-01-06T00:00:00Z",
+        "updateTime": "2024-01-06T00:00:00Z",
+    },
+]
+
+MOCK_YUM_ARTIFACTS = [
+    {
+        "name": "projects/test-project/locations/us-east1/repositories/yum-repo/packages/bash/versions/5.2.26",
+        "packageName": "bash",
+        "createTime": "2024-01-07T00:00:00Z",
+        "updateTime": "2024-01-07T00:00:00Z",
     },
 ]
 
--- a/tests/integration/cartography/intel/gcp/test_artifact_registry.py
+++ b/tests/integration/cartography/intel/gcp/test_artifact_registry.py
@@ -2,12 +2,16 @@ from unittest.mock import MagicMock
 from unittest.mock import patch
 
 from cartography.intel.gcp.artifact_registry import sync
+from cartography.intel.gcp.artifact_registry.artifact import transform_apt_artifacts
 from cartography.intel.gcp.artifact_registry.artifact import transform_docker_images
 from cartography.intel.gcp.artifact_registry.artifact import transform_maven_artifacts
+from cartography.intel.gcp.artifact_registry.artifact import transform_yum_artifacts
+from tests.data.gcp.artifact_registry import MOCK_APT_ARTIFACTS
 from tests.data.gcp.artifact_registry import MOCK_DOCKER_IMAGES
 from tests.data.gcp.artifact_registry import MOCK_HELM_CHARTS
 from tests.data.gcp.artifact_registry import MOCK_MAVEN_ARTIFACTS
 from tests.data.gcp.artifact_registry import MOCK_REPOSITORIES
+from tests.data.gcp.artifact_registry import MOCK_YUM_ARTIFACTS
 from tests.integration.util import check_nodes
 from tests.integration.util import check_rels
 
@@ -19,9 +23,13 @@ TEST_DOCKER_REPO_ID = (
 TEST_MAVEN_REPO_ID = (
     "projects/test-project/locations/us-central1/repositories/maven-repo"
 )
+TEST_APT_REPO_ID = "projects/test-project/locations/us-east1/repositories/apt-repo"
+TEST_YUM_REPO_ID = "projects/test-project/locations/us-east1/repositories/yum-repo"
 TEST_DOCKER_IMAGE_ID = "projects/test-project/locations/us-central1/repositories/docker-repo/dockerImages/my-app@sha256:abc123"
 TEST_HELM_CHART_ID = "projects/test-project/locations/us-central1/repositories/docker-repo/dockerImages/my-chart@sha256:xyz789"
 TEST_MAVEN_ARTIFACT_ID = "projects/test-project/locations/us-central1/repositories/maven-repo/mavenArtifacts/com.example:my-lib:1.0.0"
+TEST_APT_ARTIFACT_ID = "projects/test-project/locations/us-east1/repositories/apt-repo/packages/curl/versions/7.88.1"
+TEST_YUM_ARTIFACT_ID = "projects/test-project/locations/us-east1/repositories/yum-repo/packages/bash/versions/5.2.26"
 TEST_PLATFORM_IMAGE_AMD64_ID = f"{TEST_DOCKER_IMAGE_ID}@sha256:def456"
 TEST_PLATFORM_IMAGE_ARM64_ID = f"{TEST_DOCKER_IMAGE_ID}@sha256:ghi789"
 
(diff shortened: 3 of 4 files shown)


docs/root/modules/aibom/schema.md
```diff
@@ -0,0 +1,56 @@
...
+| scanner_version | Scanner version |
+| scan_scope | Scanner input scope |
+| **manifest_digest** | Manifest-list digest used for graph linking |
+
+#### Relationships
+
+- An AIBOMComponent is detected in a manifest-list ECR image.
+
+    ```
+    (AIBOMComponent)-[:DETECTED_IN]->(ECRImage::ImageManifestList)
```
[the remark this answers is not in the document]
kunaals: Addressed in `06a6b78bb`. I replaced the invalid `::` example with valid Cypher syntax in the schema docs.

cartography/models/aibom/component.py
```diff
@@ -0,0 +1,79 @@
...
+    rel_label: str = "IN_WORKFLOW"
+    properties: AIBOMComponentInWorkflowRelProperties = (
+        AIBOMComponentInWorkflowRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class AIBOMComponentSchema(CartographyNodeSchema):
+    label: str = "AIBOMComponent"
+    scoped_cleanup: bool = False
```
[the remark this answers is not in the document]
kunaals: I do not think this one is valid. AIBOM nodes are modeled as unscoped feed-style data here: there is no tenant-like sub-resource relationship, and cleanup is keyed only by UPDATE_TAG. That matches other global ingestion models that also use `scoped_cleanup=False` when there is no tenant anchor to scope against.

cartography/intel/aibom/__init__.py
```diff
@@ -0,0 +1,199 @@
+import json
```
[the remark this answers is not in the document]
kunaals: I kept this conditional intentionally. Because AIBOM cleanup is unscoped, running it after a batch where every document was skipped for invalid input, decode errors, or unmatched images would delete previously ingested AIBOM data on a bad run. The conditional is there to avoid that data-loss mode.

cartography/models/aibom/workflow.py
```diff
@@ -0,0 +1,25 @@
...
+    function: PropertyRef = PropertyRef("function")
+    file_path: PropertyRef = PropertyRef("file_path")
+    line: PropertyRef = PropertyRef("line")
+    distance: PropertyRef = PropertyRef("distance")
+
+
+@dataclass(frozen=True)
+class AIBOMWorkflowSchema(CartographyNodeSchema):
+    label: str = "AIBOMWorkflow"
+    scoped_cleanup: bool = False
```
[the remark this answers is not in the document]
kunaals: I do not think this one is valid for the same reason as the component schema. `AIBOMWorkflow` has no tenant/sub-resource relationship to scope cleanup against, so unscoped cleanup is intentional here rather than accidental.

cartography/intel/aibom/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+            results_dir,
+        )
+        return
+
+    processed_reports = 0
+    for file_path in json_files:
+        try:
+            with open(file_path, encoding="utf-8") as file_pointer:
+                document = json.load(file_pointer)
+        except (OSError, json.JSONDecodeError) as exc:
```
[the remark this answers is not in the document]
kunaals: Addressed in `06a6b78bb`. Both local-file and S3 ingestion paths now catch `UnicodeDecodeError`, and I added integration coverage for each path in `tests/integration/cartography/intel/aibom/test_aibom.py`.

cartography/intel/aibom/parser.py
```diff
@@ -0,0 +1,243 @@
...
+    scan_scope: str | None
+    skip_reason: str | None
+    components: list[ParsedAIBOMComponent]
+    workflows: list[ParsedAIBOMWorkflow]
+
+
+def _as_dict(value: Any) -> dict[str, Any]:
+    if isinstance(value, dict):
+        return value
+    return {}
```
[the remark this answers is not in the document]
kunaals: Addressed in `06a6b78bb`. The parser now validates top-level wrappers, source payloads, and workflow lists explicitly instead of silently coercing malformed structures to empty dicts/lists, and I added parser coverage in `tests/unit/cartography/intel/aibom/test_parser.py`.

docs/root/modules/aibom/schema.md
```diff
@@ -0,0 +1,57 @@
...
+| workflow_id | Original workflow id from AIBOM output |
+| function | Workflow function name |
+| file_path | File path for the workflow |
+| line | Line number for the workflow |
+| distance | Workflow distance reported by AIBOM |
+
+### Linking constraints
+
+- AIBOM ingestion prefers `ECRImage` nodes with `type = "manifest_list"`.
+- If no manifest list exists for a tag, AIBOM ingestion falls back to the single linked `ECRImage` node with `type = "image"`.
```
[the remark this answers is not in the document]
kunaals: Addressed in 6af346e70. I updated the fallback wording to remove the uniqueness claim and match the current canonical-target selection behavior.

docs/root/modules/aibom/schema.md
```diff
@@ -0,0 +1,57 @@
...
+| line_number | Line number reported by the scanner |
+| **source_image_uri** | Source image URI associated with the report |
+| scanner_name | Scanner name |
+| scanner_version | Scanner version |
+| scan_scope | Scanner input scope |
+| **manifest_digest** | Digest of the canonical ECRImage used for graph linking |
+
+#### Relationships
+
+- An AIBOMComponent is detected in the canonical ECR image for the tag.
```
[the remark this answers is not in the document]
kunaals: Addressed in 6af346e70. I updated the relationship wording to describe canonical resolution from the source image URI, which covers both tag-qualified and digest-qualified inputs.

cartography/intel/aibom/loader.py
```diff
@@ -0,0 +1,236 @@
...
+) -> dict[str, list[dict[str, str]]]:
+    if not image_uris:
+        return {}
+
+    query = """
+    UNWIND $image_uris AS image_uri
+
+    // Try tag-based lookup: ECRRepositoryImage.id == image_uri
+    OPTIONAL MATCH (:ECRRepositoryImage {id: image_uri})-[:IMAGE]->(tag_img:ECRImage)
+    WHERE tag_img.type IN ['manifest_list', 'image']
```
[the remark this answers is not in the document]
kunaals: I’m keeping the current behavior here. The fallback to `type="image"` is intentional so single-platform images still get a canonical AIBOM attachment when no manifest list exists; manifest lists are still preferred when present.

cartography/intel/aibom/loader.py
```diff
@@ -0,0 +1,194 @@
...
+
+    For digest-based URIs (repo@sha256:...), extracts the digest directly.
+    For tag-based URIs (repo:tag), looks up via ECRRepositoryImage → ECRImage.
+
+    Returns the digest string or None if no match is found.
+    """
+    # Fast path: digest is in the URI itself
+    digest = _extract_digest(image_uri)
+    if digest:
+        return digest
```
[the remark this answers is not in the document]
kunaals: Addressed — `_resolve_digest_for_source` now verifies the digest exists as an `ECRImage` node before accepting it. Digest-qualified URIs that don't match a graph node are treated as unmatched.

cartography/intel/aibom/loader.py
```diff
@@ -0,0 +1,194 @@
...
+            )
+            stat_handler.incr("aibom_sources_unmatched")
+            continue
+
+        # Transform: resolve image URI to an ECRImage digest.
+        # For digest-based URIs (repo@sha256:...) this is pure string
+        # splitting. For tag-based URIs it falls back to a graph lookup.
+        # The data model (AIBOMComponentDetectedInRel) handles the actual
+        # relationship creation via ECRImage.digest == manifest_digest.
+        manifest_digest = _resolve_digest_for_source(
```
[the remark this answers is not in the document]
kunaals: Not addressing — the "N" here is sources per report (typically 1–3 per AIBOM document), not components. This is not a scaling concern in practice.

cartography/intel/aibom/parser.py
```diff
@@ -0,0 +1,230 @@
...
+    sources_obj = analysis_obj.get("sources")
+    if not isinstance(sources_obj, dict):
+        raise ValueError("AIBOM document has invalid sources format")
+
+    parsed_sources: list[ParsedAIBOMSource] = []
+
+    for source_key_raw, source_payload_obj in sources_obj.items():
+        source_key = str(source_key_raw)
+        if not isinstance(source_payload_obj, dict):
+            continue
```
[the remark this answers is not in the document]
kunaals: Intentional — a single malformed source entry within an otherwise valid document should not abort parsing of the remaining sources. Top-level structural issues (missing `aibom_analysis`, invalid `sources` format) still raise `ValueError`. This is consistent with how non-dict workflow and component objects are handled throughout the parser.

tests/unit/cartography/intel/aibom/test_parser.py
```diff
@@ -0,0 +1,30 @@
...
+    document: dict[str, Any] = {
+        "aibom_analysis": {
+            "sources": {
+                "000000000000.dkr.ecr.us-east-1.amazonaws.com/example:v1": [],
+            },
+        },
+    }
+
+    result = parse_aibom_document(document)
+    assert result == []
```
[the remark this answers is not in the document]
kunaals: The test reflects the intentional design change: a non-dict source payload is now skipped rather than aborting the entire document parse. See reply on `parser.py:181` for rationale.

cartography/models/aibom/component.py
```diff
@@ -0,0 +1,79 @@
...
+    name: PropertyRef = PropertyRef("name")
+    category: PropertyRef = PropertyRef("category", extra_index=True)
+    instance_id: PropertyRef = PropertyRef("instance_id")
+    assigned_target: PropertyRef = PropertyRef("assigned_target")
+    file_path: PropertyRef = PropertyRef("file_path")
+    line_number: PropertyRef = PropertyRef("line_number")
+    model_name: PropertyRef = PropertyRef("model_name")
+    framework: PropertyRef = PropertyRef("framework")
+    label: PropertyRef = PropertyRef("label")
+    metadata_json: PropertyRef = PropertyRef("metadata_json")
```
achantavy: Let's avoid storing the json. I'd rather expand them out to individual properties as needed to make them queryable.

I recognize this may be hard to generalize since this might be a way to store arbitrary key-value pairs and that can be useful, but I think for now let's forgo that in favor of being explicit.

docs/root/modules/aibom/schema.md
```diff
@@ -0,0 +1,190 @@
...
+
+Representation of one detected AI component occurrence within a source.
+
+| Field | Description |
+|-------|-------------|
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | Stable hash of source id + component identity fields |
+| name | Detected symbol name |
+| **category** | Category emitted by AIBOM (for example `agent`, `model`, `tool`, `memory`, `prompt`, `other`) |
```
achantavy, line 73: I think it might be worth promoting these to be their own labels for discoverability.

AIAgent, AIModel, AITool, etc. I don't think we necessarily need to extract it out to a full ontology node (I feel like that'd be overengineering).

Rather, I think multiple labels are simple enough for this. Example, we could have a multi label setup like `AIBOMComponent::AIAgent` if we wanted to encode the fact that this was pulled using the AIBOM library.

docs/root/modules/aibom/schema.md
```diff
@@ -0,0 +1,190 @@
...
+| firstseen | Timestamp of when a sync job first discovered this node |
+| lastupdated | Timestamp of the last time the node was updated |
+| **id** | Stable hash of source id + workflow id |
+| workflow_id | Original workflow id from AIBOM output |
+| function | Workflow function name |
+| file_path | File path for the workflow |
+| line | Line number for the workflow |
+| distance | Workflow distance reported by AIBOM |
+
+### AIBOMRelationship
```
achantavy: I get that this is a faithful representation of the AIBOM library, but I think here we should be opinionated since we do have an actual graph available.

E.g. `(:AIAgent)-[:USES_TOOL]->(:AITool)` instead of

```
(:AIAgent)<-[:FROM_COMPONENT]-(:AIBOMRelationship)-[:TO_COMPONENT]->(:AITool)
```

The latter feels too much

cartography/intel/aibom/loader.py
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    source_payloads_by_id: dict[str, dict[str, Any]] = {}
+    component_payloads_by_id: dict[str, dict[str, Any]] = {}
+    relationship_payloads_by_id: dict[str, dict[str, Any]] = {}
+    component_category_counts: Counter[str] = Counter()
+    relationship_type_counts: Counter[str] = Counter()
+    workflow_payloads_by_id: dict[str, dict[str, Any]] = {}
+
+    for source in document.sources:
```
achantavy, line 77: Nit on implementation detail: this data transformation belongs in a transform if possible since it makes the code cleaner, unless there's a performance reason.

cartography/intel/aibom/loader.py
```diff
@@ -0,0 +1,352 @@
...
+            source_workflow_ids.append(workflow_hash)
+
+        should_load_components = source_status == "completed" and bool(manifest_digest)
+        for component in source.components:
+            if not should_load_components:
+                continue
+
+            component_hash_input = "|".join(
+                [
+                    source_id,
```
achantavy: I think this dedupe doesn't work as intended.

`source_id` is unique per scan per image. So IIUC, if we scan the same image twice, we will store 2 different AIBOMComponent nodes. 

To keep the dedupe, we might want to use `name, category, file_path` as the key instead.

docs/root/modules/aibom/schema.md
achantavy: Then this would enable things like 

```cypher
MATCH (img:ECRImage)<-[:DETECTED_IN]-(agent:AIAgent)-[:USES_TOOL]->(tool:AITool)
RETURN img.digest, agent.name, tool.name
```

which feels hella cool

kunaals: done

docs/root/modules/aibom/schema.md
kunaals: done

cartography/models/aibom/component.py
kunaals: removed

cartography/intel/aibom/loader.py
kunaals: done

cartography/intel/aibom/parser.py
```diff
@@ -0,0 +1,442 @@
...
+        )
+
+    primary_prefix_values = (
+        _as_str(relationship.get(f"{primary_key}_instance_id"))
+        or _as_str(relationship.get(f"{primary_key}_id")),
+        _as_str(relationship.get(f"{primary_key}_name")),
+        _as_str(relationship.get(f"{primary_key}_category")),
+    )
+    if any(primary_prefix_values):
+        return primary_prefix_values
```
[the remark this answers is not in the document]
kunaals: addressed in d71d728ba. `_extract_rel_endpoint()` now merges partial `source_*` / `target_*` data with `from_*` / `to_*` fallbacks instead of returning early, and `tests/unit/cartography/intel/aibom/test_parser.py` covers the mixed-field case.

cartography/models/aibom/component.py
kunaals: agreed and addressed earlier. `metadata_json` was removed so the native component model only keeps explicit queryable fields.

cartography/intel/aibom/cleanup.py
```diff
@@ -29,3 +25,4 @@ def cleanup_aibom(
     GraphJob.from_node_schema(AIBOMWorkflowSchema(), common_job_parameters).run(
         neo4j_session,
     )
+    neo4j_session.run("MATCH (n:AIBOMScan) DETACH DELETE n")
```
achantavy: Should delete this line and leave it out since this PR was never merged to master branch.

cartography/models/aibom/__init__.py
```diff
@@ -1,13 +1,11 @@
 from cartography.models.aibom.component import AIBOMComponentSchema
 from cartography.models.aibom.relationship import AIBOMRelationshipSchema
-from cartography.models.aibom.scan import AIBOMScanSchema
 from cartography.models.aibom.source import AIBOMSourceSchema
 from cartography.models.aibom.workflow import AIBOMWorkflowSchema
 
 __all__ = [
     "AIBOMComponentSchema",
     "AIBOMRelationshipSchema",
```
achantavy: Let's remove this schema

cartography/models/aibom/relationship.py
```diff
@@ -62,10 +62,9 @@ class AIBOMRelationshipToComponentRel(CartographyRelSchema):
...
 
 @dataclass(frozen=True)
-# Native AIBOM relationships are modeled as nodes rather than direct graph edges.
-# The scanner emits relationship objects with their own payload and source-defined
-# types, so we preserve that source fidelity here and leave normalized edges to
-# ontology or analysis jobs.
+# Legacy schema kept so cleanup removes stale AIBOMRelationship nodes created by
+# earlier revisions of this module before direct component-to-component edges
+# became the primary graph representation.
 class AIBOMRelationshipSchema(CartographyNodeSchema):
```
achantavy: This is a zombie now since it's not used, let's remove

cartography/intel/aibom/transform.py
```diff
@@ -0,0 +1,295 @@
...
+                    relationship.target_instance_id or relationship.target_name,
+                )
+                continue
+
+            relationship_field = _RELATIONSHIP_TARGET_FIELD_BY_TYPE.get(
+                relationship.relationship_type,
+            )
+            if relationship_field is None:
+                logger.info(
+                    "Skipping unsupported AIBOM relationship type %s",
```
achantavy, line 244: Should this be 'warning'?

achantavy (changes requested):
Almost there! just some small things to delete

cartography/intel/aibom/cleanup.py
kunaals: addressed in dae6fae03a. the leftover `AIBOMScan` cleanup delete is gone; this branch never shipped that node type, so carrying special-case cleanup for it was unnecessary.

cartography/models/aibom/__init__.py
kunaals: addressed in https://github.com/cartography-cncf/cartography/commit/dae6fae03a3ad4e61efe794327da3b03bd017884. AIBOMRelationshipSchema is removed from the module exports now that direct USES_* edges are the only supported representation in this branch.

cartography/models/aibom/relationship.py
kunaals: addressed in https://github.com/cartography-cncf/cartography/commit/dae6fae03a3ad4e61efe794327da3b03bd017884. this legacy file is deleted; it was only hanging around for cleanup of stale nodes from a version of the branch that never merged.

cartography/intel/aibom/transform.py
kunaals: keeping this as info. unsupported custom AIBOM relationship types are expected because the source format is extensible, and warning on every unsupported type would create noisy scans for behavior we already document and count on AIBOMSource.total_relationships.

PR #2448, opened 2026-03-04
fix(identitycenter): scope ALLOWED_BY matchlinks by identity_store_id

jychp:
### Type of change

### Summary

The `ALLOWED_BY` MatchLink relationships between `AWSRole` and `AWSSSOUser`/`AWSSSOGroup` matched target nodes only by `id` (`UserId`/`GroupId`). When multiple Identity Center instances exist, this could cause cross-instance leaks where users from one instance get incorrectly linked to roles they shouldn't have access to.

This PR adds `identity_store_id` to the MatchLink `target_node_matcher` for both `AWSRoleToSSOUserMatchLink` and `AWSRoleToSSOGroupMatchLink`, and propagates `IdentityStoreId` through the data pipeline in `_sync_role_assignments`.

### How was this tested?

- Added `test_allowed_by_scoped_to_identity_store` integration test that creates users in two different identity stores and verifies only the correct user gets the `ALLOWED_BY` relationship
- All 14 existing identity center integration tests continue to pass

#### General

#### Proof of functionality

### Notes for reviewers

The generated MatchLink Cypher now includes `identity_store_id` in the MATCH clause:
```cypher
MATCH (to:AWSSSOUser{id: item.UserId, identity_store_id: item.IdentityStoreId})
```
This ensures the relationship is only created for users belonging to the correct Identity Center instance.


tests/integration/cartography/intel/aws/test_identitycenter.py
```diff
@@ -841,3 +842,169 @@ def test_multi_account_permission_set_assignments(
...
+):
+    """
+    Test that ALLOWED_BY relationships are scoped by identity_store_id,
+    preventing cross-instance leaks when multiple Identity Center instances exist.
+
+    Scenario:
+    - Two Identity Center instances (A and B) with different identity stores
+    - Both stores contain a user with the SAME UserId (simulating UUID collision)
+    - Only instance A has a permission set assignment for the user
+    - After full sync: instance A's ALLOWED_BY relationship is created correctly
```
kunaals: this docstring says we verify mismatched `identity_store_id` matchlinks are rejected, but we don't assert that anymore. can we either add a negative assertion for that case or update the docstring to match what we actually test?

tests/integration/cartography/intel/aws/test_identitycenter.py
```diff
@@ -841,3 +842,169 @@ def test_multi_account_permission_set_assignments(
...
+    # Assert 1: ALLOWED_BY was created during instance A's sync
+    allowed_by = neo4j_session.run(
+        """
+        MATCH (user:AWSSSOUser {id: $uid})<-[:ALLOWED_BY]-(role:AWSRole)
+        RETURN role.arn as arn
+        """,
+        uid=SHARED_USER_ID,
+    )
+    actual_arns = {r["arn"] for r in allowed_by}
+    assert (
```
kunaals, line 1018: this proves the happy path (`allowed_by` exists), but not the regression path. could we add an assertion that a wrong-store role assignment does *not* create or keep an `allowed_by` edge?

cartography/models/aws/identitycenter/awspermissionset.py
```diff
@@ -114,7 +114,10 @@ class AWSRoleToSSOUserMatchLink(CartographyRelSchema):
     # Standard CartographyRelSchema fields for AWSSSOUser as target
     target_node_label: str = "AWSSSOUser"
     target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
-        {"id": PropertyRef("UserId")},
+        {
+            "id": PropertyRef("UserId"),
+            "identity_store_id": PropertyRef("IdentityStoreId"),
```
kunaals, line 119: nice addition on the matcher. one concern: user/group node identity is still keyed only by `id`, so if the same principal id appears in multiple identity stores, later syncs can overwrite `identity_store_id` on the same node. should we model principal identity as composite (`identity_store_id` + principal id) to fully prevent cross-instance leakage?

kunaals (approved):
approving to unblock since the fix is directionally right and tests are green. i left a few follow-up inline notes around regression coverage and long-term principal identity modeling.

Issue #2101, opened 2025-11-28
[GitHub] GraphQL query to GitHub API regularly fails with 502 Bad Gateway error


pvasilevich:
### Expected behavior
Cartography should fetch data from GitHub in a more efficient way. 

### Actual behavior
Fetching data from GitHub is slow. Requests to GitHub API regularly failed with 502 error, causing additional requests, timeouts, sleeps.

### To Reproduce
Have GitHub organization with Go, or PHP applications (with go.mod and composer.lock files)

### Logs

```
INFO:cartography.intel.github.repos:Syncing GitHub repos
WARNING:cartography.intel.github.util:GitHub: Received 502 response. Reducing page size to 25 and retrying.
WARNING:cartography.intel.github.util:GitHub: Received 502 response. Reducing page size to 12 and retrying.
WARNING:cartography.intel.github.util:GitHub: Received 502 response. Reducing page size to 6 and retrying.
WARNING:cartography.intel.github.util:GitHub: Received 502 response. Reducing page size to 3 and retrying.
WARNING:cartography.intel.github.util:GitHub: Received 502 response. Reducing page size to 1 and retrying.
ERROR:cartography.intel.github.util:GitHub: Could not retrieve page of resource `repositories` due to HTTP error after 5 retries. Raising exception.
NoneType: None
ERROR:cartography.sync:Unhandled exception during sync stage 'github'
Traceback (most recent call last):
  File "/var/cartography/.local/share/uv/tools/cartography/lib/python3.10/site-packages/cartography/sync.py", line 152, in run
    stage_func(neo4j_session, config)
  File "/var/cartography/.local/share/uv/tools/cartography/lib/python3.10/site-packages/cartography/util.py", line 225, in timed
    return method(*args, **kwargs)
  File "/var/cartography/.local/share/uv/tools/cartography/lib/python3.10/site-packages/cartography/intel/github/__init__.py", line 70, in start_github_ingestion
    cartography.intel.github.repos.sync(
  File "/var/cartography/.local/share/uv/tools/cartography/lib/python3.10/site-packages/cartography/intel/github/repos.py", line 1234, in sync
    repos_json = get(github_api_key, github_url, organization)
  File "/var/cartography/.local/share/uv/tools/cartography/lib/python3.10/site-packages/cartography/util.py", line 225, in timed
    return method(*args, **kwargs)
[...]

PR #2450, opened 2026-03-04
fix(github): extract dependency graph into per-repo paginated queries

jychp:
### Type of change

### Summary

The GitHub repos GraphQL query regularly fails with 502 Bad Gateway because the `dependencyGraphManifests(first: 20)` block with nested `dependencies(first: 100)` is fetched inline for each repo in a page of 50. This causes GitHub API to timeout on orgs with repos containing large manifest files (Go, PHP, Node).

This PR extracts dependency graph fetching into separate per-repo API calls with two levels of pagination and resilient error handling:

1. **Per-repo isolation**: dependency manifests are fetched individually per repo instead of inline in the batch repos query. If one repo's manifests fail, others are unaffected.
2. **Per-manifest pagination**: manifests are fetched one at a time (`first: 1`). If a single manifest times out (e.g. large `go.mod`), previously fetched manifests are kept.
3. **Per-dependency pagination**: dependencies within each manifest are paginated (`first: 50`) instead of capped at 100. Repos with >100 deps per manifest now get all their dependencies.
4. **GraphQL timeout retry**: detects GraphQL-level timeouts (HTTP 200 with null data + errors) and retries up to 3 times with exponential backoff, in addition to HTTP error retries.
5. **Null resource handling**: `fetch_all` now handles null resources gracefully instead of crashing with `AttributeError`.

### Related issues or links

- Fixes GitHub GraphQL 502 errors on large organizations
- Closes: https://github.com/cartography-cncf/cartography/issues/2101

### How was this tested?

- All 12 unit tests pass
- All 7 integration tests pass (updated to mock the new `_get_dep_manifests_for_repos` function)
- All linters pass (`pre-commit run --all-files`)
- Tested on real environment: sync completes successfully, repos with heavy manifests are gracefully skipped after retries

#### General

#### Proof of functionality

### Notes for reviewers

The main change pattern follows the existing `_get_repo_collaborators` approach — per-repo GraphQL queries with retry logic. The custom `_fetch_manifest_page` / `_get_repo_dep_manifests` functions replace `fetch_all` for this use case to enable per-manifest and per-dependency-page error recovery that `fetch_all` doesn't support.

### DIFF
--- a/cartography/data/permission_relationships.yaml
+++ b/cartography/data/permission_relationships.yaml
@@ -27,3 +27,18 @@
   - redshift:GetClusterCredentials
   - redshift:JoinGroup
   relationship_name: CAN_ADMINISTER
+# Map principals that can query RDS instances via IAM database authentication.
+- target_label: RDSInstance
+  permissions:
+  - rds-db:connect
+  relationship_name: CAN_QUERY
+# Map principals that can administer RDS instances.
+- target_label: RDSInstance
+  permissions:
+  - rds:*
+  - rds:ModifyDBInstance
+  - rds:DeleteDBInstance
+  - rds:RebootDBInstance
+  - rds:StopDBInstance
+  - rds:StartDBInstance
+  relationship_name: CAN_ADMINISTER
--- a/cartography/rules/data/rules/__init__.py
+++ b/cartography/rules/data/rules/__init__.py
@@ -40,6 +40,46 @@ from cartography.rules.data.rules.cis_google_workspace import (
 from cartography.rules.data.rules.cis_google_workspace import (
     cis_gw_4_1_1_3_user_2sv_not_enforced,
 )
+from cartography.rules.data.rules.cis_kubernetes_rbac import (
+    cis_k8s_5_1_1_cluster_admin_usage,
+)
+from cartography.rules.data.rules.cis_kubernetes_rbac import cis_k8s_5_1_2_secret_access
+from cartography.rules.data.rules.cis_kubernetes_rbac import (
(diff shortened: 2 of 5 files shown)


cartography/intel/github/repos.py
```diff
@@ -198,6 +189,290 @@
...
+    while has_next_manifest:
+        # Save cursor before this manifest so we can re-query the same position
+        # when paginating its dependencies.
+        prev_manifest_cursor = manifest_cursor
+
+        resp = _fetch_manifest_page(
+            token,
+            api_url,
+            organization,
+            repo,
```
kunaals, line 325: if github returns {"data": null, "errors": [...]}, this check passes because "data" exists, then we crash on resp["data"]["organization"] below. can we treat null data as a failed page and keep the manifests collected so far?

cartography/intel/github/repos.py
```diff
@@ -198,6 +189,290 @@
...
+            )
+        except (
+            requests.exceptions.Timeout,
+            requests.exceptions.HTTPError,
+            requests.exceptions.ConnectionError,
+            requests.exceptions.ChunkedEncodingError,
+        ):
+            if attempt + 1 >= retries:
+                return None
+            time.sleep(2 ** (attempt + 1))
```
kunaals, line 266: small edge case: organization can be present but null, so this chained .get(...).get(...) can raise attributeerror before retry handling kicks in. could we normalize organization/repository to {} when they are null?

kunaals (approved):
approving to unblock. left two small inline follow-ups around null-graph response handling to harden the retry path.

Issue #1888, opened 2025-09-08
AWS IAM get_account_summary


achantavy:
## Description
> Describe your idea.  Please be detailed.  If a feature request, please
> describe the desired behavior, what scenario it enables, and how it
> would be used.

We should audit AWS IAM `get_account_summary` so that we monitor whether the root account has MFA enabled. 

## Motivation
> Why is this feature needed? What problem does it solve or opportunity does it unlock?

This is an important compliance check.

## Alternatives Considered
> List other approaches or ideas considered, and why they were not chosen.

Could also do this with `generate_credential_report` + `get_credential_report`, but this way requires polling.

## Relevant Links
> Any extra documentation required to understand the issue.

krishi-agrawal:
Hi, I would like to work on this issue!

achantavy:
@krishi-agrawal - no need to ask for permission, just do it haha. 

The one tricky part is we need to represent the root user of the account as its own node and then we need to add an attribute to it to indicate whether it has MFA enabled

PR #2451, opened 2026-03-04
feat(aws/iam): add get_account_summary ingestion

jychp:
### Type of change
- [ ] Documentation update

### Summary

Ingests AWS IAM `GetAccountSummary` data by enriching existing `AWSAccount` nodes with account-level IAM summary properties. This enables compliance checks like root account MFA status monitoring.

Properties added to `AWSAccount` nodes: `account_mfa_enabled`, `mfa_devices`, `mfa_devices_in_use`, `account_access_keys_present`, `account_signing_certificates_present`, `users`, `groups`, `roles`, `policies`, `instance_profiles`, `providers`, `server_certificates`, `policy_versions_in_use`.

Design choice: rather than creating a separate `AWSAccountSummary` node, we enrich the existing `AWSAccount` node directly since there is a 1:1 relationship. This follows the same pattern as `AWSPrincipalServiceAccessSchema` which enriches existing `AWSPrincipal` nodes.

### Related issues or links

- Closes https://github.com/cartography-cncf/cartography/issues/1888

### How was this tested?

- New integration test `test_iam_account_summary.py` verifying properties are merged onto `AWSAccount` nodes
- Full IAM integration test suite passes (18/18)
- Lint passes (`make test_lint`)

#### General

#### Proof of functionality

### Notes for reviewers

No cleanup job is needed for this schema since it only merges properties onto existing `AWSAccount` nodes (whose lifecycle is managed by the account sync). No new node type or relationship is created.

### DIFF
--- a/cartography/data/permission_relationships.yaml
+++ b/cartography/data/permission_relationships.yaml
@@ -27,3 +27,18 @@
   - redshift:GetClusterCredentials
   - redshift:JoinGroup
   relationship_name: CAN_ADMINISTER
+# Map principals that can query RDS instances via IAM database authentication.
+- target_label: RDSInstance
+  permissions:
+  - rds-db:connect
+  relationship_name: CAN_QUERY
+# Map principals that can administer RDS instances.
+- target_label: RDSInstance
+  permissions:
+  - rds:*
+  - rds:ModifyDBInstance
+  - rds:DeleteDBInstance
+  - rds:RebootDBInstance
+  - rds:StopDBInstance
+  - rds:StartDBInstance
+  relationship_name: CAN_ADMINISTER
--- a/cartography/intel/github/repos.py
+++ b/cartography/intel/github/repos.py
@@ -1,5 +1,6 @@
 import configparser
 import logging
+import time
 from collections import defaultdict
 from collections import namedtuple
 from string import Template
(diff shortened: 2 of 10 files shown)


jychp:
tested in production:
Screenshot 2026-03-03 at 5 51 53 PM

cartography/intel/aws/iam.py
```diff
@@ -1305,6 +1306,55 @@ def load_server_certificates(
     )
 
 
+@timeit
+def get_account_summary(boto3_session: boto3.Session) -> Dict[str, int]:
+    client = boto3_session.client("iam")
+    response = client.get_account_summary()
```
kunaals, line 1312: can we add a test that locks in this behavior when summarymap is missing? this now raises (which seems intentional), but we don’t have coverage for that path yet.

tests/integration/cartography/intel/aws/iam/test_iam_account_summary.py
```diff
@@ -0,0 +1,38 @@
...
+
+import cartography.intel.aws.iam
+import tests.data.aws.iam.account_summary
+from tests.integration.util import check_nodes
+
+TEST_ACCOUNT_ID = "123456789012"
+TEST_UPDATE_TAG = 123456789
+
+
+@patch("cartography.intel.aws.iam.get_account_summary")
```
kunaals, line 13: nice happy-path test. can we add a failure-path case too (e.g. missing summarymap) and assert the expected exception so this doesn’t regress later?

kunaals (approved):
approving to unblock since this looks good overall. i left 2 small follow-ups inline for coverage on the missing summarymap path.

PR #2452, opened 2026-03-04
feat(aws): add support for AWS Beanstalk Environments

kirkj-lightspeed:
### Type of change

- [x] Documentation update

### Summary

Some AWS DNS records point to Elastic Beanstalk environments, and that made me realise there's no link from the environment's CNAME to the ALB or EC2 instance etc that the traffic is routed to.

This PR attempts to add support for Elastic Beanstalk Environments, and tries to connect the relationship between the environment and the resources it has created (EC2 instances, load balancers, etc), and then links the DNS Record to the environment.

This PR also has a potential BREAKING CHANGE in the re-ordering of the resources in cartography/intel/aws/resources.py. Please review and provide guidance on whether this is safe to do.

### Breaking changes

This PR also has a potential BREAKING CHANGE in the re-ordering of the resources in cartography/intel/aws/resources.py. The Route53 model is moved last, so that it can create the relationships between DNS Record and the Elastic Beanstalk Environment. 

The reason for this change is: the Elastic Beanstalk Environment node needs to be created after all the generated resources (EC2 Instance, Load Balancers etc), so that it can create relationships to each of these nodes. However, DNS Record needs to happen after the Elastic Beanstalk Environment because it needs to point to the environment.

Please review and provide guidance on whether this is safe to do. 

### How was this tested?

Simple unit test to check resource tree is flattened into lists of IDs.

Manual testing against an account of mine that has Elastic Beanstalk in use.

#### General

#### Proof of functionality

image

- [x] New or updated unit/integration tests. Only unit tests.

  - Example:

```
INFO:cartography.intel.aws.elasticbeanstalk:Syncing ElasticBeanstalk environments for region 'ap-northeast-1' in account '<SNIP>'.
INFO:cartography.intel.aws.elasticbeanstalk:Loading 29 ElasticBeanstalk environments for region 'ap-northeast-1' into graph.
INFO:cartography.client.core.tx:Loaded 29 ElasticBeanstalkEnvironment nodes
DEBUG:cartography.graph.job:Starting job 'Cleanup ElasticBeanstalkEnvironment'.
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #1
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #2
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #3
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #4
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #5
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #6
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #7
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #8
INFO:cartography.graph.statement:Completed ElasticBeanstalkEnvironment statement #9
INFO:cartography.graph.job:Finished job ElasticBeanstalkEnvironment
```

- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md). Not sure where this one is?

### Notes for reviewers

This is my first contribution, so I'm interested in feedback! Style, relationship naming, completeness etc are all welcome. Cheers!


cartography/models/aws/elasticbeanstalk/environment.py
jychp, line 1: Missing `__init__.py` in `cartography/models/aws/elasticbeanstalk/`. Every other model subdirectory (34 of them) has one. Without it, package imports may fail depending on Python package resolution.

cartography/models/aws/elasticbeanstalk/environment.py
```diff
@@ -0,0 +1,232 @@
...
+class ElasticBeanstalkToAWSLoadBalancerV2RelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class ElasticBeanstalkToAWSLoadBalancerV2Rel(CartographyRelSchema):
+    target_node_label: str = "AWSLoadBalancerV2"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        # This LOADBALANCER_NAMES field is set by transform_elasticbeanstalk_environments
+        {"arn": PropertyRef("LOADBALANCER_NAMES", one_to_many=True)},
```
jychp, line 179: `LOADBALANCER_NAMES` is populated from `res["Name"]` in `describe_environment_resources`. For `AWSLoadBalancerV2`, matching on `arn` only works if AWS returns the ARN in the `Name` field for ALB/NLB environments (which it does in some cases). 

Was this tested with ALB/NLB-backed environments? If so, please add a comment here explaining that for v2 load balancers, `describe_environment_resources` returns the ARN in the `Name` field — this is non-obvious and both automated reviewers flagged it as a bug.

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,172 @@
...
+from cartography.util import merge_module_sync_metadata
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+stat_handler = get_stats_client(__name__)
+
+
+@timeit
+@aws_handle_regions
+def get_elasticbeanstak_environments(
```
jychp: Typo: `get_elasticbeanstak_environments` → `get_elasticbeanstalk_environments` (missing 'l').

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,172 @@
...
+def load_environments(
+    neo4j_session: neo4j.Session,
+    data: list[dict],
+    region: str,
+    current_aws_account_id: str,
+    update_tag: int,
+) -> None:
+
+    logger.info(
+        f"Loading {len(data)} ElasticBeanstalk environments for region '{region}' into graph."
```
jychp: f-string in `logger.info` evaluates eagerly even when filtered by log level. Use lazy `%` formatting like the other log calls in this file:
```python
logger.info(
    "Loading %d ElasticBeanstalk environments for region '%s' into graph.",
    len(data),
    region,
)
```

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,172 @@
...
+            )
+            environment["Resources"] = resources.get("EnvironmentResources")
+            environments.append(environment)
+
+    return environments
+
+
+def transform_elasticbeanstalk_environments(
+    environments: list[dict[str, Any]], region: str
+) -> List[Dict[str, Any]]:
```
jychp: Mixed type hint styles: the return type uses `List[Dict[str, Any]]` from `typing` while the rest of the file uses builtin `list[dict[...]]`. Use `list[dict[str, Any]]` and remove the unused `from typing import Dict, List` imports on lines 3-4.

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,172 @@
...
+            "Status": environment.get("Status"),
+            "AbortableOperationInProgress": environment.get(
+                "AbortableOperationInProgress"
+            ),
+            "Health": environment.get("Health"),
+            "HealthStatus": environment.get("HealthStatus"),
+            "Region": region,
+        }
+
+        if environment.get("Resources"):
```
jychp, line 82: When `environment.get("Resources")` is falsy, the relationship fields (`ASG_NAMES`, `INSTANCE_IDS`, `LOADBALANCER_NAMES`, etc.) are never set on the record dict. Consider initializing them as empty lists before the `if` block so the data shape is consistent regardless of whether resources exist.

tests/unit/cartography/intel/aws/test_elasticbeanstalk_transform.py
jychp, line 1: The project standard for new intel modules includes integration tests using `check_nodes`/`check_rels` (see `tests/integration/cartography/intel/`) to verify nodes and relationships are correctly loaded into Neo4j. Unit tests for the transform are a good addition but not a substitute for integration tests.

cartography/intel/aws/resources.py
```diff
@@ -149,5 +149,10 @@
         "cognito": cognito.sync,
         "eventbridge": eventbridge.sync,
         "glue": glue.sync,
+        # ElasticBeanstalk must run after ec2:instance, ec2:autoscalinggroup, ec2:launch_templates,
+        # ec2:load_balancer, ec2:load_balancer_v2 and sqs, so that the nodes exist when elasticbeanstalk
+        # makes relationships, but before route53 as it has a relationship to this.
+        "elasticbeanstalk": elasticbeanstalk.sync,
+        "route53": route53.sync,
```
jychp, line 160: The PR description flags this reordering as a potential breaking change. Is there any existing sync module that depends on `route53` running before it? The dependency graph should be verified — moving `route53` to the very end could affect users with custom sync configurations that assumed the previous ordering.

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,172 @@
...
+            resources = environment["Resources"]
+
+            # These fields are set so that relationships can match on them (e.g. ElasticBeanstalkToInstanceRel)
+            environment_record["ASG_NAMES"] = [
+                str(res["Name"]) for res in resources.get("AutoScalingGroups", [])
```
jychp: Why do you enforce str(), if a value is None, this will cast the value to "None" instead of an empty property

jychp (changes requested):
Thanks for your contributions, some changes are requested.

cartography/models/aws/elasticbeanstalk/environment.py
kirkj-lightspeed: Ooops, sorry about that! Added now.

cartography/intel/aws/elasticbeanstalk.py
kirkj-lightspeed: Ooops! Good catch 👍

cartography/intel/aws/elasticbeanstalk.py
kirkj-lightspeed: Wow, that's cool. Thanks for pointing that out!

cartography/intel/aws/elasticbeanstalk.py
kirkj-lightspeed: Dang! Thanks for catching that... I'm a python noob!

cartography/intel/aws/elasticbeanstalk.py
kirkj-lightspeed: Thanks, will do.

cartography/intel/aws/resources.py
kirkj-lightspeed: Thanks @jychp! Is there any dependency graph that I could refer to, or is it something I should manually verify?

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,181 @@
...
+            ]
+            environment_record["LOADBALANCER_NAMES"] = [
+                str(res["Name"]) for res in resources.get("LoadBalancers", [])
+            ]
+            environment_record["QUEUE_URLS"] = [
+                str(res["URL"]) for res in resources.get("Queues", [])
+            ]
+            environment_record["TRIGGER_NAMES"] = [
+                str(res["Name"]) for res in resources.get("Triggers", [])
+            ]
```
jychp, line 106: blocking: `str()` wrapping will convert `None` to the string `"None"`, which could create incorrect relationship matches. If `res["Name"]` or `res["Id"]` is ever `None`, it will be stored as `"None"` and attempt to MERGE against nodes with that literal value.

Remove the `str()` calls — the values from the AWS API are already strings when present.

cartography/models/aws/elasticbeanstalk/environment.py
jychp, line 179: blocking: This matches `AWSLoadBalancerV2.arn` against `LOADBALANCER_NAMES`, which is populated from `res["Name"]` in `describe_environment_resources`. This only works because for ALB/NLB-backed environments, AWS returns the ARN in the `Name` field — a non-obvious API behavior.

Two things needed:
1. Confirm this was tested with ALB/NLB-backed environments (not just classic ELBs).
2. Add a code comment here explaining the AWS API quirk, e.g.:
```python
# For v2 load balancers (ALB/NLB), describe_environment_resources returns
# the load balancer ARN in the Name field, so we match on arn here.
```

Both automated reviewers (codex, cubic) flagged this as a potential bug. It's correct if the above holds, but needs explicit documentation.

tests/unit/cartography/intel/aws/test_elasticbeanstalk_transform.py
jychp, line 1: blocking: The project standard for new intel modules requires integration tests using `check_nodes`/`check_rels` to verify that nodes and relationships are correctly loaded into Neo4j. See `tests/integration/cartography/intel/aws/` for examples.

The unit test for the transform logic is a good addition, but it doesn't validate that the schema produces the correct graph structure — which is where relationship matching bugs (like the LBv2 arn issue) would surface.

cartography/models/aws/elasticbeanstalk/environment.py
```diff
@@ -0,0 +1,232 @@
...
+class ElasticBeanstalkEnvironmentNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("EnvironmentId")
+    arn: PropertyRef = PropertyRef("EnvironmentArn", extra_index=True)
+    environment_name: PropertyRef = PropertyRef("EnvironmentName")
+    application_name: PropertyRef = PropertyRef("ApplicationName")
+    version_label: PropertyRef = PropertyRef("VersionLabel")
+    solution_stack_name: PropertyRef = PropertyRef("SolutionStackName")
+    platform_arn: PropertyRef = PropertyRef("PlatformArn")
+    endpoint_url: PropertyRef = PropertyRef("EndpointURL")
+    cname: PropertyRef = PropertyRef("CNAME")
```
jychp: suggestion: `cname` is used as a target matcher property by `AWSDNSRecordToElasticBeanstalkEnvironmentRel` in the Route53 model. Without `extra_index=True`, the DNS relationship join will require a full label scan instead of an index lookup.

```suggestion
    cname: PropertyRef = PropertyRef("CNAME", extra_index=True)
```

cartography/intel/aws/resources.py
jychp: I checked module_dependencies and the full RESOURCE_FUNCTIONS ordering. No module lists route53 as a dependency, and no module after route53's original position creates relationships TO DNS records. The reordering appears safe.

jychp (changes requested):

cartography/models/aws/elasticbeanstalk/environment.py
kirkj-lightspeed: Thanks, I've added the index in the lasted commit.

cartography/models/aws/elasticbeanstalk/environment.py
kirkj-lightspeed: I have only tested this on an environment that uses ALB's. I'm trying to figure out how to create a test environment with some more variations (v1 ELB, v2 NLB).

cartography/intel/aws/resources.py
kirkj-lightspeed: Thanks @jychp , that's great to know!

cartography/intel/aws/elasticbeanstalk.py
kirkj-lightspeed: Thanks, I've removed those.

cartography/intel/aws/elasticbeanstalk.py
kirkj-lightspeed: Thanks, I've removed those str calls.

cartography/models/aws/elasticbeanstalk/environment.py
```diff
@@ -0,0 +1,232 @@
...
+class ElasticBeanstalkToAWSLoadBalancerV2RelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class ElasticBeanstalkToAWSLoadBalancerV2Rel(CartographyRelSchema):
+    target_node_label: str = "AWSLoadBalancerV2"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        # This LOADBALANCER_NAMES field is set by transform_elasticbeanstalk_environments
+        {"arn": PropertyRef("LOADBALANCER_NAMES", one_to_many=True)},
```
jychp, line 179: blocking: Still unresolved. `AWSLoadBalancerV2.arn` is matched against `LOADBALANCER_NAMES`, which is populated from `res["Name"]`. This only works because `describe_environment_resources` returns the ARN in the `Name` field for v2 LBs — a non-obvious AWS quirk. Please either (a) add an explicit code comment documenting this and confirm you have verified it with an NLB-backed environment, or (b) transform the value into a clearly-named field (e.g. `LOADBALANCER_V2_ARNS`) so that the matcher reads correctly.

tests/unit/cartography/intel/aws/test_elasticbeanstalk_transform.py
```diff
@@ -0,0 +1,67 @@
+from cartography.intel.aws.elasticbeanstalk import (
```
jychp, line 1: blocking: Still unresolved. The project standard requires integration tests under `tests/integration/cartography/intel/aws/` using `check_nodes` and `check_rels` to verify nodes and all 8 relationships are correctly loaded into Neo4j. The current unit test only validates the transform dict shape, so it would not catch the LBv2 matcher bug or any broken relationship.

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,181 @@
...
+    boto3_session: boto3.session.Session,
+    region: str,
+) -> list[dict[str, Any]]:
+    client = boto3_session.client("elasticbeanstalk", region_name=region)
+    paginator = client.get_paginator("describe_environments")
+    environments: list[dict[str, Any]] = []
+    for page in paginator.paginate():
+        # I'm not sure why describe_environments doesn't return resources as well,
+        # the docs suggest it should. Instead, do a call for each environment
+        # and add the resources to them
```
jychp, line 33: suggestion: Replace the informal comment with a neutral one-liner, e.g. `# describe_environments does not return resource details, so fetch them per-environment.`

cartography/models/aws/elasticbeanstalk/environment.py
```diff
@@ -0,0 +1,232 @@
...
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class ElasticBeanstalkEnvironmentNodeProperties(CartographyNodeProperties):
+    id: PropertyRef = PropertyRef("EnvironmentId")
```
jychp, line 16: question: Using `EnvironmentId` as the node `id` assumes Elastic Beanstalk `e-xxxxxxxxxx` IDs are globally unique across AWS accounts. If two synced accounts ever happen to hold the same `EnvironmentId`, they will collide on MERGE. Using `EnvironmentArn` (which is account+region+id-scoped) as the node id is safer. Can you confirm uniqueness, or switch to ARN?

jychp (changes requested):

cartography/models/aws/elasticbeanstalk/environment.py
kirkj-lightspeed: Hi @jychp , sorry for taking a while to address this. I did set up a test environment with v1 and v2 loadbalancers, and capture some data that I've started incorporating into integration tests. I'll try and get that done soon!

cartography/models/aws/elasticbeanstalk/environment.py
```diff
@@ -0,0 +1,232 @@
...
+class ElasticBeanstalkToAutoScalingGroupRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class ElasticBeanstalkToAutoScalingGroupRel(CartographyRelSchema):
+    target_node_label: str = "AutoScalingGroup"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        # This ASG_NAMES field is set by transform_elasticbeanstalk_environments
+        {"name": PropertyRef("ASG_NAMES", one_to_many=True)},
```
jychp, line 91: blocking: `one_to_many` matchers in `other_relationships` match globally across every node of the target label, not within the synced account/region. `AutoScalingGroupName` is only unique per account/region, so when more than one account is synced this creates `HAS_AUTO_SCALING_GROUP` edges to same-named ASGs in other accounts. Same defect for `HAS_LAUNCH_CONFIG` (line 113, matches `name`) and the classic `HAS_LOAD_BALANCER` (line 157, matches `name`). Match on a globally unique key (e.g. the ARN) or scope the match to the account.

cartography/models/aws/elasticbeanstalk/environment.py
```diff
@@ -0,0 +1,232 @@
...
+class ElasticBeanstalkToQueueRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class ElasticBeanstalkToQueueRel(CartographyRelSchema):
+    target_node_label: str = "SQSQueue"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        # This QUEUE_URLS field is set by transform_elasticbeanstalk_environments
+        {"url": PropertyRef("QUEUE_URLS", one_to_many=True)},
```
jychp, line 201: nitpick: `SQSQueue.url` has no `extra_index`, so this `one_to_many` match runs as a full label scan over all SQSQueue nodes. The queue ARN (`SQSQueue.arn`, indexed) is globally unique and indexed; prefer matching on it if `describe_environment_resources` exposes the ARN, otherwise note the scan cost.

cartography/intel/aws/elasticbeanstalk.py
```diff
@@ -0,0 +1,181 @@
...
+            "PlatformArn": environment.get("PlatformArn"),
+            "EndpointURL": environment.get("EndpointURL"),
+            "CNAME": environment.get("CNAME"),
+            "Status": environment.get("Status"),
+            "AbortableOperationInProgress": environment.get(
+                "AbortableOperationInProgress"
+            ),
+            "Health": environment.get("Health"),
+            "HealthStatus": environment.get("HealthStatus"),
+            "Region": region,
```
jychp, line 70: nitpick: `"Region": region` on the record is dead data. The schema's `region` property uses `set_in_kwargs=True` and reads the `Region=region` kwarg passed to `load()`, so this per-record key is never consumed. Remove it.

cartography/models/aws/route53/dnsrecord.py
```diff
@@ -213,6 +213,26 @@ class AWSDNSRecordToElasticIPAddressRel(CartographyRelSchema):
...
+class AWSDNSRecordToElasticBeanstalkEnvironmentRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class AWSDNSRecordToElasticBeanstalkEnvironmentRel(CartographyRelSchema):
+    target_node_label: str = "ElasticBeanstalkEnvironment"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {
+            "cname": PropertyRef("value"),
```
jychp, line 226: question: DNS record `value` for a CNAME may carry a trailing dot, while the Beanstalk `CNAME` field does not. Confirm the join works for plain CNAME records (not just alias records, whose value is dot-stripped at route53.py:113), or normalize before matching.

jychp (changes requested):

PR #2462, opened 2026-03-05
feat(gcp): Add Selective Sync for GCP

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

 - Added --gcp-requested-syncs CLI flag accepting a comma-separated list of GCP resources to sync (e.g.   "compute,iam,storage")                                                        
  - When specified, only requested resources sync — others are skipped even if enabled on the project                                                                                  
  - When not specified, all enabled resources sync (existing behavior unchanged)                                                                                                       
  - Invalid resource names in the flag raise a validation error with the full list of valid values
  - Logs a warning when a requested module's dependencies aren't also requested (e.g. policy_bindings without iam)                                                                     
  - Analysis jobs are skipped when their dependent module wasn't requested

### How was this tested?

- Ran mix of live GCP syncs

### DIFF
--- a/tests/unit/cartography/intel/gcp/test_selective_sync.py
+++ b/tests/unit/cartography/intel/gcp/test_selective_sync.py
@@ -28,6 +28,7 @@ def test_selective_sync_warns_permission_relationships_missing_deps(
         cartography.intel.gcp.start_gcp_ingestion(
             MagicMock(),
             config,
+            credentials=MagicMock(),
         )
 
     warning_records = [


cartography/intel/gcp/__init__.py
```diff
@@ -672,6 +691,34 @@ def start_gcp_ingestion(
...
+            parse_and_validate_gcp_requested_syncs(config.gcp_requested_syncs),
+        )
+        logger.info(
+            "GCP selective sync enabled for: %s", ", ".join(sorted(requested_syncs))
+        )
+
+        # Warn if modules are requested without their dependencies
+        module_dependencies = {
+            "policy_bindings": ["iam"],
+            "permission_relationships": ["iam"],
```
kunaals: nice addition overall. one non-blocking follow-up: `permission_relationships` appears to depend on both `iam` and `policy_bindings`, but the selective-sync dependency handling here only models `iam`. on a fresh graph, a run like `--gcp-requested-syncs iam,permission_relationships` can end up producing no permission relationships because `permission_relationships.sync()` reads `gcppolicybinding` nodes from neo4j rather than fetching that data itself.

prompt for agent:
verify the selective-sync dependency chain for gcp `permission_relationships` end to end. confirm which upstream data it consumes at runtime, then update the selective-sync behavior so this mode cannot silently produce incomplete results on a fresh graph. either enforce or normalize the required dependencies, or skip with a clear warning when prerequisites are missing. add a targeted test that covers a fresh-graph selective-sync run where `permission_relationships` is requested without all required prerequisites.

PR #2471, opened 2026-03-06
feat(cli,config): add neo4j liveness check timeout option to prevent connection issues

### LINKED ISSUE #1978: Support configuring `max_connection_idle_time` in Neo4j driver
# Support configuring `max_connection_idle_time` in Neo4j driver

## Description
Currently, Cartography allows configuration of the Neo4j driver via parameters such as `max_connection_lifetime` and `connection_timeout`, but does not expose `max_connection_idle_time`. This setting was introduced in newer versions of the Neo4j Python driver to proactively close idle Bolt connections before the server does.

Supporting `max_connection_idle_time` in Cartography’s Neo4j driver initialization would let deployments avoid frequent `SessionExpired` or `ConnectionResetError` issues that occur when Aura or clustered Neo4j instances close idle connections.

The desired behavior is to allow users to configure this via the standard `--neo4j-max-connection-idle-time` CLI argument or environment variable, which then gets passed through to the `GraphDatabase.driver()` call.

Example:
```python
driver = GraphDatabase.driver(
    uri,
    auth=auth,
    max_connection_lifetime=config.neo4j_max_connection_lifetime,
    max_connection_idle_time=config.neo4j_max_connection_idle_time,
)
```

## Motivation
This feature helps prevent connection resets and `SessionExpired` errors that occur after several minutes of inactivity between stages in long-running sync jobs (e.g. during large GitHub ingestions).  
Neo4j Aura and other managed clusters frequently close idle connections after ~3–5 minutes. Without this setting, the driver keeps stale sockets in its pool, which only fail on the next use, causing sync interruptions.

Allowing `max_connection_idle_time` aligns with Neo4j best practices and improves stability for large or multi-tenant syncs that may have uneven workloads across stages.

## Alternatives Considered
- **Manual retry logic:** Works but only recovers after failure
- **Refactoring to allow the Driver to be accessed throughout syncs instead of Session**: very big and invasive refactor
- **Reducing `max_connection_lifetime`:** Helps somewhat but does not address idle sockets specifically.

Adding `max_connection_idle_time` is the cleanest, officially supported solution.

## Relevant Links
- Neo4j Python driver docs: https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.GraphDatabase.driver
- Neo4j Aura recommendations on idle connections
- Related Cartography issue: #1957

Denyme24:
### Type of change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

### Summary

adds support for configuring the `liveness_check_timeout` parameter in the Neo4j driver. this controls how long a connection can remain idle before the driver performs a liveness check (`RESET` ping) before reusing it.

this helps prevent issues like `SessionExpired` and `ConnectionResetError`, especially on Aura and clustered Neo4j deployments where idle connections may be closed server side.

the parameter can be configured through:

• `--neo4j-liveness-check-timeout` CLI flag
• `neo4j_liveness_check_timeout` configuration attribute

if not specified, the Neo4j driver default behavior is used, meaning no liveness check is performed.

---

### Related issues or links

- Fixes #1978 

### How was this tested?

added an integration test `test_cli_neo4j_liveness_check_timeout` in `tests/integration/cartography/test_cli.py` to verify that the `--neo4j-liveness-check-timeout` flag is parsed correctly and the value is properly passed to the `Config` object.

also ran the full CLI integration test suite against a local Neo4j 5.15 instance to confirm everything works as expected.

#### General
- [X] I have read the [contributing guidelines](https://cartography-cncf.github.io/cartography/dev/developer-guide.html).
- [X] The linter passes locally (`make lint`).
- [X] I have added/updated tests that prove my fix is effective or my feature works.

#### Proof of functionality

- [X] New or updated unit/integration tests.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -596,6 +596,18 @@ class CLI:
             # =================================================================
             # GCP Options
             # =================================================================
+            gcp_requested_syncs: Annotated[
+                str | None,
+                typer.Option(
+                    "--gcp-requested-syncs",
+                    help=(
+                        "Comma-separated list of GCP resources to sync. "
+                        'Example: "compute,iam,storage". See cartography.intel.gcp.resources for full list.'
+                    ),
+                    rich_help_panel=PANEL_GCP,
+                    hidden=PANEL_GCP not in visible_panels,
+                ),
+            ] = None,
             gcp_permission_relationships_file: Annotated[
                 str,
                 typer.Option(
@@ -1605,6 +1617,14 @@ class CLI:
 
                 parse_and_validate_aws_regions(aws_regions)
 
+            # Validate GCP options
+            if gcp_requested_syncs:
+                from cartography.intel.gcp.util import (
+                    parse_and_validate_gcp_requested_syncs,
+                )
+
+                parse_and_validate_gcp_requested_syncs(gcp_requested_syncs)
+
             # Read Azure client secret
             azure_client_secret = None
             if azure_sp_auth and azure_client_secret_env_var:
@@ -2018,6 +2038,7 @@ class CLI:
                 digitalocean_token=digitalocean_token,
                 permission_relationships_file=permission_relationships_file,
                 azure_permission_relationships_file=azure_permission_relationships_file,
+                gcp_requested_syncs=gcp_requested_syncs,
                 gcp_permission_relationships_file=gcp_permission_relationships_file,
                 jamf_base_uri=jamf_base_uri,
                 jamf_user=jamf_user,
--- a/cartography/config.py
+++ b/cartography/config.py
@@ -89,6 +89,8 @@ class Config:
     :param permission_relationships_file: File path for the resource permission relationships file. Optional.
     :type azure_permission_relationships_file: str
     :param azure_permission_relationships_file: File path for the Azure permission relationships file. Optional.
+    :type gcp_requested_syncs: str
+    :param gcp_requested_syncs: Comma-separated list of GCP resources to sync. Optional.
     :type gcp_permission_relationships_file: str
     :param gcp_permission_relationships_file: File path for the GCP resource permission relationships file. Optional.
     :type jamf_base_uri: string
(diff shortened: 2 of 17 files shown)


Denyme24:
hey @jychp @kunaals , Could you please review this PR when you have a chance? Thanks!

achantavy:
Hi @Denyme24, thanks for the PR, I'm just curious but are you able to share an example of how this change helped you? Like, did you experience a crash in prod and then you configured this param and things were fixed?

Asking because Neo4j docs say:

> You normally should not need to tune this parameter. No connection liveliness check is done by default ([None](https://docs.python.org/3/library/constants.html#None)). A value of 0 means connections will always be tested for validity. Negative values are not allowed.

https://neo4j.com/docs/api/python-driver/current/api.html#liveness-check-timeout

Appreciate the context!

Denyme24:
> Hi @Denyme24, thanks for the PR, I'm just curious but are you able to share an example of how this change helped you? Like, did you experience a crash in prod and then you configured this param and things were fixed?
> 
> Asking because Neo4j docs say:
> 
> > You normally should not need to tune this parameter. No connection liveliness check is done by default ([None](https://docs.python.org/3/library/constants.html#None)). A value of 0 means connections will always be tested for validity. Negative values are not allowed.
> 
> https://neo4j.com/docs/api/python-driver/current/api.html#liveness-check-timeout
> 
> Appreciate the context!

@achantavy no, i haven’t hit a production crash because of this. the motivation was more around how Cartography is often run on a schedule with long gaps between syncs. in those cases, connections can get silently dropped by firewalls or NAT gateways while the driver still thinks they are valid. when the next sync starts, the failure only shows up when the connection is reused.

the Neo4j docs do mention that this usually isn’t required, which is fair. the idea here was just to make it configurable for environments where idle connections may be closed externally. it stays optional, and the default driver behavior remains unchanged.

achantavy:
Thanks for the PR, merging in! I think it's worth exposing this param

PR #2478, opened 2026-03-08
feat(aws): Add CloudFormation Stack support

### LINKED ISSUE #2265: feat(aws): Add support for Add CloudFormation Stack
## Summary

Add support for AWS CloudFormation Stack resources to enable detection of privilege escalation paths involving infrastructure-as-code deployments.

## Motivation

CloudFormation is one of the common AWS services exploited for privilege escalation. With `iam:PassRole` + `cloudformation:CreateStack`, an attacker can deploy arbitrary AWS resources (Lambda functions, EC2 instances, etc.) with embedded malicious code that executes with elevated IAM role permissions.

The [pathfinding.cloud dataset](https://github.com/DataDog/pathfinding-cloud) documents **5 CloudFormation-based privilege escalation paths**:
- cloudformation-001: `iam:PassRole` + `cloudformation:CreateStack`
- cloudformation-002: `cloudformation:UpdateStack`
- cloudformation-003: `cloudformation:CreateChangeSet` + `cloudformation:ExecuteChangeSet`
- cloudformation-004: Nested stack privilege escalation
- cloudformation-005: Transform-based code execution

## References

- [Pathfinding.cloud - CloudFormation paths](https://github.com/DataDog/pathfinding-cloud/tree/main/paths/cloudformation)
- [AWS CloudFormation Stack API](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Stack.html)
- [AWS Privilege Escalation via CloudFormation](https://rhinosecuritylabs.com/aws/aws-privilege-escalation-methods-mitigation/)
- [HackingTheCloud - PassRole + CloudFormation CreateStack](https://hackingthe.cloud/aws/exploitation/iam_privilege_escalation/#iampassrole-cloudformationcreatestack)

## Privilege Escalation Mappings

**File**: `cartography/data/permission_relationships.yaml`

```yaml
# CloudFormation - Deploy resources with embedded code
- target_label: CloudFormationStack
  permissions:
  - iam:PassRole
  - cloudformation:CreateStack
  relationship_name: CAN_EXEC

- target_label: CloudFormationStack
  permissions:
  - cloudformation:UpdateStack
  relationship_name: CAN_EXEC

- target_label: CloudFormationStack
  permissions:
  - cloudformation:CreateChangeSet
  - cloudformation:ExecuteChangeSet
  relationship_name: CAN_EXEC
```

yashasviyadav30:
### Type of change

### Summary

Add support for AWS CloudFormation Stack ingestion to enable detection of privilege escalation paths involving infrastructure-as-code deployments.

CloudFormation is one of the key AWS services exploited for privilege escalation. With `cloudformation:UpdateStack`, an attacker can deploy or modify arbitrary resources with embedded malicious code that executes under elevated IAM role permissions. This PR makes those attack paths visible in the Cartography graph.

**Changes:**

- **Node:** `CloudFormationStack` with core properties (`id`, `arn`, `stack_name`, `stack_status`, `role_arn`, `region`, etc.)
- **Relationships:**
  - `(AWSAccount)-[:RESOURCE]->(CloudFormationStack)` — ownership
  - `(CloudFormationStack)-[:HAS_EXECUTION_ROLE]->(AWSRole)` — maps the IAM role CloudFormation assumes during stack operations
- **Permission mappings** (`CAN_EXEC`) for privilege escalation via:
  - `cloudformation:UpdateStack`
- **Integration tests** covering node creation, both relationships, and the edge case where `RoleARN` is absent
- **Schema documentation** updated

### Related issues or links

- Fixes #2265
- References: [pathfinding.cloud - CloudFormation paths](https://pathfinding.cloud/paths/cloudformation-001.html)

### How was this tested?

**1. Integration tests:**

`pytest tests/integration/cartography/intel/aws/test_cloudformation.py -v` → 1 passed in 14.51s

**2. Real AWS sync (us-east-1):**

Synced IAM + CloudFormation together — stack got picked up correctly:

> INFO:cartography.intel.aws.cloudformation:Loading 1 CloudFormation stacks for region 'us-east-1' into graph.
> INFO:cartography.client.core.tx:Loaded 1 CloudFormationStack nodes
> INFO:cartography.graph.job:Finished job CloudFormationStack

**3. Neo4j verification:**

`MATCH (a:AWSAccount)-[:RESOURCE]->(c:CloudFormationStack)-[:HAS_EXECUTION_ROLE]->(r:AWSRole) RETURN a.id AS Account, c.stack_name AS Stack, c.role_arn AS StackRoleARN, r.arn AS RoleARN`

Got back 3 nodes, 2 relationships as expected. Screenshots below.

#### General

#### Proof of functionality

### Notes for reviewers
- Followed the EMR module pattern (GET → TRANSFORM → LOAD → CLEANUP) to keep things consistent.
- `HAS_EXECUTION_ROLE` points to `AWSRole` instead of `AWSPrincipal` — intentional since IAM roles have both `id` and `arn` set from the role ARN.
- `Tags` is serialized with `json.dumps()` because boto3 returns it as a list of dicts. Skipped `Outputs`, `Parameters`, etc. for now to keep the node simple — can always add them later.
- As requested, I have removed the `AND` permission combinations (like `PassRole + CreateStack`) from `permission_relationships.yaml` since the YAML schema uses OR logic. This PR now accurately maps the single `UpdateStack` permission vector.

### Known Limitations

- **CAN_EXEC over-broadness:** The `CAN_EXEC` edge is created for any principal with `cloudformation:UpdateStack` permission on a stack. However, true privilege escalation only occurs when the target stack has a service role (`role_arn` is set), because stacks without a service role execute with the caller's own permissions. The current YAML format cannot express role-gating; this is documented in `schema.md` and can be scoped in a follow-up PR. Downstream consumers should filter on `role_arn IS NOT NULL` for high-confidence escalation paths.

WhatsApp Image 2026-03-08 at 9 58 15 PM

WhatsApp Image 2026-03-08 at 9 59 42 PM


tests/data/aws/cloudformation.py
```diff
@@ -0,0 +1,38 @@
+TEST_ACCOUNT_ID = "000000000000"
+TEST_REGION = "us-east-1"
+TEST_UPDATE_TAG = 123456789
+
+DESCRIBE_STACKS = [
+    {
+        "StackId": f"arn:aws:cloudformation:{TEST_REGION}:{TEST_ACCOUNT_ID}:stack/test-stack-1/11111111-1111-1111-1111-111111111111",
+        "StackName": "test-stack-1",
+        "Description": "Test CloudFormation stack for cartography",
+        "CreationTime": "2024-01-01T12:00:00Z",
```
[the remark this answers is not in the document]
yashasviyadav30: Noted, will fix the datetime fixtures in a follow-up PR.

yashasviyadav30:
Hi @jychp
just a gentle ping on this PR.
All checks are passing and tested against real AWS with sync logs and Neo4j screenshots as suggested.
Ready for review whenever you get a chance.
Thanks!

yashasviyadav30:
Hi @jychp, checking in on this PR — let me know if any changes are needed. Thanks!

cartography/intel/aws/cloudformation.py
```diff
@@ -0,0 +1,115 @@
+import json
+import logging
+from typing import Any
+
+import boto3
+import neo4j
+
+from cartography.client.core.tx import load
+from cartography.graph.job import GraphJob
+from cartography.intel.aws.ec2.util import get_botocore_config
```
jychp: **blocking**: Wrong import path `cartography/intel/aws/ec2/util.py` has been replaced.

```python
from cartography.intel.aws.util.botocore_config import get_botocore_config
```

cartography/intel/aws/resources.py
```diff
@@ -131,6 +132,7 @@
         "apigateway": apigateway.sync,
         "apigatewayv2": apigatewayv2.sync,
         "bedrock": bedrock.sync,
+        "cloudformation": cloudformation.sync,
```
jychp: **blocking**: Sync ordering issue`RESOURCE_FUNCTIONS` is an `OrderedDict` and modules execute in insertion order. Here, `permission_relationships` (line 130) runs *before* `cloudformation` (line 135), so `CloudFormationStack` nodes won't exist when the permission analysis evaluates `CAN_EXEC` rules. The edges will never be created during a normal full sync.

Fix: move `"cloudformation": cloudformation.sync` before `"permission_relationships": permission_relationships.sync` (same pattern as the EC2/ECS dependency documented on line 118).

cartography/data/permission_relationships.yaml
```diff
@@ -42,3 +42,18 @@
...
+  relationship_name: CAN_EXEC
+- target_label: CloudFormationStack
+  permissions:
+  - cloudformation:UpdateStack
+  relationship_name: CAN_EXEC
+- target_label: CloudFormationStack
+  permissions:
+  - cloudformation:CreateChangeSet
+  - cloudformation:ExecuteChangeSet
+  relationship_name: CAN_EXEC
```
jychp: **blocking**: The permission engine (`evaluate_policy_for_permissions`) uses **OR** semantics it returns `True` on the *first* matching permission in the list (confirmed by the file header: "principals that have _any_ of the below permissions").

This means:
- **Entry 1**: A principal with only `iam:PassRole` (without `cloudformation:CreateStack`) gets marked `CAN_EXEC` on all stacks. `iam:PassRole` is a very common permission, so this creates many false-positive escalation paths.
- **Entry 3** (lines 55-59): A principal with only `cloudformation:CreateChangeSet` (without `ExecuteChangeSet`) also gets `CAN_EXEC` incorrectly.
- **Entry 2** (`cloudformation:UpdateStack` alone) is correct.

The YAML format cannot express AND semantics. I'd recommend keeping only the single-permission entry (`cloudformation:UpdateStack`) and deferring entries 1 and 3 to a follow-up PR that adds AND support to the permission framework. The issue template (#2265) provided this same YAML, so this is an inherited design gap not your fault.

cartography/intel/aws/resources.py
jychp: And add a comment to prevent someone else to move that dependency in the future.

jychp (changes requested):

cartography/intel/aws/cloudformation.py
yashasviyadav30: Fixed - Updated import to use cartography.intel.aws.util.botocore_config

cartography/intel/aws/resources.py
yashasviyadav30: Fixed. Moved cloudformation before permission_relationships and added dependency comment.

cartography/data/permission_relationships.yaml
yashasviyadav30: Fixed.

yashasviyadav30:
Hi @jychp,
Thank you for the thorough feedback and clear guidance on the issues. I've addressed all three points you raised and the PR is ready for review.

cartography/intel/aws/cloudformation.py
```diff
@@ -0,0 +1,115 @@
...
+) -> list[dict[str, Any]]:
+    client = boto3_session.client(
+        "cloudformation",
+        region_name=region,
+        config=get_botocore_config(),
+    )
+    stacks: list[dict[str, Any]] = []
+    paginator = client.get_paginator("describe_stacks")
+    for page in paginator.paginate():
+        stacks.extend(page["Stacks"])
```
jychp: **blocking**: `DescribeStacks` returns stacks in `DELETE_COMPLETE` for up to 90 days after deletion. As written, this loads every returned stack, so the graph will retain non-existent `CloudFormationStack` nodes for months and `permission_relationships` will emit `CAN_EXEC` edges pointing at stacks that can no longer be updated (false-positive escalation paths).

Filter out deleted stacks before `load()`, e.g.:

```python
stacks.extend(s for s in page["Stacks"] if s.get("StackStatus") != "DELETE_COMPLETE")
```

cartography/intel/aws/cloudformation.py
```diff
@@ -0,0 +1,115 @@
...
+@aws_handle_regions
+def get_cloudformation_stacks(
+    boto3_session: boto3.session.Session,
+    region: str,
+) -> list[dict[str, Any]]:
+    client = boto3_session.client(
+        "cloudformation",
+        region_name=region,
+        config=get_botocore_config(),
+    )
```
jychp: We recently migrate to a unified session builder:

```python
from cartography.intel.aws.util.botocore_config import create_boto3_client
...
client = create_boto3_client(boto3_session, "cloudformation", region_name=region)
```

cartography/intel/aws/cloudformation.py
```diff
@@ -0,0 +1,115 @@
...
+    return stacks
+
+
+def transform_cloudformation_stacks(
+    stacks: list[dict[str, Any]],
+) -> list[dict[str, Any]]:
+    for stack in stacks:
+        if stack.get("Tags") is not None:
+            stack["Tags"] = json.dumps(stack["Tags"])
+    return stacks
```
jychp: **nitpick**: This function both mutates the input list and returns it. Pick one pattern — either mutate in place and return `None`, or build and return a new list. Returning the mutated input is easy to misread as "the input is untouched".

cartography/data/permission_relationships.yaml
```diff
@@ -42,3 +42,8 @@
   - rds:StopDBInstance
   - rds:StartDBInstance
   relationship_name: CAN_ADMINISTER
+# Map principals that can execute CloudFormation stacks.
+- target_label: CloudFormationStack
+  permissions:
+  - cloudformation:UpdateStack
+  relationship_name: CAN_EXEC
```
jychp, line 49: **suggestion**: The PR description still advertises "Permission mappings (`CAN_EXEC`) for three documented privilege escalation paths" and lists `CreateStack+PassRole`, `UpdateStack`, and `CreateChangeSet+ExecuteChangeSet`. The final YAML only maps `UpdateStack` (the other two were intentionally dropped because the YAML format can't express AND semantics). Please update the PR description to reflect the actual scope so the merge commit and release notes don't claim coverage that isn't there.

jychp (changes requested):

cartography/intel/aws/cloudformation.py
yashasviyadav30: Fixed. I intentionally kept them initially because AWS retains deleted stacks for 90 days, and I thought it would be useful for historical forensic visibility in the graph. However, I completely agree with your point—from an attack-path modeling perspective, mapping CAN_EXEC edges to deleted stacks just creates false positives. I've added the filter to prioritize security accuracy.

cartography/intel/aws/cloudformation.py
yashasviyadav30: Fixed

cartography/intel/aws/cloudformation.py
yashasviyadav30: Fixed. I initially designed it to return the list so it could support functional pipeline chaining, but I agree that in Python, returning a mutated input creates ambiguity. I've updated it to strictly mutate in-place and return None for clearer semantics.

cartography/data/permission_relationships.yaml
yashasviyadav30: fixed

yashasviyadav30:
Hi @jychp, just pushed the updates based on your feedback. Thanks for the detailed review and helping me refine this!

cartography/intel/aws/cloudformation.py
```diff
@@ -0,0 +1,119 @@
...
+        region_name=region,
+    )
+    stacks: list[dict[str, Any]] = []
+    paginator = client.get_paginator("describe_stacks")
+    for page in paginator.paginate():
+        # Filtering DELETE_COMPLETE stacks to prevent false-positive CAN_EXEC escalation paths
+        stacks.extend(
+            s
+            for s in page.get("Stacks", [])
+            if s.get("StackStatus") != "DELETE_COMPLETE"
```
jychp: **suggestion**: The comment claims this filter prevents false-positive `CAN_EXEC` paths, but `DELETE_COMPLETE` is only one of several non-updatable terminal states. `ROLLBACK_COMPLETE`, `DELETE_FAILED`, `UPDATE_ROLLBACK_FAILED`, etc. all reject `UpdateStack` with `ValidationError`, so the same false-positive concern applies. Either expand the allow-list to actually-updatable states (or block-list all non-actionable terminal states), or drop the filter and address the broader gating in the same follow-up that scopes CAN_EXEC properly.

tests/integration/cartography/intel/aws/test_cloudformation.py
```diff
@@ -0,0 +1,147 @@
...
+            f"arn:aws:cloudformation:{TEST_REGION}:{TEST_ACCOUNT_ID}:stack/test-stack-1/11111111-1111-1111-1111-111111111111",
+            f"arn:aws:iam::{TEST_ACCOUNT_ID}:role/CloudFormationExecutionRole",
+        ),
+    }
+
+
+def test_cleanup_cloudformation_stacks(neo4j_session):
+    # Arrange: Create account and load a stack through the normal load path
+    from datetime import datetime
+    from datetime import timezone
```
jychp: **nitpick**: Move `from datetime import datetime` and `from datetime import timezone` to the top of the file with the other imports. Inline imports inside test bodies are inconsistent with the rest of the test suite and the project import style.

tests/integration/cartography/intel/aws/test_cloudformation.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+
+    boto3_session = MagicMock()
+    create_test_account(neo4j_session, TEST_ACCOUNT_ID, TEST_UPDATE_TAG)
+
+    # Create a mock AWSRole for the relationship test
+    neo4j_session.run(
+        "MERGE (r:AWSRole {arn: $arn})",
+        arn=f"arn:aws:iam::{TEST_ACCOUNT_ID}:role/CloudFormationExecutionRole",
+    )
```
jychp, line 37: Use `run_write_query` (or the IAM role load helper) instead of a raw `neo4j_session.run`

jychp (changes requested):

yashasviyadav30:
Thanks for the detailed review @jychp, really appreciate the thoroughness.

Addressed all four points:

1. **CAN_EXEC over-broadness** — added a note in `schema.md` under the CAN_EXEC relationship and a "Known Limitations" section in the PR description. Keeping the YAML as-is since it can't express role-gating, happy to tackle that in a follow-up.

2. **State filter** — swapped the `DELETE_COMPLETE` check for a proper allow-list of updatable states (`CREATE_COMPLETE`, `UPDATE_COMPLETE`, `UPDATE_ROLLBACK_COMPLETE`, `IMPORT_COMPLETE`, `IMPORT_ROLLBACK_COMPLETE`). Moved it to a module-level constant. Left out `ROLLBACK_COMPLETE` since those stacks only accept delete, not update.

3. **Inline imports** — moved to the top of the file.

4. **`run_write_query`** — done, replaced the raw `neo4j_session.run`.

Let me know if anything else needs adjusting!

tests/integration/cartography/intel/aws/test_cloudformation.py
```diff
@@ -0,0 +1,148 @@
...
+    cartography.intel.aws.cloudformation.load_cloudformation_stacks(
+        neo4j_session,
+        stale_stack_data,
+        TEST_REGION,
+        TEST_ACCOUNT_ID,
+        TEST_UPDATE_TAG,
+    )
+
+    # Pre-assert: stale node exists
+    pre = neo4j_session.run(
```
jychp, line 131: nitpick: These pre/post-cleanup assertions use raw `neo4j_session.run` reads. Use `check_nodes()` like the rest of the suite for consistency (you already moved the write to `run_write_query`).

jychp (approved):

PR #2479, opened 2026-03-08
feat(aws): Add AppRunner service support

### LINKED ISSUE #2264: feat(aws): Add support for AppRunner
## Summary

Add support for AWS App Runner Service resources to enable detection of privilege escalation paths involving containerized web application deployments.

## Motivation

AWS App Runner is a fully managed service for deploying containerized web applications. With `iam:PassRole` + `apprunner:CreateService`, an attacker can deploy a malicious container with an elevated IAM role, then access the running service to execute arbitrary code with those permissions.

## References

- [Pathfinding.cloud - App Runner paths](https://github.com/DataDog/pathfinding-cloud/tree/main/paths/apprunner)
- [AWS App Runner Service API](https://docs.aws.amazon.com/apprunner/latest/api/API_Service.html)
- [App Runner Security Documentation](https://docs.aws.amazon.com/apprunner/latest/dg/security.html)
- [AWS App Runner IAM roles](https://docs.aws.amazon.com/apprunner/latest/dg/security-iam-service-with-iam.html)

## Privilege Escalation Mappings

**File**: `cartography/data/permission_relationships.yaml`

```yaml
# App Runner - Execute code via service deployment
- target_label: AppRunnerService
  permissions:
  - iam:PassRole
  - apprunner:CreateService
  relationship_name: CAN_EXEC

- target_label: AppRunnerService
  permissions:
  - apprunner:UpdateService
  relationship_name: CAN_EXEC
```

raajheshkannaa:
Closes #2264

### Summary
- Adds `AppRunnerService` intel module — lists and describes services, flattens nested config fields, loads to graph
- Adds permission relationships for `CreateService` + `PassRole` and `UpdateService`
- Integration tests for node creation, relationships, and cleanup

### Type of change

### How was this tested?
- Integration tests: `test_load_apprunner_services_nodes`, `test_load_apprunner_services_relationships`, `test_cleanup_apprunner`
- Pre-commit passes (isort, black, flake8, pyupgrade)


raajheshkannaa:
Thanks for the review. Addressing each finding:

**#4 (ServiceId to ServiceArn) and #5 (permission_relationships.yaml OR semantics)**: Both valid, fixed in c925767.

**#1 (ClientError swallowing in describe loop)**: False positive. This matches the established pattern in `cartography/intel/aws/emr.py:57-66`. Log-and-skip individual items so a single deleted/inaccessible service doesn't fail the entire sync.

**#2 (explicit transform() stage)**: False positive. There's no project-wide requirement for a separate `transform()` function. Most modules (EMR, ECS, etc.) do transformation inline. EKS happens to have one, but it's not a convention.

**#3 (tests should call sync() not load())**: False positive. Calling `load()` directly is the common pattern across existing integration tests (e.g., `test_ecs.py` calls `load_ecs_clusters()` directly). Both approaches exist in the codebase.

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,133 @@
...
+    common_job_parameters: Dict[str, Any],
+) -> None:
+    for region in regions:
+        logger.info(
+            f"Syncing AppRunner for region '{region}' in account '{current_aws_account_id}'.",
+        )
+
+        services = get_apprunner_services(boto3_session, region)
+
+        load_apprunner_services(
```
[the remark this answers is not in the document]
raajheshkannaa: Fixed in 73e55b2. Extracted field flattening into transform_apprunner_services(), matching the get -> transform -> load -> cleanup pattern in secretsmanager, ecr, and kms.

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,133 @@
...
+            ingress_config = network_config.get("IngressConfiguration", {})
+            svc["IsPubliclyAccessible"] = ingress_config.get("IsPubliclyAccessible")
+
+            described_services.append(svc)
+        except botocore.exceptions.ClientError as e:
+            code = e.response["Error"]["Code"]
+            msg = e.response["Error"]["Message"]
+            logger.warning(
+                f"Could not run AppRunner describe_service due to boto3 error {code}: {msg}. Skipping.",
+            )
```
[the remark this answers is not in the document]
raajheshkannaa: Fixed in 73e55b2. Removed the blanket ClientError catch. Errors now propagate, matching ecr and secretsmanager patterns.

jychp:
> **#4 (ServiceId to ServiceArn) and #5 (permission_relationships.yaml OR semantics)**: Both valid, fixed in [c925767](https://github.com/cartography-cncf/cartography/commit/c925767a998396f515c7bc6de392b66031deaa5c).
> 
> **#1 (ClientError swallowing in describe loop)**: False positive. This matches the established pattern in `cartography/intel/aws/emr.py:57-66`. Log-and-skip individual items so a single deleted/inaccessible service doesn't fail the entire sync.
> 
> **#2 (explicit transform() stage)**: False positive. There's no project-wide requirement for a separate `transform()` function. Most modules (EMR, ECS, etc.) do transformation inline. EKS happens to have one, but it's not a convention.
> 
> **#3 (tests should call sync() not load())**: False positive. Calling `load()` directly is the common pattern across existing integration tests (e.g., `test_ecs.py` calls `load_ecs_clusters()` directly). Both approaches exist in the codebase.

Please follow Cubic PR review.

transform is mandatory in new modules to avoid sending potential or unwanted sensitive data to neo4j.

testing sync function is an actual requirement for any new module

tests/data/aws/apprunner.py
jychp, line 1: Test data should match the raw `describe_service` API response format (nested `SourceConfiguration`, `InstanceConfiguration`, `NetworkConfiguration`), and tests should call `sync()` with a mocked `get_apprunner_services` so that `transform_apprunner_services()` gets exercised. Currently the transform function has zero test coverage.

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,140 @@
+import logging
+from typing import Any
+from typing import Dict
+from typing import List
```
jychp: Use builtin `list` and `dict` type hints instead of importing from `typing`. The `typing.List`/`typing.Dict` aliases are deprecated since Python 3.9.

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,140 @@
...
+def load_apprunner_services(
+    neo4j_session: neo4j.Session,
+    data: List[Dict[str, Any]],
+    region: str,
+    current_aws_account_id: str,
+    aws_update_tag: int,
+) -> None:
+    logger.info(
+        f"Loading AppRunner {len(data)} services for region '{region}' into graph.",
+    )
```
jychp, line 86: Use `%s` formatting for lazy evaluation per project conventions, e.g.:
```python
logger.info("Loading AppRunner %s services for region '%s' into graph.", len(data), region)
```

jychp (changes requested):

tests/data/aws/apprunner.py
raajheshkannaa: Fixed in 06288de6. Test data now mirrors raw describe_service response with nested structures, and tests exercise sync() with mocked get_apprunner_services so transform gets tested too.

cartography/intel/aws/apprunner.py
raajheshkannaa: Fixed in 06288de6. Removed typing imports, using builtin list/dict throughout.

cartography/intel/aws/apprunner.py
raajheshkannaa: Fixed in 06288de6. Switched to %s lazy evaluation for all logger calls.

raajheshkannaa:
@jychp rebased on master. All 3 items from your review are addressed in the earlier commits. Would appreciate a re-review when you get a chance.

raajheshkannaa:
Rebased on latest master. All 3 items from your review are addressed in 06288de6 (test data shape, builtin type hints, lazy logger formatting). Ready for another look when you get a chance.

raajheshkannaa:
Rebased on latest master and fixed the CI failure. The root cause was an incorrect import path for `get_botocore_config`. It was importing from `cartography.intel.aws.ec2.util` (which does not exist), instead of `cartography.intel.aws.util.botocore_config` (the correct path used by every other module).

All your review items from before (raw test data shape, sync() in tests, builtin type hints, lazy logger formatting) remain addressed. Unit tests pass locally after the import fix.

raajheshkannaa:
@jychp all three review items are addressed and CI is green. Mind taking another look when you get a chance?

raajheshkannaa (commented):
Hi @jychp, all feedback from your review has been addressed in commit 06288de6. Would you mind taking another look when you get a chance? Thanks!

raajheshkannaa (commented):
Rebased on master. All review items from the previous round were addressed in prior commits. Ready for another look when you have time.

cartography/models/aws/apprunner.py
```diff
@@ -0,0 +1,59 @@
...
+class AppRunnerServiceNodeProperties(CartographyNodeProperties):
+    access_role_arn: PropertyRef = PropertyRef("AccessRoleArn")
+    arn: PropertyRef = PropertyRef("ServiceArn", extra_index=True)
+    auto_deployments_enabled: PropertyRef = PropertyRef("AutoDeploymentsEnabled")
+    cpu: PropertyRef = PropertyRef("Cpu")
+    created_at: PropertyRef = PropertyRef("CreatedAt")
+    egress_type: PropertyRef = PropertyRef("EgressType")
+    id: PropertyRef = PropertyRef("ServiceArn")
+    image_identifier: PropertyRef = PropertyRef("ImageIdentifier")
+    instance_role_arn: PropertyRef = PropertyRef("InstanceRoleArn")
```
jychp: The motivation in #2264 is detecting `iam:PassRole + apprunner:CreateService` privilege escalation to the service's execution role. `access_role_arn` and `instance_role_arn` are stored as plain string properties, but no relationship is created to the corresponding `AWSRole` nodes.

Without those edges, the graph can't traverse `Principal -[CAN_EXEC]-> AppRunnerService -[USES_ROLE]-> AWSRole`, which is the attack path this module is meant to enable detecting. Is this planned as a follow-up, or should it land here?

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,142 @@
...
+    transformed: list[dict[str, Any]] = []
+    for svc in services:
+        svc = dict(svc)
+
+        source_config = svc.get("SourceConfiguration", {})
+        image_repo = source_config.get("ImageRepository", {})
+        svc["ImageIdentifier"] = image_repo.get("ImageIdentifier")
+        svc["AutoDeploymentsEnabled"] = source_config.get("AutoDeploymentsEnabled")
+        auth_config = source_config.get("AuthenticationConfiguration", {})
+        svc["AccessRoleArn"] = auth_config.get("AccessRoleArn")
```
jychp, line 54: `SourceConfiguration` may contain `CodeRepository` (GitHub source) instead of `ImageRepository`. For code-source services, `image_identifier` will always be `None` and the source pointer is silently dropped. Consider extracting `CodeRepository.RepositoryUrl` alongside `ImageIdentifier`.

cartography/models/aws/apprunner.py
jychp, line 1: No `### AppRunnerService` section added to `docs/root/modules/aws/schema.md`. Recent modules (e.g. Bedrock, EventBridge) document the node properties and relationships there — please add a matching entry.

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,142 @@
...
+@aws_handle_regions
+def get_apprunner_services(
+    boto3_session: boto3.session.Session,
+    region: str,
+) -> list[dict[str, Any]]:
+    client = boto3_session.client(
+        "apprunner",
+        region_name=region,
+        config=get_botocore_config(),
+    )
```
jychp: We recently migrate to a unified client creation:

```python
from cartography.intel.aws.util.botocore_config import create_boto3_client

client = create_boto3_client(boto3_session, "apprunner", region_name=region)
```

This is the only module calling `boto3_session.client(...)` directly — please switch for consistency.

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,142 @@
...
+    services: list[dict[str, Any]] = []
+    kwargs: dict[str, Any] = {}
+    while True:
+        response = client.list_services(**kwargs)
+        services.extend(response.get("ServiceSummaryList", []))
+        next_token = response.get("NextToken")
+        if not next_token:
+            break
+        kwargs["NextToken"] = next_token
```
jychp: Nit: prefer the boto3 paginator for consistency with other modules:

```python
paginator = client.get_paginator("list_services")
services: list[dict[str, Any]] = []
    services.extend(page.get("ServiceSummaryList", []))
```

jychp: Note: if paginate is not available please explain it with a comment

jychp (changes requested):
Thx for the followup, still some elements to change but we are close.

cartography/models/aws/apprunner.py
raajheshkannaa: Good catch. Added two explicit edges in d611eb75: `(AppRunnerService)-[USES_ACCESS_ROLE]->(AWSRole)` for the ECR image pull role and `(AppRunnerService)-[USES_INSTANCE_ROLE]->(AWSRole)` for the runtime role. Split labels instead of a single `USES_ROLE` so queries can distinguish the privesc surface (instance role is the iam:PassRole + apprunner:CreateService target; access role is less interesting). Covered in the new integration test.

cartography/intel/aws/apprunner.py
raajheshkannaa: Done in d611eb75. The transform now extracts `CodeRepository.RepositoryUrl` alongside `ImageRepository.ImageIdentifier`, and the property is exposed as `code_repository_url` on the node. Added a third test service using `CodeRepository` to cover the branch.

cartography/models/aws/apprunner.py
raajheshkannaa: Added in d611eb75. New `### AppRunnerService` section in `docs/root/modules/aws/schema.md` documents the node properties and the three relationships (RESOURCE, USES_ACCESS_ROLE, USES_INSTANCE_ROLE), matching the style of the EventBridge entries.

cartography/intel/aws/apprunner.py
raajheshkannaa: Switched to `create_boto3_client(boto3_session, "apprunner", region_name=region)` in d611eb75. Dropped the explicit `get_botocore_config()` call since the helper already applies the default config.

cartography/intel/aws/apprunner.py
raajheshkannaa: Switched to the boto3 paginator in d611eb75. The `describe_service` loop stays a plain for-loop since that API returns a single service and has no paginator.

raajheshkannaa: `list_services` does support pagination, so using the paginator directly. No fallback comment needed.

cartography/models/aws/apprunner.py
```diff
@@ -0,0 +1,100 @@
...
+    memory: PropertyRef = PropertyRef("Memory")
+    name: PropertyRef = PropertyRef("ServiceName")
+    region: PropertyRef = PropertyRef("Region", set_in_kwargs=True)
+    service_url: PropertyRef = PropertyRef("ServiceUrl")
+    status: PropertyRef = PropertyRef("Status")
+    updated_at: PropertyRef = PropertyRef("UpdatedAt")
+
+
+@dataclass(frozen=True)
+class AppRunnerServiceToAWSAccountRelRelProperties(CartographyRelProperties):
```
jychp: Typo: `AppRunnerServiceToAWSAccountRelRelProperties` has `Rel` twice. Both other relationship-properties classes in this file (`AppRunnerServiceToAWSRoleRelProperties`) and the rest of the codebase (e.g. `EventBridgeRuleToAwsAccountRelProperties`) use a single `Rel`. Rename to `AppRunnerServiceToAWSAccountRelProperties` (and update the references on lines 50-51).

tests/integration/cartography/intel/aws/test_apprunner.py
```diff
@@ -0,0 +1,230 @@
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+import cartography.intel.aws.apprunner
+import cartography.intel.aws.iam
```
jychp: `import cartography.intel.aws.iam` is unused; `load_role_data` and `transform_role_trust_policies` are pulled in via the explicit `from ... import` lines just below. Drop this bare module import.

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,134 @@
...
+    client = create_boto3_client(boto3_session, "apprunner", region_name=region)
+    paginator = client.get_paginator("list_services")
+    services: list[dict[str, Any]] = []
+    for page in paginator.paginate():
+        services.extend(page.get("ServiceSummaryList", []))
+
+    described_services: list[dict[str, Any]] = []
+    for service in services:
+        desc_response = client.describe_service(ServiceArn=service["ServiceArn"])
+        described_services.append(desc_response["Service"])
```
jychp, line 35: `describe_service` is N+1 with no throttling. AppRunner's default `DescribeService` quota is 10 req/s. The outer `aws_handle_regions` decorator only retries the entire `get_apprunner_services` call, so any throttle restarts pagination from scratch. Is that the intended trade-off, or should the loop sleep / use boto retry config between calls?

cartography/intel/aws/apprunner.py
```diff
@@ -0,0 +1,134 @@
...
+    common_job_parameters: dict[str, Any],
+) -> None:
+    for region in regions:
+        logger.info(
+            "Syncing AppRunner for region '%s' in account '%s'.",
+            region,
+            current_aws_account_id,
+        )
+
+        services = get_apprunner_services(boto3_session, region)
```
jychp, line 138: Nit: AppRunner is only available in a subset of regions. Other recent modules pre-filter via `filter_regions_to_supported_service_regions` (`cartography/intel/aws/util/service_regions.py`) to avoid wasted API calls and noisy `aws_handle_regions` warnings in unsupported regions. Worth a follow-up.

jychp (changes requested):

cartography/models/aws/apprunner.py
raajheshkannaa: Done in ac7df1a4. Renamed to AppRunnerServiceToAWSAccountRelProperties and updated the references on lines 50-51.

tests/integration/cartography/intel/aws/test_apprunner.py
raajheshkannaa: Dropped in ac7df1a4. Module-level import removed; the explicit `from ... import` lines below cover everything used in the test.

cartography/intel/aws/apprunner.py
raajheshkannaa: Intended trade-off, with one caveat. The boto3 client is created via `create_boto3_client` which applies the project default `max_attempts=3, mode=standard` (cartography/intel/aws/util/botocore_config.py). Standard mode retries throttling exceptions with exponential backoff and jitter, so the inner `describe_service` calls already get retries before bubbling up. Followed the same N+1 pattern as emr.describe_cluster.

If a service hits throttling hard enough to exhaust all 3 inner retries, then yes, `aws_handle_regions` restarts pagination from scratch for that region. Worth a follow-up if it shows up in practice (eg. an inner sleep between describes, or wiring CARTOGRAPHY_AWS_MAX_ATTEMPTS up via env). Happy to tackle that in a separate PR if you'd prefer.

cartography/intel/aws/apprunner.py
raajheshkannaa: Addressed in ac7df1a4. Added `filter_regions_to_supported_service_regions` to `sync` and an info log for skipped regions, matching the codebuild pattern.

raajheshkannaa:
@jychp thanks for the review. Pushed updates addressing your four points:

1. **Typo rename** (c4b4bc94): `AppRunnerServiceToAWSAccountRelRelProperties` → `AppRunnerServiceToAWSAccountRelProperties`.
2. **Unused import** (c4b4bc94): dropped `import cartography.intel.aws.iam` from `tests/integration/cartography/intel/aws/test_apprunner.py`.
3. **Region prefilter** (c4b4bc94): added `filter_regions_to_supported_service_regions` in `sync_apprunner` so we only iterate regions where AppRunner is offered.
4. **N+1 `describe_service`**: kept as-is. This mirrors the existing cartography pattern (see `cartography/intel/aws/kms.py` which does `describe_key` per key from the paginated list, and `acm.py` doing the same). AppRunner accounts in practice have small service counts, so the 10 req/s `DescribeService` quota is rarely a concern. If a throttle does fire, the outer `aws_handle_regions` retry restarts pagination, which is the same trade-off the other modules accept. Happy to switch to a different pattern if you want a stricter throttle here — let me know.

Also rebased on latest master (8 commits behind). CI is green.

raajheshkannaa:
@jychp rebased on latest master (97f5a3af) after upstream PR #2833 landed; conflict in permission_relationships.yaml resolved by keeping both the upstream PassRole block and the AppRunner CAN_EXEC block. All four review items from your last round remain addressed. CI green. Ready for another look when you get a chance.

raajheshkannaa:
Rebased on latest master after #2478 (CloudFormation Stack support) landed. The conflict was in `permission_relationships.yaml`; resolved by keeping both the upstream CloudFormationStack CAN_EXEC block and the AppRunnerService block. All prior review items remain addressed. Ready for another look when you have time.

tests/integration/cartography/intel/aws/test_apprunner.py
raajheshkannaa: Removed the bare `import cartography.intel.aws.iam`. The test now relies only on the `from cartography.intel.aws.iam import load_role_data, transform_role_trust_policies` imports, which is enough for the role-loading setup.

cartography/intel/aws/apprunner.py
raajheshkannaa: The client comes from `create_boto3_client`, which sets botocore's `standard` retry mode (3 attempts). That retries throttling responses with exponential backoff, so the per-service `describe_service` calls back off automatically rather than failing on the 10 req/s quota. This matches how the other AWS modules handle their describe loops.

cartography/intel/aws/apprunner.py
raajheshkannaa: Added the pre-filter. Regions now pass through `filter_regions_to_supported_service_regions` before the sync loop, so unsupported regions are skipped.

raajheshkannaa:
Rebased on latest master to resolve a conflict in `permission_relationships.yaml` (upstream added the SSM Session Manager `CAN_START_SESSION` block). Kept both entries; the AppRunner block is unchanged. CI is green and the PR is conflict-free again. Ready for another look whenever you have time, @jychp.

raajheshkannaa:
@jychp all review threads have replies from the June round, and I've now rebased onto master. It was 169 commits behind and broken against current `main`:

- **#3020 provider prefixes.** The graph label is now `AWSAppRunnerService`. Class names stay unprefixed, matching `AWSCodeBuildProject` and `AWSELBV2TargetGroup`. No legacy alias, since this label has never shipped. Also updated `permission_relationships.yaml`.
- The integration test seeded `EC2Instance`, now a deprecated alias, so it's `AWSEC2Instance`.
- **#3021 generated schema docs.** Added descriptions to all 17 node properties.
- **`docs/root/modules/aws/schema.md` is gone**, so the manual schema entry is dropped.

Unit suite is at parity with a clean `master` checkout: same 27 pre-existing failures, zero introduced here. `pre-commit` passes on all changed files. No Docker locally so the integration tests are on CI.

Ready for another look whenever you have time.

PR #2483, opened 2026-03-10
feat(rules): expand EOL software rule for Kubernetes and EC2

kunaals:
### Type of change

### Summary
Expand the `eol_software` rule to detect end-of-life infrastructure software across Kubernetes and EC2.

Current coverage includes:
- `EKSCluster` nodes using the Amazon EKS provider lifecycle
- `KubernetesCluster` nodes using the upstream Kubernetes support window when the cluster is not matched to EKS
- `EC2Instance` nodes using AWS SSM `InstanceInformation` to detect Amazon Linux 2 after the vendor end-of-life date

The Kubernetes fact excludes clusters that overlap with an `EKSCluster` so EKS-backed clusters are evaluated once and against the correct lifecycle source. The EC2 fact is intentionally narrow and uses `SSMInstanceInformation.platform_name` plus `platform_version` instead of AMI-name heuristics.

### Related issues or links
- N/A for issue tracking
- Kubernetes version support policy: https://kubernetes.io/releases/version-skew-policy/
- Kubernetes releases: https://kubernetes.io/releases/
- Amazon EKS Kubernetes version lifecycle: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
- Amazon Linux 2 release notes: https://docs.aws.amazon.com/AL2/latest/relnotes/relnotes-al2.html
- Amazon Linux 2023 release cadence: https://docs.aws.amazon.com/linux/al2023/ug/release-cadence.html
- AWS Systems Manager InstanceInformation API: https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_InstanceInformation.html

### Breaking changes
None.

### How was this tested?
- `make test_lint`
- `uv run --frozen pytest -q tests/unit/rules/test_eol_software.py tests/unit/rules`
- Manual query validation against SubImage and Brookhealth graphs:
  - confirmed EKS and raw Kubernetes are evaluated with different support cutoffs
  - confirmed overlapping `KubernetesCluster` / `EKSCluster` pairs exist and are suppressed by the dedupe logic
  - confirmed the provider-first policy no longer flags EKS `1.32` as EOL
  - confirmed `SSMInstanceInformation` provides `platform_name`, `platform_version`, and `HAS_INFORMATION` coverage for EC2 instances
  - confirmed Amazon Linux 2 returns `0` findings as of March 10, 2026 because the AWS vendor EOL date is June 30, 2026
  - confirmed the EC2 query matches live Amazon Linux 2 instances in both environments when evaluated against a simulated post-EOL date

#### General

#### Proof of functionality

N/A, this PR only adds rules and tests.

N/A, this PR does not change schema.

N/A, this PR does not add an intel module.

### Notes for reviewers
This rule intentionally splits lifecycle authority by resource type:
- raw `KubernetesCluster` uses upstream Kubernetes support
- `EKSCluster` uses Amazon EKS support
- `EC2Instance` uses AWS SSM host OS metadata and vendor lifecycle dates

As of March 10, 2026:
- upstream-supported Kubernetes minors are `1.35`, `1.34`, and `1.33`
- Amazon EKS still supports `1.29` through `1.34`
- Amazon Linux 2 remains supported until June 30, 2026


cartography/rules/data/rules/eol_software.py
```diff
@@ -0,0 +1,250 @@
...
+           toString(e.version) AS software_version,
+           1 AS software_major,
+           kubernetes_minor AS software_minor,
+           e.region AS location,
+           'provider' AS support_basis,
+           'eol' AS support_status
+    ORDER BY asset_name
+    """,
+    cypher_visual_query=f"""
+    MATCH (e:EKSCluster)
```
achantavy: Since this is a visual query, we could make it more interesting and attach these to their owning AWS accounts, or other things that are interesting.

cartography/rules/data/rules/eol_software.py
```diff
@@ -0,0 +1,250 @@
...
+           toString(k.version) AS software_version,
+           1 AS software_major,
+           kubernetes_minor AS software_minor,
+           NULL AS location,
+           'upstream' AS support_basis,
+           'eol' AS support_status
+    ORDER BY asset_name
+    """,
+    cypher_visual_query=f"""
+    MATCH (k:KubernetesCluster)
```
achantavy, line 150: Same here about having a potentially more interesting visual query that isn't just nodes and no paths

achantavy (approved):

cartography/rules/data/rules/eol_software.py
kunaals: updated
Screenshot 2026-03-13 at 12 19 30 PM

cartography/rules/data/rules/eol_software.py
kunaals: similarly updated

PR #2488, opened 2026-03-11
feat(docker_scout): Coverage for Docker Scout 

shyammukund:
### Type of change

- [x] Documentation update

### Summary

This PR adds Docker Scout recommendation ingestion based on the standard text output of:

```bash
docker scout recommendations --output <FILE> <IMAGE>
```

The module now ingests recommendation reports only. The old `sbom` / `cves` JSON-based flow was removed.

This PR introduces the following Docker Scout entities:

- `DockerScoutPublicImage`
- `DockerScoutPublicImageTag`

And the following relationships:

- `(Image)-[BUILT_ON]->(DockerScoutPublicImage)`
  matching is done from the Docker Scout target digest prefix to `Image._ont_digest`
- `(DockerScoutPublicImage)-[BUILT_FROM]->(DockerScoutPublicImageTag)`
- `(DockerScoutPublicImage)-[SHOULD_UPDATE_TO]->(DockerScoutPublicImageTag)`

Additional modeling details:

- `alternative_tags` is stored on both `DockerScoutPublicImage` and `DockerScoutPublicImageTag`
- `benefits`, `fix_critical`, `fix_high`, `fix_medium`, and `fix_low` are stored on `SHOULD_UPDATE_TO`
- files can be named using any convention; the module does not rely on filenames
- files that are not valid Docker Scout recommendation reports are skipped

### Related issues or links

- Fixes #

### How was this tested?

Executed:

```bash
uv run --frozen pre-commit run --files cartography/intel/docker_scout/__init__.py cartography/intel/docker_scout/recommendation_parser.py cartography/intel/docker_scout/scanner.py cartography/models/docker_scout/image.py cartography/models/docker_scout/public_image_tag.py tests/integration/cartography/intel/docker_scout/test_docker_scout.py tests/unit/cartography/intel/test_docker_scout_recommendations.py docs/root/modules/docker_scout/config.md docs/root/modules/docker_scout/schema.md docs/root/modules/ontology/schema.md
uv run --frozen pytest tests/unit/cartography/intel/test_docker_scout_recommendations.py tests/integration/cartography/intel/docker_scout/test_docker_scout.py
```

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

`BUILT_ON` is modeled in the schema so `GraphJob` can clean stale relationships automatically.

The actual relationship creation still uses a targeted Cypher query because Docker Scout reports only a partial target digest, so a standard equality matcher on `Image._ont_digest` is not sufficient to create the edge.

### DIFF
--- a/cartography/rules/data/rules/__init__.py
+++ b/cartography/rules/data/rules/__init__.py
@@ -93,6 +93,7 @@ from cartography.rules.data.rules.database_instance_exposed import (
 from cartography.rules.data.rules.delegation_boundary_modifiable import (
     delegation_boundary_modifiable,
 )
+from cartography.rules.data.rules.eol_software import eol_software
 from cartography.rules.data.rules.identity_administration_privileges import (
     identity_administration_privileges,
 )
@@ -142,6 +143,7 @@ RULES = {
     compute_instance_exposed.id: compute_instance_exposed,
     database_instance_exposed.id: database_instance_exposed,
     delegation_boundary_modifiable.id: delegation_boundary_modifiable,
+    eol_software.id: eol_software,
     identity_administration_privileges.id: identity_administration_privileges,
     inactive_user_active_accounts.id: inactive_user_active_accounts,
     missing_mfa_rule.id: missing_mfa_rule,
--- a/cartography/rules/data/rules/eol_software.py
+++ b/cartography/rules/data/rules/eol_software.py
@@ -1,0 +1,254 @@
+from cartography.rules.spec.model import Fact
+from cartography.rules.spec.model import Finding
+from cartography.rules.spec.model import Maturity
+from cartography.rules.spec.model import Module
+from cartography.rules.spec.model import Rule
+from cartography.rules.spec.model import RuleReference
+
+_OLDEST_SUPPORTED_UPSTREAM_KUBERNETES_MINOR = 33
+_OLDEST_SUPPORTED_EKS_KUBERNETES_MINOR = 29
+_AMAZON_LINUX_2_EOL_DATE = "2026-06-30"
+
+EOL_SOFTWARE_REFERENCES = [
+    RuleReference(
+        text="Kubernetes Version Skew Policy",
+        url="https://kubernetes.io/releases/version-skew-policy/",
+    ),
+    RuleReference(
+        text="Kubernetes Releases",
+        url="https://kubernetes.io/releases/",
+    ),
+    RuleReference(
+        text="Amazon EKS Kubernetes version lifecycle",
+        url="https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html",
+    ),
+    RuleReference(
+        text="Amazon Linux 2 release notes",
+        url="https://docs.aws.amazon.com/AL2/latest/relnotes/relnotes-al2.html",
+    ),
+    RuleReference(
+        text="Amazon Linux 2023 release cadence",
+        url="https://docs.aws.amazon.com/linux/al2023/ug/release-cadence.html",
+    ),
+    RuleReference(
+        text="AWS Systems Manager InstanceInformation API",
+        url="https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_InstanceInformation.html",
+    ),
+]
+
+
+def _build_ec2_instance_amazon_linux_2_eol_query(
+    current_date_expression: str = "date()",
+) -> str:
+    return f"""
+    MATCH (ec2:EC2Instance)-[:HAS_INFORMATION]->(ssm:SSMInstanceInformation)
+    WHERE toLower(trim(coalesce(ssm.platform_name, ''))) = 'amazon linux'
+      AND trim(toString(ssm.platform_version)) = '2'
+      AND {current_date_expression} > date('{_AMAZON_LINUX_2_EOL_DATE}')
+    RETURN ec2.id AS asset_id,
+           coalesce(ec2.instanceid, ec2.id) AS asset_name,
+           'EC2Instance' AS asset_type,
+           'amazon-linux' AS software_name,
+           trim(toString(ssm.platform_version)) AS software_version,
+           2 AS software_major,
+           NULL AS software_minor,
+           coalesce(ssm.region, ec2.region) AS location,
+           'vendor' AS support_basis,
+           'eol' AS support_status
+    ORDER BY asset_name
+    """
+
+
+_eks_cluster_kubernetes_version_eol = Fact(
+    id="eks_cluster_kubernetes_version_eol",
+    name="EKS clusters running end-of-life Kubernetes versions",
+    description=(
+        "Detects EKS clusters running Kubernetes minor versions that are no longer "
+        "supported by Amazon EKS. As of 2026-03-10, EKS-supported Kubernetes "
+        "minors are 1.34, 1.33, 1.32, 1.31, 1.30, and 1.29."
+    ),
+    cypher_query=f"""
+    MATCH (e:EKSCluster)
+    WITH e,
+         CASE
+             WHEN e.version IS NULL OR size(split(toString(e.version), '.')) < 2 THEN NULL
+             ELSE toInteger(split(split(toString(e.version), '.')[1], '-')[0])
+         END AS kubernetes_minor
+    WHERE kubernetes_minor IS NOT NULL
+      AND kubernetes_minor < {_OLDEST_SUPPORTED_EKS_KUBERNETES_MINOR}
+    RETURN e.id AS asset_id,
+           e.name AS asset_name,
+           'EKSCluster' AS asset_type,
+           'kubernetes' AS software_name,
+           toString(e.version) AS software_version,
+           1 AS software_major,
+           kubernetes_minor AS software_minor,
+           e.region AS location,
+           'provider' AS support_basis,
+           'eol' AS support_status
+    ORDER BY asset_name
+    """,
+    cypher_visual_query=f"""
+    MATCH account_path=(a:AWSAccount)-[:RESOURCE]->(e:EKSCluster)
+    WITH account_path, e,
+         CASE
+             WHEN e.version IS NULL OR size(split(toString(e.version), '.')) < 2 THEN NULL
+             ELSE toInteger(split(split(toString(e.version), '.')[1], '-')[0])
+         END AS kubernetes_minor
+    WHERE kubernetes_minor IS NOT NULL
+      AND kubernetes_minor < {_OLDEST_SUPPORTED_EKS_KUBERNETES_MINOR}
+    OPTIONAL MATCH worker_path=(ec2:EC2Instance)-[:MEMBER_OF_EKS_CLUSTER]->(e)
+    WITH account_path, e, head(collect(worker_path)) AS worker_path
+    RETURN e AS cluster, account_path, worker_path
+    """,
+    cypher_count_query="""
+    MATCH (e:EKSCluster)
+    RETURN COUNT(e) AS count
+    """,
+    asset_id_field="asset_id",
+    module=Module.AWS,
+    maturity=Maturity.EXPERIMENTAL,
+)
+
+_kubernetes_cluster_kubernetes_version_eol = Fact(
+    id="kubernetes_cluster_kubernetes_version_eol",
+    name="Kubernetes clusters running end-of-life Kubernetes versions",
+    description=(
+        "Detects Kubernetes clusters running end-of-life minor versions. "
+        "If a native KubernetesCluster is the same EKS-backed cluster already "
+        "represented as an EKSCluster, it is excluded so managed clusters are "
+        "evaluated against the EKS provider lifecycle instead of upstream support."
+    ),
+    cypher_query=f"""
+    MATCH (k:KubernetesCluster)
+    WITH k,
+         CASE
+             WHEN k.version_minor IS NULL THEN NULL
+             ELSE toInteger(replace(toString(k.version_minor), '+', ''))
+         END AS kubernetes_minor
+    WHERE kubernetes_minor IS NOT NULL
+      AND kubernetes_minor < {_OLDEST_SUPPORTED_UPSTREAM_KUBERNETES_MINOR}
+      AND NOT EXISTS {{
+          MATCH (e:EKSCluster)
+          WHERE e.id = k.external_id
+             OR e.name = k.external_id
+             OR (k.api_server_url IS NOT NULL AND e.endpoint = k.api_server_url)
+      }}
+    RETURN k.id AS asset_id,
+           k.name AS asset_name,
+           'KubernetesCluster' AS asset_type,
+           'kubernetes' AS software_name,
+           toString(k.version) AS software_version,
+           1 AS software_major,
+           kubernetes_minor AS software_minor,
+           NULL AS location,
+           'upstream' AS support_basis,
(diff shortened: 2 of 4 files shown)


tests/integration/cartography/intel/docker_scout/test_docker_scout.py
```diff
@@ -0,0 +1,250 @@
...
+    )
+
+
+def test_docker_scout_sync(neo4j_session):
+    # Arrange: create external registry nodes that Docker Scout rels connect to
+    _create_ecr_image(neo4j_session, TEST_IMAGE_DIGEST, TEST_UPDATE_TAG)
+    _create_gitlab_container_image(neo4j_session, TEST_GITLAB_IMAGE_ID, TEST_UPDATE_TAG)
+
+    # Transform
+    public_image = transform_public_image(MOCK_SBOM_DATA, TEST_IMAGE_DIGEST)
```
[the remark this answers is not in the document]
shyammukund: fixed

cartography/models/docker_scout/image.py
```diff
@@ -0,0 +1,95 @@
...
+    rel_label: str = "BUILT_ON"
+    properties: DockerScoutPublicImageToImageRelProperties = (
+        DockerScoutPublicImageToImageRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class DockerScoutPublicImageSchema(CartographyNodeSchema):
+    label: str = "DockerScoutPublicImage"
+    scoped_cleanup: bool = False
```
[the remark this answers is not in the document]
shyammukund: This was done intentionally because with a scoped cleanup you would need a parent node to point the subresource relationship at. Since once package can relate to images in different registries it doesn't really make sense to do a scoped cleanup.

cartography/intel/docker_scout/scanner.py
```diff
@@ -0,0 +1,371 @@
+import json
```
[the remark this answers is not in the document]
shyammukund: again. same reason about needing un scoped cleanup here.

tests/integration/cartography/intel/docker_scout/test_docker_scout.py
```diff
@@ -0,0 +1,256 @@
...
+        neo4j_session,
+        TEST_IMAGE,
+        TEST_UPDATE_TAG,
+    )
+
+    # Verify get_sbom was called twice (scanned image, then public image)
+    assert mock_get_sbom.call_count == 2
+    assert mock_get_sbom.call_args_list[0].args[0] == TEST_IMAGE
+    assert mock_get_sbom.call_args_list[1].args[0] == "python:3.12-slim"
+    mock_get_cves.assert_called_once_with(TEST_IMAGE)
```
jychp: **suggestion**: Remove mock call-count/argument assertions from integration tests.

These lines test implementation details (how many times `get_sbom` was called and with what arguments) rather than behavior. The graph-state assertions below (`check_nodes`/`check_rels`) are comprehensive and sufficient. Integration tests should validate outcomes, not mock interactions — this makes the test fragile to refactoring.

tests/data/docker_scout/mock_data.py
```diff
@@ -0,0 +1,128 @@
+TEST_IMAGE = "000000000000.dkr.ecr.us-east-1.amazonaws.com/test-app:latest"
+TEST_IMAGE_DIGEST = (
+    "sha256:0000000000000000000000000000000000000000000000000000000000000000"
+)
+TEST_PUBLIC_IMAGE_ID = "python:3.12-slim"
+TEST_UPDATE_TAG = 123456789
+
+# GitLab container image ID matching the scanned image digest
+TEST_GITLAB_IMAGE_ID = TEST_IMAGE_DIGEST
```
jychp: **nitpick**: `TEST_GITLAB_IMAGE_ID = TEST_IMAGE_DIGEST` — using the same value for both means the test doesn't truly validate that ECR and GitLab relationships are independently matched. Consider using a distinct digest for the GitLab image to strengthen coverage for cross-registry relationship matching.

tests/integration/cartography/intel/docker_scout/test_docker_scout.py
jychp, line 1: Missing `__init__.py` files in `tests/data/docker_scout/` and `tests/integration/cartography/intel/docker_scout/`. Every other module in the project has them in both test directories. CI passes without them but this breaks project convention.

docs/root/modules/docker_scout/schema.md
jychp, line 1: We need and index.md file + config.md that explain how to setup auth.

cartography/cli.py
```diff
@@ -1254,6 +1215,18 @@ def run(
                 ),
             ] = "https://api.airbyte.com/v1",
             # =================================================================
+            # Docker Scout Options
+            # =================================================================
+            docker_scout_images: Annotated[
```
jychp: we also need to pass docker login + token for authentification

cartography/intel/docker_scout/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+from cartography.client.gitlab.container_images import get_gitlab_container_tags
+from cartography.config import Config
+from cartography.intel.docker_scout.scanner import cleanup
+from cartography.intel.docker_scout.scanner import sync
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+def _get_images_from_graph(neo4j_session: Session) -> set[str]:
```
jychp: Can we replace that with ontology to make only one query ?

cartography/cli.py
shyammukund: since we switched getting json instead of running docker we shouldn't need to pass anymore right

cartography/intel/docker_scout/scanner.py
```diff
@@ -0,0 +1,336 @@
...
+            name = artifact.get("name", "")
+            version = artifact.get("version", "")
+            purl_to_pkg_id[purl] = f"{version}|{name}"
+
+    for vuln_pkg in vuln_packages:
+        pkg_purl = vuln_pkg.get("purl", "")
+        package_id = purl_to_pkg_id.get(pkg_purl)
+
+        for vuln in vuln_pkg.get("vulnerabilities", []):
+            source_id = vuln["source_id"]
```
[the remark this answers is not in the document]
shyammukund: this is contradicting to a previous piece of feedback you gave me to not skip this because it is essential data and we need it to fail loudly. 

dont let this happen again ;)

cartography/intel/docker_scout/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+            results_dir,
+        )
+        raise ValueError("No Docker Scout json results found on disk")
+
+    logger.info("Processing %d local Docker Scout result files", len(json_files))
+
+    synced_count = 0
+    for file_path in json_files:
+        with open(file_path, encoding="utf-8") as f:
+            scout_data = json.load(f)
```
jychp: **suggestion**: Per-file JSON error handling is inconsistent with Trivy.

In Trivy's `sync_trivy_from_dir`, corrupt JSON files are caught and skipped:
```python
# Trivy (cartography/intel/trivy/__init__.py:302-308)
for file_path in json_files:
    try:
        with open(file_path, encoding="utf-8") as f:
            trivy_data = json.load(f)
    except json.JSONDecodeError as e:
        logger.error(...)
        continue
```

Here, a single corrupt file will `json.JSONDecodeError` and abort the entire Docker Scout ingestion. Since this module explicitly follows the Trivy pattern (reusing `get_json_files_in_dir`, same dir/S3 structure), consider matching Trivy's error resilience for local files.

I understand the "fail loudly" feedback from cubic, but note that Trivy itself doesn't follow that for JSON decode errors on local files — it skips and continues. The S3 path in both Trivy and Docker Scout lacks this handling, so at least there's consistency there.

cartography/intel/docker_scout/scanner.py
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+
+def transform_packages(
+    sbom_data: dict[str, Any],
+    image_digest: str,
+    public_image_id: str,
+) -> list[dict[str, Any]]:
+    """
+    Transform SBOM artifacts into DockerScoutPackage dicts.
```
jychp: From what I understand of this function and the fact that SBOM artifacts is empty in the mock data, we only ingest vulnerabilities. However, it could be interesting to have all packages (including non-vulnerable ones) represented in the graph IMHO

cartography/intel/docker_scout/scanner.py
jychp: **positive**: Agree with keeping direct indexing here. `source_id` is a required field in Docker Scout CVE output — failing loudly on malformed data is the correct approach for structured, first-party data.

tests/integration/cartography/intel/docker_scout/test_docker_scout.py
```diff
@@ -0,0 +1,213 @@
...
+        MERGE (g:GitLabContainerImage{id: $image_id})
+        ON CREATE SET g.firstseen = timestamp()
+        SET g.lastupdated = $update_tag
+        """,
+        image_id=image_id,
+        update_tag=update_tag,
+    )
+
+
+def test_docker_scout_sync_from_file(neo4j_session):
```
jychp, line 26: **suggestion**: Consider adding a cleanup test. This test verifies nodes and relationships are created, but doesn't verify that stale data is removed when cleanup runs with a newer `update_tag`. Other modules (e.g., Trivy) test this pattern. Something like:

```python
def test_docker_scout_cleanup(neo4j_session):
    # Ingest with tag 1
    _create_ecr_image(neo4j_session, TEST_ECR_IMAGE_DIGEST, TEST_UPDATE_TAG)
    sync_from_file(neo4j_session, MOCK_ECR_COMBINED_FILE_DATA, "test.json", TEST_UPDATE_TAG)
    
    # Run cleanup with tag 2 (simulating a new sync that didn't re-ingest)
    cleanup(neo4j_session, {"UPDATE_TAG": TEST_UPDATE_TAG + 1})
    
    # Assert stale nodes are removed
    assert check_nodes(neo4j_session, "DockerScoutPublicImage", ["id"]) == set()
```

docs/root/modules/docker_scout/config.md
```diff
@@ -0,0 +1,91 @@
...
+    docker login
+    ```
+
+    For CI environments, use a [Docker Hub access token](https://docs.docker.com/security/for-developers/access-tokens/):
+
+    ```bash
+    echo "$DOCKER_HUB_TOKEN" | docker login --username "$DOCKER_HUB_USERNAME" --password-stdin
+    ```
+
+1. Ensure your graph is populated with the container images you want to scan. For example, with AWS ECR:
```
jychp: **nitpick**: Step 3 mentions "Ensure your graph is populated with the container images you want to scan" — this is correct for relationship matching (BUILT_ON, DEPLOYED, AFFECTS need target nodes), but the phrasing could be clearer that it's about *having registry nodes in the graph for relationship creation*, not about scan target discovery (which is now driven by the JSON files, not the graph).

Consider something like: "Ensure the container image registry nodes (ECR, GCP, GitLab) are already synced into the graph so Docker Scout can create relationships to them."

docs/root/modules/docker_scout/config.md
```diff
@@ -0,0 +1,91 @@
...
+
+    ```bash
+    cartography --selected-modules aws --aws-requested-syncs ecr
+    ```
+
+### Generating scan results
```
jychp, line 36: Instead of combining files here, can we use two separate files with two separate argument for s3 prefix or maybe a naming convention ?

jychp (changes requested):

cartography/intel/docker_scout/__init__.py
shyammukund: Decided to raise the error just so we have the error handling but won't just continue on.

cartography/intel/docker_scout/scanner.py
shyammukund: Agreed. I have updated the code to get all packages. In the transform we now insert the cve packages first then overwrite them with an SBOM entry if there is a duplicate.

docs/root/modules/docker_scout/config.md
shyammukund: I think if we seperate the files the sync_from_file logic might become more complex. We would have to start matching the sbom files to the cve files. I think it would be better to keep them combined, but happy to defer if you feel strongly about this.

cartography/models/docker_scout/public_image_tag.py
```diff
@@ -0,0 +1,84 @@
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "SHOULD_UPDATE_TO"
+    properties: DockerScoutPublicImageShouldUpdateToPublicImageTagRelProperties = (
+        DockerScoutPublicImageShouldUpdateToPublicImageTagRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class DockerScoutPublicImageTagSchema(CartographyNodeSchema):
+    label: str = "DockerScoutPublicImageTag"
```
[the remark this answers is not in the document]
jychp: This module has never been deployed before

jychp:
@cubic-dev-ai review this PR again

cartography/intel/docker_scout/recommendation_parser.py
```diff
@@ -0,0 +1,274 @@
...
+
+    field_matches = re.findall(
+        r"^\s*([A-Za-z]+)\s*│\s*(.*?)\s*$", section, flags=re.MULTILINE
+    )
+    fields = {key: value.strip() for key, value in field_matches}
+
+    vulnerabilities = parse_vulnerabilities(fields.get("Vulnerabilities", ""))
+    flavor = fields.get("Flavor")
+    alt_tags_match = re.search(
+        r"supported tag\(s\)\s+(.+?)\.\s+If you want to display recommendations",
```
kunaals: nit: i hit this with a real local `docker scout recommendations --output ... neo4j:5` run. long supported-tag output can wrap across lines, and this regex ends up capturing embedded newlines / `│` prefixes inside the tag text. might be worth normalizing that wrapped block before extracting backtick-delimited tags.

cartography/intel/docker_scout/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+            "Docker Scout sync was configured, but no report files were found in %s.",
+            results_dir,
+        )
+        raise ValueError("No Docker Scout recommendation reports found on disk")
+
+    logger.info("Processing %d local Docker Scout report files", len(report_files))
+
+    synced_count = 0
+    for file_path in report_files:
+        raw_recommendation = Path(file_path).read_text(encoding="utf-8")
```
kunaals: nit: this currently hard-fails the whole sync on a non-utf8 file before we even decide whether it’s a scout report. i’d be inclined to catch decode errors here and skip unreadable files so one stray binary under the path doesn’t abort ingestion.

cartography/intel/docker_scout/recommendation_parser.py
kunaals: more concretely, perhaps we could change from “regex over the whole prose blob” to “extract the block, normalize wrapped lines, then parse tags”

1. find the supported-tags paragraph boundaries
1. strip the left gutter from each line in that paragraph
1. join continuation lines back into plain text
1. extract backtick-delimited tags from the normalized text

PR #2492, opened 2026-03-11
feat(ontology): add PermissionRole, NetworkAccessControl, and DNSZone semantic labels

jychp:
## Summary

- Add three new semantic label categories to the Cartography ontology for unified cross-cloud querying:
  - **PermissionRole**: IAM roles across 7 modules (AWSRole, AWSPermissionSet, AzureRoleDefinition, GCPRole, KeycloakRole, KubernetesRole, KubernetesClusterRole, CloudflareRole, OCIPolicy)
  - **NetworkAccessControl**: Security groups and firewall rules across 3 modules (EC2SecurityGroup, GCPFirewall, AzureNetworkSecurityGroup, AzureFirewall, GCPCloudArmorPolicy)
  - **DNSZone**: DNS zones across 3 modules (AWSDNSZone, GCPDNSZone already had the label; CloudflareZone added)
- Add ontology field mappings (`_ont_name`, `_ont_type`, `_ont_scope`, `_ont_direction`, `_ont_public`) with appropriate special handling per node
- Document all new semantic labels in the ontology schema and add ontology mapping blockquotes to all 17 affected node schema docs

## Test plan

- [x] All pre-commit checks pass (isort, black, flake8, mypy)
- [x] 213 integration/unit tests pass across all affected modules (AWS IAM, EC2 SG, Route53, Identity Center, Azure RBAC/Firewall/Network, GCP IAM/Compute/DNS, Cloudflare, Keycloak, Kubernetes, OCI, ontology, graph querybuilder)
- [ ] Verify ontology `_ont_*` properties are set correctly on nodes during ingestion


cartography/models/ontology/mapping/data/roles.py
```diff
@@ -0,0 +1,205 @@
...
+gcp_mapping = OntologyMapping(
+    module_name="gcp",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="GCPRole",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="title", required=True
+                ),
+                OntologyFieldMapping(ontology_field="type", node_field="role_type"),
```
kunaals: this copies gcp's provider-native values (`basic`/`predefined`/`custom`, `global`/`organization`/`project`) straight into `_ont_type` and `_ont_scope`, but the point of `permissionrole` is cross-platform normalization. as written, queries like `match (r:PermissionRole) where r._ont_type = 'custom' and r._ont_scope = 'project'` will miss gcp roles unless callers know the gcp-specific vocabulary. can we translate these to the shared lowercase values here, or omit them until we have that mapping?

cartography/models/ontology/mapping/data/roles.py
```diff
@@ -0,0 +1,205 @@
...
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="name", required=True
+                ),
+                OntologyFieldMapping(
+                    ontology_field="type",
+                    node_field="",
+                    special_handling="static_value",
+                    extra={"value": "custom"},
+                ),
+                OntologyFieldMapping(
```
kunaals: `ocipolicy` can live at the tenancy level or the compartment level, and the existing model already preserves that distinction via `compartment_id` and the parent rels. setting `_ont_scope = "compartment"` for every policy will mislabel tenancy-level policies and make ontology scope filters inaccurate. i think this should either derive scope from the actual parent or leave `_ont_scope` unset for now.

PR #2494, opened 2026-03-12
fix(okta): use next link presence for users pagination

### LINKED ISSUE #1195: Okta Users: Missing Last Page

raajheshkannaa:
## Summary
- Fix pagination in Okta users module that skips the final page of results (#1195)
- Replace `is_last_page()` SDK method with direct Link header check, matching the approach used by groups and applications modules
- Add unit tests for multi-page and single-page pagination

## Context
PR #1243 (Okta module refactor) was expected to address this but is blocked on an upstream SDK issue (okta/okta-sdk-python#481). This targeted fix resolves the pagination bug independently.

The old code called `paged_users.is_last_page()` from the Okta SDK, which had issues with the last page. The new approach checks `response.links.get("next")` directly, which is safe because `requests.Response.links` always returns an empty dict (never None) when no Link headers are present.

Note: groups.py and applications.py use a slightly different pattern (the `is_last_page()` utility from `utils.py` + chained `.get("next").get("url")`). The approach here avoids the chained `.get()` call that would raise AttributeError without the guard. A follow-up PR could unify all three modules.

## Test plan
- [x] Unit test: multi-page pagination (3 pages, all collected)
- [x] Unit test: single-page results
- [x] Existing transform tests pass (7 tests)
- [x] black, isort, flake8 clean

Fixes #1195

### DIFF
--- a/cartography/intel/okta/users.py
+++ b/cartography/intel/okta/users.py
@@ -45,13 +45,12 @@ def _get_okta_users(user_client: UsersClient) -> List[Dict]:
     user_list: List[Dict] = []
     paged_users = user_client.get_paged_users()
 
-    # TODO: Fix bug, we miss last page :(
     while True:
         user_list.extend(paged_users.result)
         check_rate_limit(paged_users.response)
-        if not paged_users.is_last_page():
-            # Keep on fetching pages of users until the last page
-            paged_users = user_client.get_paged_users(url=paged_users.next_url)
+        next_link = paged_users.response.links.get("next")
+        if next_link:
+            paged_users = user_client.get_paged_users(url=next_link["url"])
         else:
             break
 
--- a/tests/unit/cartography/intel/okta/test_user.py
+++ b/tests/unit/cartography/intel/okta/test_user.py
@@ -1,3 +1,7 @@
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from cartography.intel.okta.users import _get_okta_users
 from cartography.intel.okta.users import transform_okta_user
 from tests.data.okta.users import create_test_user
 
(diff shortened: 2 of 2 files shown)


raajheshkannaa:
This is superseded by #1243, which landed on 2026-08-29.

That refactor replaced the hand-rolled loop in `_get_okta_users` with the shared `collect_paginated` helper in `cartography/intel/okta/common.py`. The helper extends its result list with each page's data *before* checking for a next cursor, and breaks only when `PaginationHelper.extract_next_cursor` comes back empty, so the last page is included. Its docstring reaches the same conclusion this PR did, that the SDK's own `has_next()`/`is_last_page()` helpers can't be relied on.

The branch also conflicts now: `tests/unit/cartography/intel/okta/test_user.py` was deleted in the refactor and the code this PR patches no longer exists.

Happy to close this. Worth noting #1195 is still open, so it should probably be closed as fixed by #1243 rather than left hanging.

Maintainers, let me know if you'd rather I rework it into something else against the new module.

raajheshkannaa:
Closing this out.

#1243 landed on 2026-08-30 and replaced the hand-rolled loop in `_get_okta_users` with the shared `collect_paginated` helper in `cartography/intel/okta/common.py`. That helper extends its result list with each page before it looks for a cursor, and it breaks only once `PaginationHelper.extract_next_cursor` comes back empty, so the final page is included. `users.py` calls it directly now and the code this PR patched is gone.

One loose end: #1195 is still open and #1243 fixes it, so it can probably be closed too.

Issue #2424, opened 2026-02-24
GCP: introduce canonical HttpError classification and align CAI fallback semantics


kunaals:
## Goal
Create a single, well-tested GCP HttpError classification utility and migrate modules to it incrementally while preserving intended behavior.

## Proposed Work
1. Add canonical classifier helper(s) in `cartography/intel/gcp/util.py` (for example, `classify_gcp_http_error`) returning normalized categories such as:
   - `api_disabled`
   - `forbidden`
   - `not_found`
   - `invalid`
   - `transient`
   - `unknown`
2. Add table-driven unit tests covering varied Google error payload shapes and malformed/missing payloads.
3. Migrate GCP modules from direct reason-string checks to the classifier (compute, `cartography/intel/gcp/__init__.py`, and related modules).
4. Align/document CAI fallback decision logic against classifier outputs.
5. Remove any remaining ad-hoc HttpError parsing/branching that duplicates classifier semantics.

## Acceptance Criteria
- No GCP module performs ad-hoc HttpError JSON parsing or scattered reason-string classification.
- CAI fallback triggers and skip conditions are documented and behavior-tested.
- Existing sync behavior remains stable unless an intentional behavior change is documented.
- GCP unit/integration tests continue to pass.

## Notes
This is a follow-up refactor task and should remain separate from feature/modeling PRs for focused review.

kunaals:
PR #2502 covers the first slice of this work:
- adds the canonical GCP `HttpError` classifier
- adds table-driven classifier/retry tests
- migrates DNS, GKE, and Vertex models to the helper

Remaining follow-up for this issue:
- migrate remaining GCP call sites still doing ad hoc classification/parsing, including `compute.py`, `cartography/intel/gcp/__init__.py`, BigQuery, Artifact Registry, and others
- align and document CAI fallback behavior against classifier outputs
- remove remaining duplicated `HttpError` branching once those migrations are complete

Leaving this issue open since PR #2502 is explicitly a partial step, not the full resolution.

Denyme24:
hey @kunaals, if possible, could you please update this as the scope has been changed after this [PR](https://github.com/cartography-cncf/cartography/pull/2530), so it would be better to follow up.
thanks

> PR [#2502](https://github.com/cartography-cncf/cartography/pull/2502) covers the first slice of this work:
> 
> * adds the canonical GCP `HttpError` classifier
> * adds table-driven classifier/retry tests
[...]

PR #2502, opened 2026-03-13
feat(gcp): implement `classify_gcp_http_error` utility and update error handling in DNS, GKE, and Vertex AI modules

Denyme24:
### Type of change
- [X] Refactoring (no functional changes)

### Summary
Adds a canonical GCP `HttpError` classifier and uses it to fix several broken call sites while preserving behavior.

This PR introduces `classify_gcp_http_error(e)` in `cartography/intel/gcp/util.py`, returning normalized categories (`api_disabled`, `forbidden`, `not_found`, `invalid`, `transient`, `unknown`) and reusing existing helpers like `is_api_disabled_error` and `get_error_reason` to avoid duplicated logic.  It then migrates `dns.py`, `gke.py`, and `vertex/models.py` away from ad‑hoc JSON/header parsing to this helper, fixing incorrect field access and `KeyError`/logging bugs while keeping their high‑level “permission error → skip results for this project” behavior unchanged. 

### Related issues or links

- Part of #2424

### How was this tested?

- Added `TestClassifyGcpHttpError` and `TestIsRetryableGcpHttpError` in
  `tests/unit/cartography/intel/gcp/test_util.py` with table-driven unit tests
  covering all six canonical categories (`api_disabled`, `forbidden`,
  `not_found`, `invalid`, `transient`, `unknown`), both standard REST and
  gRPC-transcoded error payload shapes (including quota `RATE_LIMIT_EXCEEDED` /
  `USER_RATE_LIMIT_EXCEEDED`), and malformed/non-JSON bodies to verify the
  "never raises" guarantee.
- Added `tests/unit/cartography/intel/gcp/test_dns.py` to assert that DNS
  helpers return `[]` on `forbidden`/`api_disabled` errors and re-raise on
  unexpected errors, making the current-state cleanup behavior explicit.
- Added `tests/unit/cartography/intel/gcp/test_gke.py` to assert that the GKE
  helper returns `{}` on `forbidden`/`api_disabled` errors and re-raises on
  unexpected errors.

**table-driven unit tests :** 
WhatsApp Image 2026-03-14 at 1 05 05 PM
WhatsApp Image 2026-03-14 at 1 04 34 PM

**targeted module-level tests to make the current-state cleanup behavior explicit for dns/gke**
image

#### General
- [X] I have read the [contributing guidelines](https://cartography-cncf.github.io/cartography/dev/developer-guide.html).
- [X] The linter passes locally (`make lint`).
- [X] I have added/updated tests that prove my fix is effective or my feature works.

#### Proof of functionality

- [] Screenshot showing the graph before and after changes.
- [X] New or updated unit/integration tests.

  - Example:
    ```
    ```


Denyme24:
hey team @kunaals @jychp @achantavy. this PR implements the canonical GCP `HttpError` classifier (`classify_gcp_http_error`) and migrates `dns.py`, `gke.py`, and `vertex/models.py` to use it, addressing part of the work described in #2424 (the remaining module migrations and CAI fallback alignment will be handled in follow-up PRs).

cartography/intel/gcp/util.py
```diff
@@ -25,6 +25,7 @@
 GCP_PERMISSION_DENIED_REASONS = frozenset(
     {"forbidden", "insufficientPermissions", "IAM_PERMISSION_DENIED"}
 )
+GCP_QUOTA_EXCEEDED_REASONS = frozenset({"rateLimitExceeded", "userRateLimitExceeded"})
```
kunaals: this should also handle grpc/errorinfo-style quota reasons like `RATE_LIMIT_EXCEEDED` and `USER_RATE_LIMIT_EXCEEDED`. right now some transient 403 quota failures will still fall through to `forbidden` instead of being retried.

cartography/intel/gcp/dns.py
```diff
@@ -33,19 +34,11 @@ def get_dns_zones(dns: Resource, project_id: str) -> List[Dict]:
             )
         return zones
     except HttpError as e:
-        err = json.loads(e.content.decode("utf-8"))["error"]
-        if (
-            err.get("status", "") == "PERMISSION_DENIED"
-            or err.get("message", "") == "Forbidden"
```
kunaals, line 40: this behavior makes sense if we want current-state semantics, but i think we should say that explicitly in a short comment or test. returning empty results here intentionally allows cleanup to remove previously ingested data when access is lost. this is good so that we dont keep stale data in the graph

cartography/intel/gcp/util.py
Denyme24: @kunaals thanks for pointing this out. i’ve updated `GCP_QUOTA_EXCEEDED_REASONS` to include the gRPC / `ErrorInfo` variants (`RATE_LIMIT_EXCEEDED`, `USER_RATE_LIMIT_EXCEEDED`).

i also updated the tests so both the classifier and the retry helper treat these 403 errors as transient and retryable.

cartography/intel/gcp/dns.py
Denyme24: added comments in `dns.py` and `gke.py` explaining the current-state cleanup behavior, and added tests ensuring DNS/GKE return empty results on `forbidden` / `api_disabled` and re-raise on unexpected errors.

Denyme24:
@kunaals thanks for reviewing my changes. i have made changes as per your feedback/suggestions, PTAL.

PR #2509, opened 2026-03-17
fix(aws) use recommended function to fetch s3 bucket region

### LINKED ISSUE #2508: (aws) s3: get_bucket_location is no longer recommended by boto3
### Expected behavior
The `Region` property for every bucket should be set when the bucket list is being fetched in `get_s3_bucket_list`.

### Actual behavior
The `Region` property for some buckets is set to `None` due to the API `get_bucket_location` returning null as explained in the [note section of boto3 S3 docs](https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/get_bucket_location.html). 

> In a bucket’s home Region, calls to the GetBucketLocation operation are governed by the bucket’s policy. In other Regions, the bucket policy doesn’t apply, which means that cross-account access won’t be authorized. However, calls to the HeadBucket operation always return the bucket’s location through an HTTP response header, whether access to the bucket is authorized or not. Therefore, we recommend using the HeadBucket operation for bucket Region discovery and to avoid using the GetBucketLocation operation.

This can be fixed by calling [HeadBucket](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html) instead and accessing the region from `BucketRegion` key in response.

- Commit hash: 9b19e0054df6ce4523876964dfbe23b973816583

- Python version: 3.10.20

aliahmed58:
### Type of change
- [ ] Documentation update

### Summary
Currently, S3 bucket region is fetched using `get_bucket_location` which is not recommended by AWS Boto3 anymore. This was resulting in some buckets not having their `region` property set since it returned None for some buckets. As stated [here](https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/get_bucket_location.html):
> In a bucket’s home Region, calls to the GetBucketLocation operation are governed by the bucket’s policy. In other Regions, the bucket policy doesn’t apply, which means that cross-account access won’t be authorized. However, calls to the HeadBucket operation always return the bucket’s location through an HTTP response header, whether access to the bucket is authorized or not. Therefore, we recommend using the HeadBucket operation for bucket Region discovery and to avoid using the GetBucketLocation operation.

This is fixed by using the recommended method `head_bucket` to fetch a bucket's region.

### Related issues or links

- Closes #2508

### How was this tested?
Ran the sync before and after making the function change by:
`cartography --selected-modules aws --aws-requested-syncs s3`
Ran all tests by:
`make test`

#### General

#### Proof of functionality

Graph queries on s3 sync with same configuration and data before and after the change.
Before:
image
After:
image

    ```
    INFO:cartography.intel.aws.s3:Syncing S3 for account 'xxxxxxxxx'
    INFO:cartography.graph.statement:Completed aws_s3_details statement #1
    INFO:cartography.graph.job:Finished job aws_s3_details
    INFO:cartography.client.core.tx:Loaded 30 S3Bucket nodes
    INFO:cartography.client.core.tx:Loaded 30 S3Acl nodes
    INFO:cartography.graph.statement:Completed aws_s3acl_analysis statement #1
    INFO:cartography.graph.statement:Completed aws_s3acl_analysis statement #2
    INFO:cartography.graph.statement:Completed aws_s3acl_analysis statement #3
    INFO:cartography.graph.statement:Completed aws_s3acl_analysis statement #4
    INFO:cartography.graph.statement:Completed aws_s3acl_analysis statement #5
    INFO:cartography.graph.job:Finished job aws_s3acl_analysis
    INFO:cartography.client.core.tx:Loaded 19 S3PolicyStatement nodes
    INFO:cartography.graph.statement:Completed S3Bucket statement #1
    INFO:cartography.graph.statement:Completed S3Bucket statement #2
    INFO:cartography.graph.job:Finished job S3Bucket
    INFO:cartography.graph.statement:Completed S3Acl statement #1
    INFO:cartography.graph.statement:Completed S3Acl statement #2
    INFO:cartography.graph.statement:Completed S3Acl statement #3
    INFO:cartography.graph.job:Finished job S3Acl
    INFO:cartography.graph.statement:Completed S3PolicyStatement statement #1
    INFO:cartography.graph.statement:Completed S3PolicyStatement statement #2
    INFO:cartography.graph.statement:Completed S3PolicyStatement statement #3
    INFO:cartography.graph.job:Finished job S3PolicyStatement
    INFO:cartography.intel.aws.s3:Syncing S3 bucket notifications
    INFO:cartography.intel.aws.s3:Loading 0 S3 bucket notifications into Neo4j
    ```


cartography/intel/aws/s3.py
```diff
@@ -77,10 +77,16 @@ def get_s3_bucket_list(boto3_session: boto3.session.Session) -> List[Dict]:
     buckets = client.list_buckets()
     for bucket in buckets["Buckets"]:
         try:
-            bucket["Region"] = client.get_bucket_location(Bucket=bucket["Name"])[
-                "LocationConstraint"
-            ]
+            bucket["Region"] = client.head_bucket(Bucket=bucket["Name"])["BucketRegion"]
```
jychp: blocking: `["BucketRegion"]` raises `KeyError` (not `ClientError`) on botocore <1.31 — which is within the declared minimum (`>=1.18.1`). The `except ClientError` block won't catch it, so S3 sync crashes before any fallback runs.

Use `.get("BucketRegion")` with a fallback to the response header (always present):

```python
resp = client.head_bucket(Bucket=bucket["Name"])
bucket["Region"] = resp.get("BucketRegion") or resp["ResponseMetadata"]["HTTPHeaders"].get("x-amz-bucket-region")
```

cartography/intel/aws/s3.py
jychp, line 1: suggestion: The existing tests mock `get_s3_bucket_list` entirely, so neither the happy path nor the new error-header extraction path (lines 82-89) are exercised. Consider adding a unit test for the 403-with-header fallback to cover this new logic.

jychp, line 1: nitpick: The comment at lines 145-147 still references `get_bucket_location()` — now stale after this PR's switch to `head_bucket`. With `head_bucket`, the region is always returned (including `us-east-1`), so the "sometimes None" caveat no longer applies. Consider updating or removing it.

PR #2513, opened 2026-03-17
feat(semgrep: Secrets): Include secret findings from semgrep

serge-wq:
### Type of change

- [x]  New feature (non-breaking change that adds functionality)
- [ ] Documentation update

### Summary
This PR adds support for ingesting Semgrep Secrets findings into Cartography. It introduces a new `SemgrepSecretsFinding` node schema and the corresponding sync logic to fetch, transform, and load secrets data from the Semgrep API (`/api/v1/deployments/{deploymentId}/secrets`). 

The new node is linked to the `SemgrepDeployment` and optionally to the `GitHubRepository` where the secret was found. This allows users to query and analyze hardcoded secrets discovered by Semgrep across their repositories within the Cartography graph.

### How was this tested?

- Added unit/integration tests in `tests/integration/cartography/intel/semgrep/test_secrets.py` to verify the ingestion logic, node creation, and relationship mapping.
- Validated the schema with mock data in `tests/data/semgrep/secrets.py`.
- Ran the local test suite using `pytest`.
- Ran the semgrep intelmodule locally successfully.

#### General

#### Proof of functionality

3DF770BE-7689-459B-BCEE-8409270C1307_1_201_a

  - Example:
    ```
    ```

### Notes for reviewers
The Semgrep Secrets API uses cursor-based pagination, which is handled in `get_secret_findings`. We chose not to link secrets to the `SemgrepFindingAssistant` node as the API response for secrets does not contain the same level of assistant data as SAST/SCA findings.

### DIFF
--- a/cartography/intel/aws/cloudtrail.py
+++ b/cartography/intel/aws/cloudtrail.py
@@ -6,6 +6,12 @@ from typing import List
 
 import boto3
 import neo4j
+from botocore.exceptions import ClientError
+from botocore.exceptions import ConnectionClosedError
+from botocore.exceptions import ConnectTimeoutError
+from botocore.exceptions import EndpointConnectionError
+from botocore.exceptions import ReadTimeoutError
+from botocore.parsers import ResponseParserError
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
@@ -15,6 +21,32 @@ from cartography.util import aws_handle_regions
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
+
+_RETRYABLE_HTTP_STATUS_CODES = {429, 500, 502, 503, 504}
+
+
+class CloudTrailTransientRegionFailure(Exception):
+    pass
+
+
+def _is_retryable_cloudtrail_error(error: ClientError) -> bool:
+    error_code = error.response.get("Error", {}).get("Code", "")
+    status_code = error.response.get("ResponseMetadata", {}).get("HTTPStatusCode")
+    return (
+        error_code
+        in {
+            "RequestLimitExceeded",
+            "RequestThrottled",
+            "RequestTimeout",
+            "RequestTimeoutException",
+            "ServiceUnavailable",
+            "ServiceUnavailableException",
+            "Throttling",
+            "ThrottlingException",
+            "TooManyRequestsException",
+        }
+        or status_code in _RETRYABLE_HTTP_STATUS_CODES
+    )
 
 
 @timeit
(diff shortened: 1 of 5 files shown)


cartography/models/semgrep/secrets.py
```diff
@@ -0,0 +1,86 @@
...
+    type: PropertyRef = PropertyRef("type", extra_index=True)
+    validation_state: PropertyRef = PropertyRef("validationState", extra_index=True)
+    status: PropertyRef = PropertyRef("status", extra_index=True)
+    finding_path: PropertyRef = PropertyRef("findingPath", extra_index=True)
+    finding_path_url: PropertyRef = PropertyRef("findingPathUrl")
+    ref_url: PropertyRef = PropertyRef("refUrl")
+    mode: PropertyRef = PropertyRef("mode")
+    created_at: PropertyRef = PropertyRef("createdAt")
+    updated_at: PropertyRef = PropertyRef("updatedAt")
+    repository_visibility: PropertyRef = PropertyRef("repositoryVisibility")
```
heryxpc, line 33: Could you add the `findings[].repository.scmType` as a new attribute?

cartography/intel/semgrep/secrets.py
```diff
@@ -0,0 +1,195 @@
...
+
+@timeit
+def load_semgrep_secret_findings(
+    neo4j_session: neo4j.Session,
+    findings: List[Dict[str, Any]],
+    deployment_id: str,
+    update_tag: int,
+) -> None:
+    logger.debug(
+        f"Loading {len(findings)} SemgrepSecretsFinding objects into the graph."
```
achantavy: Use `%d` now. We found out that this helps a lot with the log system

docs/root/modules/semgrep/schema.md
```diff
@@ -89,6 +95,45 @@ Represents a [Semgrep SAST](https://semgrep.dev/docs/semgrep-code/getting-starte
...
+- A SemgrepSecretsFinding belongs to a SemgrepDeployment
+
+    ```
+    (SemgrepDeployment)-[RESOURCE]->(SemgrepSecretsFinding)
+    ```
+
+- A SemgrepSecretsFinding connected to a GitHubRepository (optional)
+
+    ```
+    (SemgrepSecretsFinding)-[FOUND_IN]->(GitHubRepository)
```
achantavy, line 136: Ooh this is exciting

cartography/models/semgrep/secrets.py
```diff
@@ -0,0 +1,86 @@
...
+class SemgrepSecretsFindingToGithubRepoRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:SemgrepSecretsFinding)-[:FOUND_IN]->(:GitHubRepository)
+class SemgrepSecretsFindingToGithubRepoRel(CartographyRelSchema):
+    target_node_label: str = "GitHubRepository"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"fullname": PropertyRef("repositoryName")},
```
heryxpc: I just realized this.
Can you use GitHub id (full url) and Semgrep repository url instead to match?

I'm a bit worried that findings form GitLab with a matching repository name in Semgrep are confused with repositories in GitHub.

We probably need to do the same thing for other Semgrep models :-(

achantavy: Good catch on the fullname - yes definitely should fix.

The target node label here is set to "GitHubRepository" though, so the matcher will only look to join on GitHubRepository nodes. I don't think a broader problem exists

serge-wq: I can work on this in a follow up PR, but good catch.

PR #2526, opened 2026-03-20
refactor(aws): split ECR provenance into its own sync phase

kunaals:
### Type of change

### Summary
This follow-up PR separates ECR inventory, provenance, and layer enrichment into distinct sync phases.

It keeps `ecr` focused on repository/image inventory and manifest relationships, moves attestation parsing and `BUILT_FROM` creation into a new `ecr:provenance` phase, and keeps `ecr:image_layers` focused on layer enrichment only. It also extracts shared ECR manifest/blob fetch helpers into `ecr_shared.py` so the two deep-inspection phases do not depend on each other.

The branch also includes follow-up cleanup fixes for the split:
- preserve stale `HAS_LAYER` cleanup
- clear legacy provenance properties left behind on upgraded graphs

### Related issues or links
- Follow-up to #2525

### How was this tested?
- `uv run pytest tests/integration/cartography/intel/aws/test_ecr.py tests/integration/cartography/intel/aws/test_ecr_provenance.py`
- `uv run pytest tests/integration/cartography/intel/aws/test_ecr_layers.py tests/integration/cartography/intel/aws/test_init.py`

#### General

#### Proof of functionality

### Notes for reviewers
This is intentionally separate from #2525. The smaller PR keeps the production fix narrowly scoped, while this branch carries the larger module-boundary change so it can be reviewed on its own terms.

### DIFF
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -137,7 +137,13 @@ def _sync_one_account(
         if func_name not in requested_syncs_set:
             continue
         # Skip permission relationships and tags for now because they rely on data already being in the graph
-        if func_name == "ecr:image_layers":
+        if func_name == "ecr:provenance":
+            # ecr:provenance uses the standard AWS sync args plus aioboto3_session
+            RESOURCE_FUNCTIONS[func_name](
+                **sync_args,
+                aioboto3_session=aioboto3_session,
+            )
+        elif func_name == "ecr:image_layers":
             # has a different signature than the other functions (aioboto3_session replaces boto3_session)
             RESOURCE_FUNCTIONS[func_name](
                 neo4j_session,
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -12,6 +12,7 @@ from cartography.graph.job import GraphJob
 from cartography.intel.aws.ecr_shared import INDEX_MEDIA_TYPES
 from cartography.intel.container_arch import normalize_architecture
 from cartography.models.aws.ecr.image import ECRImageBaseSchema
+from cartography.models.aws.ecr.image import ECRImageSchema
 from cartography.models.aws.ecr.repository import ECRRepositorySchema
 from cartography.models.aws.ecr.repository_image import ECRRepositoryImageSchema
 from cartography.util import aws_handle_regions
@@ -362,7 +363,9 @@ def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
     GraphJob.from_node_schema(ECRRepositoryImageSchema(), common_job_parameters).run(
         neo4j_session
     )
-    GraphJob.from_node_schema(ECRImageBaseSchema(), common_job_parameters).run(
+    # Use ECRImageSchema (not ECRImageBaseSchema) for cleanup so that
+    # stale HAS_LAYER relationships are also removed when images change.
+    GraphJob.from_node_schema(ECRImageSchema(), common_job_parameters).run(
         neo4j_session
     )
 
--- a/cartography/intel/aws/ecr.py
+++ b/cartography/intel/aws/ecr.py
@@ -9,9 +9,9 @@ import neo4j
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.aws.ecr_shared import INDEX_MEDIA_TYPES
 from cartography.intel.container_arch import normalize_architecture
 from cartography.models.aws.ecr.image import ECRImageBaseSchema
-from cartography.models.aws.ecr.image import ECRImageSchema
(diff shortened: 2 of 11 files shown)


cartography/intel/aws/ecr.py
```diff
@@ -363,12 +363,14 @@ def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
     GraphJob.from_node_schema(ECRRepositoryImageSchema(), common_job_parameters).run(
         neo4j_session
     )
+    # Use ECRImageSchema (not ECRImageBaseSchema) for cleanup so that
+    # stale HAS_LAYER relationships are also removed when images change.
     GraphJob.from_node_schema(ECRImageSchema(), common_job_parameters).run(
```
jychp, line 368: **blocking**: `ECRImageSchema` includes `ECRImageToParentImageRel` (BUILT_FROM) in `other_relationships`. The framework's `build_cleanup_queries` generates `DELETE r WHERE r.lastupdated <> $UPDATE_TAG` for every relationship in the schema. Since `ecr` never creates BUILT_FROM edges, running `--syncs ecr` without `ecr:provenance` will silently delete all previously ingested provenance relationships.

Consider a dedicated cleanup schema that excludes BUILT_FROM, or remove BUILT_FROM from `ECRImageSchema.other_relationships` and rely solely on `ecr_provenance.cleanup()` (which already handles it).

cartography/intel/aws/ecr_provenance.py
```diff
@@ -0,0 +1,424 @@
...
+    r.from_attestation = item["from_attestation"],
+    r.parent_image_uri = item["parent_image_uri"],
+    r.confidence = item["confidence"]
+"""
+
+
+STALE_ECR_IMAGE_PROVENANCE_QUERY = """
+MATCH (:AWSAccount {id: $AWS_ID})-[:RESOURCE]->(img:ECRImage)
+WHERE img.region = $Region
+  AND img.source_uri IS NOT NULL
```
jychp, line 64: **blocking**: The stale-provenance cleanup only matches `img.source_uri IS NOT NULL`, but `_extract_provenance_from_attestation()` can populate `invocation_uri` (line 194-198, via `runDetails.builder.id`) without setting `source_uri`. Those images will never have stale invocation fields cleared.

Suggestion: `AND (img.source_uri IS NOT NULL OR img.invocation_uri IS NOT NULL)`

cartography/intel/aws/ecr_provenance.py
```diff
@@ -0,0 +1,424 @@
+import asyncio
+import logging
+from typing import Any
+from typing import Dict
```
jychp, line 4: **suggestion**: Use Python 3.9+ builtin generics (`dict`, `str | None`) instead of `Dict` and `Optional` from `typing`. Affects lines 4-5, 87, 113, 271.

cartography/intel/aws/ecr_provenance.py
```diff
@@ -0,0 +1,424 @@
...
+def sync(
+    neo4j_session: neo4j.Session,
+    boto3_session: boto3.session.Session,
+    regions: list[str],
+    current_aws_account_id: str,
+    update_tag: int,
+    common_job_parameters: dict[str, Any],
+    aioboto3_session: aioboto3.Session | None = None,
+) -> None:
+    del common_job_parameters
```
jychp, line 395: **nitpick**: `del common_job_parameters` to mark the parameter as unused is unconventional. Consider a leading underscore on the parameter name or a `# noqa` comment instead. (Though the standard sync signature may require the exact name.)

cartography/intel/aws/__init__.py
```diff
@@ -137,7 +137,13 @@ def _sync_one_account(
         if func_name not in requested_syncs_set:
             continue
         # Skip permission relationships and tags for now because they rely on data already being in the graph
-        if func_name == "ecr:image_layers":
+        if func_name == "ecr:provenance":
```
jychp, line 140: **suggestion**: `ecr:provenance` and `ecr:image_layers` are missing from the `module_dependencies` dict (line 109-122). Running `--syncs ecr:provenance` without `ecr` produces a silent no-op (no images in graph to attach provenance to). Adding `"ecr:provenance": ["ecr"], "ecr:image_layers": ["ecr"]` would warn users.

cartography/intel/aws/ecr_image_layers.py
jychp, line 1: **suggestion**: Multiple `logger.info(f"...")` calls throughout this file use f-strings that are eagerly evaluated regardless of log level. Use `logger.info("... %d ...", len(x))` %-formatting for consistency with the rest of the codebase. (Applies to lines 319, 566, 619, 622, 731, 736, 743, 772.)

**nitpick**: The `if repo_images_list:` guard at line 742 is redundant — the `continue` at line 739 already guarantees the list is non-empty at that point.

jychp (changes requested):

PR #2538, opened 2026-03-20
feat(cve_metadata): add CVE metadata enrichment intel module

jychp:
### Type of change
- [ ] Documentation update

### Summary

New `cve_metadata` intel module that enriches existing CVE nodes in the graph with metadata from external sources:

- **NVD**: CVSS scores, descriptions, references, weaknesses, and CISA KEV (Known Exploited Vulnerabilities) data from the NIST NVD API v2.0
- **EPSS**: Exploit Prediction Scoring System scores from FIRST.org

Unlike the existing `cve` module which imports all CVEs from NVD, this module only fetches metadata for CVEs already present in the graph (created by other modules like CrowdStrike, Semgrep, SentinelOne, etc.).

Key design points:
- Creates `CVEMetadata` nodes linked to `CVE` via `ENRICHES` relationship
- Extensible source architecture: each source (nvd, epss) is a separate module
- CLI option `--cve-metadata-src` to select which sources to enable (all by default)
- NVD fetching uses date-range batching to efficiently retrieve data
- EPSS fetching uses batch queries (100 CVEs per request)
- Marks the existing `cve` module as deprecated (following the gsuite pattern)

### Related issues or links

- Replaces the obsolete `cve` module with a targeted enrichment approach

### Breaking changes

None. The existing `cve` module is deprecated but still functional.

### How was this tested?

- 11 unit tests covering NVD transform (filtering, CVSS extraction, KEV fields, weaknesses, references) and EPSS (fetching, batching, merging)
- 2 integration tests covering graph queries and full sync pipeline (CVEMetadata node creation, relationships, properties)
- All pre-commit checks pass (isort, black, flake8, mypy)

#### General

#### Proof of functionality

### Notes for reviewers

- The `cve` module deprecation follows the exact same pattern used for `gsuite` → `googleworkspace`
- CISA KEV data comes embedded in NVD API v2.0 responses, so no separate CISA API call is needed
- The 2-year lookback window for NVD fetching is a pragmatic default; it could be made configurable later


jychp:
carbon (1)
Screenshot 2026-03-25 at 2 01 52 PM

cartography/intel/cve_metadata/nvd.py
```diff
@@ -0,0 +1,184 @@
...
+        if len(parts) >= 2:
+            years.add(parts[1])
+    return years
+
+
+@timeit
+def _download_nvd_feed(http_session: Session, year: str) -> dict[Any, Any]:
+    """Download and extract a yearly NVD JSON feed zip into a temp directory."""
+    url = f"{NVD_FEED_BASE_URL}/nvdcve-2.0-{year}.json.zip"
+    logger.debug("Downloading NVD feed for year %s from %s", year, url)
```
heryxpc, line 59: The NVD website flags that the best preferred approach is using the 2.0 NVD API. The data is fresher and allows to search based on CPE (product/package).
For such case, it's recommended to use an API key, otherwise the calls will be ratelimited fast.

cartography/intel/cve_metadata/nvd.py
heryxpc, line 1: If the approach is to ingest only the CVEs already found in the graph, I would recommend using the API for sure, as it does not download in bulk and ingest not found CVEs.

cartography/intel/cve_metadata/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+    :param config: A cartography.config object
+    """
+    sources = set(config.cve_metadata_src) if config.cve_metadata_src else ALL_SOURCES
+    invalid = sources - ALL_SOURCES
+    if invalid:
+        raise ValueError(
+            f"Invalid CVE metadata sources: {invalid}. Valid sources: {ALL_SOURCES}",
+        )
+
+    # Step 1: Get all CVE IDs from the graph — this is the authoritative list
```
kunaals, line 96: i think this needs a cleanup path. right now we load cvemetadata but never run the normal schema cleanup, so if a cve disappears from the graph the old :CVEMetadata node sticks around forever. the early return when there are no :CVE nodes makes that worse because we skip the module entirely and leave previous metadata behind. can we add the usual cleanup flow here and cover it with a second-sync regression test?

cartography/intel/cve_metadata/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+logger = logging.getLogger(__name__)
+stat_handler = get_stats_client(__name__)
+
+CVE_METADATA_FEED_ID = "CVE_METADATA"
+ALL_SOURCES = {"nvd", "epss"}
+
+
+@timeit
+def get_cve_ids_from_graph(neo4j_session: neo4j.Session) -> list[str]:
```
kunaals, line 30: i think this is too narrow right now. this module only reads and matches cves via cve_id, but not every existing :CVE producer sets that field today. for example crowdstrike's cve schema keys the node by id, so those nodes won't be enriched even though the docs call out crowdstrike as a source. can we either support both id and cve_id here or narrow the docs and tests to the actually supported shape?

kunaals (changes requested):

kunaals (approved):

PR #2540, opened 2026-03-24
fix(gcp): retry cleanup and graph job writes

kunaals:
### Type of change

### Summary
This PR hardens GCP sync cleanup and graph-job write paths against transient Neo4j write failures.

The branch now does three things:
- routes `GraphStatement` writes through `execute_write_with_retry()` so cleanup and analysis jobs retry transient write/service-availability failures instead of aborting a project sync,
- switches the GCP label cleanup query to `run_write_query()` so that the shared `GCPLabel` cleanup path gets the same retry behavior,
- adds unit coverage for both retry paths.

I also merged the latest `master` into this branch and audited the merged GCP code for similar gaps. After the merge, the only remaining direct `neo4j_session.run(...)` in the relevant area is the intentional autocommit index creation path in `cartography/client/core/tx.py`; there are no other GCP module-local write paths bypassing the retry helpers.

### Breaking changes
None.

### How was this tested?
- Focused live validation from source against live GCP data plus local Neo4j:
  - loaded a real GCP organization and one real project into a fresh local Neo4j,
  - fetched live zonal instance inventory for one redacted GCP project,
  - ran `load_gcp_instances(...)`, `sync_labels(...)`, `cleanup_gcp_instances(...)`, and `run_scoped_analysis_job('gcp_compute_exposure.json', ...)`,
  - confirmed the run completed without errors and produced the expected redacted counts below.

Redacted live-validation excerpt:
```text
INFO:validation:Fetched live zonal GCP instance inventory for [REDACTED_PROJECT_1] from us-east1-b.
INFO:validation:Transformed 1 instances and 2 candidate label rows before ingestion.
Result counts: project_count=1, instance_count=1, instance_label_count=2, firewall_count=0
```

#### General

#### Proof of functionality

- [x] Included Cartography sync logs from a real environment demonstrating successful synchronization of the new/modified entity. Logs are redacted to avoid exposing project identifiers.

### Notes for reviewers
- The GCP-specific change remains intentionally narrow. `GCPLabel` is shared across multiple resource schemas, so its cleanup still needs a scoped manual query rather than a plain `GraphJob.from_node_schema(...)` cleanup.

### DIFF
--- a/cartography/util.py
+++ b/cartography/util.py
@@ -39,10 +39,6 @@ from cartography.stats import ScopedStatsClient
 
 logger = logging.getLogger(__name__)
 
-# Explicit re-exports preserve the long-standing util API after moving these helpers.
-backoff_handler = helpers.backoff_handler
-batch = helpers.batch
-
 
 def is_service_control_policy_explicit_deny(
     error: botocore.exceptions.ClientError,
@@ -63,6 +59,17 @@ def is_service_control_policy_explicit_deny(
 STATUS_SUCCESS = 0
 STATUS_FAILURE = 1
 DEFAULT_MAX_PAGES = 10000
+
+
+def backoff_handler(details: Dict) -> None:
+    """
+    Compatibility wrapper for cartography.helpers.backoff_handler.
+
+    Internal callers should import this helper from cartography.helpers.
+    This wrapper preserves the long-standing cartography.util API for
+    external callers.
+    """
+    helpers.backoff_handler(details)
 
 
 def run_analysis_job(
@@ -865,6 +872,20 @@ def camel_to_snake(name: str) -> str:
         The converted snake_case string.
     """
     return re.sub(r"(?<!^)(?=[A-Z])", "_", name).lower()
+
+
+def batch(
+    items: Iterable,
+    size: int = helpers.DEFAULT_BATCH_SIZE,
+) -> Iterable[List[Any]]:
+    """
+    Compatibility wrapper for cartography.helpers.batch.
+
+    Internal callers should import this helper from cartography.helpers.
+    This wrapper preserves the long-standing cartography.util API for
+    external callers.
+    """
+    return helpers.batch(items, size)
 
 
 def is_throttling_exception(exc: Exception) -> bool:
--- a/cartography/intel/aws/inspector.py
+++ b/cartography/intel/aws/inspector.py
@@ -13,6 +13,7 @@ import neo4j
 from cartography.client.core.tx import load
 from cartography.client.core.tx import load_matchlinks
 from cartography.graph.job import GraphJob
+from cartography.helpers import batch
 from cartography.intel.aws.util.botocore_config import create_boto3_client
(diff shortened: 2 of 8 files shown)


cartography/util.py
```diff
@@ -108,6 +106,9 @@ def run_analysis_job(
         The job file must be a valid JSON file containing GraphJob-compatible
         query definitions.
     """
+    from cartography.graph.job import GraphJob
+    from cartography.graph.statement import get_job_shortname
```
kunaals: we moved those imports into the helper functions to break a circular import after hoisting `execute_write_with_retry` to the top of statement.py: statement -> tx -> util -> graph.job/graph.statement -> statement. keeping `graphjob` and `get_job_shortname` lazy in util.py lets statement.py import cleanly at module load time without changing runtime behavior

tests/unit/cartography/data/jobs/test_statement.py
```diff
@@ -15,3 +18,36 @@ def test_create_from_json():
     assert statement.parent_job_name == "my_job_name"
     assert statement.query == "Query goes here"
     assert statement.parent_job_sequence_num == 1
+
+
+@patch("cartography.graph.statement.execute_write_with_retry")
```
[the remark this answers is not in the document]
kunaals: stale now. those statement helper-call tests were removed from the branch.

tests/unit/cartography/test_util.py
```diff
@@ -22,30 +22,33 @@
 
 
 def test_run_analysis_job_default_package(mocker):
-    mocker.patch("cartography.util.GraphJob")
+    run_from_json_mock = mocker.patch("cartography.graph.job.GraphJob.run_from_json")
```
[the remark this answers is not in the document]
kunaals: good catch. switched these to assert package selection plus session execution instead of graphjob mocks.

tests/unit/cartography/intel/gcp/test_labels.py
```diff
@@ -41,3 +45,28 @@ def test_get_labels_unknown_resource_type():
     """
     labels = get_labels([{"id": "x", "labels": {"a": "b"}}], "unknown_type")
     assert labels == []
+
```
shyammukund, line 48: minor: In all these tests it seems like the non-iterative path is tested but maybe there should be a test with an iterative path as well. However, this does seem to be covered in the integration test so maybe not necessary. Up to you.

cartography/util.py
achantavy: No way around this? Open to doing a broader refactor to avoid this, looks nasty, esp in a core component

kunaals: will look into it

kunaals: broke into helpers.py and util.py to avoid the circular dep

tests/unit/cartography/intel/gcp/test_labels.py
kunaals: made the test addition 👍

PR #2546, opened 2026-03-25
refactor(ontology): convert device hostname rels to conditional matchlinks

jychp:
## Summary
- Replace hostname-based `other_relationships` with conditional matchlinks in the Device ontology
- Serial_number remains the primary matching strategy via `other_relationships`
- Hostname matchlinks only run when: (1) serial_number links confirm module data is in the graph, and (2) target hostnames are unique (no duplicates)
- Modules without serial_number support (Duo, Bigfix) only check hostname uniqueness

## Motivation
Hostname matching can produce incorrect OBSERVED_AS links when multiple devices share the same hostname across sources. By making hostname a conditional fallback, we ensure it's only used when it's reliable (unique hostnames) and when serial_number-based matching has already validated the module's presence in the graph.

## Test plan
- [x] All 13 existing device-related tests pass
- [ ] Verify hostname matchlinks are skipped when target nodes have duplicate hostnames
- [ ] Verify hostname matchlinks run correctly as fallback for devices without serial_number matches
- [ ] Verify stale hostname matchlink relationships are cleaned up

### DIFF
--- a/cartography/intel/ontology/devices.py
+++ b/cartography/intel/ontology/devices.py
@@ -119,7 +119,7 @@ def _run_hostname_matchlinks(
     """
     device_hostnames = _get_device_hostnames(neo4j_session, update_tag)
 
-    for target_label, hostname_field, matchlink, _has_serial_rel in HOSTNAME_MATCHLINKS:
+    for target_label, hostname_field, matchlink in HOSTNAME_MATCHLINKS:
         if not device_hostnames:
             continue
 
@@ -151,7 +151,7 @@ def cleanup(
         neo4j_session,
     )
     # Clean up stale hostname matchlink relationships
-    for _, _, matchlink, _ in HOSTNAME_MATCHLINKS:
+    for _, _, matchlink in HOSTNAME_MATCHLINKS:
         GraphJob.from_matchlink(
             matchlink,
             MATCHLINK_SUB_RESOURCE_LABEL,
--- a/cartography/models/ontology/device.py
+++ b/cartography/models/ontology/device.py
@@ -276,24 +276,22 @@ class DeviceToGoogleWorkspaceDeviceHostnameMatchLink(CartographyRelSchema):
 
 
 # Configuration for hostname matchlinks used by the intel module.
-# Each tuple: (target_label, target_hostname_field, matchlink_schema, has_serial_rel)
-HOSTNAME_MATCHLINKS: list[tuple[str, str, CartographyRelSchema, bool]] = [
-    ("CrowdstrikeHost", "hostname", DeviceToCrowdstrikeHostHostnameMatchLink(), True),
-    ("KandjiDevice", "device_name", DeviceToKandjiDeviceHostnameMatchLink(), True),
-    ("SnipeitAsset", "name", DeviceToSnipeitAssetHostnameMatchLink(), True),
-    ("TailscaleDevice", "hostname", DeviceToTailscaleDeviceHostnameMatchLink(), True),
+# Each tuple: (target_label, target_hostname_field, matchlink_schema)
+HOSTNAME_MATCHLINKS: list[tuple[str, str, CartographyRelSchema]] = [
+    ("CrowdstrikeHost", "hostname", DeviceToCrowdstrikeHostHostnameMatchLink()),
+    ("KandjiDevice", "device_name", DeviceToKandjiDeviceHostnameMatchLink()),
+    ("SnipeitAsset", "name", DeviceToSnipeitAssetHostnameMatchLink()),
+    ("TailscaleDevice", "hostname", DeviceToTailscaleDeviceHostnameMatchLink()),
     (
         "GoogleWorkspaceDevice",
         "hostname",
         DeviceToGoogleWorkspaceDeviceHostnameMatchLink(),
-        True,
     ),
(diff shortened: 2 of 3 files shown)


cartography/models/ontology/device.py
```diff
@@ -29,188 +31,253 @@ class DeviceToNodeRelProperties(CartographyRelProperties):
...
-class DeviceToBigfixComputerRel(CartographyRelSchema):
-    target_node_label: str = "BigfixComputer"
-    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
-        {"computername": PropertyRef("hostname")},
+@dataclass(frozen=True)
+class DeviceSchema(CartographyNodeSchema):
+    label: str = "Device"
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Ontology"])
+    properties: DeviceNodeProperties = DeviceNodeProperties()
+    scoped_cleanup: bool = False
```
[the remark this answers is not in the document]
jychp: Device are ontology nodes, not scopped to a tenant

tests/unit/cartography/graph/test_querybuilder_indexcreation.py
```diff
@@ -30,4 +30,9 @@ def test_build_create_index_queries_for_emr():
         "CREATE INDEX IF NOT EXISTS FOR (n:AWSAccount) ON (n.id);",
         "CREATE INDEX IF NOT EXISTS FOR (n:EMRCluster) ON (n.arn);",
         "CREATE INDEX IF NOT EXISTS FOR (n:ComputeCluster) ON (n.id);",
+        # _ont_ field indexes on the ComputeCluster semantic label
```
kunaals: do we need this one off when we're making indexes for ontology modules programmatically now?

cartography/intel/ontology/devices.py
```diff
@@ -27,6 +32,7 @@ def sync(
         update_tag,
     )
     link_ontology_nodes(neo4j_session, "devices", update_tag)
+    _run_hostname_matchlinks(neo4j_session, update_tag)
```
kunaals, line 34: could this ordering mean link_ontology_nodes() is deriving `owns` before these hostname-based `observed_as` rels exist? i’m also wondering whether stale `observed_as `rels from a previous run could still be picked up here since cleanup happens later.

cartography/intel/ontology/devices.py
```diff
@@ -44,6 +50,116 @@ def load_devices(
...
+    3. Hostnames on Device nodes are unique (no duplicates on source side)
+    """
+    if has_serial_rel:
+        result = neo4j_session.run(
+            f"MATCH (:Device)-[r:OBSERVED_AS]->(:{target_label}) "
+            "WHERE r.lastupdated = $update_tag "
+            "RETURN count(r) > 0 as has_links",
+            update_tag=update_tag,
+        ).single()
+        if not result or not result["has_links"]:
```
kunaals: i may be misreading this, but does this make hostname matching supplement-only rather than fallback for serial-capable providers? if serial matching yields 0 observed_as rels in a run, it looks like hostname matching gets skipped entirely even when hostnames are unique.

kunaals (commented):
left a few questions inline on the hostname matchlink flow and the new indexing logic

cartography/intel/ontology/devices.py
jychp: good catch, logic flaw

cartography/intel/ontology/devices.py
jychp: good catch, this is a more global issue will scope link creation to lastupdated tag

tests/unit/cartography/graph/test_querybuilder_indexcreation.py
jychp: good catch, kept the coverage for programmatic semantic-label indexing but removed the overly specific exact-set assertiom

cartography/models/ontology/device.py
```diff
@@ -29,188 +31,269 @@ class DeviceToNodeRelProperties(CartographyRelProperties):
...
+        {"hostname": PropertyRef("hostname")},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "OBSERVED_AS"
+    properties: DeviceHostnameMatchLinkProperties = DeviceHostnameMatchLinkProperties()
+
+
+# Configuration for hostname matchlinks used by the intel module.
+# Each tuple: (target_label, target_hostname_field, matchlink_schema, has_serial_rel)
+HOSTNAME_MATCHLINKS: list[tuple[str, str, CartographyRelSchema, bool]] = [
```
kunaals: looks good now. one small cleanup: since the serial gate is gone, do we still need the 4th tuple field here?

docs/root/modules/ontology/schema.md
```diff
@@ -245,7 +245,6 @@ It generalizes concepts like EC2 Instances, DigitalOcean Droplets, and Scaleway
 
 | Field | Description |
 |-------|-------------|
-| _ont_id | The unique identifier for the instance. |
```
kunaals, line 248: worth calling this out somewhere user-facing too since anyone querying on `_ont_id` will need to switch.

kunaals: eh actually this is so new that we don't need to. i was originally thinking in release notes but i think its not worth.

PR #2553, opened 2026-03-27
fix(aws): lower retry attempts and harden transient failures

kunaals:
### Type of change

### Summary
This refreshes the stacked AWS transient-failure hardening branch on top of current `master`.

It keeps the region or sub-resource skip behavior for API Gateway v2, Lambda, and CloudTrail management events when transient AWS failures exhaust retries, lowers the shared AWS botocore retry budget from 10 attempts to 3, and moves CloudTrail trail and management-event clients onto a standard retry profile so unhealthy regions fail and skip faster. It also sanitizes the GitHub WebIdentity test fixtures to use generic example repository names.

### Related issues or links
- follow up on  #2550

#### General

#### Proof of functionality

### Notes for reviewers
- This merge resolves the branch on top of current `master`.
- Shared AWS client defaults now use `max_attempts=3`.
- CloudTrail and CloudTrail management-event clients now use standard retries so transient regional failures are skipped faster instead of spending as long in retry backoff.


cartography/intel/aws/lambda_function.py
```diff
@@ -284,17 +430,27 @@ def sync_aliases(
...
+        try:
+            aliases = get_function_aliases(lambda_function, client)
+        except LambdaSubResourceTransientFailure as error:
+            cleanup_safe = False
+            logger.warning(
+                "Skipping Lambda aliases for function %s after transient alias fetch failure: %s",
+                function_arn,
+                error,
+            )
+            continue
```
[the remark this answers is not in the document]
kunaals: keeping this as-is because the point of this pr is to preserve last-known-good subresource state for transient lambda failures instead of failing the whole regional sync.

cartography/intel/aws/apigatewayv2.py
```diff
@@ -13,6 +19,40 @@
...
+            "ServiceUnavailable",
+            "ServiceUnavailableException",
+            "Throttling",
+            "ThrottlingException",
+            "TooManyRequestsException",
+        }
+        or status_code in _RETRYABLE_HTTP_STATUS_CODES
+        or (
+            error_code == "AuthorizerConfigurationException"
+            and "internal server error" in error_message.lower()
```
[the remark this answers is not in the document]
kunaals: good catch. fixed this.

kunaals:
@cubic-dev-ai can you review again?

cartography/intel/aws/apigatewayv2.py
```diff
@@ -23,8 +63,25 @@ def get_apigatewayv2_apis(
...
+    except (
+        ConnectionClosedError,
+        ConnectTimeoutError,
+        EndpointConnectionError,
+        ReadTimeoutError,
+        ResponseParserError,
+    ) as error:
+        raise APIGatewayV2TransientRegionFailure(
+            "Encountered a transient regional API Gateway v2 endpoint failure while calling GetApis"
+        ) from error
```
jychp, line 82: This block swallows `EndpointConnectionError` that `aws_handle_regions` intentionally converts to an empty result for regions where API Gateway v2 is not available. Re-raising it as `APIGatewayV2TransientRegionFailure` means any unsupported/opt-in region now marks the whole account cleanup-unsafe, so stale `APIGatewayV2API` nodes can never be removed on that run. Let `EndpointConnectionError` (and arguably `ConnectTimeoutError` / `ReadTimeoutError`) fall through to `aws_handle_regions` instead of reclassifying them as transient.

cartography/intel/aws/lambda_function.py
```diff
@@ -25,19 +32,76 @@
...
+    except (
+        ConnectionClosedError,
+        ConnectTimeoutError,
+        EndpointConnectionError,
+        ReadTimeoutError,
+        ResponseParserError,
+    ) as error:
+        raise LambdaTransientRegionFailure(
+            "Encountered a transient regional Lambda endpoint failure while calling ListFunctions"
+        ) from error
```
jychp, line 103: Same concern as apigatewayv2: `aws_handle_regions` already turns `EndpointConnectionError` / `ConnectTimeoutError` / `ReadTimeoutError` into a regional skip when the service is not available in a partition/region. Reclassifying them here as `LambdaTransientRegionFailure` means an account that scans a partition without a Lambda endpoint will permanently disable Lambda cleanup, leaving stale functions/aliases/layers in the graph even though the absence of the service is permanent rather than transient. The same issue exists in `get_lambda_image_uris` at lines 149–158.

cartography/intel/aws/lambda_function.py
```diff
@@ -469,4 +709,16 @@ def sync(
...
+        common_job_parameters,
+        aliases_cleanup_safe=aliases_cleanup_safe,
+        event_source_mappings_cleanup_safe=event_source_mappings_cleanup_safe,
+        layers_cleanup_safe=layers_cleanup_safe,
+        functions_cleanup_safe=(
+            functions_cleanup_safe
+            and aliases_cleanup_safe
+            and event_source_mappings_cleanup_safe
+            and layers_cleanup_safe
+        ),
```
jychp: Chaining `functions_cleanup_safe and aliases_cleanup_safe and event_source_mappings_cleanup_safe and layers_cleanup_safe` means a single transient `list_aliases` or `list_event_source_mappings` failure on one function will indefinitely preserve every deleted `AWSLambda` node, even though `list_functions` already returned a complete inventory for every region. Only the sub-resource cleanups need to be held back in that scenario — function cleanup should depend on its own safety flag.

cartography/intel/aws/lambda_function.py
```diff
@@ -226,10 +350,9 @@ def get_lambda_permissions(
                 }
         except client.exceptions.ResourceNotFoundException:
             logger.debug(f"No policy found for Lambda function {function_name}")
-            pass
-        except Exception as e:
+        except Exception as error:
             logger.warning(
-                f"Error getting policy for Lambda function {function_name}: {e}"
+                f"Error getting policy for Lambda function {function_name}: {error}"
             )
```
jychp, line 384: `get_lambda_permissions` still catches every exception from `get_policy()` and logs a warning, leaving `AnonymousAccess` as `None` / `[]`. When the policy lookup fails transiently after retries are exhausted, that silently overwrites a previously correct "anonymous access = True" value with a false negative — the exact situation this PR is otherwise hardening against. Consider treating retry-exhausted `ClientError`s the same way as the sub-resource fetchers (raise a transient exception and mark functions cleanup unsafe).

cartography/intel/aws/util/botocore_config.py
```diff
@@ -1,8 +1,50 @@
...
+READ_TIMEOUT_ENV_VAR = "CARTOGRAPHY_AWS_READ_TIMEOUT"
+LAMBDA_READ_TIMEOUT = 30
+LAMBDA_MAX_ATTEMPTS = 2
+
+
+def _get_int_env(name: str, default: int) -> int:
+    raw_value = os.getenv(name)
+    if raw_value in (None, ""):
+        return default
+    assert raw_value is not None
```
jychp: Nit: `assert raw_value is not None` is redundant right after `if raw_value in (None, ""): return default`. Same at line 39 in `_get_retry_mode_env`. Can drop it or leave a `# type: ignore` if it's purely for mypy.

cartography/intel/aws/util/botocore_config.py
```diff
@@ -1,8 +1,50 @@
...
 from typing import Any
 
 import botocore.config
 
+DEFAULT_READ_TIMEOUT = 120
+DEFAULT_MAX_ATTEMPTS = 3
+DEFAULT_RETRY_MODE = "standard"
+VALID_RETRY_MODES = frozenset({"adaptive", "legacy", "standard"})
+
+RETRY_MODE_ENV_VAR = "CARTOGRAPHY_AWS_RETRY_MODE"
```
jychp, line 12: I like using varenv for configuration but's it's a total change from the actual tool philosophy where everything is a cli parameter.

jychp (changes requested):

cartography/intel/aws/apigatewayv2.py
kunaals: good catch. i changed this locally so endpointconnectionerror falls back to aws_handle_regions again, since that can just mean the service is not available in that region or partition. we still keep the cleanup-unsafe path for the actual retry-exhausted and timeout-style failures.

cartography/intel/aws/util/botocore_config.py
kunaals: yep, removed.

cartography/intel/aws/lambda_function.py
kunaals: agree. i changed this locally so alias and event source mapping transient failures only block cleanup for those subresources. function cleanup now depends on the function inventory path being safe, not on alias or esm fetches succeeding.

cartography/intel/aws/lambda_function.py
kunaals: agree. i tightened get_lambda_permissions locally so retry-exhausted and transport failures now raise a transient region failure instead of quietly writing incomplete anonymous access data. resource not found still stays a normal skip.

cartography/intel/aws/lambda_function.py
kunaals: yeah, i agree on the top-level regional lambda path. i changed get_lambda_data locally to let endpointconnectionerror fall through to aws_handle_regions again so unsupported endpoints still skip cleanly. i left the cleanup-unsafe handling in place for the ambiguous retry-exhausted and timeout cases, and for the later per-function image metadata calls.

cartography/intel/aws/util/botocore_config.py
kunaals: true, i think this is the right move here though, we can make this more commonplace. this is for power users

jychp (approved):
LGTM!

PR #2582, opened 2026-04-04
fix(aws): match S3 object-scoped GetObject resources to bucket relationships

### LINKED ISSUE #1639: AWS permission relationships: S3 GetObject does not work if `resource` refers to path under bucket

NikkeTryHard:
### Type of change
- [ ] Documentation update

### Summary
Fix AWS permission relationship evaluation for S3 object-scoped `s3:GetObject` resources.

Cartography models `CAN_READ` on `S3Bucket` nodes, but IAM policies often express `s3:GetObject` access on object ARNs such as `arn:aws:s3:::bucket/*` or `arn:aws:s3:::bucket/prefix/*`. Before this change, those resource clauses were evaluated directly against the bucket ARN `arn:aws:s3:::bucket`, so valid object read access did not produce the expected bucket relationship.

This change keeps the fix local to `cartography.intel.aws.permission_relationships` by projecting S3 object-scoped `resource` and `notresource` clauses onto the parent bucket ARN during permission evaluation. It also adds focused regressions for both the direct evaluator and the IAM integration path.

### Related issues or links
- Fixes #1639
- https://github.com/cartography-cncf/cartography/issues/1639

### Breaking changes
None.

### How was this tested?
- Direct reproducer: verified that `evaluate_policy_for_permissions()` now treats `arn:aws:s3:::bucket/*` and `arn:aws:s3:::bucket/prefix/*` as matching the parent `arn:aws:s3:::bucket` for `s3:GetObject`.
- `uv run pytest tests/unit/cartography/intel/aws/test_permission_relationships.py -q`
- `uv run pytest tests/integration/cartography/intel/aws/test_permission_relationships.py -q`
- `make test_lint`
- `make test_unit`
- `make test_integration`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
This is intentionally narrow.

The patch only changes AWS permission relationship matching for S3 bucket relationships when IAM statements use object-scoped resource forms. It does not change schema, graph shape, or unrelated permission evaluation behavior.

I also mirrored the S3 object-to-bucket projection for `NotResource` so the bucket-level approximation stays consistent for both allow and deny evaluation.

### DIFF
--- a/cartography/intel/aws/permission_relationships.py
+++ b/cartography/intel/aws/permission_relationships.py
@@ -20,8 +20,10 @@ from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
 
+S3_BUCKET_ARN_RE = re.compile(r"^(arn:[^:]+:s3:::)([^/]+)$", flags=re.IGNORECASE)
 
-def evaluate_clause(clause: str, match: str) -> bool:
+
+def evaluate_clause(clause: str | Pattern, match: str) -> bool:
     """Evaluates the a clause in IAM. Clauses can be AWS [not]actions and [not]resources
 
     Arguments:
@@ -35,6 +37,33 @@ def evaluate_clause(clause: str, match: str) -> bool:
     """
     result = compile_regex(clause).fullmatch(match)
     return result is not None
+
+
+def evaluate_resource_clause(clause: str | Pattern, resource_arn: str) -> bool:
+    if evaluate_clause(clause, resource_arn):
+        return True
+
+    if not S3_BUCKET_ARN_RE.fullmatch(resource_arn):
+        return False
+
+    clause_pattern = (
+        clause.pattern
+        if isinstance(clause, re.Pattern)
+        else compile_regex(clause).pattern
+    )
+    s3_arn_index = clause_pattern.lower().find("s3:::")
+    if s3_arn_index == -1:
+        return False
+
+    object_suffix_index = clause_pattern.find("/", s3_arn_index)
+    if object_suffix_index == -1:
+        return False
+
+    clause_bucket_pattern = clause_pattern[:object_suffix_index]
+    return (
+        re.compile(clause_bucket_pattern, flags=re.IGNORECASE).fullmatch(resource_arn)
+        is not None
+    )
 
 
 def evaluate_notaction_for_permission(statement: Dict, permission: str) -> bool:
@@ -62,7 +91,7 @@ def evaluate_resource_for_permission(statement: Dict, resource_arn: str) -> bool
     if "resource" not in statement:
         return False
     for clause in statement["resource"]:
-        if evaluate_clause(clause, resource_arn):
+        if evaluate_resource_clause(clause, resource_arn):
             return True
     return False
 
@@ -72,7 +101,7 @@ def evaluate_notresource_for_permission(statement: Dict, resource_arn: str) -> b
     if "notresource" not in statement:
         return False
     for clause in statement["notresource"]:
-        if evaluate_clause(clause, resource_arn):
+        if evaluate_resource_clause(clause, resource_arn):
             return True
     return False
 
@@ -173,7 +202,6 @@ def principal_allowed_on_resource(
         )
 
         if explicit_deny:
-
             return False
         if not granted and allowed:
             granted = True
@@ -213,7 +241,7 @@ def calculate_permission_relationships(
     return allowed_mappings
 
 
-def compile_regex(item: str) -> Pattern:
+def compile_regex(item: str | Pattern) -> Pattern:
     r"""Compile a clause into a regex. Clause checking in AWS is case insensitive
     The following regex symbols will be replaced to make AWS * and ? matching a regex
     * -> .* (wildcard)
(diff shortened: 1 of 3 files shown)


cartography/intel/aws/permission_relationships.py
```diff
@@ -72,7 +101,7 @@ def evaluate_notresource_for_permission(statement: Dict, resource_arn: str) -> b
     if "notresource" not in statement:
         return False
     for clause in statement["notresource"]:
-        if evaluate_clause(clause, resource_arn):
+        if evaluate_resource_clause(clause, resource_arn):
             return True
     return False
```
jychp, line 106: Reusing `evaluate_resource_clause` for `notresource` regresses a case that worked before this PR. For `Allow s3:GetObject on *` with `NotResource: arn:aws:s3:::bucket/private/*`, the bucket ARN previously produced a `CAN_READ` edge (allow matched the bucket, notresource did not). After this PR the projection makes notresource match the bucket and cancels the allow, so the edge is dropped even though the principal can still read most of the bucket.

Consider keeping the bucket projection on the allow path only, or inverting it for notresource so a narrower object-scoped exception is treated as "doesn't apply to the bucket".

cartography/intel/aws/permission_relationships.py
```diff
@@ -37,6 +39,33 @@ def evaluate_clause(clause: str, match: str) -> bool:
...
+
+    object_suffix_index = clause_pattern.find("/", s3_arn_index)
+    if object_suffix_index == -1:
+        return False
+
+    clause_bucket_pattern = clause_pattern[:object_suffix_index]
+    return (
+        re.compile(clause_bucket_pattern, flags=re.IGNORECASE).fullmatch(resource_arn)
+        is not None
+    )
```
jychp, line 66: Symmetric projection for deny statements misrepresents AWS semantics. `Allow arn:aws:s3:::bucket/*` combined with `Deny arn:aws:s3:::bucket/private/*` now marks the bucket as explicitly denied, because the deny's resource clause projects onto the bucket ARN. The principal can still read everything outside `private/`, but the bucket-level `CAN_READ` edge is suppressed.

The projection is only safe in one direction (object-scoped allow → bucket allow). A narrower object-scoped deny should not become a bucket-wide deny.

jychp, line 66: `re.compile(clause_bucket_pattern, flags=re.IGNORECASE)` runs on every evaluation. `compile_statement()` exists to pre-compile each clause once for performance; this fallback undoes that for every S3 object-scoped clause on every (principal, resource) pair. Consider caching the derived bucket pattern alongside the original compiled clause.

cartography/intel/aws/permission_relationships.py
```diff
@@ -37,6 +39,33 @@ def evaluate_clause(clause: str, match: str) -> bool:
     return result is not None
 
 
+def evaluate_resource_clause(clause: str | Pattern, resource_arn: str) -> bool:
```
jychp, line 42: Missing docstring. A short note explaining the S3 object-to-bucket projection (and why it only exists here, not in `evaluate_clause`) would help future readers understand the intended scope.

tests/unit/cartography/intel/aws/test_permission_relationships.py
```diff
@@ -380,7 +481,7 @@ def test_full_policy_explicit_allow():
         "ListAllow": [
             {
                 "action": [
-                    "s3:listobject" "dynamodb:query",
+                    "s3:listobjectdynamodb:query",
```
jychp, line 484: Nit / out of scope: the prior `"s3:listobject" "dynamodb:query"` was an implicit-concat bug from a missing comma — the intent was probably two separate actions. Collapsing to `"s3:listobjectdynamodb:query"` silences the lint warning but keeps the single nonsense action. If you want to address it here, splitting into two list entries would be the real fix; otherwise leaving it untouched keeps the PR narrow. Same in `test_full_multiple_principal` below.

cartography/intel/aws/permission_relationships.py
```diff
@@ -20,8 +20,10 @@
 
 logger = logging.getLogger(__name__)
 
+S3_BUCKET_ARN_RE = re.compile(r"^(arn:[^:]+:s3:::)([^/]+)$", flags=re.IGNORECASE)
```
jychp, line 23: Nit: the two capture groups are never referenced — `re.fullmatch` alone is sufficient. Use non-capturing groups `(?:...)` or drop the parentheses entirely.

jychp (changes requested):

PR #2606, opened 2026-04-10
fix(microsoft): reduce detected app sync memory pressure

kunaals:
### Type of change
- [ ] Documentation update

### Summary
This change reduces peak memory usage during Microsoft syncs for larger tenants.

The main fix is in Intune detected app ingestion: instead of batching large DetectedApp objects with expanded managed-device collections and only flattening them at flush time, the sync now flattens each app into relationship rows immediately and flushes by row count. This keeps large app-to-device fanout from accumulating in memory.

The change also closes Azure credentials promptly after each Microsoft sync phase so long-running Microsoft syncs do not retain credential resources longer than necessary.

### Related issues or links
- N/A

### How was this tested?
- Unit and integration tests

#### General

#### Proof of functionality

### Notes for reviewers
The detected_apps change is intentionally scoped to batching and object lifetime only. It does not change the resulting node or relationship model, but it does add a unit test that verifies batching is now driven by flattened rows rather than source app count.


cartography/intel/microsoft/entra/app_role_assignments.py
```diff
@@ -26,30 +24,23 @@
...
-    RETURN sp.id as service_principal_id
-    """
-    service_principal_id = neo4j_session.execute_read(
-        read_single_value_tx, query, app_id=app_id
-    )
-
     if not service_principal_id:
         logger.warning(
-            f"No service principal found in graph for application {app_id}. Continuing."
+            f"No service principal found for application {app_id}. Continuing."
```
achantavy, line 43: [nonblock] %s formatting

cartography/intel/microsoft/entra/groups.py
```diff
@@ -138,53 +138,58 @@ async def sync_entra_groups(
...
+                        groups_batch, user_member_map, group_member_map, group_owner_map
+                    )
+                )
+                load_groups(neo4j_session, transformed_groups, update_tag, tenant_id)
 
-        users, subgroups = await call_with_retries(get_group_members, client, group.id)
-        user_member_map[group.id] = users
-        group_member_map[group.id] = subgroups
+                # Clear the batch and maps for processed groups
+                for g in groups_batch:
```
achantavy, line 178: [non block] needing explicit mem cleaning doesn't feel very pythonic but if it's needed, it's needed

achantavy (approved):
did a quick pass, stamping to unblock. i'm not a huge fan of needing the explicit mem management and a bunch of logic branches, but microsoft APIs are what they are

kunaals:
this actually didn't help much :( need to go back to drawing board

kunaals:
superceded by https://github.com/cartography-cncf/cartography/pull/2643

PR #2612, opened 2026-04-14
feat(gitlab): align code-to-cloud graph modeling

jychp:
### Type of change

- [ ] Documentation update

### Summary

This PR aligns the GitLab module with Cartography's generic code-to-cloud model so GitLab repositories can map cleanly to both GitLab Container Registry images and external registries such as ECR.

It switches GitLab core entities to numeric GitLab IDs, adds `CodeRepository` to `GitLabProject`, adds generic registry relationships (`REPO_IMAGE` and `IMAGE`) alongside deprecated legacy GitLab-specific edges, and ingests provenance directly onto `GitLabContainerImage` so `PACKAGED_FROM` matching can use the same graph pattern as the GitHub/ECR flow.

It also updates the GitLab schema docs, ontology URL mapping, and the affected integration/unit tests. A small AWS test stability fix is included so the full suite remains deterministic.

### Related issues or links

- Fixes #

### Breaking changes

- `GitLabOrganization`, `GitLabGroup`, `GitLabProject`, and `GitLabUser` now use numeric GitLab IDs as their primary `id` instead of `web_url`.
- Queries or downstream logic that treated `GitLabProject.id` (or the other GitLab node IDs) as a web URL must switch to the new `web_url` property.
- GitLab container registry ingestion now emits the generic `REPO_IMAGE` and `IMAGE` relationships. The legacy `HAS_TAG` and `REFERENCES` relationships are still emitted for backward compatibility and are marked deprecated for removal in `v1.0.0`.
- Commit attribution now matches GitLab users by email instead of display name to avoid false-positive author matches.

Migration path:
- Use `web_url` when you need the browser URL for GitLab entities.
- Prefer the generic graph pattern `(:ContainerRegistry)-[:REPO_IMAGE]->(:ImageTag)-[:IMAGE]->(:Image)` for code-to-cloud queries.

### How was this tested?

- `uv run pytest -q tests/integration/cartography/intel/gitlab tests/unit/cartography/intel/gitlab tests/integration/cartography/intel/trivy/test_trivy_gitlab_disk.py`
- `uv run pytest -q`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

Review the GitLab supply-chain path as two separate concerns:
1. identity/model changes (`id`, `web_url`, ontology mapping, `CodeRepository`)
2. registry/provenance alignment (`REPO_IMAGE`, `IMAGE`, provenance on `GitLabContainerImage`)

The AWS diff is limited to test determinism and does not change Cartography's AWS ingestion behavior.

### DIFF
--- a/cartography/intel/aws/util/botocore_config.py
+++ b/cartography/intel/aws/util/botocore_config.py
@@ -37,11 +37,7 @@ def _get_botocore_config(
     }
     if max_pool_connections is not None:
         kwargs["max_pool_connections"] = max_pool_connections
-    return _normalize_retries(
-        botocore.config.Config(**kwargs),
-        max_attempts=max_attempts,
-        retry_mode=retry_mode,
-    )
+    return botocore.config.Config(**kwargs)
 
 
 def get_botocore_config(
--- a/cartography/intel/aws/util/botocore_config.py
+++ b/cartography/intel/aws/util/botocore_config.py
@@ -4,8 +4,24 @@ from typing import Any
 import botocore.config
 
 
+def _normalize_retries(
+    config: botocore.config.Config,
+    *,
+    max_attempts: int,
+    retry_mode: str,
+) -> botocore.config.Config:
+    # Botocore mutates Config.retries during client creation by replacing
+    # max_attempts with total_max_attempts. Reset the shared config object so
+    # callers always see the contract Cartography expects.
+    config.retries = {
+        "max_attempts": max_attempts,
+        "mode": retry_mode,
+    }
+    return config
+
+
 @lru_cache(maxsize=None)
-def get_botocore_config(
+def _get_botocore_config(
     *,
     read_timeout: int = 120,
     max_attempts: int = 10,
@@ -21,7 +37,30 @@ def get_botocore_config(
     }
     if max_pool_connections is not None:
         kwargs["max_pool_connections"] = max_pool_connections
-    return botocore.config.Config(**kwargs)
+    return _normalize_retries(
+        botocore.config.Config(**kwargs),
+        max_attempts=max_attempts,
+        retry_mode=retry_mode,
+    )
+
+
+def get_botocore_config(
+    *,
+    read_timeout: int = 120,
+    max_attempts: int = 10,
+    retry_mode: str = "adaptive",
+    max_pool_connections: int | None = None,
+) -> botocore.config.Config:
+    return _normalize_retries(
+        _get_botocore_config(
+            read_timeout=read_timeout,
+            max_attempts=max_attempts,
+            retry_mode=retry_mode,
+            max_pool_connections=max_pool_connections,
+        ),
+        max_attempts=max_attempts,
+        retry_mode=retry_mode,
+    )
 
 
 def _create_client(
--- a/cartography/intel/gitlab/container_image_attestations.py
+++ b/cartography/intel/gitlab/container_image_attestations.py
@@ -354,8 +354,9 @@ def transform_image_provenance_records(
         parent_image_digest = attestation.get("parent_image_digest")
         if not attests_digest or (not source_uri and not parent_image_digest):
             continue
-        provenance_by_digest[str(attests_digest)] = {
-            "digest": attests_digest,
+        digest = str(attests_digest)
+        existing = provenance_by_digest.get(digest, {"digest": attests_digest})
+        candidate = {
             "source_uri": source_uri,
             "source_revision": attestation.get("source_revision"),
             "source_file": attestation.get("source_file"),
@@ -364,6 +365,8 @@ def transform_image_provenance_records(
             "from_attestation": True,
             "confidence": 1.0,
         }
+        existing.update({k: v for k, v in candidate.items() if v is not None})
+        provenance_by_digest[digest] = existing
 
     records = list(provenance_by_digest.values())
     logger.info("Transformed %d image provenance record(s)", len(records))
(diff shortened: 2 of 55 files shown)


jychp:
Follow-up on the issues identified by cubic:

Fixed:
- `cartography/intel/gitlab/__init__.py`: cleanup calls are now aligned with the numeric-ID refactor and pass both the scoped numeric ID and `gitlab_url`.
- `cartography/intel/gitlab/projects.py`: removed the synthetic `project.get("id", 0)` fallback; project IDs now use the canonical required GitLab ID.
- `cartography/intel/gitlab/supply_chain.py`: kept generic cross-registry matching, but added GitLab-org scoping for GitLab Container Registry images and fixed the `GraphJob.from_matchlink()` typing/cleanup path.
- `docs/root/modules/gitlab/schema.md`: documented the indexed `web_url` / `gitlab_url` fields, bolded indexed fields, and added the legacy `HAS_TAG` / `REFERENCES` edges that are still emitted for backward compatibility.

Not applied intentionally:
- cubic suggested removing the `fetch_registry_blob` mock from `tests/integration/cartography/intel/gitlab/test_container_registry.py`.
- Short rationale: that test already exercises the attestation parsing and provenance-ingestion path; unmocking `fetch_registry_blob()` would only move the seam one layer lower into mocked HTTP/blob transport, make the test more brittle, and would not add meaningful integration coverage in this test harness.

Validation after the fixes:
- `uv run pre-commit run --all-files`
- `uv run pytest -q tests/integration/cartography/intel/gitlab tests/unit/cartography/intel/gitlab tests/integration/cartography/intel/trivy/test_trivy_gitlab_disk.py`
- `uv run pytest -q`

Current local result: `2031 passed, 73 warnings`.

jychp:
@cubic-dev-ai review this PR again

cartography/intel/gitlab/supply_chain.py
```diff
@@ -25,47 +26,66 @@
...
 
-    Scoped to the current GitLab organization via the graph path:
-    GitLabOrganization → RESOURCE → GitLabContainerRepository → HAS_TAG → Tag → REFERENCES → Image.
+    Uses the generic ontology labels and relationship labels so the same query works
+    for ECR and GitLab Container Registry.
 
     Returns one image per container repository (preferring 'latest' tag, then most recent).
     Excludes images that already have a PACKAGED_FROM relationship created in the current
     sync iteration (i.e. by provenance matching). Images with stale PACKAGED_FROM from
     previous iterations are included so they can be re-matched before cleanup removes them.
```
kunaals, line 50: i think this may be depending on generic tag fields that GitLab tags don't actually set. from what i can tell, the GitLab tag model/API gives us `name`, `location`, `created_at`, etc., but not `tag`, `uri`, or `image_pushed_at`, so this ordering may end up being arbitrary here. if those are getting populated elsewhere, ignore me.

cartography/intel/gitlab/container_image_attestations.py
```diff
@@ -222,48 +231,157 @@ def get_container_image_attestations(
     return all_attestations, summary
 
 
+def _extract_predicate_from_attestation(
+    attestation: dict[str, Any],
+    gitlab_url: str,
+    token: str,
+) -> dict[str, Any] | None:
+    """
+    Decode a DSSE attestation blob and return the embedded predicate when present.
```
kunaals: i think this may only handle the DSSE-style attestation shape where the blob has a base64 `payload`. BuildKit attestations are often stored as raw `application/vnd.in-toto+json` statements, so in that case we'd skip provenance extraction entirely. might be worth accepting both formats unless GitLab always normalizes this for us.

cartography/intel/gitlab/container_image_attestations.py
```diff
@@ -222,48 +231,157 @@ def get_container_image_attestations(
...
+    """
+    registry_url = attestation.get("_registry_url")
+    repository_name = attestation.get("_repository_name")
+    layers = attestation.get("layers", [])
+    if not registry_url or not repository_name or not layers:
+        return None
+
```
kunaals, line 251: i think this guard may be a little tighter than we want. `extract_container_parent_image()` can still give us a valid parent image even when there's no `source_uri`, but this path drops the record unless `source_uri` is present. that seems like it could prevent us from creating `BUILT_FROM` for attestations that identify the base image but not the source repo.

kunaals (commented):
a few things i think may be worth another look

cartography/intel/aws/util/botocore_config.py
```diff
@@ -21,7 +37,30 @@ def get_botocore_config(
...
+
+
+def get_botocore_config(
+    *,
+    read_timeout: int = 120,
+    max_attempts: int = 10,
+    retry_mode: str = "adaptive",
+    max_pool_connections: int | None = None,
+) -> botocore.config.Config:
+    return _normalize_retries(
```
kunaals, line 50: i think this ends up normalizing retries twice now: once in `_get_botocore_config()` and again here. not a blocker, but if the extra reset isn't intentional it seems like we could probably do it in one place and keep the same behavior.

docs/root/modules/gitlab/schema.md
```diff
@@ -196,7 +204,7 @@ ORDER BY project_count DESC
     - **last_commit_date**: Timestamp of the user's most recent commit to the project
     - **first_commit_date**: Timestamp of the user's oldest commit to the project
 
-    Commit authors are matched to GitLab users by display name. Only commits from current members are tracked.
+    Commit authors are matched to GitLab users by email address when available, with a display-name fallback for current members. Only commits from current members are tracked.
```
kunaals, line 207: i think there’s still one stale sentence a bit lower in this `COMMITTED_TO` section that says commit authors are matched "by name". might be worth updating that too so the docs are consistent with the new email-first / name-fallback behavior.

kunaals (approved):
approving to unblock. i think this is in good shape now; just left a couple minor nits that seem worth cleaning up before merge.

PR #2616, opened 2026-04-15
fix(aws): reduce SageMaker regional retry tail

kunaals:
### Type of change

### Summary

This change reduces repeated SageMaker regional retries during AWS syncs.

It filters requested regions to SageMaker-supported regions, treats transient regional SageMaker failures as first-class skip signals, carries failed-region state forward across SageMaker submodules for the same account, and suppresses cleanup when any SageMaker region was transiently skipped so last-known-good data is preserved.

### Related issues or links

- Fixes #

### Breaking changes

None.

### How was this tested?

- Added new unit tests

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### DIFF
--- a/cartography/intel/aws/sagemaker/__init__.py
+++ b/cartography/intel/aws/sagemaker/__init__.py
@@ -89,6 +89,6 @@ def sync(
             current_aws_account_id,
             update_tag,
             common_job_parameters,
-            skip_regions.copy(),
+            skip_regions,
         )
         skip_regions.update(newly_failed_regions)
--- a/cartography/intel/aws/sagemaker/__init__.py
+++ b/cartography/intel/aws/sagemaker/__init__.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Dict
 from typing import List
+from typing import Set
 
 import boto3
 import neo4j
@@ -17,6 +18,7 @@ from cartography.intel.aws.sagemaker.notebook_instances import sync_notebook_ins
 from cartography.intel.aws.sagemaker.training_jobs import sync_training_jobs
 from cartography.intel.aws.sagemaker.transform_jobs import sync_transform_jobs
 from cartography.intel.aws.sagemaker.user_profiles import sync_user_profiles
+from cartography.intel.aws.sagemaker.util import get_available_sagemaker_regions
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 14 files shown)


cartography/intel/aws/sagemaker/transform_jobs.py
```diff
@@ -7,18 +7,19 @@
...
 )
-from cartography.util import aws_handle_regions
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
 
 
 @timeit
-@aws_handle_regions
+@sagemaker_handle_regions
```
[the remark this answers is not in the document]
kunaals: keeping this as-is because sagemaker needs custom transient-region handling here, and the shared decorator would swallow the failures this pr is fixing.

cartography/intel/aws/sagemaker/util.py
```diff
@@ -1,4 +1,209 @@
...
+    get_partition_for_region = getattr(boto3_session, "get_partition_for_region", None)
+    if callable(get_partition_for_region):
+        for region in regions:
+            try:
+                partitions.add(get_partition_for_region(region))
+            except Exception:
+                logger.debug(
+                    "Could not determine AWS partition for region '%s' while filtering SageMaker regions.",
+                    region,
+                )
```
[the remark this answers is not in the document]
kunaals: good catch. narrowed this to unknownregionerror and added a test.

cartography/intel/aws/sagemaker/util.py
```diff
@@ -1,4 +1,210 @@
...
+            "ServiceUnavailableException",
+            "Throttling",
+            "ThrottlingException",
+            "TooManyRequestsException",
+        }
+        or status_code in _RETRYABLE_HTTP_STATUS_CODES
+    )
+
+
+def _is_region_unsupported_error(
```
shyammukund: Question: Some of these functions seem like they are similar to functions that are already defined in the cartography/util.py file. liek there is a is_regions_unsupported_unknown_operations function, but I am assuming that the intention was to seperate concerns for sagemaker because it has some differences to the general aws modules?

cartography/intel/aws/sagemaker/__init__.py
```diff
@@ -46,102 +48,47 @@ def sync(
...
+
+    for sync_submodule in submodule_syncs:
+        newly_failed_regions = sync_submodule(
+            neo4j_session,
+            boto3_session,
+            sagemaker_regions,
+            current_aws_account_id,
+            update_tag,
+            common_job_parameters,
+            skip_regions.copy(),
```
shyammukund: sync_sagemaker_resource only iterates over skip_regions — it never mutates the set passed in. I believe the .copy () is overly defensive.

shyammukund (approved):

cartography/intel/aws/sagemaker/util.py
kunaals: yeah sagemaker seems a bit different, i will abstract this out though to remove duplicate logic

cartography/intel/aws/sagemaker/__init__.py
kunaals: good call will remove

PR #2619, opened 2026-04-16
feat(gcp): Update CloudRunJob and CloudRunRevision to have container ontology label

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- This PR gives both GCPCloudRunJob and GCPCloudRunRevision the :Container Ontology label 

- The reason that both these nodes need to have the label is because in GCP world both these node types have container-image workloads. Since the analysis job that computes RESOLVED_IMAGE relises on the container label if either of these node types don't have the label it will ignore the HAS_IMAGE rels that originate from a GCPCloudRunJob/GCPCloudRunRevision and will will make those Container-Image workloads invisible. This double label situation gives us parity with how K8s, ECS, and AzureGroupContainers handle the RESOLVED_IMAGE rel. It just has 2 types of "containers" instead of one. 

### Related issues or links

- Closes gap in coverage mentioned in https://github.com/cartography-cncf/cartography/pull/2617

### How was this tested?

- Integration Tests updated

#### General

#### Proof of functionality

### DIFF
--- a/cartography/models/gcp/cloudrun/revision.py
+++ b/cartography/models/gcp/cloudrun/revision.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 from cartography.models.core.common import PropertyRef
 from cartography.models.core.nodes import CartographyNodeProperties
 from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.nodes import ExtraNodeLabels
 from cartography.models.core.relationships import CartographyRelProperties
 from cartography.models.core.relationships import CartographyRelSchema
 from cartography.models.core.relationships import LinkDirection
@@ -143,6 +144,7 @@ class CloudRunRevisionToArtifactRegistryContainerImageRel(CartographyRelSchema):
 class GCPCloudRunRevisionSchema(CartographyNodeSchema):
     label: str = "GCPCloudRunRevision"
     properties: GCPCloudRunRevisionProperties = GCPCloudRunRevisionProperties()
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Container"])
     sub_resource_relationship: ProjectToCloudRunRevisionRel = (
         ProjectToCloudRunRevisionRel()
     )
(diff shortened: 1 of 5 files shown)


tests/integration/cartography/intel/ontology/test_resolved_images.py
```diff
@@ -43,6 +43,76 @@ def test_resolved_image_analysis_creates_rel_via_has_image(neo4j_session):
     ) == {("container-1", "sha256:deadbeef")}
 
 
+def test_resolved_image_analysis_creates_rel_for_gcp_cloud_run_revision(neo4j_session):
```
kunaals: i think this test may be bypassing the actual change here a bit. since it creates `:Container` nodes directly in cypher, it looks like it would still pass even if the schema no longer added the label. could we run this through the cloud run load path so it proves the model change too?

cartography/models/gcp/cloudrun/revision.py
```diff
@@ -3,6 +3,7 @@
 from cartography.models.core.common import PropertyRef
 from cartography.models.core.nodes import CartographyNodeProperties
 from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.nodes import ExtraNodeLabels
```
kunaals, line 6: i think this is enough to make cloud run jobs/revisions participate in `RESOLVED_IMAGE`, but i had one ontology follow-up: should these also be added to `cartography/models/ontology/mapping/data/containers.py`? as written they get the `:Container` label, but i don’t think they get the normalized container `_ont_*` fields yet, so cross-provider container queries may still be incomplete. feels non-blocking for this pr, but maybe worth doing here if the intent is full container ontology support rather than just analysis-job participation.

kunaals (approved):
approving with some nits. i think the label placement makes sense for the current cloud run model; the two follow-ups below are the only things i’d still consider tightening.

cartography/models/gcp/cloudrun/revision.py
shyammukund: Agreed we should add these to the ontology mapping to get the normalized container fields along with AzureGroupContainer should be added to the mapping to be apart of the full ontology for container and  not just have the label. Will create an issue for a follow up PR

PR #2620, opened 2026-04-16
feature(reports): Add unified report source options for report ingestion

kunaals:
### Type of change

### Summary

Add a unified `--*-source` configuration path for bucket-backed report ingestion and keep the legacy module-specific local/S3 flags as compatibility shims.

This builds on the shared read-only object-store layer by letting Trivy, Syft, AIBOM, and Docker Scout accept a single source value that can point at a local path, `s3://...`, `gs://...`, or `azblob://...`. The legacy `*-results-dir` and `*-s3-*` flags are still supported for backward compatibility, but they now normalize into the same internal source model and emit deprecation warnings for removal in Cartography v1.0.0.

The goal is to keep the user-facing configuration smaller and provider-native while preserving existing behavior during the migration window.

### Related issues or links

- N/A

### How was this tested?

- Tested locally against a live bucket in each cloud provider and local files. Also tested legacy flags

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

- The new cloud source formats are `s3://bucket/prefix`, `gs://bucket/prefix`, and `azblob://account/container/prefix`; any other value is treated as a local filesystem path.
- Native GCS and Azure blob readers are included behind the shared interface, but this PR keeps the current module behavior intact and focuses on the ingestion-source/config surface.
- There is no `make lint` target in this repo, so the enforced pre-commit hooks were run directly instead.

#### What we keep as backward-compat shims (and why)

The intent is that **CLI flags + Config fields are the only user-facing back-compat surface**; everything else funnels through the new unified path. Three categories of shim exist in this PR, each with different lifetimes:

1. **CLI flags** (`--trivy-results-dir`, `--trivy-s3-bucket`, `--trivy-s3-prefix`, and the equivalents for syft/aibom/docker-scout): kept as syntactic sugar that normalizes into `--*-source` internally. Hidden from `--help`. Deprecation warning at parse time. Removed in v1.0.0.
2. **`Config` fields** (`trivy_results_dir`, `trivy_s3_bucket`, `trivy_s3_prefix`, etc.): kept on the `Config` ctor for embedders that construct Config directly. The CLI forwards both the resolved `*_source` and the original legacy fields so post-parse `config.trivy_s3_bucket` introspection still works (existing tests document this behavior). Deprecation warning at Config init. Removed in v1.0.0.
3. **Module-level wrappers** (`sync_aibom_from_dir`, `sync_aibom_from_s3`, plus the trivy/syft/docker-scout equivalents and the trivy scanner internals): **no production code in cartography calls these.** The CLI/Config flow goes `start_*_ingestion → parse_report_source → build_report_reader_for_source → sync_*_from_report_reader`. These wrappers exist purely so external embedders that import them directly don't break on this release. They forward to the unified `sync_*_from_report_reader`. Marked `# DEPRECATED` for v1.0.0 removal.

The unified production path is `sync_*_from_report_reader(neo4j_session, reader, ...)`, where `reader` is one of `LocalReportReader`, `S3BucketReader`, `GCSBucketReader`, or `AzureBlobContainerReader`. All readers are context managers and release their backend client connections on exit.


tests/integration/cartography/intel/syft/test_syft.py
```diff
@@ -123,3 +126,36 @@ def test_sync_syft_from_dir(
     ).single()
 
     assert result["count"] == 3
+
+
+@patch(
+    "cartography.intel.syft.S3BucketReader.list_objects",
```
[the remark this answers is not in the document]
kunaals: good catch. changed the test to drive the real s3bucketreader path and only mock the paginator/get_object calls.

kunaals:
@cubic-dev-ai brotha take another look

cartography/intel/docker_scout/__init__.py
```diff
@@ -99,41 +86,41 @@ def sync_docker_scout_from_s3(
...
     synced_count = 0
-    s3_client = boto3_session.client("s3")
-    for s3_key in report_files:
-        response = s3_client.get_object(Bucket=s3_bucket, Key=s3_key)
+    for ref in report_refs:
         try:
-            raw_recommendation = response["Body"].read().decode("utf-8")
-        except UnicodeDecodeError:
+            raw_recommendation = read_text_document(reader, ref)
+        except ObjectStoreParseError:
```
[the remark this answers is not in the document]
kunaals: keeping this as-is. this path intentionally skips unreadable reports, and that matches the pre-refactor s3 behavior.

cartography/intel/aibom/__init__.py
```diff
@@ -76,18 +53,15 @@ def _iter_documents_from_dir(
...
-        source = f"s3://{s3_bucket}/{object_key}"
+    for ref in json_files:
+        source = ref.uri
         try:
-            response = s3_client.get_object(Bucket=s3_bucket, Key=object_key)
-            scan_data_json = response["Body"].read().decode("utf-8")
-            document = json.loads(scan_data_json)
-        except (OSError, UnicodeDecodeError, json.JSONDecodeError) as exc:
+            document = read_json_document(reader, ref)
+        except ObjectStoreParseError as exc:
```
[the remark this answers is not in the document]
kunaals: keeping this as-is. per-object unreadable report handling already behaved this way before this refactor, so this is not a new regression.

cartography/intel/common/object_store.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+        return refs
+
+    def read_bytes(self, ref: ObjectRef) -> bytes:
+        account_name, _sep, container_name = ref.bucket.partition("/")
+        if not account_name or not container_name:
+            raise ObjectStoreParseError(
+                ref.uri,
+                "Azure blob reference is missing account or container information",
+            )
```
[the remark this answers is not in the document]
kunaals: good catch. added account validation before the blob download.

cartography/intel/common/report_source.py
```diff
@@ -0,0 +1,179 @@
+import os
+import re
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from cartography.intel.common.object_store import BucketReader
+
+_SOURCE_SCHEME_RE = re.compile(r"^(?P<scheme>[a-z][a-z0-9+.-]*)://(?P<rest>.*)$")
```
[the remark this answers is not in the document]
kunaals: good catch. source scheme parsing is case-insensitive now.

cartography/config.py
```diff
@@ -540,9 +556,33 @@ def __init__(
...
+        # These legacy source flags will be removed in Cartography v10.0.0.
+        self.docker_scout_source = docker_scout_source or (
+            docker_scout_results_dir
+            if docker_scout_results_dir
+            else (
+                f"s3://{docker_scout_s3_bucket}/{docker_scout_s3_prefix}"
+                if docker_scout_s3_bucket and docker_scout_s3_prefix is not None
+                else (
+                    f"s3://{docker_scout_s3_bucket}" if docker_scout_s3_bucket else None
+                )
```
jychp, line 569: Suggestion: this inline shim is inconsistent with `build_s3_source`:
- It doesn't `lstrip("/")` the prefix, so a caller passing `s3_prefix="/foo"` ends up with `s3://bucket//foo`.
- When `s3_prefix == ""`, it emits `s3://bucket/` (trailing slash), while the CLI path returns `s3://bucket`.

Reuse `build_s3_source(...)` / `parse_report_source(...)` here so programmatic and CLI callers resolve identically. Same applies to the trivy/syft/aibom shims below.

cartography/config.py
```diff
@@ -540,9 +556,33 @@ def __init__(
...
+        # DEPRECATED: `docker_scout_results_dir` and `docker_scout_s3_*` compatibility shim.
+        # These legacy source flags will be removed in Cartography v10.0.0.
+        self.docker_scout_source = docker_scout_source or (
+            docker_scout_results_dir
+            if docker_scout_results_dir
+            else (
+                f"s3://{docker_scout_s3_bucket}/{docker_scout_s3_prefix}"
+                if docker_scout_s3_bucket and docker_scout_s3_prefix is not None
+                else (
+                    f"s3://{docker_scout_s3_bucket}" if docker_scout_s3_bucket else None
```
jychp, line 568: Suggestion: the CLI emits a `DEPRECATED: ...` warning when legacy flags are used, but this shim is silent. Embedders constructing `Config(...)` directly with `docker_scout_results_dir` / `docker_scout_s3_*` won't see any v10.0.0 notice. Consider emitting `logger.warning` (or `warnings.warn(..., DeprecationWarning)`) from the shim for parity.

cartography/intel/common/report_source.py
```diff
@@ -0,0 +1,182 @@
...
+
+    authenticator = Authenticator()
+    if azure_sp_auth:
+        credentials = authenticator.authenticate_sp(
+            tenant_id=azure_tenant_id,
+            client_id=azure_client_id,
+            client_secret=azure_client_secret,
+        )
+    else:
+        credentials = authenticator.authenticate_cli()
```
jychp, line 168: `Authenticator().authenticate_cli()` runs `SubscriptionClient.subscriptions.list()`, which requires subscription-list permission that's unrelated to blob storage access. A user with only blob-data roles will hit the `return None` path and fail with the generic "Azure credentials are not available" error. For blob-only flows, instantiating `AzureCliCredential()` (or `DefaultAzureCredential()`) directly should be enough.

cartography/intel/aibom/__init__.py
```diff
@@ -76,18 +57,15 @@ def _iter_documents_from_dir(
 
 
 def _iter_documents_from_s3(
-    json_files: set[str],
-    s3_bucket: str,
-    s3_client: Any,
+    json_files: list[ObjectRef],
+    reader: BucketReader,
 ) -> Iterator[tuple[str, dict[str, Any]]]:
     """Yield (source_label, document) pairs from S3 objects."""
```
jychp, line 63: Nit: this helper now accepts any `BucketReader` (GCS / Azure / S3), so the `_s3` suffix is misleading. Rename to `_iter_documents_from_bucket_reader` to match the other renamed helpers (`sync_aibom_from_bucket_reader`, etc.).

cartography/intel/common/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    "build_s3_source",
+    "GCSBucketReader",
+    "GCSReportSource",
+    "LocalReportSource",
+    "ObjectRef",
+    "ObjectStoreParseError",
+    "parse_report_source",
+    "S3BucketReader",
+    "S3ReportSource",
+    "ReportSource",
```
jychp, line 37: Nit: `ReportSource` sits between `S3ReportSource` and `filter_object_refs`, breaking the otherwise-alphabetical ordering of `__all__`.

cartography/config.py
jychp: Typo: v1.0.0 not v10.0.0

jychp (changes requested):

cartography/config.py
kunaals: good catch. the config shim now goes through the same source builder/parser path, so direct config callers normalize the s3 source the same way as the cli.

cartography/config.py
kunaals: good catch. the config shim warns now too, and i fixed the removal version to v1.0.0.

cartography/intel/common/report_source.py
kunaals: good catch. blob-only sources now use azureclicredential directly instead of the subscription-list auth path.

cartography/intel/aibom/__init__.py
kunaals: renamed this to match the bucket-reader abstraction.

cartography/intel/common/__init__.py
kunaals: fixed the export ordering

cartography/intel/common/object_store.py
```diff
@@ -0,0 +1,276 @@
...
+                ),
+            )
+        return refs
+
+    def read_bytes(self, ref: ReportRef) -> bytes:
+        with open(ref.uri, "rb") as file_pointer:
+            data = file_pointer.read()
+        if isinstance(data, str):
+            return data.encode("utf-8")
+        return data
```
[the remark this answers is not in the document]
kunaals: good catch. updated the test mocks to encode utf-8 so they match the binary-mode contract, and dropped the dead guard in production.

cartography/intel/common/report_source.py
```diff
@@ -0,0 +1,281 @@
...
+        )
+    if local_path and (s3_bucket or s3_prefix):
+        raise ValueError(
+            f"Cannot use both {names.local} and {names.s3_bucket}/{names.s3_prefix}. "
+            f"Use {names.source} instead.",
+        )
+    if s3_prefix and not s3_bucket:
+        raise ValueError(f"{names.s3_prefix} requires {names.s3_bucket}.")
+
+    if source:
```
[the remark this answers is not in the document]
kunaals: good catch. switched the source presence check to `is not None` so an explicit empty string either fails the legacy-flag conjunction check or falls through to `parse_report_source`, which raises. added unit coverage for both paths.

kunaals:
@cubic-dev-ai can you take one more look?

cartography/cli.py
```diff
@@ -8,6 +8,8 @@
 from typing_extensions import Annotated
 
 from cartography.config import Config
+from cartography.intel.common.report_source import LegacyReportSourceNames
```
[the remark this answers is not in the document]
kunaals: good catch. split the cloud-reader builder out so cli can keep top-level imports without loading cloud sdks on startup.

tests/integration/cartography/intel/aibom/test_aibom.py
```diff
@@ -686,8 +711,13 @@ def test_sync_aibom_skips_s3_unicode_decode_errors(
     boto3_session.client.return_value = s3_client
 
     with patch(
-        "cartography.intel.aibom._get_json_files_in_s3",
-        return_value={"reports/aibom-bad-encoding.json"},
+        "cartography.intel.aibom.S3BucketReader.list_reports",
```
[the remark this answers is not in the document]
kunaals: good catch. changed this test to exercise the s3bucketreader listing path and only mock the s3 paginator/get_object calls.

cartography/intel/common/report_source.py
```diff
@@ -0,0 +1,274 @@
...
+            f"Use {names.source} instead.",
+        )
+    if s3_prefix and not s3_bucket:
+        raise ValueError(f"{names.s3_prefix} requires {names.s3_bucket}.")
+
+    if source is not None:
+        parse_report_source(source)
+        return source
+
+    if local_path:
```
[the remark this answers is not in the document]
kunaals: good catch. fixed explicit empty legacy fields to validate instead of being ignored.

cartography/intel/common/object_store.py
```diff
@@ -0,0 +1,251 @@
...
+        for path in self._root.rglob("*"):
+            if not path.is_file() or path.name.startswith("."):
+                continue
+            refs.append(
+                ReportRef(
+                    uri=str(path),
+                    name=str(path.relative_to(self._root)),
+                ),
+            )
+        return refs
```
jychp, line 117: `LocalReportReader.list_reports` only handles directories. The new `--trivy-source` / `--syft-source` / `--aibom-source` / `--docker-scout-source` flags are documented as accepting a local path, but `Path(file).rglob('*')` returns nothing for a file, so passing a single report file silently produces "no reports found". Either accept a file path here (when `self._root.is_file()`, yield a single ref) or restrict the docs/help text to directories only.

cartography/intel/common/report_reader_builder.py
```diff
@@ -0,0 +1,61 @@
...
+            boto3.Session(),
+            source.bucket,
+            source.prefix,
+            source.uri,
+        )
+
+    if isinstance(source, GCSReportSource):
+        return object_store.GCSBucketReader(source.bucket, source.prefix, source.uri)
+
+    if azure_sp_auth:
```
jychp: replace the implicit Azure fallthrough with an explicit `isinstance(source, AzureBlobReportSource)` check and raise on anything else. As written, any future variant added to `ReportSource` will silently route to the Azure branch and fail with `AttributeError` on `source.account_name` instead of a clear unsupported-source error.

cartography/cli.py
```diff
@@ -1361,120 +1385,168 @@ def run(
...
                 typer.Option(
                     "--syft-results-dir",
-                    help="Local directory containing Syft JSON results.",
+                    help="DEPRECATED: use --syft-source with a local path. Will be removed in Cartography v1.0.0.",
                     rich_help_panel=PANEL_SYFT,
-                    hidden=PANEL_SYFT not in visible_panels,
+                    hidden=True,
                 ),
             ] = None,
             # AIBOM Options
```
jychp, line 1513: Nit: the AIBOM section is missing the leading `# ===` divider that all the other option groups in this file use (see Docker Scout/Trivy/Syft above). Add the divider line for consistency.

jychp (changes requested):

cartography/intel/common/object_store.py
kunaals: good catch. local report readers now accept a single file path and there is coverage for that path.

cartography/intel/common/report_reader_builder.py
kunaals: good catch. made the azure branch explicit and added a clear unsupported-source error.

cartography/cli.py
kunaals: fixed, added the missing divider.

cartography/config.py
```diff
@@ -1,3 +1,28 @@
+from cartography.intel.common.report_source import LegacyReportSourceNames
+from cartography.intel.common.report_source import (
```
jychp: [blocking] `config.py` imports `LegacyReportSourceNames` and `resolve_report_source_with_legacy_fields` at module scope. This file is imported by every sync path in cartography. Every invocation, even ones that never touch Trivy/Syft/AIBOM/Docker Scout, pays the import cost of `report_source.py` and its transitive dependencies. Move the two imports inside `_resolve_report_source_config`.

cartography/intel/common/object_store.py
```diff
@@ -0,0 +1,254 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+from typing import Callable
+from typing import Iterable
+from typing import Protocol
+
+import boto3
+from azure.storage import blob as azure_blob
+from google.cloud import storage
```
jychp: Can we do lazy import to avoid loading all SDK ?

cartography/intel/common/object_store.py
```diff
@@ -0,0 +1,254 @@
...
+                    ),
+                )
+        return refs
+
+    def read_bytes(self, ref: ReportRef) -> bytes:
+        try:
+            response = self._client.get_object(Bucket=self._bucket, Key=ref.name)
+            return response["Body"].read()
+        except Exception as exc:
+            raise ObjectStoreParseError(ref.uri, "Failed to read S3 report") from exc
```
jychp: Use SDK specific errors to avoid swallowing error like AttributeError

cartography/intel/common/report_reader_builder.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+        if credentials is None:
+            raise RuntimeError(
+                "Azure Blob report source was configured, but Azure credentials are not available.",
+            )
+        credential = credentials.credential
+    else:
+        credential = azure_identity.AzureCliCredential()
+
```
jychp, line 77: [suggestion] When `azure_sp_auth` is falsy but `azure_tenant_id`/`azure_client_id`/`azure_client_secret` are non-None, those credentials are silently ignored and `AzureCliCredential()` is used instead. Add a `logger.warning` when SP credential fields are set but `azure_sp_auth` is falsy so misconfigured deployments produce an observable signal.

cartography/intel/trivy/scanner.py
```diff
@@ -511,11 +504,13 @@ def sync_single_image_from_s3(
...
         boto3_session: boto3 session for S3 operations
     """
-    s3_client = boto3_session.client("s3")
-
+    # DEPRECATED: sync_single_image_from_s3() will be removed in v1.0.0.
     logger.debug(f"Reading scan results from S3: s3://{s3_bucket}/{s3_object_key}")
-    response = s3_client.get_object(Bucket=s3_bucket, Key=s3_object_key)
-    scan_data_json = response["Body"].read().decode("utf-8")
+    reader = S3BucketReader(boto3_session, s3_bucket, "")
+    scan_data_json = read_text_report(
```
jychp: [suggestion] Creating `S3BucketReader(boto3_session, s3_bucket, "")` with an empty prefix just to fetch a single specific key via `ref.name=s3_object_key` is semantically odd. This is a deprecated shim so the risk is bounded, but add a comment explaining why prefix is empty to avoid confusing the next maintainer.

cartography/intel/common/report_source.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+
+
+CloudReportSource = S3ReportSource | GCSReportSource | AzureBlobReportSource
+ReportSource = LocalReportSource | CloudReportSource
+
+
+def build_s3_source(bucket: str, prefix: str | None = None) -> str:
+    normalized_prefix = (prefix or "").lstrip("/")
+    if normalized_prefix:
```
jychp, line 85: [nitpick] `build_s3_source("bucket", "/reports/")` silently strips the leading slash and returns `s3://bucket/reports/`. A caller passing a prefix from an environment variable with a leading slash gets silent normalization with no indication. Add a `logger.debug` or raise `ValueError` for leading-slash prefixes to make the normalization observable.

jychp (changes requested):

cartography/config.py
kunaals: good call. moved config report-source imports inside the compatibility resolver.

cartography/intel/common/object_store.py
kunaals: good call. object-store readers now lazy-import provider sdks only when that reader is constructed or reads.

cartography/intel/common/object_store.py
kunaals: good catch. reads now wrap provider-specific sdk errors instead of broad exceptions.

cartography/intel/common/report_reader_builder.py
kunaals: added the warning for ignored service principal fields when azure_sp_auth is disabled.

cartography/intel/trivy/scanner.py
kunaals: added a comment explaining the empty prefix in the deprecated single-key s3 wrapper.

cartography/intel/common/report_source.py
kunaals: added debug logging when leading slashes are normalized from cloud prefixes.

cartography/intel/syft/__init__.py
```diff
@@ -53,83 +60,37 @@ def sync_single_syft(
...
 
-    for file_path in json_files:
-        with open(file_path, encoding="utf-8") as f:
-            syft_data = json.load(f)
+    for ref in json_files:
+        logger.debug(
+            "Reading scan results from report source: %s",
+            ref.uri,
+        )
+        syft_data = read_json_report(reader, ref)
```
jychp: [blocking] `read_json_report` is not wrapped in a try/except, so a single unreadable or malformed JSON aborts the whole sync mid-loop. Trivy and AIBOM both catch `ObjectStoreParseError` and continue. Worse: when this raises, prior iterations have already loaded data and `cleanup_syft` is skipped, leaving partial/stale `SyftPackage` nodes in the graph. Catch `ObjectStoreParseError` (and the partial-failure decision should mirror trivy's "skip cleanup if any failed").

cartography/intel/common/report_source.py
```diff
@@ -0,0 +1,232 @@
...
+        return LocalReportSource(raw=raw_source, path=os.path.expanduser(source))
+
+    scheme = scheme_match.group("scheme").lower()
+    remainder = scheme_match.group("rest")
+
+    if scheme == "s3":
+        bucket, _sep, prefix = remainder.partition("/")
+        if not bucket:
+            raise ValueError("S3 report source must include a bucket name.")
+        return S3ReportSource(raw=raw_source, bucket=bucket, prefix=prefix)
```
jychp: [suggestion] `parse_report_source` preserves a leading slash in the prefix (e.g. `s3://bucket//reports/` → `prefix="/reports/"`), which is then passed straight to `list_objects_v2` / `list_blobs` and matches nothing. The legacy `*_s3_prefix="/reports/"` path normalized this via `_normalize_cloud_prefix`. Run prefixes through `_normalize_cloud_prefix` here too (S3, GCS, Azure) so cloud URIs behave the same as legacy flags.

cartography/cli.py
```diff
@@ -2231,37 +2304,43 @@ def run(
...
+                s3_bucket=syft_s3_bucket,
+                s3_prefix=syft_s3_prefix,
+            )
+            resolved_aibom_source = _resolve_report_source_option(
+                module="aibom",
+                source=aibom_source,
+                local_path=aibom_results_dir,
+                s3_bucket=aibom_s3_bucket,
+                s3_prefix=aibom_s3_prefix,
+            )
```
jychp, line 2335: [suggestion] `resolved_*_source` is computed (purely for validation + deprecation-warning side effects), debug-logged, then thrown away — `Config(...)` below receives the raw legacy fields and re-runs the same `resolve_report_source_with_legacy_fields` call internally with `_warn_on_legacy_report_source=False`. Pass the resolved values as `*_source=resolved_*_source` and stop forwarding the legacy `*_results_dir` / `*_s3_bucket` / `*_s3_prefix` kwargs (or drop them from the `Config` ctor entirely). One source of truth, one normalization, one validation pass.

cartography/intel/common/object_store.py
```diff
@@ -0,0 +1,263 @@
...
+        self.source_uri = source_path
+        self._root = Path(source_path)
+
+    def list_reports(self) -> list[ReportRef]:
+        if self._root.is_file():
+            return [ReportRef(uri=str(self._root), name=self._root.name)]
+
+        refs: list[ReportRef] = []
+        for path in self._root.rglob("*"):
+            if not path.is_file() or path.name.startswith("."):
```
jychp: [suggestion] `LocalReportReader.list_reports` silently skips dot-prefixed files, but the legacy `_get_json_files_in_dir` (still used by the deprecated `sync_*_from_dir` paths) does not. A user migrating from `--trivy-results-dir`/`--aibom-results-dir` to `--*-source=/path` will silently drop any `.foo.json` reports they were ingesting before. Either drop the dotfile filter or document the new behavior in the deprecation notes for the `*_results_dir` flags.

cartography/intel/common/report_source.py
```diff
@@ -0,0 +1,232 @@
...
+    container_name: str
+    prefix: str
+
+    @property
+    def uri(self) -> str:
+        return build_azblob_source(
+            self.account_name,
+            self.container_name,
+            self.prefix,
+        )
```
jychp, line 64: [nitpick] The `raw: str` field on every report-source dataclass is stored but never read anywhere in production code or tests. Drop it; the `uri` property already produces the canonical form.

cartography/intel/common/object_store.py
```diff
@@ -0,0 +1,263 @@
...
+        pass
+
+    def read_bytes(self, ref: ReportRef) -> bytes:
+        pass
+
+
+class ObjectStoreParseError(ValueError):
+    def __init__(self, source: str, message: str) -> None:
+        super().__init__(f"{message}: {source}")
+        self.source = source
```
jychp: [nitpick] `ObjectStoreParseError` is raised on read I/O failures (S3 `BotoCoreError`/`ClientError`, GCS `GoogleAPIError`, Azure `AzureError`, local `OSError`) — the name implies parse-only failures. Either rename to `ObjectStoreError` / `ReportReadError`, or split into `ReportReadError` + `ReportParseError` so callers can distinguish "couldn't fetch" from "fetched but malformed".

jychp (changes requested):

cartography/intel/syft/__init__.py
kunaals: done — wrapped in try/except, mirrors trivy's skip-cleanup-on-any-failure (467c1ef).

cartography/intel/common/report_source.py
kunaals: done — `parse_report_source` now runs prefixes through `_normalize_cloud_prefix` for s3/gs/azblob (467c1ef).

cartography/cli.py
kunaals: partially done — cli now passes resolved values to `Config(...)` and no longer forwards legacy `*_results_dir` / `*_s3_*` kwargs. kept config's legacy-field path since it's documented public api with embedder tests; clean removal slated for v1.0.0.

cartography/intel/common/object_store.py
kunaals: done — dropped the dotfile filter from `LocalReportReader` so it matches legacy walk behavior.

cartography/intel/common/report_source.py
kunaals: done — dropped from all four dataclasses.

cartography/intel/common/object_store.py
kunaals: done — renamed to `ObjectStoreError` and reparented to `Exception` (was `ValueError`) since it covers both i/o and parse failures.

cartography/intel/aibom/__init__.py
```diff
@@ -136,19 +113,18 @@ def sync_aibom_from_dir(
...
+    refs = [
+        ReportRef(
+            uri=file_path,
+            name=os.path.relpath(file_path, results_dir),
         )
-        return
-
-    _ingest_aibom_reports(
+        for file_path in _get_json_files_in_dir(results_dir)
+    ]
```
jychp: [nitpick] Inconsistent with `sync_syft_from_dir` and `sync_docker_scout_from_dir`, which pass `LocalReportReader(results_dir)` directly to the unified sync. Here you walk files manually with `_get_json_files_in_dir` and wrap them in a `ListedReportReader`. `LocalReportReader.list_reports()` already enumerates the same files, so the indirection (and `_get_json_files_in_dir`) can go. Same applies to `sync_trivy_from_dir`.

cartography/intel/syft/__init__.py
```diff
@@ -221,33 +189,25 @@ def start_syft_ingestion(neo4j_session: Session, config: Config) -> None:
         neo4j_session: Neo4j session
         config: Configuration object with syft_results_dir or syft_s3_bucket/prefix
```
jychp: [nitpick] Stale docstring: this entry point now reads `config.syft_source`, not `syft_results_dir`/`syft_s3_bucket`. Update or drop.

jychp (approved):
Approved with 2 comments

cartography/intel/aibom/__init__.py
kunaals: done — removed the local-dir `ListedReportReader` indirection for aibom and trivy so both wrappers delegate directly to `LocalReportReader`.

cartography/intel/syft/__init__.py
kunaals: done — updated the stale syft docstring to refer to `syft_source`.

PR #2624, opened 2026-04-16
fix(entra): Skip Group when a group gets deleted between listing and fetching

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- Issue: sync_entra_groups has this flow - first get all groups (get_entra_groups) and then fetch owners and members for that group (get_group_owners and get_group_members). If a group gets deleted or is stale the get functions that fetch owners and members will propagate a 404 error and the sync crashes. 

- Fix: Catch the error gracefully, log a warning and skip the deleted group so that the sync can continue. 

### How was this tested?

- Regression test added to verify this behavior


cartography/intel/microsoft/entra/groups.py
```diff
@@ -152,13 +156,25 @@ async def sync_entra_groups(
 
     # First pass: collect groups and their owners/members
     async for group in get_entra_groups(client):
-        groups_batch.append(group)
+        # Fetch owners and members for this group.
+        # The group may no longer exist by the time we fetch details,
+        # returning a 404.  Skip it and move on.
```
jychp, line 161: Nit: double space after `404.`

cartography/intel/microsoft/entra/groups.py
```diff
@@ -152,13 +156,25 @@ async def sync_entra_groups(
...
+            owners = await call_with_retries(get_group_owners, client, group.id)
+            users, subgroups = await call_with_retries(
+                get_group_members, client, group.id
+            )
+        except APIError as e:
+            if e.response_status_code == 404:
+                logger.warning(
+                    "Group %s not found (404); skipping.",
+                    group.id,
+                )
```
jychp, line 172: Suggestion: include `group.display_name` in the warning to make prod debugging easier.

tests/integration/cartography/intel/microsoft/entra/test_groups.py
```diff
@@ -177,3 +180,85 @@ async def test_sync_entra_groups(
...
+async def _mock_get_entra_groups_with_deleted(client):
+    """Mock async generator that includes a group that no longer exists."""
+    for group in MOCK_ENTRA_GROUPS_WITH_DELETED:
+        yield group
+
+
+def mock_get_group_owners_404_side_effect(client, group_id: str) -> list[str]:
+    """Return owners for valid groups, raise 404 for the deleted group."""
+    if group_id == MOCK_DELETED_GROUP.id:
+        err = APIError()
```
jychp, line 194: Nit: for consistency with the existing `test_detected_apps.py` pattern (`APIError("fallback failed")`), consider passing a message like `APIError("not found")` here and on line 206.

cartography/intel/microsoft/entra/groups.py
```diff
@@ -152,13 +156,25 @@ async def sync_entra_groups(
...
+        # Fetch owners and members for this group.
+        # The group may no longer exist by the time we fetch details,
+        # returning a 404.  Skip it and move on.
+        try:
+            owners = await call_with_retries(get_group_owners, client, group.id)
+            users, subgroups = await call_with_retries(
+                get_group_members, client, group.id
+            )
+        except APIError as e:
+            if e.response_status_code == 404:
```
jychp, line 168: Question: should 410 Gone also be handled here? Microsoft Graph occasionally returns 410 for recently deleted objects. If 404 is the only documented response for this race, feel free to skip, just flagging.

jychp (approved):
Approved to unblock, some non blocking comments.

cartography/intel/microsoft/entra/groups.py
```diff
@@ -152,13 +156,27 @@ async def sync_entra_groups(
...
+            owners = await call_with_retries(get_group_owners, client, group.id)
+            users, subgroups = await call_with_retries(
+                get_group_members, client, group.id
+            )
+        except APIError as e:
+            if e.response_status_code in (404, 410):
+                logger.warning(
+                    "Group %s (%s) not found (%d); skipping.",
+                    group.id,
+                    group.display_name,
```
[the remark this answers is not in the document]
shyammukund: this was added due to request in above comment.

PR #2625, opened 2026-04-16
feat(jamf): sync device inventory and memberships

kunaals:
### Type of change
- [x] Documentation update

### Summary
This expands the Jamf module beyond Classic API computer groups so Cartography can ingest Jamf computer inventory, mobile device inventory, and group membership relationships.

It also keeps the earlier Jamf auth modernization in place by reusing a single API session, preferring bearer token auth, and handling both Classic and modern Jamf API base paths from the same configured base URI.

To keep the change safe for older tenants, the module now uses progressive capability fallback:
- `POST /api/v1/auth/token` with fallback to legacy Basic auth on `404/405`
- `/api/v1/groups` with fallback to Classic group endpoints on `404/405`
- computer and mobile inventory endpoints are treated as optional and skipped when the tenant does not support them

The new test fixtures preserve the real Jamf response shape while using fully sanitized dummy data.

### Related issues or links
- [PR #2625](https://github.com/cartography-cncf/cartography/pull/2625)

### How was this tested?
- `make test_lint`
- ran locally
```
INFO:cartography.intel.jamf.util:Jamf: authenticated to the Classic API using bearer token auth.
INFO:cartography.client.core.tx:Loaded <> JamfTenant nodes
INFO:cartography.client.core.tx:Loaded <> JamfComputerGroup nodes
INFO:cartography.client.core.tx:Loaded <> JamfMobileDeviceGroup nodes
INFO:cartography.graph.statement:Completed JamfComputerGroup statement #1
INFO:cartography.graph.statement:Completed JamfComputerGroup statement #2
INFO:cartography.graph.job:Finished job JamfComputerGroup
INFO:cartography.graph.statement:Completed JamfMobileDeviceGroup statement #1
INFO:cartography.graph.statement:Completed JamfMobileDeviceGroup statement #2
INFO:cartography.graph.job:Finished job JamfMobileDeviceGroup
INFO:cartography.client.core.tx:Loaded <> JamfTenant nodes
INFO:cartography.client.core.tx:Loaded <> JamfComputer nodes
INFO:cartography.graph.statement:Completed JamfComputer statement #1
INFO:cartography.graph.statement:Completed JamfComputer statement #2
INFO:cartography.graph.statement:Completed JamfComputer statement #3
INFO:cartography.graph.job:Finished job JamfComputer
INFO:cartography.client.core.tx:Loaded <> JamfTenant nodes
INFO:cartography.client.core.tx:Loaded <> JamfMobileDevice nodes
INFO:cartography.graph.statement:Completed JamfMobileDevice statement #1
INFO:cartography.graph.statement:Completed JamfMobileDevice statement #2
INFO:cartography.graph.statement:Completed JamfMobileDevice statement #3
INFO:cartography.graph.job:Finished job JamfMobileDevice

```

#### General

#### Proof of functionality

### Notes for reviewers
- The sanitized Jamf fixtures are intentionally shaped after real API payloads, but all names, serials, emails, and identifiers are dummy values.
- `JamfComputer` and `JamfMobileDevice` now contribute to the `Device` ontology via serial number and hostname/display-name matching.
- The live validation rerun after the mobile `GROUPS` fix showed full parity for `JamfMobileDevice -> JamfMobileDeviceGroup` memberships as well.

### DIFF
--- a/cartography/intel/jamf/computers.py
+++ b/cartography/intel/jamf/computers.py
@@ -70,7 +70,7 @@ def transform(api_result: list[dict[str, Any]]) -> list[dict[str, Any]]:
 
         result.append(
             {
-                "id": device.get("id"),
+                "id": device["id"],
                 "udid": device.get("udid"),
                 "name": general.get("name"),
                 "platform": general.get("platform"),
--- a/cartography/intel/jamf/computers.py
+++ b/cartography/intel/jamf/computers.py
@@ -9,6 +9,7 @@ from cartography.graph.job import GraphJob
 from cartography.intel.jamf.tenant import load_tenant
 from cartography.intel.jamf.util import get_http_status_code
 from cartography.intel.jamf.util import get_paginated_jamf_results
+from cartography.intel.jamf.util import normalize_group_id
 from cartography.models.jamf.computer import JamfComputerSchema
 from cartography.util import timeit
 
@@ -34,16 +35,6 @@ def _get_nested(data: dict[str, Any], *keys: str) -> Any:
             return None
         current = current.get(key)
     return current
-
-
-def _normalize_group_id(value: Any) -> int | str | None:
-    if value is None:
-        return None
-    if isinstance(value, int):
-        return value
-    if isinstance(value, str) and value.isdigit():
-        return int(value)
-    return value
 
 
 @timeit
@@ -121,7 +112,7 @@ def transform(api_result: list[dict[str, Any]]) -> list[dict[str, Any]]:
                 "group_ids": [
                     group_id
                     for group_id in (
-                        _normalize_group_id(group.get("groupId"))
+                        normalize_group_id(group.get("groupId"))
                         for group in (device.get("groupMemberships") or [])
                     )
                     if group_id is not None
--- a/cartography/intel/jamf/computers.py
+++ b/cartography/intel/jamf/computers.py
@@ -2,6 +2,7 @@ import logging
 from typing import Any
 
 import neo4j
+import requests
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
@@ -15,11 +16,10 @@ logger = logging.getLogger(__name__)
 
 @timeit
 def get(
+    api_session: requests.Session,
     jamf_base_uri: str,
-    jamf_user: str,
-    jamf_password: str,
 ) -> dict[str, Any]:
-    return call_jamf_api("/computergroups", jamf_base_uri, jamf_user, jamf_password)
+    return call_jamf_api("/computergroups", jamf_base_uri, api_session)
 
 
 def transform(api_result: dict[str, Any]) -> list[dict[str, Any]]:
@@ -78,14 +78,13 @@ def cleanup(
 @timeit
 def sync(
     neo4j_session: neo4j.Session,
+    api_session: requests.Session,
     jamf_base_uri: str,
-    jamf_user: str,
-    jamf_password: str,
     update_tag: int,
     common_job_parameters: dict[str, Any],
 ) -> None:
     # 1. GET
-    raw_data = get(jamf_base_uri, jamf_user, jamf_password)
+    raw_data = get(api_session, jamf_base_uri)
     # 2. TRANSFORM
     computer_groups = transform(raw_data)
     # 3. LOAD
--- a/cartography/intel/jamf/groups.py
+++ b/cartography/intel/jamf/groups.py
@@ -1,0 +1,155 @@
+import logging
+from typing import Any
+
+import neo4j
+import requests
+
+from cartography.client.core.tx import load
+from cartography.graph.job import GraphJob
+from cartography.intel.jamf.tenant import load_tenant
+from cartography.intel.jamf.util import call_jamf_api
+from cartography.intel.jamf.util import get_http_status_code
+from cartography.intel.jamf.util import get_paginated_jamf_results
+from cartography.models.jamf.computergroup import JamfComputerGroupSchema
+from cartography.models.jamf.mobiledevicegroup import JamfMobileDeviceGroupSchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+def _normalize_group_id(value: Any) -> int | str | None:
+    if value is None:
+        return None
+    if isinstance(value, int):
+        return value
+    if isinstance(value, str) and value.isdigit():
+        return int(value)
+    return value
+
+
+@timeit
+def get(
+    api_session: requests.Session,
+    jamf_base_uri: str,
+) -> list[dict[str, Any]]:
+    try:
+        return get_paginated_jamf_results(
+            "/api/v1/groups",
+            jamf_base_uri,
+            api_session,
+        )
+    except requests.HTTPError as err:
+        if get_http_status_code(err) not in {404, 405}:
+            raise
+
+        logger.info(
+            "Jamf: /api/v1/groups unavailable; falling back to legacy Classic API computer groups.",
+        )
+        classic_groups = call_jamf_api(
+            "/computergroups",
+            jamf_base_uri,
+            api_session,
+        ).get("computer_groups", [])
+        return [
+            {
+                "groupDescription": None,
+                "groupJamfProId": group.get("id"),
+                "groupName": group.get("name"),
+                "groupType": "COMPUTER",
+                "membershipCount": None,
+                "smart": group.get("is_smart"),
+            }
+            for group in classic_groups
+        ]
+
+
+def transform(
+    api_result: list[dict[str, Any]],
+) -> tuple[list[dict[str, Any]], list[dict[str, Any]]]:
+    computer_groups: list[dict[str, Any]] = []
+    mobile_device_groups: list[dict[str, Any]] = []
+    for group in api_result:
+        transformed_group = {
+            "id": _normalize_group_id(group.get("groupJamfProId")),
+            "name": group.get("groupName"),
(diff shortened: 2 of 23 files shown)


cartography/intel/jamf/groups.py
```diff
@@ -0,0 +1,155 @@
...
+) -> None:
+    GraphJob.from_node_schema(JamfComputerGroupSchema(), common_job_parameters).run(
+        neo4j_session,
+    )
+    GraphJob.from_node_schema(
+        JamfMobileDeviceGroupSchema(),
+        common_job_parameters,
+    ).run(neo4j_session)
+
+    # DEPRECATED: remove orphaned pre-migration computer groups without RESOURCE rel.
```
kunaals, line 139: this is not new - just moved

kunaals:
@cubic-dev-ai take another look pls

cartography/intel/jamf/mobile_devices.py
```diff
@@ -0,0 +1,134 @@
...
+            api_session,
+            params=_SECTION_PARAMS,
+        )
+    except requests.HTTPError as err:
+        if get_http_status_code(err) not in {404, 405}:
+            raise
+        logger.info(
+            "Jamf: /api/v2/mobile-devices/detail unavailable; skipping mobile device sync.",
+        )
+        return []
```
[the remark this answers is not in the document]
kunaals: keeping this as-is because 404/405 is the intentional capability fallback for optional jamf endpoints. other errors still raise.

cartography/intel/jamf/computers.py
```diff
@@ -2,51 +2,138 @@
...
+            api_session,
+            params=_SECTION_PARAMS,
+        )
+    except requests.HTTPError as err:
+        if get_http_status_code(err) not in {404, 405}:
+            raise
+        logger.info(
+            "Jamf: /api/v1/computers-inventory unavailable; skipping computer inventory sync.",
+        )
+        return []
```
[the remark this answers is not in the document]
kunaals: keeping this as-is because 404/405 is the intentional capability fallback for optional jamf endpoints. other errors still raise.

cartography/intel/jamf/mobile_devices.py
```diff
@@ -0,0 +1,134 @@
...
+    result: list[dict[str, Any]] = []
+    for device in api_result:
+        general = device.get("general") or {}
+        hardware = device.get("hardware") or {}
+        security = device.get("security") or {}
+        user_and_location = device.get("userAndLocation") or {}
+
+        result.append(
+            {
+                "id": device.get("mobileDeviceId"),
```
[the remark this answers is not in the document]
kunaals: good catch. fixed this.

cartography/intel/jamf/util.py
```diff
@@ -1,35 +1,152 @@
...
+    normalized_uri = _normalize_jamf_base_uri(jamf_base_uri)
+    if normalized_uri.endswith(_CLASSIC_API_PATH):
+        return normalized_uri[: -len(_CLASSIC_API_PATH)]
+    return normalized_uri
+
+
+def _get_request_base_uri(jamf_base_uri: str, api_path: str) -> str:
+    if api_path.startswith("/api/"):
+        return _get_jamf_instance_uri(jamf_base_uri)
+    return _normalize_jamf_base_uri(jamf_base_uri)
```
jychp: The Classic-API fallback in `groups.get()` calls `/computergroups` and `/mobiledevicegroups`, but the new `--jamf-base-uri` example drops `/JSSResource`. With the documented config `https://tenant.jamfcloud.com`, requests will go to `https://tenant.jamfcloud.com/computergroups` instead of `.../JSSResource/computergroups` and the fallback will 404 on the very tenants that need it.

Either keep `/JSSResource` in the documented base-URI example, or have `_get_request_base_uri` append `/JSSResource` when the path isn't `/api/*`.

cartography/intel/jamf/computers.py
```diff
@@ -2,51 +2,138 @@
...
+        general = device.get("general") or {}
+        hardware = device.get("hardware") or {}
+        operating_system = device.get("operatingSystem") or {}
+        security = device.get("security") or {}
+        disk_encryption = device.get("diskEncryption") or {}
+        user_and_location = device.get("userAndLocation") or {}
 
+        result.append(
+            {
+                "id": device.get("id"),
```
jychp: Use `device["id"]` (subscript) to fail fast on missing ids, matching `mobile_devices.transform` (and commit 671d1062). As written, any record without `id` produces a `JamfComputer` node keyed on `None`, collapsing multiple missing-id records into a single node.

cartography/intel/jamf/groups.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def transform(
+    api_result: list[dict[str, Any]],
+) -> tuple[list[dict[str, Any]], list[dict[str, Any]]]:
+    computer_groups: list[dict[str, Any]] = []
+    mobile_device_groups: list[dict[str, Any]] = []
+    for group in api_result:
+        transformed_group = {
+            "id": normalize_group_id(group.get("groupJamfProId")),
```
jychp: `group.get("groupJamfProId")` silently yields `None` for missing ids, which then becomes a group node keyed on `None`. Fail loud here: `group["groupJamfProId"]`.

cartography/intel/jamf/util.py
```diff
@@ -1,35 +1,152 @@
...
+
+    try:
+        response.raise_for_status()
+    except requests.HTTPError:
+        session.close()
+        raise
+
+    raise RuntimeError(
+        "Jamf auth token request failed without returning a handled status"
+    )
```
jychp: This `raise RuntimeError(...)` is unreachable. At this point `response.ok` is False (so `status_code >= 400`) and the 404/405 branch has already returned, so `response.raise_for_status()` above is guaranteed to raise. Drop the trailing `raise`.

cartography/intel/jamf/mobile_devices.py
```diff
@@ -0,0 +1,134 @@
...
+
+        result.append(
+            {
+                "id": device["mobileDeviceId"],
+                "display_name": general.get("displayName"),
+                "managed": general.get("managed"),
+                "supervised": general.get("supervised"),
+                "last_inventory_update_date": general.get("lastInventoryUpdateDate"),
+                "last_enrolled_date": general.get("lastEnrolledDate"),
+                "platform": device.get("deviceType"),
```
jychp, line 66: `deviceType` produces `"iPhone"`/`"iPad"` here, while `computers.py` populates `platform` with Jamf's OS family (e.g. `"macOS"`). The ontology `Device.platform` mapping pulls from both — mixing device-model names with OS family names will make cross-module queries inconsistent. Consider deriving `"iOS"`/`"iPadOS"` from `deviceType` instead.

cartography/models/ontology/mapping/data/devices.py
```diff
@@ -226,6 +226,50 @@
...
+                    required=True,
+                ),
+            ],
+        ),
+        OntologyNodeMapping(
+            node_label="JamfMobileDevice",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="hostname",
+                    node_field="display_name",
```
jychp: `display_name` on a Jamf mobile device is user-set (e.g. `"Bart's iPhone"`) and rarely a DNS hostname. Mapping it to `hostname` — plus the hostname matchlink in `ontology/device.py` — will produce very few real matches and risks false positives. Consider dropping the hostname mapping for mobile devices and relying on `serial_number`.

cartography/intel/jamf/mobile_devices.py
```diff
@@ -0,0 +1,134 @@
...
+) -> list[dict[str, Any]]:
+    try:
+        return get_paginated_jamf_results(
+            "/api/v2/mobile-devices/detail",
+            jamf_base_uri,
+            api_session,
+            params=_SECTION_PARAMS,
+        )
+    except requests.HTTPError as err:
+        if get_http_status_code(err) not in {404, 405}:
```
jychp, line 42: may be we should move that logic to the get_paginatied_jamf_results, that way we only define once the "soft failing" error codes

jychp (changes requested):

cartography/intel/jamf/util.py
kunaals: good catch. fixed this.

cartography/intel/jamf/computers.py
kunaals: good catch. fixed this.

cartography/intel/jamf/groups.py
kunaals: good catch. fixed this.

cartography/intel/jamf/util.py
kunaals: good catch. fixed this.

cartography/intel/jamf/mobile_devices.py
kunaals: keeping this as-is because the live jamf tenant is returning ios for devicetype, not iphone/ipad, so this would be speculative normalization.

cartography/models/ontology/mapping/data/devices.py
kunaals: good catch. dropped the mobile hostname mapping and matchlink.

cartography/intel/jamf/mobile_devices.py
kunaals: keeping this local because 404/405 handling is endpoint-specific capability fallback, not generic pagination behavior.

jychp (approved):
LGTM

PR #2627, opened 2026-04-16
feat(kubernetes): add IRSA service account role mapping

jychp:
### Type of change

- [x] Documentation update

### Summary
- add IRSA annotation parsing to Kubernetes service account syncs via `eks.amazonaws.com/role-arn`
- model `(:KubernetesServiceAccount)-[:ASSUMES_ROLE]->(:AWSRole)` using the existing schema-based matcher pattern used for Kubernetes users/groups
- document the new field/relationship and add RBAC integration coverage for annotated and unannotated service accounts

### Related issues or links

- Fixes #

### Breaking changes
None.

### How was this tested?
- `uv run --frozen pytest -vvv tests/integration/cartography/intel/kubernetes/test_pods.py tests/integration/cartography/intel/kubernetes/test_rbac.py`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
- This intentionally follows the existing Kubernetes user/group AWS-role mapping behavior: the relationship is schema-based and only materializes when the target `AWSRole` already exists in the graph.

### DIFF
--- a/tests/integration/cartography/intel/kubernetes/test_rbac.py
+++ b/tests/integration/cartography/intel/kubernetes/test_rbac.py
@@ -21,6 +21,7 @@ from tests.data.kubernetes.rbac import KUBERNETES_CLUSTER_1_ROLE_BINDINGS_RAW
 from tests.data.kubernetes.rbac import KUBERNETES_CLUSTER_1_ROLE_IDS
 from tests.data.kubernetes.rbac import KUBERNETES_CLUSTER_1_ROLES_RAW
 from tests.data.kubernetes.rbac import KUBERNETES_CLUSTER_1_SERVICE_ACCOUNT_IDS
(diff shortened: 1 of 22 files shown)


kunaals (approved):
.

tests/integration/cartography/intel/kubernetes/test_rbac.py
```diff
@@ -124,6 +125,21 @@ def test_sync_rbac_end_to_end(
     )
     assert expected_service_accounts.issubset(actual_service_accounts)
 
+    expected_service_account_role_arns = {
```
kunaals, line 136: nit: could we make this a positive assertion instead? right now the new coverage only checks that `ASSUMES_ROLE` is empty, so the irsa mapping itself isn’t actually exercised. if we seed an `AWSRole` here and assert the annotated service account links to it, this test would protect the new behavior.

kunaals (approved):
approving to unblock, but one small test improvement below would make the new irsa mapping coverage much stronger.

PR #2628, opened 2026-04-16
feat(vercel): add Vercel ingestion module

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds a new intel module that syncs [Vercel](https://vercel.com) into the graph. Coverage spans 20 node types across identity, projects, deployments, DNS, secrets metadata, firewall, integrations, RBAC, tokens, webhooks, log drains, networks, aliases, and edge configs.

The module goes beyond inventory and wires 12 cross-resource relationships so the graph reflects actual usage, plus ontology integration so Vercel entities participate in cross-provider queries.

**Nodes:**
- `VercelTeam` (Tenant), `VercelUser` (UserAccount), `VercelProject`, `VercelDeployment`
- `VercelDomain` (DNSZone), `VercelDNSRecord` (DNSRecord)
- `VercelEnvironmentVariable`, `VercelSharedEnvironmentVariable` (values never stored)
- `VercelFirewallConfig`, `VercelFirewallBypassRule`
- `VercelIntegration`, `VercelAccessGroup` (Group), `VercelAuthToken` (APIKey)
- `VercelWebhook`, `VercelLogDrain`, `VercelSecureComputeNetwork`, `VercelAlias`
- `VercelEdgeConfig`, `VercelEdgeConfigToken` (token values never stored)

**Notable relationships:**
- `VercelUser -MEMBER_OF-> VercelTeam` (with role/confirmed/joined_from on the rel)
- `VercelProject -HAS_DOMAIN-> VercelDomain` (per-project redirect/branch/verified on the rel, MatchLink pattern; also upserts minimal `VercelDomain` for auto-generated `*.vercel.app` and CNAMEd external domains)
- `VercelDeployment -CREATED_BY-> VercelUser`, `VercelFirewallBypassRule -CREATED_BY-> VercelUser`
- `VercelEnvironmentVariable -REFERENCES-> VercelEdgeConfig`
- `VercelIntegration -CONFIGURED_FOR-> VercelProject`, `VercelAccessGroup -HAS_MEMBER-> VercelUser`, `VercelAccessGroup -HAS_ACCESS_TO-> VercelProject`
- `VercelWebhook -WATCHES-> VercelProject`, `VercelLogDrain -MONITORS-> VercelProject`, `VercelSecureComputeNetwork -CONNECTS-> VercelProject`
- `VercelAlias -DEPLOYED_TO-> VercelDeployment`, `VercelAlias -BELONGS_TO_PROJECT-> VercelProject`

**Ontology semantic label mappings:**
- `VercelTeam` -> `Tenant`
- `VercelUser` -> `UserAccount` (email, username, fullname, active)
- `VercelDomain` -> `DNSZone` (name, public)
- `VercelAuthToken` -> `APIKey` (name, created_at, expires_at, last_used_at)
- `VercelAccessGroup` -> `Group` (name)

### Related issues or links

- Documentation: https://vercel.com/docs/rest-api
- OpenAPI: https://openapi.vercel.sh/

### Breaking changes

None. New module is opt-in via `--vercel-token-env-var` / `--vercel-team-id` CLI flags.

### How was this tested?

- 20 new integration tests covering every sub-module, all passing against a local Neo4j testcontainer
- 1288 unit tests passing (including `test_doc` which validates schema docs are registered)
- `make lint` clean (isort, black, flake8, pyupgrade, mypy)

Commands:
```
uv run --frozen pytest tests/integration/cartography/intel/vercel/
uv run --frozen pytest tests/unit
make test_lint
```

#### General

#### Proof of functionality

### Notes for reviewers

- Per-project sub-resources (deployments, env vars, project domains, firewall config, firewall bypass rules) and per-domain DNS records use the `for parent in parents: sync(...)` pattern established by existing modules like Cloudflare.
- `VercelProject -[HAS_DOMAIN]-> VercelDomain` uses the MatchLink pattern (with a composite minimal schema upsert for the node) so a single `VercelDomain` represents a domain regardless of source (team-level `/v5/domains` or project-level `/v9/projects/{id}/domains`). Per-project configuration lives on the relationship.
- Environment variable and token values are intentionally never stored — only metadata (key, type, target, created_at, expires_at, etc.).
- Access groups enrich each group via sub-fetches of `/members` and `/projects` to populate `HAS_MEMBER` and `HAS_ACCESS_TO` relationships.
- Deployments are limited to 100 most recent per project to avoid unbounded historical growth.
- `VercelFirewallConfig` has a synthesized ID (`{project_id}_firewall`) since the API returns a singleton config object per project.

Sync logs from a real environment will be attached separately.


jychp:
Screenshot 2026-04-17 at 11 56 05 AM

jychp:
Screenshot 2026-04-17 at 11 59 33 AM

cartography/intel/vercel/authtokens.py
```diff
@@ -0,0 +1,73 @@
...
+
+@timeit
+def get(
+    api_session: requests.Session,
+    base_url: str,
+    team_id: str,
+) -> list[dict[str, Any]]:
+    return paginated_get(
+        api_session,
+        f"{base_url}/v5/user/tokens",
```
kunaals: i think this endpoint being /v5/user/tokens makes the ownership a little ambiguous. should these maybe hang off users rather than teams?

cartography/intel/vercel/accessgroups.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+        group["member_ids"] = [m["uid"] for m in members]
+
+        projects = paginated_get(
+            api_session,
+            f"{base_url}/v1/access-groups/{group_id}/projects",
+            "projects",
+            team_id,
+        )
+        group["project_ids"] = [p["projectId"] for p in projects]
```
kunaals: i think we may be dropping the most important part of the access-group model here, which is the per-project role assignment.

cartography/models/vercel/edgeconfig.py
```diff
@@ -0,0 +1,48 @@
...
+    properties: VercelEdgeConfigToTeamRelProperties = (
+        VercelEdgeConfigToTeamRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class VercelEdgeConfigSchema(CartographyNodeSchema):
+    label: str = "VercelEdgeConfig"
+    properties: VercelEdgeConfigNodeProperties = VercelEdgeConfigNodeProperties()
+    sub_resource_relationship: VercelEdgeConfigToTeamRel = VercelEdgeConfigToTeamRel()
```
kunaals, line 48: i think we're maybe missing the project usage side of edge config here. the team ownership makes sense, but the project linkage feels important too.

cartography/models/vercel/domain.py
```diff
@@ -0,0 +1,105 @@
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: VercelDomainToTeamRelProperties = VercelDomainToTeamRelProperties()
+
+
+@dataclass(frozen=True)
+class VercelDomainSchema(CartographyNodeSchema):
+    label: str = "VercelDomain"
+    properties: VercelDomainNodeProperties = VercelDomainNodeProperties()
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["DNSZone"])
```
kunaals: i think this may be conflating a few different things. not every project-attached hostname feels like a dnszone to me, especially if vercel isn't authoritative for dns.

cartography/models/vercel/securecomputenetwork.py
```diff
@@ -0,0 +1,70 @@
...
+
+@dataclass(frozen=True)
+# (:VercelSecureComputeNetwork)-[:CONNECTS]->(:VercelProject)
+class VercelNetworkToProjectRel(CartographyRelSchema):
+    target_node_label: str = "VercelProject"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("project_ids", one_to_many=True)},
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "CONNECTS"
```
kunaals, line 65: i think this may be flattening the secure compute attachment a bit too much. my read of the docs is that it's environment-specific and can be active/passive.

kunaals (commented):
few questions

cartography/models/vercel/edgeconfig.py
jychp: We have that link: `(project)-[:RESOURCE]->(env)-[:REFERENCES]->(edgeconfig)`

kunaals: ah okay, i missed the env -> edge config hop. that does give us a project usage path, so i'll resolve this comment.

kunaals (approved):
approving to unblock as most of the questions that i asked earlier are nits and much of this will get sorted out as large organizations use this module and we address feedback

cartography/models/vercel/edgeconfig.py
jychp: The project ↔ edge-config link is already represented indirectly via env vars: `(:VercelProject)-[:RESOURCE]->(:VercelEnvironmentVariable {edge_config_id})-[:REFERENCES]->(:VercelEdgeConfig)`. That's how Vercel itself exposes the relationship (there's no dedicated `projects-by-edge-config` or `edge-configs-by-project` endpoint — a project "uses" an Edge Config iff one of its env vars has `edgeConfigId` set). Let me know if you'd prefer a direct `USED_BY` edge too, but it would be fully derivable from the env var path.

PR #2635, opened 2026-04-17
feat(aws-ssm): ingest allowlisted public parameters

kunaals:
### Type of change

### Summary
Adds ingestion of AWS-managed public SSM parameters for Bottlerocket and EKS optimized AMI recommendations, modeled as shared regional `SSMParameter` nodes so the same public data is fetched once per region and not duplicated per account. Private SSM parameters continue to behave as account-owned resources, while public `/aws/service/...` parameters use their AWS ARN as node identity and remain queryable through the existing `SSMParameter` label.

### Breaking changes
None.

### How was this tested?
- added unit and integration tests for the shared public SSM path
- ran focused pytest locally
- validated read-only against live AWS public SSM data using a disposable local Neo4j database

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
- Public SSM parameter ingestion now defaults to the Bottlerocket and EKS optimized AMI namespaces, with an explicit empty-string escape hatch to disable it.
- AWS-managed public parameters are synced once per region as shared `PublicSSMParameter:SSMParameter` nodes with ARN identity.
- Private SSM parameters continue to attach to `AWSAccount` with `RESOURCE`; shared public parameters intentionally do not.


cartography/intel/aws/ssm.py
```diff
@@ -123,11 +153,75 @@ def get_ssm_parameters(
...
+                ):
+                    continue
+                if (
+                    not ingest_secure_strings
+                    and parameter.get("Type") == "SecureString"
+                ):
+                    logger.debug(
+                        "Skipping SecureString SSM parameter %s in region %s.",
+                        parameter_name,
+                        region,
```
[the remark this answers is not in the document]
kunaals: fixed, the debug log no longer includes the parameter name.

cartography/intel/aws/__init__.py
```diff
@@ -414,13 +415,26 @@ def _perform_aws_analysis(
...
 @timeit
 def start_aws_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
+    aws_ssm_public_parameter_prefix_allowlist = (
+        config.aws_ssm_public_parameter_prefix_allowlist
+        if config.aws_ssm_public_parameter_prefix_allowlist is not None
+        else os.getenv("AWS_SSM_PUBLIC_PARAMETER_PREFIX_ALLOWLIST")
+    )
+    aws_ssm_ingest_secure_strings = (
+        config.aws_ssm_ingest_secure_strings
+        if config.aws_ssm_ingest_secure_strings
```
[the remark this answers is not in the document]
kunaals: good catch, fixed this by treating only none as unset so an explicit false now wins over the env var.

cartography/models/aws/ssm/parameters.py
```diff
@@ -15,8 +15,9 @@
 class SSMParameterNodeProperties(CartographyNodeProperties):
 
     arn: PropertyRef = PropertyRef("ARN", extra_index=True)
-    id: PropertyRef = PropertyRef("ARN")
+    id: PropertyRef = PropertyRef("Id")
```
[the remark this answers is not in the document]
kunaals: updated the aws schema docs for the deterministic id and the new value field.

tests/unit/cartography/intel/aws/test_ssm.py
```diff
@@ -0,0 +1,124 @@
+from datetime import datetime
```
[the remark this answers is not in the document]
kunaals: added integration coverage for allowlisted public parameters and tightened the helper test around the paginated requests.

cartography/intel/aws/__init__.py
```diff
@@ -222,6 +224,65 @@ def _autodiscover_account_regions(
...
+
+    first_profile_name, first_account_id = next(iter(aws_accounts.items()))
+    boto3_session = _get_boto3_session_for_profile(
+        default_boto3_session,
+        first_profile_name,
+    )
+    regions = configured_regions or _autodiscover_account_regions(
+        boto3_session,
+        first_account_id,
+    )
```
jychp, line 267: It seems that you pins the public SSM sync to the first configured profile. There is several cases where that can drop the shared parameters:
- aws_sync_all_profiles=True witjout an explicit aws_regions
- profiles with different opt-in regions
- first profile that lacks `ec2:DescribeRegions` / `ssm:GetParametersByPath`

cartography/intel/aws/ssm.py
```diff
@@ -123,6 +160,58 @@ def get_ssm_parameters(
...
+            Recursive=True,
+            WithDecryption=False,
+            PaginationConfig={"PageSize": 10},
+        ):
+            for parameter in page.get("Parameters", []):
+                parameter_name = parameter.get("Name", "")
+                if not _parameter_matches_allowlist_prefixes(
+                    parameter_name,
+                    [prefix],
+                ):
```
jychp, line 190: This filter is dead code. `paginator.paginate(Path=prefix, Recursive=True)` already guarantees every returned parameter name starts with `prefix`, and the allowlist passed in here is a single-element list with exactly that same prefix. The `.startswith(prefix)` check can't fail. Drop it (and the helper call) to simplify the loop.

cartography/intel/aws/ssm.py
```diff
@@ -123,6 +160,58 @@ def get_ssm_parameters(
...
+                    and parameter.get("Type") == "SecureString"
+                ):
+                    logger.debug(
+                        "Skipping SecureString SSM parameter in region %s.",
+                        region,
+                    )
+                    continue
+                # Public AWS-managed parameters are expected to be plain Strings.
+                # If SecureStrings are explicitly allowed, WithDecryption remains
+                # disabled here so values stay in the service-returned form.
```
jychp, line 203: `aws_ssm_ingest_secure_strings` is misleading as wired. It only gates the skip of `SecureString` entries on the public path, but the adjacent comment states public AWS-managed params are always plain Strings so no entries are ever skipped in practice. Even when `True`, `WithDecryption=False` is hard-coded, so no secret value is ever exposed either. Consider dropping the flag.

cartography/models/aws/ssm/parameters.py
```diff
@@ -82,3 +84,18 @@ class SSMParameterSchema(CartographyNodeSchema):
...
+    label: str = "PublicSSMParameter"
+    properties: SSMParameterNodeProperties = SSMParameterNodeProperties()
+    sub_resource_relationship: None = None
+    scoped_cleanup: bool = False
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["SSMParameter"])
+    other_relationships: OtherRelationships = OtherRelationships(
+        [
+            SSMParameterToKMSKeyRel(),
+        ],
+    )
```
jychp, line 101: `SSMParameterToKMSKeyRel` on this schema is dead weight. Public `/aws/service/...` parameters are plain Strings, `KMSKeyIdShort` is always `None`, so the edge never resolves. Drop `other_relationships` here to keep the shared-node model minimal.

cartography/models/aws/ssm/parameters.py
```diff
@@ -17,6 +18,7 @@ class SSMParameterNodeProperties(CartographyNodeProperties):
     arn: PropertyRef = PropertyRef("ARN", extra_index=True)
     id: PropertyRef = PropertyRef("ARN")
     name: PropertyRef = PropertyRef("Name")
+    value: PropertyRef = PropertyRef("Value")
```
jychp, line 21: Nit: `value` is now on the shared node properties, but `describe_parameters` (private path) never returns a `Value`, so private `SSMParameter` nodes will carry `value=null`. Worth calling that out in the schema docs so users don't expect meaningful values on private parameters.

jychp (changes requested):

cartography/intel/aws/__init__.py
```diff
@@ -222,6 +224,65 @@ def _autodiscover_account_regions(
...
+    aws_accounts: Dict[str | None, str],
+    requested_syncs: List[str],
+    common_job_parameters: Dict[str, Any],
+    configured_regions: list[str] | None,
+    aws_best_effort_mode: bool,
+) -> None:
+    if "ssm" not in requested_syncs:
+        return
+
+    first_profile_name, first_account_id = next(iter(aws_accounts.items()))
```
jychp: blocking: Pinning the shared public SSM sync to the first profile drops parameters in several cases: `aws_sync_all_profiles=True` without an explicit `aws_regions`, profiles with different opt-in regions, and a first profile that lacks `ec2:DescribeRegions` / `ssm:GetParametersByPath`. Iterate profiles/regions (union of enabled regions) instead of arbitrarily picking the first.

cartography/intel/aws/ssm.py
```diff
@@ -123,6 +153,50 @@ def get_ssm_parameters(
     return ssm_parameters_data
 
 
+@timeit
+@aws_handle_regions
+def get_public_ssm_parameters_by_path(
+    boto3_session: boto3.session.Session,
+    region: str,
+    allowlist_prefixes: list[str],
+    ingest_secure_strings: bool,
```
jychp: suggestion: `ingest_secure_strings` is effectively dead as wired. `WithDecryption=False` is hard-coded (no secret value is ever exposed) and AWS-managed public params under `/aws/service/...` are plain Strings, so the SecureString skip never triggers. Drop the flag and its config plumbing.

cartography/models/aws/ssm/parameters.py
```diff
@@ -82,3 +84,18 @@ class SSMParameterSchema(CartographyNodeSchema):
...
+
+@dataclass(frozen=True)
+class PublicSSMParameterSchema(CartographyNodeSchema):
+
+    label: str = "PublicSSMParameter"
+    properties: SSMParameterNodeProperties = SSMParameterNodeProperties()
+    sub_resource_relationship: None = None
+    scoped_cleanup: bool = False
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["SSMParameter"])
+    other_relationships: OtherRelationships = OtherRelationships(
```
jychp: suggestion: `SSMParameterToKMSKeyRel` never resolves here. Public `/aws/service/...` params are plain Strings, so `KMSKeyIdShort` is always `None` and the `ENCRYPTED_BY` edge is dead weight. Drop `other_relationships` to keep the shared-node model minimal.

docs/root/modules/aws/schema.md
```diff
@@ -5383,10 +5383,11 @@ Representation of an AWS Systems Manager Parameter as returned by the [`describe
 |-------|-------------|
 | **firstseen**| Timestamp of when a sync job first discovered this node  |
 | lastupdated |  Timestamp of the last time the node was updated |
-| **id** | The ARN of the parameter |
+| **id** | The AWS parameter ARN. |
 | region | The region of the parameter. |
 | **arn** | The Amazon Resource Name (ARN) of the parameter. |
 | name | The parameter name. |
+| value | The parameter value. |
```
jychp: nitpick: `value` is only populated on public parameters. Private `SSMParameter` nodes come from `describe_parameters`, which never returns a value, so they always carry `value=null`. Call that out here so users don't expect meaningful values on private parameters.

cartography/config.py
```diff
@@ -458,6 +467,10 @@ def __init__(
         )
         self.experimental_aws_inspector_batch = experimental_aws_inspector_batch
         self.aws_tagging_api_cleanup_batch = aws_tagging_api_cleanup_batch
+        self.aws_ssm_public_parameter_prefix_allowlist = (
```
jychp, line 646: question: `aws_ssm_public_parameter_prefix_allowlist` and `aws_ssm_ingest_secure_strings` are added to `Config` but have no corresponding `cli.py` arguments, so a CLI user can only set them via the `AWS_SSM_*` env vars. Intentional, or is the CLI wiring missing? Other AWS config options are exposed as CLI flags.

jychp (changes requested):

cartography/models/aws/ssm/parameters.py
```diff
@@ -82,3 +84,18 @@ class SSMParameterSchema(CartographyNodeSchema):
...
     )
+
+
+@dataclass(frozen=True)
+class PublicSSMParameterSchema(CartographyNodeSchema):
+
+    label: str = "PublicSSMParameter"
+    properties: SSMParameterNodeProperties = SSMParameterNodeProperties()
+    sub_resource_relationship: None = None
+    scoped_cleanup: bool = False
```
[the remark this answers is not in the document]
kunaals: keeping this unscoped because these are aws-managed regional catalog nodes shared across accounts; account-scoped cleanup could remove data still used by another account.

cartography/models/aws/ssm/parameters.py
```diff
@@ -82,3 +84,18 @@ class SSMParameterSchema(CartographyNodeSchema):
...
         ],
     )
+
+
+@dataclass(frozen=True)
+class PublicSSMParameterSchema(CartographyNodeSchema):
+
+    label: str = "PublicSSMParameter"
+    properties: SSMParameterNodeProperties = SSMParameterNodeProperties()
+    sub_resource_relationship: None = None
```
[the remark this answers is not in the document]
kunaals: keeping this without a tenant relationship because the nodes are aws-managed regional catalog data shared across accounts.

cartography/intel/aws/ssm.py
kunaals: good catch. removed the redundant prefix filter.

cartography/intel/aws/ssm.py
kunaals: good catch. removed the secure-string flag and its plumbing.

cartography/models/aws/ssm/parameters.py
kunaals: good catch. documented that values are only populated for public parameters.

cartography/intel/aws/__init__.py
kunaals: good catch. this now unions regions across profiles and falls back across profile sessions per region.

cartography/intel/aws/ssm.py
kunaals: good catch. removed the secure-string flag and its config plumbing.

cartography/models/aws/ssm/parameters.py
kunaals: good catch. removed the public kms relationship; the private secure-string relationship remains.

docs/root/modules/aws/schema.md
kunaals: good catch. clarified this in the schema docs.

cartography/config.py
kunaals: good catch. added the public allowlist cli flag and removed the dead secure-string option.

jychp (approved):

PR #2636, opened 2026-04-17
fix(jamf): map mobile display fields into device ontology

kunaals:
### Type of change
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [x] Documentation update
- [ ] Other (please describe):

### Summary
Jamf mobile devices already ingest useful display-facing fields, but the `Device` ontology was only carrying `serial_number`, `model`, `os_version`, and `platform` forward for `JamfMobileDevice` nodes. That left canonical `(:Device)` rows with empty `hostname` and `os` values even though the upstream Jamf mobile record already had `display_name` and `platform`.

This change maps:
- `JamfMobileDevice.display_name -> Device.hostname`
- `JamfMobileDevice.platform -> Device.os`

Why this belongs in Cartography:
- these values already exist on the synced Jamf mobile node
- downstream consumers that read the canonical `Device` ontology should not need Jamf-specific fallbacks
- this improves ontology completeness without hardcoding manufacturer/OS family values or adding speculative matchlinks

### Related issues or links
- Fixes #

### How was this tested?
- `uv run pytest tests/integration/cartography/intel/ontology/test_devices.py tests/integration/cartography/intel/jamf/test_jamf.py tests/unit/cartography/intel/ontology/test_ontology_mapping.py -q`
- validated against a real Jamf tenant using a disposable local Neo4j database and the device ontology sync

#### General
- [x] I have read the [contributing guidelines](https://cartography-cncf.github.io/cartography/dev/developer-guide.html).
- [x] The linter passes locally (`make lint`).
- [x] I have added/updated tests that prove my fix is effective or my feature works.

#### Proof of functionality
- [x] New or updated unit/integration tests.

#### If you are adding or modifying a synced entity
- [ ] Included Cartography sync logs from a real environment demonstrating successful synchronization of the new/modified entity. Logs should show:
  - The sync job starting and completing without errors
  - The number of nodes/relationships created or updated

#### If you are changing a node or relationship
- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules).
- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md).

### Notes for reviewers
This intentionally does not:
- hardcode `manufacturer = Apple`
- hardcode `os = iOS`
[...]

PR #2638, opened 2026-04-17
fix(jamf): normalize mobile os before ontology mapping

kunaals:
### Type of change
- [x] Documentation update

### Summary
Follow-up to #2636.

`JamfMobileDevice` now contributes `display_name -> Device.hostname`, which improves the canonical `Device` rows for Jamf mobile inventory. But the current `Device.os <- JamfMobileDevice.platform` mapping still relies on Jamf's raw `deviceType` field, and that field is not consistently an OS-family value across environments and fixtures.

In the live tenant we validated, `deviceType` comes through as `iOS`, so the existing mapping happens to work. In Cartography's sanitized integration fixtures, the same field is `iPhone` / `iPad`, which means the canonical ontology can end up with semantically incorrect `Device.os` values even though the data is otherwise valid.

This follow-up fixes that at the Jamf provider layer by:
- normalizing known Jamf mobile `deviceType` values into an explicit `JamfMobileDevice.os` field
- mapping `Device.os` from that normalized provider field instead of from `platform`
- keeping `platform` unchanged so the raw Jamf device family is still preserved separately

Why this belongs here:
- it keeps the `Device` ontology aligned with its own `os` vs `platform` semantics
- it avoids hardcoded canonical assumptions in the ontology layer itself
- it keeps downstream consumers from having to special-case Jamf mobile device-family strings

### Related issues or links
- Follow-up to #2636

### How was this tested?
- tested locally on Jamf tenant

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/jamf/mobile_devices.py
+++ b/cartography/intel/jamf/mobile_devices.py
@@ -26,6 +26,33 @@ _SECTION_PARAMS = {
 }
 
 
+def _normalize_mobile_os(device_type: str | None) -> str | None:
+    """Normalize Jamf mobile device family values into OS-family values.
+
+    Jamf's v2 mobile inventory uses ``deviceType`` for mobile family/platform
+    values. In practice this may already be an OS-family value such as ``iOS``,
+    but some tenants and fixtures return hardware-family values such as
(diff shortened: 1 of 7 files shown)


jychp (approved):
LGTM

PR #1043, opened 2022-12-09
Fix #1042: AWS: only specify profile_name when syncing multiple accounts

achantavy:
Bandaid fix for #1042: specifies `profile_name` when creating a boto3 session only if we are syncing more than one AWS account. This ensures that boto3 is able to properly find credentials.

Issue #1042, opened 2022-12-09
boto3 fails to use environment for creds when profile_name is specified


achantavy:
# Description
> What issue is being seen?  Describe what should be happening instead of the bug, for example: Cartography should not crash, the expected value isn't returned, the data schema is wrong, etc.

I am unable to run the cartography CLI with AWS credentials set in environment variables. I believe this is because boto3 fails to infer that I want to use env vars when a profile_name is specified.

# To Reproduce

> Steps to reproduce the behavior.  Provide all data and inputs required to reproduce the issue.

## Expected:

If I set my boto3 environment variables correctly, I am able to open a python console and call AWS API calls successfully:

```
>>> import boto3
>>> sess = boto3.Session()
>>> client = sess.client('ec2')
>>> client.describe_regions()
{'Regions': [{'Endpoint': 'ec2.ap-south-1.amazonaws.com', 'RegionName': 'ap-south-1', 'OptInStatus': 'opt-in-not-required'},
...
```

Therefore I expect that when I run the `cartography` command, the AWS sync will succeed.

## Actual:

I get a fat stacktrace:

```
$ $ AWS_PROFILE=myprofile cartography --aws-requested-syncs s3
INFO:cartography.sync:Starting sync with update tag '1670565876'
INFO:cartography.sync:Starting sync stage 'create-indexes'
INFO:cartography.intel.create_indexes:Creating indexes for cartography node types.
INFO:cartography.sync:Finishing sync stage 'create-indexes'
INFO:cartography.sync:Starting sync stage 'aws'
INFO:cartography.intel.aws:{'ROLE_NAME': 'ACCOUNT_NUM'}
INFO:cartography.intel.aws:Syncing AWS accounts: ACCOUNT_NUM
INFO:cartography.intel.aws:Syncing AWS account with ID 'ACCOUNT_NUM' using configured profile 'ROLE_NAME'.
INFO:cartography.intel.aws:Trying to autodiscover accounts.
WARNING:botocore.credentials:Refreshing temporary credentials failed during mandatory refresh period.
Traceback (most recent call last):
  File "/PATH/cart/lib/python3.10/site-packages/botocore/credentials.py", line 570, in _protected_refresh
    metadata = self._refresh_using()
  File "/PATH/cart/lib/python3.10/site-packages/botocore/credentials.py", line 717, in fetch_credentials
    return self._get_cached_credentials()
[...]

PR #2641, opened 2026-04-17
fix(aws): honor configured profile even when only one is discovered

### LINKED ISSUE #1142: Allow the use of aws profile even if we have one account
*Description*:

We actually can't use profile when using `--aws-sync-all-profiles` with one account, the default profile will be used
*To Reproduce*:

Configure config file with one account:
```
[profile XXXXX]
role_arn = arn:aws:iam::XXXXX:role/XXXXX
region = eu-west-1
output = json
credential_source = EcsContainer
```
then launch cartography cli
```
cartography -v  --aws-sync-all-profiles --neo4j-uri bolt://$NEO4J_URI:7687  --neo4j-password-env-var=NEO4J_PASSWORD --neo4j-user=neo4j
```
*Logs*:
```
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the ListUsers operation: User: arn:aws:sts::XXXX:assumed-role/XXXX is not authorized to perform: iam:ListUsers on resource: arn:aws:iam::XXXXX:user/ because no identity-based policy allows the iam:ListUsers action
```

>  - Cartography release version [0.75.0] 
>  - Python version: [3.9.16]

*Additional context*:
A pull request will be proposed right away

### LINKED ISSUE #1185: Cartography Multiple AWS Account Setup always discover hub account
**Cartography Multiple AWS Account Setup always discover hub account even though it shows acc id of spoke account on terminal but it actually discovers hub account**

*Description*:
> I am trying to use cartography for multiple aws accounts, i want to use hub account to deploy my cartography service and neo4j db (in ec2) and want to discover resources of spoke accounts using sts assume role. I am following the [documentation](https://lyft.github.io/cartography/modules/aws/config.html#multiple-aws-account-setup) step by step, all the permissions and policies are set accordingly but It fails to discover spoke account resources and always bring hub account resources on neo4j DB. 
The terminal show account id of spoke account, but its misleading, it actually discover hub account. 

Keep in mind that my hub and spoke account are not inside one organisation, but rather are two separate AWS accounts. 

*Screenshots*:
> attaching some screenshots below for reference

>  - Cartography release version = '0.79.2'
>  - Python version: Python 3.10.6
>  - OS Ubuntu (ec2 T2.medium)

Screenshot from 2023-05-30 01-27-15
Screenshot from 2023-05-30 01-29-52
Screenshot from 2023-05-30 01-27-35
23-05-30 01-32-06](https://github.com/lyft/cartography/assets/18714447/1cba4a8e-fe1a-48eb-86f3-d61a25e58445)
Screenshot from 2023-05-30 01-29-04

jychp:
### Type of change
- [ ] Documentation update

### Summary

`_sync_multiple_accounts` used to fall back to the default boto3 session whenever exactly one account was being synced, which broke two documented flows:

- **#1142**: `--aws-sync-all-profiles` with a single configured profile silently reverted to the default profile (raising `AccessDenied` because the default creds had no IAM permissions on the target account).
- **#1185**: Hub/spoke STS assume-role setups logged the spoke account ID but actually ingested the hub's data because the spoke profile was dropped in favor of the default session.

Both had the same root cause: the `num_accounts == 1` discriminator was introduced in #1043 to dodge `ProfileNotFound` when env-var-only credentials are used without `~/.aws/config`, but it was too aggressive — it also dropped legitimately configured profiles whenever only one was discovered.

This PR replaces the discriminator with an explicit `use_explicit_profile` flag plumbed from `config.aws_sync_all_profiles` through `start_aws_ingestion` → `_sync_multiple_accounts`:

- If the user passed `--aws-sync-all-profiles`, honor each discovered profile by name (fixes #1142 and #1185).
- Otherwise, keep using the default session (preserves the #1042 fix for env-var-only credentials).

### Release note

Users running `--aws-sync-all-profiles` against a single configured profile may observe a change in auth behavior after upgrading: cartography now correctly uses the named profile instead of silently falling back to the default session. If you were unintentionally relying on that fallback (e.g., the default session happened to have the right credentials), you'll need to either drop `--aws-sync-all-profiles` or make sure the named profile resolves to credentials with `SecurityAudit` on the target account.

### Related issues or links

- Fixes #1142
- Fixes #1185
- Preserves behavior from #1043 / #1042

### Breaking changes

None at the API level. See the release note above for a behavioral change that may affect misconfigured single-profile setups.

### How was this tested?

- `uv run --frozen pytest tests/integration/cartography/intel/aws/test_init.py` → 11 passed.
- `uv run --frozen pre-commit run --files cartography/intel/aws/__init__.py tests/integration/cartography/intel/aws/test_init.py` → all hooks passed.
- Added three regression tests:
  - `test_sync_multiple_accounts_single_profile_uses_profile_name` — asserts that with `use_explicit_profile=True` and a single-entry accounts dict, `boto3.Session` / `aioboto3.Session` are called with `profile_name=\"spoke1\"`.
  - `test_sync_multiple_accounts_default_path_uses_default_session` — asserts that with `use_explicit_profile=False` no `profile_name` kwarg is passed (preserves the #1042 env-var-only path).
  - `test_sync_multiple_accounts_profile_session_is_usable` — moto-backed smoke test that exercises a real `boto3.Session` (no mocks on `boto3.Session`). It writes a shared credentials file with a distinctive `[spoke1]` profile, captures the session built by `_sync_multiple_accounts`, and verifies (a) `session.profile_name == \"spoke1\"`, (b) `get_credentials()` returns the profile's creds (not moto's default \`testing\` creds), and (c) an STS call succeeds through moto. This catches the class of bug that #1042/#1142/#1185 belong to — verifying the session is actually usable end-to-end, not just that `profile_name` was forwarded.
- Updated the existing `test_sync_multiple_accounts` to assert per-iteration `profile_name` kwargs and to fix a pre-existing signature mismatch (it was passing a `Config` object where `aws_best_effort_mode: bool` is expected; it only passed because `Config` was truthy).

#### General

#### Proof of functionality

### Notes for reviewers

- The single-line comments on the two branches of the session-kwargs construction document *why* each path exists — happy to expand or trim if you prefer.
- The `use_explicit_profile` parameter defaults to `False` so other callers (if any emerge) keep the safer default-session behavior.
- The `\"default\"` profile is still skipped in `get_aws_accounts_from_botocore_config`, unchanged here.
- Today `use_explicit_profile` mirrors `aws_sync_all_profiles` 1:1; it's named separately (with an inline comment at the call site) so `_sync_multiple_accounts` stays decoupled from the CLI option should the two ever diverge.

### DIFF
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -266,14 +266,13 @@ def _sync_multiple_accounts(
     aws_best_effort_mode: bool,
     aws_requested_syncs: List[str] = [],
     regions: list[str] | None = None,
+    use_profile_for_session: bool = False,
 ) -> bool:
     logger.info("Syncing AWS accounts: %s", ", ".join(accounts.values()))
     organizations.sync(neo4j_session, accounts, sync_tag, common_job_parameters)
 
     failed_account_ids = []
     exception_tracebacks = []
-
-    num_accounts = len(accounts)
 
     for profile_name, account_id in accounts.items():
         logger.info(
@@ -282,13 +281,14 @@ def _sync_multiple_accounts(
             profile_name,
         )
         common_job_parameters["AWS_ID"] = account_id
-        if num_accounts == 1:
-            # Use the default boto3 session because boto3 gets confused if you give it a profile name with 1 account
-            boto3_session = boto3.Session()
-            aioboto3_session = aioboto3.Session()
-        else:
+        if use_profile_for_session:
+            # Honor explicit profiles even for single-profile setups, so hub/spoke STS assume-role configs sync the spoke.
             boto3_session = boto3.Session(profile_name=profile_name)
             aioboto3_session = aioboto3.Session(profile_name=profile_name)
+        else:
+            # Default session keeps env-var-only credentials working when ~/.aws/config is absent (see #1042).
+            boto3_session = boto3.Session()
+            aioboto3_session = aioboto3.Session()
 
         _autodiscover_accounts(
             neo4j_session,
@@ -477,6 +477,7 @@ def start_aws_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
         config.aws_best_effort_mode,
         requested_syncs,
         regions=regions,
+        use_profile_for_session=config.aws_sync_all_profiles,
     )
 
     if sync_successful:
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -266,7 +266,7 @@ def _sync_multiple_accounts(
     aws_best_effort_mode: bool,
     aws_requested_syncs: List[str] = [],
     regions: list[str] | None = None,
-    use_profile_for_session: bool = False,
+    use_explicit_profile: bool = False,
 ) -> bool:
     logger.info("Syncing AWS accounts: %s", ", ".join(accounts.values()))
     organizations.sync(neo4j_session, accounts, sync_tag, common_job_parameters)
@@ -281,14 +281,11 @@ def _sync_multiple_accounts(
             profile_name,
         )
         common_job_parameters["AWS_ID"] = account_id
-        if use_profile_for_session:
-            # Honor explicit profiles even for single-profile setups, so hub/spoke STS assume-role configs sync the spoke.
-            boto3_session = boto3.Session(profile_name=profile_name)
-            aioboto3_session = aioboto3.Session(profile_name=profile_name)
-        else:
-            # Default session keeps env-var-only credentials working when ~/.aws/config is absent (see #1042).
-            boto3_session = boto3.Session()
-            aioboto3_session = aioboto3.Session()
+        # When use_explicit_profile is set, honor configured profiles (hub/spoke STS assume-role configs, #1142/#1185).
+        # Otherwise fall back to the default session so env-var-only credentials keep working when ~/.aws/config is absent (#1042).
+        session_kwargs = {"profile_name": profile_name} if use_explicit_profile else {}
+        boto3_session = boto3.Session(**session_kwargs)
+        aioboto3_session = aioboto3.Session(**session_kwargs)
 
         _autodiscover_accounts(
             neo4j_session,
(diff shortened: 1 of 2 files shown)


cartography/intel/aws/__init__.py
```diff
@@ -266,29 +266,29 @@ def _sync_multiple_accounts(
     aws_best_effort_mode: bool,
     aws_requested_syncs: List[str] = [],
     regions: list[str] | None = None,
+    use_profile_for_session: bool = False,
```
kunaals, line 269: naming nit: `use_explicit_profile` reads a touch more intent-y than `use_profile_for_session`. not worth churn, just flagging

cartography/intel/aws/__init__.py
```diff
@@ -266,29 +266,29 @@ def _sync_multiple_accounts(
...
             account_id,
             profile_name,
         )
         common_job_parameters["AWS_ID"] = account_id
-        if num_accounts == 1:
-            # Use the default boto3 session because boto3 gets confused if you give it a profile name with 1 account
-            boto3_session = boto3.Session()
-            aioboto3_session = aioboto3.Session()
-        else:
+        if use_profile_for_session:
```
kunaals, line 284: could collapse to `kwargs = {"profile_name": profile_name} if use_profile_for_session else {}` and splat into both sessions. taste thing, fine as is

cartography/intel/aws/__init__.py
```diff
@@ -477,6 +477,7 @@ def start_aws_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
         config.aws_best_effort_mode,
         requested_syncs,
         regions=regions,
+        use_profile_for_session=config.aws_sync_all_profiles,
```
kunaals, line 480: worth a one-liner here that this flag and `aws_sync_all_profiles` are the same signal today - so future readers don't wonder why there are two names

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -118,6 +116,81 @@ def test_sync_multiple_accounts(
     assert mock_cleanup.call_count == 1
 
 
+@mock.patch.object(cartography.intel.aws.organizations, "sync", return_value=None)
```
kunaals, line 119: these verify plumbing (that `profile_name` gets forwarded) but not actual boto3 session behavior. a moto-backed smoke test would catch the class of bug #1042/#1142/#1185 belong to. not blocking, just nice to have

kunaals (approved):
approving to unblock

nits inline, all non-blocking. one follow-up worth calling out at the review level: this restores the documented behavior for single-profile --aws-sync-all-profiles setups. users who were unintentionally relying on the old default-session fallback may see different auth behavior after upgrading, so a short changelog/release note could be helpful

PR #2647, opened 2026-04-20
feat(tailscale): model grants and resolve effective access relationships

### LINKED ISSUE #1762: Model Tailscale Grants relationships (who can access what)
### Summary

The Tailscale module currently fetches data from the API, but does not parse or model Grants. This prevents Cartography from answering access questions such as "who can access what" for Tailscale-managed resources.

This feature aims to parse and model Tailscale Grants to surface effective access relationships, similar to what is already implemented for AWS IAM.

### Motivation

Understanding identity-to-resource access relationships is a key capability for security teams using Cartography. By modeling Tailscale Grants, users will be able to:

- Query effective access paths across overlay networks.
- Detect overly permissive or misconfigured Grants.
- Build least-privilege analysis, compliance checks, or dynamic policy audits.

### Proposed Solution

_Note: Tailscale ACLs are now deprecated in favor of Tailscale Grants._

- Parse Tailscale Grants and resolve them into effective identity-to-resource relationships.
- Model relationships using MatchLinks to connect `TailscaleUser` nodes to the corresponding `TailscaleNode` or `TailscaleService`.
- Use directionality and tenant scoping consistent with AWS.
- Follow Cartography’s standard sync pattern (`get → transform → load → cleanup`), and document the schema.

### Additional Context

- The Tailscale module and endpoint integration already exist, but Grants are not yet parsed.
- Related modeling patterns exist in: `cartography/intel/aws/iam/`
- Background: ["IAM: Whatever You Say I Am"](https://lyft.github.io/cartography/blog/iam-whatever-you-say-i-am)
- Reference: [Tailscale Grants documentation](https://tailscale.com/kb/1324/grants)

jychp:
### Type of change
- [ ] Documentation update

### Summary

Model Tailscale Grants and resolve **effective access relationships** in the graph. This enables graph queries to answer "who can access what" for Tailscale-managed resources, supporting least-privilege analysis and access auditing.

**New node types:**
- `TailscaleGrant` — represents each grant rule from the ACL/policy file with source/destination relationships to groups, users, and tags
- `TailscaleService` — represents Tailscale Services fetched from `GET /api/v2/tailnet/{tailnet}/services`

**Resolved access (CAN_ACCESS MatchLinks):**
- `(:TailscaleUser)-[:CAN_ACCESS]->(:TailscaleDevice)` — user to device
- `(:TailscaleGroup)-[:CAN_ACCESS]->(:TailscaleDevice)` — group to device
- `(:TailscaleDevice)-[:CAN_ACCESS]->(:TailscaleDevice)` — device to device (tag source)
- `(:TailscaleUser)-[:CAN_ACCESS]->(:TailscaleService)` — user to service
- `(:TailscaleGroup)-[:CAN_ACCESS]->(:TailscaleService)` — group to service

**Transitive group membership:**
- `(:TailscaleUser)-[:INHERITED_MEMBER_OF]->(:TailscaleGroup)` — resolved via `MEMBER_OF*1..` graph traversal (same pattern as Google Workspace)

**Grant source/destination selectors covered:**
| Selector | Type | Status |
|----------|------|--------|
| `user@example.com` | Source | ✅ |
| `group:xxx` / `autogroup:xxx` | Source | ✅ |
| `tag:xxx` | Source (device-to-device) | ✅ |
| `*` (wildcard) | Source/Destination | ✅ |
| `tag:xxx` | Destination | ✅ |
| `group:xxx` / `autogroup:xxx` | Destination | ✅ |
| `autogroup:self` | Destination | ✅ |
| `svc:xxx` | Destination | ✅ |
| IP address / CIDR | Destination | ✅ |
| `srcPosture` | Filter | ✅ (filters non-compliant sources) |

**Design decisions:**
- Grant IDs use content-hash (`grant:<sha256[:12]>`) for stability across ACL reordering
- `CAN_ACCESS.granted_by` is a list of grant IDs, aggregating multiple grants per pair
- `srcPosture` filtering uses posture_matches from `postureresolution.sync()` to exclude non-compliant sources
- `autogroup:internet` and `via` routing are intentionally not covered (no corresponding graph entity)

### Related issues or links

- Fixes https://github.com/cartography-cncf/cartography/issues/1762

### Breaking changes

None. All changes are additive — new node types, new relationships, new intel modules. Existing Tailscale ingestion is unchanged except:
- `acls.sync()` now returns 4 values instead of 2 (added `grants` and `groups`)
- `postureresolution.sync()` now returns `posture_matches` instead of `None`

### How was this tested?

- **76 total tests** (58 unit + 18 integration) all passing
- **Unit tests** (`tests/unit/cartography/intel/tailscale/test_grants.py`): 58 tests covering:
  - `ACLParser.get_grants()` parsing for all selector types
  - `resolve_access()` for all source/destination combinations
  - IP/CIDR resolution (exact, /32, wide CIDR, IPv6, invalid)
  - `srcPosture` filtering (compliant/non-compliant users, groups, devices)
  - `granted_by` aggregation and deduplication
  - Service destination resolution
  - Edge cases (empty grants, unknown users, self-loops)
- **Integration tests** (`tests/integration/cartography/intel/tailscale/test_grants.py`): 5 tests verifying end-to-end Neo4j loading of grants, relationships, inherited membership, and effective access
- Linting passes (`pre-commit run` with isort, black, flake8, mypy)

#### General

#### Proof of functionality

### Notes for reviewers

- The `TailscaleService` module is a net-new intel module (`GET /api/v2/tailnet/{tailnet}/services`) that was added to unblock `svc:xxx` grant resolution.
- `INHERITED_MEMBER_OF` follows the same MatchLink pattern as `cartography/intel/googleworkspace/groups.py`.
- The `srcPosture` filtering is conservative: if posture data is unavailable, posture-gated grants produce no access (fail-closed).
- Schema documentation updates are not included in this PR and should be done as a follow-up.


cartography/intel/tailscale/acls.py
```diff
@@ -28,22 +29,38 @@
...
+    groups, tags, postures, posture_conditions, grants = transform(raw_acl, users)
     load_groups(
         neo4j_session,
         groups,
         common_job_parameters["UPDATE_TAG"],
         org,
     )
+    # Clean stale group nodes/edges before computing transitive membership,
+    # so INHERITED_MEMBER_OF is based only on current MEMBER_OF edges.
+    cleanup_groups(neo4j_session, common_job_parameters)
```
kunaals: since we already have the full group structure in memory here, could we compute transitive membership from the transformed groups payload instead of writing the direct edges to neo4j and then reading them back with raw cypher? i think that would give us the same inherited_member_of rows without the graph roundtrip.

cartography/intel/tailscale/acls.py
```diff
@@ -195,12 +215,77 @@ def load_posture_conditions(
 
 
 @timeit
-def cleanup(
-    neo4j_session: neo4j.Session, common_job_parameters: Dict[str, Any]
+def get_inherited_member_relationships(
```
kunaals: i think this helper goes away if we do the transitive expansion in python. we can still materialize inherited_member_of, but from python-generated rows rather than a recursive query embedded in the module.

cartography/intel/tailscale/grants.py
```diff
@@ -0,0 +1,644 @@
...
+) -> dict[str, list[str]]:
+    """Build a mapping from tag ID to list of device IDs."""
+    tag_to_devices: dict[str, list[str]] = {}
+    for device in devices:
+        for tag in device.get("tags", []):
+            tag_to_devices.setdefault(tag, []).append(device["nodeId"])
+    return tag_to_devices
+
+
+def _build_group_members_map(
```
kunaals, line 452: if we switch to python-derived transitive membership, i think resolve_access should consume that effective membership too. right now this map is direct members only, so nested groups still will not fully expand for can_access even if we materialize inherited_member_of separately.

cartography/intel/tailscale/services.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    transformed: list[dict[str, Any]] = []
+    for service in raw_services:
+        name = service["name"]
+        addrs = service.get("addrs", [])
+        tags = service.get("tags", [])
+
+        transformed.append(
+            {
+                "id": f"svc:{name}",
```
kunaals: name is already returned with the `svc:` prefix per the live api spec (`VIPServiceInfo.name` uses `svc:example`). prepending `svc:` here produces `svc:svc:...`, which won’t match `svc:xxx` grant selectors. the fixtures hide this because they use bare dns-labels.

cartography/intel/tailscale/grants.py
```diff
@@ -0,0 +1,650 @@
...
+            if user_login not in all_user_logins:
+                continue
+            if not _user_meets_posture(
+                user_login,
+                required_postures,
+                user_to_devices,
+                device_postures,
+            ):
+                continue
+            for device_id in dest_device_ids:
```
kunaals, line 338: i think we’re collapsing app-level and network-level grants together a bit too much here. this resolver will still emit `can_access` edges for app-only grants even when the grant doesn’t actually contribute network connectivity.

cartography/intel/tailscale/utils.py
```diff
@@ -130,9 +126,102 @@ def get_tags(self) -> List[Dict[str, Any]]:
...
+            destinations = grant.get("dst", [])
+
+            # Classify sources
+            source_users: list[str] = []
+            source_groups: list[str] = []
+            source_tags: list[str] = []
+            for src in sources:
+                if src.startswith("group:") or src.startswith("autogroup:"):
+                    source_groups.append(src)
+                elif src.startswith("tag:"):
```
kunaals, line 158: i don’t think rewriting `src: ["*"]` to `autogroup:member` is equivalent to tailscale semantics. the docs describe `*` more broadly than just member users, so this looks like it would under-model some valid grant sources.

cartography/intel/tailscale/grants.py
```diff
@@ -0,0 +1,650 @@
...
+            # Handled by _resolve_destination_services()
+            pass
+        elif dst.startswith("tag:"):
+            # Tag selector: find devices with this tag
+            # Handle "tag:web:443" format by stripping port suffix
+            parts = dst.split(":")
+            tag_id = parts[0] + ":" + parts[1]
+            devices_with_tag = tag_to_devices.get(tag_id, [])
+            dest_device_ids.update(devices_with_tag)
+        elif dst.startswith("group:") or dst.startswith("autogroup:"):
```
kunaals, line 600: i think this is still narrower than tailscale’s documented selector surface. things like host aliases, `ipset:`, and some other valid destination selectors seem to fall through and disappear from the derived access graph here unless we explicitly scope them out or fail loudly.

kunaals (approved):
approving to unblock but few more things to iron out before merge

PR #2653, opened 2026-04-20
feat: scope Container to individual containers; route Cloud Run + image-based Functions through Function

jychp:
### Type of change

- [x] Documentation update

### Summary

Audit of every node carrying the `:Container` semantic label revealed two classes of bugs:

1. **`:Container` was attached to group-level nodes.** `AzureContainerInstance` (actually Azure's *container group*), `GCPCloudRunRevision`, and `GCPCloudRunJob` were labeled `:Container` despite being workload groupings rather than individual containers. `AzureGroupContainer` — the node that *does* model an individual container — had `:Container` but no ontology mapping at all, so `_ont_image`/`_ont_image_digest` were never populated anywhere on the Azure side.
2. **`_ont_image`/`_ont_image_digest` were missing from `:Function`.** Container-deployed Functions (AWS Lambda `PackageType=Image`, Azure Function Apps with `DOCKER|...`, Cloud Run Jobs) exposed no image reference at the ontology layer, so cross-cloud image queries silently skipped them.

This PR fixes both, in two commits:

**Azure (5577794a)** — Move the Container ontology mapping from `AzureContainerInstance` (the group) to `AzureGroupContainer` (the individual container); remove `:Container` from `AzureContainerInstance`. Consistent with `ECSTask`/`KubernetesPod` which don't carry `:Container` either.

**Cloud Run + Functions (b5b4e686)**:
- `GCPCloudRunRevision` / `GCPCloudRunJob` lose `:Container`. Job keeps `:Function` (its true nature); Revision becomes a plain internal versioning artifact (no ontology label).
- `resolved_image_analysis.json` gains Function-scoped statements, including a traversal `(Service:Function)-[:HAS_REVISION]->(Revision)-[:HAS_IMAGE]->(Image)` so Services still get RESOLVED_IMAGE even though Revisions no longer carry `:Container`.
- `AWSLambda` (`PackageType=Image`) and `AzureFunctionApp` (`linuxFxVersion=DOCKER|...`) now populate `image_uri` / `image_digest` / `architecture_normalized` and carry `HAS_IMAGE` edges to `ECRImage` / `GitLabContainerImage` / `GCPArtifactRegistryContainerImage` — symmetric with Cloud Run and AzureGroupContainer.
- Shared helper `cartography.intel.container_image.parse_image_uri` handles digest extraction and Azure's `DOCKER|` prefix.
- Function ontology gains `_ont_image` / `_ont_image_digest`, mapped on Lambda / FunctionApp / CloudRunJob.

Ontology-level normalization handles the per-cloud field-name differences — module nodes keep their natural property names (`image_uri` on Lambda, `container_image` on CloudRunJob, etc.) and the ontology mapping translates them to the common `_ont_image` surface.

### Related issues or links

N/A — proactive ontology audit.

### Breaking changes

The following nodes **no longer match** `MATCH (n:Container) ...`:

- `AzureContainerInstance`
- `GCPCloudRunRevision`
- `GCPCloudRunJob`

Downstream queries:

- Cross-cloud container queries should now use `:Container` for individual containers (`ECSContainer`, `KubernetesContainer`, `AzureGroupContainer`) and `:Function` for container-deployed serverless workloads (`GCPCloudRunJob`, container-image `AWSLambda`, container `AzureFunctionApp`).
- `RESOLVED_IMAGE` no longer lands on `GCPCloudRunRevision` nodes directly; it is created on the parent `GCPCloudRunService` by traversing `HAS_REVISION`.
- `_ont_*` properties that previously lived on `AzureContainerInstance` migrate to `AzureGroupContainer`.

These are all documented in `docs/root/modules/ontology/schema.md` (Container and Function sections) and each module's schema doc.

### How was this tested?

- `uv run --frozen pre-commit run --all-files` → passes (black, flake8, mypy, isort, pyupgrade).
- `cubic review` → no issues.
- 41 targeted unit + integration tests pass, including:
  - New `tests/unit/cartography/intel/test_container_image.py` (7 tests for `parse_image_uri` covering all digest/prefix forms).
  - New integration test `test_container_image_lambda_has_image_and_resolved_image` in `tests/integration/cartography/intel/aws/test_lambda.py` — full end-to-end sync of a `PackageType=Image` Lambda, asserts HAS_IMAGE to ECRImage and RESOLVED_IMAGE via `:Function`.
  - New integration test `test_container_function_app_has_image_and_resolved_image` in `tests/integration/cartography/intel/azure/test_functions.py` — container-deployed FunctionApp sync with mocked `get_configuration`, asserts HAS_IMAGE + RESOLVED_IMAGE via `:Function`.
  - Updated Cloud Run test `test_resolved_image_analysis_creates_rel_for_cloud_run` — verifies RESOLVED_IMAGE lands on `:Function` (Service via `HAS_REVISION`, Job directly) and is **not** created on Revision.
  - New `test_container_ontology_mapping` in `tests/integration/cartography/intel/azure/test_group_containers.py` — verifies `_ont_name`, `_ont_image`, `_ont_image_digest`, `_ont_source=azure` land on `AzureGroupContainer`.

#### General

#### Proof of functionality

- [ ] Included Cartography sync logs from a real environment demonstrating successful synchronization of the new/modified entity. (Not included — this PR adds image fields and relationships to existing synced entities rather than a new entity; coverage is via integration tests using mocked API responses.)

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (AWS, Azure, GCP, Ontology).
- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md). (Not applicable — README is unchanged by these edits.)

### Notes for reviewers

- Two commits kept separate for review ergonomics: the Azure-only fix (5577794a) is a straightforward bug fix; the Cloud Run + Function image plumbing (b5b4e686) is where most of the surface area and design decisions live.
- Design question that drove this: `GCPCloudRunRevision` is treated as an internal versioning artifact of `GCPCloudRunService` rather than a standalone ontology entity, consistent with how Cloud Run positions Revisions. RESOLVED_IMAGE flows to the Service via traversal to preserve cross-cloud parity.
- `AzureFunctionApp` picks up an N+1 `get_configuration` API call — one per function app. This is the canonical way to read `linuxFxVersion` since the list endpoint does not return site config.
- `GCPCloudFunction` is intentionally left without `_ont_image` mapping: both Gen 1 and Gen 2 are code-based even when Gen 2 runs on Cloud Run internally.


cartography/models/gcp/cloudrun/revision.py
```diff
@@ -144,7 +143,6 @@ class CloudRunRevisionToArtifactRegistryContainerImageRel(CartographyRelSchema):
 class GCPCloudRunRevisionSchema(CartographyNodeSchema):
     label: str = "GCPCloudRunRevision"
     properties: GCPCloudRunRevisionProperties = GCPCloudRunRevisionProperties()
-    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Container"])
```
kunaals, line 147: i think this only changes the label assignment for newly ingested nodes. from what i can tell our load path adds labels with set but doesn't remove old ones, so existing cloud run revisions could keep :Container after upgrade unless there's some explicit cleanup/migration step for the old label state. maybe an analysis/cleanup query is enough here, but i wanted to call out the upgrade path.

cartography/intel/azure/functions.py
```diff
@@ -72,6 +151,11 @@ def transform_function_apps(function_apps_response: List[Dict]) -> List[Dict]:
                 "state": app.get("state"),
                 "default_host_name": app.get("default_host_name"),
                 "https_only": app.get("https_only"),
+                "is_container": is_container,
+                "deployment_type": "container" if is_container else "code",
```
kunaals: i think this may treat "couldn't fetch site config" as "definitely code-deployed". if get_configuration fails for a container-backed function app, we fall back to {} here, then set is_container false / deployment_type "code" and clear the image fields. that seems like it could silently misclassify a real container app on a transient azure error.

kunaals (approved):

PR #2655, opened 2026-04-21
fix(kubernetes): process templated EKS aws-auth entries

### LINKED ISSUE #1854: Process Templated Entries in aws-auth configmap in Kubernetes Module
*Issue*: Right now kubernetes module (specifically the eks functionality) only processes non-templated entries from the aws auth map. Being able to process the templated entries would give much better coverage 

*Description*:
> According to [this aws-auth template](https://github.com/kubernetes-sigs/aws-iam-authenticator?tab=readme-ov-file#full-configuration-format) there are three cases of templating for map roles:

1. {{AccountID}}
2. {{SessionName}}
3. {{SessionNameRaw}}

- For {{AccountID}} Just replace with AWS account ID that is in the rolearn
- If an AWS role maps to a username or group with {{SessionName}} or {{SessionNameRaw}}, we should map that AWS role to all Kube groups or kube users in rolebindings that could match the format of the username. Some regex likely required here.

[optional *Relevant Links*:]
> [Existing Kube Functionality](https://github.com/cartography-cncf/cartography/pull/1825)

jychp:
### Type of change
- [ ] Documentation update

### Summary
This fixes EKS `aws-auth` processing so supported templated `mapRoles` entries are no longer dropped during parsing.
It now resolves `{{AccountID}}` directly from the role ARN and maps `{{SessionName}}` and `{{SessionNameRaw}}` entries to existing Kubernetes users and groups that match the generated subject patterns.

### Related issues or links
- Fixes #1854

### Breaking changes
None.

### How was this tested?
- `uv run pytest tests/integration/cartography/intel/kubernetes/test_eks.py`
- `uv run pytest tests/integration/cartography/intel/kubernetes/test_rbac.py tests/integration/cartography/intel/kubernetes/test_eks.py`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
`{{EC2PrivateDNSName}}` remains unsupported here. This change is scoped to the templates called out in the issue: `{{AccountID}}`, `{{SessionName}}`, and `{{SessionNameRaw}}`.


cartography/intel/kubernetes/eks.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import re
```
kunaals, line 2: i think this is keying the lookup on the wrong cluster identifier. `sync_eks()` gets the eks external ref/arn here, but the rbac users/groups loaded earlier use `client.name` for `cluster_name`, so the session-template match will come back empty on the real eks path even though this direct test passes.

kunaals (changes requested):

kunaals (approved):

PR #2656, opened 2026-04-21
fix(gcp): skip default Apps Script project folders

kunaals:
### Type of change

### Summary
Exclude only documented default Apps Script folder parents from GCP CRM project enumeration.

This keeps Cartography on the Resource Manager list APIs while reducing work spent enumerating auto-created Apps Script default projects in large organizations. Standard user-managed projects remain included because the filter only applies when the discovered folder lineage matches `system-gsuite > apps-script`.

### Related issues or links
- N/A

### Breaking changes
None.

### How was this tested?
- Added integration tests

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
This change intentionally does not add a `sys-` project ID fallback. If Cartography cannot see the documented folder lineage, it leaves project enumeration unchanged rather than risking exclusion of standard projects.


cartography/intel/gcp/crm/folders.py
```diff
@@ -13,6 +13,47 @@
...
+            parent_folder = folder_by_name.get(parent_name)
+            if not parent_folder:
+                break
+            if (
+                parent_folder.get("displayName")
+                == _DEFAULT_APPS_SCRIPT_SYSTEM_FOLDER_DISPLAY_NAME
+            ):
+                matching_folders.add(folder["name"])
+                break
+            parent_name = parent_folder.get("parent")
```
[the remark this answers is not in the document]
kunaals: good catch. tightened this to require the exact direct `system-gsuite > apps-script` parent relationship and added a regression test for the nested case.

cartography/intel/gcp/crm/folders.py
```diff
@@ -13,6 +13,40 @@
...
+        parent_name = folder.get("parent")
+        parent_folder = folder_by_name.get(parent_name) if parent_name else None
+        if (
+            parent_name
+            and parent_name.startswith("folders/")
+            and parent_folder
+            and parent_folder.get("displayName")
+            == _DEFAULT_APPS_SCRIPT_SYSTEM_FOLDER_DISPLAY_NAME
+        ):
+            matching_folders.add(folder["name"])
```
jychp, line 46: nit: exclusion relies entirely on the display names `system-gsuite > apps-script`. Display names are user-settable, so if a customer legitimately creates folders with these names, every project directly under that folder will be silently dropped from the sync. Would checking the folder resource ID pattern (default Apps Script folders have predictable `folders/sys-*` IDs) give a stronger signal than display names alone?

jychp (approved):
LGTM, only a small non blocking comment

cartography/intel/gcp/crm/folders.py
kunaals: yeah i looked around for this and unfortunately i didn't find anything :(

PR #2657, opened 2026-04-21
refactor: refine Container/Function ontology for Azure container groups and Cloud Run

jychp:
### Type of change

- [x] Documentation update

### Summary

Follow-up to #2653 fixing two classes of mismatch between cartography's ontology and the underlying provider semantics — both surfaced while reviewing the result of that PR.

**1. Azure container group/instance labels were inverted.**
`AzureContainerInstance` actually models the *container group* (the Azure ACI resource that holds N containers), and `AzureGroupContainer` actually models each individual container inside it. The Azure SDK calls the group `container_groups` (`ContainerInstanceManagementClient.container_groups.list()`) which is what made the cartography names land backwards historically. Swap them so the graph labels match the semantics: `AzureContainerInstance` = individual container (carries `:Container`), `AzureGroupContainer` = group.

**2. Cloud Run Service / Job were modeled as `:Function`, but they are orchestrators.**
- `GCPCloudRunJob` is a managed batch orchestrator (closer to AWS Batch or an ECS Fargate task than to Lambda). It is now a pure grouping node. Each container declared in `job.template.template.containers` becomes a `GCPCloudRunContainer` child carrying `:Container` and `HAS_IMAGE`.
- `GCPCloudRunService` is a long-running HTTP orchestrator with multi-container deployments and traffic-splitting versioning (closer to `ECSService` than to Lambda). It loses `:Function`. Each container declared in `service.template.containers` (the `latestReadyRevision` spec, which the v2 API returns inline) becomes a `GCPCloudRunContainer` child.
- `GCPCloudRunRevision` was carrying `HAS_IMAGE` and image fields despite being just a versioned config snapshot of the Service. Strip it to pure metadata; per-container image data lives on the Service's container children.

This matches the pattern established by ECS in cartography — only individual runtime containers carry `:Container` / `HAS_IMAGE`; orchestrators (`ECSService`, `ECSTask`, `ECSTaskDefinition`) do not. Cloud Run does not expose runtime containers, so the ``latestReadyRevision`` spec is treated as the canonical "what runs" for cross-cloud parity.

Three commits kept separate for review ergonomics:
1. `refactor(azure): swap AzureContainerInstance and AzureGroupContainer labels`
2. `refactor(gcp-cloudrun): split Job into grouping node with GCPCloudRunContainer children`
3. `refactor(gcp-cloudrun): drop :Function from Service, model containers as Service children, strip Revision`

### Related issues or links

Follow-up to #2653 — no GitHub issue.

### Breaking changes

This PR is intentionally breaking; release notes need to call this out clearly.

**Compatibility migrations are bundled.** Two analysis jobs run automatically on the next sync to relabel existing graphs:
- ``azure_container_label_swap_migration.json`` — swaps the Azure labels via a temporary intermediate label to avoid collisions, then strips any stale ``:Container`` from groups.
- ``gcp_cloudrun_label_migration.json`` — strips ``:Function`` from existing ``GCPCloudRunService`` and ``GCPCloudRunJob`` nodes; strips stale ``:Container`` from ``GCPCloudRunRevision``.

**Cypher changes that downstream queries must adopt:**

*Azure*
- ``MATCH (n:Container) ...`` now matches ``AzureContainerInstance`` instead of ``AzureGroupContainer``.
- ``(g:AzureContainerInstance)-[:CONTAINS]->(c:AzureGroupContainer)`` becomes ``(g:AzureGroupContainer)-[:CONTAINS]->(c:AzureContainerInstance)``.
- ``ip_address`` / ``ip_address_type`` / subnet ``ATTACHED_TO`` / tags now live on ``AzureGroupContainer`` (the group) — previously on ``AzureContainerInstance``.
- ``image`` / ``image_digest`` / cpu/memory request+limit live on ``AzureContainerInstance`` (the individual container).

*GCP Cloud Run Job*
- ``GCPCloudRunJob`` no longer carries ``:Function``. Cross-cloud queries that filter on ``:Function`` must additionally include ``:Container`` (or query ``GCPCloudRunContainer`` directly) to include Job containers.
- ``(j:GCPCloudRunJob)-[:HAS_IMAGE]->(i)`` and ``(j:GCPCloudRunJob)-[:RESOLVED_IMAGE]->(i)`` are gone. New traversal: ``(j:GCPCloudRunJob)-[:CONTAINS]->(c:GCPCloudRunContainer)-[:HAS_IMAGE|RESOLVED_IMAGE]->(i)``.
- The flat array properties on Job (``container_image``, ``container_images``, ``image_digest``, ``image_digests``, ``architecture*``) are removed; these fields live on each ``GCPCloudRunContainer`` child instead.

*GCP Cloud Run Service*
- ``GCPCloudRunService`` no longer carries ``:Function``. Cross-cloud ``:Function`` queries no longer include Cloud Run Services.
- ``(s:GCPCloudRunService)-[:RESOLVED_IMAGE]->(i)`` (created previously by traversing ``HAS_REVISION``) is gone. New traversal: ``(s:GCPCloudRunService)-[:CONTAINS]->(c:GCPCloudRunContainer)-[:RESOLVED_IMAGE]->(i)``. The Service-via-``HAS_REVISION`` statements in ``resolved_image_analysis.json`` are removed.

*GCP Cloud Run Revision*
- ``GCPCloudRunRevision`` no longer carries ``HAS_IMAGE`` to any image registry node, and the flat image / architecture properties (``container_image``, ``container_images``, ``image_digest``, ``image_digests``, ``architecture*``) are removed. Revision is now a pure versioning marker.

**``GCPCloudRunContainer`` parent semantics:** the new ``GCPCloudRunContainer`` schema is shared between Job-children and Service-children. Each node has exactly one parent; ``job_id`` and ``service_id`` are mutually exclusive (the rel matcher whose target id is ``None`` does not resolve, so only one ``CONTAINS`` rel is created per node).

### How was this tested?

- ``uv run --frozen pre-commit run --files <touched files>`` → passes (black, isort, flake8, mypy, pyupgrade).
- 17 targeted integration tests pass:
  - ``tests/integration/cartography/intel/azure/test_container_instances.py`` (2)
  - ``tests/integration/cartography/intel/azure/test_group_containers.py`` (2)
  - ``tests/integration/cartography/intel/azure/test_compute_exposure.py`` (2)
  - ``tests/integration/cartography/intel/gcp/test_cloudrun.py`` (2) — image prerequisites test now asserts ``HAS_IMAGE`` on ``GCPCloudRunContainer`` parented to Service AND Job
  - ``tests/integration/cartography/intel/ontology/test_resolved_images.py`` (3) — Cloud Run RESOLVED_IMAGE now lands on ``:Container`` for both Service and Job containers; no Function path for any GCP node
- Test mocks updated to include ``service.template.containers[]`` (matches the actual v2 API response shape — see ``RevisionTemplate`` in ``projects.locations.services``).

#### General

#### Proof of functionality

- [ ] Included Cartography sync logs from a real environment.
  (Not included — this PR reshapes existing synced entities; coverage is via integration tests using mocked API responses, including the ``service.template.containers`` shape that exists but was previously unread.)

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (Azure, GCP, Ontology).
  (Not applicable — README is unchanged by these edits.)

### Notes for reviewers

- The Azure "swap" looks aggressive in the diff but it is a near-mechanical rename: file contents move between the two model files, class names follow, graph labels follow. The two test files swap their assertions accordingly. The intent is to make ``AzureContainerInstance`` finally mean what its name suggests in a Cypher query.
- The Cloud Run changes are driven by the realization that "Service / Job / Revision are orchestrators, not workloads" — the actual containers are what should carry ``:Container``. ``GCPCloudRunContainer`` is shared between the two parents (Job and Service) to avoid duplicating five HAS_IMAGE rel classes per parent type.
- The PR is intentionally breaking but bundled: the two migration jobs are auto-run by ``cartography.intel.azure.__init__`` and ``cartography.intel.gcp.__init__`` so users get a clean graph after the next sync. Old Cypher queries written against the previous label set will need updating per the breaking-changes section above.


cartography/intel/gcp/cloudrun/job.py
```diff
@@ -189,4 +225,5 @@ def sync_jobs(
 
     cleanup_job_params = common_job_parameters.copy()
     cleanup_job_params["project_id"] = project_id
+    cleanup_containers(neo4j_session, cleanup_job_params)
```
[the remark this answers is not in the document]
jychp: Acknowledging — this is a pre-existing pattern (this PR only inherits it for the new `GCPCloudRunContainer` cleanup). Tracked separately in #2658 so this PR stays scoped to the ontology refactor.

cartography/intel/gcp/cloudrun/service.py
```diff
@@ -163,4 +218,5 @@ def sync_services(
 
         cleanup_job_params = common_job_parameters.copy()
         cleanup_job_params["project_id"] = project_id
+        cleanup_containers(neo4j_session, cleanup_job_params)
```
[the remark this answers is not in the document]
jychp: Fixed in 86bd0be9 — split the schema into `GCPCloudRunJobContainer` and `GCPCloudRunServiceContainer` so each module owns its own cleanup, matching the convention followed by the rest of `cartography/intel/gcp/`. The shared `GCPCloudRunContainer` is gone; no cross-module cleanup interaction remains.

cartography/intel/gcp/cloudrun/service.py
```diff
@@ -1,16 +1,23 @@
...
 
 import neo4j
 from googleapiclient.discovery import Resource
 from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.container_arch import ARCH_SOURCE_PLATFORM_REQUIREMENT
+from cartography.intel.container_arch import normalize_architecture
+from cartography.intel.container_image import parse_image_uri
```
kunaals, line 13: i think this collapses cloud run service containers down to a single desired template in a way that can lose still-serving images. `service.template` is the revision template, but a service can still be serving traffic from older revisions during reconciliation or traffic splitting. since this pr also strips image data from `GCPCloudRunRevision`, those still-serving revisions would stop contributing `HAS_IMAGE` / `RESOLVED_IMAGE`. can we make sure we are modeling the serving image set here rather than just the latest desired template?

cartography/intel/azure/container_instances.py
```diff
@@ -38,42 +33,38 @@ def get_container_instances(
...
-            if subnet_id:
-                subnet_ids.append(subnet_id)
-
-        transformed_instance = {
-            "id": group.get("id"),
-            "name": group.get("name"),
-            "location": group.get("location"),
-            "type": group.get("type"),
-            "provisioning_state": group.get(
-                "provisioning_state",
```
kunaals, line 58: i think we’re dropping per-container runtime state here. azure aci does expose container-level state via `instanceView.currentState.state`, separate from the group provisioning state, and that seems like the right source for `_ont_state` now that `AzureContainerInstance` is the actual `:Container` node. as written, these nodes will never get `_ont_state`.

cartography/models/ontology/mapping/data/containers.py
```diff
@@ -54,26 +54,55 @@
...
-            node_label="AzureGroupContainer",
+            node_label="AzureContainerInstance",
             fields=[
                 OntologyFieldMapping(ontology_field="name", node_field="name"),
                 OntologyFieldMapping(ontology_field="image", node_field="image"),
                 OntologyFieldMapping(
                     ontology_field="image_digest", node_field="image_digest"
                 ),
-                # state: Not per-container on Azure; provisioning_state lives on the parent AzureContainerInstance (container group)
+                # state: Not per-container on Azure; provisioning_state lives on the parent AzureGroupContainer (container group)
```
kunaals: i think we should make `_ont_state` work for the new container labels introduced in this pr before merging. after this refactor, `AzureContainerInstance`, `GCPCloudRunServiceContainer`, and `GCPCloudRunJobContainer` all carry `:Container`, but none of them map `state`, so the documented container ontology contract is missing for every new runtime node added here. for azure, the api seems to give us a direct per-container state source. for cloud run, even if the right answer is a static value for service/job container specs, i think we should encode that explicitly and add integration assertions rather than leaving `_ont_state` unset across the board.

kunaals (changes requested):

cartography/intel/gcp/cloudrun/service.py
jychp: I think this is acceptable for now to reduce complexity (the case you describe will be transcient)

kunaals (approved):

PR #2659, opened 2026-04-22
fix(gcp): connect GCPPolicyBinding to bound resource via APPLIES_TO

jychp:
### Type of change
- [ ] Documentation update

### Summary

`GCPPolicyBinding.resource` stores Cloud Asset full resource names (e.g. `//storage.googleapis.com/buckets/my-bucket`, `//cloudresourcemanager.googleapis.com/projects/my-project`) that do not match the `id` convention used by the target resource nodes (`GCPBucket.id = "my-bucket"`, `GCPProject.id = "my-project"`). As a result, there was no way to traverse from a binding to the resource it governs without string-manipulating the raw field inside each Cypher query.

This PR adds a `(GCPPolicyBinding)-[:APPLIES_TO]->(ResourceNode)` edge created at ingest time via `load_matchlinks` with a dynamic `target_node_label`. The full resource name is parsed into `(label, id)` and the edge is created only when a matching node is already present in the graph. Matching on `(id, label)` rather than just `id` avoids cross-type collisions (e.g. a bucket and a project sharing the same name).

Covered target labels in this PR:
- `GCPProject`, `GCPFolder`, `GCPOrganization`
- `GCPBucket`
- `GCPKeyRing`, `GCPCryptoKey` (with cryptoKey winning over keyRing for nested paths)
- `GCPSecretManagerSecret`, `GCPSecretManagerSecretVersion`
- `GCPArtifactRegistryRepository`
- `GCPCloudRunService`
- `GCPInstance`, `GCPVpc`, `GCPSubnet`, `GCPFirewall`

Extending to more types is a one-line append to `_FULL_NAME_MAPPINGS`.

### Related issues or links

- Fixes #

### How was this tested?

- `make test_unit` and `make test_integration` locally — new parametrized unit suite (`tests/unit/cartography/intel/gcp/test_policy_bindings.py`) covers each supported label plus edge cases (unknown service, empty suffix, marker absent, nested sub-resource resolving to the nearest ancestor present in the ontology).
- Extended existing integration test `tests/integration/cartography/intel/gcp/test_policy_bindings.py` to create a test `GCPBucket` node and assert the `(GCPPolicyBinding)-[:APPLIES_TO]->(GCPProject|GCPBucket)` edges are produced.
- `make lint` clean.

#### General

#### Proof of functionality

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (`docs/root/modules/gcp/schema.md` — GCPPolicyBinding section).

### Notes for reviewers

- No migration is needed: the edge is created purely from existing ingest data, and the cleanup job runs `GraphJob.from_matchlink` for every supported label so stale `APPLIES_TO` edges are pruned even when a given sync produces no bindings for that label.
- Nested resource types resolve to the nearest ancestor present in the ontology. For example, a policy attached to a `cryptoKeyVersion` produces an edge to the owning `GCPCryptoKey`; a policy attached to a secret version produces an edge to the owning `GCPSecretManagerSecretVersion` (version beats secret because versions are nested under secrets).
- `GCPSqlInstance` and `GKECluster` are intentionally not covered: their node `id` is a `selfLink` URL that does not match the Cloud Asset full-resource-name format without a separate lookup property. BigQuery datasets/tables use a `{project}:{dataset}` id that would also need explicit synthesis. These can be added in a follow-up.


cartography/intel/gcp/policy_bindings.py
```diff
@@ -19,13 +20,173 @@
...
+        "buckets",
+        "GCPBucket",
+        "last_segment",
+    ),
+    # KMS — cryptoKey wins over keyRing (nested).
+    _FullNameMapping(
+        "//cloudkms.googleapis.com/",
+        "cryptoKeys",
+        "GCPCryptoKey",
+        "full_path",
```
kunaals, line 90: i think this needs a sync-order fix before we can rely on these new target labels. `load_matchlinks()` only connects to target nodes that already exist, but policy bindings run before secret manager, artifact registry, and cloud run in the main gcp project sync. for those resources, a newly discovered target won't get `APPLIES_TO` until a later sync, even though the docs/tests imply it works on the same run. could we either create these matchlinks after the supported resource modules run, add a post-resource pass, or limit this mapping to resource types that already exist before policy bindings?

kunaals (approved):
approving with one nit

PR #2664, opened 2026-04-22
fix(microsoft): migrate intune detected apps to report exports

kunaals:
### Type of change
- [ ] Documentation update

### Summary
Replace the Intune detected-app N+1 crawl with Intune report exports.

This switches detected app ingestion to use `AppInvAggregate` for `IntuneDetectedApp` nodes and `AppInvRawData` for `HAS_APP` relationships. It adds a focused export-job helper with bounded polling and download/parsing, updates the node model so `IntuneDetectedApp.id` is keyed by report `ApplicationKey`, and refreshes the unit/integration fixtures to use fake but report-shaped data.

### Related issues or links
- Follow-up to #2643

### Breaking changes
`IntuneDetectedApp.id` now stores the Intune report `ApplicationKey` instead of the direct Graph `detectedApp.id` resource identifier.

On the first sync after upgrade, existing `IntuneDetectedApp` nodes and `HAS_APP` edges will be rematerialized under the new identity.

### How was this tested?
- `make test_lint`
- `./.venv/bin/pytest tests/unit/cartography/intel/microsoft/intune/test_reports.py tests/unit/cartography/intel/microsoft/intune/test_detected_apps.py tests/unit/cartography/test_doc.py tests/integration/cartography/intel/microsoft/intune/test_intune.py`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
The main semantic change to review is the identifier migration: `IntuneDetectedApp.id` now comes from Intune report `ApplicationKey`. The rest of the graph shape stays the same.


cartography/intel/microsoft/intune/reports.py
```diff
@@ -0,0 +1,172 @@
...
+                decoded = io.TextIOWrapper(
+                    report_file, encoding="utf-8-sig", newline=""
+                )
+                reader = csv.DictReader(decoded)
+                if not reader.fieldnames:
+                    raise ValueError(
+                        f"Export for {report_name} did not contain CSV headers."
+                    )
+                rows = [
+                    {key: value or "" for key, value in row.items() if key is not None}
```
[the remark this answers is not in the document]
kunaals: fixed this in the latest patch.

cartography/intel/microsoft/intune/detected_apps.py
```diff
@@ -19,105 +18,62 @@
...
-        next_page_url = page.odata_next_link
-        page.value = None
-        page = await client.device_management.detected_apps.with_url(
-            next_page_url,
-        ).get()
-
 
-@timeit
-async def get_managed_device_ids_for_detected_app(
+async def get_detected_app_raw_rows(
```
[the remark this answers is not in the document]
kunaals: good catch. fixed this.

cartography/intel/microsoft/intune/detected_apps.py
```diff
@@ -19,105 +18,62 @@
...
+]
+APPINVRAWDATA_COLUMNS = [
+    "ApplicationKey",
+    "DeviceId",
 ]
 
 
-@timeit
-async def get_detected_apps(
+async def get_detected_app_aggregate_rows(
```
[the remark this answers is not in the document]
kunaals: good catch. fixed this.

cartography/intel/microsoft/intune/detected_apps.py
```diff
@@ -19,105 +18,66 @@
...
-        "version": app.version,
-        "size_in_byte": app.size_in_byte,
-        "device_count": app.device_count,
-        "publisher": app.publisher,
-        "platform": app.platform.value if app.platform else None,
+        "id": _get_required_value(row, "ApplicationKey", APPINVAGGREGATE_REPORT_NAME),
+        "application_id": _get_optional_value(row, "ApplicationId"),
+        "display_name": _get_optional_value(row, "ApplicationName"),
+        "version": _get_optional_value(row, "ApplicationVersion"),
+        "size_in_byte": None,
```
jychp: I think we can remove this property from the schema if we harcode the value to None.

cartography/intel/microsoft/intune/reports.py
```diff
@@ -0,0 +1,206 @@
...
+def download_export_report_rows(
+    download_url: str,
+    report_name: str,
+) -> ExportedReportRows:
+    with requests.get(
+        download_url,
+        stream=True,
+        timeout=EXPORT_DOWNLOAD_TIMEOUT_SECONDS,
+    ) as response:
+        response.raise_for_status()
```
jychp: May be we should use a Retry logic here.

cartography/intel/microsoft/intune/reports.py
```diff
@@ -0,0 +1,206 @@
...
+DEFAULT_REPORT_TIMEOUT_SECONDS = 300
+EXPORT_DOWNLOAD_TIMEOUT_SECONDS = 60
+MAX_REPORT_ARCHIVE_BYTES = 64 * 1024 * 1024
+MAX_REPORT_UNCOMPRESSED_BYTES = 256 * 1024 * 1024
+
+
+@dataclass(frozen=True)
+class ExportedReportRows:
+    fieldnames: tuple[str, ...]
+    rows: list[dict[str, str | None]]
```
jychp, line 44: Everything will be store in memory here, not sure if there is a huge impact or not.

cartography/intel/microsoft/intune/reports.py
```diff
@@ -0,0 +1,206 @@
...
+    return ExportedReportRows(fieldnames=tuple(reader.fieldnames), rows=rows)
+
+
+def _report_status_name(status: DeviceManagementReportStatus | None) -> str:
+    if status is None:
+        return "unknown"
+    value = status.value
+    if isinstance(value, tuple):
+        return str(value[0])
+    return str(value)
```
jychp: **nit** — `DeviceManagementReportStatus` is a standard `str` Enum, so `status.value` is always a `str`. The `isinstance(value, tuple)` branch is dead; `return str(status.value)` is enough.

jychp (commented):
non blocking comments

Additionaly:
`docs/root/modules/entra/schema.md:342-357` has a second `IntuneDetectedApp` table that wasn't touched here.

It still describes `id` as "Unique identifier for the detected app" and is missing the new `application_id` field. Worth a quick cleanup (or dropping the  duplicate) in a follow-up so the two schema docs don't drift.

cartography/intel/microsoft/intune/reports.py
kunaals: keeping this materialized for now because the sync needs repeated access to both exports to build the unioned app set. the explicit compressed and uncompressed size caps bound the memory cost.

cartography/intel/microsoft/intune/detected_apps.py
kunaals: good catch. removed the always-null field from the schema and docs.

cartography/intel/microsoft/intune/reports.py
kunaals: good catch. added retryable download handling for the export artifact.

cartography/intel/microsoft/intune/reports.py
kunaals: good catch. simplified this.

jychp (approved):
LGTM

PR #2666, opened 2026-04-23
fix(aws): preserve S3 buckets on head_bucket timeouts

kunaals:
### Type of change
- [ ] Documentation update

### Summary
Preserve S3 buckets when `head_bucket()` region discovery hits a transport timeout or connection error.

This keeps the sync on the per-bucket failure path used elsewhere in the module instead of aborting the entire account sync because one bucket endpoint is slow or unreachable.

### Related issues or links
- N/A

### Breaking changes
None.

### How was this tested?
- `uv run pytest tests/unit/cartography/intel/aws/test_s3.py -q`
- `uv run pytest tests/integration/cartography/intel/aws/test_s3.py -q`
- Commit hooks passed during `git commit`, including `isort`, `black`, `flake8`, and `mypy`.

#### General

#### Proof of functionality

### Notes for reviewers
- I kept this out of `aws_handle_regions()` because this code path is global and bucket-by-bucket rather than region-scoped.
- Returning `[]` for the whole call on one timeout would make cleanup semantics worse than preserving the bucket with `Region=None` and continuing.

### DIFF
--- a/tests/unit/cartography/intel/aws/test_s3.py
+++ b/tests/unit/cartography/intel/aws/test_s3.py
@@ -87,18 +87,36 @@ def test_get_s3_bucket_list_common_exception_sets_region_none(mock_is_common):
     assert result["Buckets"][0]["Region"] is None
 
 
-def test_get_s3_bucket_list_connect_timeout_sets_region_none():
-    """Transport timeouts during region discovery should preserve the bucket."""
+def test_get_s3_bucket_list_connect_timeout_preserves_other_buckets():
+    """A timeout on one bucket should not stop region discovery for surrounding buckets."""
     mock_session = MagicMock()
     mock_client = mock_session.client.return_value
 
     mock_client.list_buckets.return_value = {
-        "Buckets": [{"Name": "slow-bucket"}],
+        "Buckets": [
+            {"Name": "first-bucket"},
+            {"Name": "slow-bucket"},
+            {"Name": "last-bucket"},
+        ],
     }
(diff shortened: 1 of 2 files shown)


tests/unit/cartography/intel/aws/test_s3.py
```diff
@@ -84,3 +85,20 @@ def test_get_s3_bucket_list_common_exception_sets_region_none(mock_is_common):
...
+    mock_client.list_buckets.return_value = {
+        "Buckets": [{"Name": "slow-bucket"}],
+    }
+    mock_client.head_bucket.side_effect = ConnectTimeoutError(
+        endpoint_url="https://slow-bucket.s3.me-south-1.amazonaws.com/",
+        error="timed out",
+    )
+
+    result = get_s3_bucket_list(mock_session)
+    assert result["Buckets"][0]["Region"] is None
```
jychp: This test doesn't actually prove the stated goal of the PR.
You should add a multi-bucket case where a middle bucket times out and the surrounding buckets still resolve their regions.

jychp (approved):
look good but I think we can improve tests

tests/unit/cartography/intel/aws/test_s3.py
kunaals: good catch. added a multi-bucket case so the timeout is in the middle and the surrounding buckets still resolve.

kunaals: yes good call

PR #2667, opened 2026-04-23
feat(ontology): map missing nodes and add FileStorage semantic label

jychp:
### Type of change
- [x] Documentation update

### Summary

Wire seven module nodes into the existing ontology catalog and introduce a new `FileStorage` semantic label for network file systems and shares.

**New semantic label mappings:**
- `SSMParameter` &rarr; `Secret` — applied conditionally via `ConditionalNodeLabel` only when `type == "SecureString"`. `String` and `StringList` parameters are plaintext config and stay unlabeled.
- `EntraServicePrincipal` &rarr; `ServiceAccount`.
- `ScalewayPermissionSet` &rarr; `PermissionRole`.
- `WorkdayHuman` &rarr; `UserAccount` (alongside existing `Human`).
- `EfsFileSystem`, `AzureDataLakeFileSystem`, `AzureStorageFileShare` &rarr; `FileStorage` (new).

**New `FileStorage` semantic label:**
- Added `cartography/models/ontology/mapping/data/file_storage.py` with `_ont_name` (required), `_ont_location`, `_ont_encrypted`.
- Registered under the `filestorage` category in `SEMANTIC_LABELS_MAPPING`.
- Documented in `docs/root/modules/ontology/schema.md` (mermaid + dedicated section) and `docs/root/agents/enrich-ontology.md`.

`FileStorage` is deliberately distinct from `ObjectStorage` (S3 / GCS / Azure Blob) and from block storage (EBS, managed disks) so cross-provider queries for NFS / SMB file systems stay precise.

### Breaking changes

None. Existing nodes gain additional labels and `_ont_*` properties; no label is removed or renamed.

### How was this tested?

- `uv run --frozen pytest tests/unit/cartography/intel/ontology/test_ontology_mapping.py tests/unit/cartography/graph/test_ontology_fields.py` &rarr; 34 passed.
- `uv run --frozen pytest tests/unit/cartography/graph/` &rarr; 75 passed.
- `uv run --frozen pytest tests/unit/cartography/intel/workday/ tests/unit/cartography/intel/microsoft/ tests/unit/cartography/intel/scaleway/` &rarr; 22 passed.
- `uv run --frozen pre-commit run --files ...` &rarr; all hooks passed (black, isort, flake8, mypy, pyupgrade).

#### General

#### Proof of functionality

### Notes for reviewers

- The `SSMParameter` mapping uses `ConditionalNodeLabel` so only `SecureString` parameters carry the `Secret` label at ingest time. `_ont_*` properties are still set on the full node set (harmless, and makes the intent traceable).
- For `ScalewayPermissionSet` the `_ont_scope` field is intentionally left unmapped: `scope_type` is a Scaleway enum (`PROJECTS`, `ACCOUNTS`, ...) that would need a value-mapping table to normalize into ontology scopes.
- For the Azure `FileStorage` nodes, `_ont_location` and `_ont_encrypted` are not mapped because those live on the parent storage account, not on the filesystem / share itself.


cartography/models/workday/human.py
```diff
@@ -71,7 +71,7 @@ class WorkdayHumanToManagerRel(CartographyRelSchema):
 @dataclass(frozen=True)
 class WorkdayHumanSchema(CartographyNodeSchema):
     label: str = "WorkdayHuman"
-    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Human"])
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Human", "UserAccount"])
```
kunaals: i think this changes the meaning of `WorkdayHuman` a bit too much. in our docs/workday model this is a person record, not really a system account, and adding `UserAccount` means it can start participating in the canonical `User` ontology flow as an account source. is that semantic collapse actually something we want here?

cartography/models/azure/data_lake_filesystem.py
```diff
@@ -67,6 +68,7 @@ class AzureDataLakeFileSystemToSubscriptionRel(CartographyRelSchema):
 class AzureDataLakeFileSystemSchema(CartographyNodeSchema):
     label: str = "AzureDataLakeFileSystem"
     properties: AzureDataLakeFileSystemProperties = AzureDataLakeFileSystemProperties()
```
kunaals, line 69: i think `AzureDataLakeFileSystem` may not fit this new `FileStorage` contract. the docs here describe `FileStorage` as managed network file systems / shares, but ADLS Gen2 filesystems are still blob-storage capabilities with file-system semantics, not really the same thing as EFS or Azure Files. this feels like it would make cross-provider `FileStorage` queries too broad.

kunaals (approved):
approving to unblock, two questions

PR #2670, opened 2026-04-23
feat(aws): add ELBV2TargetGroup node and ECSService target registrations

jychp:
## Summary

- Adds a first-class `ELBV2TargetGroup` node created during LBv2 sync, with `(AWSLoadBalancerV2)-[:ELBV2_TARGET_GROUP]->(ELBV2TargetGroup)` edges
- Adds a `TARGETS` MatchLink from `ELBV2TargetGroup` to `ECSService` created during ECS service sync, carrying `container_name` and `container_port` properties
- Enables the graph path `(ECSService)<-[:TARGETS]-(ELBV2TargetGroup)<-[:ELBV2_TARGET_GROUP]-(AWSLoadBalancerV2)` which was previously missing

### Motivation

Currently there is no graph path from `ECSService` to `AWSLoadBalancerV2`. The `loadBalancers[].targetGroupArn` field from `ecs:DescribeServices` is silently dropped, and there is no `ELBV2TargetGroup` node at all. This prevents tracing ECS services to the load balancers fronting them.

This is needed by SubImage's attack-path engine (SUB-971), which currently falls back to string matching on service/LB names. Adding first-class TG nodes and edges fixes this properly.

### No breaking changes

- Existing `(LBv2)-[:EXPOSE]->(...)` edges and `aws_lb_container_exposure.json` analysis job are unchanged
- No new boto3 calls — all data comes from existing `describe_load_balancers` / `describe_target_groups` / `describe_services` responses

## Test plan

- [x] New integration test `test_load_load_balancer_v2_target_group_nodes` — verifies 4 TG nodes and ELBV2_TARGET_GROUP edges
- [x] New integration test `test_load_ecs_services_target_group_registrations` — verifies TARGETS edge with container_name/container_port
- [x] All 17 existing LBv2 + ECS integration tests pass
- [x] Lint clean
- [x] `cubic review` clean


cartography/intel/aws/ecs.py
```diff
@@ -305,6 +309,44 @@ def load_ecs_services(
         AWS_ID=current_aws_account_id,
         lastupdated=aws_update_tag,
     )
+    _load_ecs_service_target_group_registrations(
```
kunaals, line 312: i think we should add `ec2:load_balancer_v2` to the dependency warning for `ecs` now that ECS creates this matchlink. requested syncs are exact keys, not prefix-expanded, so `--aws-requested-syncs ecs` won’t imply any `ec2:*` syncs. full sync order is fine, but selective ECS syncs will otherwise skip these TARGETS edges silently when the TG nodes aren’t present.

kunaals (approved):
approving with one nit

PR #2676, opened 2026-04-23
feat(ontology): promote DNSRecord to semantic label and add cross-provider DNS linking

jychp:
### Type of change

### Summary

Promote `DNSRecord` from a bare `extra_node_label` to a proper ontology semantic label, and add a cross-provider analysis job that creates `DNS_POINTS_TO` relationships between DNS records and the infrastructure they resolve to.

**Changes:**

1. **DNSRecord ontology mapping** — Define standardized fields (`name`, `type`, `value`) for all four DNS record providers: AWS Route53, GCP Cloud DNS, Cloudflare, and Vercel.

2. **KubernetesIngress `host_names` property** — Extract hostnames from ingress rules into a first-class `list[str]` property (previously buried in a JSON blob), enabling Cypher matching.

3. **`ontology_dnsrecords_linking` analysis job** — Cross-provider `DNS_POINTS_TO` linking:
   - `DNSRecord.name` → `KubernetesIngress` (hostname match)
   - `DNSRecord.value` → `AWSLoadBalancerV2`, `AWSLoadBalancer` (LB DNS name match)
   - `DNSRecord.value` → `AWSCloudFrontDistribution` (distribution domain match)
   - `DNSRecord.value` → `EC2Instance` (public DNS name match)
   - `DNSRecord.name` → `AzureAppService`, `AzureFunctionApp` (default hostname match)

4. **Cleanup** — Remove stale `VercelDomain` entry from the `DNSZone` ontology mapping (the schema never had the `DNSZone` extra label).

This enables queries like:
```cypher
MATCH (dns:DNSRecord)-[:DNS_POINTS_TO]->(ing:KubernetesIngress)-[:TARGETS]->(svc:KubernetesService)-[:TARGETS]->(pod:KubernetesPod)<-[:CONTAINS]-(c:KubernetesContainer)
RETURN dns.name, pod.name, c.image
```

### Related issues or links

- Enables cross-provider DNS-to-compute visibility (e.g. Cloudflare → Kubernetes, Route53 → Azure)

### How was this tested?

- All existing ontology unit and integration tests pass (40 tests)
- All Kubernetes tests pass (57 tests)
- All Cloudflare tests pass (6 tests)

#### General

#### Proof of functionality

### Notes for reviewers

The `DNS_POINTS_TO` relationship label is reused from the existing `AWSDNSRecord` provider-specific relationships. The analysis job uses the ontology `DNSRecord` label so it works cross-provider — a Cloudflare CNAME pointing at an AWS ALB hostname will now be linked.

The KubernetesIngress `host_names` extraction follows the same pattern as the existing `load_balancer_dns_names` property on the same node.


cartography/data/jobs/analysis/ontology_dnsrecords_linking.json
```diff
@@ -0,0 +1,40 @@
...
+  "name": "Ontology - DNSRecords relationship linking",
+  "statements": [
+    {
+      "__comment": "Link DNSRecord to KubernetesIngress where the record name matches an ingress host",
+      "query": "MATCH (dns:DNSRecord) WHERE dns.lastupdated = $UPDATE_TAG AND dns.name IS NOT NULL WITH dns MATCH (ing:KubernetesIngress) WHERE dns.name IN ing.host_names MERGE (dns)-[r:DNS_POINTS_TO]->(ing) ON CREATE SET r.firstseen = timestamp() SET r.lastupdated = $UPDATE_TAG",
+      "iterative": false
+    },
+    {
+      "__comment": "Link DNSRecord to AWSLoadBalancerV2 where the record value matches the LB DNS name",
+      "query": "MATCH (dns:DNSRecord) WHERE dns.lastupdated = $UPDATE_TAG AND dns.value IS NOT NULL WITH dns MATCH (lb:AWSLoadBalancerV2) WHERE toLower(dns.value) = toLower(lb.dnsname) MERGE (dns)-[r:DNS_POINTS_TO]->(lb) ON CREATE SET r.firstseen = timestamp() SET r.lastupdated = $UPDATE_TAG",
```
kunaals: i think this still misses gcp. these matches all use `dns.value`, but `GCPRecordSet` stores rrdata in `data`, not `value`. also, gcp rrdata is list-valued, so even a semantic-field version would still need unwind/normalization instead of a straight scalar compare.

cartography/data/jobs/analysis/ontology_dnsrecords_linking.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "Ontology - DNSRecords relationship linking",
+  "statements": [
+    {
+      "__comment": "Link DNSRecord to KubernetesIngress where the record name matches an ingress host",
+      "query": "MATCH (dns:DNSRecord) WHERE dns.lastupdated = $UPDATE_TAG AND dns.name IS NOT NULL WITH dns MATCH (ing:KubernetesIngress) WHERE dns.name IN ing.host_names MERGE (dns)-[r:DNS_POINTS_TO]->(ing) ON CREATE SET r.firstseen = timestamp() SET r.lastupdated = $UPDATE_TAG",
```
kunaals: i think we also need cleanup wiring for these new `DNS_POINTS_TO` edges. this job creates links to `KubernetesIngress`, `AzureAppService`, `AzureFunctionApp`, and `CloudFrontDistribution`, but i don’t see the cleanup-only rel schemas for them, so stale analysis-created links can stick around after the match stops being true.

kunaals (approved):
approving to unblock, two nits

PR #2678, opened 2026-04-24
feat(scan): Add DEPLOYED rel between syft/trivy and images

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- This PR adds the following 

    - TrivyPackage ---DEPLOYS--->Image
    - SyftPackage ---DEPLOYS--->Image
   
    - Added propagation query: Package → Image via TrivyPackage DEPLOYED path                                                                                                          
  - Added propagation query: Package → Image via SyftPackage DEPLOYED path 
  - This helps create the Package ---DEPLOYED---->Image shortcut
  - Removed 4 provider-specific propagation queries (ECR, GCP Container, GCP Platform, GitLab) and the respective relationship classes. 
  
- Behavior change: TrivyPackage/TrivyImageFinding no longer link to multi-platform image nodes (manifest lists / image indexes). Edges are now created only to single-platform :Image nodes via _ont_digest. Intentional — scanners resolve to a platform image before scanning. Callers querying manifest-list digests directly will see empty results.

### How was this tested?

image
image

#### General

#### Proof of functionality

### Notes for reviewers

- The trivy/syft data was mocked in order to match images that existed in my local database in order to prove that this relationship could be created.


cartography/models/trivy/package.py
```diff
@@ -77,6 +77,17 @@ class TrivyPackageToGitLabImageRel(CartographyRelSchema):
     properties: TrivyPackageToImageRelProperties = TrivyPackageToImageRelProperties()
 
 
+@dataclass(frozen=True)
+class TrivyPackageToOntologyImageRel(CartographyRelSchema):
```
kunaals, line 37: i think we probably need the same generic :Image wiring for TrivyImageFinding too. this makes TrivyPackage registry-agnostic, but findings are still only attached to ECR/GAR/GitLab-specific image nodes, so new registries that only map into :Image still won't get finding->image edges.

cartography/models/trivy/package.py
```diff
@@ -77,6 +77,17 @@ class TrivyPackageToGitLabImageRel(CartographyRelSchema):
     properties: TrivyPackageToImageRelProperties = TrivyPackageToImageRelProperties()
 
 
+@dataclass(frozen=True)
```
kunaals, line 36: i think this only fixes the scanner-package layer. the ontology package propagation is still registry-specific today, so Package -> DEPLOYED won't follow this for new registries yet. should we generalize the ontology package layer too?

kunaals, line 36: after the finding/image wiring and ontology package propagation are generalized, i think there is some follow-up cleanup to do too: the registry-specific Package -> DEPLOYED propagation/cleanup paths can probably collapse into one generic :Image path. not blocking this part first, but it seems like good cleanup once the broader change lands.

kunaals (changes requested):
left a few blocking inline comments. i think this is close, but it still misses the generic image path in a couple places.

cartography/models/trivy/package.py
shyammukund: Deleted 4 provider-specific rel classes in models/ontology/package.py and deleted 4 provider-specific queries in models/ontology/mapping/data/packages.py 

Tests are now updated to use ontology Image instead of any provider specific images.

shyammukund: Went ahead and adde this into the ontology layers.  ontology package propagation  is not longer registry specific

cartography/models/trivy/package.py
shyammukund: added (TrivyImageFinding)-[:AFFECTS]->(Image)

cartography/models/trivy/package.py
```diff
@@ -106,6 +117,7 @@ class TrivyPackageSchema(CartographyNodeSchema):
             TrivyPackageToGCPImageRel(),
             TrivyPackageToGCPPlatformImageRel(),
             TrivyPackageToGitLabImageRel(),
+            TrivyPackageToOntologyImageRel(),
```
kunaals, line 72: if the intent is to attach scanner packages only to ontology `:Image` now, i think the provider-specific `DEPLOYED` rels should be removed from this schema. `TrivyPackageToOntologyImageRel` covers the intended generic image path via `_ont_digest`, and keeping the ECR/GCP/GitLab rels means we still materialize the old provider-specific shape too.

cartography/models/trivy/findings.py
```diff
@@ -106,5 +117,6 @@ class TrivyImageFindingSchema(CartographyNodeSchema):
             TrivyFindingToGCPImageRel(),
             TrivyFindingToGCPPlatformImageRel(),
             TrivyFindingToGitLabImageRel(),
+            TrivyFindingToOntologyImageRel(),
```
kunaals, line 72: same thing here for findings: if finding/image associations should only go through ontology `:Image`, i think the provider-specific `AFFECTS` rels can be removed now that `TrivyFindingToOntologyImageRel` exists. otherwise this still creates both the old provider-specific edges and the new generic edge.

cartography/models/ontology/package.py
```diff
@@ -64,41 +64,8 @@ class PackageToSocketDevDependencyRel(CartographyRelSchema):
...
-    )
-    direction: LinkDirection = LinkDirection.OUTWARD
-    rel_label: str = "DEPLOYED"
-    properties: PackageToNodeRelProperties = PackageToNodeRelProperties()
-
-
-@dataclass(frozen=True)
-class PackageToGitLabContainerImageRel(CartographyRelSchema):
-    target_node_label: str = "GitLabContainerImage"
+class PackageToOntologyImageRel(CartographyRelSchema):
```
kunaals, line 67: i think this relationship is cleanup-only in practice. `Package.id` values are package ids like `npm|express|4.18.2`, so this matcher won't load-time match real `Image.id` values; the actual `Package -> DEPLOYED -> Image` edge is created by `ontology_packages_linking.json` through the scanner package path. can we add a short cleanup-only comment here, or use a `_cleanup_*` placeholder PropertyRef like the other ontology cleanup-only relationships, so future readers don't treat this matcher as load-bearing?

kunaals (changes requested):

kunaals (approved):
looks good, one small line to add to the PR desc so that callers are aware of the changed behavior:
```
Behavior change: TrivyPackage/TrivyImageFinding no longer link to multi-platform image nodes (manifest lists / image indexes). Edges are now created only to single-platform :Image nodes via _ont_digest. Intentional — scanners resolve to a platform image before scanning. Callers querying manifest-list digests directly will see empty results.
```

PR #2679, opened 2026-04-24
fix(gcp): Handle CAI policy binding rate limits in GCP sync

kunaals:
### Type of change

### Summary

Handle Cloud Asset Inventory policy-binding quota exhaustion in the GCP sync path by adding targeted CAI gRPC retry/backoff for rate-limit and transient failures, adding light CAI-specific throttling for the two policy-binding RPCs, and preserving existing policy-binding and derived permission-relationship data when a project cannot be refreshed completely.

### Related issues or links

- None

### Breaking changes

None.

### How was this tested?

- `uv run pytest tests/unit/cartography/intel/gcp/test_policy_bindings.py tests/unit/cartography/intel/gcp/test_selective_sync.py tests/integration/cartography/intel/gcp/test_policy_bindings.py -q`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

The main behavior change is that incomplete Cloud Asset policy-binding refreshes no longer clean up or recompute derived permission relationships for the affected project.

### DIFF
--- a/cartography/intel/gcp/__init__.py
+++ b/cartography/intel/gcp/__init__.py
@@ -507,11 +507,8 @@ def _sync_project_resources(
                 and policy_bindings_status
                 != policy_bindings.PolicyBindingsSyncStatus.SUCCESS
             ):
-                status_label = (
-                    policy_bindings_status.value.replace("_", " ")
-                    if policy_bindings_status is not None
-                    else "not_run"
-                )
+                assert policy_bindings_status is not None
+                status_label = policy_bindings_status.value.replace("_", " ")
                 logger.warning(
                     "Skipping GCP permission relationships for project %s because policy bindings sync was %s. "
                     "Preserving existing permission relationships for this project.",
--- a/cartography/intel/gcp/__init__.py
+++ b/cartography/intel/gcp/__init__.py
@@ -172,6 +172,7 @@ def _sync_project_resources(
     for project in projects:
         project_id = project["projectId"]
         common_job_parameters["PROJECT_ID"] = project_id
+        policy_bindings_status: policy_bindings.PolicyBindingsSyncStatus | None = None
         enabled_services = _services_enabled_on_project(
             build_client("serviceusage", "v1", credentials=credentials),
             project_id,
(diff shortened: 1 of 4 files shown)


cartography/intel/gcp/policy_bindings.py
```diff
@@ -17,6 +26,91 @@
...
+        last_call = _CAI_POLICY_BINDINGS_LAST_CALL_BY_OPERATION.get(operation)
+        if last_call is not None:
+            sleep_for = max(0.0, min_interval - (now - last_call))
+            if sleep_for > 0:
+                logger.debug(
+                    "Throttling Cloud Asset policy bindings %s for %.2f seconds.",
+                    operation,
+                    sleep_for,
+                )
+                time.sleep(sleep_for)
```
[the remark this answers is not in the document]
kunaals: good catch. moved the sleep outside the lock and reserve the next per-operation slot under the lock so unrelated calls do not block each other.

cartography/intel/gcp/policy_bindings.py
```diff
@@ -17,6 +26,98 @@
...
+            operation,
+            exc,
+        ),
+    )
+
+
+def _is_rate_limit_retry_error(exc: Exception) -> bool:
+    if isinstance(exc, ResourceExhausted):
+        return True
+    return isinstance(exc, RetryError) and isinstance(exc.cause, ResourceExhausted)
```
jychp, line 116: nitpick: The first branch (`isinstance(exc, ResourceExhausted)`) is unreachable from the only call site — it's inside `except RetryError as e:` at line 341, so `exc` is always a `RetryError`. Could simplify to just the `RetryError`-cause check.

cartography/intel/gcp/__init__.py
```diff
@@ -483,16 +495,36 @@ def _sync_project_resources(
...
+            if (
+                policy_bindings_requested
+                and policy_bindings_status
+                != policy_bindings.PolicyBindingsSyncStatus.SUCCESS
+            ):
+                status_label = (
+                    policy_bindings_status.value.replace("_", " ")
+                    if policy_bindings_status is not None
+                    else "not_run"
+                )
```
jychp: nitpick: The `"not_run"` fallback is unreachable. When `policy_bindings_requested` is `True`, every code path through the `if policy_bindings_requested:` block above sets `policy_bindings_status` before reaching here.

tests/integration/cartography/intel/gcp/test_policy_bindings.py
jychp, line 1: nitpick: The docstring for `test_sync_gcp_policy_bindings_permission_denied` (line 288) still says "sync should return False" but the function now returns a `PolicyBindingsSyncStatus` enum.

jychp (approved):
LGTM, some dead code and drifting docstring

cartography/intel/gcp/policy_bindings.py
kunaals: yep. simplified this to just the retryerror cause check.

cartography/intel/gcp/__init__.py
kunaals: agreed. dropped the unreachable fallback and made the invariant explicit for mypy.

tests/integration/cartography/intel/gcp/test_policy_bindings.py
kunaals: good catch. updated the docstring to match the enum return value.

PR #2680, opened 2026-04-24
fix(aws): preserve ELB state on transient regional failures

kunaals:
### Type of change

### Summary

Preserve last-known-good load balancer state when regional ELB endpoints return ambiguous transient failures.

Classic ELB and ELBV2 already use `aws_handle_regions()`, which retries `ClientError` failures for up to 600 seconds. After those retries are exhausted, retryable regional failures now cause the affected region to be skipped and account-level cleanup to be suppressed for that sync, preventing stale cleanup from deleting previously synced load balancer data.

S3 keeps its existing `FETCH_FAILED` behavior and now applies it to connect/read timeouts from per-bucket detail fetches. This keeps the change scoped to skipping failed S3 detail groups on transient transport failures without changing S3 cleanup semantics or bucket-list return shape.

### Related issues or links

- N/A

### Breaking changes

None.

### How was this tested?

- `uv run --frozen pytest tests/unit/cartography/intel/aws/ec2/test_load_balancers.py tests/unit/cartography/intel/aws/ec2/test_load_balancer_v2s.py tests/unit/cartography/intel/aws/test_s3.py`
- `uv run --frozen pre-commit run --files cartography/intel/aws/ec2/load_balancers.py cartography/intel/aws/ec2/load_balancer_v2s.py cartography/intel/aws/s3.py tests/unit/cartography/intel/aws/ec2/test_load_balancers.py tests/unit/cartography/intel/aws/ec2/test_load_balancer_v2s.py tests/unit/cartography/intel/aws/test_s3.py`
- Live AWS validation against a real multi-account sync where transient Middle East regional ELB and S3 endpoint failures were observed. Account details omitted.

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

This intentionally stays local to classic ELB, ELBV2, and S3 detail fetches. It does not change `aws_handle_regions()` globally, does not add a shared transient-error abstraction, and does not change S3 cleanup behavior. In particular, this PR does not try to change the existing S3 derived exposure cleanup behavior; it only prevents transport timeouts from aborting per-bucket detail fetches.


cartography/intel/aws/s3.py
```diff
@@ -101,9 +100,15 @@ def get_s3_bucket_list(boto3_session: boto3.session.Session) -> List[Dict]:
                     "skipping bucket='{}' due to exception.".format(bucket["Name"]),
                 )
                 continue
-            else:
-                raise
-        except (ConnectTimeoutError, EndpointConnectionError, ReadTimeoutError):
+            if is_retryable_aws_client_error(e):
+                bucket["Region"] = None
```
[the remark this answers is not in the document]
kunaals: good catch. fixed this by skipping buckets whose region could not be rediscovered, then preserving the existing bucket, acl, and policy-statement state so cleanup does not drop the last-known-good region or related s3 state.

cartography/intel/aws/s3.py
```diff
@@ -1336,6 +1404,12 @@ def sync(
         current_aws_account_id,
         update_tag,
     )
+    preserve_s3_buckets_with_failed_region_discovery(
+        neo4j_session,
+        bucket_data.get("FailedBuckets", []),
+        current_aws_account_id,
+        update_tag,
+    )
```
jychp: blocking: `preserve_s3_buckets_with_failed_region_discovery()` runs after `load_s3_details()`, but `load_s3_details()` executes `aws_s3_details.json` cleanup internally (line 734) which REMOVEs `anonymous_access` and `anonymous_actions` from ALL `S3Bucket` nodes in the account including the ones being preserved here.

cartography/intel/aws/ec2/load_balancers.py
```diff
@@ -152,8 +159,19 @@ def get_loadbalancer_data(
...
+    except ClientError as error:
+        if is_retryable_aws_client_error(error):
+            raise ELBTransientRegionFailure(
+                "AWS SDK retries were exhausted for transient DescribeLoadBalancers failure"
+            ) from error
+        raise
+    except TRANSIENT_REGION_EXCEPTIONS as error:
+        raise ELBTransientRegionFailure(
+            "Encountered a transient regional ELB endpoint failure while calling DescribeLoadBalancers"
+        ) from error
```
jychp, line 198: Catching and re-raising ELBTransientRegionFailure will block the backoff decorattor in @aws_handle_regions.

cartography/intel/aws/ec2/load_balancer_v2s.py
```diff
@@ -431,22 +482,39 @@ def sync_load_balancer_v2s(
...
+                current_aws_account_id,
+                region,
+                error,
+            )
+            continue
         load_load_balancer_v2s(
             neo4j_session,
             data,
             region,
             current_aws_account_id,
```
jychp, line 586: `cleanup_safe` is an account-wide boolean if any single region has a transient failure, cleanup is skipped for ALL regions. A per-region approach (e.g., tracking `failed_regions: set[str]` and scoping cleanup to exclude only those regions) would avoid accumulating stale data in healthy regions.

cartography/intel/aws/s3.py
```diff
@@ -93,25 +94,94 @@ def get_s3_bucket_list(boto3_session: boto3.session.Session) -> List[Dict]:
...
+        except TRANSIENT_REGION_EXCEPTIONS:
+            failed_bucket_names.append(bucket["Name"])
             logger.warning(
                 "skipping bucket='%s' region discovery due to transport timeout/connection error.",
                 bucket["Name"],
             )
+            continue
+        resolved_buckets.append(bucket)
+    buckets["Buckets"] = resolved_buckets
+    buckets["FailedBuckets"] = failed_bucket_names
```
jychp: Previously, `head_bucket` failures still ingested the bucket with `Region=None`, so at least its existence was represented in the graph. Now the bucket is completely excluded from `Buckets`, and `preserve_s3_buckets_with_failed_region_discovery()` only updates nodes that already exist in Neo4j.

cartography/intel/aws/s3.py
```diff
@@ -93,25 +94,94 @@ def get_s3_bucket_list(boto3_session: boto3.session.Session) -> List[Dict]:
...
+    existing bucket and related ACL/policy nodes keeps cleanup from deleting
+    previously known state during transient failures.
+    """
+    if not bucket_names:
+        return
+
+    preserve_bucket_query = """
+    MATCH (:AWSAccount{id: $AWS_ID})<-[:RESOURCE]-(b:S3Bucket)
+    WHERE b.name IN $bucket_names
+    SET b.lastupdated = $UPDATE_TAG
```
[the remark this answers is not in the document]
jychp: Need to be fixed.

jychp (changes requested):

cartography/intel/aws/s3.py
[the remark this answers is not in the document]
kunaals: good catch. fixed this. the preservation path now updates the bucket plus the related resource/applies_to/policy_statement relationship lastupdated values so cleanup keeps the full last-known-good s3 state.

cartography/intel/aws/ec2/load_balancer_v2s.py
kunaals: keeping this account-wide for now. this matches the existing lambda/cloudtrail cleanup contract: once a regional read is ambiguous we skip cleanup for the whole resource family rather than mixing fresh data with partial deletes.

cartography/intel/aws/s3.py
kunaals: good catch. fixed this. the anonymous_access cleanup now excludes buckets whose region discovery failed, so we no longer strip those properties before preserving the bucket state.

cartography/intel/aws/ec2/load_balancers.py
kunaals: keeping this as-is. the custom exception is intentional so retry exhaustion does not get converted into [] by aws_handle_regions, and that matches the existing lambda/cloudtrail transient-region pattern.

cartography/intel/aws/s3.py
kunaals: good catch. fixed this too. failed region discovery now preserves existing buckets, and if the bucket was not already in neo4j we create the minimal bucket/resource state without writing an ambiguous region value.

kunaals:
@cubic-dev-ai can you do one more review?

cartography/intel/aws/s3.py
kunaals: small correction to my earlier reply: this now preserves existing buckets only. if region discovery fails for a bucket that is not already in neo4j, we skip creating a new bucket node because we do not have a trustworthy region value to write.

jychp (approved):
LGTM

PR #2686, opened 2026-04-24
feat(ontology): map OktaAdministrationRole, ScalewayPermissionSet, and WorkOSRole to roles

jychp:
## Summary
- Add ontology field mappings (`_ont_name`, `_ont_type`, `_ont_scope`) and the `PermissionRole` semantic label for **OktaAdministrationRole**, **ScalewayPermissionSet**, and **WorkOSRole**
- For schema-based nodes (Scaleway, WorkOS): add `ExtraNodeLabels(["PermissionRole"])` so the querybuilder auto-generates the label and `_ont_*` fields
- For legacy Cypher node (Okta): add `:PermissionRole` label and `_ont_*` SET clauses directly in the ingest queries

## Test plan
- [x] All ontology unit tests pass (37/37)
- [x] All Okta, Scaleway, WorkOS unit tests pass
- [x] Full lint suite passes (pre-commit, mypy, flake8, black, isort)
- [x] `cubic review` passes clean


cartography/models/ontology/mapping/data/roles.py
```diff
@@ -214,6 +214,87 @@
...
+                    ontology_field="type",
+                    node_field="",
+                    special_handling="static_value",
+                    extra={"value": "builtin"},
+                ),
+                OntologyFieldMapping(
+                    ontology_field="scope",
+                    node_field="",
+                    special_handling="static_value",
+                    extra={"value": "org"},
```
kunaals: i think this should use `scope_type` rather than a static `org`. scaleway permission sets are split between org-scoped and project-scoped sets, and the model already stores `scope_type`; hardcoding this makes project-scoped permission sets look org-scoped in `PermissionRole` queries.

cartography/models/ontology/mapping/data/roles.py
```diff
@@ -214,6 +214,87 @@
...
+            node_label="WorkOSRole",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="name", required=True
+                ),
+                OntologyFieldMapping(
+                    ontology_field="type",
+                    node_field="",
+                    special_handling="static_value",
+                    extra={"value": "builtin"},
```
kunaals: i think this still flattens two different WorkOS concepts. `listOrganizationRoles` can return both environment roles and organization-specific custom roles, but this maps all of them to `builtin` / `org`. could we derive this from `role.type`, or leave the field unset until we can map it accurately?

cartography/intel/okta/roles.py
```diff
@@ -105,9 +105,13 @@ def _load_user_role(
     MATCH (user:OktaUser{id: $USER_ID})<-[:RESOURCE]-(org:OktaOrganization)
     WITH user,org
     UNWIND $ROLES_DATA as role_data
-    MERGE (role_node:OktaAdministrationRole{id: role_data.type})
+    MERGE (role_node:OktaAdministrationRole:PermissionRole{id: role_data.type})
     ON CREATE SET role_node.type = role_data.type, role_node.firstseen = timestamp()
-    SET role_node.label = role_data.label, role_node.lastupdated = $okta_update_tag
+    SET role_node.label = role_data.label,
+        role_node.lastupdated = $okta_update_tag,
+        role_node._ont_name = role_data.label,
```
kunaals, line 112: since this path bypasses the querybuilder, can we also set `_ont_source = 'okta'` here? schema-based semantic labels get that automatically, but these handwritten Cypher loads currently only set `_ont_name`, `_ont_type`, and `_ont_scope`. an Okta integration assertion for the ontology fields would catch this too.

kunaals (changes requested):

cartography/intel/okta/roles.py
kunaals: i guess could also wait till Okta is fully migrated to data model and then this will not be relevant?

kunaals (approved):

PR #2690, opened 2026-04-24
feat(gcp): Add code-to-cloud for GCP Artifact Registry

jychp:
### Type of change

### Summary

Add code-to-cloud tracing for GCP Artifact Registry container images. This enables linking GCP-hosted container images back to their source code repositories, completing the code-to-cloud visibility chain for GCP.

**How it works:**
- During GCP Artifact Registry sync, fetches OCI image configs from the Docker Registry API in a single async pass
- Extracts build provenance from OCI config labels (`org.opencontainers.image.source`), BuildKit VCS metadata, and SLSA attestation referrers
- Sets `source_uri` and `source_revision` on GCP image nodes, enabling the existing GitHub/GitLab supply chain modules to automatically create `PACKAGED_FROM` relationships via provenance matching
- Extracts `layer_diff_ids` and layer history, creating `ImageLayer` nodes for Dockerfile analysis fallback
- Adds a shared `get_unmatched_gcp_images_with_history()` helper so GitHub and GitLab supply chain modules include GCP images in Dockerfile command-similarity matching

**Files added:**
- `cartography/intel/gcp/artifact_registry/supply_chain.py` — OCI config fetching, provenance extraction, ImageLayer creation
- `cartography/models/gcp/artifact_registry/image_layer.py` — GCPArtifactRegistryImageLayer node schema

**Files modified:**
- `cartography/models/gcp/artifact_registry/container_image.py` — Provenance enrichment schema (two-schema pattern like ECR)
- `cartography/intel/gcp/artifact_registry/manifest.py` — Refactored URI parsing into reusable helpers
- `cartography/intel/gcp/artifact_registry/artifact.py` — Added `docker_images_raw` to sync result
- `cartography/intel/gcp/artifact_registry/__init__.py` — Wired supply chain sync
- `cartography/intel/supply_chain.py` — Added shared OCI label extraction and GCP unmatched image query helpers
- `cartography/intel/github/supply_chain.py` — Added GCP images to Dockerfile analysis
- `cartography/intel/gitlab/supply_chain.py` — Added GCP images to Dockerfile analysis

### How was this tested?

- All 1499 unit tests pass (`make test_unit`)
- Pre-commit hooks pass (flake8, black, isort, mypy, pyupgrade)
- Code reviewed with cubic review (no issues)

#### General

#### Proof of functionality

### Notes for reviewers

- The provenance enrichment uses the two-schema pattern (like ECR) to prevent basic API loads from nulling out provenance fields
- The supply chain module reuses existing shared utilities from `cartography/intel/supply_chain.py` (normalize_vcs_url, extract_image_source_provenance, etc.)
- GCP images get the `Image` ontology label via conditional labels, so existing GitHub/GitLab provenance matchlinks (`Image.source_uri == repo_url`) work across providers with no new matchlink models needed


cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,412 @@
...
+    accept: str | None = None,
+) -> dict[str, Any] | None:
+    headers: dict[str, str] = {"Authorization": f"Bearer {auth_token}"}
+    if accept:
+        headers["Accept"] = accept
+    try:
+        resp = await http_client.get(url, headers=headers, timeout=30.0)
+        resp.raise_for_status()
+        return resp.json()
+    except (httpx.HTTPError, json.JSONDecodeError) as e:
```
kunaals: i think we should distinguish "this image has no enrichment" from "we failed to fetch this image's config/blob". the per-image failures can stay non-fatal, but if the enrichment pass is partial we probably shouldn't treat the resulting layer set as authoritative for stale layer cleanup.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,412 @@
...
+            layer_digest = layer.get("digest")
+            if not layer_digest:
+                continue
+
+            blob_url = build_blob_url(registry, image_path, layer_digest)
+            blob = await _fetch_json(http_client, blob_url, auth_token)
+            if not blob:
+                continue
+
+            predicate = blob.get("predicate", {})
```
kunaals: i think this misses the dsse/cosign envelope shape that our GitLab attestation path already handles, where the blob has a base64 `payload` containing the in-toto statement. could we reuse/generalize that decoding before extracting provenance here?

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,412 @@
...
+    return result
+
+
+async def _fetch_all_image_provenance(
+    credentials: GoogleCredentials | None,
+    docker_artifacts_raw: list[dict[str, Any]],
+    project_id: str,
+    max_concurrent: int = 50,
+) -> list[dict[str, Any]]:
+    resolved = _resolve_credentials(credentials)
```
kunaals, line 330: for large registries, a token captured once here can expire mid-run. i think we should retry a 401 after refreshing credentials so a long enrichment pass doesn't silently degrade into partial data.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,412 @@
...
+        history_by_idx: dict[int, str | None] = {}
+        non_empty_idx = 0
+        for entry in history_entries:
+            if entry.get("empty_layer", False):
+                continue
+            history_by_idx[non_empty_idx] = entry.get("created_by") or None
+            non_empty_idx += 1
+
+        for idx, diff_id in enumerate(diff_ids):
+            if diff_id not in layers_by_diff_id:
```
kunaals: small correctness issue: if the first occurrence of a shared diff id has no history but a later image has a populated `created_by`, we keep the empty one forever. can we prefer populated history when we re-encounter an existing diff id?

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,412 @@
...
+                layers_by_diff_id[diff_id] = {
+                    "diff_id": diff_id,
+                    "is_empty": False,
+                    "history": history_by_idx.get(idx),
+                }
+
+    return list(layers_by_diff_id.values())
+
+
+@timeit
```
kunaals, line 424: can we add focused tests for this new enrichment path? i'd expect at least coverage for oci label provenance, dsse/referrers provenance, layer history alignment, partial-fetch cleanup suppression, and the empty-enrichment cleanup case.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,412 @@
...
+        loop = asyncio.get_event_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+
+    enrichments = loop.run_until_complete(
+        _fetch_all_image_provenance(credentials, docker_artifacts_raw, project_id),
+    )
+
+    if not enrichments:
```
kunaals: can we still run stale layer cleanup when discovery completed but there are no enrichments? otherwise deleted images, or projects that now only have non-enrichable artifacts, leave old `GCPArtifactRegistryImageLayer` nodes behind forever.

kunaals (changes requested):
few changes

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: though im not entirely sure what is best here because i dont think we should make the cleanup too complicated

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,412 @@
...
+            if entry.get("empty_layer", False):
+                continue
+            history_by_idx[non_empty_idx] = entry.get("created_by") or None
+            non_empty_idx += 1
+
+        for idx, diff_id in enumerate(diff_ids):
+            if diff_id not in layers_by_diff_id:
+                layers_by_diff_id[diff_id] = {
+                    "diff_id": diff_id,
+                    "is_empty": False,
```
kunaals: i think `is_empty` is confusing here because every layer node is keyed by a rootfs diff id, and empty history entries do not have diff ids. since this is always false, could we either drop it from the GCP layer schema or document/compute a value with real meaning?

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -0,0 +1,515 @@
...
+async def _fetch_json(
+    http_client: httpx.AsyncClient,
+    url: str,
+    token_manager: _TokenManager,
+    accept: str | None = None,
+) -> dict[str, Any]:
+    """Fetch a JSON document. Refreshes credentials once on 401, otherwise raises."""
+    for attempt in range(2):
+        observed_generation = token_manager.generation
+        headers: dict[str, str] = {"Authorization": f"Bearer {token_manager.token}"}
```
kunaals: small auth fidelity thing: could `_TokenManager` expose a helper that calls `credentials.apply(headers)` instead of manually setting `Authorization`? google-auth's `apply()` also preserves things like `x-goog-user-project` when `quota_project_id` is set, so this would stay closer to the official credential behavior while keeping the async `httpx` path.

kunaals (approved):
approving with 1 nit

PR #2692, opened 2026-04-25
feat(AIBOM): Add DETECTED_BY and SCANNED_BY rels between AIBOM and Image 

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- AIBOM sources and components now connect to the provider-agnostic :Image ontology label via _ont_digest instead of targeting ECRImage directly. This means AIBOM scan results automatically link to images from any provider (ECR, GCP Artifact Registry, GitLab Container Registry, etc.) as long as they carry the :Image semantic label.
- Digest-based URIs in the AIBOM report(repo@sha256:...) extract the digest directly and verify it against any :Image node in the graph. This is the primary path and is fully provider-agnostic.
- Tag-based URIs in the AIBOM report (repo:tag) require a fallback resolution path that traverses provider-specific graph relationships (ECRRepositoryImage → ECRImage) to resolve the tag to image digests. This path must remain provider-specific because there is no unified ontology-level tag-to-image relationship — each provider models tag resolution differently. For manifest lists, the fallback traverses CONTAINS_IMAGE to resolve to all child single-platform images rather than the manifest list itself.

### How was this tested?

 - Integration Test Suite testing all 3 scenarios for both AIBOM---SCANNED_IMAGE--->Image and AIBOMComponent---DETECTED_IN--->Image  (digest fast path, tag single platform slow path, tag manifest list slow path)

**AIBOM Digest Based URI (AIBOMSource---SCANNED_IMAGE--->Image)**
image

**AIBOM Digest Based URI (AIBOMComponent---DETECTED_IN--->Image)**
image

**AIBOM Tag Based URI Single Platform Fallback (AIBOMSource---SCANNED_IMAGE--->Image)**
image

**AIBOM Tag Based URI Single Platform Fallback (AIBOMComponent---DETECTED_IN--->Image)**
image

**AIBOM Tag Based URI Manifest List Fallback (AIBOMSource---SCANNED_IMAGE--->Image)**
583665025-738592b3-e9f9-4192-99d1-82c448416451

**AIBOM Tag Based URI Manifest List Fallback (AIBOMComponent---DETECTED_IN--->Image)**
583665170-4cb2e120-adf4-4eb2-a96b-a83d67b03150

#### General

#### Proof of functionality

### Notes for reviewers

 - Looking to make a follow up PR once this is merged to add unified ontology-level tag-to-image relationship so that we can remove all provider specific queries.


tests/integration/cartography/intel/aibom/test_aibom.py
```diff
@@ -786,3 +828,242 @@ def test_cleanup_aibom_removes_stale_nodes(
...
+        MERGE (repo_img)-[r:IMAGE]->(img)
+        SET r.lastupdated = $lastupdated
+        """,
+        image_uri=TEST_TAG_SINGLE_PLATFORM_IMAGE_URI,
+        digest=TEST_TAG_SINGLE_PLATFORM_DIGEST,
+        lastupdated=TEST_UPDATE_TAG,
+    )
+
+
+def _seed_tag_manifest_list_graph(neo4j_session) -> None:
```
kunaals: can we seed this through the existing ecr sync path (or some other image registry) instead of manually creating the image graph? this test is validating the integration between aibom and the image graph, so it would be stronger if `ECRRepositoryImage`, `Image` / `ImageManifestList`, `_ont_digest`, and `CONTAINS_IMAGE` came from the same sync code that creates them in real runs. this is the same test-fidelity idea google calls out here: prefer the real implementation when it's practical, because hand-built test behavior can drift from production behavior. https://testing.googleblog.com/2024/02/increase-test-fidelity-by-avoiding-mocks.html

kunaals (changes requested):

tests/integration/cartography/intel/aibom/test_aibom.py
shyammukund: Decided to seed using the existing ECR sync path data.

kunaals (approved):
approving to unblock, i think `_seed_image_resolution()` and `_seed_multi_image_resolution_graph()` is dead code now - i dont see them used by the unit tests

PR #2700, opened 2026-04-27
perf(matchlinks): Support opt-in scoped MatchLinks

kunaals:
### Type of change
- [x] Documentation update

### Summary
h/t @jychp for the idea

Adds an opt-in MatchLink sub-resource scope so selected source and/or target node matches can be anchored through their tenant/account/project relationship instead of matching only by label-wide endpoint properties.

This keeps existing MatchLinks unchanged by default and enables the scoped query shape for two high-confidence cases where the endpoint nodes are already scoped to the same sub-resource used by `load_matchlinks()`.

### Related issues or links

N/A

### Breaking changes

None.

### How was this tested?

- Unit and integration tests

#### General

#### Proof of functionality

### Notes for reviewers

This does not change the graph schema or MatchLink cleanup semantics. `_sub_resource_label` and `_sub_resource_id` are still relationship properties used for cleanup; the new fields only opt specific MatchLink endpoint matches into an anchored query shape.

### DIFF
--- a/cartography/models/core/relationships.py
+++ b/cartography/models/core/relationships.py
@@ -236,6 +236,8 @@ class MatchLinkSubResource:
 
     This helper lets a MatchLink schema opt into matching an endpoint through its
     sub-resource relationship, such as:
+        MATCH (sub_resource:Parent {id: $PARENT_ID})
+        UNWIND $DictList as item
         MATCH (from:Child {id: item.child_id})<-[:RESOURCE]-(sub_resource)
 
     Direction follows the same convention as CartographyRelSchema: it describes
--- a/cartography/models/core/relationships.py
+++ b/cartography/models/core/relationships.py
@@ -247,6 +247,15 @@ class MatchLinkSubResource:
     direction: LinkDirection
     rel_label: str
 
+    def __post_init__(self) -> None:
+        for key, property_ref in vars(self.target_node_matcher).items():
+            if not property_ref.set_in_kwargs:
+                raise ValueError(
+                    "MatchLinkSubResource target_node_matcher PropertyRefs must "
+                    f"have set_in_kwargs=True. Got {key}="
+                    f"PropertyRef({property_ref.name!r})."
+                )
+
 
 @dataclass(frozen=True)
 class CartographyRelSchema(abc.ABC):
--- a/cartography/models/core/relationships.py
+++ b/cartography/models/core/relationships.py
@@ -230,6 +230,25 @@ def make_source_node_matcher(key_ref_dict: Dict[str, PropertyRef]) -> SourceNode
 
 
 @dataclass(frozen=True)
+class MatchLinkSubResource:
+    """
+    Sub-resource scoping metadata for load_matchlinks() source or target matches.
+
+    This helper lets a MatchLink schema opt into matching an endpoint through its
+    sub-resource relationship, such as:
+        MATCH (from:Child {id: item.child_id})<-[:RESOURCE]-(sub_resource)
+
+    Direction follows the same convention as CartographyRelSchema: it describes
+    the relationship from the MatchLink endpoint node to the sub-resource node.
+    """
+
+    target_node_label: str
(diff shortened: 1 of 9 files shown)


cartography/models/core/relationships.py
```diff
@@ -229,6 +229,25 @@ def make_source_node_matcher(key_ref_dict: Dict[str, PropertyRef]) -> SourceNode
...
+    sub-resource relationship, such as:
+        MATCH (from:Child {id: item.child_id})<-[:RESOURCE]-(sub_resource)
+
+    Direction follows the same convention as CartographyRelSchema: it describes
+    the relationship from the MatchLink endpoint node to the sub-resource node.
+    """
+
+    target_node_label: str
+    target_node_matcher: TargetNodeMatcher
+    direction: LinkDirection
```
jychp, line 249: The sub-resource MATCH is placed before the `UNWIND $DictList as item` in the generated query. If a schema author uses a `PropertyRef` without `set_in_kwargs=True` here, the match clause will reference `item.xxx` which is unbound at that point — Neo4j will error at runtime.

jychp: You should check that target_node_matcher have set_in_kwargs=True, that will prevent this edge case.
This is required for sub_resource relationship anyway.

jychp (changes requested):

cartography/models/core/relationships.py
kunaals: good catch. added validation for this and documented the kwargs requirement.

tests/unit/cartography/graph/test_matchlink.py
```diff
@@ -44,6 +56,110 @@ def test_build_matchlink_query(_mock_get_cartography_version):
...
+                r._module_version = "3.14.16",
+                r.lastupdated = $UPDATE_TAG,
+                r.permission_action = item.permission_action,
+                r._sub_resource_label = $_sub_resource_label,
+                r._sub_resource_id = $_sub_resource_id;
+    """
+
+    actual_query = remove_leading_whitespace_and_empty_lines(link_query)
+    expected_query = remove_leading_whitespace_and_empty_lines(expected)
+    assert actual_query == expected_query
```
jychp, line 190: Suggestion: add a unit test for the case where source and target sub-resources are both set but unequal (e.g., different `target_node_label`). The current tests cover source-only, target-only, and the equal/dedup'd path, but the `else` branch in `build_matchlink_query` that emits two distinct `MATCH` statements is uncovered, so a future regression there would slip through.

cartography/models/core/relationships.py
```diff
@@ -229,6 +229,34 @@ def make_source_node_matcher(key_ref_dict: Dict[str, PropertyRef]) -> SourceNode
...
 
 
+@dataclass(frozen=True)
+class MatchLinkSubResource:
+    """
+    Sub-resource scoping metadata for load_matchlinks() source or target matches.
+
+    This helper lets a MatchLink schema opt into matching an endpoint through its
+    sub-resource relationship, such as:
+        MATCH (from:Child {id: item.child_id})<-[:RESOURCE]-(sub_resource)
```
jychp, line 241: Nitpick: the docstring example shows the sub-resource as a bare `(sub_resource)` node with no inline match clause, which hides the fact that it is matched in a separate `MATCH (sub_resource:Label{id: $_sub_resource_id})` statement hoisted above `UNWIND`. Show the full two-line shape so readers don't assume `sub_resource` is bound from the endpoint MATCH itself.

jychp (approved):
LGTM thx for handling that, some minor comments (feel free to discard them)

cartography/models/core/relationships.py
kunaals: makes sense. updated the example to show the hoisted match shape.

tests/unit/cartography/graph/test_matchlink.py
kunaals: good call. added coverage for the unequal sub-resource branch.

PR #2706, opened 2026-04-27
perf(gcp): optimize Artifact Registry image writes

kunaals:
### Type of change

### Summary
This improves GCP Artifact Registry image ingestion tail behavior and observability for projects with large image volumes.

The change stays local to GAR and keeps the existing generic loader/querybuilder APIs intact. GAR now writes container and platform image nodes separately from relationship attachment, but the relationship phases are modeled as Cartography MatchLinks rather than handwritten GAR Cypher or new generic `load()` modes.

For GAR image nodes, the loader builds the existing schema-driven node-only ingestion query once with `build_ingestion_query(..., selected_relationships=set())`, ensures indexes once, and feeds batches through existing `load_graph_data()`. GAR relationships use small GAR MatchLink schemas and the existing MatchLink querybuilder path. This avoids re-MERGE/re-SET node work during relationship phases and avoids the previous per-parent manifest grouping concern.

This primarily optimizes write tail behavior and production observability rather than reducing the raw number of graph entities written. It preserves the existing graph shape, labels, ontology fields, cleanup behavior, idempotency, and expected relationship metadata while making long-running GAR image phases visible through per-batch progress logs. The MatchLink-managed GAR relationship phases also add relationship cleanup metadata via `_sub_resource_label` and `_sub_resource_id`.

### Related issues or links
None.

### Breaking changes
None.

### How was this tested?
- Live read-only validation against an existing GCP Artifact Registry project using a disposable local Neo4j database. The run exercised `cartography.intel.gcp.artifact_registry.sync()` twice and validated container image nodes, platform image nodes, project/repository/parent relationships, labels, relationship metadata, and stable `firstseen` across reruns.

#### General

#### Proof of functionality

### Notes for reviewers
This does not change the GAR node schema or graph shape. The MatchLink-managed GAR relationship phases add cleanup metadata via `_sub_resource_label` and `_sub_resource_id`, so the schema docs and schema README are unchanged. The sync-log checkbox is left unchecked because this PR changes write shape and observability for existing GAR entities, not the external API fetch behavior or graph model.


cartography/intel/gcp/artifact_registry/artifact.py
```diff
@@ -638,13 +649,58 @@ def load_docker_images(
...
+            f"Artifact Registry container image nodes for project {project_id}"
+        ),
+        lastupdated=update_tag,
+        PROJECT_ID=project_id,
+    )
+    # Container image conditional labels are scoped through the project RESOURCE
+    # relationship, so apply them once after nodes and that relationship exist.
+    load_matchlinks_with_progress(
+        neo4j_session,
+        GCPArtifactRegistryProjectToContainerImageRel(),
```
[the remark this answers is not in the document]
kunaals: good catch on wanting this covered. existing node-schema cleanup handles these rels by lastupdated, and i added integration coverage for stale container/platform matchlink-managed rel cleanup.

cartography/models/gcp/artifact_registry/container_image.py
```diff
@@ -8,8 +8,10 @@
 from cartography.models.core.relationships import CartographyRelProperties
```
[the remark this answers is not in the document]
kunaals: keeping this as-is. matchlink is the existing data-model primitive that gives us relationship-only batched writes with firstseen/metadata and cleanup metadata, without raw cypher or shared loader changes.

cartography/intel/gcp/artifact_registry/util.py
```diff
@@ -27,6 +42,133 @@
...
+        started_at = time.monotonic()
+        load_graph_data(
+            neo4j_session,
+            query,
+            data_batch,
+            batch_size=len(data_batch),
+            **kwargs,
+        )
+        cumulative += len(data_batch)
+        logger.info(
```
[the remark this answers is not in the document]
kunaals: keeping this at info intentionally. the production issue was an opaque write tail, so per-batch info-level progress is part of the observability fix.

cartography/intel/gcp/artifact_registry/artifact.py
```diff
@@ -638,15 +642,60 @@ def load_docker_images(
...
+        apply_labels=False,
+        progress_description=(
+            f"Artifact Registry container image nodes for project {project_id}"
+        ),
+        lastupdated=update_tag,
+        PROJECT_ID=project_id,
+    )
+    # Container image conditional labels are scoped through the project RESOURCE
+    # relationship, so apply them once after nodes and that relationship exist.
+    load_matchlinks_with_progress(
```
jychp, line 663: I did not find where we clean these relationships.

jychp (changes requested):

cartography/intel/gcp/artifact_registry/artifact.py
kunaals: this is still cleaned by the existing node-schema cleanup path: cleanup_docker_images() runs GraphJob.from_node_schema(GCPArtifactRegistryContainerImageSchema(), ...), and these matchlink writes use the same RESOURCE/CONTAINS relationship shapes and lastupdated fields declared on that schema. i added integration coverage for stale matchlink-managed rel cleanup too

kunaals: if you think there is a better way to do this or something more explicit let me know!

kunaals: hm actually i'll add the explicit matchlink cleanup

kunaals: good point. i made this more explicit now: the existing node-schema cleanup still runs first for stale node deletion, then each gar matchlink-managed relationship runs `GraphJob.from_matchlink()` cleanup scoped to the gcp project.

jychp (approved):
LGTM

PR #2710, opened 2026-04-28
feat(github): add GHCR container packages sync

### LINKED ISSUE #2315: feat(github): Add support for GitHub Container Registry (GHCR)
## Summary

Add support for GitHub Container Registry (GHCR) to ingest container images into the graph, similar to existing Amazon ECR and GitLab Container Registry modules.

## Motivation

GHCR is widely used for storing container images alongside GitHub repositories. Adding support would enable users to:
- Map container images to their source repositories
- Track image metadata (tags, digests, visibility)
- Correlate container assets with GitHub organizations

## References

- [Amazon ECR module](https://github.com/lyft/cartography/tree/master/cartography/intel/aws/ecr)
- [GitLab Container Registry module](https://github.com/lyft/cartography/tree/master/cartography/intel/gitlab)
- [GitHub Packages API docs](https://docs.github.com/en/rest/packages)

jychp:
### Type of change

### Summary

Sync container packages hosted on GitHub Container Registry (`ghcr.io`) for each configured GitHub organization. Cartography already covers ECR, GCP Artifact Registry, and GitLab container registries; GHCR was the missing piece, and without it images pushed to GHCR could not be tied to their `GitHubRepository` source or to deployed assets.

The new pipeline mirrors the existing GitLab container-registry split inside `cartography/intel/github/`:

1. **Packages** → `GitHubPackage` (one per `/orgs/{org}/packages?package_type=container`). Carries the `ContainerRegistry` extra label and is linked to its `GitHubRepository` via `HAS_PACKAGE` when the package payload includes a repo.
2. **Versions → Container images** → `GitHubContainerImage` with conditional `Image` / `ImageManifestList` labels. Manifests come from `ghcr.io/v2/{org}/{name}/manifests/{digest}`; image config blobs supply `layer_diff_ids`, `architecture`, `os`.
3. **Layers** → `GitHubContainerImageLayer` with `HAS_LAYER` / `HEAD` / `TAIL` / `NEXT` rels (linked-list ordering matching the ECR/GitLab pattern).
4. **Tags** → `GitHubContainerImageTag` with `(:ContainerRegistry)-[:REPO_IMAGE]->(:ImageTag)-[:IMAGE]->(:Image)` so the existing supply-chain matcher picks GHCR images up without changes.
5. **Attestations** (SLSA via `/orgs/{org}/attestations/{digest}`) enrich images with `source_uri` / `source_revision` / `source_file` through a provenance-only schema (same trick as `GCPArtifactRegistryContainerImageProvenanceSchema` to avoid nulling out enrichment fields on re-load).

API call shape is bounded — there is no bulk endpoint, so the sync de-duplicates per-run by digest and config-digest, and skips manifest/config/attestation fetches for digests already enriched in a prior run, dropping steady-state cost from `O(versions)` to `O(new versions)`.

The configured GitHub token is reused for GHCR; the OCI Distribution endpoints accept the same Bearer credential, just base64-encoded. The token must have the `read:packages` scope. No CLI surface change.

### Related issues or links

- Closes https://github.com/cartography-cncf/cartography/issues/2315

### How was this tested?

- `uv run --frozen pytest -vvv tests/integration/cartography/intel/github/test_packages.py` — new end-to-end test exercising packages + manifests + layers + tags + attestations through the full pipeline (mocked REST and registry calls).
- `uv run --frozen pytest -vvv tests/integration/cartography/intel/github/` — full GitHub integration suite (29 tests, no regressions).
- `make test_lint` — pre-commit/black/isort/mypy clean on changed files.

### Notes for reviewers

- The supply-chain Cypher in `cartography/intel/github/supply_chain.py:60-67` matches on the generic `Image` / `ImageTag` / `ContainerRegistry` labels; once GHCR nodes carry those labels, no Cypher changes are needed for cross-registry digest matching.
- `fetch_ghcr_manifest` / `fetch_ghcr_blob` were added next to `call_github_rest_api` in `cartography/intel/github/util.py` and reuse `_resolve_token` so AppCredential refresh continues to work.
- `repos_json` and the workflow list are passed in from `repos.sync` / `actions.sync`; no extra REST calls beyond the GHCR-specific endpoints.

### Test plan

- [ ] Run `make test_lint` and `uv run --frozen pytest -vvv tests/integration/cartography/intel/github/`.
- [ ] Run cartography against a real org with GHCR images using a token that has `read:packages`. Verify in the Neo4j browser:
  - `MATCH (p:GitHubPackage) RETURN count(p)` is non-zero.
  - `MATCH (i:GitHubContainerImage)-[:HAS_LAYER]->(l) RETURN count(*)` is non-zero for multi-layer images.
  - `MATCH (img:GitHubContainerImage)-[:PACKAGED_FROM]->(repo:GitHubRepository) RETURN repo.url, count(img)` confirms cross-registry supply-chain matching picks up GHCR images.
- [ ] Re-run the sync twice and confirm cleanup deletes stale tags/versions (delete a tag in GHCR between runs to verify).

### DIFF
--- a/cartography/intel/github/__init__.py
+++ b/cartography/intel/github/__init__.py
@@ -132,7 +132,7 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
 
         # Sync GHCR (container packages, image manifests, tags, attestations).
         # Runs before supply_chain.sync so the latter can correlate digests.
-        ghcr_packages = cartography.intel.github.packages.sync_packages(
+        ghcr_result = cartography.intel.github.packages.sync_packages(
             neo4j_session,
             token,
             api_url,
@@ -140,7 +140,7 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
             common_job_parameters["UPDATE_TAG"],
             common_job_parameters,
         )
-        if ghcr_packages:
+        if ghcr_result.packages:
             (
                 ghcr_manifests,
                 _ghcr_manifest_lists,
@@ -150,7 +150,7 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
                 token,
                 api_url,
                 org_name,
-                ghcr_packages,
+                ghcr_result.packages,
                 common_job_parameters["UPDATE_TAG"],
                 common_job_parameters,
             )
--- a/cartography/intel/github/__init__.py
+++ b/cartography/intel/github/__init__.py
@@ -8,6 +8,10 @@ import neo4j
 
 import cartography.intel.github.actions
 import cartography.intel.github.commits
+import cartography.intel.github.container_image_attestations
+import cartography.intel.github.container_image_tags
+import cartography.intel.github.container_images
+import cartography.intel.github.packages
 import cartography.intel.github.repos
 import cartography.intel.github.supply_chain
 import cartography.intel.github.teams
@@ -125,6 +129,48 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
         )
         # Filter out None entries
         valid_repos = [r for r in repos_json if r is not None]
+
+        # Sync GHCR (container packages, image manifests, tags, attestations).
+        # Runs before supply_chain.sync so the latter can correlate digests.
+        ghcr_packages = cartography.intel.github.packages.sync_packages(
+            neo4j_session,
+            token,
+            api_url,
+            org_name,
+            common_job_parameters["UPDATE_TAG"],
+            common_job_parameters,
+        )
+        if ghcr_packages:
+            (
+                ghcr_manifests,
+                _ghcr_manifest_lists,
+                ghcr_tag_rows,
+            ) = cartography.intel.github.container_images.sync_container_images(
+                neo4j_session,
+                token,
+                api_url,
+                org_name,
+                ghcr_packages,
+                common_job_parameters["UPDATE_TAG"],
+                common_job_parameters,
+            )
+            cartography.intel.github.container_image_tags.sync_container_image_tags(
+                neo4j_session,
+                org_name,
+                ghcr_tag_rows,
+                common_job_parameters["UPDATE_TAG"],
+                common_job_parameters,
+            )
+            cartography.intel.github.container_image_attestations.sync_container_image_attestations(
+                neo4j_session,
+                token,
+                api_url,
+                org_name,
+                ghcr_manifests,
+                common_job_parameters["UPDATE_TAG"],
+                common_job_parameters,
+            )
+
         if valid_repos:
             cartography.intel.github.supply_chain.sync(
                 neo4j_session,
--- a/cartography/intel/github/__init__.py
+++ b/cartography/intel/github/__init__.py
@@ -132,6 +132,11 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
 
         # Sync GHCR (container packages, image manifests, tags, attestations).
         # Runs before supply_chain.sync so the latter can correlate digests.
+        # Gate on cleanup_safe — not on the packages list — so an org that
+        # legitimately has zero packages still gets its stale GHCR images,
+        # tags, and attestations reaped. An endpoint outage or missing-scope
+        # condition flips cleanup_safe to False, which disables both the
+        # fetches and the downstream cleanups.
         ghcr_result = cartography.intel.github.packages.sync_packages(
             neo4j_session,
             token,
@@ -140,11 +145,12 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
             common_job_parameters["UPDATE_TAG"],
             common_job_parameters,
         )
-        if ghcr_result.packages:
+        if ghcr_result.cleanup_safe:
             (
                 ghcr_manifests,
                 _ghcr_manifest_lists,
                 ghcr_tag_rows,
+                ghcr_observed_and_skipped,
             ) = cartography.intel.github.container_images.sync_container_images(
                 neo4j_session,
                 token,
@@ -169,6 +175,7 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
                 ghcr_manifests,
                 common_job_parameters["UPDATE_TAG"],
                 common_job_parameters,
+                additional_observed_digests=ghcr_observed_and_skipped,
             )
 
         if valid_repos:
--- a/cartography/intel/github/container_image_attestations.py
+++ b/cartography/intel/github/container_image_attestations.py
@@ -54,8 +54,7 @@ def _digests_already_enriched(
     org_url: str,
 ) -> set[str]:
     query = """
-    MATCH (org:GitHubOrganization {id: $org_url})
-    MATCH (img:GitHubContainerImage)-[:RESOURCE]->(org)
+    MATCH (org:GitHubOrganization {id: $org_url})-[:RESOURCE]->(img:GitHubContainerImage)
     WHERE img.source_uri IS NOT NULL
     RETURN img.digest
     """
--- a/cartography/intel/github/container_image_attestations.py
+++ b/cartography/intel/github/container_image_attestations.py
@@ -1,0 +1,307 @@
+"""
+GitHub Container Image Attestations Intelligence Module.
+
+Pulls SLSA attestations for image digests via the GitHub Attestations API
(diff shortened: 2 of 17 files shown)


jychp:
Screenshot 2026-04-29 at 11 57 45 AM

cartography/intel/github/packages.py
```diff
@@ -0,0 +1,214 @@
...
+        # Older GitHub Enterprise versions don't expose this endpoint.
+        if err.response is not None and err.response.status_code == 404:
+            logger.warning(
+                "GitHub Packages endpoint not available for org %s (404). "
+                "GHCR sync will be skipped and cleanup deferred so previously "
+                "synced packages are not purged by an endpoint outage.",
+                organization,
+            )
+            return ContainerPackagesFetchResult(packages=[], cleanup_safe=False)
+        raise
```
kunaals, line 95: verified live: hit /orgs/<org>/packages?package_type=container with a token missing read:packages and got

  HTTP/2.0 403 Forbidden
  {"message":"You need at least read:packages scope to list packages.",...}

docs: https://docs.github.com/rest/packages/packages#list-packages-for-an-organization

i think this except is unreachable — fetch_all_rest_api_pages (util.py:465-471) already swallows 403 → []. we'd then go down the success branch with packages=[] and cleanup_safe=True, and cleanup would delete every GitHubPackage for the org. fine-grained PATs hit this too (per github community #52868), so afaict it's the most common token misconfig in the wild.

i'd lean toward a `raise_on_status=(403,404)` escape hatch on fetch_all_rest_api_pages and using it from this caller, but a single-page probe before paginating would also work — open to either.

cartography/intel/github/container_images.py
```diff
@@ -0,0 +1,469 @@
...
+
+    if layers:
+        load_container_image_layers(neo4j_session, layers, org_url, update_tag)
+    if images:
+        load_container_images(neo4j_session, images, org_url, update_tag)
+
+    cleanup_params = dict(common_job_parameters)
+    cleanup_params["org_url"] = org_url
+    cleanup_container_image_layers(neo4j_session, cleanup_params)
+    cleanup_container_images(neo4j_session, cleanup_params)
```
kunaals, line 564: i think the skip at line 189 leads to data loss: digests already in `_existing_digests_with_layers` never make it into `raw_manifests`, so transform emits no rows, lastupdated stays old, and cleanup at 468 (cleanupbuilder.py:280-282 matches `lastupdated <> $UPDATE_TAG`) sweeps them. layers go too. tag rows survive because they're appended at 175-186 before the skip, so we'd end up with orphan tags pointing at deleted images.

to be clear on the principle: if a digest is gone from ghcr, the versions api stops returning it, the loop never visits it, and cleanup correctly reaps it — that behavior should stay. the bug is the *other* case: digest still exists in ghcr (still in versions), we skipped the expensive manifest+config re-fetch as a cross-run optimization, but cleanup then wrongly reaps the still-live image.

so the fix should only bump `lastupdated` for digests **observed in versions this run and chosen to skip** — not blindly for every entry in `_existing_digests_with_layers`. roughly:

```python
observed_and_skipped: set[str] = set()
for version in versions:
    digest = ...
    ...
    if digest in skip_digests:
        observed_and_skipped.add(digest)
        continue
    ...
# before cleanup:
# MATCH (n:GitHubContainerImage) WHERE n.digest IN $observed_and_skipped
# SET n.lastupdated = $UPDATE_TAG
```

i'd lean toward this over dropping the skip (which re-introduces the manifest+config fetches the optimization was added to avoid).

ci doesn't catch this — the integration test only runs sync once. could we add a "run sync twice, assert nodes survive" case?

cartography/intel/github/container_image_attestations.py
```diff
@@ -0,0 +1,316 @@
...
+            len(attestation_rows),
+            organization,
+        )
+        load_attestations(neo4j_session, attestation_rows, org_url, update_tag)
+    if provenance_rows:
+        load_image_provenance(neo4j_session, provenance_rows, org_url, update_tag)
+
+    cleanup_params = dict(common_job_parameters)
+    cleanup_params["org_url"] = org_url
+    cleanup_attestations(neo4j_session, cleanup_params)
```
kunaals, line 390: same shape as container_images.py:443 — skip at 289, unconditional cleanup at 316. i think the same lastupdated-bump fix applies.

cartography/intel/github/__init__.py
```diff
@@ -125,6 +129,48 @@ def start_github_ingestion(neo4j_session: neo4j.Session, config: Config) -> None
...
+        ghcr_result = cartography.intel.github.packages.sync_packages(
+            neo4j_session,
+            token,
+            api_url,
+            org_name,
+            common_job_parameters["UPDATE_TAG"],
+            common_job_parameters,
+        )
+        if ghcr_result.packages:
+            (
```
kunaals, line 149: i think gating on `ghcr_result.packages` is too aggressive: if an org legitimately deletes its last package between runs, none of the downstream cleanups (images, tags, attestations) run and stale ghcr nodes get stranded indefinitely. gating on `not cleanup_safe` instead would still skip on endpoint outages but reap empty-package orgs correctly. could be wrong on the intended behavior — let me know.

cartography/intel/github/container_image_attestations.py
```diff
@@ -0,0 +1,316 @@
...
+    ``/orgs/{org}/attestations/{digest}`` is the documented response when the
+    digest has no attestations (it is *not* an error). Only that case is
+    converted into an empty list; every other HTTP error propagates so the
+    sync aborts before cleanup runs and existing attestation nodes are not
+    silently purged.
+    """
+    base_url = rest_api_base_url(api_url)
+    endpoint = _attestations_endpoint(organization, digest)
+    try:
+        response = call_github_rest_api(endpoint, token, base_url)
```
kunaals: endpoint embeds `?per_page=100` (the documented max per https://docs.github.com/rest/orgs/attestations#list-attestations) but `call_github_rest_api` is single-page — i think >100 attestations would silently truncate.

switch to `fetch_all_rest_api_pages`, but only after the 403/404 fix above lands, since 404 here is the documented "no attestations or no access" response and must remain a non-error.

cartography/intel/github/container_image_attestations.py
```diff
@@ -0,0 +1,316 @@
...
+                {
+                    "id": row_id,
+                    "bundle_id": attestation.get("id"),
+                    "predicate_type": predicate_type,
+                    "attests_digest": digest,
+                    "source_uri": provenance["source_uri"],
+                    "source_revision": provenance["source_revision"],
+                    "source_file": provenance["source_file"],
+                },
+            )
```
kunaals, line 221: pulled a real attestation against /repos/cli/cli/attestations/sha256:<...> for v2.92.0 release artifacts:

  {"repository_id": 212613049,
   "bundle_url":   "https://tmaproduction.blob.core.windows.net/...",
   "initiator":    "user",
   "bundle":       null}

i think there's a bigger issue here than a row_id nit:

1. no top-level `id` — `attestation.get("id")` at line 202 is always None, so `bundle_id` is always null in the graph.

2. `bundle` is None and the actual bundle is lazy-served from `bundle_url` (signed short-lived, snappy-compressed `.json.sn` containing a sigstore bundle). line 190 `bundle = attestation.get("bundle") or {}` silently gives `{}`, envelope `{}`, dsse decode `{}`, `_extract_provenance` empty — every row loads with no predicate_type / source_uri / provenance.

i didn't find a container image attestation with the inline-bundle shape to compare, so could be wrong about whether ghcr attestations switched too — worth probing a few real digests before fixing the cleanup bugs above. if they did switch, we'd need to follow `bundle_url` and pick row_id from inside the bundle.

kunaals (changes requested):
a few blockers around silent data loss on the ghcr cleanup paths and an api response shape mismatch i hit when probing live attestations

cartography/intel/github/container_images.py
```diff
@@ -0,0 +1,543 @@
...
+    """
+    if not digests:
+        return
+    query = """
+    MATCH (org:GitHubOrganization {id: $org_url})-[r_org:RESOURCE]->(img:GitHubContainerImage)
+    WHERE img.digest IN $digests
+    SET img.lastupdated = $update_tag,
+        r_org.lastupdated = $update_tag
+    WITH img, org
+    OPTIONAL MATCH (img)-[r_layer:HAS_LAYER|HEAD|TAIL]->(layer:GitHubContainerImageLayer)
```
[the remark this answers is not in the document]
kunaals: I think this has merit

kunaals (approved):
approving to unblock, i think that cubic comment is correct though, lets fix that before merging

PR #2714, opened 2026-04-29
feat(gitlab): ingest CI/CD runners, variables, environments, and pipelines config

jychp:
### Type of change

### Summary

Adds GitLab CI/CD ingestion to cartography, reaching parity with the existing GitHub Actions module. Five new node types:

- **`GitLabRunner`** at instance / group / project scopes — security-relevant flags (`run_untagged`, `locked`, `access_level`) surface unprotected runners that accept jobs from any branch.
- **`GitLabCIVariable`** at group / project scopes (`HAS_CI_VARIABLE` semantic edge + `RESOURCE` cleanup edge). The variable's `value` is intentionally never stored — only metadata (`protected`, `masked`, `environment_scope`, ...). GitLab does not distinguish secrets from variables at the API level, so the same node type covers both.
- **`GitLabEnvironment`** with a standard `HAS_CI_VARIABLE` relationship (one_to_many) to applicable CI variables, matching exact `environment_scope` + wildcard `*`.
- **`GitLabCIConfig`** + **`GitLabCIInclude`** parsed from `.gitlab-ci.yml` (via `/ci/lint` with raw fallback). Each include carries an `is_pinned` flag — a `project:` include without a 40-char SHA `ref` is the security signal.

The CI/CD sync runs right after users (before containers / dependencies) so the static security signals land in the graph even if the slower container scan fails.

### Related issues or links

- Fixes #

### Breaking changes

None.

### How was this tested?

- `make test_lint` — passes
- `make test_unit` — passes (53 new unit tests across runners, ci_variables, environments, ci_config + parser)
- `make test_integration` — passes via CI (Neo4j fixture)
- Manual sync against a GitLab.com test group with a project that has runners, group-level variables (`protected=true`), project-level variables (mixed `environment_scope`), environments (`production`, `staging`), and a `.gitlab-ci.yml` with mixed includes (local, project pinned/unpinned, remote, template). Verified with the example queries documented in `docs/root/modules/gitlab/schema.md` (e.g. `MATCH (c:GitLabCIConfig)-[:USES_INCLUDE]->(i) WHERE i.include_type = 'project' AND i.is_pinned = false RETURN ...`).

#### General

#### Proof of functionality

- [x] Cartography sync logs from a real environment — reproduced locally; example output:
  ```
  INFO:cartography.intel.gitlab.runners:Syncing GitLab instance-level runners
  INFO:cartography.intel.gitlab.runners:Loaded 7 instance-level runners
  INFO:cartography.intel.gitlab.runners:Syncing GitLab group-level runners for 4 groups
  INFO:cartography.intel.gitlab.runners:Syncing GitLab project-level runners for 12 projects
  INFO:cartography.intel.gitlab.runners:GitLab runners sync completed
  INFO:cartography.intel.gitlab.ci_variables:Syncing GitLab CI/CD variables for 4 groups and 12 projects
  INFO:cartography.intel.gitlab.ci_variables:GitLab CI/CD variables sync completed
  INFO:cartography.intel.gitlab.environments:Syncing GitLab environments for 12 projects
  INFO:cartography.intel.gitlab.environments:GitLab environments sync completed
  INFO:cartography.intel.gitlab.ci_config:Syncing GitLab CI configs for 12 projects
  INFO:cartography.intel.gitlab.ci_config:GitLab CI configs sync completed
  ```

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (`docs/root/modules/gitlab/schema.md` — mermaid + tables + example queries for the 5 new nodes; `docs/root/modules/gitlab/config.md` — admin scope note for instance runners, Maintainer scope note for `/ci/lint`).
- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md) — N/A, the README only lists top-level integrations and `gitlab` is already listed.

### Notes for reviewers

A few design decisions worth a second look:

- **Three schemas / one label for runners and variables**: `GitLabRunner` and `GitLabCIVariable` exist at multiple scopes (instance/group/project for runners, group/project for variables). I followed the `GitHubActionsSecret` pattern — three schemas sharing the Neo4j label but each with its own `sub_resource_relationship`. This keeps cleanup correctly bounded to the parent of each scope.

- **403 sentinel**: `get_*` helpers return `None` (not `[]`) on 403, and the `sync_*` functions surface a per-scope "skipped" set. The cleanup phase honours that set so a transient permission failure does not delete previously-ingested data. This addresses Cubic's P1 finding.

- **Standard relationships, not MatchLinks, for env→variable and config→variable**: both endpoints share the same sub-resource (the project), so a `one_to_many=True` `OtherRelationships` does the right thing without a separate MatchLink load + cleanup. Mirrors `GitHubWorkflowToSecretRel`.

- **YAML parser is pure** (`cartography/intel/gitlab/ci_config_parser.py`), mirroring `cartography/intel/github/workflow_parser.py`. All HTTP I/O lives in the orchestration module, with a `/ci/lint` → raw `.gitlab-ci.yml` fallback for tokens that lack Maintainer scope.

- **Composite IDs** include `project_id` or `environment_scope` where GitLab's native IDs are not unique on the dimension we need (env IDs are per-project; variables can repeat the same key per environment_scope).

- **Out of scope**: `GitLabPipeline` runtime data (the equivalent of `GitHubWorkflowRun`) — same parity decision as the GitHub module, which models definitions only.

### DIFF
--- a/cartography/intel/gitlab/ci_config.py
+++ b/cartography/intel/gitlab/ci_config.py
@@ -95,12 +95,15 @@ def _try_raw_ci_yaml(
     project_id: int,
     ref: str | None,
     file_path: str = DEFAULT_FILE_PATH,
-) -> str | None:
+) -> tuple[str | None, bool]:
     """
     Fetch the raw `.gitlab-ci.yml` from the repository as a fallback when
-    /ci/lint is unavailable. Returns None on 403 / 404 (the file is absent
-    or the token can't read repository files for this project — both
-    non-fatal for the sync).
+    /ci/lint is unavailable. Returns ``(text, denied)``:
+
+    - ``(text, False)`` on success
+    - ``(None, False)`` on 404 — the file does not exist, authoritative
+    - ``(None, True)`` on 403 — the token cannot read repository files
+    Other errors propagate.
     """
     encoded = quote(file_path, safe="")
     endpoint = f"/api/v4/projects/{project_id}/repository/files/{encoded}/raw"
@@ -117,15 +120,20 @@ def _try_raw_ci_yaml(
         )
         response.raise_for_status()
     except requests.exceptions.HTTPError as e:
-        if e.response is not None and e.response.status_code in (403, 404):
+        if e.response is not None and e.response.status_code == 403:
             logger.warning(
-                "Raw .gitlab-ci.yml not available for project %s (status %s).",
+                "Raw .gitlab-ci.yml denied for project %s.",
                 project_id,
-                e.response.status_code,
             )
-            return None
+            return None, True
+        if e.response is not None and e.response.status_code == 404:
+            logger.warning(
+                "Raw .gitlab-ci.yml not found for project %s.",
+                project_id,
+            )
+            return None, False
         raise
-    return response.text
+    return response.text, False
 
 
 @timeit
@@ -160,17 +168,17 @@ def fetch_ci_config_yaml(
 
     # Both 404 and 403 from /ci/lint fall back to the raw file. The raw
     # endpoint uses `read_repository`, broader than the Maintainer scope
-    # /ci/lint needs — but if the token also lacks `read_repository` for
-    # this project, _try_raw_ci_yaml returns None on 403 too.
-    raw = _try_raw_ci_yaml(gitlab_url, token, project_id, ref)
+    # /ci/lint needs.
+    raw, raw_denied = _try_raw_ci_yaml(gitlab_url, token, project_id, ref)
     if raw is not None:
         return raw, None, False, False
 
-    # Both paths failed. If lint specifically reported 403 we treat the
-    # whole project as denied (skip cleanup). If lint was 404 (no pipeline)
-    # and raw was also missing, the project legitimately has no config and
-    # cleanup may proceed.
-    return None, None, False, lint_denied
+    # Both paths failed. We only flag the project as ``denied`` if neither
+    # path could authoritatively confirm the absence of a config — i.e.
+    # both returned 403. A 404 from either endpoint (especially raw 404)
+    # is authoritative absence and cleanup may run normally.
+    denied = lint_denied and raw_denied
+    return None, None, False, denied
 
 
 def transform_ci_config(
--- a/cartography/intel/gitlab/ci_config.py
+++ b/cartography/intel/gitlab/ci_config.py
@@ -25,6 +25,7 @@ import requests
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.intel.gitlab.ci_config_parser import parse_ci_config
+from cartography.intel.gitlab.ci_config_parser import parse_lint_includes
 from cartography.intel.gitlab.ci_config_parser import ParsedCIConfig
 from cartography.intel.gitlab.util import make_request_with_retry
 from cartography.models.gitlab.ci_config import GitLabCIConfigSchema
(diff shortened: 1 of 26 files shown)


jychp:
Screenshot 2026-04-29 at 2 18 25 PM

cartography/intel/gitlab/runners.py
```diff
@@ -0,0 +1,372 @@
...
+    token: str,
+    group_id: int,
+) -> list[dict[str, Any]] | None:
+    """
+    Fetch group-level runners for a specific group. Returns ``None`` on 403.
+    """
+    return _list_runners_tolerant(
+        gitlab_url,
+        token,
+        f"/api/v4/groups/{group_id}/runners",
```
kunaals, line 103: i think these list calls need to pass `type=group_type` / `type=project_type`. gitlab returns inherited runners here by default, so `/projects/:id/runners` can return instance runners and this would attach them as project-scoped runners for every project. i validated this with glab: the unfiltered project endpoint returned only `instance_type` runners, while `?type=project_type` returned none.

cartography/intel/gitlab/runners.py
```diff
@@ -0,0 +1,372 @@
...
+    """
+    For each runner in the list, fetch its detail record. Skips runners whose
+    detail call fails with 403/404.
+    """
+    enriched: list[dict[str, Any]] = []
+    for runner in listed_runners:
+        runner_id = runner.get("id")
+        if runner_id is None:
+            continue
+        details = get_runner_details(gitlab_url, token, runner_id)
```
kunaals, line 212: can we fall back to the listed runner record here instead of dropping it? the list endpoint just confirmed the runner exists, so if the detail endpoint has a transient 403/404, skipping it means the later cleanup can delete a real runner from the graph. the detail fields can be filled in when the detail call succeeds again.

cartography/intel/gitlab/ci_config.py
```diff
@@ -0,0 +1,386 @@
...
+            return None, None, True
+        if e.response is not None and e.response.status_code == 404:
+            logger.warning(
+                "ci/lint returned 404 for project %s (no pipeline / not enabled).",
+                project_id,
+            )
+            return None, None, False
+        raise
+
+    body = response.json()
```
kunaals, line 92: i think the lint response's `includes` need to be preserved separately from `merged_yaml`. gitlab expands includes out of `merged_yaml`, so parsing only that string drops `GitLabCIInclude` nodes and `USES_INCLUDE` edges on the preferred lint path. i validated with `glab api -X POST .../ci/lint`: `includes` was populated while `merged_yaml` no longer had an `include:` section.

cartography/intel/gitlab/ci_variables.py
```diff
@@ -0,0 +1,264 @@
...
+        transformed = transform_variables(raw, "group", group_id, gitlab_url)
+        load_group_variables(
+            neo4j_session, transformed, group_id, gitlab_url, update_tag
+        )
+
+    project_variables: dict[int, list[dict[str, Any]]] = {}
+    for project in projects:
+        project_id: int = project["id"]
+        raw = get_project_variables(gitlab_url, token, project_id)
+        if raw is None:
```
kunaals, line 250: this still turns a project variable 403 into an empty downstream variable set. env/config sync then reloads nodes with no variable ids, so their relationship cleanup can delete existing env/config -> variable edges even though the variable read was permission-denied. i think the skipped project needs to propagate to those downstream syncs, or they need to avoid refreshing variable-derived edges for that project.

kunaals (changes requested):

cartography/intel/gitlab/ci_config_parser.py
```diff
@@ -0,0 +1,358 @@
...
+    when present, falling back to ``context_sha`` (the resolved commit
+    that GitLab fetched).
+    """
+    if not lint_includes:
+        return []
+    results: list[ParsedCIInclude] = []
+    for entry in lint_includes:
+        if not isinstance(entry, dict):
+            continue
+        include_type = entry.get("type") or ""
```
kunaals, line 305: thanks, this preserves the lint includes now, but i think project includes still need one more normalization step. gitlab returns project includes from `/ci/lint` as `type: "file"` with `extra.project` / `extra.ref`, so this currently stores them as `include_type="file"` and loses the project part of the location. can we map those back to the same shape as the raw parser, e.g. `include_type="project"` and `location="<project>:<file>"`, before creating `GitLabCIInclude` nodes?

kunaals (approved):
approving with one more change to make

cartography/intel/gitlab/ci_config_parser.py
jychp: Done in 8e9b0250 — `parse_lint_includes` now detects `type: "file"` with `extra.project` and rewrites to `include_type="project"`, `location="<project>:<file>"` before building the `GitLabCIInclude` record, so the lint and raw paths produce equivalent nodes. Added a unit test covering both pinned and unpinned cases.

PR #2716, opened 2026-04-29
feat(semgrep): Support for OSS Semgrep SAST Findings 

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- This PR adds support for ingesting OSS Semgrep SAST findings from explicit Semgrep JSON report artifacts described in a repository mapping YAML file.
    - OSS findings are represented with the same `SemgrepSASTFinding` node label as Cloud findings, so they are unified in the graph for querying.

    - Design choices
        - `OSSSemgrepSASTFindingSchema` uses `label = "SemgrepSASTFinding"` so OSS and Cloud findings coexist as the same node type, but the models remain separate because the Semgrep Cloud API and OSS CLI JSON output expose different fields.
        - OSS findings do not come from a real Semgrep Cloud deployment, so a synthetic `SemgrepDeployment` node with `id="oss"` is created as the sub-resource parent. This keeps OSS findings in the same Semgrep data model while remaining isolated from Cloud ingestion.
        - `--semgrep-oss-source` now points to a repository mapping YAML file rather than a directory/prefix of reports. The mapping file contains a top-level `repositories` list, and each repository entry must include `provider`, `owner`, `repo`, `url`, `branch`, and a non-empty `reports` list.
        - Each `reports` entry must resolve to exactly one Semgrep OSS JSON artifact. A single mapping file can describe multiple repositories, and a single repository entry can aggregate multiple explicit report artifacts for that repository.
        - OSS reports do not include Semgrep Cloud finding IDs, so a stable synthetic ID is generated and prefixed with `semgrep-oss-sast-`. IDs prefer Semgrep's `extra.fingerprint` when usable and otherwise fall back to a SHA-256 hash of `check_id`, `path`, start/end location, and `repository_url`.
        - Placeholder fingerprint values such as `requires login` are treated as unusable and fall back to the location-based hash. This prevents unrelated findings in the same repo/rule from collapsing when Semgrep OSS emits placeholder fingerprints.
        - Cartography derives:
            - `repository` as `owner/repo`
            - `repository_url` from `url`
            - `branch` from `branch`
        - OSS categories are normalized to the same list shape used by Cloud `SemgrepSASTFinding.categories`.
        - OSS cleanup is now scoped per `repository_url`. One repository entry is treated as the intended snapshot for that repository in the current run. Cleanup only runs for repository entries where all listed reports were successfully processed; if any listed report fails, cleanup is skipped for that repository to avoid deleting findings from an incomplete snapshot.
        - `FOUND_IN` relationships are created when matching `GitHubRepository` nodes already exist with `id` equal to the repository URL from the mapping entry.
        - Uses the same `parse_report_source` / `build_report_reader_for_source` pattern as trivy, syft, aibom, and docker_scout for provider-agnostic report ingestion.

image

### Related issues or links

- Uses provider agnostic report ingestion from https://github.com/cartography-cncf/cartography/pull/2620

### How was this tested?

- Added and updated unit/integration tests for OSS Semgrep ingestion:
    - unit test for repository mapping file happy path
    - unit test for repo-level report aggregation happy path
    - unit tests proving synthetic OSS IDs prefer usable fingerprints and remain distinct across repositories
    - unit tests for failure paths where the mapping source contains multiple YAML files and where a report source resolves to multiple artifacts
    - integration test for single-repo OSS sync using a repository mapping file and the real report-reader path end to end
    - integration test for multi-entry / multi-report OSS sync using a repository mapping file
- Ran:
    - `uv run pytest tests/unit/cartography/intel/semgrep/test_ossfindings.py`
    - `uv run pytest tests/integration/cartography/intel/semgrep/test_findings.py`
- Manually tested OSS Semgrep ingestion against live Semgrep OSS reports for two repositories using a repository mapping file.

#### General

#### Proof of functionality


cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,227 @@
...
+        all_findings.extend(transform_oss_semgrep_sast_report(document))
+
+    if all_findings:
+        logger.info(
+            "Transformed %d total OSS Semgrep SAST findings.", len(all_findings)
+        )
+        load_oss_semgrep_sast_findings(neo4j_session, all_findings, update_tag)
+
+    common_job_parameters["DEPLOYMENT_ID"] = OSS_DEPLOYMENT_ID
+    cleanup_oss_semgrep_sast_findings(neo4j_session, common_job_parameters)
```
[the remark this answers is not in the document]
shyammukund: OSS Semgrep path is intentionally modeled as a scoped snapshot sync under DEPLOYMENT_ID="oss". An empty current report set should converge that scoped graph slice to empty; otherwise stale OSS findings persist indefinitely.

tests/integration/cartography/intel/semgrep/test_findings.py
```diff
@@ -439,3 +445,131 @@ def test_sync_findings(
             "INFO",
         ),
     }
+
+
+@patch.object(
+    cartography.intel.semgrep.ossfindings,
+    "get_semgrep_oss_reports",
```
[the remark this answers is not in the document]
shyammukund: addressed by calling the sync the way it is called in orchestration

cartography/cli.py
```diff
@@ -1179,6 +1179,15 @@ def run(
                     hidden=PANEL_SEMGREP not in visible_panels,
                 ),
             ] = None,
+            semgrep_oss_source: Annotated[
```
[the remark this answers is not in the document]
shyammukund: This comment doesn't align with how scanner modules handle this currently in cartography. I believe my pattern is more consistent.

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,228 @@
...
+    branch = repository.get("branch") if isinstance(repository, dict) else None
+
+    transformed: list[dict[str, Any]] = []
+
+    for result in raw_results:
+        if not _is_oss_sast_result(result):
+            continue
+
+        row = dict(result)
+        row["id"] = _build_oss_sast_finding_id(result)
```
kunaals: i think the synthetic id needs to include repository identity before this is safe for bucket/directory ingestion. right now the hash is only check_id + path + start/end location, so two repos with the same finding in the same relative file path will merge into one SemgrepSASTFinding. can we include repositoryUrl, preferably, in the hash and add a regression test with two reports from different repos producing two distinct nodes?

cartography/models/semgrep/ossfindings.py
```diff
@@ -0,0 +1,91 @@
...
+    severity: PropertyRef = PropertyRef("extra.severity")
+    confidence: PropertyRef = PropertyRef("extra.metadata.confidence")
+    file_path: PropertyRef = PropertyRef("path", extra_index=True)
+    start_line: PropertyRef = PropertyRef("start.line")
+    start_col: PropertyRef = PropertyRef("start.col")
+    end_line: PropertyRef = PropertyRef("end.line")
+    end_col: PropertyRef = PropertyRef("end.col")
+    cwe_names: PropertyRef = PropertyRef("extra.metadata.cwe")
+    owasp_names: PropertyRef = PropertyRef("extra.metadata.owasp")
+    categories: PropertyRef = PropertyRef("extra.metadata.category")
```
kunaals: this should preserve the same shape as the existing Cloud SemgrepSASTFinding.categories property. Cloud findings store categories as a list, but OSS maps extra.metadata.category as a scalar string here, so queries like `"security" IN n.categories` will work for Cloud rows and fail for OSS rows under the same label. can we normalize OSS category to a list during transform and map this property from that normalized field?

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,228 @@
...
+            reader.source_uri,
+        )
+        return []
+
+    reports: list[tuple[ReportRef, dict[str, Any]]] = []
+
+    for ref in refs:
+        try:
+            document = read_json_report(reader, ref)
+        except ObjectStoreError as exc:
```
kunaals: this path skips unreadable/malformed reports but still runs cleanup at the end, which can delete previously ingested OSS findings if the configured source is wrong or every JSON file fails to read/parse. can we track failed_report_count and processed_report_count separately, then only run cleanup when at least one Semgrep OSS report was successfully read? an intentionally empty Semgrep report can still clear stale findings, but a failed/no-report run should not.

kunaals (changes requested):

cartography/intel/semgrep/ossfindings.py
shyammukund: discussed offline: repo info will be required to be provided as a YAML file that lives in the same bucket as the data. See schema.md for semgrep for details

cartography/intel/semgrep/ossfindings.py
shyammukund: Addressed by having get_semgrep_oss_reports() return both the successfully parsed report list and a processed_reports signal, then having cleanup be conditional on that signal in sync_oss_semgrep_sast_findings(). That means an intentionally empty but valid OSS report can still converge findings to empty, while unreadable/malformed/no-report runs skip cleanup and preserve previously ingested OSS findings.

cartography/intel/semgrep/ossfindings.py
shyammukund: regression test added with two reports from different repos producing two distinct nodes

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,312 @@
...
+            end_line,
+            end_col,
+            repository_url,
+        ],
+    )
+    digest = hashlib.sha256(raw_id.encode("utf-8")).hexdigest()
+    return f"semgrep-oss-sast-{digest}"
+
+
+def _get_semgrep_oss_repo_context(reader: ReportReader) -> dict[str, str]:
```
kunaals: hmm, i think this is the main design issue. this ties repo metadata to the same source that holds the json reports and caps `--semgrep-oss-source` at one repo context per invocation. could we make `--semgrep-oss-source` point at a yaml manifest instead, with entries like `{provider, owner, repo, url, branch, findings}` and then resolve each `findings` value through `parse_report_source()` / `build_report_reader_for_source()`? that lets one manifest cover many repos and lets the metadata live separately from findings in local/s3/gs/azblob. this is the design bucket that gets hard to undo once documented: https://google.github.io/eng-practices/review/reviewer/looking-for.html

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,312 @@
...
+            logger.warning("Skipping unreadable Semgrep report %s: %s", ref.uri, exc)
+            continue
+
+        if not _looks_like_semgrep_oss_report(document):
+            logger.debug("Skipping %s: not a Semgrep OSS JSON report", ref.uri)
+            continue
+
+        reports.append((ref, document))
+
+    return reports, bool(reports)
```
kunaals: nit: the second tuple element is just `bool(reports)`. i'd drop it and let the call site check `if reports:` unless this starts tracking something stronger like `had_failures` / cleanup safety.

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,312 @@
...
+        all_findings.extend(transform_oss_semgrep_sast_report(document, repo_context))
+
+    if all_findings:
+        logger.info(
+            "Transformed %d total OSS Semgrep SAST findings.", len(all_findings)
+        )
+        load_oss_semgrep_sast_findings(neo4j_session, all_findings, update_tag)
+
+    if processed_reports:
+        common_job_parameters["DEPLOYMENT_ID"] = OSS_DEPLOYMENT_ID
```
kunaals: this still feels unsafe on partial failures. if report a parses and report b fails to read, cleanup can delete b’s previous findings even though we didn’t successfully observe b’s current snapshot. i think cleanup should only run for scopes that were fully read, or explicit malformed report entries should fail loudly rather than being treated like absence

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,312 @@
...
+
+    # Empty results is still a valid Semgrep OSS report.
+    if not results:
+        return True
+
+    first = results[0]
+    if not isinstance(first, dict):
+        return False
+
+    required_keys = {"check_id", "path", "start", "end", "extra"}
```
kunaals: same required key set appears here and in `_is_oss_sast_result()`. if we keep the shape-sniffing path, can we share a module constant? if we move to explicit manifest entries, this file-level sniffer probably goes away and referenced files should be validated directly.

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,312 @@
...
+    end_line: str,
+    end_col: str,
+    repository_url: str,
+) -> str:
+    """
+    Build a stable synthetic ID for OSS findings since Semgrep OSS CLI output
+    does not include the Semgrep Cloud finding ID. Include repository URL
+    so identical findings in different repositories do not collide.
+    """
+    raw_id = "|".join(
```
kunaals: minor correctness/churn point: this id rotates on whitespace-only edits that move line/col. semgrep includes `extra.fingerprint` in the fixture, which should be more stable when present. could prefer that in the hash and fall back to the location hash; at minimum this churn is worth documenting.

cartography/intel/semgrep/__init__.py
```diff
@@ -20,36 +25,50 @@ def start_semgrep_ingestion(
...
+        source = parse_report_source(config.semgrep_oss_source)
+        with build_report_reader_for_source(
+            source,
+            config=config,
+        ) as reader:
+            sync_oss_semgrep_sast_findings(
+                neo4j_session,
+                reader,
+                config.update_tag,
+                common_job_parameters,
```
kunaals: both cloud and oss paths mutate `common_job_parameters["DEPLOYMENT_ID"]`. works in the current order, but a per-sync copy would make the contract clearer and avoid future ordering surprises. related shape: https://www.hyrumslaw.com/

cartography/models/semgrep/ossfindings.py
```diff
@@ -0,0 +1,91 @@
...
+    title: PropertyRef = PropertyRef("check_id")
+
+
+@dataclass(frozen=True)
+class OSSSemgrepSASTFindingToSemgrepDeploymentRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:SemgrepSASTFinding)<-[:RESOURCE]-(:SemgrepDeployment)
```
kunaals, line 45: not blocking: this comment placement is a little odd, but the cloud semgrep model uses the same pattern, so i don't think we need to spend review capital on it unless you're already touching this block.

cartography/models/semgrep/ossfindings.py
```diff
@@ -0,0 +1,91 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class OSSSemgrepSASTFindingNodeProperties(CartographyNodeProperties):
```
kunaals: not blocking: this mostly mirrors `SemgrepSASTFindingNodeProperties`. a shared base could reduce drift eventually, but i don't think this PR needs to take that refactor unless a third source appears.

cartography/models/semgrep/ossfindings.py
```diff
@@ -0,0 +1,91 @@
...
+class OSSSemgrepSASTFindingToGithubRepoRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+# (:SemgrepSASTFinding)-[:FOUND_IN]->(:GitHubRepository)
+class OSSSemgrepSASTFindingToGithubRepoRel(CartographyRelSchema):
+    target_node_label: str = "GitHubRepository"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("repositoryUrl")},
```
kunaals, line 69: tiny/non-blocking: this matcher works because `GitHubRepository.id` is the repository url. a one-line comment could save some head scratching, but this is too small to block on.

tests/integration/cartography/intel/semgrep/test_findings.py
```diff
@@ -439,3 +449,256 @@ def test_sync_findings(
...
+
+def test_sync_oss_sast_findings(neo4j_session, tmp_path):
+    # Arrange
+    neo4j_session.run("MATCH (n) DETACH DELETE n")
+    create_github_repos(neo4j_session)
+    common_job_parameters = {
+        "UPDATE_TAG": TEST_UPDATE_TAG,
+    }
+    fixture_path = (
+        Path(__file__).resolve().parents[4]
```
kunaals: tiny: `parents[4]` is brittle if this file moves. most cartography integration tests pull fixtures through `tests.data.*` helpers or use direct repo-relative paths; either would be easier to maintain.

tests/integration/cartography/intel/semgrep/test_findings.py
```diff
@@ -439,3 +449,256 @@ def test_sync_findings(
...
+    )
+
+    source_root_b = tmp_path / "github" / "different-org" / "different-repo" / "main"
+    source_root_b.mkdir(parents=True, exist_ok=True)
+    (source_root_b / "oss_sast_report.json").write_text(
+        fixture_path.read_text(),
+        encoding="utf-8",
+    )
+    (source_root_b / "repo_metadata.yaml").write_text(
+        "\n".join(
```
kunaals: tiny: this inline yaml string can drift from `repo_metadata.yaml`. `yaml.safe_dump({...})` would keep the shape explicit without copy-pasting the fixture format.

tests/integration/cartography/intel/semgrep/test_findings.py
```diff
@@ -439,3 +449,256 @@ def test_sync_findings(
...
+    ids_for_repo_b = check_nodes(
+        neo4j_session,
+        "SemgrepSASTFinding",
+        ["id"],
+    )
+
+    # Assert
+    assert ids_for_repo_a
+    assert ids_for_repo_b
+    assert ids_for_repo_a != ids_for_repo_b
```
kunaals: this test only proves the ids differ after the second sync; because the second sync uses a newer update tag under the same `DEPLOYMENT_ID`, it can clean repo a's findings and still pass. can this assert both repo a and repo b coexist after sequential syncs, or move cleanup scope per repo?

tests/unit/cartography/intel/semgrep/test_ossfindings.py
```diff
@@ -0,0 +1,58 @@
...
+Unit tests for Semgrep OSS metadata parsing.
+"""
+
+from cartography.intel.common.object_store import ListedReportReader
+from cartography.intel.common.object_store import ReportRef
+from cartography.intel.semgrep.ossfindings import _build_oss_sast_finding_id
+from cartography.intel.semgrep.ossfindings import _get_semgrep_oss_repo_context
+
+
+def test_get_semgrep_oss_repo_context_happy_path():
```
kunaals: happy path is covered, but the new failure branches could use tests too: malformed metadata/missing fields, duplicate yaml files, json present but not semgrep-shaped, and zero json files. these are the places where the cleanup contract and operator feedback matter: https://abseil.io/resources/swe-book/html/ch11.html

kunaals (changes requested):

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,502 @@
...
+
+        if repo_findings:
+            logger.info(
+                "Transformed %d OSS Semgrep SAST findings for repository %s.",
+                len(repo_findings),
+                repository_mapping.repository_name,
+            )
+            load_oss_semgrep_sast_findings(neo4j_session, repo_findings, update_tag)
+
+    if cleanup_repository_urls:
```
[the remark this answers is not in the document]
shyammukund: This was a design decision made intentionally.

docs/root/modules/semgrep/schema.md
```diff
@@ -52,37 +52,65 @@ Represents a Semgrep [Deployment](https://semgrep.dev/api/v1/docs/#tag/Deploymen
...
+Represents a [Semgrep SAST](https://semgrep.dev/docs/semgrep-code/getting-started/) finding. This is a code-level security issue discovered either from the Semgrep Cloud Platform or from OSS Semgrep CLI JSON reports ingested via `--semgrep-oss-source`. Cloud findings come from the Semgrep Findings API. OSS findings come from explicit Semgrep OSS JSON report artifacts described in a repository mapping file passed to `--semgrep-oss-source`.
+
+> **Note**: The OSS Semgrep JSON report does not include repository identity. Cartography therefore requires `--semgrep-oss-source` to point to a repository mapping YAML file that declares the repository metadata and the explicit report artifact(s) associated with that repository. The mapping file must be valid UTF-8 YAML, must contain a top-level `repositories` list, and each repository entry must include `provider`, `owner`, `repo`, `url`, `branch`, and a non-empty `reports` list.
```
kunaals: can we add the oss setup to `docs/root/modules/semgrep/config.md` too? this schema doc explains the graph shape well, but config.md is where users will look for "how do i configure semgrep ingestion." i think it should mention `--semgrep-oss-source`, say it points to a repository mapping yaml file, and include a minimal mapping + command example.

docs/root/modules/semgrep/schema.md
```diff
@@ -52,37 +52,65 @@ Represents a Semgrep [Deployment](https://semgrep.dev/api/v1/docs/#tag/Deploymen
...
+>     owner: "different-org"
+>     repo: "different-repo"
+>     url: "https://github.com/different-org/different-repo"
+>     branch: "main"
+>     reports:
+>       - "gs://security-artifacts/semgrep/different-repo/job-1.json"
+>       - "gs://security-artifacts/semgrep/different-repo/job-2.json"
+> ```
+>
+> Each `reports` entry must point to exactly one Semgrep OSS JSON artifact for the repository it is nested under. Cartography treats one repository entry as the intended snapshot for that repository in the current run. If all listed reports for a repository are successfully processed, Cartography runs cleanup for stale OSS findings scoped to that repository URL. If any listed report for that repository fails to resolve, fails to parse, or is not Semgrep-shaped, Cartography skips cleanup for that repository to avoid deleting findings from an incomplete snapshot.
```
kunaals: this is mostly clear, but i'd make the single-artifact behavior explicit for monorepo/sharded scans. something like: "for sharded or monorepo scans, list each generated json artifact separately; a `reports` entry must not be a directory or object-store prefix containing multiple json files." that makes it even more clear for that use case.

cartography/intel/semgrep/ossfindings.py
```diff
@@ -0,0 +1,502 @@
...
+
+logger = logging.getLogger(__name__)
+
+OSS_DEPLOYMENT_ID = "oss"
+_SEMGREP_OSS_RESULT_REQUIRED_KEYS = {"check_id", "path", "start", "end", "extra"}
+_SEMGREP_OSS_UNUSABLE_FINGERPRINTS = {"requires login"}
+
+
+class SemgrepOSSRepositoryMappingEntry(BaseModel):
+    provider: str
```
kunaals, line 42: since `provider` is part of the mapping now, can we create a follow-up github issue for adding gitlab support? the current graph matching path is github-only via `GitHubRepository`, which is fine for this PR, but tracking gitlab explicitly would make the provider field's future direction clear.

tests/unit/cartography/intel/semgrep/test_ossfindings.py
```diff
@@ -0,0 +1,127 @@
...
+    assert report_collection.successful_sources == 1
+    assert report_collection.failed_sources == 0
+    assert report_collection.all_sources_succeeded is True
+    assert report_collection.any_reports_processed is True
+    assert len(report_collection.reports) == 1
+    assert report_collection.reports[0][0].name == "oss_sast_report.json"
+    assert isinstance(report_collection.reports[0][1]["results"], list)
+
+
+def test_get_semgrep_oss_repository_mappings_rejects_multiple_yaml_files():
```
kunaals, line 62: nice improvement on the mapping tests. i think we still want a few negative cases around the contract that now matters: missing required mapping fields, empty `reports`, non-semgrep json, and zero artifacts. those are the operator-error paths this feature is most likely to hit.

tests/integration/cartography/intel/semgrep/test_findings.py
```diff
@@ -439,3 +444,146 @@ def test_sync_findings(
...
+            "SemgrepSASTFinding",
+            "rule_id",
+            "FOUND_IN",
+            rel_direction_right=False,
+        )
+        == expected_found_in_rels
+    )
+
+
+def test_sync_oss_sast_findings_multi_entry_multi_report(neo4j_session):
```
kunaals, line 511: can we add one cleanup-focused integration case? seed stale oss findings for two repos, run a mapping where repo a fully succeeds and repo b has one bad/missing report, then assert repo a cleans up and repo b's stale findings are preserved. that would lock in the partial-failure behavior described in the docs.

kunaals (changes requested):

docs/root/modules/semgrep/schema.md
kunaals:
```suggestion
> **Note**: The OSS Semgrep JSON report does not include repository identity. Cartography therefore requires `--semgrep-oss-source` to point to a repository mapping YAML file. See more in the [Semgrep config](./config.md#oss-semgrep-configuration)
```
btw `./config.md#oss-semgrep-configuration` is a same directory link with a fragment so it will take them straight to setup

The schema should be intentionally light about setup related stuff

docs/root/modules/semgrep/schema.md
```diff
@@ -52,37 +52,65 @@ Represents a Semgrep [Deployment](https://semgrep.dev/api/v1/docs/#tag/Deploymen
...
+>   - provider: "github"
+>     owner: "different-org"
+>     repo: "different-repo"
+>     url: "https://github.com/different-org/different-repo"
+>     branch: "main"
+>     reports:
+>       - "gs://security-artifacts/semgrep/different-repo/job-1.json"
+>       - "gs://security-artifacts/semgrep/different-repo/job-2.json"
+> ```
+>
```
kunaals:
```suggestion
```
lets remove this, already in config.md

docs/root/modules/semgrep/schema.md
```diff
@@ -52,37 +52,65 @@ Represents a Semgrep [Deployment](https://semgrep.dev/api/v1/docs/#tag/Deploymen
...
+> Each `reports` entry must point to exactly one Semgrep OSS JSON artifact for the repository it is nested under. For sharded or monorepo scans, list each generated JSON artifact separately under `reports`; a `reports` entry must not be a directory or object-store prefix containing multiple JSON files. Cartography treats one repository entry as the intended snapshot for that repository in the current run. If all listed reports for a repository are successfully processed, Cartography runs cleanup for stale OSS findings scoped to that repository URL. If any listed report for that repository fails to resolve, fails to parse, or is not Semgrep-shaped, Cartography skips cleanup for that repository to avoid deleting findings from an incomplete snapshot.
```
kunaals: lets move this to config.md

kunaals (approved):
approving to unblock, some nits about the docs that would be good to change

PR #2722, opened 2026-04-30
refactor(reports): Refactor Report Reader to not need azure specific args

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

This PR simplifies unified report-source reading for scanner modules by removing Azure-specific auth argument plumbing from the scanner entrypoints.

Scanner modules now:

- parse the report source
- pass the source plus Config into build_report_reader_for_source()

This keeps scanner modules from knowing about Azure-specific auth fields individually while preserving compatibility with existing Cartography Azure service-principal config for azblob:// report sources.

The new shared helper is:
-build_azure_blob_credential_from_config(config)

Behavior after this change:

- if Azure service-principal auth is configured in Cartography, the helper builds ClientSecretCredential(...)
- if Azure service-principal auth is not configured, the helper builds AzureCliCredential()
- if Azure service-principal fields are provided while azure_sp_auth is disabled, the helper logs the same warning as before and still uses AzureCliCredential()
- AzureBlobContainerReader then preserves the previous fallback behavior and creates AzureCliCredential() when no credential is supplied

**Tests**
tests/unit/cartography/intel/common/test_report_source.py
- test_build_azure_blob_credential_from_config_returns_sp_credential
    -verifies that Azure SP config builds ClientSecretCredential(...)
- test_build_azure_blob_credential_from_config_warns_and_returns_cli_credential_when_disabled_with_sp_fields
     -verifies that disabled Azure SP auth with tenant/client/secret set logs the compatibility warning and returns AzureCliCredential()
- test_build_azure_blob_credential_from_config_requires_all_sp_fields
      -  verifies that missing Azure SP fields raise the expected ValueError
- test_build_report_reader_for_azure_uses_reader_defaults
      - verifies that an AzureBlobReportSource uses AzureBlobContainerReader with the AzureCliCredential built from the disabled-SP config path
- test_build_report_reader_for_azure_uses_config_credential
       - verifies that Azure SP config is translated into a ClientSecretCredential and passed through to AzureBlobContainerReader
 
tests/unit/cartography/intel/common/test_object_store.py

- test_azure_blob_reader_creates_cli_credential_when_none_supplied
      - verifies that AzureBlobContainerReader(..., credential=None) creates AzureCliCredential()
- test_azure_blob_reader_uses_supplied_credential
       - verifies that AzureBlobContainerReader(..., credential=fake_credential) uses the supplied credential instead of creating a default one
 
- Removed Tests     
   
   The old builder-level Azure auth tests were removed because build_report_reader_for_source() no longer accepts raw Azure auth kwargs directly such as azure_sp_auth, azure_tenant_id, azure_client_id, and azure_client_secret.

Instead, Azure auth from Config is translated by build_azure_blob_credential_from_config(config) inside the Azure branch of build_report_reader_for_source().

### Related issues or links

- Builds on functionality introduced in https://github.com/cartography-cncf/cartography/pull/2620

### How was this tested?

 - Unit Tests updated
 - `uv run pytest tests/unit/cartography/intel/common/test_report_source.py tests/unit/cartography/intel/common/test_object_store.py`

#### General

#### Proof of functionality


cartography/intel/trivy/__init__.py
```diff
@@ -351,15 +355,17 @@ def start_trivy_ingestion(neo4j_session: Session, config: Config) -> None:
         return
 
     source = parse_report_source(config.trivy_source)
+    azure_blob_credential = (
```
kunaals: i think this still leaves the scanner module coupled to azure storage auth. the azure-specific branch moved out of the function args, but now each scanner has to import `AzureBlobReportSource` and know when to build azure credentials.

could we move this into a common config-aware reader builder instead, so callers only do `source -> reader`? that keeps trivy/syft/aibom/docker scout ignorant of provider auth, and future storage auth changes stay in one place.

cartography/intel/common/object_store.py
```diff
@@ -271,7 +273,9 @@ def __init__(
         self._account_name = account_name
         self._container_name = container_name
         self._prefix = prefix
-        from azure.storage import blob as azure_blob
+
+        if credential is None:
+            credential = DefaultAzureCredential()
```
kunaals: i think we should preserve the existing fallback here and keep this PR as a pure interface refactor. the PR is marked as no functional changes, but switching the non-SP path from `AzureCliCredential` to `DefaultAzureCredential` changes auth resolution order: env/workload/managed identity can be selected before azure cli.

could we keep the previous behavior for now (`azure_sp_auth=True` -> `ClientSecretCredential`, otherwise `AzureCliCredential`) and just move that decision behind the common report-reader builder? if we want `DefaultAzureCredential`, i think that should be a separate explicit behavior change with docs/tests around identity precedence.

kunaals (changes requested):
requesting changes on the caller boundary and auth compatibility. the refactor direction is good, but this should stay a pure interface cleanup and preserve the existing azure credential behavior.

cartography/intel/trivy/__init__.py
kunaals: like i think this makes azure the odd one out in the report-reader abstraction. s3 and gcs auth/session setup stays inside the common reader-building layer, but azure now requires each scanner entrypoint to import `AzureBlobReportSource`, branch on it, and build a credential before asking for a reader.

benefit in the future is that auth/storage changes stay in one place instead of being repeated across every scanner module

shyammukund: removed the azure credential build from ingestion and to the reader-building layer.

cartography/intel/common/object_store.py
shyammukund: restoring AzureCliCredential in the helper and in the reader default path.

cartography/intel/common/object_store.py
```diff
@@ -271,7 +273,9 @@ def __init__(
         self._account_name = account_name
         self._container_name = container_name
         self._prefix = prefix
-        from azure.storage import blob as azure_blob
+
+        if credential is None:
+            credential = AzureCliCredential()
```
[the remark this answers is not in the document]
shyammukund: This contradicts @kunaals comment above

cartography/intel/common/report_reader_builder.py
```diff
@@ -9,45 +9,37 @@
...
-    azure_sp_auth: bool | None,
-    azure_tenant_id: str | None,
-    azure_client_id: str | None,
-    azure_client_secret: str | None,
-) -> Any:
-    # Blob data-plane access; we deliberately avoid the Azure resource-graph
-    # Authenticator here because it requires subscription-list permission.
-    from azure import identity as azure_identity
+def build_azure_blob_credential_from_config(config: "Config") -> Any | None:
+    if not config.azure_sp_auth:
```
kunaals, line 21: could we preserve the previous warning here when service-principal fields are provided but `azure_sp_auth` is disabled? before this refactor, that path still used cli credentials, but logged that the SP fields were being ignored.

the behavior is now back to cli credentials, which is good, but keeping the warning preserves the old compatibility signal and makes misconfiguration easier to diagnose. i'd add a small test for this too: `azure_sp_auth=False` with tenant/client/secret set should warn and still return/use `AzureCliCredential`.

kunaals (dismissed):
this is much closer now. the scanner call sites look clean to me: they just pass `source` and `config` into the common reader builder, and azure is no longer the odd one out in the scanner modules.

one remaining thing: the PR description looks stale after the latest changes. it still says the reader falls back to `DefaultAzureCredential()` and that disabled azure SP auth returns `None`, but the code now uses `AzureCliCredential()` for the non-SP path. can we update the description/testing section so it matches the current behavior?

cartography/intel/common/report_reader_builder.py
shyammukund: Added back the exact same warning, just in build_azure_blob_credential_from_config() instead of _build_azure_credential() since that no longer exists 

the unit test now is updated to check the "should warn and still return AzureCliCredential". I think that the existing Azure builder test (test_build_report_reader_for_azure_uses_reader_defaults) covers the scenario that if  azure_sp_auth=False the Azure reader gets the AzureCliCredential result so i did not include that in this test.

kunaals (approved):

cartography/intel/common/object_store.py
```diff
@@ -7,6 +7,8 @@
 from typing import Iterable
 from typing import Protocol
 
+from azure.identity import AzureCliCredential
```
kunaals, line 10: could we keep these azure sdk imports lazy inside `AzureBlobContainerReader.__init__()`? `object_store.py` is shared by local/s3/gcs report readers too, so importing azure at module load makes non-azure paths pay that dependency/import cost.

azure is currently a normal dependency, so this is not breaking today, but keeping provider sdk imports local preserves the option to split cloud providers into extras later and matches the existing provider-boundary pattern.

kunaals (approved):
approved, though i just saw one more nit, would recommend fixing this and then good to merge

PR #2726, opened 2026-04-30
feat(gcp): GCP attack-path parity — IAM permission edges, SA keys, public-exposure rules, GKE Workload Identity

jychp:
### Type of change
- [ ] Documentation update

### Summary

Brings GCP coverage closer to AWS parity for attack-path / findings work. Each commit is self-contained and reviewable on its own:

- `feat(gcp): extend IAM permission_relationships to BigQuery, KMS, and Artifact Registry` — adds `(:GCPPrincipal)-[:CAN_*]->(target)` edges for BigQuery datasets / tables, KMS crypto keys (new `CAN_DECRYPT` / `CAN_ENCRYPT` capacities), and Artifact Registry repositories. YAML-only, no engine changes.
- `feat(gcp): add CAN_IMPERSONATE edge from GCPPrincipal to GCPServiceAccount` — models the GCP equivalent of `sts:AssumeRole`, the primary privilege escalation path on GCP.
- `feat(gcp): ingest user-managed service account keys as GCPServiceAccountKey` — new node + intel + cleanup wiring for SA keys (USER_MANAGED only, system-managed keys are intentionally skipped).
- `feat(gcp,rules): add GCP Facts for object_storage_public, compute_instance_exposed, database_instance_exposed` — each rule previously had AWS-only (or AWS+Azure) coverage despite a cross-cloud name. Adds GCP Facts plus the schema enrichments they need (`iam_config_public_access_prevention` on `GCPBucket`, `is_public` on `GCPPolicyBinding`, new `GCPCloudSQLAuthorizedNetwork`).
- `feat(kubernetes,gcp): wire GKE Workload Identity to GCPServiceAccount` — mirrors the existing AWS IRSA wiring on `KubernetesServiceAccount` so a Pod's GCP impersonation chain is queryable.

Internal review surfaced four issues that are addressed in the trailing four commits:

- `fix(gcp): skip GCP service account key cleanup on partial sync failures` — a transient quota/permission failure on a single SA used to silently delete the rest of the project's keys. The orchestrator now skips key cleanup when any SA's keys couldn't be re-ingested.
- `fix(gcp): use full resource path in permission_relationships scope keys` — the engine reduced scopes to the last `/`-separated segment, causing CAN_* over-attribution between sibling resources sharing a leaf name (two BigQuery tables named `events` in different datasets, two KMS crypto keys named `default` in different keyrings, etc).
- `fix(rules): tighten GCP compute_instance_exposed to firewall + management ports` — the GCP fact previously matched any instance with a public IP, drifting outside the rule's "Common Management Ports" semantics. Now requires an enabled INGRESS firewall on the VPC allowing 0.0.0.0/0 to one of `[22, 3389, 3306, 5432, 6379, 9200, 27017]`.
- `fix(rules): exclude conditional bindings from gcp_bucket_public fact` — a binding bound to `allUsers` with an IAM Condition (e.g. expired temporal scope) shouldn't be flagged as unconditionally public. Fact now requires `has_condition = false`.

A trailing `docs(gcp,kubernetes): document new nodes, properties, and edges` commit updates the GCP and Kubernetes schema docs.

### Related issues or links

N/A.

### How was this tested?

```
make test_lint    # pre-commit clean across all files
uv run --frozen pytest -q tests/unit/ tests/integration/cartography/intel/gcp/ tests/integration/cartography/intel/kubernetes/
# 1823 passed
```

Per-feature coverage:

- New IAM permission edges: extended `tests/integration/cartography/intel/gcp/test_permission_relationships.py` with a project-level role bundle on alice and assertions on each new target label, plus a sibling-`events` regression test for resource scope uniqueness.
- `CAN_IMPERSONATE`: same test file, bob bound to `roles/iam.serviceAccountTokenCreator` at project level, edge to the project's SA asserted.
- `GCPServiceAccountKey`: new `test_sync_gcp_iam_service_account_keys` covering USER_MANAGED-only filtering, `(GCPProject)-[:RESOURCE]->` and `(GCPServiceAccount)-[:HAS_KEY]->` rels. Partial-failure case covered by `test_sync_gcp_iam_service_account_keys_partial_failure`.
- Bucket `iam_config_public_access_prevention` and CloudSQL `GCPCloudSQLAuthorizedNetwork`: extended `test_sync_gcp_buckets` and `test_sync_sql`.
- `is_public` on `GCPPolicyBinding`: extended `test_sync_gcp_policy_bindings`.
- GKE Workload Identity: extended `test_sync_rbac_end_to_end` with an annotated SA and a pre-seeded `GCPServiceAccount` target.

#### General

#### Proof of functionality

### Notes for reviewers

- The PR is structured as 11 small commits on purpose: each can be reviewed and reverted independently. The scope-uniqueness fix and the rule-tightening fixes are deliberately separate from the feature commits so it stays clear what the engineering review surfaced; the trailing docs commit is split out the same way.
- **Production sync logs**: I've validated the new code paths via the integration test suite only; I don't have a real GCP project handy with all of BigQuery / KMS / Artifact Registry / Cloud SQL / GKE Workload Identity in one place to capture end-to-end logs. Let me know if that's a hard merge requirement.
- The `compute_instance_exposed` GCP fact is intentionally widened by VPC scope (no `target_tags` / target SA filtering yet) and the description calls that limitation out. Tightening that further would mean modeling the full firewall-to-instance applicability graph and is better as a separate change.
- Two existing files needed assertion updates because mock data was extended to exercise the new behaviour: `test_policy_bindings.py` (new bindings) and `test_rbac.py` (new annotated SA). Those are not behavioural changes to those modules.


kunaals (approved):
LGTM

PR #2727, opened 2026-05-01
feat(ecr): Infer ECR provenance from CircleCI labels

kunaals:
### Type of change

- [ ] Documentation update

### Summary

This adds a narrow ECR image provenance fallback for images that do not publish SLSA/in-toto attestation provenance but do include source metadata in image config labels.

The fallback reads labels from the same ECR image config blob already used for layer and history extraction, recognizes CircleCI label names by final segment so custom namespaces work, and maps the values into the existing `ECRImage.source_uri`, `source_revision`, and `source_file` fields. GitHub SSH repository URLs are normalized to HTTPS URLs so the existing GitHub supply-chain matcher can create `PACKAGED_FROM` relationships without a schema change.

Attestation provenance remains preferred. Label-derived values only fill fields that were not already populated from stronger provenance data. For manifest lists, the fallback is applied to child/platform image digests rather than the manifest-list node.

### Related issues or links

N/A

### Breaking changes

None.

### How was this tested?

- `python3 -m compileall cartography/intel/aws/ecr_image_layers.py`
- `/opt/homebrew/bin/uv run pre-commit run isort --files cartography/intel/aws/ecr_image_layers.py tests/unit/cartography/intel/aws/test_ecr_image_layers.py tests/integration/cartography/intel/aws/test_ecr_layers.py`
- `/opt/homebrew/bin/uv run pre-commit run black --files cartography/intel/aws/ecr_image_layers.py tests/unit/cartography/intel/aws/test_ecr_image_layers.py tests/integration/cartography/intel/aws/test_ecr_layers.py`
- `/opt/homebrew/bin/uv run pre-commit run flake8 --files cartography/intel/aws/ecr_image_layers.py tests/unit/cartography/intel/aws/test_ecr_image_layers.py tests/integration/cartography/intel/aws/test_ecr_layers.py`
- `/opt/homebrew/bin/uv run pytest tests/unit/cartography/intel/aws/test_ecr_image_layers.py -q`
- `/opt/homebrew/bin/uv run pytest tests/integration/cartography/intel/aws/test_ecr_layers.py -q`
- `/opt/homebrew/bin/uv run pytest tests/integration/cartography/intel/aws/test_ecr_layers.py tests/unit/cartography/intel/aws/test_ecr_image_layers.py -q`

Manual local Cartography-as-library validation against AWS ECR image metadata also confirmed that label-derived `source_uri`, `source_revision`, and `source_file` values can populate local `ECRImage` nodes and produce `PACKAGED_FROM` relationships to matching local `GitHubRepository` nodes.

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

No public schema fields or relationship labels are added. This reuses the existing image provenance fields and existing GitHub supply-chain matching path.


cartography/intel/aws/ecr_image_layers.py
```diff
@@ -94,6 +95,82 @@ class ECRLayerFetchTransientError(Exception):
...
+    expected = label_name.lower()
+    for key, value in labels.items():
+        final_segment = str(key).rsplit(".", 1)[-1].lower()
+        if final_segment != expected:
+            continue
+        if value is None:
+            return None
+        value_str = str(value).strip()
+        return value_str or None
+    return None
```
jychp: **[blocking]** `_get_label_value` stops at the first matching key and silently ignores any subsequent matching keys. If an image carries two labels with the same final segment under different namespaces (e.g. `com.foo.CIRCLE_REPOSITORY_URL` and `com.bar.CIRCLE_REPOSITORY_URL`), the winner is determined by iteration order, not intent. Detect and log a warning when multiple keys match the same target label, and return `None` in that case.

cartography/intel/aws/ecr_image_layers.py
```diff
@@ -614,8 +694,11 @@ def transform_ecr_image_layers(
                 # Source file (Dockerfile path) from configSource
                 if provenance.get("source_file"):
                     membership["source_file"] = provenance["source_file"]
-                membership["from_attestation"] = True
-                membership["confidence"] = "explicit"
+                if provenance.get("parent_image_uri") or provenance.get(
+                    "parent_image_digest"
+                ):
+                    membership["from_attestation"] = True
```
jychp, line 717: **[blocking]** `from_attestation=True` and `confidence="explicit"` are now gated on `parent_image_uri / parent_image_digest` being present, but that is a regression from master where any provenance data triggered those fields. An image enriched only with `source_uri` + `source_revision` from a SLSA attestation that does not carry parent image data will now leave `from_attestation` and `confidence` unset. These flags must be set whenever provenance comes from attestation, regardless of whether parent image data is present.

cartography/intel/aws/ecr_image_layers.py
```diff
@@ -94,6 +95,82 @@ class ECRLayerFetchTransientError(Exception):
     httpx.WriteTimeout,
 )
 MAX_BLOB_DOWNLOAD_ATTEMPTS = 3
+CIRCLECI_LABEL_SOURCE_URI = "CIRCLE_REPOSITORY_URL"
+CIRCLECI_LABEL_SOURCE_REVISION = "CIRCLE_SHA1"
+CIRCLECI_LABEL_SOURCE_FILE = "DOCKERFILE"
```
jychp, line 100: **[nitpick]** `CIRCLECI_LABEL_SOURCE_FILE = "DOCKERFILE"` matches any label with the final segment `dockerfile` regardless of CI system. Gate extraction of this field on the presence of at least one unambiguously CircleCI label (e.g. `CIRCLE_SHA1`) before treating `DOCKERFILE` as a CircleCI provenance signal.

cartography/intel/aws/ecr_image_layers.py
```diff
@@ -94,6 +95,82 @@ class ECRLayerFetchTransientError(Exception):
...
+def _normalize_git_repository_url(url: str) -> str:
+    url = url.strip()
+
+    github_ssh_match = re.fullmatch(r"git@github\.com:(?P<path>.+)", url)
+    if github_ssh_match:
+        url = f"https://github.com/{github_ssh_match.group('path')}"
+    else:
+        github_ssh_url_match = re.fullmatch(r"ssh://git@github\.com/(?P<path>.+)", url)
+        if github_ssh_url_match:
+            url = f"https://github.com/{github_ssh_url_match.group('path')}"
```
jychp, line 145: **[suggestion]** `_normalize_git_repository_url` strips `.git` after SSH-to-HTTPS rewrites, but plain HTTPS clone URLs (e.g. `https://github.com/org/repo.git`) pass through both SSH regex branches unchanged and still carry `.git`. The `.endswith(".git")` strip already handles this, but the function does not strip trailing `.git` from HTTPS URLs that do not match either regex. Verify that the strip on line 134 executes for all URL forms, including plain HTTPS.

tests/data/github/repos.py
```diff
@@ -190,6 +190,36 @@
     },
 ]
 
+GET_REPOS_CIRCLECI_PROVENANCE: List[dict[str, Any]] = [
```
jychp: **[nitpick]** `GET_REPOS_CIRCLECI_PROVENANCE: List[dict[str, Any]]` uses `List` from `typing` (imported at line 3). The project style guide requires lowercase `list[dict[str, Any]]` (PEP 585). Change to `list[dict[str, Any]]` and remove the `List` import if it is no longer used elsewhere.

kunaals:
@cubic-dev-ai can you review?

jychp (approved):

cartography/intel/aws/ecr_image_layers.py
kunaals: good catch. fixed this by warning and failing closed when multiple suffix-matched labels are present.

cartography/intel/aws/ecr_image_layers.py
kunaals: good catch. fixed this by tracking attestation provenance separately from parent image fields.

cartography/intel/aws/ecr_image_layers.py
kunaals: good catch. gated dockerfile label extraction on an actual circleci source or revision signal.

cartography/intel/aws/ecr_image_layers.py
kunaals: verified and added unit coverage for plain https .git normalization; the existing strip path handles it.

tests/data/github/repos.py
kunaals: fixed.

tests/integration/cartography/intel/aws/test_ecr_layers.py
```diff
@@ -6,7 +6,10 @@
```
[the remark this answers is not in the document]
kunaals: stale now. the test runs github.repos.sync and supply_chain.sync; only external fetch boundaries are mocked.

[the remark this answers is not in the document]
kunaals: this one is not applicable to the github-link assertion; that multi-arch test does not load a repo, and the link test uses matching normalized data.

PR #2729, opened 2026-05-01
docs: ship intel-module procedures as auto-loaded skills + prefer uv for native install

### LINKED ISSUE #1940: Update install.md native install docs with uv instructions

jychp:
### Type of change
- [x] Documentation update

### Summary

Two related docs changes:

1. **Replace `docs/root/agents/` with auto-loaded skills under `.agents/skills/`.**
   The eight procedure docs (create-module, add-node-type, add-relationship, analysis-jobs, create-rule, enrich-ontology, refactor-legacy, troubleshooting) become Claude skills with YAML frontmatter triggers. Long-form examples and cheat sheets move into per-skill `references/` so the first-load context stays small (Anthropic skill spec progressive-disclosure pattern). Skill-aware coding agents auto-load the right one when a relevant task starts; humans keep getting the same content but indexed via `AGENTS.md` instead of a Sphinx subtree.

   - `AGENTS.md` drops the procedure-link table and now references skills by name only.
   - Sphinx toctrees in `docs/root/index.md` and `docs/root/dev/index.md` drop `agents/index`.
   - `docs/root/agents/` is removed entirely (no duplicate source of truth).

2. **Prefer `uv tool install cartography` for native install (Fixes #1940).**
   Replaces the `pip install cartography` step in Option 3 with a `uv tool install cartography` step, matching the project's existing uv-based dev/CI workflow. Mentions `uv tool upgrade cartography` for updates. `pip install cartography` is kept as a fallback. No content duplication, no new image references.

### Related issues or links

- Fixes #1940

### How was this tested?

- `uv run sphinx-build docs/root /tmp/out -c docs` → "build succeeded, 6 warnings". The 6 are pre-existing (`_static`, missing `logo-vertical.svg`, unsupported theme options); zero new warnings, zero `agents/index` toctree warnings.
- Repo-wide `grep -rn "docs/root/agents\|docs/agents"` returns nothing in tracked files.
- Each `SKILL.md` validated: kebab-case folder name matches `name:` frontmatter, descriptions under 1024 chars and free of `<` / `>`, no `claude-` / `anthropic-` prefix.

#### General
- [ ] I have added/updated tests that prove my fix is effective or my feature works. (N/A — docs only)


AGENTS.md
```diff
@@ -6,26 +6,24 @@ This guide teaches you how to write intel modules for Cartography using the mode
...
 2. [AI Assistant Quick Reference](#ai-assistant-quick-reference) - Key concepts and imports
 3. [Git and Pull Request Guidelines](#git-and-pull-request-guidelines) - Commit signing and PR templates
 4. [Quick Start](#quick-start-copy-an-existing-module) - Copy an existing module
 5. [Quick Reference Cheat Sheet](#quick-reference-cheat-sheet) - Copy-paste templates
 
-## Procedure Documentation
+## Procedure Skills
 
-Detailed procedures are available in separate documents:
+Procedures for building and extending Cartography intel modules ship as Claude skills under `.agents/skills/`. Skill-aware agents auto-load each skill from its YAML frontmatter when a relevant task starts; you do not need to open the files manually. The available skills are:
```
kunaals, line 17: let's add a symlink from .claude/skills to .agents/skills so that Claude can autodiscover these?

.agents/skills/add-node-type/SKILL.md
```diff
@@ -0,0 +1,159 @@
+---
+name: add-node-type
+description: Define a new node schema in `cartography/models/<module>/`, including required properties, sub-resource relationships, extra labels, conditional labels, scoped cleanup, and one-to-many transforms. Use when the user asks to add a node type, model a new resource, configure extra Neo4j labels (Identity, Asset, UserAccount, Tenant), or wire scoped vs global cleanup.
```
kunaals: this description currently fails the skill validator because it contains angle brackets in `cartography/models/<module>/`. can we rewrite the placeholder without xml-style brackets, e.g. `cartography/models/MODULE_NAME/`, so the skill remains valid and discoverable?

.agents/skills/create-module/references/sync-pattern.md
```diff
@@ -0,0 +1,228 @@
+# Sync pattern reference
```
kunaals, line 1: this reference is long enough that it should probably start with a short `## Contents` section. the skill authoring guidance recommends that for reference files over 100 lines so agents can see the available sections even when previewing only the top of the file. same issue applies to the other >100-line reference files in this PR.

kunaals (changes requested):

kunaals (approved):

PR #2735, opened 2026-05-01
refactor: unify workload chain across providers via WORKLOAD_PARENT

jychp:
### Type of change
- [x] Documentation update

### Summary

Introduce a uniform `child -[:WORKLOAD_PARENT]-> parent` relationship across every provider that models a compute workload chain (AWS ECS, Kubernetes, GCP Cloud Run, Azure Container Instances), and tag intermediate nodes with role-distinct ontology labels (`ComputeService`, `ComputeNamespace`, `ComputePod`) joining the existing `ComputeCluster`, `Container`, and `Function` vocabulary.

The goal is to let any consumer walk a workload from leaf to top with a single variable-length Cypher pattern instead of bespoke per-provider queries:

```cypher
MATCH path = (c:Container)-[:WORKLOAD_PARENT*1..6]->(top)
WHERE NOT (top)-[:WORKLOAD_PARENT]->()
RETURN path
```

The new chains wired in this PR:

| Provider | Chain |
|---|---|
| AWS ECS | `ECSContainer -> ECSTask -> ECSService -> ECSCluster` (standalone tasks fall back directly to `ECSCluster`) |
| Kubernetes | `KubernetesContainer -> KubernetesPod -> KubernetesNamespace -> KubernetesCluster` |
| GCP Cloud Run | `GCPCloudRun{Service,Job}Container -> GCPCloudRun{Service,Job}` (serverless: chain stops at the orchestrator) |
| Azure ACI | `AzureContainerInstance -> AzureGroupContainer` (serverless: chain stops at the group) |

Existing chain rels (`HAS_SERVICE`, `HAS_TASK`, `HAS_CONTAINER`, namespace/pod/container `CONTAINS`) stay in place in parallel and are flagged with `# DEPRECATED: replaced by WORKLOAD_PARENT, will be removed in v1.0.0` so consumers have time to migrate.

A few implementation notes:

- For ECS, service-attached tasks chain through the service; standalone tasks chain directly to the cluster. The ECS loader's `transform_ecs_tasks` sets a transient `_workload_parent_cluster_arn` only when the task lacks `serviceName`, gating the cluster matcher so each task gets exactly one `WORKLOAD_PARENT` edge.
- Kubernetes namespace and pod sub-resource `RESOURCE` rels stay untouched because they back scoped cleanup. Only the `CONTAINS` chain rels (`namespace -> pod`, `pod -> container`) are flagged deprecated.
- The PR is split commit-by-commit per provider plus a final ontology doc commit, so reviewers can step through one cloud at a time.

### Related issues or links

- Fixes #

### Breaking changes

None. Legacy chain relationships are kept in parallel and only flagged as deprecated; their removal is deferred to v1.0.0.

### How was this tested?

- `make test_lint` (pre-commit clean: black, isort, flake8, mypy)
- `uv run pytest tests/unit/cartography/intel/{aws,kubernetes,gcp,azure}` -> 450 passed
- `uv run pytest tests/integration/cartography/intel/aws/test_ecs.py tests/integration/cartography/intel/kubernetes/{test_namespaces,test_pods,test_containers}.py tests/integration/cartography/intel/gcp/test_cloudrun.py tests/integration/cartography/intel/azure/{test_container_instances,test_group_containers}.py` -> 28 passed

After re-syncing a tenant with all four providers active, the following Cypher queries should report:

```cypher
// New labels populated
MATCH (n) WHERE n:ComputeService OR n:ComputeNamespace OR n:ComputePod
RETURN labels(n)[0] AS primary, count(*) AS count;
// Expect non-zero counts per active provider.

// New relationship type exists and chains terminate at the right top
MATCH path = (c:Container)-[:WORKLOAD_PARENT*1..6]->(top)
WHERE NOT (top)-[:WORKLOAD_PARENT]->()
RETURN labels(c)[0], labels(top) LIMIT 20;
// Expect ECSContainer/KubernetesContainer chains end at ECSCluster/KubernetesCluster;
// Cloud Run / ACI containers end at their service / group (no cluster).

// Legacy rels still present in parallel
MATCH (s:ECSService)<-[:HAS_SERVICE]-(:ECSCluster) RETURN count(*);
// Expect non-zero. Same for HAS_TASK, HAS_CONTAINER, CONTAINS.
```

#### General

#### Proof of functionality

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (per-provider docs + ontology schema doc).

### Notes for reviewers

- The PR is structured as one commit per provider plus a closing ontology doc commit; reviewing one cloud at a time should keep the diff manageable.
- The biggest design call is around ECS tasks with vs. without a service. The transform-side `_workload_parent_cluster_arn` field is the cleanest way I found to keep "exactly one `WORKLOAD_PARENT` per task" without splitting the data into two load passes; happy to revisit if you'd prefer two separate load steps instead.
- Kubernetes namespace/pod sub-resource `RESOURCE` rels are intentionally not deprecated: they back scoped cleanup, and removing them would break the cleanup boundary. Only the `CONTAINS` chain rels are flagged.

### DIFF
--- a/cartography/models/azure/group_container.py
+++ b/cartography/models/azure/group_container.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from cartography.models.core.common import PropertyRef
 from cartography.models.core.nodes import CartographyNodeProperties
 from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.nodes import ExtraNodeLabels
 from cartography.models.core.relationships import CartographyRelProperties
 from cartography.models.core.relationships import CartographyRelSchema
 from cartography.models.core.relationships import LinkDirection
@@ -66,6 +67,7 @@ class AzureGroupContainerToSubnetRel(CartographyRelSchema):
 @dataclass(frozen=True)
 class AzureGroupContainerSchema(CartographyNodeSchema):
     label: str = "AzureGroupContainer"
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["ComputeService"])
     properties: AzureGroupContainerNodeProperties = AzureGroupContainerNodeProperties()
     sub_resource_relationship: AzureGroupContainerToSubscriptionRel = (
         AzureGroupContainerToSubscriptionRel()
--- a/cartography/models/azure/group_container.py
+++ b/cartography/models/azure/group_container.py
@@ -67,7 +67,7 @@ class AzureGroupContainerToSubnetRel(CartographyRelSchema):
 @dataclass(frozen=True)
 class AzureGroupContainerSchema(CartographyNodeSchema):
     label: str = "AzureGroupContainer"
-    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["ComputeService"])
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["ComputePod"])
     properties: AzureGroupContainerNodeProperties = AzureGroupContainerNodeProperties()
     sub_resource_relationship: AzureGroupContainerToSubscriptionRel = (
         AzureGroupContainerToSubscriptionRel()
--- a/cartography/intel/aws/ecs.py
+++ b/cartography/intel/aws/ecs.py
@@ -231,6 +231,12 @@ def transform_ecs_tasks(tasks: list[dict[str, Any]]) -> list[dict[str, Any]]:
         if group and group.startswith("service:"):
             task["serviceName"] = group.split("service:", 1)[1]
 
+        # Standalone tasks (no service) attach WORKLOAD_PARENT directly to the
+        # cluster; service-attached tasks chain through the service instead so
+        # the matcher stays null and only one parent edge fires.
+        if not task.get("serviceName"):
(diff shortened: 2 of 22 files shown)


cartography/models/azure/group_container.py
```diff
@@ -66,6 +67,7 @@ class AzureGroupContainerToSubnetRel(CartographyRelSchema):
 @dataclass(frozen=True)
 class AzureGroupContainerSchema(CartographyNodeSchema):
     label: str = "AzureGroupContainer"
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["ComputeService"])
```
kunaals: i think this should be `ComputePod` rather than `ComputeService`. azure describes an aci container group as a set of containers scheduled together on the same host with shared lifecycle/network/storage, and explicitly says it is similar to a kubernetes pod. the schema docs here say the same thing: this is analogous to an ecs task or kubernetes pod. if we label it as `ComputeService`, generic workload queries will miss aci when walking `Container -> ComputePod`, and service inventories will include a pod-shaped resource.

kunaals (approved):
approving to unblock but there is one change i think we should make, im open to pushback

PR #2737, opened 2026-05-02
refactor: enforce naming and sub_resource conventions in graph models

jychp:
### Type of change
- [ ] Documentation update

### Summary

The naming and `sub_resource_relationship` checks in `tests/unit/cartography/graph/test_model.py` were emitting `UserWarning`s that no one acted on. This branch:

1. Brings every model schema, intel loader, cleanup, and doc into alignment with those conventions.
2. Tightens the test from soft warnings to hard `assert` failures so any future regression breaks CI.

The bulk of the work is in three shapes:

- **Naming fixes (chore)**: rename AWS Bedrock relationship classes to end with `Rel` and PagerDuty `RelProperties` classes to add the missing `Rel` infix; relax the convention test to also accept `MatchLinkProperties` as a valid suffix.
- **`sub_resource_relationship` fixes (refactor)**: flip GSuite `OUTWARD`->`INWARD`, scope `AWSPolicyStatement` to `AWSAccount`, scope GitHub `BranchProtectionRule` / `Dependency` / `DependencyGraphManifest` / repo-level Actions `Secret` and `Variable` to `GitHubOrganization`, introduce a `CrowdstrikeTenant` parent for `CrowdstrikeHost` and `SpotlightVulnerability`. Each refactor collapses a per-repo (or per-policy / per-CID) cleanup loop into a single org-scoped `GraphJob` and ships an analysis migration that backfills the new `RESOURCE` edges so existing graphs aren't orphaned from the new cleanup pattern.
- **Test improvements**: add explicit allowlists for known intentional exceptions (canonical ontology nodes, AWS-owned shared resources, `KubernetesOIDCProvider`'s `TRUSTS` rel label, scanner modules without a tenant API, GitHub nodes that can exist outside any org), aggregate per-label rather than per-class so aliasing/facet schemas no longer count as roots, and convert every warning into a collected assertion error.

### Related issues or links

- Fixes #

### Breaking changes

None at the API level. Three behavioural changes are worth calling out for operators:

1. After upgrade, the first sync of GSuite, AWS IAM, GitHub (org/repo data), and CrowdStrike runs a one-shot analysis job that creates new `RESOURCE` edges (and `CrowdstrikeTenant` nodes) so the new scoped cleanup can see existing data. These jobs are idempotent (`MERGE`) and labelled `DEPRECATED ... Remove in v1.0.0` in the JSON and call sites.
2. CrowdStrike `cleanup_repo_level` was removed (folded into the org-level cleanup); the public function `cleanup_repo_level` no longer exists.
3. `cartography.intel.aws.iam.load_policy_statements` and `cartography.intel.github.repos.load_branch_protection_rules` (and the deps/manifests/secrets/variables loaders) gained an `owner_org_id` / `current_aws_account_id` parameter; existing call sites in this repo are updated.

### How was this tested?

- `uv run pytest tests/unit/cartography/graph/test_model.py` is now `3 passed, 0 warnings` and fails hard on regressions.
- `uv run pytest tests/integration/cartography/intel/aws/iam/` (10 passed), `tests/integration/cartography/intel/github/test_repos.py` (7 passed), `tests/integration/cartography/intel/github/test_actions.py` (9 passed), and `tests/integration/cartography/intel/crowdstrike/` (3 passed) all green.
- `pre-commit run` clean across every touched file (black/isort/flake8/mypy/pyupgrade).

#### General

#### Proof of functionality

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (AWS, GitHub, CrowdStrike).
- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md). _Not relevant: structure of the README is unchanged; only per-module schemas and rel patterns moved._

### Notes for reviewers

- The `sub_resource_relationship` refactors all follow the same pattern: add `*ToOrgRel` (or `*ToAccountRel` / `*ToTenantRel`), demote the previous edge to `other_relationships`, add an analysis migration JSON to backfill `RESOURCE` edges, and simplify the intel cleanup. Reviewing the AWS PolicyStatement commit first is the easiest entry point because it's the smallest of that shape; the GitHub commits then follow the same recipe.
- Three exception lists in `tests/unit/cartography/graph/test_model.py` are now the canonical place to document intentional architectural exceptions: `SUB_RESOURCE_REL_LABEL_EXCEPTIONS`, `MODULES_WITHOUT_TENANT_ROOT`, and `GLOBAL_NODE_LABELS`. Each entry has a short comment explaining why it's allowlisted.


cartography/models/github/dependencies.py
```diff
@@ -69,9 +93,12 @@ class DependencyGraphManifestToDependencyRel(CartographyRelSchema):
 class GitHubDependencySchema(CartographyNodeSchema):
     label: str = "Dependency"
     properties: GitHubDependencyNodeProperties = GitHubDependencyNodeProperties()
-    sub_resource_relationship: GitHubDependencyToRepositoryRel = (
-        GitHubDependencyToRepositoryRel()
+    sub_resource_relationship: GitHubDependencyToOrganizationRel = (
```
kunaals: i think this makes `Dependency` cleanup unsafe because the node id is still package-global (`canonical_name|requirements`), not repo/org-scoped. `GraphJob.from_node_schema()` will now detach-delete a stale dependency through one org's `RESOURCE` edge even if another repo/org still has `REQUIRES` or `HAS_DEP` edges to the same node. could we either keep cleanup relationship-only for these shared package nodes, or make the dependency id include the owning scope if we want org-scoped node deletion? same shape is safer for `DependencyGraphManifest` too, even though its `repo_url#path` id is much less likely to be shared across orgs.

cartography/models/crowdstrike/tenant.py
```diff
@@ -0,0 +1,24 @@
...
+    points to this tenant; CrowdstrikeTenant is the cleanup scope for those
+    nodes.
+    """
+
+    id: PropertyRef = PropertyRef("id")
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class CrowdstrikeTenantSchema(CartographyNodeSchema):
```
kunaals, line 23: should this tenant root get `ExtraNodeLabels(["Tenant"])` and a tenant ontology mapping? most tenant root nodes expose the shared `Tenant` label even when the mapping only has a minimal/id-backed name, and this new `CrowdstrikeTenant` looks like the same kind of organizational boundary. if it is meant to be cleanup-only and not ontology-visible, can we call that out explicitly?

tests/unit/cartography/graph/test_model.py
```diff
@@ -22,88 +21,145 @@
...
+    # than RESOURCE. Statement IDs of AWS-managed policies are global (the same
+    # statement is shared by every account that attaches the policy), so making
+    # the account the cleanup scope would risk a delete in account A removing a
+    # statement still referenced by account B; the per-policy STATEMENT scope
+    # avoids that cross-account interference.
+    "AWSPolicyStatement",
+    # OIDC providers express a trust relationship with the cluster rather than
+    # a strict ownership; an OIDC provider can in principle be referenced by
+    # multiple clusters.
+    "KubernetesOIDCProvider",
```
kunaals: can you confirm the multi-cluster trust scenario actually occurs in practice? if OIDC providers are effectively 1:1 with clusters in real data, the standard `RESOURCE` sub-resource plus `TRUSTS` as an `other_relationship` is cleaner. if they really can be shared, this exception is fine, but the comment is doing load-bearing modeling work and i'd rather keep the exception list from growing without that being explicit.

kunaals (changes requested):

jychp:
Closing/reopening to retrigger Test Suite workflow which didn't fire on the latest push.

kunaals (approved):

PR #2743, opened 2026-05-04
fix(gcp): Refactor GCP Artifact Registry image modeling

kunaals:
### Type of change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Refactoring (no functional changes)
- [x] Documentation update
- [ ] Other (please describe):

### Summary
- Split GCP Artifact Registry Docker image records into repository-scoped `GCPArtifactRegistryRepositoryImage` nodes and digest-scoped `GCPArtifactRegistryImage` nodes.
- Use ontology-aligned image labels and relationships: `ContainerRegistry`, `ImageTag`, `Image`, `ImageManifestList`, `REPO_IMAGE`, `IMAGE`, and `CONTAINS_IMAGE`.
- Keep the GAR-native `CONTAINS` repository edge while also adding `ContainerRegistry -[:REPO_IMAGE]-> ImageTag` for generic registry queries.
- Retarget runtime image relationships to canonical digest-scoped images and preserve pullable GAR URIs on repository image references.
- Add cleanup for old GAR image nodes and orphan canonical images after the new graph shape is loaded.
- Update Artifact Registry, runtime image, ontology, AIBOM, and schema documentation coverage.

### Related issues or links
N/A

### Breaking changes
This removes the old `GCPArtifactRegistryContainerImage` and `GCPArtifactRegistryPlatformImage` model path. New ingests emit `GCPArtifactRegistryRepositoryImage` for pullable repository references and `GCPArtifactRegistryImage` for digest-scoped image content. Existing graphs are cleaned by migration/orphan cleanup jobs after the replacement nodes and relationships exist.

### How was this tested?
- `uv run --frozen pytest tests/integration/cartography/intel/aibom/test_aibom.py tests/integration/cartography/intel/gcp/test_artifact_registry.py tests/unit/cartography/intel/gcp/test_artifact_registry_artifact.py -q`
- `uv run --frozen pytest tests/unit/cartography/intel/ontology/test_ontology_mapping.py tests/integration/cartography/intel/ontology/test_resolved_images.py tests/integration/cartography/intel/ontology/test_packages.py tests/integration/cartography/intel/ontology/test_packages_end_to_end.py -q`
- `uv run --frozen pytest tests/integration/cartography/intel/gcp/test_artifact_registry.py tests/unit/cartography/intel/test_supply_chain.py -q`
[...]

PR #2745, opened 2026-05-04
feat(gcp): derive Artifact Registry provenance from SBOMs

kunaals:
### Type of change

### Summary
Adds a GCP Artifact Registry provenance fallback for digest-specific SPDX SBOM image tags. When OCI labels and OCI referrers do not expose source metadata, Cartography now checks for legacy ko-style SBOM tags, including separate SBOM image artifacts tagged with the subject image digest.

The fallback validates that the SPDX document is tied to the image digest and only populates source_uri when the source repository can be corroborated from the SBOM package data. This feeds the existing image provenance path, so existing GitHub/GitLab PACKAGED_FROM matching can use the derived source URI without new node or relationship types.

### Related issues or links
N/A

### Breaking changes
None.

### How was this tested?
- Ran locally
- updated tests

#### General

#### Proof of functionality

N/A

N/A

N/A

### Notes for reviewers
This PR is stacked on the GCP Artifact Registry repository image model PR (#2743). The fallback intentionally avoids dependency-only package guesses: for separate SBOM image artifacts, the source repository must match the import-path-shaped SBOM image name and appear in the SPDX package data before source_uri is populated.


cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,232 @@ async def _fetch_attestation_provenance(
...
+    image itself was packaged from those repositories.
+    """
+    digest_value = (
+        image_digest.split(":", 1)[1] if ":" in image_digest else image_digest
+    )
+    sbom_identity = " ".join(
+        str(sbom.get(field, "")) for field in ("name", "documentNamespace")
+    )
+    if image_digest not in sbom_identity and digest_value not in sbom_identity:
+        return {}
```
jychp: Drop the digest match in the SBOM document identity. The fallback already proves the SBOM belongs to the image (via the `sha256-<digest>.sbom` tag, or via `documentDescribes`). Requiring the digest to also appear in `name`/`documentNamespace` rejects valid SPDX outputs whose generators emit generic document names like `sbom.spdx.json`.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,232 @@ async def _fetch_attestation_provenance(
...
+    for artifact in docker_artifacts_raw:
+        for tag in artifact.get("tags") or []:
+            if not isinstance(tag, str):
+                continue
+            if not tag.startswith("sha256-") or not tag.endswith(".sbom"):
+                continue
+            digest_value = tag.removeprefix("sha256-").removesuffix(".sbom")
+            if not digest_value:
+                continue
+            sbom_artifacts.setdefault(f"sha256:{digest_value}", artifact)
```
jychp: `setdefault` keeps only the first SBOM artifact per subject digest, so subsequent candidates (image copies, multi-variant SBOMs) are silently ignored. If the first one lacks a usable SPDX layer, the second fallback fails even when later artifacts would have worked. Store a list and try them in order.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -316,6 +546,59 @@ async def _process_single_image(
...
+        subject_digest = uri.split("@")[-1] if "@" in uri else manifest_digest
+        if subject_digest and subject_digest.startswith("sha256:"):
+            try:
+                sbom_provenance = await _fetch_legacy_sbom_provenance(
+                    http_client,
+                    token_manager,
+                    registry,
+                    image_path,
+                    subject_digest,
+                )
```
jychp, line 731: This calls the registry once per image whose `source_uri` is still empty after labels and Referrers, even when no `sha256-<digest>.sbom` tag exists in the project. On large repos that is N guaranteed 404 round-trips. Gate on `sbom_artifacts_by_digest` (which already enumerates digest-tagged SBOMs) so the legacy URL fetch only happens when there's a candidate.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,232 @@ async def _fetch_attestation_provenance(
...
+def _extract_source_from_spdx_sbom(
+    sbom: dict[str, Any],
+    image_digest: str,
+    expected_source_uri: str | None = None,
+) -> dict[str, str]:
+    """Extract source repo from a digest-specific SPDX SBOM.
+
+    Only the package(s) named by documentDescribes are considered. Dependency
+    packages can also contain GitHub URLs, but they are not evidence that the
+    image itself was packaged from those repositories.
```
jychp: Docstring says only described packages are considered, but the `expected_source_uri` branch below scans every package, so a dependency with a matching URL falsely confirms provenance. Either restrict that loop to `described_ids`, or rewrite the docstring to describe the two modes (this is the same concern flagged by cubic-dev-ai).

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,232 @@ async def _fetch_attestation_provenance(
...
+    except ValueError:
+        return None
+
+    if len(parts) <= github_idx + 2:
+        return None
+
+    owner = parts[github_idx + 1]
+    repo = parts[github_idx + 2]
+    if not owner or not repo:
+        return None
```
jychp, line 297: ko / GitLab combinations get nothing: only `github.com` paths and `pkg:golang/github.com/...` purls are recognized. Worth a TODO or follow-up if GitLab provenance is in scope; otherwise the `extract_image_source_provenance` GitLab path that exists upstream is reachable only via labels/Referrers, not this fallback.

jychp (changes requested):

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: good catch. dropped the document-name digest gate and added coverage for generic sbom document names.

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: good catch. this now keeps all matching sbom artifacts for a digest and tries them in order.

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: good catch. gated the legacy sbom probe on discovered sbom candidates so we avoid guaranteed 404s.

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: yep, updated the docstring to make the ko expected-source mode explicit.

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: keeping this github-only in this pr. gitlab can still come from labels/referrers; sbom gitlab support can be a follow-up.

kunaals:
@cubic-dev-ai another review please

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,227 @@ async def _fetch_attestation_provenance(
...
+        repo_url = _repo_url_from_golang_purl(locator)
+        if repo_url:
+            return repo_url
+
+    return None
+
+
+def _extract_source_from_spdx_sbom(
+    sbom: dict[str, Any],
+    image_digest: str,
```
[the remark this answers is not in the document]
kunaals: good catch. removed the unused parameter and updated the call sites/tests.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,225 @@ async def _fetch_attestation_provenance(
...
+    return sbom_artifacts
+
+
+async def _fetch_spdx_layer_provenance(
+    http_client: httpx.AsyncClient,
+    token_manager: _TokenManager,
+    registry: str,
+    image_path: str,
+    reference: str,
+    subject_digest: str,
```
jychp, line 519: `subject_digest` is accepted here but never read inside the helper. Either validate that the SPDX document's described package matches this digest (so the docstring's digest-anchored evidence claim actually holds) or drop the parameter. Without one of those, a mis-tagged or reused `sha256-<digest>.sbom` artifact silently attributes the wrong source repo to this image. Same dead-parameter shape that was just removed from `_fetch_legacy_sbom_provenance` in 1db47a66.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,225 @@ async def _fetch_attestation_provenance(
...
+
+    try:
+        return await _fetch_spdx_layer_provenance(
+            http_client,
+            token_manager,
+            registry,
+            image_path,
+            sbom_tag,
+            image_digest,
+            expected_source_uri=_repo_url_from_github_path(image_path),
```
jychp, line 588: For typical GAR image paths (`project/repo/image`), `_repo_url_from_github_path(image_path)` returns `None`, so `_extract_source_from_spdx_sbom` falls into the `documentDescribes`-only branch. Real ko SBOMs describe the OCI image package and put the source module in a dependency package, so this legacy `sha256-<digest>.sbom` fallback never extracts source for actual ko output at the same image path. The unit test only covers it because `_spdx_sbom` (not `_ko_spdx_sbom`) is fed in. Either resolve `expected_source_uri` from another signal here, or scan dependency packages when `documentDescribes` points at the OCI image.

jychp (changes requested):

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: good catch. added coverage for the normal gar path plus ko-shaped sbom case, and now we can use the source dependency only after the described oci package matches the image digest.

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: good catch. fixed this by validating described oci packages against the subject digest before using dependency source hints.

kunaals:
@cubic-dev-ai can you do one more full review?

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -242,10 +250,299 @@ async def _fetch_attestation_provenance(
...
+            matches
+            for package in described_packages
+            if (
+                matches := _spdx_package_matches_subject_digest(package, subject_digest)
+            )
+            is not None
+        ]
+        if False in subject_digest_matches:
+            return {}
+        subject_digest_verified = any(subject_digest_matches)
```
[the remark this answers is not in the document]
kunaals: good catch. tightened this so expected-source matches on dependency packages require verified subject digest; described source packages can still match directly.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -292,37 +597,97 @@ async def _process_single_image(
...
+    # Some older build flows publish SPDX SBOMs as digest-specific image tags
+    # instead of OCI referrers. Use them only when the document itself ties back
+    # to this image digest and names one described source package.
+    sbom_artifacts = (
+        sbom_artifacts_by_digest.get(subject_digest_str)
+        if subject_digest_str and sbom_artifacts_by_digest
+        else None
+    )
+    if not provenance.get("source_uri") and sbom_artifacts and subject_digest_str:
+        try:
```
jychp, line 697: Step 1 (_fetch_legacy_sbom_provenance) fires a request against the main image path every time sbom_artifacts is non-empty, even when every sbom_artifact URI resolves to a different image path. For the common case where the SBOM lives at a separate github.com/... artifact, this produces a guaranteed 404 per image before step 2 does the useful fetch. Gate step 1 so it only runs when at least one sbom_artifact URI parses to the same image_path as the main image.

jychp (approved):
approved to unblock, one remaining change before merging

cartography/intel/gcp/artifact_registry/supply_chain.py
kunaals: good catch. added the same-path gate so the legacy tag fetch only runs when a discovered sbom artifact is on the main image path; separate sbom artifact paths go straight to the explicit artifact fetch loop.

PR #2747, opened 2026-05-04
perf(sync): defer provider SDK loading until a stage runs

jychp:
### Type of change

### Summary

Importing `cartography.sync` (used by both the CLI and library callers) used to eagerly load all 46 `cartography.intel.*` provider packages, pulling in `boto3` + `aioboto3`, 20+ `azure-mgmt-*` + `azure-identity`, `google-cloud-*` + `googleapiclient`, `msgraph-sdk`, `kubernetes`, `oci`, `okta`, `crowdstrike-falconpy`, and ~15 more SDKs at startup even when only one provider was being used.

This PR replaces the direct attribute references in `TOP_LEVEL_MODULES` with a tiny `_LazyStage` callable that resolves `cartography.intel.<provider>.<func>` via `importlib` at first invocation. The public surface (`TOP_LEVEL_MODULES` keys, `build_default_sync`, `build_sync`, `Sync.run` signature, stage call signature `(neo4j_session, config) -> None`) is unchanged.

`Sync.list_intel_modules()` already does deep eager iteration by design (discovery utility), so its three direct references to `create_indexes` / `ontology` / `analysis` were swapped to `importlib.import_module` rather than re-introduced as top-level imports.

### Perf impact

Measured on Python 3.13.13 / macOS, median of 5 runs (each in a fresh subprocess):

| Scenario                  | Before    | After    | Delta         |
|---------------------------|-----------|----------|---------------|
| `import cartography.sync` | 5409.8 ms | 177.6 ms | **-96.7%, 30x** |
| `build_sync('azure')`     | 5443.8 ms | 173.5 ms | -96.8%        |
| `build_default_sync()`    | 5402.7 ms | 174.9 ms | -96.8%        |
| RSS, `import cartography.sync`     | 578.5 MB  | 79.3 MB  | **-499 MB** |
| RSS, `build_sync('azure')`         | 578.9 MB  | 79.4 MB  | -499 MB       |
| Python heap (`tracemalloc`)        | 463.3 MB  | 47.4 MB  | -415 MB       |

`python -X importtime -c "import cartography.sync"` drops from **8385 import lines / 6.04 s cumulative** to **660 lines / 0.18 s cumulative**.

When a stage actually runs, its provider package and the SDK behind it load on first call, so steady-state cost during a real sync is unchanged.

### How was this tested?

- `make test_unit` — 1740 tests pass.
- `make test_lint` — clean (ruff / black / isort / mypy / flake8 / pyupgrade).
- Smoke: `from cartography.sync import build_default_sync, build_sync, TOP_LEVEL_MODULES` works; `TOP_LEVEL_MODULES['aws']` is callable; `analysis` is still last; `build_sync('azure')` produces a single-stage Sync without resolving the AWS / GCP / Microsoft modules.
- The contributor-facing `create-module` skill was updated with a "Register the module" step that mandates `_LazyStage` and warns against re-introducing a top-level `import cartography.intel.<service>` to `sync.py`.

#### General
- [x] I have added/updated tests that prove my fix is effective or my feature works. (no new behavior; existing tests cover the public surface — `tests/unit/cartography/test_sync.py`, `tests/unit/rules/test_model.py`, `tests/unit/cartography/intel/ontology/test_ontology_mapping.py`).

### DIFF
--- a/.agents/skills/create-module/SKILL.md
+++ b/.agents/skills/create-module/SKILL.md
@@ -47,7 +47,23 @@ In `cartography/cli.py`:
 
 In `cartography/config.py`, extend `Config.__init__` with the new fields. Then in your module entry point, validate them and short-circuit with `logger.info("... not configured - skipping module")` when missing.
 
-### Step 3 — Implement the sync pattern
+### Step 3 — Register the module in `cartography/sync.py`
+
+Add one entry to `TOP_LEVEL_MODULES` using the lazy wrapper. **Do not add a top-level `import cartography.intel.your_service` to `sync.py`** — that defeats lazy SDK loading and reintroduces the slow-startup problem.
+
+```python
+TOP_LEVEL_MODULES = OrderedDict({
+    ...
+    "your_service": _LazyStage("cartography.intel.your_service", "start_your_service_ingestion"),
+    ...
+    # `analysis` must remain last
+    "analysis": _LazyStage("cartography.intel.analysis", "run"),
+})
+```
+
+Pick a sensible position relative to neighbors (cloud providers grouped together, etc.). The provider's heavy SDK imports stay where they are — they only fire when this stage is selected and run.
+
+### Step 4 — Implement the sync pattern
 
 For each domain (users, devices, projects, ...):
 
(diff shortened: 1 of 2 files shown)


kunaals (approved):
awesome improvement!

PR #2748, opened 2026-05-04
fix(azure): skip transient data factory failures

kunaals:
### Type of change

### Summary
Azure Data Factory listing calls can fail with transient Azure Resource Manager errors. Previously, those errors could fail the full Azure sync, and treating them as empty results would risk stale cleanup from an incomplete API response.

This change adds a small Data Factory-scoped retry helper for transient HTTP statuses, raises a sanitized internal skip signal after retry exhaustion, and skips the Data Factory subtree for the affected subscription. Successful empty responses still go through normal load and cleanup.

### Related issues or links
- N/A

### Breaking changes
None.

### How was this tested?
- `uv run --frozen pytest -vvv tests/unit/cartography/intel/azure/test_data_factory.py`
- `uv run --frozen pytest -vvv tests/integration/cartography/intel/azure/test_data_factory.py`
- `make test_lint`

#### General
- [x] The linter passes locally (`make test_lint`; this checkout does not define `make lint`).

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
This does not change Azure Data Factory models or cleanup semantics for successful API responses. It only prevents cleanup from running when the Data Factory API result is known to be incomplete after transient retry exhaustion.

A transient failure in any Data Factory list stage skips the remaining Data Factory subtree for that subscription for the current run. The next successful run will resume normal load and cleanup behavior.

### DIFF
--- a/cartography/intel/azure/__init__.py
+++ b/cartography/intel/azure/__init__.py
@@ -92,7 +92,8 @@ def _sync_data_factory(
     except AzureDataFactoryTransientError as error:
         logger.warning(
             "Skipping Azure Data Factory sync after transient API failures "
-            "(operation=%s, status_code=%s).",
+            "for subscription %s (operation=%s, status_code=%s).",
+            subscription_id,
             error.operation,
             error.status_code,
         )
--- a/cartography/intel/azure/__init__.py
+++ b/cartography/intel/azure/__init__.py
@@ -46,6 +46,58 @@ from .util.credentials import Credentials
 logger = logging.getLogger(__name__)
 
 
+def _sync_data_factory(
+    neo4j_session: neo4j.Session,
+    credentials: Credentials,
+    subscription_id: str,
+    update_tag: int,
+    common_job_parameters: Dict,
+) -> None:
+    try:
+        factories_raw = data_factory.sync_data_factories(
+            neo4j_session,
+            credentials,
+            subscription_id,
+            update_tag,
+            common_job_parameters,
+        )
+        linked_services_by_factory = (
+            data_factory_linked_service.sync_data_factory_linked_services(
+                neo4j_session,
+                credentials,
+                factories_raw,
+                subscription_id,
+                update_tag,
+                common_job_parameters,
+            )
+        )
+        datasets_by_factory = data_factory_dataset.sync_data_factory_datasets(
+            neo4j_session,
+            credentials,
+            factories_raw,
+            linked_services_by_factory,
+            subscription_id,
+            update_tag,
+            common_job_parameters,
+        )
+        data_factory_pipeline.sync_data_factory_pipelines(
+            neo4j_session,
+            credentials,
+            factories_raw,
+            datasets_by_factory,
+            subscription_id,
+            update_tag,
+            common_job_parameters,
+        )
+    except AzureDataFactoryTransientError as error:
+        logger.warning(
+            "Skipping Azure Data Factory sync after transient API failures "
+            "(operation=%s, status_code=%s).",
+            error.operation,
+            error.status_code,
+        )
+
+
 def _sync_one_subscription(
     neo4j_session: neo4j.Session,
     credentials: Credentials,
@@ -152,49 +204,13 @@ def _sync_one_subscription(
         update_tag,
         common_job_parameters,
     )
-    try:
-        factories_raw = data_factory.sync_data_factories(
-            neo4j_session,
-            credentials,
-            subscription_id,
-            update_tag,
-            common_job_parameters,
-        )
-        linked_services_by_factory = (
-            data_factory_linked_service.sync_data_factory_linked_services(
-                neo4j_session,
-                credentials,
-                factories_raw,
-                subscription_id,
-                update_tag,
-                common_job_parameters,
-            )
-        )
-        datasets_by_factory = data_factory_dataset.sync_data_factory_datasets(
-            neo4j_session,
-            credentials,
-            factories_raw,
-            linked_services_by_factory,
-            subscription_id,
-            update_tag,
-            common_job_parameters,
-        )
-        data_factory_pipeline.sync_data_factory_pipelines(
-            neo4j_session,
-            credentials,
-            factories_raw,
-            datasets_by_factory,
-            subscription_id,
-            update_tag,
-            common_job_parameters,
-        )
-    except AzureDataFactoryTransientError as error:
-        logger.warning(
-            "Skipping Azure Data Factory sync after transient API failures "
-            "(operation=%s, status_code=%s).",
-            error.operation,
-            error.status_code,
-        )
+    _sync_data_factory(
+        neo4j_session,
+        credentials,
+        subscription_id,
+        update_tag,
+        common_job_parameters,
+    )
     data_lake.sync(
         neo4j_session,
         credentials,
--- a/cartography/intel/azure/__init__.py
+++ b/cartography/intel/azure/__init__.py
@@ -39,6 +39,7 @@ from . import storage
 from . import subscription
 from . import synapse
 from . import tenant
+from .data_factory_util import AzureDataFactoryTransientError
 from .util.credentials import Authenticator
 from .util.credentials import Credentials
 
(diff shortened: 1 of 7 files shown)


cartography/intel/azure/data_factory_util.py
```diff
@@ -0,0 +1,62 @@
...
+    if status_code is None:
+        return None
+
+    try:
+        return int(status_code)
+    except (TypeError, ValueError):
+        return None
+
+
+def call_data_factory_operation(operation: str, func: Callable[[], T]) -> T:
```
jychp: Cartography already depends on `backoff` (see `pyproject.toml`) and ships `cartography.helpers.backoff_handler`. The Azure-adjacent `cartography/intel/microsoft/entra/utils.py` uses `@backoff.on_exception(backoff.expo, ..., on_backoff=backoff_handler)` for exactly this shape of problem. Hand-rolling a retry loop here is inconsistent with that established pattern; reuse the existing helper unless there is a concrete reason it does not fit.

cartography/intel/azure/data_factory_util.py
```diff
@@ -0,0 +1,62 @@
...
+    for attempt in range(1, _MAX_ATTEMPTS + 1):
+        try:
+            return func()
+        except HttpResponseError as error:
+            status_code = _get_status_code(error)
+            if status_code not in _RETRYABLE_STATUS_CODES:
+                raise
+
+            if attempt == _MAX_ATTEMPTS:
+                raise AzureDataFactoryTransientError(operation, status_code) from None
```
jychp: `from None` discards the underlying `HttpResponseError`, which carries the Azure request id, error code, and server message. After three failed retries that is exactly what an operator needs in the logs. Drop `from None` (or use `from error`); the wrapper message stays sanitized either way. The unit test asserting `__cause__ is None` should be updated accordingly.

cartography/intel/azure/data_factory_util.py
```diff
@@ -0,0 +1,62 @@
...
+
+            logger.warning(
+                "Transient Azure Data Factory API error during %s "
+                "(status_code=%s). Retrying.",
+                operation,
+                status_code,
+            )
+            time.sleep(_RETRY_BACKOFF_SECONDS**attempt)
```
jychp: Pure `2**attempt` backoff with no jitter means parallel runners against the same throttled endpoint will synchronize their retries and amplify pressure. Add randomized jitter, or rely on `backoff.expo`'s built-in jitter per the helper-reuse comment above.

cartography/intel/azure/data_factory_util.py
```diff
@@ -0,0 +1,62 @@
...
+
+    raise RuntimeError("unreachable")
```
jychp: This line is unreachable: every iteration of the loop either returns or raises. Convert to `while True:` with an explicit break after a successful return, or drop the sentinel — static analyzers flag this kind of dead code.

cartography/intel/azure/__init__.py
```diff
@@ -39,12 +39,65 @@
...
+            credentials,
+            factories_raw,
+            datasets_by_factory,
+            subscription_id,
+            update_tag,
+            common_job_parameters,
+        )
+    except AzureDataFactoryTransientError as error:
+        logger.warning(
+            "Skipping Azure Data Factory sync after transient API failures "
```
jychp, line 92: Include `subscription_id` in this warning. In multi-subscription syncs this single line is the only signal that a subscription's Data Factory subtree is stale, and without the subscription id it is not actionable for triage.

tests/unit/cartography/intel/azure/test_data_factory.py
```diff
@@ -0,0 +1,210 @@
...
+        data_factory.get_factories(client)
+
+    assert excinfo.value.operation == "list data factories"
+    assert excinfo.value.status_code == 503
+    assert excinfo.value.__cause__ is None
+    assert client.factories.list.call_count == 3
+    assert mock_sleep.call_count == 2
+
+
+def test_sync_data_factory_skips_child_syncs_after_transient_error() -> None:
```
jychp, line 96: Coverage is missing for the realistic mid-sync failure: `sync_data_factories` succeeds (factories loaded + cleaned up), then a downstream child sync (e.g. `sync_data_factory_linked_services`) exhausts retries. Add a test that verifies `_sync_data_factory` swallows the error, the later child syncs are not invoked, and no exception escapes.

jychp (changes requested):

cartography/intel/azure/data_factory_util.py
kunaals: good catch. switched this to the shared backoff helper pattern.

cartography/intel/azure/data_factory_util.py
kunaals: good catch. changed this to chain from the underlying azure error.

cartography/intel/azure/data_factory_util.py
kunaals: good catch. using backoff.expo now, so retries use the library jitter instead of fixed sleeps.

cartography/intel/azure/data_factory_util.py
kunaals: fixed by removing the hand-rolled retry loop.

cartography/intel/azure/__init__.py
kunaals: good catch. added the subscription id to the skip warning.

tests/unit/cartography/intel/azure/test_data_factory.py
kunaals: good catch. added coverage for a child data factory sync failing after factories are fetched.

cartography/intel/azure/__init__.py
```diff
@@ -39,12 +39,66 @@
...
 logger = logging.getLogger(__name__)
 
 
+def _sync_data_factory(
+    neo4j_session: neo4j.Session,
+    credentials: Credentials,
+    subscription_id: str,
+    update_tag: int,
+    common_job_parameters: Dict,
+) -> None:
```
jychp, line 53: The `_sync_data_factory` signature uses `Dict` from `typing`. Replace with built-in `dict`

cartography/intel/azure/data_factory_util.py
```diff
@@ -0,0 +1,65 @@
...
+    @backoff.on_exception(  # type: ignore[misc]
+        backoff.expo,
+        HttpResponseError,
+        max_tries=_MAX_TRIES,
+        giveup=lambda error: not _is_retryable_data_factory_error(error),
+        on_backoff=backoff_handler,
+        logger=None,
+    )
+    def _call() -> T:
+        return func()
```
jychp: `_call` is redefined and re-decorated by `@backoff.on_exception` on every invocation of `call_data_factory_operation`. Refactor so the decorated wrapper is created once at module level instead of per-call.

cartography/intel/azure/data_factory_util.py
```diff
@@ -0,0 +1,65 @@
...
+
+    try:
+        return _call()
+    except HttpResponseError as error:
+        if not _is_retryable_data_factory_error(error):
+            raise
```
jychp, line 66: `_is_retryable_data_factory_error` (and its `_get_status_code` call inside) is invoked twice per exhausted-retry path: once by the `giveup` lambda and once here. Call `_get_status_code(error)` once in this `except` block and pass the result to both the check and `AzureDataFactoryTransientError`.

jychp (approved):
LGTM, some nit comments

cartography/intel/azure/data_factory_util.py
kunaals: good catch. moved the decorated wrapper to module scope.

cartography/intel/azure/data_factory_util.py
kunaals: good catch. the final exception path now gets the status code once and reuses it.

cartography/intel/azure/__init__.py
kunaals: good catch. switched these annotations to built-in types.

PR #2749, opened 2026-05-04
feat(azure): add Application Gateway intel module

### LINKED ISSUE #1736: feat(azure): improve Azure coverage (Application Gateway)

jychp:
### Type of change

### Summary

Adds an Azure Application Gateway intel module modelled on the existing Load Balancer module. This was the last unchecked item on issue #1736's Azure coverage checklist.

The module ingests one parent and five sub-resources:

- `AzureApplicationGateway` (with `LoadBalancer` ontology label, like `AzureLoadBalancer`)
- `AzureApplicationGatewayFrontendIPConfiguration`
- `AzureApplicationGatewayBackendPool`
- `AzureApplicationGatewayHTTPListener`
- `AzureApplicationGatewayBackendHTTPSettings`
- `AzureApplicationGatewayRequestRoutingRule`

Cross-resource edges: `IN_SUBNET` to `AzureSubnet`, `ASSOCIATED_WITH` to `AzurePublicIPAddress`, `ROUTES_TO` to `AzureNetworkInterface`. Tags reuse the shared `AzureTag` model.

`PathBasedRouting` rules resolve their effective backend pool and HTTP settings through the referenced `url_path_map`'s defaults, so `ROUTES_TO` and `USES_SETTINGS` edges are populated for both `Basic` and `PathBasedRouting` rules. Per-path overrides are not modeled as separate nodes and that limitation is documented.

Cleanup runs at subscription scope after the per-gateway loop so stale child nodes are deleted even when every gateway disappears between syncs.

### Related issues or links

- Closes #1736

### How was this tested?

- `make test_lint` (full pre-commit incl. mypy): clean
- `pytest tests/unit -q`: 1740 passed
- `pytest tests/integration/cartography/intel/azure/test_application_gateway.py -vvv`: 2 passed (covers both `Basic` and `PathBasedRouting` rules, all node types and relationships, plus the tag loader)

#### General
- [x] The linter passes locally (\`make lint\`).

#### Proof of functionality

### Notes for reviewers

The module deliberately mirrors `cartography/intel/azure/load_balancers.py` so the patterns (transform helpers, scoped cleanup at subscription level, `transform_tags()` reuse, ontology label) are easy to compare side-by-side.

Out of scope (intentional follow-ups):
- WAF / `AzureWebApplicationFirewallPolicy` cross-resource edges; `firewall_policy_id` is kept as a property only.
- Analysis jobs (exposure, firewall protection) analogous to `azure_lb_exposure.json`.
- URL path maps as separate nodes, redirect configurations, rewrite rule sets, SSL certificate nodes, probes.


cartography/intel/azure/application_gateways.py
```diff
@@ -0,0 +1,514 @@
...
+            or rule.get("properties", {}).get("url_path_map")
+            or {}
+        )
+
+        listener_id = listener_ref.get("id")
+        backend_pool_id = backend_pool_ref.get("id")
+        backend_http_settings_id = backend_settings_ref.get("id")
+        url_path_map_id = url_path_map_ref.get("id")
+
+        # PathBasedRouting: fall back to the url_path_map's defaults when the
```
kunaals: i think this makes the path-based rule look more complete than it is. the url path map default only applies when no path rule matches; individual path rules can route to different backend pools/settings, so this `ROUTES_TO` edge can point at only the fallback backend and miss the actual matched backends. can we either model the path rules explicitly, or avoid emitting this edge for `PathBasedRouting` until we can represent it accurately?

cartography/intel/azure/application_gateways.py
```diff
@@ -0,0 +1,514 @@
...
+                    config, "private_ip_allocation_method"
+                ),
+                "public_ip_address_id": public_ip_ref.get("id"),
+                "subnet_id": subnet_ref.get("id"),
+            }
+        )
+    return transformed
+
+
+def transform_backend_pools(application_gateway: dict) -> list[dict]:
```
kunaals, line 113: app gateway backend pools can also target ip/fqdn/app service style backends, but the only traversable edge we create here is for nic-derived targets. that means common pools will have `fqdns`/`ip_addresses` as properties but no backend edge in the graph. if that’s intentional for this first pass, can we document `ROUTES_TO` as nic-only for now? otherwise we should model those target types too.

kunaals (approved):
approving to unblock with some small followup qs below

PR #2757, opened 2026-05-05
fix(gitlab): ingest dependency artifacts by job id

kunaals:
### Type of change
- [x] Documentation update

### Summary
GitLab dependency ingestion currently discovers a successful dependency scanning job but downloads artifacts through the ref/job-name endpoint. That can miss the selected job when the latest matching job is not in the latest default-branch successful pipeline, and it also misses newer/gzipped CycloneDX artifact names.

This changes the dependency artifact download path to use the selected job ID, paginates successful job discovery, prefers default-branch dependency scanning jobs, and parses CycloneDX SBOM artifacts from ZIP archives or direct artifact responses, including gzipped reports. Invalid gzipped CycloneDX artifacts now surface as parse failures instead of being treated as an empty dependency list. The GitLab module docs now also call out the CI artifact download permissions needed for dependency ingestion.

Docs referenced while validating this change:
- GitLab Job Artifacts API: https://docs.gitlab.com/api/job_artifacts/
- GitLab job artifact access controls: https://docs.gitlab.com/ci/jobs/job_artifacts/
- GitLab artifacts:access: https://docs.gitlab.com/ci/yaml/#artifactsaccess
- GitLab dependency scanning SBOM artifacts: https://docs.gitlab.com/user/application_security/dependency_scanning/dependency_scanning_sbom/

### Related issues or links
N/A

### Breaking changes
None.

### How was this tested?
- Local GitLab.com sync against a test group with a successful `gemnasium-dependency_scanning` job and a CycloneDX artifact archive
- Updated tests

#### General
- [x] The linter passes locally (`make test_lint`; this checkout does not define `make lint`).

#### Proof of functionality

Redacted local GitLab.com validation excerpt:

```text
INFO:cartography.sync:Starting sync stage 'gitlab'
INFO:cartography.intel.gitlab:Starting GitLab sync for organization <gitlab_group_id> at https://gitlab.com
INFO:cartography.client.core.tx:Loaded 1 GitLabOrganization nodes
INFO:cartography.client.core.tx:Loaded 2 GitLabGroup nodes
INFO:cartography.client.core.tx:Loaded 4 GitLabProject nodes

INFO:cartography.intel.gitlab.dependency_files:Syncing GitLab dependency files for 4 projects
DEBUG:cartography.intel.gitlab.dependency_files:Found manifest file: package-lock.json
DEBUG:cartography.intel.gitlab.dependency_files:Found manifest file: package.json
INFO:cartography.client.core.tx:Loaded 2 GitLabDependencyFile nodes

INFO:cartography.intel.gitlab.dependencies:Syncing GitLab dependencies for 4 projects
DEBUG:cartography.intel.gitlab.dependencies:Found dependency scanning job 'gemnasium-dependency_scanning' (ID <job_id>, ref main) for project ID <project_id>
DEBUG:cartography.intel.gitlab.dependencies:Downloading artifacts for project ID <project_id> job ID <job_id>
DEBUG:cartography.intel.gitlab.dependencies:Artifacts download response status: 200
DEBUG:cartography.intel.gitlab.dependencies:Parsing CycloneDX SBOM file: gl-sbom-npm-npm.cdx.json
DEBUG:cartography.intel.gitlab.dependencies:Successfully parsed 2 dependencies from CycloneDX SBOM artifact(s) for project ID <project_id>
INFO:cartography.intel.gitlab.dependencies:Transformed 2 dependencies
INFO:cartography.client.core.tx:Loaded 2 GitLabDependency nodes
INFO:cartography.intel.gitlab.dependencies:GitLab dependencies sync completed
INFO:cartography.intel.gitlab:GitLab ingestion completed for organization <gitlab_group_id>
```

Graph verification from the local validation database:

```text
GitLabDependency nodes:
- npm:express@4.18.2
- npm:lodash@4.17.21

GitLabDependencyFile HAS_DEP relationships:
- package-lock.json -> npm:express@4.18.2
- package-lock.json -> npm:lodash@4.17.21
```

- [x] Included Cartography sync logs from a real environment demonstrating successful synchronization of the modified ingestion path. Logs show:
  - The number of dependency file and dependency nodes created or updated

### Notes for reviewers
This does not add GitLab Dependencies API support or manifest-file dependency parsing. It keeps dependency ingestion on dependency scanning artifacts and only changes artifact discovery/download/parsing behavior.

### DIFF
--- a/cartography/intel/gitlab/dependencies.py
+++ b/cartography/intel/gitlab/dependencies.py
@@ -375,9 +375,6 @@ def get_dependencies(
     except zipfile.BadZipFile as e:
         logger.error(f"Invalid ZIP file for job ID {job_id}: {e}")
         return []
-    except gzip.BadGzipFile as e:
-        logger.error(f"Invalid gzip file in CycloneDX SBOM for job ID {job_id}: {e}")
-        return []
     except json.JSONDecodeError as e:
         logger.error(f"Invalid JSON in CycloneDX SBOM: {e}")
         return []
--- a/cartography/intel/gitlab/dependencies.py
+++ b/cartography/intel/gitlab/dependencies.py
@@ -96,26 +96,22 @@ def _get_successful_jobs(
         jobs.extend(page_jobs)
 
         next_page = response.headers.get("X-Next-Page")
-        if next_page:
-            page = int(next_page)
-            continue
-        if len(page_jobs) < DEPENDENCY_SCAN_JOBS_PER_PAGE:
+        if not next_page:
             break
-        page += 1
+        page = int(next_page)
 
     return jobs
 
 
 def _is_cyclonedx_artifact(path: str) -> bool:
     filename = path.rsplit("/", 1)[-1]
-    if not filename.startswith("gl-sbom"):
-        return False
-    if not filename.endswith(CYCLONEDX_ARTIFACT_SUFFIXES):
-        return False
     return filename in {
         "gl-sbom.cdx.json",
         "gl-sbom.cdx.json.gz",
-    } or filename.startswith("gl-sbom-")
+    } or (
+        filename.startswith("gl-sbom-")
+        and filename.endswith(CYCLONEDX_ARTIFACT_SUFFIXES)
+    )
 
 
 def _load_cyclonedx_json(content: bytes, path: str) -> dict[str, Any]:
@@ -371,9 +367,6 @@ def get_dependencies(
 
     except requests.exceptions.RequestException as e:
         logger.error(f"Error downloading artifacts for job ID {job_id}: {e}")
-        return []
-    except zipfile.BadZipFile as e:
-        logger.error(f"Invalid ZIP file for job ID {job_id}: {e}")
         return []
     except json.JSONDecodeError as e:
         logger.error(f"Invalid JSON in CycloneDX SBOM: {e}")
--- a/cartography/intel/gitlab/dependencies.py
+++ b/cartography/intel/gitlab/dependencies.py
@@ -4,6 +4,7 @@ GitLab Dependencies Intelligence Module
 Fetches and parses individual dependencies from dependency scanning job artifacts.
 """
 
+import gzip
 import io
 import json
 import logging
@@ -33,18 +34,23 @@ DEFAULT_DEPENDENCY_SCAN_JOB_NAMES = frozenset(
         AUTODEVOPS_MAVEN_DEPENDENCY_SCAN_JOB_NAME,
     )
 )
+DEPENDENCY_SCAN_JOBS_PER_PAGE = 100
+CYCLONEDX_ARTIFACT_SUFFIXES = (".cdx.json", ".cdx.json.gz")
 
 
 def _select_dependency_scan_job(
     jobs: list[dict[str, Any]],
     dependency_scan_job_name: str | None,
+    default_branch: str | None = None,
 ) -> dict[str, Any] | None:
     """
     Select the latest dependency scan job from successful jobs.
 
     If no custom job name is configured, support all known GitLab dependency
     scan job names (default + AutoDevOps language-specific names). For any
-    custom job name, only that specific name is matched.
+    custom job name, only that specific name is matched. Prefer jobs that ran
+    on the project default branch so branch-scoped dependency data wins over
+    newer merge request or feature branch jobs.
     """
     candidate_names: set[str] | frozenset[str]
     if dependency_scan_job_name is None:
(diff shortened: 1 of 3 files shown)


cartography/intel/gitlab/dependencies.py
```diff
@@ -33,29 +34,193 @@
...
+        next_page = response.headers.get("X-Next-Page")
+        if next_page:
+            page = int(next_page)
+            continue
+        if len(page_jobs) < DEPENDENCY_SCAN_JOBS_PER_PAGE:
+            break
+        page += 1
+
+    return jobs
+
```
jychp, line 104: [suggestion] The loop has two exit conditions for the same scenario. When `X-Next-Page` is present, `continue` restarts the loop before reaching the `len(page_jobs) < DEPENDENCY_SCAN_JOBS_PER_PAGE` check. When the header is absent and the page is exactly full (100 jobs), neither condition triggers a `break`, so `page += 1` fires and the next request returns an empty list, wasting a round trip. Either rely exclusively on `X-Next-Page` to advance (break when the header is absent) or remove `page += 1` and treat `len < per_page` as the sole no-header terminator.

cartography/intel/gitlab/dependencies.py
```diff
@@ -33,29 +34,193 @@
...
+    if not filename.startswith("gl-sbom"):
+        return False
+    if not filename.endswith(CYCLONEDX_ARTIFACT_SUFFIXES):
+        return False
+    return filename in {
+        "gl-sbom.cdx.json",
+        "gl-sbom.cdx.json.gz",
+    } or filename.startswith("gl-sbom-")
+
+
```
jychp, line 116: [nitpick] The first guard (`not filename.startswith('gl-sbom')`) already rejects anything that does not start with `gl-sbom`, so the inner `filename.startswith('gl-sbom-')` branch of the final `return` is always true at that point. Simplify: after the two prefix/suffix guards, return True unconditionally, or collapse the explicit set check and the `startswith` into a single clear expression.

cartography/intel/gitlab/dependencies.py
```diff
@@ -143,66 +303,70 @@ def get_dependencies(
...
-            raise requests.exceptions.HTTPError(
-                "Unauthorized (401) - token may need 'api' or 'read_api' scope",
-                response=response,
-            )
+        if response.status_code in (401, 403):
+            _log_artifact_permission_warning(project_id, job_id, response.status_code)
+            return []
 
         response.raise_for_status()
         check_rate_limit_remaining(response)
```
jychp, line 332: This PR now silently fail on permissions errors, is it excepted ?

cartography/intel/gitlab/dependencies.py
```diff
@@ -143,66 +303,70 @@ def get_dependencies(
...
+            all_dependencies = _parse_cyclonedx_artifacts_zip(
+                response.content,
+                dependency_files,
+            )
+        except zipfile.BadZipFile:
+            all_dependencies = _parse_cyclonedx_artifact(
+                response.content,
+                (
+                    "gl-sbom.cdx.json.gz"
+                    if response.content.startswith(b"\x1f\x8b")
```
jychp, line 344: The inner `except zipfile.BadZipFile` is caught here and used for the magic-bytes fallback, but the outer `except zipfile.BadZipFile as e` at the end of the function body is now unreachable for this same exception because the inner handler already consumed it. The outer handler was the old error-logging path; it still exists in the code and will never trigger. Remove it to avoid misleading dead code.

jychp (approved):
Approved to unblock, but some suggestion to consider

cartography/intel/gitlab/dependencies.py
kunaals: good catch. fixed this to rely on the gitlab next-page header and avoid the extra empty request.

cartography/intel/gitlab/dependencies.py
kunaals: yes, this is expected here. artifact 401/403 is scoped to a single project job, so we log an actionable warning and continue instead of failing the whole gitlab sync. 404 still means no served artifact.

cartography/intel/gitlab/dependencies.py
kunaals: good catch. removed the dead outer badzip handler now that the direct-artifact fallback owns that case.

cartography/intel/gitlab/dependencies.py
kunaals: good call. simplified this into the explicit exact-name-or-hyphenated-sbom check.

PR #2758, opened 2026-05-06
fix(graph): apply conditional labels during ingestion

kunaals:
### Type of change
- [x] Documentation update

### Summary
Apply `ConditionalNodeLabel` declarations as part of the generated ingestion query instead of running a separate post-load label maintenance pass.

Conditional labels are part of the Cartography data model: module authors describe which semantic labels a node may carry, and the querybuilder turns that schema into Cypher. Previously, normal ingestion wrote the node first and then ran separate `MATCH ... REMOVE` / `MATCH ... SET` statements across all nodes matching the schema, optionally scoped through `sub_resource_relationship`.

That worked for many tenant-owned resource types, but it made conditional labels depend on a second graph-wide or scope-wide write after the actual node load. At larger graph sizes and higher ingestion concurrency, that extra sweep can touch many nodes outside the current batch and contend with concurrent node and relationship writes. In practice, this makes label maintenance more expensive and more fragile than the row-local node write that produced the relevant properties in the first place.

For shared or digest-keyed nodes, adding a fake ownership relationship just to scope the label query would be misleading, and leaving the query unscoped means one load can relabel nodes outside the current batch.

This change keeps the abstraction at the data-model layer and moves the execution to the row-local node in the generated `UNWIND` ingestion query:
(in one write transaction)
- remove all conditional labels declared by the schema from the loaded node;
- reapply only the conditional labels whose predicates match the node properties just written;
- keep normal modules declarative, without adding handwritten label-maintenance Cypher to intel modules.

The main design choice is to make conditional labels behave like the rest of node ingestion: the schema describes the desired state for the row being loaded, and the generated query updates that row in the same write. This is narrower, easier to reason about, and avoids relying on post-load sweeps for correctness.

### Related issues or links
N/A

### Breaking changes
None.

#### General

#### Proof of functionality

### Notes for reviewers
The standalone conditional-label query builder is left in place for explicit maintenance-query callers, but normal ingestion now applies conditional labels row-locally in the generated `UNWIND` query.


jychp:
Already landed in master

PR #2764, opened 2026-05-06
feat(tailscale): authenticate via OAuth client

rajsinghtech:
Cartography's Tailscale module currently authenticates with a long-lived API access token generated by a specific user. This PR lets it use an OAuth client instead, matching Tailscale's [recommended pattern for service integrations](https://tailscale.com/kb/1215/oauth-clients) (tag-scoped, not tied to a user, short-lived access tokens).

Two new flags:

```
--tailscale-oauth-client-id-env-var TS_OAUTH_CLIENT_ID
--tailscale-oauth-client-secret-env-var TS_OAUTH_CLIENT_SECRET
```

When both are set, cartography POSTs to `{base_url}/oauth/token` with `grant_type=client_credentials` and uses the returned access token as the `Bearer` for the rest of the sync. `--tailscale-token-env-var` still works; if both are configured, the OAuth client wins and a warning is logged.

### Breaking changes

None. Existing `--tailscale-token-env-var` deployments are unaffected.


docs/root/modules/tailscale/config.md
```diff
@@ -1,9 +1,32 @@
...
-1. If your have a self hosted instance, configure the API Url using `--tailscale-base-url` (default: `https://api.tailscale.com/api/v2`)
+- **OAuth client (recommended)** — tag-scoped, not tied to a user, exchanged at sync time for a short-lived bearer token. Matches Tailscale's [recommended pattern for service integrations](https://tailscale.com/kb/1215/oauth-clients).
+- **API access token** — long-lived, tied to a user account.
+
+In both cases, pass `--tailscale-org <tailnet-name>` (find it under [Settings → General](https://login.tailscale.com/admin/settings/general)). For self-hosted instances, set `--tailscale-base-url` (default `https://api.tailscale.com/api/v2`); the same base URL is used for the OAuth token endpoint.
+
+### OAuth client (recommended)
+
+1. Create an OAuth client at [Settings → OAuth clients](https://login.tailscale.com/admin/settings/oauth) with the read-only scopes you want cartography to ingest (typically `devices:core:read`, `users:read`, `users:read_all`, `dns:read`, `routes:read`, `policy_file:read`).
```
kunaals: this new oauth scope list looks off against the current tailscale scope names. i don’t see `users:read_all` or `routes:read` in the trust-credential scope list, and the existing sync also calls `/device/:id/attributes`, which needs `devices:posture_attributes:read`. users following these docs could mint a token successfully but then fail during sync.

cartography/intel/tailscale/__init__.py
```diff
@@ -28,23 +49,41 @@ def start_tailscale_ingestion(neo4j_session: neo4j.Session, config: Config) -> N
...
         total=5,
         backoff_factor=1,
         status_forcelist=[429, 500, 502, 503, 504],
-        allowed_methods=["GET"],
+        allowed_methods=["GET", "POST"],
     )
     api_session.mount("https://", HTTPAdapter(max_retries=retry_policy))
-    api_session.headers.update({"Authorization": f"Bearer {config.tailscale_token}"})
+
+    if has_oauth_client:
```
kunaals: this mints the oauth access token once for the whole tailscale run, but tailscale access tokens expire after one hour. large tailnets can spend enough time in the per-device attribute fetches to cross that boundary and start failing with 401s mid-sync. can we either refresh based on `expires_in` or retry once with a newly minted token on 401?

kunaals (dismissed):
sorry for the delay and thank you for the contribution! few small changes required

docs/root/modules/tailscale/config.md
rajsinghtech: good catch, the scope list was wrong. replaced it with the canonical names tied to the endpoints we actually hit:

- `devices:core:read` — /tailnet/:org/devices
- `devices:posture_attributes:read` — /device/:id/attributes
- `users:read` — /tailnet/:org/users
- `policy_file:read` — /tailnet/:org/acl
- `feature_settings:read` — /tailnet/:org/settings and /tailnet/:org/posture/integrations

dropped the bogus `users:read_all`/`routes:read` and `dns:read`

cartography/intel/tailscale/__init__.py
rajsinghtech: went with retry-once-on-401 rather than tracking `expires_in`

cartography/intel/tailscale/__init__.py
```diff
@@ -18,6 +19,66 @@
...
+            return response
+        logger.info("Tailscale returned 401; re-minting OAuth bearer and retrying.")
+        new_token = _mint_oauth_bearer(
+            api_session, base_url, client_id, client_secret,
+        )
+        api_session.headers["Authorization"] = f"Bearer {new_token}"
+        retried = response.request.copy()
+        retried.headers["Authorization"] = f"Bearer {new_token}"
+        retried.headers[_REAUTH_HEADER] = "1"
+        return api_session.send(retried)
```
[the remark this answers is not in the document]
rajsinghtech: good catch, forwarded `**kwargs` from the hook into `api_session.send` so timeout/verify/cert/etc carry through. test now asserts the kwargs round-trip. 6a82396

cartography/intel/tailscale/__init__.py
```diff
@@ -18,6 +19,66 @@
...
+    client_secret: str,
+) -> str:
+    response = api_session.post(
+        f"{base_url.rstrip('/')}/oauth/token",
+        data={
+            "grant_type": "client_credentials",
+            "client_id": client_id,
+            "client_secret": client_secret,
+        },
+        headers={_REAUTH_HEADER: "1"},
```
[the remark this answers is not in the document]
rajsinghtech: fixed — `_mint_oauth_bearer` now sets `Authorization: None` in the per-request headers, which removes the session-level bearer for the token call. test asserts that. 6a82396

kunaals (approved):

PR #2773, opened 2026-05-07
fix(gcp): reduce policy binding write pressure

kunaals:
### Type of change

### Summary
Reduces Neo4j write pressure from GCP policy binding syncs when multiple projects are processed concurrently.

This keeps the existing GCPPolicyBinding and APPLIES_TO graph model intact while making the write-heavy phase less bursty:
- limits concurrent policy binding graph load/cleanup work within a process
- uses smaller batches for policy binding node and APPLIES_TO relationship writes
- replaces per-target-label APPLIES_TO cleanup with one scoped relationship cleanup query

#### General

#### Proof of functionality

### Notes for reviewers
This intentionally avoids changing GCPPolicyBinding identity or APPLIES_TO semantics. The goal is to reduce concurrent write contention while preserving existing traversals and cleanup scope.

### DIFF
--- a/cartography/intel/gcp/policy_bindings.py
+++ b/cartography/intel/gcp/policy_bindings.py
@@ -526,6 +526,9 @@ def _cleanup_applies_to_relationships(
     project_id: str,
     update_tag: int,
 ) -> None:
+    # APPLIES_TO can point at several resource labels. Clean up by relationship
+    # scope directly instead of running the same stale-edge cleanup once per
+    # possible target label.
     GraphStatement(
         """
         MATCH (:GCPPolicyBinding)-[r:APPLIES_TO]->()
--- a/cartography/intel/gcp/policy_bindings.py
+++ b/cartography/intel/gcp/policy_bindings.py
@@ -3,6 +3,7 @@ import logging
 import time
 from dataclasses import dataclass
 from enum import Enum
+from threading import BoundedSemaphore
 from threading import Lock
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


cartography/intel/gcp/policy_bindings.py
```diff
@@ -505,12 +514,41 @@ def load_bindings(
...
+def _cleanup_applies_to_relationships(
+    neo4j_session: neo4j.Session,
+    project_id: str,
+    update_tag: int,
+) -> None:
+    # APPLIES_TO can point at several resource labels. Clean up by relationship
+    # scope directly instead of running the same stale-edge cleanup once per
+    # possible target label.
+    GraphStatement(
+        """
```
kunaals, line 533: we changed the `applies_to` cleanup to delete stale edges by project scope instead of doing one cleanup pass per possible target label

each `applies_to` edge has scope fields on it: `_sub_resource_label = "GCPProject"`, `_sub_resource_id = <project_id>`, and `lastupdated`. so cleanup can just find stale `(:GCPPolicyBinding)-[:APPLIES_TO]->()` edges for that project and delete them in small batches

same behavior, less repeated graph work. this matters because policy bindings can apply to lots of different resource types, so the old label-by-label cleanup made us scan for the same project’s stale edges many times

kunaals: im not sure if there is a better way to do this but if there is please let me know

jychp: The cleanup semantics here are fine; this query is functionally equivalent to running `from_matchlink` once per label, with the bonus of also catching stale APPLIES_TO edges pointing to labels that got dropped from `_FULL_NAME_MAPPINGS`. The deviation worth flagging is from the declarative pattern that AGENTS.md endorses ("prefer GraphJob for cleanup"). The cleaner long-term fix is to extend `GraphJob.from_matchlink` (or add a sibling) to support a label-agnostic target node so this can stay schema-driven. If you keep the handwritten query, drop a short code comment explaining why the schema helper isn't a fit.

jychp (approved):
LGTM

PR #2776, opened 2026-05-08
feat(ontology): add blockstorage, identityproviders, cicdpipelines categories

jychp:
### Type of change

### Summary
Add three new ontology semantic labels covering categories that were modelled in cartography but not yet unified across providers:

- **`BlockStorage`**: EBSVolume (AWS), AzureDisk (Azure), ScalewayVolume (Scaleway) — fields `_ont_name`, `_ont_size_gb`, `_ont_encrypted`, `_ont_region`, `_ont_state`. Enables cross-cloud volume encryption posture queries (e.g. CIS / EBS encryption checks).
- **`IdentityProvider`**: AWSSAMLProvider (AWS), KubernetesOIDCProvider (Kubernetes), KeycloakIdentityProvider (Keycloak) — fields `_ont_name`, `_ont_protocol`, `_ont_issuer`, `_ont_enabled`. Surfaces federation trust relationships across providers for cross-platform posture queries.
- **`CICDPipeline`**: CodeBuildProject (AWS), GitHubWorkflow (GitHub), GitLabCIConfig (GitLab), AzureDataFactoryPipeline (Azure), SpaceliftStack (Spacelift) — fields `_ont_name`, `_ont_type` (`build`/`deploy`/`iac`), `_ont_status`. Unified inventory of build, deploy, and IaC pipelines across platforms for supply-chain visibility.

A few small node-schema additions are included to support full coverage where the upstream API exposes the field but cartography did not yet store it:
- `AzureDisk.state` (from Azure managed-disk `disk_state`).
- `ScalewayVolume.size_gb` (derived in the transform from byte-valued `size`; original `size` retained).
- `AWSSAMLProvider.name` (parsed from the ARN suffix; the IAM `list_saml_providers` API does not return a friendly name).
- `CodeBuildProject.name` (already present in the boto3 `BatchGetProjects` response, simply added to the transform output).

For `CICDPipeline` we deliberately scope to *pipeline definitions* and not their components or executions: `GitHubAction` (a step component referenced inside a workflow) and `SpaceliftRun` (an execution of a Stack) are intentionally excluded. `GitLabCIConfig` is included as the GitLab analog of `GitHubWorkflow` to round out coverage.

### Related issues or links

N/A

### How was this tested?

- `make test_lint` (pre-commit) passes locally on the changed files.
- `make test_unit` passes locally (1899 tests). The existing assertions in `tests/unit/cartography/intel/ontology/test_ontology_mapping.py` automatically validate every new mapping: that each `node_field` exists on the schema, that `static_value` / `equal_boolean` declarations are well-formed, and that module names are valid `TOP_LEVEL_MODULES`.

### Test plan
- [ ] Run a sync that exercises EBS / Azure Disk / Scaleway volumes and verify:
  ```cypher
  MATCH (n:BlockStorage)
  RETURN n._ont_source, count(n), avg(n._ont_size_gb), sum(CASE WHEN n._ont_encrypted THEN 1 ELSE 0 END) AS encrypted_count
  ```
  Expect rows from at least two `_ont_source` values, with `_ont_encrypted` populated for EBS and Azure.
- [ ] Run a sync that exercises AWS SAML / K8s OIDC (EKS) / Keycloak and verify:
  ```cypher
  MATCH (n:IdentityProvider) RETURN n._ont_source, n._ont_name, n._ont_protocol, n._ont_issuer, n._ont_enabled
  ```
  Expect at least three protocols / sources represented; healthy EKS providers should show `_ont_enabled = true` (matched against the uppercase `ACTIVE` AWS returns).
- [ ] Run a sync that exercises CodeBuild / GitHub / GitLab / Azure Data Factory / Spacelift and verify:
  ```cypher
  MATCH (n:CICDPipeline) RETURN n._ont_source, n._ont_type, count(n)
  ```
  Expect rows for `build`, `deploy`, and `iac` types from the corresponding sources.

#### General
- [x] I have added/updated tests that prove my fix is effective or my feature works. (Existing ontology mapping unit tests cover the new mappings on import; no new tests needed.)

#### Proof of functionality
- [x] New or updated unit/integration tests. (Mapping tests exercise the new categories.)

- [x] Updated the schema documentation: central `docs/root/modules/ontology/schema.md` (sections + mermaid diagram for the three new labels) and the touched provider schema docs.

### Notes for reviewers

- Each of the three feature commits is self-contained and tied to one category — easy to review or revert independently. Two follow-up commits address review feedback: a fix for the EKS uppercase `ACTIVE` status casing in the `IdentityProvider.enabled` mapping, and the schema doc updates.


cartography/models/ontology/mapping/data/blockstorage.py
```diff
@@ -0,0 +1,75 @@
...
+    module_name="azure",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="AzureDisk",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="name", required=True
+                ),
+                OntologyFieldMapping(ontology_field="size_gb", node_field="disksizegb"),
+                OntologyFieldMapping(
```
kunaals: i think this is mixing two different Azure disk encryption concepts. `_ont_encrypted` is documented as encryption at rest, but `encryption_settings_collection.enabled` is the older Azure Disk Encryption-style setting, while managed disks are encrypted at rest by default with SSE. This would make normal Azure managed disks look unencrypted/null in cross-cloud volume posture queries. Can we either leave Azure `_ont_encrypted` unmapped until we store the SSE/DES posture, or split this into a more specific ontology field?

cartography/models/ontology/mapping/data/identityproviders.py
```diff
@@ -0,0 +1,92 @@
...
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="name", required=True
+                ),
+                OntologyFieldMapping(
+                    ontology_field="protocol",
+                    node_field="",
+                    special_handling="static_value",
+                    extra={"value": "SAML"},
+                ),
+                OntologyFieldMapping(ontology_field="issuer", node_field="arn"),
```
kunaals: i don’t think the ARN is the right value for `_ont_issuer`. For Kubernetes this field is the actual OIDC issuer URL, but for AWS SAML this would be the AWS-local IAM provider resource id, not the SAML issuer/entity ID. AWS exposes the real IdP metadata through `GetSAMLProvider`; if we don’t parse that yet, i think it’s better to leave `_ont_issuer` unset than make cross-provider issuer queries compare two different concepts.

cartography/models/ontology/mapping/data/cicdpipelines.py
```diff
@@ -0,0 +1,123 @@
...
+            ],
+        ),
+    ],
+)
+
+azure_mapping = OntologyMapping(
+    module_name="azure",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="AzureDataFactoryPipeline",
```
kunaals: i’d push back on labeling Azure Data Factory pipelines as `CICDPipeline`. ADF pipelines are data movement/transformation workflows, not CI/CD pipeline definitions like GitHub workflows or GitLab pipeline configs. From a security graph perspective this will make CI/CD inventory queries include ETL/data workflows as deploy pipelines. Can we drop ADF from this label, or rename the abstraction to something broader if the intent is generic automation/workflow inventory?

kunaals (changes requested):

kunaals (approved):

PR #2786, opened 2026-05-08
feat(rules): expand ISO 27001 coverage

jychp:
### Type of change
- [ ] Documentation update

### Summary
Adds additional ISO 27001 Annex A mappings to existing Cartography rules and expands rules coverage for ISO-aligned controls.

This PR adds:
- CIS AWS 6.1.2 coverage for CIFS/SMB port 445 exposure using existing EC2 security group data.
- ISO 27001 metadata on existing generic rules for compute/database/object-storage exposure, identity and policy administration, MFA, lifecycle, supply-chain, and monitoring gaps.
- New non-hyperscaler security rules for Cloudflare, LastPass, JumpCloud, Duo, Tailscale, Jamf, and Tailscale device posture data.
- `Module.JUMPCLOUD` so JumpCloud facts can be represented directly in rule metadata.

### Related issues or links
- Fixes #

### Breaking changes
None.

### How was this tested?
- `UV_CACHE_DIR=/private/tmp/codex-uv-cache uv run --frozen pytest tests/unit/rules/test_cis_aws.py tests/unit/rules/test_non_hyperscaler_security.py tests/unit/rules/test_iso27001_mappings.py tests/unit/rules/test_runners.py tests/unit/rules/test_cli.py`
- Ran `EXPLAIN` against local Neo4j on all newly added rule queries, visual queries, and count queries: 45 checked, 0 failures.

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
This PR only adds rule metadata and rule queries. It does not add or modify synced entities, node schemas, or relationships beyond adding the JumpCloud enum value used by rules metadata.


cartography/rules/data/rules/database_instance_exposed.py
```diff
@@ -210,4 +211,8 @@ class DatabaseInstanceExposed(Finding):
         "stride:tampering",
```
kunaals, line 211: i don't think `8.24` is defensible on this rule as written. this detects public reachability, but annex a `8.24` is use of cryptography, and there's no evidence here about encryption at rest, encryption in transit, or key management. i'd keep `8.20` and drop `8.24` unless we add an actual crypto check.

cartography/rules/data/rules/cis_kubernetes_workloads.py
```diff
@@ -89,13 +90,8 @@ class SecretsInEnvVarsOutput(Finding):
```
kunaals, line 88: i don't think `8.22` fits here. this rule is about pods using the kubernetes `default` namespace, while annex a `8.22` is segregation of networks. namespace hygiene / policy separation feels like a different control concept.

cartography/rules/data/rules/tailscale_security_configuration_gaps.py
```diff
@@ -0,0 +1,191 @@
...
+    facts=(
+        _tailscale_device_approval_disabled,
+        _tailscale_user_approval_disabled,
+        _tailscale_network_flow_logging_disabled,
+        _tailscale_device_auto_updates_disabled,
+        _tailscale_device_key_expiry_disabled,
+    ),
+    tags=("network", "device", "logging", "compliance", "stride:spoofing"),
+    version="0.1.0",
+    frameworks=(
```
kunaals: this rule feels too broad to carry one rule-level `frameworks` tuple. `Rule.frameworks` applies to every fact here, so `device_key_expiry_disabled` and `device_auto_updates_disabled` will also claim `8.15`, while the logging fact will claim `8.8`, etc. i think this needs to be split into narrower rules, or the iso mappings need to live only on rules that each represent one control concept.

kunaals (changes requested):

kunaals (approved):

PR #2788, opened 2026-05-09
feat(guardduty): Persist Service Action API fields on AWSGuardDutyFindings

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- This PR extends the AWSGuardDutyFinding node to persist the full Service.Action.AwsApiCallAction block. It also connects a GuardDuty finding to a remote account that triggered the finding if that account is within the same AWS enviornment. 

**New Rel** 
- AWSGuardDutyFinding----REMOTE_ACCOUNT--->AWSAccount

### How was this tested?

image

**New Fields on Node**
image

#### General

#### Proof of functionality


cartography/intel/aws/guardduty.py
```diff
@@ -170,6 +170,19 @@ def transform_findings(findings: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
         service = f.get("Service", {})
+        action = service.get("Action", {})
+        action_type = action.get("ActionType")
+
+        # Extract AwsApiCallAction fields when applicable
+        api_call: dict[str, Any] = {}
+        if action_type == "AWS_API_CALL":
+            api_call = action.get("AwsApiCallAction", {})
+
+        remote_ip_details = api_call.get("RemoteIpDetails", {})
```
kunaals: i think this should use `or {}` for the nested optional dicts. `dict.get("RemoteIpDetails", {})` only falls back when the key is missing, not when AWS returns the key with `null`, so a payload like `{"Country": null}` would make the later `.get("CountryName")` raise and fail the whole finding batch. something like `(api_call.get("RemoteIpDetails") or {})` and `(remote_ip_details.get("Country") or {})` would make this more defensive.

tests/integration/cartography/intel/aws/test_guardduty.py
```diff
@@ -258,3 +260,136 @@ def test_sync_guardduty_findings(
...
+    mock_get_findings,
+    mock_get_detector_details,
+    mock_get_detectors,
+    neo4j_session,
+):
+    """Test that AWS_API_CALL findings persist richer action fields and REMOTE_ACCOUNT."""
+    boto3_session = MagicMock()
+    api_call_update_tag = 987654321
+    create_test_account(neo4j_session, TEST_ACCOUNT_ID, api_call_update_tag)
+    create_test_account(neo4j_session, REMOTE_ACCOUNT_ID, api_call_update_tag)
```
kunaals, line 292: can we add a case where the remote account is not pre-created? that is probably the more important production shape: GuardDuty gives us `RemoteAccountDetails.AccountId`, but the external account is not in the graph. the test should assert the `api_call_remote_account_id` property still persists on the finding, and only the `REMOTE_ACCOUNT` edge is absent.

cartography/models/aws/guardduty/findings.py
```diff
@@ -31,6 +31,30 @@ class GuardDutyFindingNodeProperties(CartographyNodeProperties):
...
+    api_call_remote_ip: PropertyRef = PropertyRef("api_call_remote_ip")
+    api_call_remote_country: PropertyRef = PropertyRef("api_call_remote_country")
+    api_call_remote_city: PropertyRef = PropertyRef("api_call_remote_city")
+    api_call_remote_org: PropertyRef = PropertyRef("api_call_remote_org")
+    api_call_remote_asn: PropertyRef = PropertyRef("api_call_remote_asn")
+    api_call_remote_asn_org: PropertyRef = PropertyRef("api_call_remote_asn_org")
+    api_call_remote_isp: PropertyRef = PropertyRef("api_call_remote_isp")
+    api_call_remote_lat: PropertyRef = PropertyRef("api_call_remote_lat")
+    api_call_remote_lon: PropertyRef = PropertyRef("api_call_remote_lon")
+    api_call_remote_account_id: PropertyRef = PropertyRef(
```
kunaals, line 52: since this field may be the only durable way to find external remote accounts when there is no matching `AWSAccount` node, should this be `extra_index=True`? the relationship is useful when the account exists in the graph, but for the scarier external-account case users will likely query this property directly.

kunaals (changes requested):

kunaals (approved):

PR #2793, opened 2026-05-11
feat(aws): add ALB mTLS properties to ELBV2Listener

jychp:
### Type of change

### Summary

ALB / NLB listener nodes in Cartography currently expose only `port`, `protocol`, `ssl_policy`, and `targetgrouparn`. They do not surface the `MutualAuthentication` block that AWS returns from `DescribeListeners`, so there is no way to query whether mutual TLS is enforced on a listener.

This PR adds four additive properties to the existing `ELBV2Listener` node, populated from the listener's `MutualAuthentication` object:

| AWS field                       | Node property                      |
| ------------------------------- | ---------------------------------- |
| `Mode`                          | `mutual_authentication_mode`       |
| `TrustStoreArn`                 | `trust_store_arn`                  |
| `IgnoreClientCertificateExpiry` | `ignore_client_certificate_expiry` |
| `AdvertiseTrustStoreCaNames`    | `advertise_trust_store_ca_names`   |

Listeners without `MutualAuthentication` get `null` for all four, mirroring the existing behavior of `ssl_policy` and `targetgrouparn` on plain HTTP listeners. No new node, no new relationship, no migration.

### Related issues or links

(none)

### How was this tested?

- Extended the existing `tests/data/aws/ec2/load_balancer_v2s.py` fixture so the HTTPS test listener carries a `MutualAuthentication` block; HTTP and TLS listeners stay bare to cover the "no mTLS" path.
- Extended the existing `check_nodes(..., "ELBV2Listener", [...])` assertion in `tests/integration/cartography/intel/aws/ec2/test_load_balancer_v2s.py` to check the four new properties on all three test listeners.
- `make test_lint` passes.
- `uv run pytest tests/integration/cartography/intel/aws/ec2/test_load_balancer_v2s.py` passes (4/4).

#### General

#### Proof of functionality

- [ ] Updated the schema README. (No new node or relationship; only additive properties on an existing node, table-only.)

### Test plan

- [x] `make test_lint` passes
- [x] `pytest tests/integration/cartography/intel/aws/ec2/test_load_balancer_v2s.py` passes
- [ ] Against a synced dev DB, `MATCH (l:ELBV2Listener) WHERE l.mutual_authentication_mode IS NOT NULL RETURN l` returns listeners with mTLS enabled

### DIFF
--- a/cartography/intel/aws/ec2/load_balancer_v2s.py
+++ b/cartography/intel/aws/ec2/load_balancer_v2s.py
@@ -226,6 +226,7 @@ def _transform_load_balancer_v2_data(
 
         # Extract listener data
         for listener in lb.get("Listeners", []):
+            mutual_auth = listener.get("MutualAuthentication") or {}
             listener_data.append(
                 {
                     "ListenerArn": listener["ListenerArn"],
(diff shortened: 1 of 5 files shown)


kunaals (approved):
lgtm

PR #2794, opened 2026-05-11
feat(rules): normalize Kubernetes control-plane exposure and add multi-cloud rule

jychp:
### Type of change

### Summary

Adds a normalized `_ont_control_plane_public_access` field on the `ComputeCluster` ontology label across EKS, GKE, and AKS, plus a new multi-cloud rule `kubernetes_control_plane_exposed` that flags managed Kubernetes clusters whose API server is reachable from the public internet. Mirrors the structure of `serverless_workload_exposed`.

Why: a narrow `eks_public_endpoint` rule would only cover AWS and miss the equivalent control on GKE and AKS. Normalizing the signal at the ontology layer lets a single rule cover all three providers, closes the CIS EKS Benchmark 5.4.2 gap, and extends the same control to GCP and Azure.

Key design notes:
- **EKS**: maps directly from `endpoint_public_access`. The EKS fact also requires `exposed_internet = true` to preserve the existing `aws_eks_asset_exposure.json` semantics.
- **GKE**: derives from `privateClusterConfig.enablePrivateEndpoint` via `invert_boolean`. This is strictly safer than gating on `privateClusterConfig.publicEndpoint`, which is null for default (non-private) GKE clusters even though those are very much internet-reachable.
- **AKS**: now captures `apiServerAccessProfile.enablePrivateCluster` from the existing `managed_clusters.list()` response (no new API call). The new `api_server_public_access` property feeds the ontology mapping.
- **Self-managed `KubernetesCluster`**: no exposure signal available; left unset.
- **Known caveat (matches AWS EKS convention)**: clusters restricted via `masterAuthorizedNetworksConfig` (GKE), `authorizedIpRanges` (AKS), or public-access CIDRs (EKS) are still flagged.

### Related issues or links

N/A

### How was this tested?

- `make test_lint` — passes clean.
- `uv run --frozen pytest -vvv tests/unit/rules/test_kubernetes_control_plane_exposed.py` — 5 new tests pass.
- `uv run --frozen pytest -vvv tests/unit/cartography/intel/ontology/ tests/unit/cartography/graph/test_ontology_fields.py` — 37 tests pass (ontology framework still happy with the new field).
- `uv run --frozen pytest -vvv tests/integration/cartography/intel/azure/test_aks.py` — AKS integration covers both public (`enable_private_cluster: false`) and private (`enable_private_cluster: true`) cluster fixtures, asserting the new `api_server_public_access` value on each.
- `uv run --frozen pytest -vvv tests/integration/cartography/intel/aws/test_eks.py tests/integration/cartography/intel/gcp/test_gke.py` — EKS/GKE load paths still pass with the new ontology mapping.

#### General

#### Proof of functionality

- [x] Updated the schema documentation (`docs/root/modules/azure/schema.md` for the new AKS `api_server_public_access` property).

### Test plan

- [ ] CI lint + unit + integration suites pass.
- [ ] Against a populated Neo4j: `MATCH (c:ComputeCluster) WHERE c._ont_control_plane_public_access = true RETURN labels(c), c.id, c.name` returns EKS/GKE/AKS clusters with publicly-reachable control planes.
- [ ] Run the rule cypher and confirm rows for at least one EKS, one GKE, and one AKS cluster.


tests/integration/cartography/intel/azure/test_aks.py
```diff
@@ -46,20 +46,22 @@ def test_sync_aks(mock_get_clusters, mock_get_pools, neo4j_session):
...
     # Assert Clusters
+    public_cluster_id = MOCK_CLUSTERS[0]["id"]
+    private_cluster_id = MOCK_CLUSTERS[1]["id"]
     expected_nodes = {
-        (
-            "/subscriptions/00-00-00-00/resourceGroups/TestRG/providers/Microsoft.ContainerService/managedClusters/my-test-aks-cluster",
-            "my-test-aks-cluster",
-        ),
+        (public_cluster_id, "my-test-aks-cluster", True),
+        (private_cluster_id, "my-private-aks-cluster", False),
```
kunaals, line 54: i think we should add one fixture for the `publicNetworkAccess=Disabled` path here. the transform now has two independent gates, but this test only exercises `enable_private_cluster`; a field-name/value regression would still pass while reintroducing the AKS false positive this PR is trying to avoid.

kunaals (approved):

PR #2799, opened 2026-05-12
feat(gcp): add Workload Identity Federation pool and provider nodes

jychp:
### Type of change

### Summary
GCP Workload Identity Federation (WIF) pools and providers were not modeled. Their absence meant `GCPPolicyBinding.members` entries of the form `principal://iam.googleapis.com/.../workloadIdentityPools/{POOL}/...` and `principalSet://...` were silently dropped by the member parser, leaving federated identity grants invisible in the graph and breaking cross-cloud lateral-movement analysis for any GCP project that uses WIF.

This PR:

- Adds `GCPWorkloadIdentityPool` and `GCPWorkloadIdentityProvider` node schemas under `cartography/models/gcp/workload_identity.py`, both sub-resources of `GCPProject`. Provider has a `MEMBER_OF` relationship to its pool and carries the `IdentityProvider` ontology label.
- Adds `cartography/intel/gcp/workload_identity.py`, a new intel module that lists pools and providers via the IAM v1 discovery client (`iam.googleapis.com`). The sync is gated by the existing IAM enablement check (`Services.iam`), so no new entry in the `Services` namedtuple is needed.
- Teaches the policy binding member parser to recognize `principal://` and `principalSet://` URIs and emit the referenced pool resource name into a new `wif_pools` field on the binding node.
- Adds a `HAS_ALLOW_POLICY` edge from `GCPWorkloadIdentityPool` to `GCPPolicyBinding`, mirroring the existing principal-to-binding edge label so federated identity grants are traversable the same way native principal grants are.
- Maps `GCPWorkloadIdentityProvider` into the existing `identityproviders` ontology dataset, alongside `AWSSAMLProvider`, `KubernetesOIDCProvider`, and `KeycloakIdentityProvider`.
- Documents both new nodes and the new binding fields/edges in `docs/root/modules/gcp/schema.md`.

### Related issues or links
- Closes a coverage gap for GCP federated identity in the knowledge graph.

### How was this tested?
- `make test_lint` (pre-commit) passes.
- `uv run --frozen pytest tests/integration/cartography/intel/gcp/ tests/unit/cartography/intel/gcp/` — 342 tests pass, including the two new ones in `tests/integration/cartography/intel/gcp/test_workload_identity.py`:
  - `test_sync_workload_identity_pools_and_providers` exercises the full sync against mocked pool/provider list responses and asserts pool/provider nodes, `RESOURCE` and `MEMBER_OF` relationships, and the `IdentityProvider` ontology label.
  - `test_transform_bindings_extracts_wif_pools` exercises the policy binding parser end to end on a mix of `serviceAccount:` and WIF `principal://` / `principalSet://` members.

### Test plan
- [ ] Verify `(GCPProject)-[:RESOURCE]->(:GCPWorkloadIdentityPool)<-[:MEMBER_OF]-(:GCPWorkloadIdentityProvider)` after a real sync.
- [ ] Verify `(:GCPPolicyBinding)-[:HAS_ALLOW_POLICY]->(:GCPWorkloadIdentityPool)` is created for bindings whose members include `principal[Set]://...workloadIdentityPools/...` URIs.
- [ ] Verify `IdentityProvider` label + `_ont_name`, `_ont_protocol`, `_ont_issuer`, `_ont_enabled` are populated on `GCPWorkloadIdentityProvider` nodes.
- [ ] Verify cleanup correctly removes pools and providers no longer returned by the API.
- [ ] Verify projects that do not have any WIF pools complete sync without errors.

#### General
- [x] I have added/updated tests that prove the feature works.

#### Proof of functionality
- [x] New unit/integration tests.

- [ ] Sync logs from a real environment will be added before merge.

- [x] Updated the schema documentation (`docs/root/modules/gcp/schema.md`).


cartography/intel/gcp/workload_identity.py
```diff
@@ -0,0 +1,281 @@
...
+
+@timeit
+def load_pools(
+    neo4j_session: neo4j.Session,
+    pools: List[Dict[str, Any]],
+    project_id: str,
+    gcp_update_tag: int,
+) -> None:
+    if not pools:
+        return
```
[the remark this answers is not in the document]
jychp: Intentional design: this is fail-safe behavior, not a convergence bug.

If the caller loses `iam.workloadIdentityPools.list` mid-life (rotated service account, role drift, project IAM edit, etc.), running cleanup on the empty response would silently delete every `GCPWorkloadIdentityPool` / `GCPWorkloadIdentityProvider` node in the graph, breaking downstream policy-binding traversals and attack-path queries. We prefer stale-but-correct over destructive-on-permission-hiccup, same stance as the IAM-key cleanup pattern at `cartography/intel/gcp/iam.py` and the per-pool provider failure path right next to this code (`providers_sync_complete` flag).

Persistently revoked permissions are a documented misconfiguration (`docs/root/modules/gcp/config.md` now calls out the WIF permissions required); once they are restored, the next successful sync converges the graph. A fresh install simply ingests nothing for WIF until permissions are granted, which is the desired behavior.

cartography/intel/gcp/workload_identity.py
```diff
@@ -0,0 +1,317 @@
...
+
+    all_providers: List[Dict[str, Any]] = []
+    providers_sync_complete = True
+    for pool in raw_pools:
+        pool_name = pool.get("name")
+        if not pool_name:
+            continue
+        try:
+            raw_providers = get_workload_identity_providers(iam_client, pool_name)
+        except HttpError as e:
```
kunaals, line 318: i tested this against a live gcp env and gke-managed workload identity pools show up in workloadIdentityPools.list with mode=SYSTEM_TRUST_DOMAIN, but providers.list on those pools returns a deterministic 400 INVALID_ARGUMENT because ListWorkloadIdentityPoolProviders is not supported for that resource. this catch treats that as a partial provider sync and disables provider cleanup for the whole project, so any project with a gke .svc.id.goog pool can keep stale GCPWorkloadIdentityProvider nodes forever. can we skip provider listing for SYSTEM_TRUST_DOMAIN pools, or classify that specific 400 as not-applicable rather than incomplete?

cartography/intel/gcp/workload_identity.py
```diff
@@ -0,0 +1,317 @@
...
+                "state": pool.get("state"),
+                "disabled": pool.get("disabled", False),
+                "sessionDuration": pool.get("sessionDuration"),
+                "projectId": project_id,
+            },
+        )
+    return result
+
+
+def _detect_protocol(provider: Dict[str, Any]) -> str | None:
```
kunaals, line 118: google's current workload identity provider API also supports x509 providers, but _detect_protocol only handles oidc/aws/saml, so x509 providers would load with protocol=None and weak ontology data. even if we do not model the full trust store yet, i think we should at least set protocol="X509" and cover it with a transform test.

kunaals (changes requested):

kunaals (approved):

PR #2801, opened 2026-05-12
fix(rules): reduce k8s service account token noise

kunaals:
### Type of change

### Summary
This refines the CIS Kubernetes 5.1.6 service account token mount rule so it stays focused on workload cases that security engineers should review.

The rule still reports pods where service account tokens are effectively mounted, including inherited Kubernetes defaults, but it now suppresses expected token-use cases that fit the benchmark's intent:

- default service accounts, because default service account hygiene is covered separately by CIS Kubernetes 5.1.5
- service accounts in infrastructure namespaces documented by upstream Kubernetes or project installation guides
- service accounts with documented infrastructure controller names, so controller installs in non-default namespaces do not create noisy workload findings
- AWS IAM roles for service accounts, where the Kubernetes service account token is the intended web identity federation mechanism
- GKE Workload Identity Federation service accounts, where the Kubernetes service account identity is intentionally used for cloud API access

The finding output also prefers the ontology service account name when present, falling back to the raw Kubernetes service account name.

### Related issues or links
- Kubernetes service account token automount behavior: https://kubernetes.io/docs/concepts/security/service-accounts/
- Kubernetes initial namespaces: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
- Amazon EKS IAM roles for service accounts: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
- GKE Workload Identity Federation: https://cloud.google.com/kubernetes-engine/docs/concepts/workload-identity
- AWS Load Balancer Controller installation: https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/installation/
- ingress-nginx install namespace: https://kubernetes.github.io/ingress-nginx/deploy/
- Kyverno installation namespace: https://kyverno.io/docs/installation/
- Kubernetes Vertical Pod Autoscaling components: https://kubernetes.io/docs/concepts/workloads/autoscaling/vertical-pod-autoscale/

### Breaking changes
None.

### How was this tested?
- uv run pytest tests/unit/rules/test_cis_kubernetes.py
- uv run pre-commit run --files cartography/rules/data/rules/cis_kubernetes_workloads.py tests/unit/rules/test_cis_kubernetes.py
- Manually compared before/after rule counts against several live graph snapshots to confirm the filter suppresses default service account, infrastructure namespace, documented controller service account, IRSA, and GKE Workload Identity cases while preserving non-infrastructure workload findings.

#### General
- [x] The linter passes locally (focused pre-commit hooks).

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers
The namespace and controller service account allowlists are intentionally conservative and tied to public upstream or project documentation. The goal is to remove expected platform/controller token usage from this workload finding while keeping application and vendor workload service accounts visible for review.


cartography/rules/data/rules/cis_kubernetes_workloads.py
```diff
@@ -118,27 +155,57 @@ class ServiceAccountTokenMountOutput(Finding):
...
+      AND NOT (
+        service_account_name = 'default'
+        OR service_account_namespace IN {K8S_INFRASTRUCTURE_SERVICE_ACCOUNT_NAMESPACES_CYPHER}
+        OR service_account_name IN {K8S_INFRASTRUCTURE_SERVICE_ACCOUNT_NAMES_CYPHER}
+        OR service_account_assumes_aws_role
+        OR service_account_assumes_gcp_identity
+      )
     RETURN *
     """,
     cypher_count_query="""
```
jychp: blocking: cypher_count_query (immediately below) was not updated to match the new filter logic. It still excludes only three namespaces (kube-system, kube-public, kube-node-lease), while the main query now also excludes all infrastructure namespaces, named controller SAs, IRSA workloads, and GKE Workload Identity workloads. runners.py uses this count as total_assets to compute passing = total_assets - failing, which feeds the compliance percentage. The stale denominator inflates the passing count for any cluster running IRSA, istio, karpenter, or ingress-nginx. Update the count query to count the same candidate population as the main query.

cartography/rules/data/rules/cis_kubernetes_workloads.py
```diff
@@ -29,6 +29,43 @@
...
+    "istio-system",
+    "karpenter",
+    "kube-node-lease",
+    "kube-public",
+    "kube-system",
+    "kyverno",
+    "tigera-operator",
+)
+
+K8S_INFRASTRUCTURE_SERVICE_ACCOUNT_NAMESPACES_CYPHER = repr(
```
jychp: suggestion: repr(list(...)) generates a Cypher list literal only by coincidence. Python repr() uses single-quoted strings and square brackets, which happens to match Cypher list syntax. If any entry ever contains a single quote, repr() escapes it as \\' which is not valid Cypher. Use an explicit construction instead: "[" + ", ".join(f"'{ns}'" for ns in K8S_INFRASTRUCTURE_SERVICE_ACCOUNT_NAMESPACES) + "]" Apply the same fix to K8S_INFRASTRUCTURE_SERVICE_ACCOUNT_NAMES_CYPHER. No current values contain quotes, so this is not blocking today.

jychp (changes requested):

cartography/rules/data/rules/cis_kubernetes_workloads.py
kunaals: good catch. fixed this by applying the same candidate exclusions in the count query.

cartography/rules/data/rules/cis_kubernetes_workloads.py
kunaals: good call. replaced the repr usage with explicit escaped cypher string literals.

jychp (approved):

PR #2802, opened 2026-05-12
fix(rules): reduce false positives in top-volume security rules

jychp:
### Type of change

### Summary

Audit of the highest-finding-count rules surfaced a mix of false positives, finding-granularity issues, and missing utilization signal. This PR fixes the five worst offenders so the top-volume rules emit only actionable findings.

- `cis_aws_6_1_1_ebs_encryption`: require `encrypted = false` explicitly. Treating `encrypted IS NULL` as `false` was the source of ~88% false positives — most `null` rows resolved to volumes that no longer exist in the graph.
- `policy_administration_privileges`: collapse to one finding per `(principal, policy)` across AWS / GCP / Azure. Previously a single principal+policy could explode into 21 rows. `action: str` / `resource: str` become `actions: list[str]` / `resources: list[str]` (mirrors `IdentityAdministrationPrivileges`).
- `identity_administration_privileges` (AWS): tighten grouping to `(principal, policy)` only, and drop rows whose only matched action is `iam:CreateServiceLinkedRole` — a scoped action included in `PowerUserAccess`, not real identity-admin capability.
- `cis_aws_6_5_default_sg_traffic`: emit one finding per default SG, not one per rule entry per direction (~976 -> ~393 findings). Adds `has_inbound_rules`, `has_egress_rules`, and `in_use` flags so unused-VPC defaults can be filtered or downgraded without losing CIS coverage. The previous query also used a non-existent `IpPermissionEgress` label for AWS egress rules; the new query matches `AWSIpRule` excluding inbound.
- `cis_gcp_3_8_vpc_flow_logs`: add `in_use` flag derived from any non-VPC consumer attached to the subnet (GCP instance via NIC, forwarding rule, etc.) so the auto-created `PRIVATE` defaults in unused projects can be filtered without breaking compliance reporting.

### Related issues or links

N/A

### How was this tested?

- `make test_lint` clean (pre-commit on the four modified files).
- `pytest tests/unit/rules/ -q` — 587 passed.
- Sanity check: imported all four updated rules and instantiated each output model with the new field shape.

### Test plan

- [ ] Run the EBS Cypher against a populated graph and confirm only volumes with `encrypted = false` are returned; deleted-volume noise disappears on the next finding-replacement cycle.
- [ ] Run the default-SG Cypher and confirm exactly one row per default SG, with `in_use` reflecting at least one EC2 / ENI / LB / RDS attachment.
- [ ] Run the GCP flow-logs Cypher against a project with auto-created defaults and confirm `in_use=false` for the empty subnets.
- [ ] Run the policy / identity admin privileges Cyphers and confirm a single row per `(principal, policy)`; identity-admin no longer flags principals whose only matched action is `iam:CreateServiceLinkedRole`.

#### General

#### Proof of functionality


cartography/rules/data/rules/policy_administration_privileges.py
```diff
@@ -43,18 +43,22 @@
             allow_action IN all_deny_actions OR
             ANY(d IN all_deny_actions WHERE d ENDS WITH('*') AND allow_action STARTS WITH split(d,'*')[0])
         )
-        // Step 4 - Preserve (action, resource) mapping
+        // Step 4 - Aggregate one row per (account, principal, policy)
```
kunaals: i think this still drops allow statements that use `NotResource` without `Resource`. `allow_resources` will be null in that case, so `UNWIND allow_resources` emits no rows and the finding disappears. the identity-admin rule handles this with `UNWIND coalesce(stmt.resource, [null]) AS resource` and then filters nulls out after aggregation; i think we need the same shape here so `NotResource`-only policy-admin grants don't become false negatives.

cartography/rules/data/rules/policy_administration_privileges.py
```diff
@@ -204,8 +207,7 @@
             )
         ] AS matched
     WHERE size(matched) > 0
-    UNWIND matched AS action
-    RETURN DISTINCT
+    RETURN
```
kunaals, line 221: i think this can still return multiple findings for the same principal/role assignment when the role definition has multiple matching `AzurePermissions` blocks. since the goal of this PR is to collapse noisy per-permission findings into one finding per principal/policy, can we aggregate before returning here, e.g. collect distinct matched actions across the assignment/role definition and return one row?

kunaals (changes requested):

kunaals (approved):

PR #2808, opened 2026-05-14
fix(rules): five false-positive filters across rules

jychp:
### Type of change

### Summary

Five small predicate fixes across existing rules in `cartography/rules/data/rules/` to suppress known false-positive findings. Each commit is independent and targets one rule.

1. **`identity_administration_privileges` (AWS)** — exclude `iam:CreateServiceLinkedRole` and `iam:DeleteServiceLinkedRole` from the matched-action list. These actions only let a principal manage AWS-owned service-linked roles, not arbitrary identities, so PowerUserAccess-style principals were being flagged incorrectly.

2. **`unmanaged-account`** — exclude `KubernetesUser` nodes whose name starts with `eks:` / `system:` (AWS-managed Kubernetes service identities) and the `SlackUser` whose id is `USLACKBOT`. These will never link to a `User` ontology node and are not human-managed accounts.

3. **`cis_k8s_5_1_6_sa_token_mounts`** — skip pods in the `kube-system`, `kube-public`, and `kube-node-lease` namespaces. These infrastructure pods (coredns, aws-node, kube-proxy, operators) require API tokens by design; the CIS benchmark itself acknowledges this exception.

4. **`cis_aws_6_5_default_sg_traffic`** — require at least one `EC2NetworkInterface` to be `MEMBER_OF_EC2_SECURITY_GROUP` of the default SG before flagging. Default SGs in empty VPCs/regions cannot expose anything and dominate findings in unused regions.

5. **`policy_administration_privileges`** — collapse the row-per-permission shape into a row per `(principal, scope)`, aggregating matched permissions, resources, and policy names into list fields. Output model gains `actions: list[str]`, `resources: list[str]`, `policy_names: list[str]` (replacing the singular `action`/`resource`/`policy_name`), matching the shape already used by `identity_administration_privileges`.

### How was this tested?

- `make test_lint` — passes
- `uv run --frozen pytest tests/unit/rules/` — 642 passed
- Manual Cypher inspection of each predicate against representative graph data

#### General
- [x] Existing unit tests for these rules still pass; no new tests added (predicate-only changes).

### Notes for reviewers

The `policy_administration_privileges` change updates the `Finding` output model (renames `action` → `actions: list[str]`, `resource` → `resources: list[str]`, `policy_name` → `policy_names: list[str]`). Downstream consumers reading these fields will need to handle lists.


cartography/rules/data/rules/policy_administration_privileges.py
```diff
@@ -258,9 +261,9 @@ class PolicyAdministrationPrivileges(Finding):
     account: str | None = None
     account_id: str | None = None
     principal_type: str | None = None
-    policy_name: str | None = None
-    action: str | None = None
-    resource: str | None = None
+    policy_names: list[str] = []
```
kunaals: i think this is a breaking finding contract change. this rule used to expose `policy_name`, `action`, and `resource`, but now emits `policy_names`, `actions`, and `resources`. downstream consumers keyed on the old fields will silently stop seeing values, since `parse_results()` only maps returned keys into model fields. can we either keep backward-compatible singular aliases or bump the rule version as a breaking change and call it out explicitly?

cartography/rules/data/rules/policy_administration_privileges.py
```diff
@@ -44,18 +44,19 @@
...
             a.name AS account,
             a.id   AS account_id,
             principal.name AS principal_name,
             principal.arn  AS principal_identifier,
             principal_type,
-            policy.name    AS policy_name,
-            allow_action   AS action,
-            resource
-        ORDER BY account, principal_name, action, resource
+            collect(DISTINCT policy.name) AS policy_names,
```
kunaals: the pr says this now matches `identity_administration_privileges`, but that sibling rule still exposes singular `policy_name` while this one now exposes `policy_names`. i think we should align the two rule output shapes, otherwise consumers that handle both privilege rules have to special-case field names.

kunaals (changes requested):

kunaals (approved):

PR #2819, opened 2026-05-15
feat(ontology): project _ont_has_mfa and _ont_active on AWSUser

jychp:
### Type of change

### Summary

`AWSUser` nodes currently expose neither `_ont_has_mfa` nor `_ont_active`, which means the cross-cloud `UserAccount` ontology facts (e.g. `mfa-missing`, unmanaged-accounts) cannot reason over AWS IAM users the same way they reason over other identity providers.

AWS does not ingest the IAM credential report, so there is no `mfa_active` / `account_enabled` field on the `AWSUser` schema. Instead of adding new ingestion, derive both ontology fields from data that is already in the graph:

- `_ont_has_mfa` is `true` iff the user has at least one `:MFA_DEVICE` edge to an `AWSMfaDevice`.
- `_ont_active` is `true` iff the user has signed in via the console at least once (`passwordlastused_dt IS NOT NULL`) or owns at least one `AccountAccessKey` with `status = "Active"`.

The derivation runs as a small analysis job (`ontology_aws_user_projection.json`) inside the ontology users sync, right before `ontology_users_linking.json`. The AWS sync has already loaded the relevant nodes by then.

Also expanded an inline comment in the security-group ontology mapping to explain why `direction` is intentionally not projected for AWS security groups (SGs are bidirectional containers; direction lives on the individual rule nodes, unlike GCP firewalls).

### Related issues or links

None.

### How was this tested?

- `pre-commit run --files <changed files>` passes.
- `uv run pytest tests/unit/cartography/graph/test_ontology_fields.py tests/unit/cartography/intel/ontology/` passes (37 tests).

### Test plan

- Run a sync against an AWS account with a mix of IAM users (some with MFA devices, some without; some with active access keys, some without; some with console password use, some programmatic only).
- Verify with:
  ```cypher
  MATCH (u:AWSUser)
  RETURN u.name, u._ont_has_mfa, u._ont_active
  LIMIT 20;
  ```
- Expect both columns populated with booleans on every row.
- Verify the cross-cloud `missing-mfa-ontology` fact now returns AWS rows when applicable.

#### General


cartography/data/jobs/analysis/ontology_aws_user_projection.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "Ontology - AWS user projection",
+  "statements": [
+    {
+      "__comment": "Project `_ont_has_mfa` on AWSUser from the :MFA_DEVICE edge. AWS does not expose an MFA boolean on the IAM user (the credential report is not ingested), so we derive it from the presence of any AWSMfaDevice attached to the user.",
+      "query": "MATCH (u:AWSUser) SET u._ont_has_mfa = EXISTS((u)-[:MFA_DEVICE]->(:AWSMfaDevice))",
+      "iterative": false
+    },
+    {
+      "__comment": "Project `_ont_active` on AWSUser. A user is considered active if they have used the console at least once (passwordlastused_dt is set) OR own at least one access key whose status is Active.",
+      "query": "MATCH (u:AWSUser) SET u._ont_active = (u.passwordlastused_dt IS NOT NULL) OR EXISTS((u)-[:AWS_ACCESS_KEY]->(:AccountAccessKey {status: 'Active'}))",
```
kunaals: i think this should preserve unknown activity rather than writing `false`.

`passwordlastused_dt IS NULL` does not mean the IAM user is inactive; AWS says it can also mean the user has a password but has never used it, and this sync does not ingest login profiles/credential reports to distinguish that. since the rule uses `COALESCE(a._ont_active, true)`, writing `false` here will cause no-MFA AWS users with unknown activity to be skipped.

could we only set `_ont_active = true` when there is positive activity evidence, and otherwise leave it unset/null?

kunaals (changes requested):

cartography/intel/ontology/users.py
```diff
@@ -26,6 +26,14 @@ def sync(
         data,
         update_tag,
     )
+    # Derive `_ont_has_mfa` and `_ont_active` on AWSUser from related
```
kunaals, line 29: i think this changes when AWS MFA findings are available. `mfa-missing` now only checks `_ont_has_mfa`, but AWS only gets that projected during the `ontology` stage. a supported custom sync like `--selected-modules aws,analysis` would still have `AWSUser`/`AWSMfaDevice` data, but no `_ont_has_mfa`, so AWS users would silently drop out of this rule. could we either run this projection from the AWS IAM sync after MFA/access key loading, or keep the AWS-specific fact for the raw AWS graph path?

kunaals (changes requested):

cartography/data/jobs/analysis/ontology_aws_user_projection.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "Ontology - AWS user projection",
+  "statements": [
+    {
+      "__comment": "Project `_ont_has_mfa` on AWSUser from the :MFA_DEVICE edge. AWS does not expose an MFA boolean on the IAM user (the credential report is not ingested), so we derive it from the presence of any AWSMfaDevice attached to the user.",
+      "query": "MATCH (u:AWSUser) SET u._ont_has_mfa = EXISTS((u)-[:MFA_DEVICE]->(:AWSMfaDevice))",
+      "iterative": false
+    },
+    {
+      "__comment": "Project `_ont_active = true` on AWSUser when there is positive activity evidence: the user has used the console at least once (passwordlastused_dt is set) OR owns at least one access key with status = 'Active'. Absence of evidence is left as NULL (unknown) rather than `false`, because this sync does not ingest login profiles or credential reports and a NULL passwordlastused does not imply inactivity. Rules that gate on `_ont_active` use `COALESCE(_ont_active, true)`, so NULL keeps unknown-activity users in scope.",
+      "query": "MATCH (u:AWSUser) WHERE (u.passwordlastused_dt IS NOT NULL) OR EXISTS((u)-[:AWS_ACCESS_KEY]->(:AccountAccessKey {status: 'Active'})) SET u._ont_active = true",
```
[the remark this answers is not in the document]
kunaals: i think this has merit

kunaals (approved):

PR #2821, opened 2026-05-19
feat(ontology): project _ont_enabled on EntraApplication

jychp:
### Type of change

### Summary
`EntraApplication` nodes carry the `ThirdPartyApp` semantic label but `_ont_enabled` was never populated, because Microsoft Graph exposes `accountEnabled` on the **service principal** (the tenant-local instance) rather than on the application registration. Cross-cloud `ThirdPartyApp` queries therefore could not reason about whether an Entra app was enabled, the way they already do for Keycloak/Okta/Slack.

Following the same pattern as the AWS user MFA projection, derive the ontology field from data that is already in the graph by joining each `EntraApplication` to its same-tenant `EntraServicePrincipal`. Multi-tenant Entra graphs are explicitly handled: the projection scopes the join through `EntraTenant` so a foreign tenant's service principal (which can also link to the home-tenant app via the `app_id`-only relationship matcher) cannot race the home tenant's value.

This PR also fixes a long-standing inconsistency on `AnthropicApiKey` surfaced by the same audit: the schema, docs, and the NIST AI provider key hygiene rule all reference `last_used_at`, but the Anthropic Admin API (`GET /v1/organizations/api_keys`) never returns that field. Every Anthropic key was therefore classified as stale/unused by the NIST rule. Drop the dead property from the schema and the docs, and rescope the Anthropic fact to the signals it can actually evaluate (missing owner, missing workspace scope). The fact id is kept stable to avoid breaking CLI invocations that pin it.

Key changes:
- New analysis job `ontology_entra_application_projection.json`, run from `sync_service_principals` (after cleanup) so that a `--selected-modules microsoft` sync still projects the field. Tenant-scoped join + `OPTIONAL MATCH` ensure orphan apps (SP deleted upstream) get reset to `NULL` instead of keeping a stale boolean.
- `EntraApplication` ontology mapping comment updated to point at the analysis job.
- `AnthropicApiKey.last_used_at` removed from the node schema, the generated schema docs, and the Anthropic ontology mapping; mapping carries an inline comment explaining the upstream gap.
- NIST AI Anthropic fact: drop the staleness `CASE` (now emits `NULL`/`NULL` for `last_used_at`/`is_stale_or_unused`), update name + description, fact id intentionally unchanged.
- Parent rule description rewritten so that listing/metadata accurately reflects what each fact evaluates (ownership + scope for both providers, 90-day staleness for OpenAI only).

### Related issues or links

None.

### Breaking changes

`AnthropicApiKey.last_used_at` is removed. The property was always `NULL` upstream, so any consumer reading it was already getting nothing meaningful; queries that explicitly reference the property will need to drop it.

### How was this tested?

- `make test_lint` passes.
- `uv run pytest tests/unit/rules/test_nist_ai_rmf.py tests/integration/cartography/intel/anthropic/ tests/integration/cartography/intel/microsoft/entra/test_applications.py` passes.
- The Entra applications integration test was extended to cover three cases: (1) happy path (`_ont_enabled = True` projected from the linked SP), (2) orphan path (SP deleted upstream, `_ont_enabled` reset to `NULL`), (3) multi-tenant safety (a foreign-tenant SP with `account_enabled = false` linked to the home-tenant app via `app_id` must not override the home-tenant `True`).

### Test plan

- [ ] Run a sync against an Entra tenant with a mix of applications: some with active service principals, some with disabled service principals, and at least one application whose service principal was removed since the last sync.
- [ ] Verify with:
  ```cypher
  MATCH (app:EntraApplication)
  RETURN app.display_name, app._ont_enabled
  ORDER BY app.display_name;
  ```
  Expect `_ont_enabled` to be `true` for enabled SPs, `false` for disabled SPs, and `NULL` for apps whose SP no longer exists.
- [ ] Run a `--selected-modules microsoft` sync (no ontology stage) and confirm `_ont_enabled` is still set.
- [ ] Run `cartography-rules run nist_ai_provider_api_key_hygiene anthropic_nist_ai_stale_or_unscoped_api_keys` and verify it no longer flags Anthropic keys with active usage simply because the staleness signal is missing.

#### General

#### Proof of functionality


cartography/data/jobs/analysis/ontology_entra_application_projection.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Ontology - Entra application projection",
+  "statements": [
+    {
+      "__comment": "Project `_ont_enabled` on every EntraApplication from the same-tenant EntraServicePrincipal. Microsoft Graph exposes `accountEnabled` on the service principal (tenant-local instance) rather than on the application registration. The SERVICE_PRINCIPAL relationship is matched only on `app_id`, so a multi-tenant Entra graph can attach service principals from foreign tenants to a home-tenant application; selecting the service principal through the shared EntraTenant guarantees we read the home-tenant state and that writes are deterministic. OPTIONAL MATCH guarantees every application is touched: when an app has no service principal in its own tenant (e.g. the SP was just deleted by cleanup), `_ont_enabled` is reset to NULL so stale values from prior syncs cannot linger.",
+      "query": "MATCH (app:EntraApplication)<-[:RESOURCE]-(tenant:EntraTenant) OPTIONAL MATCH (tenant)-[:RESOURCE]->(sp:EntraServicePrincipal {app_id: app.app_id}) WITH app, sp SET app._ont_enabled = sp.account_enabled",
```
kunaals: this runs from the tenant-scoped service-principal sync, but the query rewrites every `EntraApplication` in the graph. in a multi-tenant graph, syncing one tenant can recalculate or clear `_ont_enabled` for apps owned by other tenants. can we scope the first match with the current tenant, e.g. `MATCH (app:EntraApplication)<-[:RESOURCE]-(tenant:EntraTenant {id: $TENANT_ID}) ...`?

kunaals (changes requested):

kunaals (approved):

PR #2822, opened 2026-05-19
feat(gcp): close ontology projection gaps for GCP

jychp:
### Type of change

### Summary

Closes five ontology projection gaps for GCP nodes so cross-cloud semantic-label rules (ComputeInstance, Database, LoadBalancer, Function, ObjectStorage) can reason over GCP resources the same way they reason over AWS / Azure / DO / Scaleway.

- **GCPCloudSQLInstance `type`**: parse `databaseVersion` (e.g. `POSTGRES_14`) into a normalised `database_engine` and project as `_ont_type`.
- **GCPInstance `type` / `created_at` / `public_ip` / `private_ip`**: lift `creationTimestamp`, the primary NIC's `networkIP` and the primary access-config's `natIP` onto the instance so the ComputeInstance projection populates the missing fields. Full GCPNetworkInterface / GCPNicAccessConfig graph is unchanged.
- **GCPForwardingRule `lb_type`**: derive from the `target` proxy collection (`targetHttpsProxies`, `targetPools`, …) with a `backendService` fallback for internal regional LBs that omit `target`. `dns_name` stays unmapped — GCP forwarding rules are addressed by IP only.
- **GCPCloudFunction `memory` / `timeout`**: copy `availableMemoryMb` and parse the protobuf Duration `timeout` (whole seconds → int, fractional → float).
- **GCPBucket `encrypted` / `public`**:
  - `encrypted` is now a static `true` (GCS is encrypted at rest by default; `default_kms_key_name` only flags CMEK).
  - `_ont_public` is computed by a new analysis job `gcp_bucket_public_projection.json` ORing two signals: legacy ACL exposure to `allUsers` / `allAuthenticatedUsers` (captured at sync time as `acl_public` after switching `buckets.list` to `projection=full`) and at least one unconditional `GCPPolicyBinding {is_public: true}`. Mirrors the existing `gcp_bucket_public` rule: buckets with `iam_config_public_access_prevention = 'enforced'` short-circuit to `false`, and conditional bindings are excluded.
  - The projection is gated on storage being requested AND every org's policy bindings actually syncing successfully — falling back to a warning if any org failed or none were discovered, so we never recompute `_ont_public` from stale `APPLIES_TO` edges.

### Related issues or links

- N/A

### How was this tested?

- `make test_lint` passes.
- Updated GCP unit + integration test fixtures and assertions to cover every new field (creation_timestamp / private_ip / public_ip on instances, lb_type on forwarding rules including the backend-service-only fallback, available_memory_mb / timeout including fractional Duration, acl_public on buckets).

### Test plan

- [ ] Run a sync against a GCP organization that has Cloud SQL, Compute, Cloud Functions, forwarding rules and Storage buckets.
- [ ] Verify with:
  ```cypher
  MATCH (n)
  WHERE n:GCPCloudSQLInstance OR n:GCPInstance OR n:GCPForwardingRule
     OR n:GCPCloudFunction OR n:GCPBucket
  RETURN labels(n)[0] AS label,
         n.id,
         n._ont_type,
         n._ont_created_at,
         n._ont_public_ip_address,
         n._ont_private_ip_address,
         n._ont_lb_type,
         n._ont_memory,
         n._ont_timeout,
         n._ont_encrypted,
         n._ont_public
  LIMIT 50;
  ```
- [ ] Expect populated values for the listed fields on the relevant rows.
- [ ] Confirm a bucket with `iam_config_public_access_prevention = 'enforced'` keeps `_ont_public = false` even when an unconditional public binding exists.
- [ ] Confirm a bucket with an `allUsers` ACL entry (uniform bucket-level access disabled) gets `_ont_public = true`.

#### General

#### Proof of functionality


cartography/intel/gcp/storage.py
```diff
@@ -36,7 +36,10 @@ def get_gcp_buckets(storage: Resource, project_id: str) -> Dict:
     :return: Storage response object
     """
     try:
-        req = storage.buckets().list(project=project_id)
+        # `projection=full` is required to get the `acl` / `defaultObjectAcl`
+        # arrays back; the default `noAcl` projection strips them, which would
+        # leave legacy-ACL public buckets undetectable.
+        req = storage.buckets().list(project=project_id, projection="full")
```
kunaals, line 47: this needs to page through `nextPageToken` before we can safely run cleanup. `projection=full` caps `buckets.list` at 200 items, but this only reads the first response; `sync_gcp_buckets()` then treats that partial set as complete and deletes anything not returned. projects with >200 buckets would lose bucket inventory after the first page. can we loop through all pages and add a regression test with a second page?

kunaals (changes requested):

kunaals (approved):

PR #2825, opened 2026-05-19
feat(aibom): Support AIBOM 1.0.0rc4 ingestion

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

Updates the AIBOM intel module to ingest raw Cisco AIBOM `1.0.0rc4` reports directly and load them into the current `AIBOMSource` / `AIBOMComponent` graph model anchored to concrete ontology `:Image` nodes.

This PR removes the legacy parser + compatibility-envelope flow and simplifies ingestion to a more standard Cartography pattern:

### Breaking changes

- The AIBOM module now expects raw top-level `aibom_analysis` reports instead of the old compatibility envelope.
- `AIBOMWorkflow` ingestion has been removed from the current rc4 implementation.

Migration impact:

- Existing downstream expectations that relied on the legacy parser or workflow nodes will need to be updated.
- AIBOM reports should be generated and supplied in raw `1.0.0rc4` form.
- Consumers should expect the currently supported report relationship types to load between `AIBOMComponent` nodes: `USES_MODEL`, `USES_TOOL`, `EXPOSES_TOOL`, and `CUSTOM`.

### How was this tested?

Unit tests
Integration test
Manual validation

#### General

#### Proof of functionality

image
image

### Notes for reviewers

- Component-to-component relationships are now resolved during transform and stored as target component id arrays on the source `AIBOMComponent` payload, which lets the standard schema relationships create the graph


cartography/intel/aibom/__init__.py
```diff
@@ -25,6 +24,76 @@
 stat_handler = get_stats_client(__name__)
 
 
+def _extract_aibom_source_map(document: dict[str, Any]) -> dict[str, Any]:
+    return document["aibom_analysis"]["sources"]
```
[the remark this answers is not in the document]
shyammukund: AIBOM module requires the raw report untouched to be the input so a malformed report should fail loudly.

shyammukund:
@cubic-dev-review

cartography/intel/aibom/transform.py
```diff
@@ -1,295 +1,447 @@
...
+    bucketed by relationship_type/label for later loading.
+    """
+    component_lookup = _build_relationship_component_lookup(document)
+    relationship_records = _parse_aibom_relationship_records(document)
+
+    payloads: list[dict[str, Any]] = []
+    for relationship_record in relationship_records:
+        source_key = relationship_record["source_key"]
+        relationship = relationship_record["relationship"]
+        relationship_type = _as_str(relationship.get("relationship_type"))
```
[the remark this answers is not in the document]
shyammukund: Pushing back on this because the only valid input for this module is a raw AIBOM report therefore we can safely assume that none of the records will be malformed.

cartography/intel/aibom/transform.py
```diff
@@ -1,295 +1,447 @@
...
+
+    for component_record in component_records:
+        source_key = component_record["source_key"]
+        component_type = component_record["component_type"]
+        component = component_record["component"]
+        normalized_component_type = (
+            _as_str(component.get("component_type")) or component_type
+        )
+        component_name = _as_str(component.get("name")) or ""
+        component_lookup[
```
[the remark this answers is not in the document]
shyammukund: Unfortunately path/line/instance don't appear in every relationship record so keying on these fields would make matching inconsistent.

cartography/intel/aibom/__init__.py
```diff
@@ -56,22 +127,42 @@ def sync_aibom_from_report_reader(
...
+        )
+        if not source_payloads:
+            logger.info("AIBOM report %s had no source payloads to ingest", source)
             continue
 
         stat_handler.incr("aibom_reports_processed")
-        load_aibom_document(neo4j_session, parsed_document, update_tag)
+        load_aibom_components(neo4j_session, component_payloads, update_tag)
+        load_aibom_sources(neo4j_session, source_payloads, update_tag)
+        for relationship_label in (
```
achantavy: [coding]  This looks very complicated. Prefer the style of just copying the code and keeping the load functions as dumb as possible.

cartography/intel/aibom/cleanup.py
```diff
@@ -18,6 +21,32 @@ def cleanup_aibom(
...
-    GraphJob.from_node_schema(AIBOMWorkflowSchema(), common_job_parameters).run(
-        neo4j_session,
-    )
+
+    # AIBOM component-to-component matchlinks are scoped by source_key rather than
+    # a top-level tenant/account id passed into sync, so we query the current run's
+    # ingested AIBOMSource nodes to discover which matchlink scopes need cleanup.
+    source_keys = [
+        record["source_key"]
+        for record in neo4j_session.run(
```
achantavy: This does not use transactions, see other examples in the code that use `read_single_value_tx()` or `read_list_of_dicts_tx()` or  

https://github.com/cartography-cncf/cartography/blob/08d15240e755acf5515f0941a69479b22f6b15ca/cartography/client/core/tx.py#L405

https://github.com/cartography-cncf/cartography/blob/08d15240e755acf5515f0941a69479b22f6b15ca/cartography/client/core/tx.py#L449

cartography/intel/aibom/cleanup.py
```diff
@@ -18,6 +21,32 @@ def cleanup_aibom(
...
+            UPDATE_TAG=common_job_parameters["UPDATE_TAG"],
+        )
+    ]
+    relationship_schemas = [
+        AIBOMComponentUsesModelRel(),
+        AIBOMComponentUsesToolRel(),
+        AIBOMComponentExposesToolRel(),
+        AIBOMComponentCustomRel(),
+    ]
+    for source_key in source_keys:
```
achantavy: [style] Don't use for loop here, it adds nesting and there's no benefit here. Just copy paste the calls.

cartography/intel/aibom/loader.py
```diff
@@ -1,144 +1,74 @@
...
+    }
+    rel_schema = rel_schema_map[relationship_label]
+
+    payloads_by_source_key: dict[str, list[dict[str, object]]] = defaultdict(list)
+    for relationship_payload in relationship_payloads:
+        if relationship_payload.get("relationship_label") != relationship_label:
+            continue
+
+        source_key = relationship_payload.get("source_key")
+        if isinstance(source_key, str):
```
achantavy: [coding] Avoid `isinstance`. Is there really a need to have this here? We have explicit types in python

cartography/intel/aibom/loader.py
```diff
@@ -1,144 +1,74 @@
...
+            payloads_by_source_key[source_key].append(relationship_payload)
+
+    for source_key, source_payloads in payloads_by_source_key.items():
```
achantavy: [coding] Avoid iterating through list here, this is not necessary and adds nesting and complexity in this situation.

achantavy (dismissed):
Edit 5/20/26 1742 PST: I was wrong about (2) below. The PR does not introduce new data objects, but it does introduce matchlinks to load data where it previously was not used. We need to confirm whether this is actually necessary, because matchlinks add a perf impact. Original post follows below.

---

Left comments on coding. let's keep the load() functions as dumb as possible. No need to have lists or for loop iterations or mapping when simple copying the code would do fine. One load function per asset type, no need to have it all tightly coupled to one function operating on multiple different types.

More broadly though, this is a large PR that if I understand correctly (please correct me here if I'm wrong) does 2 big things:

1. changes the way a cartography user will configure AIBOM
2. introduces new data objects to AIBOM

In my mind, this is 2 separate PRs. It would be cleaner to have PR 1 first support the new data ingestion method and keep the same data objects, and then have PR 2 introduce the new data objects. I realize there is time pressure to get this done but splitting would overall move faster.

cartography/intel/aibom/cleanup.py
shyammukund: addressed to make sure if we are calling neo4j we are usingi the transaction functions instead of calling neo4j directly.

cartography/intel/aibom/loader.py
shyammukund: Since matchlinks require a sub_resource_id for cleanup we need to collect the source_key and then for each load_matchlink call we need to pass in the source key as the sub resource id. Therefore I thought the best way was to collect the source keys by payload first before loading the matchlinks for each different payload. 

The component to component edges use source_key as the sub__resource_id because the relationships in the report are defined at the source level not the component.

shyammukund: After discussion. makes sense to move this collection step to the transform.

cartography/intel/aibom/__init__.py
shyammukund: Separated out this relationship match link load so each individual relationship has its own load function.

cartography/intel/aibom/cleanup.py
shyammukund: Gave relationship its own cleanup

tests/unit/cartography/intel/aibom/test_cleanup.py
```diff
@@ -0,0 +1,102 @@
...
+
+    sync_aibom_from_report_reader(
+        neo4j_session,
+        reader,
+        123,
+        {"UPDATE_TAG": 123},
+    )
+
+    mock_load_sources.assert_called_once()
+    mock_cleanup_aibom.assert_not_called()
```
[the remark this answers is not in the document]
shyammukund: Accepting limitation in order to follow trivy's all or nothing cleanup implmentation.

cartography/models/aibom/component.py
```diff
@@ -153,29 +139,14 @@ class AIBOMComponentUsesPromptRel(CartographyRelSchema):
...
-            ConditionalNodeLabel(label="AITool", conditions={"category": "tool"}),
-            ConditionalNodeLabel(label="AIMemory", conditions={"category": "memory"}),
-            ConditionalNodeLabel(
-                label="AIEmbedding",
-                conditions={"category": "embedding"},
-            ),
-            ConditionalNodeLabel(label="AIPrompt", conditions={"category": "prompt"}),
-        ],
-    )
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels([])
```
kunaals: this removes the ai-specific labels (`:AIAgent`, `:AIModel`, `:AITool`, etc.) and leaves everything as `:AIBOMComponent`. that makes the graph a lot less useful for security queries and breaks existing rule/ontology assumptions.

cartography/intel/aibom/transform.py
```diff
@@ -1,295 +1,445 @@
...
+    component_lookup: dict[tuple[str, str, str], str] = {}
+    for component_record in component_records:
+        source_key = component_record["source_key"]
+        component_type = component_record["component_type"]
+        component = component_record["component"]
+        normalized_component_type = (
+            _as_str(component.get("component_type")) or component_type
+        )
+        component_name = _as_str(component.get("name")) or ""
+        component_lookup[
```
kunaals: this resolves rel endpoints by just source/type/name. if two components share a name, we can attach the edge to the wrong node. we should use `instance_id` when it exists or fail on ambiguity.

cartography/intel/aibom/__init__.py
```diff
@@ -25,6 +25,65 @@
...
+    source_keys = tuple(sources)
+
+    image_digests = tuple(
+        digest
+        for digest in (
+            _extract_digest_from_source_key(source_key) for source_key in source_keys
+        )
+        if digest
+    )
+    if not image_digests:
```
kunaals: the image anchoring behavior feels inconsistent. if a report/source can’t anchor to a concrete `:Image`, we should either reject it loudly or load explicit unmatched provenance. right now it’s kind of halfway between both.

cartography/intel/aibom/transform.py
```diff
@@ -1,295 +1,445 @@
...
+        )
+        if resolved_component_ids is None:
+            continue
+
+        source_component_id, target_component_id = resolved_component_ids
+        source_component_payload = component_payloads_by_id.get(source_component_id)
+        if source_component_payload is None:
+            continue
+
+        target_component_ids = cast(list[str], source_component_payload[target_field])
```
kunaals, line 502: we’re dropping the evidence behind component-to-component rels. for investigations, it matters why we think an agent uses a model/tool and where that came from.

tests/integration/cartography/intel/aibom/test_aibom.py
```diff
@@ -219,89 +89,35 @@ def _seed_gar_single_platform_graph(neo4j_session) -> None:
...
-    sync_aibom_from_dir(
+    sync_aibom_from_report_reader(
         neo4j_session,
-        "/tmp",
+        LocalReportReader("/tmp"),
         TEST_UPDATE_TAG,
         {"UPDATE_TAG": TEST_UPDATE_TAG},
     )
 
     assert check_nodes(
```
kunaals, line 136: the integration coverage got stripped too far. this needs real sync/load/cleanup tests for stale cleanup, skipped reports, duplicate rel endpoints, and the labels/rels downstream rules rely on.

kunaals (changes requested):

cartography/models/aibom/component.py
shyammukund: Added back the AI specific labels that existed previously. These no longer are used to create component to component relationships but still used for the ontology/rule assumptions.

cartography/intel/aibom/transform.py
shyammukund: Right now we load these relationships as part of the component node by storing a list of target component ids on the source component payload. That works for saying “this component points to these other components,” but it only preserves the connection itself. It does not give us a separate record per relationship, so we have nowhere to store relationship-specific evidence like justification or file/line locations. To preserve that evidence, we would need standalone relationship payloads containing source id, target id, and evidence fields, which means loading relationships separately from the component node.

I think we have two options


1. Keep the current implmentation that drops evidence in favor of using the standard cartography load 
2. Reconsider using matchlinks here if we decide that evidence is a must have on the component to component relationship

cartography/intel/aibom/__init__.py
shyammukund: Decided to fail loudly if the source is not a digest and to also fail loudly if the digest found does not resolve to any Image node that exists in the graph.

cartography/intel/aibom/transform.py
shyammukund: In the sample report source_instance_id and component_instance_id are missing from every single relationship record. If we fail when it doesn't exist it would fail every time. I do not think that this is a outlier of our report since every single relationship has these fields missing.

tests/integration/cartography/intel/aibom/test_aibom.py
shyammukund: Have added another integration test that runs through the real sync (including load and cleanup) to test stale cleanup, a unit test for duplicate rel endpoints, and added coverage to the happy path integration test to verify the labels/rels that downstream rules rely on.

cartography/models/aibom/component.py
kunaals: thanks, this part looks right to me now.

cartography/intel/aibom/transform.py
kunaals: agreed we should not fail just because instance ids are missing. the thing i want to avoid is silently choosing the wrong endpoint. can we use instance ids when present, and only fall back to name/type when that resolves uniquely?

cartography/intel/aibom/__init__.py
kunaals: this direction makes sense. one gap: the current check still accepts mixed reports if at least one source key is digest-qualified. if the contract is digest-only, we should fail on any non-digest source key and add a test for that.

cartography/intel/aibom/transform.py
kunaals: yeah, i am okay not blocking this pr on relationship evidence if we treat it as a follow-up. the main thing i do not want to ship is wrong edges; missing edge-level evidence is less urgent than ambiguous endpoint resolution.

tests/integration/cartography/intel/aibom/test_aibom.py
kunaals: thanks, the integration coverage helps. i think we still need a test for the ambiguous same-name endpoint case, since the current duplicate test covers duplicate relationship records rather than endpoint ambiguity. rc4 also has more relationship types than we currently map, so we should either cover the important ones here or make the unsupported behavior explicit and tested.

cartography/intel/aibom/transform.py
shyammukund: I updated build_relationship_component_lookup to 
1. support 2 endpoint key shapes in component_lookup. One for instance id matching and one for name/type matching
2. keep track of the ambiguous name/type keys for relationships we should skip

Then per relationship we call resolve_relationship_component_ids with the source and target instance_ids if we have them or the fallback type/name keys. if the type or name keys are in the ambigous_type_name_keys collection then we skip that relationship ensuring that we didn't silently choose the wrong endpoint.

cartography/intel/aibom/__init__.py
shyammukund: Changed behavior to not allow mixed reports to ensure the contract is digest only reports. We now raise a ValueError if even one source in the report isn't a digest. Added a unit test for the mixed case failure.

tests/integration/cartography/intel/aibom/test_aibom.py
shyammukund: Added an integration test to make sure that if there is an ambiguous same name endpoint case the relationship is not created. 

Per offline discussion. will fast follow with a PR that adds in every supported Component to Component relationship, not just the ones that were produced in the sample report. Because this PR will have a fast follow that supports all realtionships i did not add a test for making sure these relationships are skipped.

shyammukund:
@cubic-dev-ai review

cartography/intel/aibom/__init__.py
shyammukund: Also added check that raises failure for reports with empty sources.

kunaals (approved):
LGTM

PR #2780, opened 2026-05-08
feat(AWS): Add AWS Organizations support

kunaals:
### Type of change
- [x] New feature (non-breaking change that adds functionality)
- [x] Refactoring (no functional changes)
- [x] Documentation update

### Summary
Adds AWS Organizations support to AWS ingestion with a hierarchy model that mirrors AWS Organizations and the existing GCP CRM style.

This creates `AWSOrganization`, `AWSOrganizationRoot`, and `AWSOrganizationalUnit` nodes, labels `AWSOrganization` as `Tenant`, enriches `AWSAccount` nodes with Organizations account metadata, creates placement relationships for active organization accounts, and creates traversable hierarchy relationships:

```cypher
(:AWSOrganization)-[:RESOURCE]->(:AWSOrganizationRoot)
(:AWSOrganizationRoot)-[:PARENT]->(:AWSOrganization)
(:AWSOrganizationRoot)-[:RESOURCE]->(:AWSOrganizationalUnit)
(:AWSOrganizationalUnit)-[:RESOURCE]->(:AWSOrganizationalUnit)
(:AWSOrganizationRoot)-[:RESOURCE]->(:AWSAccount)
(:AWSOrganizationalUnit)-[:RESOURCE]->(:AWSAccount)
(:AWSOrganizationalUnit)-[:PARENT]->(:AWSOrganizationRoot|AWSOrganizationalUnit)
(:AWSAccount)-[:PARENT]->(:AWSOrganizationRoot|AWSOrganizationalUnit)
```

[...]

PR #2828, opened 2026-05-21
fix(aws): skip org cleanup when account state is unavailable

kunaals:
### Type of change

### Summary

AWS Organizations cleanup should only run after Cartography can determine current account placement safely. This adds a defensive check that treats a hierarchy response as incomplete if an account has neither the current `State` field nor the legacy `Status` field.

That prevents stale SDK/service-model combinations from turning unknown account state into destructive hierarchy cleanup.

### Related issues or links

- Follow-up to #2780

### Breaking changes

None.

### How was this tested?

- Added basic tests to catch this case

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

This does not move cleanup into an analysis job. The guard stays in the Organizations sync path so cleanup remains coupled to complete, trustworthy hierarchy enumeration.

### DIFF
--- a/cartography/intel/aws/organizations.py
+++ b/cartography/intel/aws/organizations.py
@@ -148,6 +148,10 @@ def _get_account_state(account: dict[str, Any]) -> str | None:
     return account.get("State") or account.get("Status")
 
 
+def _has_account_state_signal(account: dict[str, Any]) -> bool:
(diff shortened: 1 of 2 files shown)


jychp (approved):
LGTM

PR #2830, opened 2026-05-21
feat(aws): add --aws-excluded-syncs to skip individual modules

mandeepsinghbaldd:
## Motivation

When a single AWS module misbehaves (slow, throttled, or temporarily broken upstream), users today must either disable the entire AWS sync or enumerate every other module via `--aws-requested-syncs`. The latter breaks silently when cartography adds a new module in a release — the user's curated allowlist will keep working but quietly stop syncing the newly-added module.

This change adds `--aws-excluded-syncs` as the inverse of `--aws-requested-syncs`. Excluded modules are subtracted from the effective sync set (either `--aws-requested-syncs` if set, or the full `RESOURCE_FUNCTIONS` default). Unknown module names raise a `ValueError` at startup, mirroring the validation behavior of `--aws-requested-syncs`.

### Motivating example

SecretsManager's sync runs a strict serial loop calling `list_secret_version_ids` per secret (`cartography/intel/aws/secretsmanager.py:238`). Under SecretsManager API throttling, each call hits cartography's nested retry/backoff — boto3's internal retries *plus* the `@aws_handle_regions` decorator's `backoff.on_exception(..., max_time=600)` in `cartography/util.py:706` — ballooning to ~3 min/secret. With ~1000 secrets in a single account/region, the module takes days and stalls every later module in the `RESOURCE_FUNCTIONS` sync order.

`--aws-excluded-syncs secretsmanager` is the minimal-disruption workaround until that loop is parallelised, and the same shape of escape hatch should work for any future module that hits comparable pathology.

## Implementation notes

- **Exclusion is applied inside `_sync_one_account` *after* its inner `_normalize_requested_syncs` call.** This is load-bearing: `_normalize_requested_syncs` has a backward-compat auto-include for `ec2:load_balancer_v2:expose` whenever `ec2:load_balancer_v2` is requested. If exclusion were applied earlier (e.g. only in `start_aws_ingestion`), the inner normalize would silently re-add the excluded module. Applying exclusion as the final filter step makes the user's exclusion authoritative.
- `aws_excluded_syncs` is plumbed through `_sync_multiple_accounts` → `_sync_one_account` as a new optional kwarg, so programmatic callers can opt into the same behavior without going through `start_aws_ingestion`.
- `_perform_aws_analysis` receives the post-exclusion list so cross-module analysis jobs whose deps were excluded are correctly skipped via the existing `run_analysis_and_ensure_deps` gating.

## Semantics

When both `--aws-requested-syncs` and `--aws-excluded-syncs` are passed, the effective set is `requested - excluded`:

```bash
cartography --aws-requested-syncs "ec2:instance,s3,iam,secretsmanager" \
            --aws-excluded-syncs  "secretsmanager"
# Effective set: ec2:instance, s3, iam
```

When only `--aws-excluded-syncs` is passed, the effective set is `ALL_MODULES - excluded`.

## Tests

- **Unit**: `parse_and_validate_aws_excluded_syncs` covered for single, multi, empty, trailing-comma, unknown-name, and garbage inputs.
- **Integration**: new tests verify (a) excluded module is not invoked even when normalize would auto-include it (the load-bearing order-of-operations case), and (b) exclusion works without `--aws-requested-syncs` (default-all minus excluded).
- All existing tests continue to pass; the only mechanical adjustments are kwarg additions in `_sync_one_account` call assertions (`aws_excluded_syncs=[]` propagated).
- `pre-commit run --all-files` clean (isort / black / flake8 / pyupgrade / mypy).

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -518,6 +518,20 @@ class CLI:
                     hidden=PANEL_AWS not in visible_panels,
                 ),
             ] = None,
+            aws_excluded_syncs: Annotated[
+                str | None,
+                typer.Option(
+                    "--aws-excluded-syncs",
+                    help=(
+                        "Comma-separated list of AWS resources to skip. Subtracted from the "
+                        "effective sync set (either --aws-requested-syncs or, if unset, all "
+                        'modules). Example: "secretsmanager". See cartography.intel.aws.resources '
+                        "for the full list of valid names."
+                    ),
+                    rich_help_panel=PANEL_AWS,
+                    hidden=PANEL_AWS not in visible_panels,
+                ),
+            ] = None,
             aws_guardduty_severity_threshold: Annotated[
                 str | None,
                 typer.Option(
@@ -1971,6 +1985,12 @@ class CLI:
                 )
 
                 parse_and_validate_aws_requested_syncs(aws_requested_syncs)
+            if aws_excluded_syncs:
+                from cartography.intel.aws.util.common import (
+                    parse_and_validate_aws_excluded_syncs,
+                )
+
+                parse_and_validate_aws_excluded_syncs(aws_excluded_syncs)
             if aws_regions:
                 from cartography.intel.aws.util.common import (
                     parse_and_validate_aws_regions,
@@ -2499,6 +2519,7 @@ class CLI:
                 entra_client_id=entra_client_id,
                 entra_client_secret=entra_client_secret,
                 aws_requested_syncs=aws_requested_syncs,
+                aws_excluded_syncs=aws_excluded_syncs,
                 aws_guardduty_severity_threshold=aws_guardduty_severity_threshold,
                 analysis_job_directory=analysis_job_directory,
                 oci_sync_all_profiles=oci_sync_all_profiles,
--- a/cartography/config.py
+++ b/cartography/config.py
@@ -100,6 +100,10 @@ class Config:
     :param entra_client_secret: Client Secret for connecting in a Service Principal Authentication approach. Optional.
     :type aws_requested_syncs: str
     :param aws_requested_syncs: Comma-separated list of AWS resources to sync. Optional.
+    :type aws_excluded_syncs: str
+    :param aws_excluded_syncs: Comma-separated list of AWS resources to skip. Subtracted from the
+        effective sync set (``aws_requested_syncs`` if provided, otherwise the default of all
+        modules). Optional.
     :type aws_guardduty_severity_threshold: str
     :param aws_guardduty_severity_threshold: GuardDuty severity threshold filter. Only findings at or above this
         severity level will be synced. Valid values: LOW, MEDIUM, HIGH, CRITICAL. Optional.
@@ -362,6 +366,7 @@ class Config:
         entra_client_id=None,
         entra_client_secret=None,
         aws_requested_syncs=None,
+        aws_excluded_syncs=None,
         aws_guardduty_severity_threshold=None,
         analysis_job_directory=None,
         oci_sync_all_profiles=None,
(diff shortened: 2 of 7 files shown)


jychp:
Why not just using `--aws-requested-syncs` ?

achantavy:
Hey @mandeepsinghbaldd, I think maintaining 2 different inclusion and exclusion lists makes for a confusing interface and adds extra logic that is hard to maintain, especially because having this pattern would set a precedence for the other modules (e.g. why do you have inclusion + exclusion for AWS but only inclusion for GCP?). I'd prefer to avoid this and I hope this reasoning makes sense.

If there are errors, you can either 

1. use `--aws-requested-syncs` to pick the ones you want
2. Use `--aws-best-effort-mode` to run everything but have the sync continue when it encounters an exception.

If 2 happens and if it is related to a failure in cartography and not a transient error, I'd love to know the error, so please file an issue to this repo when you find these.

Happy to discuss more if you have questions.

mandeepsinghbaldd:
> Why not just using `--aws-requested-syncs` ?

@jychp Honestly tried that first. The catch is an allowlist freezes the module list at whatever point you write it. When cartography adds new modules in a release, they just don't get synced. No warning, the data just doesn't show up in Neo4j. Already hit this going from 0.134 to 0.136.

Exclusion-only is different because the base set is whatever's in RESOURCE_FUNCTIONS at runtime. We name what to skip, anything new gets picked up on the next release automatically. That's the bit --aws-requested-syncs can't really do.

Happy to change my approach, if better solution exists

mandeepsinghbaldd:
> Hey @mandeepsinghbaldd, I think maintaining 2 different inclusion and exclusion lists makes for a confusing interface and adds extra logic that is hard to maintain, especially because having this pattern would set a precedence for the other modules (e.g. why do you have inclusion + exclusion for AWS but only inclusion for GCP?). I'd prefer to avoid this and I hope this reasoning makes sense.
> 
> If there are errors, you can either
> 
> 1. use `--aws-requested-syncs` to pick the ones you want
> 2. Use `--aws-best-effort-mode` to run everything but have the sync continue when it encounters an exception.
> 
> If 2 happens and if it is related to a failure in cartography and not a transient error, I'd love to know the error, so please file an issue to this repo when you find these.
> 
> Happy to discuss more if you have questions.

@achantavy Fair point on interface complexity, but want to make one case before we drop it.
The mode I actually care about is exclusion-only, no allowlist passed, just --aws-excluded-syncs secretsmanager. Then the base set is list(RESOURCE_FUNCTIONS.keys()) at runtime, so modules added in future releases automatically flow in. With --aws-requested-syncs we'd own the full list forever, and any release that adds a module silently drops out of our sync (0.134 to 0.136 already did this). For us that silent-drop-on-upgrade is worse than the maintenance cost of one extra flag.
 
  --aws-best-effort-mode doesn't help us either, because secretsmanager doesn't actually fail. It just succeeds really slowly. The per-secret loop in secretsmanager.py around line 238 calls list_secret_version_ids once per secret, and under throttling each call goes through the nested retry stack (boto3retries + the @aws_handle_regions backoff.on_exception(..., max_time=600) in util.py) and takes 2-3 minutes. With ~1000 secrets in one region that's days, and everything after it in the sync order just waits. Nothing throws, so best-effort never engages. It's a perf cliff, not an error.

I can open a separate PR for the real root cause too: bounded-concurrency aioboto3 in get_secret_versions, same shape as ecr_image_layers.sync. But even after that lands the escape hatch is useful for the next module that hits something similar.

On the GCP/Azure asymmetry, I'd flip it. The fact that other providers don't have exclusion isn't a reason to avoid it in AWS, it's a reason to add the same primitive there too. Opt-out scales O(1) per release, opt-in scales O(every new module). Happy to follow up with PRs for GCP/Azure if the AWS one lands.

PR #2833, opened 2026-05-22
feat(aws): add CAN_PASSROLE permission relationship

jychp:
### Type of change

### Summary

Adds a `(:AWSPrincipal)-[:CAN_PASSROLE]->(:AWSRole)` relationship to the AWS graph by extending the declarative permission resolver with one new entry in `cartography/data/permission_relationships.yaml`.

`iam:PassRole` is the IAM action AWS checks when a principal hands an IAM role to an AWS service (attaching an instance profile to `ec2:RunInstances`, an execution role to `lambda:CreateFunction`, a task role to `ecs:RunTask`, etc.). Combined with a service-launch permission, `iam:PassRole` is a well-known privilege-escalation primitive: the caller can launch a workload running under a more powerful role, then execute code inside that workload to inherit the role's credentials.

Until now Cartography had no edge for this. Adding it unlocks PassRole-based detections (e.g. Karpenter / IRSA workload identity abuse, where the controller's role typically holds `ec2:RunInstances` + broad `iam:PassRole` over a pool of node roles, making controller compromise equivalent to account takeover).

The fix is purely declarative; no resolver changes are required. The existing engine already filters on `Effect: Allow`, supports action wildcards (`iam:*`, `*`), and resolves resource ARNs against target node `arn` properties. `AWSRole.arn` is already the primary indexed identifier.

### Related issues or links

- N/A

### How was this tested?

- New integration test `test_permission_relationships_can_passrole` seeds a target `AWSRole`, attaches an `iam:PassRole` policy to a caller role, runs `permission_relationships.sync()`, and asserts the `CAN_PASSROLE` edge is created.
- Existing unit tests still pass (`pytest tests/unit/cartography/intel/aws/test_permission_relationships.py`, 25 passed) since the resolver code is unchanged.
- `make test_lint` passes.

### Out of scope (intentional)

- SCP / permission-boundary intersection. The resolver does not honor boundaries today; this PR matches existing behavior for `CAN_READ`, `CAN_WRITE`, etc.
- `Condition` block evaluation. Existing edges ignore conditions; same here.
- Role trust-policy compatibility (whether the launching AWS service can actually assume the passed role). Permission-only modeling, consistent with the rest of the file.

#### General

#### Proof of functionality

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (`docs/root/modules/aws/schema.md`, AWSPrincipal section).

### Notes for reviewers

The diff is intentionally minimal: one YAML entry, one integration test, one schema-doc bullet. No changes to the resolver itself.


cartography/data/permission_relationships.yaml
```diff
@@ -61,3 +61,11 @@
...
   - ecs:ExecuteCommand
   relationship_name: CAN_EXECUTE_COMMAND
+# Map principals that can pass an IAM role to an AWS service.
+# iam:PassRole is required to attach an instance profile to ec2:RunInstances,
+# an execution role to lambda:CreateFunction, ecs:RunTask, etc. Combined with
+# a service-launch permission this is a privilege-escalation primitive.
+- target_label: AWSRole
+  permissions:
+  - iam:PassRole
+  relationship_name: CAN_PASSROLE
```
kunaals: could we rename this to `CAN_PASS_ROLE`? i think most existing multi-word permission relationship labels use underscores between words, e.g. `CAN_EXECUTE_COMMAND`

kunaals (changes requested):

kunaals (approved):

PR #2838, opened 2026-05-26
perf(graph): add MatchLink Cartesian product loader

kunaals:
### Type of change
- [x] Other (please describe): core graph loading performance helper

### Summary
Adds a reusable core MatchLink Cartesian product loader for cases where every source node in a batch should be linked to every target node in a batch.

This is motivated by high-cardinality permission expansions such as GCP BigQuery table permissions, where broad grants can require direct relationships from many principals to many tables. This PR adds only the shared core loader; a follow-up GCP module change can consume it while preserving existing direct table-level permission relationship semantics.

The new helper keeps writes serial and bounded by source and target batch sizes, ensures MatchLink indexes before writing, returns the actual number of matched relationships from Neo4j, and uses the existing `GraphJob.from_matchlink()` cleanup model.

It is also likely useful for other future permission or identity inheritance loaders where a set of source nodes should connect to a set of target nodes with constant relationship properties.

### Related issues or links

N/A

### Breaking changes

None.

### How was this tested?

- Unit and integration tests

#### General

#### Proof of functionality

- [ ] Included Cartography sync logs from a real environment demonstrating successful synchronization of the new/modified entity. Not applicable; this PR adds a core loader helper and does not add or modify a synced entity.

- [ ] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules). Not needed; graph schema is unchanged.
- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md). Not needed; graph schema is unchanged.

- [ ] Used the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node). Not applicable.

### Notes for reviewers

`load_matchlinks_cartesian_product()` is intentionally narrow in v1. It supports one exact source matcher key, one exact target matcher key, no endpoint sub-resource matchers, and only relationship properties supplied from kwargs. Row-specific mappings and row-specific relationship properties should continue to use `load_matchlinks()`.


cartography/client/core/tx.py
```diff
@@ -905,3 +925,142 @@ def load_matchlinks(
...
+            )
+
+    if relationships_loaded < total_attempted_pairs:
+        logger.warning(
+            "Loaded %d relationships for %d attempted %s pairs. Some source or target "
+            "nodes were not matched.",
+            relationships_loaded,
+            total_attempted_pairs,
+            description,
+        )
```
achantavy, line 1056: [non-block]

Because cartography does not use formal `constraint`s in neo4j, 2 nodes can share property values, e.g. if a job had a problem or if there was a bug someplace else created 2 AWSRole nodes with the same `arn`. The matcher then resolves 1 input value to multiple nodes, so `relationships_loaded` can exceed `total_attempted_pairs`, and then the accounting doesn't mean much after that.

This precondition of assuming that each matching property value is unique is currently undocumented. We could document it, or adjust the warning message to be more correct, like maybe

```python
if relationships_loaded < total_attempted_pairs:
    logger.warning("Loaded %d of %d attempted %s pairs; some source/target nodes were not matched.", ...)
elif relationships_loaded > total_attempted_pairs:
    logger.warning(
        "Loaded %d relationships for only %d attempted %s pairs. A matcher key likely "
        "resolves to multiple nodes (non-unique key); relationship counts are unreliable.",
        relationships_loaded, total_attempted_pairs, description,
    )
```

cartography/client/core/tx.py
```diff
@@ -905,3 +925,142 @@ def load_matchlinks(
...
     )
+
+
+def load_matchlinks_cross_product(
+    neo4j_session: neo4j.Session,
+    rel_schema: CartographyRelSchema,
+    source_values: list[Any],
+    target_values: list[Any],
+    source_batch_size: int = 100,
+    target_batch_size: int = 1000,
```
achantavy, line 936: Would be helpful to explain why these defaults were chosen and in what situations would we tune these

achantavy (approved):
Great solution. It's bounded and scoped to a specific problem. Instead of doing the cartesian product in python-land, we do the lesser of two evils and do that in neo4j. 

Left nits on docs and clarity on logs

cartography/client/core/tx.py
```diff
@@ -905,3 +925,142 @@ def load_matchlinks(
         rel_schema.rel_label,
         rel_schema.target_node_label,
     )
+
+
+def load_matchlinks_cross_product(
```
achantavy: Oh oh, something to be pedantic about: the correct word is cartesian product and not cross product 

Image

kunaals: lmao good call will update

cartography/client/core/tx.py
kunaals: done!

cartography/client/core/tx.py
kunaals: done

cartography/client/core/tx.py
kunaals: good call, done

PR #2839, opened 2026-05-26
feat(aws): ingest EKS access entries

kunaals:
### Type of change
- [x] Documentation update

### Summary
This adds first-class AWS graph modeling for EKS Access Entries.

EKS clusters can grant Kubernetes access through the EKS API instead of, or alongside, the legacy `aws-auth` ConfigMap. Cartography already used access entries as an identity-mapping source in the Kubernetes EKS path, but the AWS EKS sync did not preserve the access entry resources themselves. This change stores the cluster authentication mode on `EKSCluster`, ingests `EKSAccessEntry` nodes from `ListAccessEntries`, enriches them with `DescribeAccessEntry` when available, and links each entry to its AWS account, EKS cluster, and IAM principal when that principal already exists in the graph.

For `CONFIG_MAP`-only clusters, the AWS EKS sync skips access entry calls because those clusters do not support the API authentication path. If `DescribeAccessEntry` is denied, Cartography still loads minimal access entry nodes from `ListAccessEntries` so SecurityAudit-style read-only roles can continue syncing EKS clusters.

### Related issues or links
N/A

### Breaking changes
None.

### How was this tested?
- Tested locally and persisted API shapes in integration test data

#### General
- [ ] The linter passes locally (`make test_lint`). Targeted pre-commit passed for the touched files.

#### Proof of functionality

- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md). No schema README exists in this checkout.

### Notes for reviewers
AWS SecurityAudit currently includes `eks:ListAccessEntries`; that is enough to create minimal `EKSAccessEntry` nodes and principal/cluster relationships. Granting `eks:DescribeAccessEntry` additionally populates access entry ARN, username, type, Kubernetes groups, and timestamps.


cartography/intel/aws/eks.py
```diff
@@ -46,6 +52,96 @@ def get_eks_describe_cluster(
...
+                cluster_name,
+            )
+            return []
+        raise
+
+    return principal_arns
+
+
+@timeit
+def get_eks_access_entries(
```
jychp, line 87: `get_eks_clusters` and `get_eks_describe_cluster` are decorated with `@aws_handle_regions`; this function is not. Only `AccessDeniedException`, `ResourceNotFoundException`, and the unsupported-auth-mode `InvalidRequestException` are caught, so any other `ClientError` from `ListAccessEntries`/`DescribeAccessEntry` will abort the entire multi-region/multi-account sync. Add `@aws_handle_regions`.

cartography/intel/aws/eks.py
```diff
@@ -202,12 +314,32 @@ def cleanup(
...
 
+def transform_access_entries(
+    access_entries: list[dict[str, Any]],
+    cluster_arn: str,
+) -> list[dict[str, Any]]:
+    transformed_entries = []
+    for entry in access_entries:
+        transformed_entry = entry.copy()
+        transformed_entry["id"] = entry.get(
+            "accessEntryArn",
```
jychp: The node `id` is the synthetic `{cluster_arn}/access-entry/{principalArn}` when only `ListAccessEntries` is available, but becomes the real `accessEntryArn` once `DescribeAccessEntry` is granted. The same access entry then changes identity across runs, orphaning the old node until cleanup and breaking any references to it. Derive `id` consistently from cluster + principal and keep `accessEntryArn` only in the `arn` property.

jychp (changes requested):

cartography/intel/aws/eks.py
kunaals: good call — added `@aws_handle_regions` and covered the regional access error path.

cartography/intel/aws/eks.py
kunaals: good catch — the id now always uses cluster + principal, and `accessEntryArn` stays on the `arn` property.

cartography/intel/aws/eks.py
```diff
@@ -46,6 +52,97 @@ def get_eks_describe_cluster(
...
+            error_code = e.response["Error"]["Code"]
+            if error_code == "ResourceNotFoundException":
+                logger.warning(
+                    "Access entry lookup failed for principal %s on cluster %s: %s",
+                    principal_arn,
+                    cluster_name,
+                    e,
+                )
+                continue
+            if error_code == "AccessDeniedException":
```
jychp, line 130: When the role has ListAccessEntries but not DescribeAccessEntry (the SecurityAudit path this PR targets), every principal fires a failing describe_access_entry call plus a WARNING: N wasted API calls and N log lines on the expected path. Short-circuit describe after the first AccessDeniedException, or drop the per-entry log to DEBUG.

cartography/models/aws/eks/access_entry.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@dataclass(frozen=True)
+class AWSPrincipalToEKSAccessEntryRel(CartographyRelSchema):
+    target_node_label: str = "AWSPrincipal"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"arn": PropertyRef("principalArn")},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "HAS_ACCESS_ENTRY"
```
jychp: AWSPrincipalToEKSAccessEntryRel reuses rel_label HAS_ACCESS_ENTRY, identical to EKSClusterToAccessEntryRel (line 58). MATCH ()-[:HAS_ACCESS_ENTRY]->(:EKSAccessEntry) then cannot distinguish the owning cluster from the granted principal without inspecting source labels. Use a distinct label for the principal edge (e.g. GRANTED_ACCESS_TO).

jychp (approved):
Approved with nits

cartography/intel/aws/eks.py
kunaals: fixed in the earlier update — the id is always derived from cluster + principal, with the arn stored separately.

cartography/intel/aws/eks.py
kunaals: good catch. describe now stops after the first access denied, and the remaining entries use the list data.

cartography/models/aws/eks/access_entry.py
kunaals: good catch. principal edges now use granted_access_to.

PR #2840, opened 2026-05-26
perf(gcp): bulk load BigQuery table permissions

kunaals:
### Type of change
- [ ] Documentation update

### Summary
This PR uses the MatchLink Cartesian product loader added in #2838 for high-cardinality GCP permission expansion.

Large GCP environments can have many BigQuery tables. When broad project- or dataset-level IAM scopes apply to BigQuery table permissions, the previous path repeatedly evaluated every table against every principal and then loaded the resulting direct table relationships in small MatchLink batches.

This change makes GCP permission relationship loading scope-aware: project-wide grants bulk-load direct relationships across the target resource set, BigQuery dataset-wide grants bulk-load direct `GCPPrincipal -> GCPBigQueryTable` `CAN_*` relationships for tables in the dataset, and exact table or non-container scopes stay on the existing row-by-row evaluator. Graph semantics and cleanup stay unchanged.

### Related issues or links
- Uses the MatchLink Cartesian product helper added in #2838.

### Breaking changes
None.

### How was this tested?
- `uv run --frozen pytest tests/unit/cartography/intel/gcp/test_permission_relationships.py -q`
- `uv run --frozen pytest tests/integration/cartography/intel/gcp/test_permission_relationships.py -q`
- `uv run --frozen pre-commit run --files cartography/intel/gcp/permission_relationships.py tests/unit/cartography/intel/gcp/test_permission_relationships.py tests/integration/cartography/intel/gcp/test_permission_relationships.py`
- `make test_lint`

#### General

#### Proof of functionality

### Notes for reviewers
No schema documentation update is needed because this preserves the existing direct permission relationships and cleanup scope.


cartography/intel/gcp/permission_relationships.py
```diff
@@ -174,6 +177,24 @@ def evaluate_policy_binding_for_permissions(
     return False
 
 
+def assignment_allows_permissions(
```
achantavy: [nonblock] Nit: demote to private function?

cartography/intel/gcp/permission_relationships.py
```diff
@@ -560,16 +760,30 @@ def sync(
...
-            neo4j_session,
-            principals,
-            resource_dict,
-            permissions,
-            matchlink_schema,
-            update_tag,
-            project_id,
-            batch_size=GCP_PERMISSION_RELATIONSHIP_BATCH_SIZE,
-        )
+        if target_label == "GCPBigQueryTable":
```
achantavy: Right now we're splitting on target label to decide whether to use cross product or a standard matchlink. I think it's more correct to split on scope. That is, if the permission scope is a project, or if the permission scope is a "container" (like bigquery dataset -> bigquery table), then we do cross product, otherwise we do the standard row-by-row matchlink logic.

There are other patterns of this "container-of" thing: GCPKeyRing->GCPCryptoKey, and GCPSecretManagerSecret->GCPSecretManagerSecretVersion (this case is a bit too finegrained imo but you get my idea)

achantavy (changes requested):

cartography/intel/gcp/permission_relationships.py
kunaals: good catch. renamed this private.

cartography/intel/gcp/permission_relationships.py
kunaals: good catch. changed this to split project/container scopes first. bigquery table handling is now the registered dataset-container loader, and exact/non-container scopes still use the row-by-row path.

achantavy (approved):
amazing!

Issue #2831, opened 2026-05-22
Azure: model RBAC role assignments at their owning scope


kunaals:
# Azure: model RBAC role assignments at their owning scope

## Description
Azure RBAC role assignments can be attached at several Azure scopes, including management group, subscription, resource group, and individual resource scopes. Cartography's current Azure RBAC ingestion is subscription-centered: role assignments, role definitions, and permission nodes are loaded from the subscription authorization client path and are modeled as subresources of `AzureSubscription`.

That works for subscription-scoped RBAC, but it becomes inaccurate for management-group-scoped RBAC. A role assignment attached to a management group is not really a subscription resource, even though it can grant effective access to descendant subscriptions and resources through Azure's scope hierarchy.

The proposed model is to keep one unified Azure RBAC node model, while materializing each role assignment at the Azure scope where it is actually attached:

- subscription-scoped role assignments stay owned by `(:AzureSubscription)-[:RESOURCE]->(:AzureRoleAssignment)`
- management-group-scoped role assignments are owned by `(:AzureManagementGroup)-[:RESOURCE]->(:AzureRoleAssignment)`
- future resource-group or resource-scoped role assignments can be added with the same pattern, without introducing separate duplicate RBAC node labels
- Azure management groups should be represented in the tenant hierarchy, with `PARENT` relationships from child scope to parent scope for traversal
- derived permission relationships, such as `(:EntraUser)-[:CAN_READ]->(:AzureSQLServer)`, should still represent effective access, but they should preserve enough provenance to explain whether the access is direct or inherited and which upstream scope granted it

Suggested hierarchy shape:

```cypher
// Tenant-owned hierarchy objects. AzureSubscription already follows this pattern today.
(:AzureTenant)-[:RESOURCE]->(:AzureManagementGroup)
(:AzureTenant)-[:RESOURCE]->(:AzureSubscription)

// Placement/hierarchy edges point from child to parent, matching AWS/GCP conventions.
(:AzureManagementGroup)-[:PARENT]->(:AzureTenant)
(:AzureManagementGroup)-[:PARENT]->(:AzureManagementGroup)
(:AzureSubscription)-[:PARENT]->(:AzureManagementGroup)
```

[...]

PR #2842, opened 2026-05-28
feat(azure): Ingest Azure Management Groups and Management Group Subscriptions

shyammukund:
### Type of change

- [ ] Documentation update

### Summary

- Added Azure management-group ingestion with a new AzureManagementGroup node model.
- Split Azure orchestration so management groups sync as their own step before subscriptions.
- Management-group hierarchy is fetched from the tenant root subtree once, rather than recursively fetching every listed management group.
- Made subscription sync enrich base subscription rows with management-group parent data from the management-group subscriptions API.
- Partial management-group subscription enrichment preserves existing subscription-to-management-group parent edges from the graph for the failed management groups only.
- Base subscription ingestion remains authoritative, so subscriptions still load even if management-group subscription enrichment fails.
- If the management-group subscription enrichment call fails entirely, Cartography preserves existing subscription-to-management-group parent edges from the graph and refreshes them through the normal load path so standard cleanup does not delete them.

### Related issues or links

- Partially Addresses - https://github.com/cartography-cncf/cartography/issues/2831

### How was this tested?

Added integration coverage for:
   - management-group hierarchy happy path
   - subscription-from-management-group happy path
   - cleanup of stale hierarchy nodes and stale subscription-parent edges
   - subscription fallback behavior when management-group enrichment fails, including preserving existing subscription-parent edges on total enrichment failure
  
Added a unit test proving the tenant-root recursive hierarchy transform preserves the root-to-tenant and child-to-parent relationships, and skips duplicate management-group IDs if Azure repeats a node in one recursive payload.

#### General

#### Proof of functionality

image

### Notes for reviewers

- Base AzureSubscription ingestion brings in all subscriptions even the ManagementGroup based ones.  Management-group data is only used to enrich subscriptions with parent_management_group_id before load().
- Management-group sync raises on incomplete hierarchy fetches so cleanup does not run on partial data, but Azure orchestration catches that failure and continues base subscription/resource ingestion.
- On total management-group subscription enrichment failure, subscription sync reuses existing graph parent mappings for the subscriptions returned by the base subscription API so the normal AzureSubscription cleanup remains safe without custom cleanup logic.
- There is a new dependency on azure-mgmt-managementgroups, and both pyproject.toml and uv.lock were updated for it.


cartography/intel/azure/subscription.py
```diff
@@ -96,10 +169,33 @@ def cleanup(neo4j_session: neo4j.Session, common_job_parameters: Dict) -> None:
...
+
+    try:
+        management_group_subscriptions = get_azure_management_group_subscriptions(
+            credentials,
+        )
+        transformed_subscriptions = transform_azure_subscriptions(
+            subscriptions,
+            management_group_subscriptions,
+        )
+    except HttpResponseError as e:
```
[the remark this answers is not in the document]
shyammukund: Following guidelines about not overly defensive and handling failures that I haven't seen in a real sync.

cartography/intel/azure/management_groups.py
```diff
@@ -0,0 +1,209 @@
...
+            )
+            results.append(expanded_group.as_dict())
+        except HttpResponseError as e:
+            logger.warning(
+                "Failed to fetch expanded Azure management group '%s'. "
+                "Falling back to basic payload. Details: %s",
+                group_name,
+                e,
+            )
+            results.append(management_group.as_dict())
```
kunaals: this fallback makes the run look cleanup-safe even though we no longer have the recursive children or reliable parent details for this group. since `sync()` still runs stale cleanup afterward, a transient/permission failure on one expanded read can delete previously correct child management groups or `PARENT` edges. please mark this run incomplete and skip cleanup, or fail the management-group sync before cleanup.

cartography/intel/azure/subscription.py
```diff
@@ -72,6 +81,80 @@ def get_current_azure_subscription(
...
+                )
+            )
+        except HttpResponseError as e:
+            logger.warning(
+                "Failed to fetch Azure management-group subscriptions for '%s'. "
+                "Skipping this management group. Details: %s",
+                group_name,
+                e,
+            )
+            continue
```
kunaals, line 116: skipping one management group's subscription lookup preserves base subscription ingestion, but it also makes parent discovery partial. because subscription cleanup still runs later, subscriptions under the skipped group can lose existing `(:AzureSubscription)-[:PARENT]->(:AzureManagementGroup)` edges. please track partial enrichment and avoid cleaning parent edges from incomplete management-group subscription data.

cartography/intel/azure/management_groups.py
```diff
@@ -0,0 +1,209 @@
...
+            )
+
+
+@timeit
+def get_azure_management_groups(credentials: Credentials) -> list[dict[str, Any]]:
+    client = ManagementGroupsAPI(credentials.credential)
+    try:
+        management_groups = list(client.management_groups.list())
+    except HttpResponseError as e:
+        raise RuntimeError(
```
kunaals, line 116: this turns missing management-group read permission into a whole Azure sync failure because management groups run before subscription/resource ingestion. for existing deployments that do not yet grant `Management Group Reader`, the additive hierarchy feature would block all Azure ingestion. please make this cleanup-unsafe and let base Azure ingestion continue, unless this is intended as a breaking operational requirement.

cartography/intel/azure/management_groups.py
```diff
@@ -0,0 +1,209 @@
...
+            "Failed to fetch Azure management groups for tenant "
+            f"'{credentials.tenant_id}': {e}"
+        ) from e
+
+    results: list[dict[str, Any]] = []
+    for management_group in management_groups:
+        group_name = management_group.name
+        try:
+            expanded_group = client.management_groups.get(
```
kunaals: this does a recursive subtree fetch for every management group returned by `list()`. dedupe may preserve the final graph, but in cloud ingestion this kind of repeated provider API work is merge-blocking because it can drive rate limits and long sync runtimes for larger tenants. can we fetch the tenant-root subtree once, or otherwise avoid recursive reads for every group?

kunaals (changes requested):

cartography/intel/azure/management_groups.py
shyammukund: get_azure_management_groups() no longer falls back to the basic payload when an expanded recursive management-group read fails. It now raises a RuntimeError immediately, which prevents the sync from continuing with incomplete hierarchy data and avoids running cleanup on an incomplete result set.

cartography/intel/azure/subscription.py
shyammukund: get_azure_management_group_subscriptions() now tracks the mgmt groups whose subscription lookups failed and returns their management group ids alongside the successfully fetched subscriptions 

If any group lookups fail, get_existing_subscription_parent_mappings() queries the graph for existing (:AzureSubscription)-[:PARENT]->(:AzureManagementGroup) relationships and builds a lookup of the current parent mappings.

Now in transform_azure_subscriptions() we only reuse an existing graph parent mapping when that subscription’s existing parent management group is one of the groups that failed this run. That keeps the fallback scoped to the incomplete portion of enrichment instead of preserving stale parent edges globally.

cartography/intel/azure/management_groups.py
shyammukund: added exception that logs and skips management group sync if it can't fetch the complete hierarchy. Now when get_azure_management_groups raises on being unable to fetch the complete hierarchy it won't run cleanup.

cartography/intel/azure/management_groups.py
shyammukund: Changed to fetch the tenant root management group subtree once instead of recursively fetching every listed group.

kunaals (approved):
Re-reviewed the latest branch. The cleanup-safety and provider-call efficiency concerns from my previous review are addressed, and the focused Azure management-group/subscription test slice passes locally.

PR #2843, opened 2026-05-28
feat(kube): Add Kubernetes Tailscale endpoint links

kunaals:
### Type of change

### Summary
Adds model-backed Kubernetes links for Tailscale Kubernetes operator endpoints.

Kubernetes ingress and service `.ts.net` load balancer hostnames are normalized during transform and linked directly to matching `TailscaleDevice.name` values with schema-visible `USES_TAILSCALE_DEVICE` relationships. Tailscale operator proxy pods also expose the parent resource labels from the operator and create `TAILSCALE_PROXY_FOR` relationships to their parent `KubernetesIngress` or `KubernetesService`.

Existing AWS `USES_LOAD_BALANCER` behavior is unchanged, and Tailscale-only endpoints are not treated as internet exposure.

### Related issues or links
- Fixes #

### Breaking changes
None.

### How was this tested?
- Added new unit and integration tests using real API data shapes

#### General

#### Proof of functionality

### Notes for reviewers
The previous analysis-job matching approach has been removed. This version keeps the Cypher generated by the declarative model layer and uses Tailscale operator labels plus exact `.ts.net` FQDN matching to avoid broad short-name matching across tailnets.


cartography/models/kubernetes/pods.py
```diff
@@ -200,5 +257,7 @@ class KubernetesPodSchema(CartographyNodeSchema):
             KubernetesPodToServiceAccountRel(),
             KubernetesPodToSecretVolumeRel(),
             KubernetesPodToSecretEnvRel(),
+            KubernetesPodToTailscaleParentIngressRel(),
+            KubernetesPodToTailscaleParentServiceRel(),
```
jychp, line 261: blocking: `TAILSCALE_PROXY_FOR` is created during pod load, but the sync order in `cartography/intel/kubernetes/__init__.py` runs `sync_pods` (line 89) before `sync_services` (98) and `sync_ingress` (106). The matchers only link to already-existing `KubernetesIngress`/`KubernetesService` nodes, so on a first sync (or whenever a parent is newly created) the target does not exist yet and the edge is silently skipped until the next run. The integration test hides this by loading services/ingresses before pods.

cartography/intel/kubernetes/pods.py
```diff
@@ -182,6 +182,39 @@ def _format_pod_labels(labels: dict[str, str]) -> str:
...
+) -> dict[str, Any]:
+    labels = labels or {}
+    parent_type = labels.get("tailscale.com/parent-resource-type")
+    parent_namespace = labels.get("tailscale.com/parent-resource-ns")
+    parent_name = labels.get("tailscale.com/parent-resource")
+
+    if not parent_type or not parent_namespace or not parent_name:
+        return {
+            "tailscale_managed": labels.get("tailscale.com/managed") == "true",
+            "tailscale_parent_resource_type": parent_type,
```
jychp, line 196: nitpick: the early-return branch stores `tailscale_parent_resource_type` un-normalized while the main branch lowercases it (line 206). When `parent_type` is set but namespace/name is missing, the node gets mixed-case `tailscale_parent_resource_type`. Normalize here too, or move the `.lower()` up before the guard.

jychp (changes requested):

PR #2704, opened 2026-04-27
feat(gcp): ingest workload identity principal sets

kunaals:
### Type of change
- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

### Summary
Adds support for GCP Workload Identity Federation IAM members discovered through the existing Cloud Asset Inventory policy-binding sync.

This introduces `GCPExternalPrincipal` nodes for supported `principal://...` and `principalSet://...` workload identity pool members, keys all GCP principals with `principal_id`, and lets existing policy-binding and permission-relationship syncs create `HAS_ALLOW_POLICY` and `CAN_*` relationships for external principals.

### Related issues or links

- Fixes #

### Breaking changes

None expected. Existing email-backed GCP principals keep their email fields; `principal_id` is added as the stable matching key for GCP IAM permission relationships.

### How was this tested?

- `uv run --frozen pytest tests/unit/cartography/intel/gcp/test_policy_bindings.py tests/unit/cartography/intel/gcp/test_permission_relationships.py tests/integration/cartography/intel/gcp/test_policy_bindings.py tests/integration/cartography/intel/gcp/test_permission_relationships.py`
- `make test_lint`

#### General
- [x] I have read the [contributing guidelines](https://cartography-cncf.github.io/cartography/dev/developer-guide.html).
- [x] The linter passes locally (`make lint`).
- [x] I have added/updated tests that prove my fix is effective or my feature works.

#### Proof of functionality
- [ ] Screenshot showing the graph before and after changes.
- [x] New or updated unit/integration tests.

#### If you are adding or modifying a synced entity
- [ ] Included Cartography sync logs from a real environment demonstrating successful synchronization of the new/modified entity. Logs should show:
  - The sync job starting and completing without errors
  - The number of nodes/relationships created or updated
  - Example:
    ```
    INFO:cartography.intel.aws.ec2:Loading 42 EC2 instances for region us-east-1
    INFO:cartography.intel.aws.ec2:Synced EC2 instances in 3.21 seconds
    ```

#### If you are changing a node or relationship
- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules).
- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md).

#### If you are implementing a new intel module
- [ ] Used the NodeSchema [data model](https://cartography-cncf.github.io/cartography/dev/writing-intel-modules.html#defining-a-node).

### Notes for reviewers

[...]

PR #2844, opened 2026-05-29
feat(gcp): model WIF external principals

kunaals:
### Type of change
- [x] Documentation update

### Summary
Adds `GCPExternalPrincipal` nodes for Workload Identity Federation IAM members discovered through CAI policy-binding sync.

This keeps the existing WIF pool/provider model from #2799 intact, while preserving per-member selectors from `principal://...` and `principalSet://...` bindings. External principals now get `HAS_ALLOW_POLICY` edges to `GCPPolicyBinding` and can participate in GCP permission relationship syncs such as `CAN_READ` without re-keying existing email-backed GCP principals.

### Related issues or links

- Supersedes #2704

### Breaking changes

None expected. Existing `GCPPrincipal` email-based matching for users, groups, and service accounts is unchanged.

### How was this tested?

- Updated tests
- Read-only live GCP validation against existing IAM policy data: found 4 WIF `principalSet` members across 2 policy-binding resources, transformed them into 4 `GCPExternalPrincipal` nodes, and loaded 4 `HAS_ALLOW_POLICY` relationships in an isolated local Neo4j validation database. Selector coverage included 3 `attribute` selectors and 1 pool-wide selector.

#### General

#### Proof of functionality

### Notes for reviewers

This intentionally avoids the broader `principal_id` migration from #2704. Email-backed GCP principals continue to match by `email`; WIF external principals use a dedicated `GCPExternalPrincipal` matchlink keyed by the full IAM principal URI.


cartography/intel/gcp/policy_bindings.py
```diff
@@ -852,6 +983,21 @@ def _cleanup_applies_to_relationships(
     ).run(neo4j_session)
 
 
+def _cleanup_orphan_external_principals(neo4j_session: neo4j.Session) -> None:
+    GraphStatement(
+        """
+        // remove this after #2850 gives gcp principals a stable lifecycle.
+        MATCH (principal:GCPExternalPrincipal)
+        WHERE NOT (principal)-[:HAS_ALLOW_POLICY]->(:GCPPolicyBinding)
```
jychp, line 987: Blocking: this global orphan delete is unscoped and races with concurrent loads. `load_bindings`/`_load_*_bindings` and `cleanup()` both run inside `_GCP_POLICY_BINDINGS_GRAPH_SEMAPHORE`, which permits `GCP_POLICY_BINDINGS_GRAPH_CONCURRENCY = 4` simultaneous workers. `GCPExternalPrincipal` nodes are loaded before their `HAS_ALLOW_POLICY` edges, so one worker can `DETACH DELETE` another worker's freshly loaded principal before its edge is written, silently dropping that WIF member nondeterministically. Scope this delete by `_sub_resource`/`UPDATE_TAG`, or run it once after all per-project work completes instead of inside the per-project cleanup.

cartography/intel/gcp/permission_relationships.py
```diff
@@ -22,6 +25,14 @@
 GCP_BIGQUERY_TABLE_PERMISSION_TABLE_BATCH_SIZE = 1000
 GCP_BIGQUERY_TABLE_PERMISSION_PRINCIPAL_BATCH_SIZE = 100
 GCPPrincipalPermissionContext = dict[str, dict[str, dict[str, Any]]]
+_WIF_PRINCIPAL_PREFIXES = (
```
jychp: `_WIF_PRINCIPAL_PREFIXES` and the prefix check duplicate the identical `_WIF_PRINCIPAL_PREFIXES` tuple in `cartography/intel/gcp/policy_bindings.py:483`. Import the existing constant (or move it to a shared module) so the two WIF prefix lists cannot drift.

jychp (changes requested):

cartography/intel/gcp/policy_bindings.py
kunaals: fixed

cartography/intel/gcp/permission_relationships.py
kunaals: fixed

cartography/models/gcp/policy_bindings.py
```diff
@@ -170,11 +191,57 @@ class GCPFolderPolicyBindingSchema(CartographyNodeSchema):
...
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "RESOURCE"
+    properties: GCPExternalPrincipalToWifPoolRelProperties = (
+        GCPExternalPrincipalToWifPoolRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class GCPExternalPrincipalSchema(CartographyNodeSchema):
+    label: str = "GCPExternalPrincipal"
```
jychp, line 237: Most of our rules / enrichment use GCPPrincipal extra label.
You should either add it this extra label (preferred IMO) or update rules/analysis jobs.

jychp (changes requested):

PR #2856, opened 2026-06-02
feat(entra)/fix(gcp): sync Entra directory roles and link GCP instances to their service accounts

jychp:
### Type of change
- [x] Documentation update

### Summary

Two independent improvements, one commit each.

**GCP: link instances to their service accounts.** `GCPInstance` already exposed `service_account_email`/`service_account_scopes` from the instance transform, but the data model had no relationship to the service account the instance runs as, so the attached identity could not be traversed. This adds a `(:GCPInstance)-[:RUNS_AS]->(:GCPServiceAccount)` relationship (matched on the service account email, which is indexed), mirroring the existing `GCPCloudFunction->GCPServiceAccount` edge. Because the edge is created when instances are loaded and matches existing `GCPServiceAccount` nodes, the IAM block now runs before `compute.sync()` within `_sync_project_resources()` (IAM cleanup stays at the end of the project, so deletion semantics are unchanged). Under selective sync (e.g. `--gcp-requested-syncs compute`) IAM may not run, so the edge is only created for service accounts already present in the graph; this is documented inline.

**Microsoft/Entra: sync directory role assignments.** The Entra module did not sync directory (Entra ID) role assignments, so the graph could not show which accounts actually hold privileged roles such as Global Administrator. This adds a directory role sync built on the Microsoft Graph unified RBAC endpoints:
- `EntraRoleDefinition` from `roleManagement/directory/roleDefinitions`
- `EntraRoleAssignment` (junction node) from `roleManagement/directory/roleAssignments`

with relationships `(:EntraUser|EntraGroup|EntraServicePrincipal)-[:HAS_ROLE]->(:EntraRoleAssignment)-[:ASSIGNED_TO]->(:EntraRoleDefinition)`, each scoped to its `EntraTenant`. The sync is wired into `start_entra_ingestion` after app role assignments.

### Related issues or links

N/A

### How was this tested?

- `make test_lint` passes locally (isort, black, flake8, mypy, pre-commit hooks).
- New/updated integration tests (require local Neo4j):
  - `tests/integration/cartography/intel/gcp/test_compute.py` - asserts the populated `service_account_email` property and the `RUNS_AS` edge, the real `iam.sync()` -> `compute.sync()` ordering producing the edge, and the selective-sync case (property set, no edge when IAM did not run).
  - `tests/integration/cartography/intel/microsoft/entra/test_directory_roles.py` - asserts `EntraRoleDefinition`/`EntraRoleAssignment` nodes and the `RESOURCE`, `ASSIGNED_TO`, and `HAS_ROLE` relationships for user/group/service-principal principals.
- New unit test `tests/unit/cartography/intel/gcp/test_gcp_init_sync_order.py` asserts IAM syncs before Compute in `_sync_project_resources`.
- Spot-check queries after a sync:
  - `MATCH (i:GCPInstance)-[:RUNS_AS]->(sa:GCPServiceAccount) RETURN count(*)`
  - `MATCH (p)-[:HAS_ROLE]->(a:EntraRoleAssignment)-[:ASSIGNED_TO]->(rd:EntraRoleDefinition) RETURN labels(p), rd.display_name LIMIT 10`

#### General

#### Proof of functionality

### Notes for reviewers

The two commits are independent (GCP compute vs Microsoft/Entra) and can be reviewed separately. For the Entra junction node, only the principal relationship whose target node actually exists will be created, so a single assignment links to exactly one of user/group/service principal.


cartography/intel/microsoft/entra/directory_roles.py
```diff
@@ -0,0 +1,210 @@
...
+    page = await call_with_retries(
+        lambda: client.role_management.directory.role_assignments.get(),
+    )
+    while page:
+        if page.value:
+            for role_assignment in page.value:
+                yield role_assignment
+        if not page.odata_next_link:
+            break
+        page = await call_with_retries(
```
kunaals: Nit/non-blocking: roleAssignments can be large, and we now have `get_paginated_values_with_expired_page_retry()` for Graph `Directory_ExpiredPageToken` cases. Consider using that helper here (and for roleDefinitions) so an expired nextLink does not cause the optional directory-role sync to be skipped by the outer `APIError` handler. Not approval-blocking; tests and live validation look good.

cartography/intel/microsoft/entra/__init__.py
```diff
@@ -156,6 +158,27 @@ async def main() -> None:
             common_job_parameters,
         )
```
kunaals, line 159: Nit/non-blocking: this is useful for tenants missing `RoleManagement.Read.Directory`, but it catches every Graph `APIError`, including non-permission failures. A future tightening to only swallow expected auth/permission errors would make real Graph failures more visible while preserving the optional dataset behavior.

kunaals (approved):

PR #2857, opened 2026-06-02
feat(gcp): resolve WIF pool permissions into capability edges

jychp:
### Type of change
- [ ] Documentation update

### Summary

A GCP Workload Identity Pool's federated identity (a `principalSet://...workloadIdentityPools/<POOL>` member) can be granted GCP IAM roles, but Cartography never resolved those grants into capability edges. The pool was a graph dead-end: it had `(GCPWorkloadIdentityPool)-[:HAS_ALLOW_POLICY]->(GCPPolicyBinding)`, but no `CAN_READ`/`CAN_WRITE`/... edges to concrete resources. Anyone reasoning over WIF-based access (e.g. an external identity that federates into the pool) could not see which GCP resources that identity can actually reach.

Root cause: the permission-resolution engine built its principal context only from `binding["members"]` (principal emails) and linked via a MatchLink whose source is `GCPPrincipal` matched on `email`. The WIF federated identity lives in `binding["wif_pools"]` (pool resource ids), and the pool is a `GCPWorkloadIdentityPool` node matched on `id`, so it was invisible to the engine.

This PR:

- **Resolves WIF pool capabilities.** `build_principals_from_policy_bindings` now also returns a permission context keyed by pool id (from each binding's `wif_pools`), sharing the same compiled assignments. `permission_relationships.sync` runs a second MatchLink pass with a `GCPWorkloadIdentityPool` source matched on `id`, reusing the existing scope-aware engine unchanged:

  ```
  (GCPWorkloadIdentityPool)-[:CAN_READ|CAN_WRITE|CAN_DECRYPT|CAN_IMPERSONATE|...]->(GCPResource)
  ```

  Because inherited org/folder bindings already flow through this engine and `resolve_gcp_scope` maps org/folder/project scope to `project/{id}/*`, org-scoped grants (e.g. `roles/storage.objectViewer` at org level) resolve to every in-scope project resource, exactly as they do for a normal principal. The pool pass always runs (even with an empty context) so its matchlink cleanup prunes pool edges left by a prior run once the last WIF grant is removed, mirroring the `GCPPrincipal` pass.

- **Adds a cross-cloud trust edge.** For AWS-protocol providers, a match-only relationship links the provider to the trusted AWS account when it is also synced (mirroring the existing `KubernetesServiceAccount`->`AWSRole` precedent). No phantom `AWSAccount` nodes are created, and OIDC/SAML providers (null `aws_account_id`) never match:

  ```
  (GCPWorkloadIdentityProvider{protocol: 'AWS'})-[:TRUSTS_AWS_ACCOUNT]->(AWSAccount)
  ```

Limitations (called out for reviewers): conditional IAM bindings remain unsupported (pre-existing engine limitation, skipped for pools as for principals), and `attribute_condition` is preserved unparsed, so the trust edge is account-level; consumers can lower confidence when `attribute_condition` is non-null.

### Related issues or links

N/A

### How was this tested?

- New unit tests: pool-keyed context from `wif_pools` (conditional / missing-role grants skipped); the `sync` pool pass builds a `GCPWorkloadIdentityPool` matchlink matched on `id`; the pool cleanup always runs (even with no WIF grants) so stale edges are pruned.
- New integration tests: a pool granted a role at project scope gets `CAN_READ` to the in-scope buckets; a stale pool `CAN_READ` edge is pruned when the WIF grant is gone; an AWS-protocol provider gets `TRUSTS_AWS_ACCOUNT` to the matching `AWSAccount` (and creates no phantom node, and OIDC providers get no edge).
- `make test_lint` (black, flake8, mypy, isort, pyupgrade) passes.
- Full GCP unit + integration suite: 387 passed.

#### General

#### Proof of functionality

### Notes for reviewers

The design intentionally reuses the existing scope-aware resolution engine rather than duplicating it: the engine is generic over a "principals" dict keyed by a string that becomes the matchlink source key, so the only additions are (1) a second context keyed by pool id and (2) a second matchlink whose source is `GCPWorkloadIdentityPool` matched on `id`. Org/project-scope inheritance therefore needs no special handling.


kunaals:
i think we should merge #2844 first since it models the actual WIF IAM members as `GCPExternalPrincipal` nodes using the full `principal://` and `principalSet://` selectors. that keeps the important detail around principal sets, instead of collapsing everything down to the pool

then for this pr, we can keep the `GCPWorkloadIdentityProvider -[:TRUSTS_AWS_ACCOUNT]-> AWSAccount` piece, but probably drop the pool-level `GCPWorkloadIdentityPool -[:CAN_*]-> resource` edges as they're very coarse

PR #2866, opened 2026-06-05
fix: remove deprecated AWS Organizations Status field

### LINKED ISSUE #2861: chore(aws): remove deprecated `status` field from Organizations sync (deadline: 2026-09-09)
## Summary

In `cartography/intel/aws/organizations.py` line 189,
there is a TODO comment indicating that the `status`
field should be removed after AWS retires it on
2026-09-09.

## Location

- File: `cartography/intel/aws/organizations.py`
- Line: 189

## Current Code

The current code contains:
# TODO: Remove status after AWS retires it on 2026-09-09.

## Proposed Fix

Remove the deprecated `status` field from the
Organizations data ingestion before the AWS
retirement deadline. This includes:

- Removing the field from data transformation
- Updating related Neo4j load queries
- Updating any tests that reference this field

## Why This Matters

This is time-sensitive — if not addressed before
2026-09-09, the field will become invalid and
could cause sync failures for users running
Cartography against AWS Organizations.

## Additional Notes

I found this while exploring the codebase as a
new contributor. I'd like to work on this fix
if maintainers agree with the approach.

g-k-s-03:
Fixes #2861. Removes the Status fallback from _get_account_state ahead of AWS retirement on 2026-09-09. All 13 unit tests pass.


g-k-s-03:
Hi @jychp  @achantavy  I have resolved the DCO sign-off issue and force pushed the branch. The commit now also includes the removal of status: PropertyRef from AWSOrganizationAccountNodeProperties in account.py, which was missing from the original commit. All 13 unit tests pass locally. Could you please approve the workflow run so the CI pipeline can execute? Happy to address any review feedback. Thank you.

g-k-s-03:
Hi @achantavy — DCO is now fixed, and the branch is clean. Ready for review whenever you get a chance

cartography/intel/aws/organizations.py
```diff
@@ -565,7 +563,6 @@ def cleanup_stale_aws_account_organization_metadata(
         SET account.arn = null,
             account.email = null,
             account.state = null,
```
jychp, line 565: question: removing `account.status = null` here (and the property from the model) means graphs that already ingested `status` keep stale values indefinitely - nothing nulls them on upgrade. If the field is truly retired, add a one-time migration/cleanup that clears `account.status`; otherwise document that stale values persist.

jychp (changes requested):
Are we sure the current dependency requirement expose the `State` property ?
If not you should also bump the minimum dependency contract.

Additional question on stale data.

g-k-s-03:
Hi @jychp  , thanks for catching that. I've restored account.status = null to the cleanup SET block so existing graph nodes have their stale status values cleared on the next sync. Regarding the dependency contract no bump is needed since we are removing the field rather than adding one. The AWS retirement is confirmed effective September 9, 2026. Happy to add a comment to make the stale-data handling more explicit if useful.

cartography/intel/aws/organizations.py
```diff
@@ -145,7 +145,7 @@ def get_aws_organization_accounts_for_parent(
 
 
 def _get_account_state(account: dict[str, Any]) -> str | None:
-    return account.get("State") or account.get("Status")
+    return account.get("State")
```
jychp, line 148: **Blocking — removing the `Status` fallback breaks users on the pinned minimum botocore.**

`State` is now the sole account-state signal, but `pyproject.toml` pins `botocore>=1.18.1` (Aug 2020). botocore parses API responses against its per-version service model and drops any member the model doesn't declare, so `State` only reaches this code if the installed botocore is new enough to know about it.

AWS added the `State` field to `ListAccounts` / `ListAccountsForParent` / `DescribeAccount` in September 2025; per AWS you need "an SDK version released after September 9, 2025" (AWS CLI v2 >= 2.29.0) to see it. `botocore>=1.18.1` predates that by ~5 years, so on the floor every account resolves to `state=None`, `_has_account_state_signal` is False for all, and `sync_aws_organization` silently skips the entire org as INCOMPLETE (the exact path asserted by `test_sync_aws_organization_returns_incomplete_when_account_state_is_unknown`).

Fix: bump the minimum botocore to the first release that ships the `State` model change (post 2025-09-09), or keep the `Status` fallback until that floor is raised. "Removing not adding" doesn't apply — the break is `State` availability, not `Status`.

Refs: https://aws.amazon.com/about-aws/whats-new/2025/09/aws-organizations-provides-account-state-information-member-accounts/ | https://docs.aws.amazon.com/organizations/latest/APIReference/API_Account.html

cartography/models/aws/account.py
```diff
@@ -31,7 +31,6 @@ class AWSOrganizationAccountNodeProperties(CartographyNodeProperties):
     arn: PropertyRef = PropertyRef("arn", extra_index=True)
     email: PropertyRef = PropertyRef("email")
     state: PropertyRef = PropertyRef("state")
```
jychp, line 33: **Blocking — stale `status` is never cleared on accounts still in the org.** Dropping `status` from the schema means the loader stops writing it, but existing values persist. `cleanup_stale_aws_account_organization_metadata` only nulls `status` for accounts NOT in `current_account_ids` (departed accounts), so nodes still in the org keep whatever earlier syncs wrote. The reply that "existing nodes get their stale status cleared on next sync" is only true for departed accounts, not the common case. Add a one-time SET-to-null over in-org account nodes, or explicitly document that stale `status` lingers on active accounts.

jychp (changes requested):

PR #2871, opened 2026-06-05
feat(rules): Add compliance_only flag

kunaals:
### Type of change
- [x] Documentation update

### Summary
Adds explicit rule metadata so catalog consumers can distinguish compliance-only hygiene/checklist controls from standalone actionable rules without relying on rule names or framework heuristics.

The new `Rule.compliance_only` flag defaults to `False`. Rules with `compliance_only=False` remain appropriate for standalone rule catalogs, even when they are also mapped to compliance frameworks.

Framework-prefixed rule definitions are marked `compliance_only=True`. This covers rule IDs/names that are authored as specific framework controls, such as `cis_*` rules and `nist_ai_*` / `NIST AI RMF: ...` rules. These are primarily compliance, governance, or hygiene checklist controls and should appear through framework views rather than the general standalone rules catalog.

Regular rule definitions that are mapped into frameworks remain standalone-visible by default. If a framework-prefixed control should become a standalone detection, it can be renamed/generalized into a regular rule and then mapped into the relevant framework.

Rule execution and framework mappings are unchanged. The serialized rule result now includes `rule_compliance_only` alongside the existing rule tags and framework metadata.

### Breaking changes

None.

### How was this tested?

- `uv run --frozen pytest -q tests/unit/rules/test_compliance_only.py tests/unit/rules/test_model.py tests/unit/rules/test_runners.py`
- `make test_lint`
- `make test_unit`

#### General

#### Proof of functionality

### Notes for reviewers

This is metadata-only for catalog surfaces. Frameworks still wrap rules the same way, and framework detail pages can continue to use the existing framework mappings. The intended boundary is that framework-prefixed controls are compliance-only, while reusable regular detections can remain visible in standalone catalogs even when mapped to one or more frameworks.

### DIFF
--- a/cartography/rules/data/rules/cis_aws_networking.py
+++ b/cartography/rules/data/rules/cis_aws_networking.py
@@ -10,7 +10,6 @@ Facts within a Rule are provider-specific implementations of the same concept.
 
 from cartography.rules.data.frameworks.cis import cis_aws
 from cartography.rules.data.frameworks.iso27001 import iso27001_annex_a
-from cartography.rules.spec.model import Catalog
 from cartography.rules.spec.model import Fact
 from cartography.rules.spec.model import Finding
 from cartography.rules.spec.model import Maturity
@@ -93,7 +92,7 @@ cis_aws_6_1_1_ebs_encryption = Rule(
     tags=("networking", "ebs", "encryption", "stride:information_disclosure"),
     version="1.0.0",
     references=CIS_REFERENCES,
-    catalog_visibility=(Catalog.COMPLIANCE,),
+    compliance_only=True,
     frameworks=(
         cis_aws("6.1.1"),
         iso27001_annex_a("8.24"),
@@ -196,7 +195,7 @@ cis_aws_6_1_2_cifs_restricted = Rule(
     ),
     version="1.0.0",
     references=CIS_REFERENCES,
-    catalog_visibility=(Catalog.COMPLIANCE,),
+    compliance_only=True,
     frameworks=(
         cis_aws("6.1.2"),
         iso27001_annex_a("8.20"),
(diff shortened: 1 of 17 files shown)


cartography/rules/data/rules/cis_aws_networking.py
```diff
@@ -299,6 +301,7 @@ class RemoteAdminIpv4Output(Finding):
     ),
     version="1.0.0",
     references=CIS_REFERENCES,
+    compliance_only=True,
```
jychp, line 304: blocking: `cis_aws_6_3_remote_admin_ipv4` detects SSH/RDP (22/3389) open to 0.0.0.0/0 - an immediate, actionable internet-exposure finding, not a compliance checklist item. `compliance_only=True` suppresses it from standalone actionable catalogs. Same applies to `default_sg_allows_traffic` and `cloudsql_public_ip` in this PR. Leave these `False`.

tests/unit/rules/test_compliance_only.py
```diff
@@ -0,0 +1,107 @@
...
+    assert _serialized_compliance_only(object_storage_public) is False
+
+
+def test_all_framework_prefixed_rules_are_compliance_only() -> None:
+    prefixed_rules = [
+        rule for rule in RULES.values() if rule.id.startswith(("cis_", "nist_ai_"))
+    ]
+
+    assert prefixed_rules
+    assert all(rule.compliance_only for rule in prefixed_rules)
```
jychp, line 81: blocking: this test hard-codes the prefix heuristic as an invariant (every `cis_*`/`nist_ai_*` rule must be `compliance_only`). That cements the exact name-based classification the flag was introduced to avoid, and makes it impossible to ever expose a `cis_`-prefixed rule as a standalone detection without renaming it. If per-rule classification is adopted, this test must be dropped.

jychp (changes requested):

PR #2873, opened 2026-06-05
feat(ontology): add Subnet, VirtualNetwork and Snapshot semantic labels

jychp:
### Type of change
- [x] Documentation update

### Summary
Adds three new cross-provider ontology **semantic-label categories** so network and storage-snapshot resources become queryable through the abstract ontology layer, regardless of provider:

- **`Subnet`** — `EC2Subnet` (AWS), `GCPSubnet` (GCP), `AzureSubnet` (Azure). Normalized `_ont_*`: `name`, `cidr_block`, `availability_zone`, `region` (where the provider exposes them).
- **`VirtualNetwork`** — `AWSVpc`, `GCPVpc`, `AzureVirtualNetwork`. Normalized `_ont_*`: `name`, `cidr`, `region`.
- **`Snapshot`** — `EBSSnapshot`, `RDSSnapshot` (AWS), `AzureSnapshot`, `ScalewayVolumeSnapshot`. Normalized `_ont_*`: `name`, `encrypted`, `public`, `source_id`, `created_at`, `region`.

Each category adds a mapping under `cartography/models/ontology/mapping/data/`, registers it in `SEMANTIC_LABELS_MAPPING`, and attaches `ExtraNodeLabels` to every source schema class that shares the label (including the composite/stub schemas, e.g. all `EC2Subnet` variants).

Fields are only mapped where the provider actually exposes them; gaps are documented inline rather than approximated. Notably:
- `Subnet._ont_is_public` is **not** mapped: no provider stores a faithful public/private flag on the subnet node (it requires route-table/IGW or NSG analysis).
- `VirtualNetwork._ont_cidr` is omitted for GCP (CIDRs live on subnets) and Azure (address space stored on subnets); `region` is omitted for GCP (VPCs are global).
- `Snapshot._ont_encrypted` is skipped for Azure (the available field is the legacy ADE flag, which would mislabel SSE-encrypted snapshots as unencrypted, mirroring the existing `AzureDisk` treatment in `blockstorage`); Azure/Scaleway also lack public/source/created_at on the snapshot node.

The constraint mechanism and CI guard already exist (`constraints.py`, `test_ontology_rel_constraints.py`), and semantic-label categories are exempt from the canonical-node id requirement, so no foundation changes were needed.

### Related issues or links

N/A

### Breaking changes

None. This only adds extra Neo4j labels and `_ont_*` properties to existing nodes.

### How was this tested?
- Unit: `pytest tests/unit/cartography/intel/ontology/ tests/unit/cartography/graph/test_ontology_rel_constraints.py`
- Integration (against a local Neo4j): new `_ont_*`/label assertions plus the full touched suites:
  `pytest tests/integration/cartography/intel/aws/ec2/test_ec2_subnets.py tests/integration/cartography/intel/aws/ec2/test_vpc.py tests/integration/cartography/intel/aws/ec2/test_ec2_snapshots.py tests/integration/cartography/intel/aws/test_rds.py tests/integration/cartography/intel/gcp/test_compute.py`
- `pre-commit` (isort/black/flake8/mypy) passes on all changed files.

#### General

#### Proof of functionality

### Notes for reviewers
- The semantic label is attached via `ExtraNodeLabels` on **every** schema class that shares a node label (the composite/stub pattern). The querybuilder looks up the ontology mapping by `node_schema.label` and skips mapped fields a given schema does not define, so partial/stub writers never null-clobber `_ont_*` set by the full writer.
- `GCPSubnetStubSchema` deliberately does **not** carry the `Subnet` label: it only holds `partial_uri`, so it would otherwise surface in `(:Subnet)` queries with a null `_ont_name`. The full `GCPSubnetSchema` attaches the label once real subnet data syncs. AWS stubs are unaffected because their ontology name resolves from `id` (always present).

### DIFF
--- a/cartography/models/aws/ec2/subnet_instance.py
+++ b/cartography/models/aws/ec2/subnet_instance.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 from cartography.models.core.common import PropertyRef
 from cartography.models.core.nodes import CartographyNodeProperties
 from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.nodes import ExtraNodeLabels
 from cartography.models.core.relationships import CartographyRelProperties
 from cartography.models.core.relationships import CartographyRelSchema
 from cartography.models.core.relationships import LinkDirection
@@ -66,6 +67,7 @@ class EC2SubnetInstanceSchema(CartographyNodeSchema):
 
     label: str = "EC2Subnet"
     properties: EC2SubnetInstanceNodeProperties = EC2SubnetInstanceNodeProperties()
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Subnet"])
     sub_resource_relationship: EC2SubnetToAWSAccountRel = EC2SubnetToAWSAccountRel()
     other_relationships: OtherRelationships = OtherRelationships(
         [
--- a/cartography/models/ontology/mapping/__init__.py
+++ b/cartography/models/ontology/mapping/__init__.py
@@ -88,6 +88,7 @@ from cartography.models.ontology.mapping.data.useraccounts import (
     USERACCOUNTS_ONTOLOGY_MAPPING,
 )
 from cartography.models.ontology.mapping.data.users import USERS_ONTOLOGY_MAPPING
+from cartography.models.ontology.mapping.data.vpcs import VPCS_ONTOLOGY_MAPPING
 from cartography.models.ontology.mapping.specs import OntologyMapping
 from cartography.models.ontology.mapping.specs import OntologyNodeMapping
 from cartography.models.ontology.package import PackageSchema
@@ -138,6 +139,7 @@ SEMANTIC_LABELS_MAPPING: dict[str, dict[str, OntologyMapping]] = {
     "secrets": SECRETS_ONTOLOGY_MAPPING,
     "securityissues": SECURITY_ISSUES_ONTOLOGY_MAPPING,
     "subnets": SUBNETS_ONTOLOGY_MAPPING,
+    "vpcs": VPCS_ONTOLOGY_MAPPING,
     "thirdpartyapps": THIRDPARTYAPPS_ONTOLOGY_MAPPING,
     "tenants": TENANTS_ONTOLOGY_MAPPING,
     "serviceaccounts": SERVICEACCOUNTS_ONTOLOGY_MAPPING,
--- a/cartography/models/ontology/mapping/__init__.py
+++ b/cartography/models/ontology/mapping/__init__.py
@@ -79,6 +79,7 @@ from cartography.models.ontology.mapping.data.security_issues import (
 from cartography.models.ontology.mapping.data.serviceaccounts import (
     SERVICEACCOUNTS_ONTOLOGY_MAPPING,
 )
+from cartography.models.ontology.mapping.data.subnets import SUBNETS_ONTOLOGY_MAPPING
 from cartography.models.ontology.mapping.data.tenants import TENANTS_ONTOLOGY_MAPPING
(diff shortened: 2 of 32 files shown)


cartography/models/aws/ec2/subnet_instance.py
```diff
@@ -3,6 +3,7 @@
 from cartography.models.core.common import PropertyRef
 from cartography.models.core.nodes import CartographyNodeProperties
 from cartography.models.core.nodes import CartographyNodeSchema
+from cartography.models.core.nodes import ExtraNodeLabels
```
kunaals, line 6: i’m a little worried about putting the `subnet` semantic label on the partial ec2 subnet schemas. these paths only know the subnet id and sometimes region, so `(:subnet)` queries can include aws subnets with no `_ont_cidr_block` or `_ont_availability_zone`. for a security workflow, “show me all subnets and their cidrs / azs” becomes noisier than expected. could we either only label the full `ec2subnetschema`, or add a test/doc note that these partial aws subnet nodes are intentionally allowed in the semantic label?

cartography/models/ontology/mapping/data/snapshots.py
```diff
@@ -0,0 +1,114 @@
+from cartography.models.ontology.mapping.specs import OntologyFieldMapping
+from cartography.models.ontology.mapping.specs import OntologyMapping
+from cartography.models.ontology.mapping.specs import OntologyNodeMapping
+
+# Snapshot fields:
+# _ont_name - The name/identifier of the snapshot
+# _ont_encrypted - Whether the snapshot is encrypted at rest
+# _ont_public - Whether the snapshot is publicly shared (a known attack vector)
```
kunaals, line 8: can we make the partial coverage here more explicit in the ontology docs? `_ont_public` and `_ont_encrypted` are really useful fields, but right now they’re aws-only for snapshots. a security engineer may naturally write something like `coalesce(s._ont_encrypted, false) = false` and accidentally flag every azure/scaleway snapshot as unencrypted, when the actual state is “unknown / not modeled.” a short note that absence is not false would make this safer to use.

cartography/models/ontology/mapping/data/subnets.py
```diff
@@ -0,0 +1,85 @@
...
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="name", required=True
+                ),
+                OntologyFieldMapping(
+                    ontology_field="cidr_block", node_field="address_prefix"
+                ),
+                # _ont_availability_zone: not mapped. Azure subnets are not
+                # zone-scoped.
+                # _ont_region: not mapped. The region/location is stored on the
```
kunaals, line 74: worth documenting this region gap somewhere. azure subnets won’t have `_ont_region`, so a region-scoped cross-cloud query like `match (s:subnet) where s._ont_region = $region` silently drops azure. that may be okay since the location lives on the parent vnet, but users should know they need to traverse through `azurevirtualnetwork` for azure subnet region filtering.

cartography/models/ontology/mapping/__init__.py
```diff
@@ -79,6 +79,10 @@
 from cartography.models.ontology.mapping.data.serviceaccounts import (
     SERVICEACCOUNTS_ONTOLOGY_MAPPING,
 )
+from cartography.models.ontology.mapping.data.snapshots import (
```
kunaals, line 82: can we add a little more end-to-end coverage for the non-aws mappings? the new tests cover aws subnet/vpc/ebs snapshot, but the gcp, azure, scaleway, and rds snapshot ontology mappings are mostly protected by field-name inspection rather than sync tests. i’d especially like a regression test for the gcp subnet stub staying unlabeled, since that’s subtle and easy to accidentally “fix” later.

kunaals (changes requested):

kunaals (approved):

PR #2874, opened 2026-06-07
chore(gcp): finish GCP HttpError classifier migration

jychp:
### Type of change
- [ ] Documentation update

### Summary
This follow-up completes the GCP HttpError classifier migration from #2424 by moving remaining GCP modules away from ad-hoc helper calls and direct status checks.

The PR updates the canonical classifier to distinguish billing-disabled errors, then routes remaining BigQuery, Bigtable, Cloud SQL, Cloud Run, Compute-adjacent, KMS, Secret Manager, Storage, Backend Service, Instance Group, and Workload Identity error handling through classify_gcp_http_error() while preserving the existing skip/re-raise behavior.

### Related issues or links
- Part of #2424

### Breaking changes
None.

### How was this tested?
- UV_CACHE_DIR=.uv-cache uv run --frozen pytest tests/unit/cartography/intel/gcp
- UV_CACHE_DIR=.uv-cache uv run --frozen black --check cartography/intel/gcp tests/unit/cartography/intel/gcp/test_util.py tests/unit/cartography/intel/gcp/test_cloudrun_util.py
- git diff --check

ruff was not available in the frozen uv environment: Failed to spawn: ruff.

#### General
- [ ] The linter passes locally (make test_lint).

#### Proof of functionality

- [ ] Updated the schema documentation.

- [ ] Used the NodeSchema data model.

### Notes for reviewers
This intentionally leaves the low-level parsing helpers in cartography/intel/gcp/util.py; callers outside the utility now use the canonical classifier instead.


cartography/intel/gcp/bigquery_dataset.py
```diff
@@ -7,8 +7,8 @@
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.gcp.util import classify_gcp_http_error
 from cartography.intel.gcp.util import gcp_api_execute_with_retry
```
kunaals, line 10: this changes exhausted 403 quota/rate-limit responses from "skip after retries" into a hard failure, since the classifier returns `transient` and this handler no longer catches it. i think this should preserve the old 403 skip behavior here.

cartography/intel/gcp/compute.py
```diff
@@ -19,7 +19,6 @@
 from cartography.intel.gcp.labels import sync_labels
 from cartography.intel.gcp.util import classify_gcp_http_error
 from cartography.intel.gcp.util import gcp_api_execute_with_retry
-from cartography.intel.gcp.util import is_permission_denied_error
 from cartography.intel.gcp.util import parse_compute_full_uri_to_partial_uri
```
kunaals, line 23: this now treats any unrecognized 403 as permission-denied and skips compute sync. the old helper only skipped known permission-denied shapes, so malformed/generic 403s now get swallowed instead of surfacing.

kunaals (changes requested):

cartography/intel/gcp/workload_identity.py
```diff
@@ -9,6 +9,7 @@
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.gcp.util import classify_gcp_http_error
 from cartography.intel.gcp.util import gcp_api_execute_with_retry
```
kunaals, line 12: this still changes the old behavior for quota/rate-limit 403s. before, wif skipped any 403; now quota 403s classify as `transient` and fall through to `raise`. should this use the same `include_transient_403=True` helper as the bigquery paths?

cartography/intel/gcp/storage.py
```diff
@@ -59,15 +58,15 @@ def get_gcp_buckets(storage: Resource, project_id: str) -> Dict:
         first_response.pop("nextPageToken", None)
         return first_response
     except HttpError as e:
-        reason = get_error_reason(e)
```
kunaals, line 62: this still broadens the old permission check. before we only skipped known permission-denied reasons via `is_permission_denied_error(e)`; now any generic/malformed 403 classifies as `forbidden` and gets swallowed.

cartography/intel/gcp/secretsmanager.py
```diff
@@ -48,30 +45,30 @@ def get_secrets(secretmanager: Resource, project_id: str) -> List[Dict]:
             )
         return secrets
```
kunaals, line 47: same concern here: `category == "forbidden"` catches generic/malformed 403s that the old `is_permission_denied_error(e)` path would not have skipped. can we preserve the narrower permission-denied check?

kunaals (changes requested):

kunaals (approved):

PR #2882, opened 2026-06-08
fix(aws): draw permission edges for object-level S3 grants (#1639)

### LINKED ISSUE #1639: AWS permission relationships: S3 GetObject does not work if `resource` refers to path under bucket

jychp:
### Type of change
- [ ] Documentation update

### Summary

Object-level S3 grants did not produce a permission edge to the S3Bucket node.

A policy granting `s3:GetObject` on `arn:aws:s3:::my-bucket/*` should draw a
`CAN_READ` edge to the bucket node `arn:aws:s3:::my-bucket`, but it did not.
The resource clause `arn:aws:s3:::my-bucket/*` compiles to the regex
`arn:aws:s3:::my-bucket/.*`, which requires the trailing `/` to be present, so
`re.fullmatch` against the bare bucket ARN fails and no edge is created.

This adds `evaluate_resource_clause`, which (in addition to a direct match)
also tests the clause against the resource ARN with a trailing path separator
appended, so `arn:aws:s3:::my-bucket/.*` matches `arn:aws:s3:::my-bucket/`.
It is used for both `resource` and `notresource` matching so object-level
exclusions are honored too.

The trailing-slash alternate is scoped to S3 ARNs (`arn:aws:s3:::` prefix) on
purpose. S3 is the only AWS resource family where object ARNs nest under the
resource node as `<bucket-arn>/<key>`. For other families a `/` is part of the
resource name itself (e.g. an IAM role `arn:aws:iam::000000000000:role/MyRole`),
so a grant on `.../MyRole/*` targets a different role path, not an object under
the role, and must not match the parent node. This avoids false edges such as
`CAN_PASS_ROLE`.

### Related issues or links

- Fixes #1639

### How was this tested?

`make test_lint` and the unit suite pass. Added unit tests covering:
- an object-level grant (`arn:aws:s3:::my-bucket/*`) matching the bucket node;
- the shared-prefix non-match (`my-bucket/*` does not match `my-bucket-other`);
- the S3 scoping guard (an `iam:PassRole` grant on `.../PassableRole/*` does NOT
  match the parent role node);
- an object-level `NotResource` exclusion applying to the bucket node.

```
$ uv run python -m pytest tests/unit/cartography/intel/aws/test_permission_relationships.py -q
29 passed
```

#### General

#### Proof of functionality


cartography/intel/aws/permission_relationships.py
```diff
@@ -59,12 +59,50 @@ def evaluate_action_for_permission(statement: Dict, permission: str) -> bool:
...
+        clause {str, re.Pattern} -- The resource clause to evaluate against.
+        resource_arn {str} -- The resource ARN to match.
+
+    Returns:
+        [bool] -- True if the clause matches the resource ARN, False otherwise
+    """
+    if evaluate_clause(clause, resource_arn):
+        return True
+    if resource_arn.startswith(_S3_BUCKET_ARN_PREFIX):
+        return evaluate_clause(clause, f"{resource_arn}/")
```
kunaals: this only matches `bucket/*`, but not common prefix grants like `bucket/logs/*`, so the fix is narrower than the comment says and still misses object-level grants under a prefix.

cartography/intel/aws/permission_relationships.py
```diff
@@ -59,12 +59,50 @@ def evaluate_action_for_permission(statement: Dict, permission: str) -> bool:
     return False
 
 
+# Prefix of every S3 bucket ARN. S3 ARNs are region- and account-less, so this
+# uniquely identifies a bucket resource. Object ARNs nest under the bucket as
```
kunaals, line 63: i don’t think we should use the object-to-bucket widening for `notresource`. `notresource: bucket/*` excludes objects, not the bucket arn itself, so this can drop bucket-level permission edges that aws would still allow.

kunaals (changes requested):

kunaals (approved):

PR #2884, opened 2026-06-08
feat(kubernetes): Process mapAccounts from aws-auth ConfigMap

### LINKED ISSUE #1857: Process mapAccounts from aws-auth-map in Kubernetes Module
**Feature request template**

*Title*: *Process mapAccounts from aws-auth-map in kubernetes module*

*Description*:
>Right now we are just processing mapUsers and mapAccounts from the aws-auth map in intel/kubernetes/eks.py. If we add functionality to process mapAccounts in the parse_aws_auth_map() function and modify we can make sure that we are getting users from accounts that are listed in mapAccounts. 

[optional *Relevant Links*:]
>Reference this PR to see current Kube Functionality( https://github.com/cartography-cncf/cartography/pull/1825)

jychp:
### Type of change
- [x] Documentation update

### Summary
The EKS identity sync only processed `mapRoles` and `mapUsers` from the `aws-auth` ConfigMap. This adds support for the third field, `mapAccounts`, which grants cluster access to every IAM principal in a listed AWS account.

For each account ID in `mapAccounts`, Cartography resolves the IAM principals already synced from that account (`AWSUser`, `AWSRole`, and the account `AWSRootPrincipal`) and maps each to a `KubernetesUser` named after the principal ARN, with no Kubernetes RBAC groups assigned. This matches the aws-iam-authenticator contract, where a matching `IsAccountAllowed(identity.AccountID)` returns the canonical ARN with no groups.

Because the mappings are resolved from AWS data already in the graph, the AWS account must be synced for them to appear. If a listed account has no principals in the graph, an info log is emitted rather than failing.

### Related issues or links
- Fixes #1857

### How was this tested?
- New unit tests for `parse_aws_auth_map` parsing `mapAccounts` (including int-to-string normalization and the missing-field case).
- New integration test (`test_eks_sync_maps_accounts_to_all_iam_principals`) asserting that every `AWSUser`/`AWSRole`/`AWSRootPrincipal` in a mapped account gets a `MAPS_TO` relationship to a `KubernetesUser` keyed by ARN, and that no Kubernetes groups are created.
- `make test_lint` passes locally; full EKS unit + integration suite passes (10 passed).

#### General

#### Proof of functionality

- [x] Updated the schema documentation (added `(:AWSRootPrincipal)-[:MAPS_TO]->(:KubernetesUser)` and `mapAccounts` config notes).

### Notes for reviewers
`mapAccounts` deliberately assigns no Kubernetes groups (matching the authenticator), and resolved `KubernetesUser` nodes are keyed by the principal ARN since that is the username the authenticator derives. Root-principal support was added so the account-wide grant is modeled faithfully rather than partially.


cartography/intel/kubernetes/eks.py
```diff
@@ -38,14 +38,16 @@ def get_aws_auth_configmap(client: K8sClient) -> V1ConfigMap:
     )
 
 
-def parse_aws_auth_map(configmap: V1ConfigMap) -> dict[str, list[dict[str, Any]]]:
+def parse_aws_auth_map(configmap: V1ConfigMap) -> dict[str, list[Any]]:
     """
-    Parse mapRoles and mapUsers from aws-auth ConfigMap.
+    Parse mapRoles, mapUsers and mapAccounts from aws-auth ConfigMap.
 
     :param configmap: V1ConfigMap containing aws-auth data
```
kunaals, line 60: this still corrupts some unquoted leading-zero account ids because pyyaml parses values like `012345670123` as octal before we can pad them. we should parse these as strings instead of formatting parsed ints.

kunaals (changes requested):

kunaals (approved):

PR #2886, opened 2026-06-08
feat(github): project GitHub dependencies into the Package ontology with provenance and lockfile fallback

### LINKED ISSUE #2406: Feature request: robust GitHub dependency normalization for ontology Package (lockfile-aware)
## Problem

To simplify merging canonical `Package` ontology support, we removed GitHub->Package propagation from the current PR because GitHub dependency graph data is currently inconsistent for canonical package matching.

Observed in real syncs:
- Many GitHub dependencies have `requirements` ranges (for example `>=`) and `packageUrl` without `@version` (for example `pkg:pypi/langchain`), so `version` and `normalized_id` are null.
- `subimagesec/subimage` dependencyGraphManifests queries timed out in GraphQL during testing.
- `uv.lock` is present in repos but is not reliably represented in dependency graph output used by current sync path.

This leads to weak ontology linkage (`Package` dedup/matching misses) and false negatives when attempting `Dependency -> Package` mapping.

## Request

Add a dedicated GitHub dependency normalization path designed for ontology-grade matching:

1. Keep preferring `packageUrl` when present.
2. Only create `normalized_id` when version is exact/reliable.
3. Add lockfile-aware fallback parsing for ecosystems where dependency graph is range-only or incomplete (at least `uv.lock` and npm lockfiles).
4. Add clear provenance fields to dependency nodes (for example `source = dependency_graph|lockfile`, `version_confidence = exact|range|unknown`).
5. Add metrics/logging for coverage and quality:
   - % dependencies with exact version
   - % dependencies with normalized_id
   - per-repo manifest fetch timeout/error counters

## Why this matters

Canonical ontology `Package` should avoid low-confidence joins. We should only project GitHub dependencies into ontology when identifier quality is high enough to avoid noisy/incorrect cross-tool matches.

## Acceptance criteria

- Integration tests for mixed manifests (exact pins + ranges).
- Tests for `uv.lock` fallback (when GraphQL data lacks exact version).
- Documented behavior for when `normalized_id` is intentionally not generated.

If helpful, we can follow up with a scoped PR that adds fallback parsers + quality gates and re-enables GitHub->Package ontology mapping behind those checks.

jychp:
### Type of change
- [x] Documentation update

### Summary

GitHub's dependency-graph ingestion already produces normalized dependency nodes (`version`, `type`, `purl`, indexed `normalized_id`), but those nodes were never projected into the canonical `Package` ontology, so GitHub dependencies could not be queried alongside Trivy, Syft, SocketDev, and GitLab packages. This PR closes that gap and hardens the version data behind it, in four self-contained commits (one per phase):

1. **Project GitHub dependencies into the Package ontology.** Add an additive `GitHubDependency` extra label to the dependency node, a `PackageToGitHubDependencyRel` matching on `normalized_id`, and a `github` entry in `PACKAGES_ONTOLOGY_MAPPING`. This mirrors how GitLab/SocketDev expose a provider-specific label. The primary `Dependency` label and all existing `REQUIRES`/`HAS_DEP` edges are untouched, and the new label avoids sweeping in the unrelated nodes that also share the generic `Dependency` label. The `(:Package)-[:DETECTED_AS]->(:GitHubDependency)` edge is created declaratively during the packages ontology sync.

2. **Provenance fields.** Add `source` (`dependency_graph` or `lockfile`) and `version_confidence` (`exact` / `range` / `unknown`) to the dependency node so confidence in each version is observable in the graph.

3. **Coverage metrics.** The per-repo dependency transform now logs a structured summary (total, percent exact, percent normalized_id, count with a purl), and per-repo manifest fetch failures are aggregated into a single WARNING.

4. **Lockfile-aware fallback.** For ecosystems with a parseable lockfile (`uv.lock` for pip, `package-lock.json` for npm), fetch the lockfile co-located with each dependency's manifest and recover an exact version for range-only dependencies, tagging them `source=lockfile`, `version_confidence=exact`. The parsers are pure (no network) and only return versions that are unambiguous for a direct dependency.

**Shared-node safety:** a `Dependency` node is keyed by `name|requirements` (not by version), so the same `name|requirements` referenced from several manifests merges into one node. If that id resolves to more than one exact version (from the dependency graph or a lockfile), the exact version is cleared so the shared node carries a null `normalized_id` and is simply not projected, rather than projected with a wrong version.

### Related issues or links

- Fixes #2406

### How was this tested?

- `uv run --frozen pytest tests/unit/cartography/intel/github/test_repos.py tests/unit/cartography/intel/github/test_lockfiles.py tests/integration/cartography/intel/github/test_repos.py tests/integration/cartography/intel/ontology/test_packages.py`
- `make test_lint`

New coverage includes: PURL extraction and the documented null-degradation for range-only deps; the `source`/`version_confidence` classification; the coverage-summary log; the lockfile parsers (uv.lock, npm v1 and v2/v3, ambiguity handling); co-located lockfile fetching in a monorepo; and the shared-id version-conflict guards, including a Neo4j-level test that loads two conflicting same-id rows and asserts the merged node keeps a null `normalized_id`. The integration test `test_load_ontology_packages` runs the real ontology packages sync and asserts the `(:Package)-[:DETECTED_AS]->(:GitHubDependency)` edge resolves.

#### General

#### Proof of functionality

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (`docs/root/modules/github/schema.md`).

### Notes for reviewers

- The lockfile fallback was built and validated against fixtures and mocked fetches, not a live GitHub sync, so it is not accompanied by real-environment sync logs. The Neo4j-backed integration tests exercise the load path end to end.
- The lockfile fetch is intentionally co-located with each manifest (a manifest under `services/api/package.json` only uses `services/api/package-lock.json`, never the repo-root lockfile) to avoid attaching one project's versions to another in a monorepo.


cartography/intel/github/lockfiles.py
```diff
@@ -0,0 +1,120 @@
...
+  transitive installs (``node_modules/a/node_modules/b``) are ignored, so a
+  transitive version can never be attached to a direct dependency of the same
+  name.
+- uv: a package name that resolves to more than one distinct version is dropped
+  rather than guessing which version a range refers to.
+"""
+
+import json
+import logging
+import tomllib
```
kunaals: `tomllib` is only stdlib on python 3.11+, but the package still declares `requires-python >=3.10` and says 3.10 support is not removed until oct 2026. this will make github imports fail on 3.10 unless we add a `tomli` fallback or bump the python floor here.

kunaals (changes requested):

kunaals (approved):

PR #2887, opened 2026-06-08
feat(rules): Separate rule identity from compliance framework mapping

kunaals:
### Type of change

- [x] Documentation update

### Summary

Separates rule identity and display names from compliance framework mappings.

This renames compliance-prefixed rule IDs and names across AWS, GCP, Google Workspace, Kubernetes, and NIST AI RMF aligned rules so the rule IDs describe the provider/security concept and rule names are plain security names. Compliance framework/control metadata remains on `frameworks`, which stays the source of truth for compliance-specific rendering and filtering.

This also adds `Framework.control_title` as optional user-facing metadata for the mapped external framework requirement/control. `Framework.requirement` remains the requirement/control id, such as `5.1.8`, `8.2`, or `govern 5`. Control titles are preserved as display copy, are serialized through rule results, and are not used for framework matching.

The framework mapping model is intentionally many-rules-to-one-control: `Rule.name` describes the Cartography security detection, while `Framework.control_title` describes the external GRC/control title. UIs can render compliance labels from `<framework-id> (<requirement>) <control_title>` without treating the control title as the rule display name.

This also adds registry-wide tests that reject compliance-prefixed rule IDs/names and prove renamed rules still keep framework mappings and framework filterability.

### Related issues or links

N/A

### Breaking changes

Rule IDs for compliance-mapped rules have changed. No migration aliases are included because these rules are not production-facing yet.

`Framework.title` has been renamed to `Framework.control_title` to avoid implying that the mapped control title is the rule display name.

### How was this tested?

Ran CLI locally, is working

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

This intentionally does not add backwards-compatible rule ID aliases.


.agents/skills/create-rule/SKILL.md
```diff
@@ -170,30 +170,26 @@ cartography-rules run my_security_rule --no-experimental
...
+    id="aws_access_keys_not_rotated",
+    name="Access Keys Not Rotated",
     # ...
     tags=("iam", "credentials", "stride:spoofing"),  # category tags only
     frameworks=(
-        Framework(
-            name="CIS AWS Foundations Benchmark",
-            short_name="CIS",
-            scope="aws",
-            revision="5.0",
```
kunaals, line 188: note that we have one active revision per benchmark. we could expand to multiple in the future

cartography/rules/data/frameworks/iso27001.py
```diff
@@ -12,11 +12,40 @@
 ISO27001_SHORT_NAME = "27001"
```
jychp: Should be renamed "ISO 27001"

tests/unit/rules/test_rule_identity.py
```diff
@@ -0,0 +1,187 @@
...
+
+def test_rule_framework_mappings_have_control_titles():
+    for rule in RULES.values():
+        for framework in rule.frameworks:
+            assert framework.control_title is not None, (rule.id, framework)
+
+
+def test_rule_definitions_do_not_inline_framework_control_titles():
+    for path in RULE_DATA_DIR.glob("*.py"):
+        assert not INLINE_CONTROL_TITLE_ARG.search(path.read_text()), str(path)
```
jychp: This test bans `control_title=` in every rule file, but the updated docs (cis-conventions.md:96, create-rule SKILL.md, compliance-frameworks.md) explicitly instruct setting `Framework(control_title=...)` inline for frameworks without a helper or known lookup title. A contributor following the docs will trip this test. Reconcile: either allow inline titles for helperless frameworks, or drop that guidance and require all titles in the central lookup dicts.

cartography/rules/data/frameworks/cis.py
```diff
@@ -20,42 +20,156 @@
...
+    "4.1.1.1": "Ensure 2-Step Verification (Multi-Factor Authentication) is enforced for all users in administrative roles",
+    "4.1.1.3": "Ensure 2-Step Verification (Multi-Factor Authentication) is enforced for all users",
+}
+
+
+def _control_title(requirement: str, titles: dict[str, str]) -> str | None:
+    return titles.get(requirement.strip().lower())
+
+
+def cis_aws(requirement: str, control_title: str | None = None) -> Framework:
```
jychp, line 129: The new `control_title` override param on every helper is unreachable from rule files, since `test_rule_definitions_do_not_inline_framework_control_titles` forbids `control_title=` there. The override is dead code in-repo. Drop the parameter or relax the test.

tests/unit/rules/test_rule_identity.py
```diff
@@ -0,0 +1,187 @@
...
+
+RULE_DATA_DIR = Path(__file__).parents[3] / "cartography" / "rules" / "data" / "rules"
+
+COMPLIANCE_NAME_PREFIX = re.compile(
+    r"^(CIS AWS|CIS GCP|CIS Google Workspace|CIS Kubernetes|CIS K8s|NIST AI RMF)\b"
+)
+INLINE_CONTROL_TITLE_ARG = re.compile(r"\bcontrol_title\s*=")
+
+
+def test_rule_ids_do_not_use_compliance_prefixes():
```
jychp, line 28: not sure is test is very usefull

jychp (changes requested):

.agents/skills/create-rule/SKILL.md
kunaals: added docs noting helpers encode one active revision per benchmark scope today.

tests/unit/rules/test_rule_identity.py
kunaals: kept the test but narrowed it to enforce helper defaults only; that covers the convention without blocking helperless mappings.

cartography/rules/data/frameworks/iso27001.py
kunaals: good catch. renamed the iso framework display name to iso 27001 while keeping revision 2022 and annex a control ids.

tests/unit/rules/test_rule_identity.py
kunaals: good catch. narrowed the test to helper calls and left direct Framework(control_title=...) valid for helperless mappings.

cartography/rules/data/frameworks/cis.py
kunaals: fixed by relaxing the test; helper overrides now require an explicit allowlist instead of being impossible.

jychp (commented):
Review in progress

tests/unit/rules/test_rule_identity.py
```diff
@@ -0,0 +1,199 @@
...
+    r")\([^)]*\bcontrol_title\s*=",
+    re.DOTALL,
+)
+ALLOWED_HELPER_CONTROL_TITLE_OVERRIDES: dict[str, tuple[str, ...]] = {}
+
+
+def test_rule_ids_do_not_use_compliance_prefixes():
+    for rule in RULES.values():
+        assert not rule.id.startswith(("cis_", "nist_ai_")), rule.id
+
```
jychp, line 31: [blocking] `ALLOWED_HELPER_CONTROL_TITLE_OVERRIDES` is empty `{}` but the authoring docs (`cis-conventions.md`, `create-rule/SKILL.md`) explicitly tell contributors to pass `control_title=` to a helper when the central lookup is incorrect. Any rule file that follows that guidance will trip `test_rule_definitions_use_framework_helper_default_control_titles` with no obvious fix path. Either (a) document that you must add the file to `ALLOWED_HELPER_CONTROL_TITLE_OVERRIDES` first, or (b) remove the guidance to pass `control_title=` to helpers inline.

cartography/rules/data/frameworks/cis.py
jychp, line 129: [suggestion] `_control_title` does `requirement.strip().lower()` before dict lookup. This is correct, but none of the CIS dict keys contain whitespace or uppercase letters, so the strip+lower is defensive only. Consider a short comment explaining this normalisation exists to match how Framework stores requirements after `__post_init__`, so future dict entries with mixed casing are handled correctly.

tests/unit/rules/test_rule_identity.py
jychp, line 31: [nitpick] `ALLOWED_HELPER_CONTROL_TITLE_OVERRIDES` is typed as `dict[str, tuple[str, ...]]` where the value is a tuple of helper-function names. A `set[str]` would be more semantically accurate (order is irrelevant, membership is what matters) and avoids having to call `set(allowed_helpers)` in the assertion.

cartography/rules/data/frameworks/iso27001.py
```diff
@@ -1,22 +1,51 @@
...
+run compliance reporting against ISO 27001. The requirement identifiers below
+are Annex A control identifiers from ISO/IEC 27001:2022, which are derived from
+and aligned with ISO/IEC 27002:2022 controls.
 """
 
 from cartography.rules.spec.model import Framework
 
-ISO27001_FRAMEWORK_NAME = "ISO/IEC 27001:2022 Annex A"
+ISO27001_FRAMEWORK_NAME = "ISO 27001"
 ISO27001_SHORT_NAME = "27001"
```
jychp: I think we should use the old name for the long name and `ISO27001` for the short one (consistent with CIS frameworks)

jychp (approved):
Approved with some nits

cartography/rules/data/frameworks/cis.py
kunaals: added a short comment explaining the normalization.

tests/unit/rules/test_rule_identity.py
kunaals: fixed this by documenting the allowlist path next to the helper override guidance.

cartography/rules/data/frameworks/iso27001.py
kunaals: done. restored the long name and changed the short name to iso27001.

tests/unit/rules/test_rule_identity.py
kunaals: good call. changed the allowlist values to set[str].

PR #2891, opened 2026-06-10
feat(iam): condition-aware permission relationships for AWS and GCP (#2250, #1643, #2312)

### LINKED ISSUE #1643: AWS permission relationships: correctly support AWS roles being able to remote into instances via SSM

### LINKED ISSUE #2250: Model AWS IAM Policy Conditions in the Graph
## Description

Cartography's AWS IAM module currently parses policy statements and creates relationships between principals, policies, and resources, but it does not model the `Condition` block of IAM policy statements. This means all permission relationships are treated as unconditional, even when policies include restrictions like:

- `aws:SourceIp` - access only from specific IP ranges
- `aws:MultiFactorAuthPresent` - access requires MFA
- `aws:PrincipalOrgID` - access limited to principals in a specific AWS Organization
- `aws:PrincipalTag` / `aws:ResourceTag` - tag-based access control
- `aws:CurrentTime` / `aws:TokenIssueTime` - time-based restrictions

**One possible idea:**

This work extends Cartography's existing [[permission relationships module](https://github.com/cartography-cncf/cartography/blob/master/cartography/intel/aws/permission_relationships.py)](https://github.com/cartography-cncf/cartography/blob/master/cartography/intel/aws/permission_relationships.py), which already evaluates IAM policy statements to create edges like `CAN_READ` and `CAN_WRITE` between principals and resources.

1. Extend `iam.py` to parse and store condition blocks on `AWSPolicyStatement` nodes
2. Extend `permission_relationships.py` to propagate relevant conditions onto the permission relationship edges it creates (e.g., `(:AWSRole)-[:CAN_READ {requires_mfa: true, source_ip_restricted: true}]->(:S3Bucket)`)
3. For conditions that can be statically evaluated (e.g., `aws:PrincipalOrgID` against known org structure), optionally filter out relationships that would never match
4. Provide example Cypher queries showing how to filter permissions by condition

**Example of the problem:**

This S3 bucket policy only allows access from a corporate VPN:

```json
{
  "Effect": "Allow",
  "Principal": "*",
  "Action": "s3:GetObject",
  "Resource": "arn:aws:s3:::sensitive-bucket/*",
  "Condition": {
    "IpAddress": {
      "aws:SourceIp": "203.0.113.0/24"
    }
  }
}
```

Today, Cartography would represent this as "everyone can read this bucket," which is misleading for security analysis.

## Motivation

Accurate permission analysis is one of Cartography's core use cases. Without condition modeling:

- **Attack path analysis overstates risk** - paths that require MFA or originate from specific IPs appear fully exploitable when they may not be
- **Least privilege audits produce false positives** - policies with tight conditions look overly permissive
- **Security teams lose trust in the data** - if basic IAM nuances aren't captured, users question what else is missing

AWS IAM conditions are widely used in production environments, especially for:
- Service control policies (SCPs) with org-wide restrictions
- Cross-account access with `aws:PrincipalOrgID`
- Sensitive operations gated by MFA
- Network-restricted access to S3/KMS

This work would make Cartography significantly more useful for real-world IAM analysis.

## Alternatives to Discuss

We can discuss these details

**1. Leave conditions unmodeled (status quo)**
- Simpler, but increasingly limits Cartography's usefulness as organizations adopt condition-based policies
- Users can't answer "show me policies that require MFA" or "which resources are IP-restricted"

**2. Model conditions as a JSON blob property**
- Lower implementation effort—just store the raw condition JSON on the policy node
- Downside: no queryability. Users would need to parse JSON in Cypher or post-process results
- Could be a reasonable interim step before full schema modeling

**3. Only model a subset of condition keys**
- Focus on the most common/impactful keys: `aws:SourceIp`, `aws:MultiFactorAuthPresent`, `aws:PrincipalOrgID`, `aws:PrincipalTag`
- Provides 80% of the value with bounded scope
- **This is probably the right approach for an initial implementation**

## Relevant Links

- [[Cartography permission_relationships.py](https://github.com/cartography-cncf/cartography/blob/master/cartography/intel/aws/permission_relationships.py)](https://github.com/cartography-cncf/cartography/blob/master/cartography/intel/aws/permission_relationships.py) - existing module that evaluates IAM statements and creates permission edges
- [[Cartography Permissions Mapping Documentation](https://lyft.github.io/cartography/modules/aws/permissions-mapping.html)](https://lyft.github.io/cartography/modules/aws/permissions-mapping.html) - how resource permission relationships work today
- [[AWS IAM Condition Element Documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition.html)](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition.html)
- [[AWS Global Condition Context Keys](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html)](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html)
- [[Cartography AWS IAM Schema](https://cartography-cncf.github.io/cartography/modules/aws/schema/iam.html)](https://cartography-cncf.github.io/cartography/modules/aws/schema/iam.html)

### LINKED ISSUE #2312: feat(gcp): GCP IAM Policy follow-ups - hierarchy evaluation and deny policies
## Summary

Follow-up improvements to GCP IAM policy handling after #1469 (org-level IAM roles). The current implementation is **correct for its supported subset** (user/serviceAccount/group with email-like identifiers, allow-only, non-conditional bindings) but **incomplete for real-world GCP IAM semantics**.

### 1. Policy Inheritance Evaluation

**Partially implemented.** `policy_bindings.py` fetches effective/inherited policies via `batch_get_effective_iam_policies`, and bindings are tagged with `resource_type` (organization/folder/project/resource).

However, `permission_relationships.py` only queries project-scoped bindings:
```cypher
MATCH (project:GCPProject{id: $ProjectId})-[:RESOURCE]->(binding:GCPPolicyBinding)
```

Inherited bindings are **stored** but not fully **evaluated** in permission relationship calculations.

### 2. Deny Policies

**Not implemented.** IAM evaluates deny before allow, so ignoring deny policies can **overstate access**. Requires separate API: `iam.googleapis.com/v2/policies/{attachment_point}/denypolicies`

From `permission_relationships.py:203-206`:
```python
# For now, we only handle allow policies
# GCP can have denied permissions and denied principals (under deny policies),
# but we'd need to fetch that separately using the IAM API.
```

### 3. Conditional Bindings

**Stored but filtered out.** `permission_relationships.py` explicitly filters `has_condition = false`, so any conditional grant is ignored and **access is undercounted**.

### 4. Dropped Principal Types

`transform_bindings()` only keeps `user:`, `serviceAccount:`, `group:` members. These valid IAM principals are dropped:

| Principal Format | Example | Use Case |
|-----------------|---------|----------|
| `allUsers` | `allUsers` | Public access |
| `allAuthenticatedUsers` | `allAuthenticatedUsers` | Any Google account |
| `domain:{domain}` | `domain:example.com` | Org-wide access |
| `deleted:*` | `deleted:user:foo@example.com?uid=123` | Deleted principals |

### 5. Non-Email Principal Formats

These valid IAM principals don't match the email-based model:

| Principal Format | Example | Use Case |
|-----------------|---------|----------|
| K8s SA | `serviceAccount:{project}.svc.id.goog[{namespace}/{ksa}]` | GKE Workload Identity |
| `principal://` | `principal://iam.googleapis.com/...` | Workforce/Workload Identity Federation |
| `principalSet://` | `principalSet://iam.googleapis.com/...` | Principal sets, "all SAs in project/folder/org" |

### 6. Group Expansion

Permission relationships connect group principals to resources, but there's **no propagation from group → member users**. User-level access derived from group membership requires additional analysis.

## Tasks

- [ ] Evaluate inherited policy bindings in permission relationships
- [ ] Add deny policy fetching and evaluation
- [ ] Handle conditional bindings (or flag them appropriately)
- [ ] Support `allUsers`, `allAuthenticatedUsers`, `domain:*` principals
- [ ] Support K8s service account identifiers (GKE Workload Identity)
- [ ] Support `principal://` and `principalSet://` formats
- [ ] Add group → member expansion for user-level access analysis

## Related

- #1734 - Implement Resource Permission Relationships for GCP
- #415 - Extend GCP support

jychp:
### Type of change
- [x] Documentation update

### Summary

Makes IAM permission-relationship edges semantics-aware across AWS and GCP. Today the engines treat every grant as unconditional, ignoring policy conditions, target state, and conditional bindings, which misrepresents real access. This PR addresses three related issues:

**#2250 - Model AWS IAM policy conditions.** IAM `Condition` blocks were stored on `AWSPolicyStatement` but never evaluated, so a grant restricted to a VPN/MFA/org looked unconditional. AWS evaluates conditions at request time and they cannot be resolved statically, so each permission edge is now annotated (keep + flag) instead of dropped:
- `has_condition` is `true` only when every matching Allow statement granting the edge is gated by a `Condition` (an unconditional Allow path wins).
- `condition_keys` lists the referenced context keys (e.g. `aws:SourceIp`).
- `conditions` carries the raw operator maps as a JSON string.

**#1643 - SSM remote access.** The issue's premise (only the "SSM-managed" check was missing) was incorrect: `ssm:StartSession` edges could never form because `EC2Instance` had no ARN and the engine matches resources only by ARN. This PR:
- Synthesizes and indexes `EC2Instance.arn` (`arn:aws:ec2:{region}:{account}:instance/{id}`), resolving a long-standing TODO.
- Adds a generic, declarative `target_precondition` to the permission-relationships engine: a relationship may require the target resource to be connected to a given label via a given relationship/direction (inert for existing entries).
- Adds a `CAN_START_SESSION` mapping gated on `(:EC2Instance)-[:HAS_INFORMATION]->(:SSMInstanceInformation)`.

**#2312 - GCP IAM follow-ups (partial).** Two of the issue's items were already implemented and verified during this work (org/folder inheritance is already evaluated via `resolve_gcp_scope`; `allUsers`/`allAuthenticatedUsers`, Workload Identity Federation, and GKE Workload Identity service accounts are already handled; group->member expansion is already traversable because `GoogleWorkspaceGroup` carries the `GCPPrincipal` label). The genuine gaps fixed here:
- **Conditional bindings** were dropped (understating access). They are now retained and the edge is flagged with `has_condition` / `condition_title` / `condition_expression`. Conditional grants are routed off the bulk Cartesian fast-paths (which can only set kwargs-bound properties) onto a new condition-capable `GCPConditionalPermissionMatchLink` used on the row-by-row path.
- **`domain:` members** were dropped; they are now retained on `GCPPolicyBinding.domains` for visibility.
- Documented the supported principal types and the group->member traversal query.

Deferred to a separate PR (not in scope here): GCP **deny policies**, which require a net-new GCP IAM API integration and deny-before-allow evaluation. For that reason this PR does not auto-close #2312.

### Related issues or links

- Fixes #2250
- Fixes #1643
- Addresses #2312 (conditional bindings + `domain:` retention; deny-policy support tracked separately)

### How was this tested?

- New and updated unit tests for AWS condition collection, the target-precondition clause builder, the EC2 ARN transform, GCP conditional-binding inclusion/aggregation, and GCP `domain:` retention.
- New integration tests (run against local Neo4j): AWS `CAN_START_SESSION` only attaches to SSM-managed instances; GCP conditional edges are flagged with their condition metadata while unconditional edges are `has_condition=false`.
- `make test_lint` passes locally (black, isort, flake8, mypy).

#### General

#### Proof of functionality

- [x] Updated the schema documentation (AWS EC2Instance/permissions-mapping, GCP permission-mapping/GCPPolicyBinding).

### Notes for reviewers

- The biggest behavioral change to review is the GCP split: conditional bindings must bypass `_split_project_scope_principals` / the BigQuery dataset-scope fast path so their per-edge condition metadata survives. Unconditional broad grants are unchanged.
- `EC2Instance.arn` is newly populated; it is additive (existing `id`-based matching is untouched).


cartography/intel/gcp/permission_relationships.py
```diff
@@ -209,7 +212,13 @@ def _split_project_scope_principals(
             if not _assignment_allows_permissions(assignment_data, permissions):
                 continue
 
-            if assignment_data["scope"].pattern == project_scope_pattern:
+            # Conditional grants must keep per-edge condition metadata, which the bulk
```
kunaals, line 215: this transition can leave stale condition metadata on an edge that becomes unconditional. if a previous sync wrote the edge through the conditional row-by-row schema with `has_condition=true`, and a later sync routes the same now-unconditional project-scope grant through the cartesian bulk path, the `MERGE` only refreshes `lastupdated` / sub-resource properties and does not clear `has_condition`, `condition_title`, or `condition_expression`. cleanup won’t remove it because the edge is current, so queries looking for unconditional access can keep missing it. can we clear the condition fields on the bulk unconditional path, or add a regression test for conditional -> unconditional on the same edge?

kunaals (changes requested):

cartography/intel/gcp/permission_relationships.py
```diff
@@ -350,6 +419,13 @@ def _split_bigquery_table_dataset_scope_principals(
             if not _assignment_allows_permissions(assignment_data, permissions):
                 continue
 
+            # As with project scope, conditional grants cannot ride the bulk loader.
+            if assignment_data.get("has_condition"):
```
kunaals, line 423: this still lets a conditional table binding overwrite an unconditional dataset-scope grant for the same principal/table. the dataset fast path writes the edge with `has_condition=false`, but the conditional binding stays in `residual_principals` and the later row-by-row load can merge the same edge back to `has_condition=true`. since unconditional access should win, can we drop residual bindings already covered by a broad unconditional grant, or otherwise preserve `has_condition=false` when the edge was already loaded unconditionally?

kunaals (changes requested):

kunaals (approved):

PR #2901, opened 2026-06-12
feat(azure): sync management group role assignments

### LINKED ISSUE #2831: Azure: model RBAC role assignments at their owning scope

kunaals:
### Type of change
- [x] Documentation update

### Summary
Adds Azure management-group-scoped RBAC role assignment sync so `AzureRoleAssignment` nodes are owned by the Azure scope where they are attached.

Management-group assignments now load under `(:AzureManagementGroup)-[:RESOURCE]->(:AzureRoleAssignment)`, while subscription assignments continue loading under `(:AzureSubscription)-[:RESOURCE]->(:AzureRoleAssignment)`. Management-group assignment cleanup is scoped per management group and skipped for that scope if Azure returns an access/API error, preserving prior graph data for that management group.

This intentionally does not change Azure role definition ownership or effective inherited permission modeling.

### Related issues or links
- Fixes #2831

### Breaking changes
None.

### How was this tested?

- Manual live validation against an Azure tenant using a fresh temporary Neo4j database. Validated management-group role assignment sync and scoped cleanup by seeding a stale management-group role assignment and confirming the next scoped cleanup removed it.

#### General

#### Proof of functionality

- [x] Included Cartography sync logs from a real environment demonstrating successful synchronization of the new/modified entity. Logs are available from manual validation and can be shared privately if needed.

- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md). Not applicable: this checkout does not include `docs/schema/README.md`.

### Notes for reviewers
This is scoped to direct management-group-owned Azure role assignments. Subscription-scoped RBAC behavior remains unchanged, and inherited/effective permission expansion remains separate follow-up work.


cartography/intel/azure/rbac.py
```diff
@@ -47,6 +57,28 @@ def get_role_assignments(
...
+    credentials: Credentials,
+    authorization_subscription_id: str,
+    scope: str,
+    management_group_id: str,
+) -> list[dict]:
+    """
+    Fetch role assignments attached directly to a given Azure scope.
+    """
+    client = get_client(credentials, authorization_subscription_id)
+    role_assignments = list(client.role_assignments.list_for_scope(scope))
```
jychp: Blocking: `list_for_scope(scope)` with no `$filter` returns assignments inherited from parent management groups AND every assignment at child scopes (subscriptions, resource groups, resources) under this management group, per the Azure REST docs ("List all role assignments that apply to a scope"; `atScope()` would still include parent-inherited ones). `get_role_assignments_for_scope` then stamps every returned assignment with this `management_group_id`, so each MG ends up directly owning ancestor and descendant assignments. Because `AzureManagementGroupRoleAssignmentSchema` shares the `AzureRoleAssignment` label and merges on `id`, subscription-scoped assignments already loaded by the subscription sync get a second `(:AzureManagementGroup)-[:RESOURCE]->` edge and have their `subscription_id`/`management_group_id` properties overwritten. Filter to direct assignments only: keep an assignment when `assignment.get("scope") == management_group_id`.

cartography/intel/azure/rbac.py
```diff
@@ -265,6 +333,72 @@ def cleanup_permissions(
...
+        "Syncing Azure RBAC role assignments for management group '%s'.",
+        management_group_id,
+    )
+    role_assignments = get_role_assignments_for_scope(
+        credentials,
+        authorization_subscription_id,
+        management_group_id,
+        management_group_id,
+    )
+    transformed_assignments = transform_role_assignments(role_assignments)
```
jychp, line 432: Blocking: this path loads assignment nodes but never fetches/loads their role definitions. For an assignment whose `role_definition_id` is a management-group/tenant custom role (or any role not already loaded by the first subscription's sync), the `AzureRoleDefinition` node will not exist, so the documented `(:AzureRoleAssignment)-[:ROLE_ASSIGNED]->(:AzureRoleDefinition)` edge and its Permission nodes are missing. Call `extract_role_definition_ids` + `get_role_definitions_by_ids` + `load_role_definitions`/`load_permissions` here as the subscription sync does. The test masks this by pre-loading a subscription role definition with the same ID.

cartography/intel/azure/rbac.py
```diff
@@ -27,6 +30,13 @@ def get_client(
     return client
 
 
+def _get_value(data: dict[str, Any], *keys: str) -> Any | None:
```
jychp: `_get_value` duplicates the identical helper already defined in `cartography/intel/azure/management_groups.py`, and is only ever called with a single key (`"id"`). Drop it and use `management_group.get("id")` directly.

cartography/intel/azure/__init__.py
```diff
@@ -445,6 +446,14 @@ def start_azure_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
             common_job_parameters,
         )
         if management_groups_synced:
+            rbac.sync_management_group_role_assignments_for_management_groups(
+                neo4j_session,
+                credentials,
+                management_group_data,
+                subscriptions[0]["subscriptionId"],
```
jychp, line 453: Question: `subscriptions[0]["subscriptionId"]` arbitrarily picks the first subscription to build the AuthorizationManagementClient for every management-group scope query. Confirm that subscription's credentials are guaranteed to have `Microsoft.Authorization/roleAssignments/read` at the management-group scopes; otherwise every MG lookup 403s (caught and skipped), silently yielding no MG assignments.

tests/data/azure/rbac.py
```diff
@@ -214,6 +214,22 @@
     },
 ]
 
+TEST_MANAGEMENT_GROUP_ID = (
+    "/providers/Microsoft.Management/managementGroups/test-management-group"
+)
+
+AZURE_MANAGEMENT_GROUP_ROLE_ASSIGNMENTS = [
```
jychp, line 241: Test gap: the mocked `get_role_assignments_for_scope` returns a single assignment whose `scope` already equals the management group, so the tests never exercise the real API output (inherited parent assignments and child-scope subscription/resource assignments). The inherited-attribution bug passes these tests. Add fixtures with mixed scopes to prove only direct (`scope == management_group_id`) assignments are attached.

jychp (changes requested):

cartography/intel/azure/rbac.py
kunaals: good catch, fixed with `filter="atScope()"` plus an explicit `assignment.scope == management_group_id` guard so only direct mg-scope assignments get the mg `RESOURCE` edge. i also re-ran live validation and saw 609 raw returned assignments vs 198 direct assignments, with 0 non-direct mg ownership edges after the fix.

cartography/intel/azure/rbac.py
kunaals: yep, agreed. fixed in 7be7ba232 by loading role definitions and permissions in the mg assignment path before loading the assignments, matching the subscription path.

cartography/intel/azure/rbac.py
kunaals: yep, removed it and just use `management_group.get("id")` now.

cartography/intel/azure/__init__.py
kunaals: i checked this with live validation. the sdk builds the request from the explicit management-group scope, and using the first accessible subscription for the auth client worked for all readable mg scopes in that tenant. if a scope 403s we still preserve prior data for that mg rather than cleaning it.

tests/data/azure/rbac.py
kunaals: agreed. added a mixed-scope fixture covering parent mg, subscription, and resource-group assignments, plus a test proving only `scope == management_group_id` is returned/attached.

cartography/intel/azure/rbac.py
```diff
@@ -265,6 +328,108 @@ def cleanup_permissions(
...
+        authorization_subscription_id,
+        role_definition_ids,
+    )
+
+    transformed_assignments = transform_role_assignments(role_assignments)
+
+    load_role_definitions_and_permissions(
+        neo4j_session,
+        role_definitions,
+        authorization_subscription_id,
```
jychp: These role definitions and permissions get loaded under `subscriptions[0]`, but that subscription is just the auth handle here, it doesn't actually own them. So an MG- or tenant-scoped role (or one from another subscription) ends up with a `RESOURCE` edge to the wrong subscription.

It also fights with the per-subscription cleanup: `cleanup_role_definitions` runs earlier during the subscription sync, so on the next run it'll delete these defs before this path recreates them, and the `ROLE_ASSIGNED` edges will flap in and out.

Probably worth not stamping them with a subscription scope, or skipping defs the subscription sync already loaded.

jychp (changes requested):

cartography/intel/azure/rbac.py
kunaals: good catch, fixedby loading mg-fetched role definitions/permissions through unscoped schemas, so the first subscription stays only an auth handle and no longer gets `RESOURCE` ownership edges for those nodes. i added test coverage for that and re-ran live validation: 198 mg assignments, 0 non-direct mg ownership edges, 0 missing role-definition edges, and 0 false subscription role-definition/permission edges.

jychp (approved):

PR #2908, opened 2026-06-12
fix(azure): support azure-mgmt-web 11 model shape

jychp:
### Type of change
- [ ] Documentation update

### Summary
This supersedes #2907.

Dependabot PR #2907 only bumps azure-mgmt-web to 11.0.0 in uv.lock. That version changes the SDK model serialization shape for Web Apps and Site Configs. The new hybrid models put App Service and Function App fields under properties, so the existing transforms would silently drop state, default_host_name, https_only, and linuxFxVersion-derived container image data.

This PR intentionally hard-pins the minimum azure-mgmt-web version to 11.0.0 and updates the Azure App Service and Function App transforms to consume the SDK 11 properties shape directly, without retaining compatibility for the pre-11 flat shape.

### Related issues or links
- Supersedes #2907

### Breaking changes
The minimum supported azure-mgmt-web version is now 11.0.0. The Azure Web transforms now expect the SDK 11.0.0 hybrid model as_dict shape with resource-specific fields under properties.

### How was this tested?
- UV_CACHE_DIR=.uv-cache uv run --frozen pytest -q tests/integration/cartography/intel/azure/test_app_service.py tests/integration/cartography/intel/azure/test_functions.py
- Smoke-tested the transforms with azure.mgmt.web.models.Site and SiteConfigResource from azure-mgmt-web 11.0.0 to verify state/defaultHostName/httpsOnly and linuxFxVersion are read from properties.

#### General
- [ ] The linter passes locally (make test_lint).

#### Proof of functionality

### Notes for reviewers
Focus area: the shape change from azure-mgmt-web 11 hybrid models. Existing Azure connectors already read provider-specific fields through properties; this brings App Service and Function Apps in line with that pattern.


cartography/intel/azure/functions.py
```diff
@@ -124,16 +124,16 @@ def transform_function_apps(
...
             # deployment) from "config fetch failed" (unknown). Silently defaulting
             # to "code" on a transient Azure error would misclassify container apps.
             site_config: dict[str, Any] | None = (
                 configurations.get(app_id) if isinstance(app_id, str) else None
             )
             if site_config is not None:
-                linux_fx_version = site_config.get(
-                    "linux_fx_version"
-                ) or site_config.get("linuxFxVersion")
+                config_properties = site_config.get("properties", {}) or {}
```
kunaals, line 135: i don’t think this matches the sdk shape for site config. with `azure-mgmt-web==11`, `SiteConfigResource(linux_fx_version="DOCKER|...").as_dict()` still returns `{"linux_fx_version": ...}` at the top level, not under `properties.linuxFxVersion`. this means container function apps get classified as code deployments and lose their image fields. the fixture was changed to the nested shape too, so the test won’t catch it.

kunaals (changes requested):

kunaals (approved):
lgtm!

PR #2910, opened 2026-06-12
fix(rules): K8s 5.1.13 noise, filter-property indexes, KubernetesNode->EC2Instance link, quieter ECR sync

jychp:
### Type of change
- [ ] Documentation update

### Summary
A batch of independent reliability, performance, and signal-to-noise improvements:

- **CIS Kubernetes 5.1.13 (service-account token creation):** the rule flagged the built-in `cluster-admin` / `admin` / `edit` ClusterRoles and the EKS-managed `eks:*` ClusterRoles. These are platform-owned defaults a tenant cannot remediate, so they were unactionable noise. Exclude them (scoped to ClusterRoles only, so user-defined namespaced Roles with those names still surface). Custom ClusterRoles granting `create` on `serviceaccounts/token` still trigger.

- **Index high-cardinality filter properties:** three model properties used as `WHERE`/inline filters lacked an index, causing full label scans:
  - `GuardDutyFinding.archived`
  - `GCPPolicyBinding.has_condition`
  - `S3Bucket.anonymous_access`

  Added `extra_index=True` to each so the filters use an index.

- **Link `KubernetesNode` to its backing `EC2Instance`:** EKS nodes expose `spec.providerID` as `aws:///<az>/<instance-id>`. We now capture `provider_id` during the node transform, parse the EC2 instance id from it, and create `(:KubernetesNode)-[:IS_INSTANCE]->(:EC2Instance)` (mirroring the existing `ECSContainerInstance` convention). The link is created only when the provider id resolves to an EC2 instance id, so non-AWS providers (GCE, Azure) are unaffected.

- **Quieter ECR image-layer sync:** the layer sync logged a WARNING for every region with no ECR repositories. Most regions legitimately have none, so this fired on nearly every region of every account. Demoted to DEBUG since an empty region is the expected case.

### Related issues or links

N/A

### Breaking changes

None.

### How was this tested?
- `make test_lint` passes locally (black, flake8, isort, mypy).
- Kubernetes integration suite passes (54 tests), including a new assertion that an EKS node links to its backing `EC2Instance` while a non-AWS node (null `provider_id`) produces no link.
- ECR image-layers unit tests pass (46 tests).
- CIS Kubernetes rule unit tests pass (364 tests).

#### General

#### Proof of functionality

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (KubernetesNode: new `provider_id` / `instance_id` fields and the `IS_INSTANCE` relationship).

### Notes for reviewers
These changes are independent and can be reviewed commit by commit. The `KubernetesNode` -> `EC2Instance` link uses the node schema's `other_relationships`, so the edge forms once the matching `EC2Instance` exists; on a fresh graph where the Kubernetes sync runs before the AWS sync, the link lands on the following run.

### DIFF
--- a/cartography/intel/aws/ecr_image_layers.py
+++ b/cartography/intel/aws/ecr_image_layers.py
@@ -1381,9 +1381,11 @@ def sync(
         )
 
         if not repo_images_list:
-            logger.warning(
-                f"No ECR images found in graph for region {region}. "
-                f"Run 'ecr' sync first to populate basic ECR data."
+            # Most regions legitimately have no ECR repositories, so this is
+            # the expected case rather than a problem worth a warning.
+            logger.debug(
+                f"No ECR images found in graph for region {region}; "
+                f"skipping layer sync for this region."
             )
             continue
 
--- a/cartography/intel/kubernetes/nodes.py
+++ b/cartography/intel/kubernetes/nodes.py
@@ -20,14 +20,29 @@ def get_nodes(client: K8sClient) -> list[V1Node]:
     return k8s_paginate(client.core.list_node)
 
 
+def _ec2_instance_id_from_provider_id(provider_id: str | None) -> str | None:
+    """Extract the EC2 instance id from an EKS node's providerID.
+
+    EKS sets ``spec.providerID`` to ``aws:///<availability-zone>/<instance-id>``.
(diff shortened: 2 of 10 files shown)


kunaals (approved):
lgtm

PR #2911, opened 2026-06-13
feat(aws): add WAFv2 web ACL support

### LINKED ISSUE #1067: [FeatureRequest] Adding the ability to map WAF/WebACL with AWS
Adding the ability to map if LB or APIGW have WAF enabled and the information of the WebACL, could be really helpful.

/assign @AlmogBaku (I'm already working on a PR :P )

raajheshkannaa:
### Type of change

### Summary
Adds WAFv2 web ACL support, following the roadmap laid out in #1067:

- New `AWSWebACL` node for REGIONAL and CLOUDFRONT scoped web ACLs, synced via `list_web_acls`
- `(:AWSWebACL)-[:PROTECTS]->(:AWSLoadBalancerV2)` from `list_resources_for_web_acl`
- `(:AWSWebACL)-[:PROTECTS]->(:APIGatewayStage)` and `(:AWSWebACL)-[:PROTECTS]->(:CloudFrontDistribution)` matched on the `webaclarn` and `web_acl_id` properties those nodes already carry, so no extra API calls are needed
- AppSync is not modeled in cartography yet, so no AppSync edge

CLOUDFRONT scoped ACLs are global and only listable via us-east-1, so they sync once outside the region loop. The module is registered after `ec2:load_balancer_v2`, `apigateway`, and `cloudfront` so the PROTECTS edges can match existing nodes.

### Related issues or links
- Closes #1067

### Breaking changes
None.

### How was this tested?
- Unit tests for pagination (including the NextMarker-on-final-page quirk), the deleted-ACL race on `list_resources_for_web_acl`, transform, and sync ordering
- Integration tests for nodes, all relationships, and cleanup of stale nodes and stale PROTECTS edges on re-sync
- `make test_lint` passes locally

#### General

#### Proof of functionality

- [ ] Sync logs from a real environment. I do not have WAFv2 resources in an account right now; happy to set some up and attach logs if needed.

### DIFF
--- a/cartography/intel/aws/resources.py
+++ b/cartography/intel/aws/resources.py
@@ -47,6 +47,7 @@ from . import ses
 from . import sns
 from . import sqs
 from . import ssm
+from . import wafv2
 from .ec2.auto_scaling_groups import sync_ec2_auto_scaling_groups
 from .ec2.elastic_ip_addresses import sync_elastic_ip_addresses
 from .ec2.images import sync_ec2_images
@@ -160,5 +161,8 @@ RESOURCE_FUNCTIONS: OrderedDict[str, Callable[..., None]] = OrderedDict(
         "cognito": cognito.sync,
         "eventbridge": eventbridge.sync,
         "glue": glue.sync,
+        # `wafv2` must run after `ec2:load_balancer_v2`, `apigateway`, and
+        # `cloudfront` so that web ACL PROTECTS edges can match those nodes.
+        "wafv2": wafv2.sync,
     }
 )
--- a/cartography/intel/aws/resources.py
+++ b/cartography/intel/aws/resources.py
@@ -47,6 +47,7 @@ from . import ses
 from . import sns
 from . import sqs
 from . import ssm
+from . import wafv2
 from .ec2.auto_scaling_groups import sync_ec2_auto_scaling_groups
 from .ec2.elastic_ip_addresses import sync_elastic_ip_addresses
 from .ec2.images import sync_ec2_images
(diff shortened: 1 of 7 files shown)


raajheshkannaa:
@jychp rebased onto master. It was 220 commits behind and genuinely broken against current `main`, so I fixed that first:

- **#3020 provider prefixes.** `AWSWebACL` already complied, but the `PROTECTS` edges targeted `APIGatewayStage` and `CloudFrontDistribution`, both now deprecated aliases. Retargeted at `AWSAPIGatewayStage` and `AWSCloudFrontDistribution`, plus the integration test's Cypher. `test_legacy_aws_labels_only_appear_in_migration_code` passes now.
- **#3021 generated schema docs.** Added descriptions to all 8 node properties and removed the `(:A)-[:REL]->(:B)` shorthand from the relationship docstrings, which `test_relationship_docstrings_are_user_facing_descriptions` rejects. Kept the prose explaining when each edge is created.
- **`docs/root/modules/aws/schema.md` is gone**, so the manual schema entry is dropped.

Unit suite is at parity with a clean `master` checkout: same 27 pre-existing failures (Microsoft entra/intune, environment-related), zero introduced by this branch. `pre-commit` passes on all changed files. I can't run the integration tests locally, no Docker here, so those are on CI.

On the `needs-tests` label: I want to make sure I fix the right thing. The PR has unit tests for pagination (including the NextMarker-on-final-page quirk), the deleted-ACL race on `list_resources_for_web_acl`, and transform/sync ordering, plus three integration tests using `check_nodes`/`check_rels` that cover all four relationship types, stale-node cleanup, stale-edge cleanup on re-sync, and the empty case.

What's missing from your point of view? If it's sync logs from a real environment, I don't have WAFv2 resources in an account right now but I'll set some up and attach them. If it's a testing pattern I've diverged from, point me at the module you'd like me to match and I'll follow it.

PR #2913, opened 2026-06-13
fix(microsoft): make Intune detected-apps export resilient to stalled jobs

jychp:
### Type of change
- [ ] Documentation update

### Summary
Microsoft Graph's async device-management report export occasionally strands an
export job in `inProgress` (or marks it `failed`) server-side, even on small
tenants whose report normally completes in a few seconds. When that happens the
job never reaches `completed`, so `wait_for_export_job` raises `TimeoutError`
after the poll budget elapses.

For the Intune detected-apps sync this had an outsized impact:

- The `TimeoutError` is not an `APIError`, so it bypassed the existing
  permission-based guards and aborted the entire Microsoft ingestion, which also
  skipped the compliance-policy sync that runs after detected apps.
- A single stalled job on an optional enrichment report took down unrelated,
  already-successful parts of the sync.

This PR makes the report export resilient:

- `export_report_rows` now resubmits a fresh export job on `TimeoutError` /
  `RuntimeError`, up to `max_attempts` (default 3). A stranded job never
  recovers, but a freshly submitted job almost always completes, so the data
  self-heals within the same run. This covers both `AppInvAggregate` and
  `AppInvRawData`.
- `start_intune_ingestion` now treats an exhausted detected-apps export as a
  non-fatal skip (logged at `warning`), mirroring the existing handling for a
  `403`, so this optional enrichment can never fail the whole Microsoft module.
  Managed devices, compliance policies, and the compliance analysis job continue
  to run.

### Related issues or links

N/A

### How was this tested?
`make test_lint` and the Intune unit tests pass locally:

```
uv run pytest tests/unit/cartography/intel/microsoft/intune/test_reports.py \
              tests/unit/cartography/intel/microsoft/intune/test_init.py
```

New unit tests cover:
- `export_report_rows` resubmitting a fresh job after a stalled job and
  recovering.
- `export_report_rows` re-raising once `max_attempts` is exhausted (no download
  attempted).
- `start_intune_ingestion` continuing through managed devices, compliance
  policies, and the analysis job when the detected-apps export fails with
  `TimeoutError` or `RuntimeError`.

#### General

#### Proof of functionality

### Notes for reviewers
No schema or node/relationship changes; this only changes the resilience of the
report-export control flow and the failure handling around the optional
detected-apps sync.


cartography/intel/microsoft/intune/__init__.py
```diff
@@ -91,6 +91,16 @@ async def main() -> None:
                 )
             else:
                 raise
+        except (TimeoutError, RuntimeError) as e:
```
kunaals: this catch is broader than the export failure case it’s trying to tolerate. `sync_detected_apps()` also validates report shape, writes app nodes/relationships, and runs cleanup, so any unrelated `RuntimeError` from those phases would now be logged as an optional export skip and the microsoft sync would continue after a possible partial write. can we make the report export path raise a specific exception, or catch this inside the export/report boundary instead of around the full detected-app sync?

kunaals (changes requested):

kunaals (approved):
lgtm

PR #2914, opened 2026-06-13
fix(aws): skip Inspector region on transient endpoint failure

jychp:
### Type of change

### Summary
`cartography.intel.aws.inspector` did not handle connection-level failures (connect/read timeouts, endpoint connection errors) when reaching a regional `inspector2` endpoint. These exceptions are **not** `botocore.exceptions.ClientError` subclasses, so neither the generic `@aws_handle_regions` decorator nor the inspector module's existing `ClientError` handler caught them.

As a result, a single unreachable regional endpoint aborted the sync for the **entire account**. This is most easily triggered by an opt-in region (e.g. `me-south-1`) that is listed in `AWS_INSPECTOR_REGIONS` but is not enabled on the target account: every call to its endpoint connect-times-out and the account fails on every run.

This PR mirrors the existing SageMaker pattern (`sagemaker_handle_regions`):

- Catch the connection-error family (`ConnectTimeoutError`, `EndpointConnectionError`, `ReadTimeoutError`, `ConnectionClosedError`) and skip only the affected region, continuing with the other regions and accounts.
- The guard is applied to **both** the `get_member_accounts` probe and the findings sync. For the opt-in-region case the failure actually surfaces first at `list_members` (inside `get_member_accounts`, which is only decorated with `@aws_handle_regions`), before the findings paginator is ever reached, so guarding only the findings path would not have fixed it.
- Track a `cleanup_safe` flag: when any region is transiently skipped, cleanup is skipped (with a warning) so findings already in the graph for the unreachable region are not deleted. This preserves last-known-good data, the same reasoning SageMaker uses for its `cleanup_safe` flag.

### How was this tested?
- `make test_lint` passes locally (black, flake8, isort, pyupgrade, mypy).
- Existing inspector unit and integration tests pass.
- Added a regression test, `test_sync_inspector_survives_connection_timeout`, that seeds a finding from a prior run, makes `get_inspector_findings` raise a `ConnectTimeoutError` during pagination, and asserts the account sync completes without raising and the pre-existing finding is preserved (cleanup skipped).

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/aws/inspector.py
+++ b/cartography/intel/aws/inspector.py
@@ -1,4 +1,5 @@
 import logging
+from functools import wraps
 from typing import Any
 from typing import Dict
 from typing import Iterator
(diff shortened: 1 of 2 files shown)


kunaals (approved):
lgtm

PR #2915, opened 2026-06-15
fix(github): synthesize token_name for classic PATs

jychp:
### Type of change

### Summary
`GitHubPersonalAccessToken` nodes created from SAML SSO credential authorizations (`token_kind = "classic"`) were written with `token_name = None`, because the SAML credential-authorizations API does not return a token name. Since `token_name` is the only human-meaningful label for these nodes, classic PATs ended up with no usable display name and were hard to tell apart in inventory and downstream consumers (in a large org, classic PATs dominate, so this affects most PATs).

This change synthesizes a name from the data classic PATs do carry (`owner_login`):

```python
"token_name": f"{owner_login} (classic PAT)" if owner_login else None,
```

The node stays self-describing, is distinguishable from fine-grained PATs, and remains null only when even `owner_login` is missing. Fine-grained PATs are unaffected (their name comes from the API).

### How was this tested?
Updated the existing unit test asserting the classic-PAT transform output. `make test_lint` and the GitHub PAT unit tests pass locally.

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/github/personal_access_tokens.py
+++ b/cartography/intel/github/personal_access_tokens.py
@@ -207,7 +207,7 @@ def _transform_saml_credential_authorization(
         "id": f"{org_url}/credential-authorizations/{credential_id}",
         "token_kind": "classic",
         "token_id": None,
-        "token_name": None,
+        "token_name": f"{owner_login} (classic PAT)" if owner_login else None,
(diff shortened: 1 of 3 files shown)


kunaals (approved):
lgtm

PR #2916, opened 2026-06-15
fix(rules): exclude account-BPA-covered buckets from S3 block public access

jychp:
### Type of change

### Summary

The CIS AWS 3.1.4 S3 Block Public Access rule (`aws_s3_block_public_access`) flagged any bucket whose bucket-level Block Public Access (BPA) settings were `NULL` or not all `true`, with no awareness of account-level BPA. AWS account-level BPA already blocks public access for every bucket in the account, so a bucket with missing bucket-level settings under an account that fully enforces BPA is not actually publicly exposed: flagging it is a false positive.

This change adds an `EXISTS` guard against the `S3AccountPublicAccessBlock` node so a bucket is not flagged when its parent `AWSAccount` has all four account-level BPA settings (`block_public_acls`, `ignore_public_acls`, `block_public_policy`, `restrict_public_buckets`) enforced.

Account-level BPA is account-global in AWS but stored one node per region with identical values, so a single fully-enforced region node is sufficient to confirm the account blocks public access. The count (denominator) query is unchanged: every bucket is still an evaluated asset.

Rule version bumped `1.0.0` -> `1.1.0`.

### How was this tested?

- New integration test `tests/integration/rules/test_cis_aws_storage.py` covering: bucket with NULL BPA under an account with full account-level BPA (not flagged), under partial account-level BPA (flagged), under no account-level BPA node (flagged), and a bucket with its own full bucket-level BPA (never flagged).
- Unit rules tests pass (1223 passed).

#### General
- [x] The linter passes locally (\`make test_lint\`).

#### Proof of functionality


kunaals (approved):
lgtm

PR #2920, opened 2026-06-16
fix(aws): apigateway client certificate early return

### LINKED ISSUE #2895: bug(aws): get_rest_api_client_certificate returns [] instead of None on missing certificate
## Summary
In `cartography/intel/aws/apigateway.py`, the function 
`get_rest_api_client_certificate` declares a return type of 
`Optional[Any]` but returns an empty list `[]` when no stage 
has a `clientCertificateId`. Callers expecting either a dict 
or `None` will receive a list instead, which can cause silent 
logic errors or downstream `AttributeError`.

## Location
- File: `cartography/intel/aws/apigateway.py`
- Line: ~150

## Current Behavior
When no stage has a `clientCertificateId`, the function returns 
`[]` (an empty list).

## Expected Behavior
The function should return `None` when no client certificate is 
found, consistent with its `Optional[Any]` return type declaration.

## Why This Matters
Type inconsistency between the declared return type and actual 
return value can cause callers to silently mishandle the result
treating a missing certificate as an empty collection rather than 
an absent value. This is a correctness issue that can mask real 
infrastructure state in the graph.

## Proposed Fix
Replace the `return []` fallback with `return None` to match 
the declared `Optional[Any]` return type.

## Additional Notes
Found during a systematic codebase audit as a new contributor.
I would like to work on this fix if maintainers agree with the approach.

g-k-s-03:
### Type of change

### Summary
`get_rest_api_client_certificate` iterated over API stages looking for a
`clientCertificateId`, but contained an `else: return []` inside the loop.
This caused the function to exit on the first stage that lacked a certificate,
silently dropping certificates from any subsequent stage. For a REST API with
stages ordered as `[dev (no cert), prod (cert)]`, the certificate for `prod`
was never retrieved and never written to the graph.

The fix removes the `else: return []` block entirely. The pre-existing
`response = None` initialisation already handles the no-certificate case
correctly and satisfies the `Optional[Any]` return type.

Two unit tests are added to `tests/unit/cartography/intel/aws/test_apigateway.py`
covering the previously broken mixed-stage ordering and the all-no-cert case.

### Related issues or links
- Fixes #2895 
Identified during a codebase audit; root cause clarified by @achantavy in the
linked issue.


cartography/intel/aws/organizations.py
```diff
@@ -145,7 +145,7 @@ def get_aws_organization_accounts_for_parent(
 
 
 def _get_account_state(account: dict[str, Any]) -> str | None:
-    return account.get("State") or account.get("Status")
+    return account.get("State")
```
jychp: This change is unrelated to the actual PR.

cartography/intel/aws/organizations.py
```diff
@@ -186,8 +186,6 @@ def transform_aws_organization_accounts(
                 "email": account.get("Email"),
                 "name": account.get("Name"),
                 "state": _get_account_state(account),
```
jychp, line 188: This change is unrelated to the actual PR.

.claude/skills
```diff
@@ -1 +0,0 @@
-../.agents/skills
```
jychp, line 1: blocking: This deletes the `.claude/skills` symlink, unrelated to the API Gateway fix, and is paired with a corrupted `.gitignore`: that file has been converted to CRLF and its final `.claude/` line is UTF-16-encoded (10 embedded null bytes), so git now classifies it as binary (`Bin 454 -> 507`) and adds bogus `!.claude/skills` / `.claude/` entries. (The `.gitignore` problem cannot be flagged inline because the file is now binary.) Both are local environment artifacts: restore the original LF `.gitignore` from master and drop this symlink deletion.

jychp (changes requested):

g-k-s-03:
Hi @jychp — all three points addressed. Dropped the Organizations change, restored the .gitignore, and removed the .claude/skills deletion. The branch now contains only the apigateway fix (1 commit, 2 files). Please re-review when you get a chance.

jychp (approved):
**Re-review: Approve (pending manual submission)**

The fix is correct. Removing the `else: return []` block lets the loop visit every stage instead of bailing on the first stage without a `clientCertificateId`. I traced every branch: there is no case where the old code returned a certificate and the new code drops one, so this is strictly more correct.

- No caller regression from `[]` -> `None`: the only consumer (`get_rest_api_details`) and `transform_rest_api_details` both gate on truthiness (`if certificate:`), and both `[]` and `None` are falsy.
- Addresses #2895 more completely than the issue proposed: replacing `return []` with `return None` alone would not have fixed the dropped-certificate bug; removing the whole `else` block fixes both the type inconsistency and the early-return drop.
- Tests are well-targeted: mixed-ordering case asserts `get_client_certificate` is called exactly once and returns the prod cert; all-no-cert case returns `None`.

Non-blocking, pre-existing (out of scope for this PR, no change requested):
- Last-cert-wins: when multiple stages each carry a distinct `clientCertificateId`, only the last is returned (the model stores one certificate per REST API). Pre-existing modeling limitation.
- `get_rest_api_client_certificate` type hint `stages: Dict` is actually a list.

No blocking issues.

g-k-s-03:
Hi @jychp  could you please approve the workflow run so CI can execute and auto-merge can proceed? Thank you

PR #2921, opened 2026-06-16
feat(circleci): add CircleCI intel module

jychp:
### Type of change
- [ ] Documentation update

### Summary
Adds a new intel module for **CircleCI** (API v2, `Circle-Token` auth) to inventory CI/CD assets for security analysis.

The module is built on the modern declarative data model (`CartographyNodeSchema`, `load()`, `GraphJob.from_node_schema()`) and follows the standard GET -> TRANSFORM -> LOAD -> CLEANUP pattern, mirroring the Vercel module.

**Nodes (20):**
- Org-scoped: `CircleCIOrganization` (`:Tenant`), `CircleCIUser`, `CircleCIContext`, `CircleCIContextEnvVar`, `CircleCIOidcConfig`, `CircleCIGroup`, `CircleCIPolicy`, `CircleCIEnvironment`, `CircleCIComponent`, `CircleCIProject`
- Project-scoped: `CircleCIProjectEnvVar`, `CircleCICheckoutKey`, `CircleCIWebhook`, `CircleCISchedule`, `CircleCIPipeline`, `CircleCIPipelineDefinition`, `CircleCITrigger`, `CircleCIProjectOidcConfig`

**Cross-module / cross-node relationships (best-effort, `OPTIONAL MATCH`):**
- `(:CircleCIProject)-[:BUILDS]->(:GitHubRepository|:GitLabProject)` joined on the repo URL
- `(:CircleCIOrganization)-[:ASSOCIATED_WITH]->(:GitHubOrganization)` joined on login
- `(:CircleCIContext)-[:RESTRICTED_TO]->(:CircleCIProject)` (one-to-many, from context restrictions)

**Design notes worth highlighting:**
- **Project discovery:** API v2 has no "list projects in an org" endpoint, so projects are discovered from each org's `/pipeline` feed and supplemented by `--circleci-project-slugs`. Because that discovery is partial, `CircleCIProject` nodes are upsert-only (never auto-deleted) to avoid removing still-valid projects; per-project sub-resources are still cleaned within their own `PROJECT_ID` scope.
- **Secrets:** the API never returns secret values in clear text. Context env vars expose no value at all; project env vars expose only a masked value (`xxxx` + last 4). The module stores only what the API returns.
- **Out of scope:** the CI -> cloud OIDC trust edge (`CircleCIOidcConfig.audience` -> AWS IAM OIDC provider) is left for a follow-up.

### Related issues or links
CircleCI API v2 reference: https://circleci.com/docs/api/v2/

### How was this tested?
- Added integration tests for every resource and relationship under `tests/integration/cartography/intel/circleci/` (21 tests). Each patches the module's `get()` boundary, calls `sync()`, and asserts nodes + relationships with `check_nodes` / `check_rels`.
- All endpoint response shapes were verified against the live CircleCI OpenAPI spec (`circleci.com/api/v2/openapi.json`), including nested objects that must be flattened (pipeline-definition `repo`, deploy `labels`) and the policy-bundle `name -> [doc]` shape.

#### General

#### Proof of functionality

- [ ] Included Cartography sync logs from a real environment. (Integration tests mock the API boundary; no real-tenant logs attached.)

- [x] Updated the schema documentation (`docs/root/modules/circleci/`, included in `docs/root/usage/schema.md`).

### Notes for reviewers
- Suggested reading order: `cartography/intel/circleci/__init__.py` (dispatch + sync ordering), then the per-resource modules.
- Relationship-modelling choice: edges whose join key lives in the source node's own data are plain `other_relationships`; a MatchLink was intentionally avoided since no edge required a third-source/independent-lifecycle link.
- Projects are synced before org-scoped resources so `RESTRICTED_TO` and `HAS_COMPONENT` links can attach to existing project nodes in the same run.


cartography/intel/circleci/organizations.py
```diff
@@ -0,0 +1,84 @@
...
+    # /me/collaborations returns a bare JSON array, not a paginated envelope.
+    req = api_session.get(f"{base_url}/me/collaborations", timeout=_TIMEOUT)
+    req.raise_for_status()
+    return req.json()
+
+
+def transform(raw: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    orgs = []
+    for org in raw:
+        vcs_type = org.get("vcs_type")
```
kunaals, line 43: CircleCI’s `/me/collaborations` response uses `vcs-type`, not `vcs_type`. As written, `vcs_type` is always `None`, so `vcs_login` is never derived and the `CircleCIOrganization -> GitHubOrganization` association won’t form. The fixture should use the real `vcs-type` shape too.

cartography/intel/circleci/checkout_keys.py
```diff
@@ -0,0 +1,88 @@
...
+    raw: list[dict[str, Any]],
+    project_slug: str,
+) -> list[dict[str, Any]]:
+    return [
+        {
+            "id": f"{project_slug}:{item['fingerprint']}",
+            "fingerprint": item["fingerprint"],
+            "type": item.get("type"),
+            "preferred": item.get("preferred"),
+            "public_key": item.get("public_key"),
```
kunaals: The checkout-key API returns `public-key` and `created-at`, not `public_key` / `created_at`. This will silently load `public_key=None` and `created_at=None` for real CircleCI data. Please read the hyphenated keys and update the fixture to match the API response.

cartography/intel/circleci/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+        api_session,
+        common_job_parameters,
+    )
+
+    # API v2 has no list-projects endpoint, so discover slugs from each org's
+    # pipeline feed and union with any operator-configured slugs.
+    project_slugs: set[str] = set(config.circleci_project_slugs or [])
+    for org in orgs:
+        if org.get("slug"):
+            project_slugs.update(
```
kunaals, line 96: This discovery call is outside `_run`/try-except, so one org’s `/pipeline` error aborts the entire CircleCI sync before configured slugs and downstream resources run. The rest of the module is careful to isolate per-resource API failures; this fan-out should get the same treatment.

kunaals (changes requested):

cartography/intel/circleci/organizations.py
jychp: Good catch to flag, but I verified this against the live API and it's the opposite: `GET /me/collaborations` actually returns `vcs_type` (snake_case), not `vcs-type`. The published OpenAPI spec / docs show the hyphenated form but they're out of date here.

Evidence from a real org:
```
$ curl -s -H "Circle-Token: $CT" https://circleci.com/api/v2/me/collaborations
[ { "vcs_type": "circleci", "slug": "circleci/...", "name": "SubImage", "id": "...", "avatar_url": "..." } ]
```
And end-to-end (real token → Neo4j) the org node came out with `vcs_type = "circleci"` populated — if the field were `vcs-type`, `org.get("vcs_type")` would have yielded `None`.

Note the spec is inconsistent across endpoints: `vcs_type` (collaborations) and `verify_tls`/`signing_secret` (webhooks) are snake_case in the live API, whereas the older checkout-key endpoint really does use hyphens (`public-key`, `created-at`) — I reverted that one per your other comment. So I've kept `vcs_type` as-is here.

cartography/intel/circleci/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    Run a resource sync, isolating expected per-resource API failures. On an HTTP
+    error the resource is skipped entirely - because its load/cleanup never runs,
+    previously-ingested data is preserved (no destructive empty-then-cleanup) - and
+    the rest of the module continues. Returns the sync's result, or None if it
+    raised, so callers can distinguish "failed" (None) from "empty" ([]).
+    """
+    try:
+        return fn(*args)
+    except requests.exceptions.HTTPError as exc:
```
kunaals: could we catch `requests.exceptions.RequestException` here instead? the retry adapter raises `RetryError` once 429/5xx retries are exhausted, and `RetryError` isn't an `HTTPError`, so a persistent rate limit or server error can still abort the whole circleci sync. the per-project handler below has the same issue.

kunaals (approved):
approving with one more nit

PR #2922, opened 2026-06-16
refactor(analysis): add typed analysis jobs

jychp:
### Type of change
- [ ] Documentation update

### Summary
This PR replaces hand-written JSON analysis jobs with typed `AnalysisJob` definitions colocated with each module's model package.

The typed model declares the effect each job creates or updates, so cleanup Cypher is generated automatically and analysis jobs can be inspected by added relationships/properties. Existing JSON migration and cleanup-only jobs remain as JSON.

The branch also fixes generated cleanup batching so iterative cleanup statements use `WITH ... LIMIT $LIMIT_SIZE`, and widens `run_scoped_analysis_job()` to accept typed `AnalysisJob` objects.

### Related issues or links

- Fixes #

### Breaking changes

None expected. The migrated analysis jobs are still invoked from the same sync paths. Some previously uncleaned analysis-created ontology edges/properties now get generated stale cleanup from their declared effects.

### How was this tested?

- `uv run --frozen pytest tests/unit/cartography/graph tests/unit/cartography/test_util.py`
- `uv run --frozen pytest tests/unit/cartography/graph/test_analysis.py tests/unit/cartography/graph/test_analysis_cleanup_coverage.py tests/unit/cartography/test_util.py`
- `uv run --frozen pytest tests/integration/cartography/intel/kubernetes/test_exposure.py tests/integration/cartography/intel/microsoft/intune/test_intune.py::test_sync_intune tests/integration/cartography/intel/ontology/test_resolved_images.py tests/integration/cartography/intel/azure/test_functions.py tests/integration/cartography/intel/aws/test_lambda.py tests/integration/cartography/intel/microsoft/entra/test_applications.py tests/integration/cartography/intel/ontology/test_devices.py tests/integration/cartography/data/jobs/test_syntax.py`
- `git diff --check`

#### General

#### Proof of functionality

### Notes for reviewers

Review focus: generated cleanup Cypher in `cartography/graph/analysis.py`, and the per-module `models/*/analysis.py` conversions. The remaining JSON analysis jobs are migration/cleanup-only jobs intentionally left in place.


cartography/models/ontology/analysis.py
```diff
@@ -0,0 +1,482 @@
...
+SUPPLY_CHAIN_SOURCE_FILE = AnalysisJob(
+    name="Enrich PACKAGED_FROM with source_file from Image provenance",
+    short_name="supply_chain_source_file",
+    statements=(
+        AnalysisStatement(
+            match="MATCH (i:Image)-[r:PACKAGED_FROM]->() WHERE r.dockerfile_path IS NULL AND i.source_file IS NOT NULL",
+            effects=(
+                SetRelationshipProperty(
+                    "r",
+                    "dockerfile_path",
```
kunaals, line 616: this generated cleanup now removes every existing `r.dockerfile_path` before re-setting only the `image.source_file` fallback case. that changes the ownership model for this property: github/gitlab matchlinks also write `dockerfile_path` with higher-fidelity dockerfile analysis data, and the old json job only filled gaps when the property was null.

can we either keep this as a raw query without generated property cleanup, or add an explicit escape hatch for relationship properties that are co-owned by ingestion/matchlinks? i'd also add a regression test with an existing matchlink-populated `dockerfile_path` to make sure this analysis job does not wipe it.

cartography/models/ontology/analysis.py
```diff
@@ -0,0 +1,482 @@
...
+            ),
+        ),
+        AnalysisStatement(
+            match="MATCH (u:User)-[:HAS_ACCOUNT]->(:OktaUser)-[:MEMBER_OF_OKTA_GROUP]->(:OktaGroup)-[:APPLICATION]->(a:ThirdPartyApp)",
+            effects=(
+                AddRelationship(
+                    "u",
+                    "AUTHORIZED",
+                    "a",
+                    properties={"scopes": []},
```
kunaals: this second merge path sets `r.scopes = []` for okta group-derived authorization. if the same user/app also has a direct `AUTHORIZED|APPLICATION` path, this statement can run after the direct one and overwrite real oauth scopes with an empty list.

the previous json job only set scopes on the direct authorization path, so i think this needs to preserve that behavior. one option is to avoid writing `scopes` from the group path, or only write the empty list when there is no current direct authorization path for the same user/app.

cartography/data/jobs/analysis/aws_ec2_asset_exposure.json
```diff
@@ -1,53 +0,0 @@
-{
```
kunaals, line 1: since this pr deletes the json analysis job files, the docs/comments that still link to them are going to point at dead paths after merge. i found references to these deleted jobs in the analysis-job developer guide and module schema docs.

can we update those references to point at the new typed job definitions, or keep a json example around if the docs are still meant to teach json-authored analysis jobs?

cartography/analysis/aws/analysis.py
```diff
@@ -0,0 +1,266 @@
...
+AWS_EC2_KEYPAIR_MATCHING_FINGERPRINT = AnalysisJob(
+    name="Analysis jobs for EC2 Key Pairs matching fingerprints",
+    short_name="aws_ec2_keypair_analysis_matching_fingerprint",
+    statements=(
+        AnalysisStatement(
+            match="MATCH (k1:EC2KeyPair) MATCH (k2:EC2KeyPair) WHERE id(k1) < id(k2) AND k1.keyfingerprint = k2.keyfingerprint",
+            effects=(
+                AddRelationship(
+                    "k1",
+                    "MATCHING_FINGERPRINT",
```
kunaals, line 234: small behavior change here: the old job used `$UPDATE_TAG` for `MATCHING_FINGERPRINT.firstseen`, but the generated relationship now falls back to neo4j `timestamp()`. other migrated jobs that had `$UPDATE_TAG` firstseen preserve it explicitly, so i think this one should pass `firstseen=Expr("$UPDATE_TAG")` too unless the timestamp change is intentional.

kunaals (dismissed):

docs/root/dev/writing-analysis-jobs.md
heryxpc, line 1: Can we add more code and details on the syntax for AnalysisJobs?
Referring existing code as an example is ok, but there are multiple primitives (`SetProperty`, `AddToSet`, `Expr`, etc.) that one single example will make it hard to capture all the building blocks around it.

A graphical description of the systanx (e.g. cause-effect diagram) in markdown could be easily read by humans and bots to understand how to build the job with the existing primitives.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+            parent_job_sequence_num=sequence_num,
+        )
+
+
+@dataclass(frozen=True)
+class SetProperty:
+    node: str
+    property: str
+    value: Any
+    label: str = ""
```
heryxpc: Is it right to have it default to ""?, this means that `not SetProperty.label` will be false and I think certain property effects won't run.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+    source: str
+    rel: str
+    target: str
+    source_label: str = ""
+    target_label: str = ""
+    rel_alias: str = "r"
+    properties: dict[str, Any] | None = None
+    undirected: bool = False
+    firstseen: Any = None
+    scoped_to: Literal["source", "target"] = "source"
```
heryxpc, line 123: how are used in conjunction `ScopedTo` and `scoped_to`?

heryxpc (approved):
Left a few comments, but looks awesome this new way to build Analysis jobs.

.agents/skills/analysis-jobs/references/examples.md
```diff
@@ -1,17 +1,23 @@
...
 def start_gcp_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
     # ... sync all orgs, folders, projects, and resources ...
 
-    run_analysis_job("gcp_compute_asset_inet_exposure.json", neo4j_session, common_job_parameters)
-    run_analysis_job("gcp_gke_asset_exposure.json",          neo4j_session, common_job_parameters)
-    run_analysis_job("gcp_gke_basic_auth.json",              neo4j_session, common_job_parameters)
-    run_analysis_job("gcp_compute_instance_vpc_analysis.json", neo4j_session, common_job_parameters)
+    run_analysis_job(GCP_COMPUTE_INSTANCE_VPC_ANALYSIS, neo4j_session, common_job_parameters)
+    for job in (GCP_GKE_ASSET_EXPOSURE, GCP_GKE_BASIC_AUTH):
+        run_analysis_job(job, neo4j_session, common_job_parameters)
```
achantavy: Nit: I don't see a lot of value in having the `for ..` construct here. It adds nesting. Just keep the explicit `run_analysis_job()` bits.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+    cleanup_iterationsize: int = 10000
+
+    def __post_init__(self) -> None:
+        if not self.statements:
+            raise ValueError("AnalysisJob requires at least one statement.")
+
+    def relationships_added(self) -> tuple[RelationshipEffect, ...]:
+        relationships: list[RelationshipEffect] = []
+        for effect in self._effects():
+            if isinstance(effect, AddRelationship):
```
achantavy: Avoid isinstance

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+                )
+                if rel_effect not in relationships:
+                    relationships.append(rel_effect)
+        return tuple(relationships)
+
+    def properties_set(self) -> tuple[PropertyEffect | RelationshipPropertyEffect, ...]:
+        properties: list[PropertyEffect | RelationshipPropertyEffect] = []
+        for effect in self._effects():
+            prop_effect: PropertyEffect | RelationshipPropertyEffect | None = None
+            if isinstance(effect, SetProperty):
```
achantavy: Avoid isinstance

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+            if isinstance(effect, (PropertyEffect, RelationshipPropertyEffect)):
+                statements.append(
+                    self._cleanup_statement(effect, parent_name, len(statements) + 1)
+                )
+
+        for offset, statement in enumerate(self.statements, start=len(statements) + 1):
+            statements.append(statement.to_graph_statement(parent_name, offset))
+
+        for effect in cleanup_effects:
+            if isinstance(effect, RelationshipEffect):
```
achantavy: Avoid isinstance

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+    def _effects(self) -> tuple[StatementEffect, ...]:
+        return tuple(
+            effect for statement in self.statements for effect in statement.effects
+        )
+
+    def _cleanup_effects(self) -> tuple[AnalysisEffect, ...]:
+        cleanups: list[AnalysisEffect] = []
+        for effect in self._effects():
+            cleanup: AnalysisEffect
+            if isinstance(effect, SetProperty):
```
achantavy: Avoid isinstance

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+    firstseen_value = _cypher_literal(effect.firstseen or Expr("timestamp()"))
+    return (
+        f"MERGE {rel}\n"
+        f"ON CREATE SET {effect.rel_alias}.firstseen = {firstseen_value}\n"
+        f"SET {effect.rel_alias}.lastupdated = $UPDATE_TAG{property_assignments}"
+    )
+
+
+def _cypher_literal(value: Any) -> str:
+    if isinstance(value, Expr):
```
achantavy: Avoid isinstance

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+    value: str
+
+
+StatementEffect = (
+    SetProperty | SetProperties | SetRelationshipProperty | AddToSet | AddRelationship
+)
+
+
+def _compile_effect(effect: StatementEffect) -> str:
+    if isinstance(effect, SetProperty):
```
achantavy: Avoid isinstance

cartography/intel/aws/__init__.py
```diff
@@ -17,6 +17,15 @@
 from cartography.intel.aws.util.common import parse_and_validate_aws_account_ids
 from cartography.intel.aws.util.common import parse_and_validate_aws_regions
 from cartography.intel.aws.util.common import parse_and_validate_aws_requested_syncs
+from cartography.models.aws.analysis import AWS_EC2_ASSET_EXPOSURE_JOBS
```
achantavy: `cartography.models` is more meant for data models i.e. subclasses of `CartographyNodeSchema` or `CartographyRelSchema`.

I think we should instead add a new top level concept: `cartography.analysis.<provider>`. This becomes a "registry" for the analysis jobs, and we keep the analysis layers all in one place.

Then it'd also be easier to call out cross-module analysis jobs, and we don't need to hunt in all `cartography.models.<provider>.<resource>.analysis` paths to find analysis jobs.

cartography/util.py
```diff
@@ -119,6 +120,12 @@ def run_analysis_job(
         The job file must be a valid JSON file containing GraphJob-compatible
         query definitions.
     """
+    if isinstance(filename, AnalysisJob):
```
achantavy: Avoid `isinstance`, especially in a core function like this. Let's take advantage of explicit type hints. If there's a transition period where we have to have a separate function or a shim that will get deprecated during this transition, that is totally fine.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+            raise ValueError(
+                "AnalysisStatement accepts query or match/effects, not both."
+            )
+        if not self.query and (not self.match or not self.effects):
+            raise ValueError("AnalysisStatement requires query or match/effects.")
+
+    def compile_query(self) -> str:
+        if self.query:
+            return self.query
+        assert self.match
```
achantavy: [question] why assert here instead of raising a value error?

cartography/graph/analysis.py
achantavy: +1, much more pythonic to have None as default arg for function params

tests/unit/cartography/graph/test_analysis.py
```diff
@@ -0,0 +1,337 @@
...
+                "HAS",
+                "e",
+                source_label="AWSLambda",
+                target_label="ECRImage",
+            ),
+        ),
+    )
+
+    # Act and assert
+    assert statement.compile_query() == (
```
achantavy: Thank you for doing these. For style, can also use `clean_query_list()`: https://github.com/cartography-cncf/cartography/blob/ae6be21fa0ba877609901e6c56ec5f7b6d460575/tests/unit/cartography/graph/test_cleanupbuilder.py#L37-L55

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+@dataclass(frozen=True)
+class AddRelationship:
+    source: str
+    rel: str
+    target: str
+    source_label: str = ""
+    target_label: str = ""
+    rel_alias: str = "r"
```
achantavy, line 118: I would add an inline docstring describing under what situations to override this

cartography/util.py
```diff
@@ -74,7 +75,7 @@ def backoff_handler(details: Dict) -> None:
 
 
 def run_analysis_job(
-    filename: str,
+    filename: str | AnalysisJob,
```
achantavy: Prefer shim or a separate function for now instead of doing faked override functions in python haha

cartography/graph/analysis.py
```diff
@@ -0,0 +1,400 @@
...
+@dataclass(frozen=True)
+class RelationshipEffect:
+    source_label: str
+    rel_label: str
+    target_label: str
+    properties: tuple[str, ...] = ()
+    direction: LinkDirection | None = LinkDirection.OUTWARD
+    scoped_to: Literal["source", "target"] = "source"
+
+    def cleanup_query(self, scope: ScopedTo | None) -> str:
```
achantavy: Currently all query builder logic in cartography is located in [cartography.graph](https://github.com/cartography-cncf/cartography/tree/master/cartography/graph), and I think we should keep it there. 

I also don't think the analysis object itself should have the behavior defined on the object. For consistency with the project, I think we should define cartography/graph/analysisbuilder.py, and then declare a set of library functions that accept `AnalysisStatement` and `RelationshipEffect` and the other new objects and generate new queries.

achantavy (changes requested):
First pass. The interface you've defined is very neat and looks easy to use - thanks for doing this.

Left comments mostly around code organization. I didn't have a chance to look deeply at query building failure cases but I'll do that later.

.agents/skills/analysis-jobs/references/examples.md
```diff
@@ -23,39 +47,39 @@ def _sync_one_account(...) -> None:
     # ... sync resources ...
 
     # scoped per-account
-    run_scoped_analysis_job("aws_ec2_iaminstanceprofile.json", neo4j_session, common_job_parameters)
+    run_typed_analysis_job(AWS_EC2_IAM_INSTANCE_PROFILE, neo4j_session, common_job_parameters)
 
     # cross-account, but called per account loop
```
achantavy: Is this comment still valid? Could remove if not. The nesting of the call to AWS_LAMBDA_ECR is now the same level as AWS_EC2_IAM_INSTANCE_PROFILE.

cartography/analysis/aws/s3/analysis.py
```diff
@@ -0,0 +1,108 @@
+from cartography.graph.analysis import AnalysisJob
+from cartography.graph.analysis import AnalysisStatement
+from cartography.graph.analysis import Expr
+from cartography.graph.analysis import ScopedTo
+from cartography.graph.analysis import SetProperties
+
+AWS_S3ACL_ANALYSIS = AnalysisJob(
+    name="AWS S3 Acl exposure analysis",
+    short_name="aws_s3acl_analysis",
+    scope=ScopedTo("AWSAccount", "AWS_ID"),
```
achantavy: Note to self: if `ScopedTo` is not specified, what happens if `$AWS_ID` is still specified?

cartography/analysis/aws/s3/analysis.py
```diff
@@ -0,0 +1,108 @@
...
+            MATCH (acl:S3Acl)-[:APPLIES_TO]->(bucket:S3Bucket)<-[:RESOURCE]-(aws:AWSAccount{id: $AWS_ID})
+            WHERE acl.uri IN ['http://acs.amazonaws.com/groups/global/AllUsers', 'http://acs.amazonaws.com/groups/global/AuthenticatedUsers']
+            AND acl.permission = 'READ'
+            """,
+            effects=(
+                SetProperties(
+                    "bucket",
+                    {
+                        "anonymous_access": True,
+                        "anonymous_actions": Expr(
```
achantavy: As the analysis job author, I just want to make sure `anonymous_actions` becomes populated with a list. I feel like there has to be a cleaner way of doing this. 

I think `Expr` is too low level of a tool.

cartography/analysis/aws/analysis.py
```diff
@@ -0,0 +1,267 @@
...
+            ),
+        ),
+    ),
+)
+AWS_LAMBDA_ECR = AnalysisJob(
+    name="Lambda functions with ECR images",
+    short_name="aws_lambda_ecr",
+    statements=(
+        AnalysisStatement(
+            match="MATCH (lmbda:AWSLambda)\nMATCH (e:ECRImage)\nWHERE e.digest = 'sha256:' + lmbda.codesha256",
```
achantavy: Let's get rid of all these `\n`s

cartography/analysis/aws/analysis.py
```diff
@@ -0,0 +1,267 @@
...
+                SetProperty(
+                    "elbv2", "exposed_internet", True, label="AWSLoadBalancerV2"
+                ),
+                SetProperty(
+                    "elbv2", "exposed_internet_type", None, label="AWSLoadBalancerV2"
+                ),
+            ),
+        ),
+        AnalysisStatement(
+            match="MATCH (cidr:AWSIpRange{range:'0.0.0.0/0'})-->(perm:AWSIpPermissionInbound)-->(sg:EC2SecurityGroup)<-[:MEMBER_OF_EC2_SECURITY_GROUP]-(elbv2:AWSLoadBalancerV2{scheme: 'internet-facing'})-->(listener:ELBV2Listener)\nWHERE perm.protocol = '-1' OR (listener.port>=perm.fromport AND listener.port<=perm.toport)",
```
achantavy: I think this analysis job should operate on specific labels only. There are some cases where it's ok to allow any rel label, but here we don't want shortcuts.

cartography/analysis/aws/analysis.py
```diff
@@ -0,0 +1,267 @@
...
+AWS_EC2_KEYPAIR_ANALYSIS_JOBS = (
+    AWS_EC2_KEYPAIR_PROPERTIES,
+    AWS_EC2_KEYPAIR_MATCHING_FINGERPRINT,
+)
+AWS_EKS_ASSET_EXPOSURE = AnalysisJob(
+    name="AWS EKS internet exposure",
+    short_name="aws_eks_asset_exposure",
+    statements=(
+        AnalysisStatement(
+            match="MATCH (cluster:EKSCluster) WHERE cluster.endpoint_public_access = true",
```
achantavy, line 254: Future idea: could be useful to have a registry that has a table of all global vs scoped jobs. 

This could enable a CI check like "analysis job touches account scopes but has no scope declared"

cartography/analysis/microsoft/entra/analysis.py
```diff
@@ -0,0 +1,24 @@
...
+
+ENTRA_APPLICATION_PROJECTION = AnalysisJob(
+    name="Ontology - Entra application projection",
+    short_name="ontology_entra_application_projection",
+    scope=ScopedTo("EntraTenant", "TENANT_ID"),
+    statements=(
+        AnalysisStatement(
+            match="MATCH (app:EntraApplication)<-[:RESOURCE]-(tenant:EntraTenant {id: $TENANT_ID}) OPTIONAL MATCH (tenant)-[:RESOURCE]->(sp:EntraServicePrincipal {app_id: app.app_id}) WITH app, sp",
+            effects=(
+                SetProperty(
```
achantavy, line 15: Side note: A neat side effect of having it in a data model is we now have an authoritative list of all properties and relationships created by analysis jobs. This can help with documentation.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,148 @@
...
+    label: str
+    id_param: str
+    id_property: str = "id"
+    rel_label: str = "RESOURCE"
+
+
+@dataclass(frozen=True)
+class AnalysisStatement:
+    query: str | None = None
+    comment: str = ""
```
achantavy: Should we init as `None`? Or we could just do 

`comment: str` 

and leave it like that.

Consider [non-block]: Ideally we'd make comments be mandatory either at the job level or the statement level since we have `__comment__` field in the json.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,148 @@
...
+    node: str
+    property: str
+    value: Any
+    label: str | None = None
+
+
+@dataclass(frozen=True)
+class SetProperties:
+    node: str
+    properties: dict[str, Any]
```
achantavy: Can we use explicit typing instead of `Any` here?

I see that it gets used this way:

```
def _(effect: SetProperties) -> str:
    assignments = ", ".join(
        f"{effect.node}.{key} = {_cypher_literal(value)}"
        for key, value in effect.properties.items()
    )
    return f"SET {assignments}"
```

so I _think_ it should be safe to assume it's a `dict[str, str]`?

cartography/graph/analysis.py
```diff
@@ -0,0 +1,148 @@
...
+    label: str | None = None
+
+
+@dataclass(frozen=True)
+class AddRelationship:
+    source: str
+    rel: str
+    target: str
+    source_label: str = ""
+    target_label: str = ""
```
achantavy: Non-blocking nit: we could leave these initialized as `None` here and then have the query building code set it to `""` if it is None.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,148 @@
...
+    properties: dict[str, Any] | None = None
+    undirected: bool = False
+    firstseen: Any = None
+    # Override to "target" when the target, not the source, is under AnalysisJob.scope.
+    scoped_to: Literal["source", "target"] = "source"
+    cleanup_where: str = ""
+
+
+@dataclass(frozen=True)
+class Expr:
```
achantavy: This definitely needs some python API docstrings to explain usage. Same for all of the other classes in this file.


----

<me being confused>
For `Expr` specifically though, I think it does too many jobs. it expands variables in python land, it passes params to neo4j, it sends calls to neo4j. 

Chatgpt is telling me that we could potentially split this out to 

- `Var('type')`
- `Param('UPDATE_TAG')`
- `RawCypher('timestamp()`)

I'm not sure I'm a fan though. Will think on this.
</me being confused>

cartography/util.py
```diff
@@ -206,6 +218,37 @@ def run_analysis_and_ensure_deps(
...
+    )
+    if not resource_dependencies.issubset(requested_syncs):
+        logger.info(
+            f"Did not run {analysis_job_name} because it needs {resource_dependencies} to be included "
+            f"as a requested sync. You specified: {requested_syncs}. If you want this job to run, please change your "
+            f"CLI args/cartography config so that all required resources are included.",
+        )
+        return
+
+    if isinstance(analysis_job, tuple):
```
achantavy: \>:( lol

cartography/graph/analysis.py
```diff
@@ -0,0 +1,148 @@
...
+from dataclasses import dataclass
+from typing import Any
+from typing import Literal
+from typing import Sequence
+
+from cartography.models.core.relationships import LinkDirection
+
+
+@dataclass(frozen=True)
+class ScopedTo:
```
achantavy: Can we rename to `CleanupScopedTo`? This way it's more clear that this only takes effect during the cleanup. I initially thought that this was related to performing the initial writes

cartography/graph/analysis.py
```diff
@@ -0,0 +1,148 @@
...
+    property: str
+    value: Any
+    label: str | None = None
+
+
+@dataclass(frozen=True)
+class SetProperties:
+    node: str
+    properties: dict[str, Any]
+    label: str | None = None
```
achantavy, line 65: Helpful to document what each field is used for. E.g. 
- `node` is a variable reference to the node and is needed for setting the property.
- `label` is necessary for the cleanup because when we remove the property we need to be able to filter by the label.

achantavy (commented):
took a deeper read

cartography/analysis/microsoft/entra/analysis.py
jychp: Yes, that is one of the goals of moving analysis jobs into typed models. Once this lands, we should be able to generate docs/listings from the declared effects. I would keep that as a follow-up PR so this one stays focused on the migration and cleanup generation.

cartography/analysis/aws/analysis.py
jychp: That's an interesting safe-guard, but it adds a lot of complexity and should be deferred to a follow-up PR IMO.

.agents/skills/analysis-jobs/SKILL.md
```diff
@@ -18,90 +18,97 @@ Use them when you need to:
...
+Typed jobs read as:
+
+```text
+AnalysisJob(scope=ScopedTo(...))
+    -> AnalysisStatement(match="MATCH ...", effects=(...))
+        -> SetProperty / AddToSet / AddRelationship / SetRelationshipProperty
 ```
 
-### Step 3 — Write the queries
+`label` is required for node-property effects so cleanup knows which label owns the property. Use `Expr("...")` for Cypher expressions like `$UPDATE_TAG`, `timestamp()`, and `coalesce(...)`; plain strings become quoted Cypher strings.
```
achantavy: Stale doc. Expr is now gone (yay)

.agents/skills/analysis-jobs/SKILL.md
```diff
@@ -18,90 +18,97 @@ Use them when you need to:
...
 
-```json
-{
-  "query": "MATCH (n:Node) WHERE n.stale = true WITH n LIMIT $LIMIT_SIZE DELETE n RETURN COUNT(*) AS TotalCompleted",
-  "iterative": true,
-  "iterationsize": 1000
-}
+```python
+AnalysisStatement(
+    query="MATCH (n:Node) WHERE n.stale = true WITH n LIMIT $LIMIT_SIZE DELETE n RETURN COUNT(*) AS TotalCompleted",
```
achantavy, line 91: I wonder if we should even let this be an option for the user anymore. The safest way is to do iterative by default since large analysis jobs can crash the database. Or we could even have the abstraction do a quick node count to check whether iteration or full computation would be faster - e.g. don't batch if the total nodes affected is less than 100k, or something.

I think this is a rough edge and defeats a bit of the purpose of having the abstraction.

cartography/analysis/aws/analysis.py
```diff
@@ -0,0 +1,292 @@
...
+)
+AWS_EC2_ASSET_EXPOSURE_AUTO_SCALING_GROUP = AnalysisJob(
+    name="AWS AutoScalingGroup internet exposure",
+    short_name="aws_ec2_asset_exposure_auto_scaling_group",
+    cleanup_iterationsize=1000,
+    statements=(
+        AnalysisStatement(
+            match="""
+            MATCH (instance:EC2Instance{exposed_internet: true})-[:MEMBER_AUTO_SCALE_GROUP]->(asg:AutoScalingGroup)
+            WITH distinct instance.exposed_internet_type as types, asg
```
achantavy: Note to self: make sure this pattern is documented among examples

cartography/analysis/kubernetes/analysis.py
```diff
@@ -0,0 +1,138 @@
...
+        ),
+    ),
+)
+K8S_CONTAINER_ASSET_EXPOSURE = AnalysisJob(
+    name="Kubernetes container internet exposure",
+    short_name="k8s_container_asset_exposure",
+    scope=CleanupScopedTo("KubernetesCluster", "CLUSTER_ID"),
+    statements=(
+        AnalysisStatement(
+            match="MATCH (cluster:KubernetesCluster{id: $CLUSTER_ID})-[:RESOURCE]->(pod:KubernetesPod{exposed_internet: true})-[:CONTAINS]->(c:KubernetesContainer) WITH DISTINCT c",
```
achantavy: Just curious: under what situations is the `with distinct` necessary vs not. I wonder if a module author could be confused by this.

cartography/graph/analysis.py
```diff
@@ -0,0 +1,213 @@
...
+@dataclass(frozen=True)
+class AnalysisStatement:
+    """One analysis statement, either raw Cypher or a match plus typed effects."""
+
+    query: str | None = None
+    comment: str | None = None
+    match: str | None = None
+    effects: Sequence[StatementEffect] = ()
+    iterative: bool = False
+    iterationsize: int = 0
```
achantavy, line 38: docstrings for all the params

achantavy: Oh also this is a cleanup iteration or a write iteration?

docs/root/dev/writing-analysis-jobs.md
```diff
@@ -7,10 +7,63 @@ In a nutshell, Analysis Jobs let you add your own customizations to Cartography
...
+Common primitives:
+
+| Primitive | Use |
+| --- | --- |
+| `SetProperty(node, property, value, label=...)` | Set one node property and derive cleanup for that label. |
+| `SetProperties(node, {property: value}, label=...)` | Set several node properties with one effect. |
+| `AddToSet(node, property, value, label=...)` | Add a value to a list property without duplicates. |
+| `AddRelationship(source, rel, target, source_label=..., target_label=...)` | MERGE a relationship, set `firstseen` on create, set `lastupdated` every run, and derive stale-edge cleanup. |
+| `SetRelationshipProperty(rel, property, value, source_label=..., rel_label=...)` | Set a relationship property and derive property cleanup. |
+| `Expr("cypher")` | Use a Cypher expression instead of a literal value. |
```
achantavy: This is now gone

achantavy (commented):
almost there! let's talk live on these to get this over the line

.agents/skills/analysis-jobs/SKILL.md
achantavy: As discussed, let this be a follow-up, because we could take advantage of this same problem for the regular clean-up.

cartography/analysis/aws/analysis.py
achantavy: I wonder if we can change this to 

```cypher
MATCH (instance:EC2Instance{exposed_internet: true})-[:MEMBER_AUTO_SCALE_GROUP]->(asg:AutoScalingGroup)
```

achantavy (approved):

PR #2924, opened 2026-06-17
feat(gcp): allow syncing specific GCP projects by ID

### LINKED ISSUE #2804: feat(gcp): Allow the sync of one GCP project
Allow the sync of one GCP project

## Description
It looks like Cartography does support only GCP ORG for the moment. We cannot do the sunc of a GCP project that does not belong to a GCP ORG.

## Motivation
For testing purpose and for some requirement, we need to sync only one specific GCP project.

rishichintala:
### Type of change

### Summary
Today Cartography can only sync GCP projects by first discovering a GCP Organization the identity can access, then walking the org/folder hierarchy. Projects that do not belong to an organization — or cases where the operator only has project-scoped credentials — cannot be synced at all (issue #2804).

This adds a `--gcp-project-ids` option (comma-separated) that syncs exactly those projects directly, bypassing organization and folder discovery:

- Each project is fetched with `get_project()` instead of `list_projects()` under an org, so it works without org access.
- Projects are loaded with a new `GCPStandaloneProjectSchema` that has no organization sub-resource. The optional `PARENT` edges to a `GCPOrganization`/`GCPFolder` are still created if those nodes already exist in the graph.
- The existing per-project resource sync (`_sync_project_resources`) and post-sync analysis jobs are reused. The analysis-jobs block was extracted into `_run_gcp_global_analysis_jobs()` so the org-based and standalone paths produce identical derived graph state.

Scope/limitations of this mode (documented in the GCP config docs):
- Organization-level data (orgs, folders, org IAM roles, inherited policy bindings) is intentionally not synced.
- Resource-level cleanup still works, scoped by `PROJECT_ID`. The standalone `GCPProject` node itself is not cleaned up, since it has no org to scope a cleanup against.

### Related issues or links
- Fixes #2804

### How was this tested?
- **Unit:** added `test_get_gcp_projects_by_ids_fetches_each_project_directly` (mocks the resource-manager client, asserts each project is fetched by name and normalised).
- **Integration:** added `test_sync_gcp_projects_by_ids_standalone` (project loads with no org in the graph) and `test_sync_gcp_projects_by_ids_links_existing_parent_org` (PARENT edge created when the org node already exists, but no org→project RESOURCE edge).
- `make test_lint` passes (black, isort, flake8, mypy).
- **Live GCP run** against a real project with `--selected-modules gcp --gcp-project-ids <project-id> --gcp-requested-syncs compute,iam,storage`. Sync completed with exit code 0; the graph contained the `GCPProject` node, **0** `GCPOrganization` nodes, and the project's Compute/IAM/Storage resources correctly attached via `(:GCPProject)-[:RESOURCE]->(...)`:

```
INFO:cartography.intel.gcp:GCP standalone project sync enabled for: <project-id>
INFO:cartography.intel.gcp.compute:Syncing GCP project <project-id> for Compute.
INFO:cartography.intel.gcp.iam:Syncing GCP project <project-id> for IAM.
INFO:cartography.intel.gcp.storage:Syncing GCP project <project-id> for Storage.
# result: 1 GCPProject, 0 GCPOrganization, 177 resources linked via :RESOURCE
```

### Checklist
- [x] Used the NodeSchema data model for the new schema/option.


cartography/intel/gcp/crm/projects.py
```diff
@@ -129,3 +134,88 @@ def sync_gcp_projects(
...
+    for project_id in project_ids:
+        proj = client.get_project(name=f"projects/{project_id}")
+        name_field = proj.name  # "projects/<number>"
+        project_number = name_field.split("/")[-1] if name_field else None
+        results.append(
+            {
+                "projectId": getattr(proj, "project_id", None),
+                "projectNumber": project_number,
+                "name": getattr(proj, "display_name", None),
+                "lifecycleState": proj.state.name,
```
jychp: suggestion: `get_project()` returns a project in any lifecycle state (e.g. `DELETE_REQUESTED`, suspended), whereas the org path relies on `list_projects()` returning only ACTIVE. A pending-deletion project ID would be ingested as a live GCPProject. Filter on `proj.state.name == "ACTIVE"` and skip/warn otherwise to match org-path semantics.

cartography/intel/gcp/crm/projects.py
```diff
@@ -129,3 +134,88 @@ def sync_gcp_projects(
...
+    Get and sync specific GCP projects by ID, without organization/folder discovery.
+
+    This powers the standalone single-project sync path. No cleanup job is run for the
+    project node itself: a standalone project has no organization sub-resource to scope
+    a cleanup against, and running a global cleanup would risk deleting projects synced
+    by the organization-based path. Resource-level cleanup (Compute, IAM, etc.) is still
+    handled per-project by the individual resource modules, scoped by PROJECT_ID.
+
+    :param project_ids: List of GCP project IDs to sync.
+    :return: List of projects synced (may be shorter than project_ids if some were skipped).
```
jychp: nitpick: docstring says the return "may be shorter than project_ids if some were skipped", but `get_gcp_projects_by_ids` raises on any inaccessible/missing ID and nothing is ever skipped. Fix the return-value description to match the fail-loud behavior.

cartography/intel/gcp/crm/projects.py
```diff
@@ -72,14 +73,18 @@ def transform_gcp_projects(data: List[Dict]) -> List[Dict]:
...
+        # Set parent fields based on parent type. A standalone project synced
+        # directly by ID may have no parent at all (it does not belong to an org
+        # or folder), in which case both fields stay None and no PARENT edge is
+        # created.
+        parent = project.get("parent") or ""
+        if parent.startswith("organizations"):
+            project["parent_org"] = parent
+        elif parent.startswith("folders"):
+            project["parent_folder"] = parent
+        elif parent:
```
jychp, line 94: nitpick: changing `else` to `elif parent:` is correct for standalone (no parent), but it also silences the previously-emitted warning when an org-path project returns an unexpected empty parent.

cartography/intel/gcp/__init__.py
```diff
@@ -847,6 +932,53 @@ def start_gcp_ingestion(
...
+        project_ids = [
+            project_id.strip()
+            for project_id in config.gcp_project_ids.split(",")
+            if project_id.strip()
+        ]
+        logger.info(
+            "GCP standalone project sync enabled for: %s", ", ".join(project_ids)
+        )
+
+        projects = sync_gcp_projects_by_ids(
```
jychp, line 966: suggestion/question: this path passes no `org_role_permissions_by_name`, and predefined roles (`roles/*`) plus org custom roles are synced only via `sync_org_iam()` in the org path. So `role_permissions_by_name` here holds only project-level custom roles, and any policy binding referencing a predefined role (`roles/owner`, `roles/editor`, ...) resolves to no permissions, leaving `permission_relationships` largely empty. The docs mention "organization-level IAM roles" but a reader would not expect Google predefined roles to be dropped too. Either sync predefined roles in standalone mode or state this limitation explicitly in docs/root/modules/gcp/config.md.

tests/integration/cartography/intel/gcp/test_crm_projects.py
```diff
@@ -197,3 +197,101 @@ def test_sync_gcp_projects_with_org_parent(
...
         == set()
     )
+
+
+@patch.object(
+    cartography.intel.gcp.crm.projects,
+    "get_gcp_projects_by_ids",
+    return_value=tests.data.gcp.crm.GCP_STANDALONE_PROJECTS,
+)
+def test_sync_gcp_projects_by_ids_standalone(_mock_get_projects, neo4j_session) -> None:
```
jychp, line 225: suggestion: these tests exercise `sync_gcp_projects_by_ids` directly, but the new `if config.gcp_project_ids:` branch in `start_gcp_ingestion` (id parsing, early return before org discovery, empty-projects warning, `_run_gcp_global_analysis_jobs` reuse) is untested. Add coverage so the wiring and early-return don't regress.

jychp (changes requested):
Thanks for your contribution, some minor changes and we should be able to merge it.

rishichintala:
Thanks for the thorough review @jychp! I've pushed `78a5397d` addressing all five comments, and merged `master` in so the branch is up to date. Detailed walkthrough below.

---

### 1. `get_project()` ingests non-ACTIVE projects (`crm/projects.py`)

You're right — `get_project()` returns a project in any lifecycle state, so a `DELETE_REQUESTED` or otherwise non-ACTIVE ID would have been ingested as a live `GCPProject`, unlike the org path which only ever sees ACTIVE projects from `list_projects()`.

Fixed in `get_gcp_projects_by_ids`: each fetched project is now checked, and non-ACTIVE ones are skipped with a warning rather than appended:

```python
lifecycle_state = proj.state.name
if lifecycle_state != "ACTIVE":
    logger.warning(
        "Skipping GCP project %s: lifecycle state is %s, not ACTIVE.",
        project_id, lifecycle_state,
    )
    continue
```

So the two code paths now have consistent ACTIVE-only semantics.

### 2. Return-value docstring (`crm/projects.py`)

Corrected. With the ACTIVE filter above, "may be shorter than project_ids" is true again — but for a *different* reason than the old wording implied. The docstring now distinguishes the two behaviours precisely:

> Return the list of projects that were synced. An inaccessible or non-existent project ID raises (fail loud), but projects that exist in a non-ACTIVE lifecycle state are skipped, so the returned list may be shorter than `project_ids`.

### 3. `elif parent:` silences the empty-parent warning (`crm/projects.py`)

Good catch — that was unintentional for the org path. `transform_gcp_projects` is shared by both paths, so I added a `warn_on_missing_parent` parameter (default `True`) rather than dropping or unconditionally restoring the warning:

- **Org path** (`load_gcp_projects`) uses the default → warns on an unexpected empty parent, as before.
- **Standalone path** (`load_standalone_gcp_projects`) passes `warn_on_missing_parent=False` → stays quiet, because a parentless project is legitimate there.

```python
elif warn_on_missing_parent:
    logger.warning(
        "Project %s has no parent; expected an organization or folder "
        "in the org-based sync path.", project["projectId"],
    )
```

### 4. Predefined/org roles dropped → empty permission relationships (`__init__.py`, `iam.py`, `models/gcp/iam.py`, docs)

This was the most important one. You're correct that the standalone path passed no `org_role_permissions_by_name`, so `role_permissions_by_name` only held project-level custom roles — any binding to a predefined role (`roles/owner`, `roles/editor`, …) resolved to no permissions, leaving `permission_relationships` largely empty. And, as you noted, a reader wouldn't expect Google *predefined* roles to be dropped.

I went with the stronger of your two options (sync rather than just document), and took it one step further than the permission map alone. Predefined roles are global and need no org access, so in the standalone path we now:

1. **Load them as `GCPRole` nodes** via a new `GCPStandalonePredefinedRoleSchema`. This is deliberately modelled like `GCPStandaloneProjectSchema`:
   - `sub_resource_relationship = None` — predefined roles are global, not owned by an org.
   - `scoped_cleanup = False`, and **no cleanup job runs** for these nodes. A global `GCPRole` cleanup would delete role nodes synced by the org-based path, so we intentionally skip it (same reasoning as the standalone `GCPProject` node).
   - Because it declares no relationships, loading it is non-destructive when both paths run: a later standalone load `MERGE`s on the role name, updates properties, and never touches the org's `RESOURCE` edge.

   The point of loading the nodes (not just the permission map) is that `(:GCPPolicyBinding)-[:GRANTS_ROLE]->(:GCPRole)` is a matchlink — without the target nodes present, those edges silently wouldn't form in standalone mode. Now they do.

2. **Thread their permission map** into `_sync_project_resources` via `org_role_permissions_by_name`, so bindings to predefined roles expand into permission relationships as they do on the org path.

Both are gated behind the same `iam` / `policy_bindings` requested-syncs check the org path uses:

```python
if requested_syncs is None or "iam" in requested_syncs or "policy_bindings" in requested_syncs:
    iam_client = build_client("iam", "v1", credentials=credentials)
    standalone_role_permissions_by_name = iam.sync_standalone_predefined_roles(
        neo4j_session, iam_client, config.update_tag,
    )
else:
    standalone_role_permissions_by_name = {}
```

Organization-level **custom** roles genuinely require org access, so they remain unresolved in this mode — that's an inherent limitation of not walking the org, not something we can fetch here. Rather than leave it implicit, I documented it in `docs/root/modules/gcp/config.md`:

> Google **predefined** roles are synced … their `GCPRole` nodes are loaded (letting `GRANTS_ROLE` edges form) and their permissions expand into permission relationships. … Organization-level **custom** roles are **not** synced in this mode (they require organization access), so any policy binding that references an org custom role will not link to a role node or expand into permission relationships.

### 5. Untested `start_gcp_ingestion` standalone branch (tests)

Added. New integration tests exercise the branch itself, not just `sync_gcp_projects_by_ids`:

- `test_start_gcp_ingestion_standalone_project_ids` — ID parsing/trim (incl. blank entries), org-discovery bypass (`get_gcp_organizations` asserted **not** called), project node load, predefined-role node load, that the permission map is passed into `_sync_project_resources`, and that the shared analysis jobs run.
- `test_start_gcp_ingestion_standalone_no_projects_returns_early` — the empty-projects early return (no resource sync, no analysis jobs).
- `test_standalone_predefined_role_enables_grants_role_edge` — end-to-end proof of comment 4: loads a predefined role + project + policy binding and asserts the `GRANTS_ROLE` edge resolves.

Plus unit coverage for the ACTIVE filter and both branches of the missing-parent warning.

---

**Verification:** all GCP unit tests (327) and the full GCP integration suite (112) pass locally against Neo4j 5. Happy to adjust any of the above — particularly the comment-4 approach if you'd prefer a narrower scope.

cartography/intel/gcp/crm/projects.py
```diff
@@ -129,3 +151,110 @@ def sync_gcp_projects(
...
+    be shorter than project_ids.
+
+    :param project_ids: List of GCP project IDs (e.g. ["my-project-1", "my-project-2"]).
+    :param credentials: GCP credentials to use for API calls.
+    :return: List of project dicts matching the shape returned by get_gcp_projects().
+    """
+    client = resourcemanager_v3.ProjectsClient(credentials=credentials)
+    results: List[Dict] = []
+    for project_id in project_ids:
+        proj = client.get_project(name=f"projects/{project_id}")
```
jychp: `get_gcp_projects_by_ids` lets `get_project()` raise on the first bad ID, so one inaccessible or misspelled project in a comma-separated `--gcp-project-ids` list aborts the entire sync and nothing loads. Either handle per-ID (log and skip the bad one, sync the rest) or document that any single invalid ID fails the whole run.

cartography/intel/gcp/iam.py
```diff
@@ -317,6 +318,72 @@ def build_role_permissions_by_name(
...
+    Unlike sync_org_iam, the loaded nodes have no organization sub-resource and are not
+    cleaned up; see GCPStandalonePredefinedRoleSchema. Organization-level *custom* roles
+    are not synced here — they require organization access.
+
+    :param neo4j_session: The Neo4j session.
+    :param iam_client: The IAM resource object created by googleapiclient.discovery.build().
+    :param gcp_update_tag: The timestamp of the current sync run.
+    :return: Mapping of predefined role name to its list of included permissions.
+    """
+    predefined_roles_raw = get_gcp_predefined_roles(iam_client)
```
jychp, line 386: `sync_standalone_predefined_roles` loads the entire predefined-role catalog (~1000+ `GCPRole` nodes) on every run and never cleans them up. In a standalone-only deployment, roles Google later deletes persist indefinitely. Confirm this is acceptable, or scope the load to roles referenced by the synced policy bindings.

tests/integration/cartography/intel/gcp/test_crm_projects.py
```diff
@@ -197,3 +215,297 @@ def test_sync_gcp_projects_with_org_parent(
...
+    cartography.intel.gcp.crm.orgs,
+    "get_gcp_organizations",
+    return_value=tests.data.gcp.crm.GCP_ORGANIZATIONS,
+)
+@patch.object(
+    cartography.intel.gcp.crm.projects,
+    "get_gcp_projects_by_ids",
+    return_value=tests.data.gcp.crm.GCP_STANDALONE_PROJECTS_WITH_ORG_PARENT,
+)
+def test_sync_gcp_projects_by_ids_links_existing_parent_org(
```
jychp, line 268: No test covers a standalone project whose `parent` references an org/folder that is not in the graph. Add a case asserting no PARENT edge is created when the referenced parent node is absent, to lock in the "skipped otherwise" behavior.

jychp (changes requested):
Thks for the followup, some nitpicking comments and we should be able to merge your contribution.

cartography/intel/gcp/crm/projects.py
rishichintala: Good catch — switched to per-ID handling. `get_gcp_projects_by_ids` now wraps `get_project()` in a try/except for `NotFound` / `PermissionDenied` / `Forbidden`: a single inaccessible or misspelled ID is logged and skipped so the remaining projects still sync. Transient errors are intentionally *not* caught and still propagate. Updated the docstring and `docs/root/modules/gcp/config.md` to match, and added a unit test (`test_get_gcp_projects_by_ids_skips_inaccessible_ids`).

cartography/intel/gcp/iam.py
rishichintala: Confirmed acceptable — and I've documented why in the code so the intent is explicit.

These predefined `GCPRole` nodes are global, cross-scope reference data shared between the standalone and organization-based paths (both `MERGE` on the role name). Not node-cleaning them follows the same convention Cartography already uses for other global reference nodes — e.g. AWS managed policies ([`aws/iam.py::cleanup_iam`](https://github.com/cartography-cncf/cartography/blob/master/cartography/intel/aws/iam.py)), which are deliberately *not* node-cleaned because deleting a shared node for one scope would erroneously remove it for another. Here, a global `GCPRole` cleanup run from the standalone path would delete role nodes the org path owns in a mixed deployment.

A few points that bound the concern:
- The catalog is bounded (~1.5k Google-managed roles) and `MERGE`'d on the role name, so it does not grow without bound across runs.
- The organization-based path already loads the same full predefined catalog via `load_org_roles`, so the node volume is consistent with existing behaviour.
- I added an explanatory comment on both `sync_standalone_predefined_roles` and `GCPStandalonePredefinedRoleSchema`.

On the "scope the load to roles referenced by policy bindings" alternative: I looked at it, but it isn't feasible without losing the `GRANTS_ROLE` edges. That edge is a matchlink created during the policy-binding load and only connects to a `GCPRole` node that already exists — but we only learn which roles are referenced *during* binding sync, so "load only referenced roles" and "form GRANTS_ROLE in one pass" are mutually exclusive without restructuring the policy-bindings flow. Happy to revisit if you'd prefer dropping the predefined-role nodes entirely (permission relationships would still resolve via the in-memory map).

tests/integration/cartography/intel/gcp/test_crm_projects.py
rishichintala: Added `test_sync_gcp_projects_by_ids_absent_parent_creates_no_edge`: it syncs a standalone project whose `parent` is `organizations/1337` *without* syncing that org, then asserts the org node is absent and no `(:GCPProject)-[:PARENT]->(:GCPOrganization)` edge is created — locking in the "skipped otherwise" matchlink behaviour.

rishichintala:
Thanks again @jychp — pushed the follow-up and merged latest `master`. Summary of each item:

**1. A single bad `--gcp-project-ids` entry aborted the whole run** — Fixed. `get_gcp_projects_by_ids` now handles each ID independently: an inaccessible or misspelled ID (`NotFound` / `PermissionDenied` / `Forbidden`) is logged and skipped so the remaining projects still sync. Transient errors are intentionally not caught and still propagate. Docstring + `config.md` updated; unit test `test_get_gcp_projects_by_ids_skips_inaccessible_ids` added.

**2. Predefined-role catalog volume / no cleanup** — Confirmed acceptable, and documented in code. These predefined `GCPRole` nodes are global, cross-scope reference data shared with the org-based path (both `MERGE` on the role name). Not node-cleaning them follows the same convention Cartography uses for AWS managed policies (`aws/iam.py::cleanup_iam`): deleting a shared node for one scope would erroneously remove it for another. The catalog is bounded (~1.5k Google-managed roles) and `MERGE`'d, and the org path already loads the same full catalog via `load_org_roles`. Scoping to only bindings-referenced roles isn't feasible without dropping the `GRANTS_ROLE` edges (the matchlink needs the role node to pre-exist the binding load, but referenced roles are only known during binding sync). Explanatory comments added on `sync_standalone_predefined_roles` and the schema.

**3. Missing absent-parent test** — Added `test_sync_gcp_projects_by_ids_absent_parent_creates_no_edge`, asserting no `PARENT` edge is created when the referenced org/folder isn't in the graph.

Full GCP unit and integration suites pass locally against Neo4j 5.

cartography/intel/gcp/crm/projects.py
```diff
@@ -129,3 +154,125 @@ def sync_gcp_projects(
...
+    :param project_ids: List of GCP project IDs (e.g. ["my-project-1", "my-project-2"]).
+    :param credentials: GCP credentials to use for API calls.
+    :return: List of project dicts matching the shape returned by get_gcp_projects().
+    """
+    client = resourcemanager_v3.ProjectsClient(credentials=credentials)
+    results: List[Dict] = []
+    for project_id in project_ids:
+        try:
+            proj = client.get_project(name=f"projects/{project_id}")
+        except (NotFound, PermissionDenied, Forbidden) as e:
```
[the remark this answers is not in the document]
rishichintala: Fixed in c9d9b011. Dropped `Forbidden` from the except tuple — since `PermissionDenied` is a subclass of `Forbidden`, catching `Forbidden` also silently swallowed other 403s (Resource Manager API or billing disabled), which are environment-level problems that should fail loudly. Now only `NotFound` and `PermissionDenied` are skipped per ID; any other `GoogleAPICallError` propagates. Removed the now-unused import.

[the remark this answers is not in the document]
rishichintala: Fixed in c9d9b011. Updated the `sync_gcp_projects_by_ids()` docstring: inaccessible/non-existent IDs (`NotFound` / `PermissionDenied`) and non-ACTIVE projects are now described as logged-and-skipped (so the returned list may be shorter than `project_ids`), and other errors still propagate and fail the sync.

cartography/intel/gcp/__init__.py
```diff
@@ -863,6 +948,76 @@ def start_gcp_ingestion(
...
+            iam_client = build_client("iam", "v1", credentials=credentials)
+            standalone_role_permissions_by_name = iam.sync_standalone_predefined_roles(
+                neo4j_session,
+                iam_client,
+                config.update_tag,
+            )
+        else:
+            standalone_role_permissions_by_name = {}
+
+        project_resources_result = _sync_project_resources(
```
jychp, line 1000: blocking: `common_job_parameters` has no `ORG_RESOURCE_NAME` in this path, so policy binding sync sends `scope=None` to CAI `batch_get_effective_iam_policies`; Google requires an organization, folder, or project scope, so the default standalone sync fails. Supply `projects/{project_id}` as the effective-policy scope for standalone projects.

cartography/intel/gcp/crm/projects.py
```diff
@@ -129,3 +153,130 @@ def sync_gcp_projects(
...
+    Load GCP projects that were fetched directly by ID (no organization context).
+
+    Uses GCPStandaloneProjectSchema, which has no organization sub-resource
+    relationship. Any PARENT edges to a GCPOrganization or GCPFolder are still
+    created if those nodes already exist in the graph, but are skipped otherwise.
+    """
+    # warn_on_missing_parent=False: a standalone project legitimately may have no
+    # parent (it does not belong to an org or folder), so a missing parent here is
+    # expected and should not emit a warning.
+    transformed_data = transform_gcp_projects(data, warn_on_missing_parent=False)
```
jychp, line 247: blocking: This load only merges the current `PARENT` edge, and the standalone schema never runs relationship cleanup. If a project changes parent or becomes parentless, its previous `PARENT` edge survives indefinitely; reconcile old parent edges for the requested project IDs before or after loading.

jychp (changes requested):

cartography/intel/gcp/__init__.py
rishichintala: Fixed in 2f7d146856ce14f272fa1f4fd48336f0ecd3fe5c.

Rather than injecting a synthetic `ORG_RESOURCE_NAME` into `common_job_parameters` (which would trigger org-binding load/cleanup side effects), I fixed it at the call site in `get_policy_bindings`:

```python
effective_scope = org_id or f"projects/{project_id}"
```

The org path is unchanged (still scopes to the org for inheritance); the standalone path now sends `projects/{project_id}` instead of `None`. Covered by a new parametrized unit test `test_get_policy_bindings_effective_scope` (org -> `organizations/1337`, standalone -> `projects/project-abc`).

cartography/intel/gcp/crm/projects.py
rishichintala: Fixed in 2f7d146856ce14f272fa1f4fd48336f0ecd3fe5c.

Added `cleanup_standalone_project_parents()`, called from `sync_gcp_projects_by_ids` right after the load. It deletes any `PARENT` edge on the synced projects whose `lastupdated` is stale:

```cypher
MATCH (p:GCPProject)-[r:PARENT]->()
WHERE p.id IN $project_ids AND r.lastupdated <> $UPDATE_TAG
DELETE r
```

So a project that changes parent or becomes parentless has its old edge reconciled away on the next run. Covered by a new integration test `test_sync_gcp_projects_by_ids_reconciles_stale_parent_edge`.

One nuance vs. the wording here: I scoped the cleanup to the projects **actually loaded** this run rather than the raw requested IDs. A requested project that was skipped (inaccessible / non-ACTIVE) is absent from the loaded set, so its still-valid edge is preserved instead of deleted from stale data — matching this module's preserve-on-failure pattern (e.g. the IAM-cleanup gating). Happy to switch to raw requested IDs if you'd prefer strict adherence to the comment.

PR #2937, opened 2026-06-20
fix(aws/route53): materialize Ip nodes so DNS_POINTS_TO relationships are created  

### LINKED ISSUE #2936: Route53 sync: A/AAAA records never create Ip nodes — DNS_POINTS_TO relationships silently missing
### Expected behavior

After a Route53 sync, querying the graph for DNS records and their resolved IPs should return results:

```cypher
MATCH (r:AWSDNSRecord)-[:DNS_POINTS_TO]->(ip:Ip) RETURN r.name, ip.id
```

`Ip` nodes should be created for every IP address in A and AAAA records, and `DNS_POINTS_TO` edges should connect them to their `AWSDNSRecord` nodes.

### Actual behavior

The query above returns no results. `Ip` nodes for Route53 A/AAAA records are never created during the sync, so the `DNS_POINTS_TO` relationship is silently never written to the graph.

### To Reproduce

1. Run a cartography sync with Route53 enabled.
2. Query the graph:
```cypher
MATCH (r:AWSDNSRecord {type: "A"})-[:DNS_POINTS_TO]->(ip:Ip) RETURN r.name, ip.id
```
3. Observe: no results, even though A records with IP addresses were synced.

- Cartography release version or commit hash: #178db0d
- Python version: Python 3.14.2
- OS: macOS

### Additional context

`AWSDNSRecordToIpRel` in `cartography/models/aws/route53/dnsrecord.py` declares `(AWSDNSRecord)-[:DNS_POINTS_TO]->(Ip)`. The `load()` framework uses `MATCH` to find relationship targets — it does not create them. No code in `cartography/intel/aws/route53.py` ever `MERGE`s `Ip` nodes, so the `MATCH` finds nothing and the edge is never created.

The TODO comment at line 238 of `route53.py` acknowledges this:

```python
# TODO consider creating IPs as a first-class node from here.
# Right now we just match on them from the A record.
```

Dreamstick9:
### Type of change

- [ ] Documentation update

### Summary

`AWSDNSRecordSchema` declares an `AWSDNSRecordToIpRel` relationship (`DNS_POINTS_TO`) from `AWSDNSRecord` nodes to `Ip` nodes. However, the `load()` framework uses `MATCH` to find relationship targets — it does not create them. No code in `route53.py` ever created `Ip` nodes, so the `MATCH` always found nothing and the `DNS_POINTS_TO` edge was silently never written to the graph.  
  
This meant that querying "what IP does this DNS A record point to?" always returned empty results, even though the IP data was present in the synced records.  
  
The fix adds a `load_ip_nodes()` function that collects all IP addresses from A and AAAA records and `MERGE`s them as `Ip` nodes before `load_a_records()` and `load_aaaa_records()` run. Once the `Ip` nodes exist, the existing schema relationship logic creates the `DNS_POINTS_TO` edges correctly. 

### Related issues or links

- Fixes #2936 

### How was this tested?

- Ran `make test_lint` — linter passes. 
- Ran existing integration tests in `tests/integration/cartography/intel/aws/test_route53_sync.py` — all pass. 

#### General

#### Proof of functionality

No schema changes were made. `AWSDNSRecordToIpRel` already existed in the schema — this PR only ensures the target `Ip` nodes are created so the declared relationship can be materialized.

### Notes for reviewers

The `Ip` node `MERGE` is idempotent and scoped only to IPs found in A/AAAA records during this sync. It does not affect `Ip` nodes created by other modules (e.g. EC2, elasticsearch). Cleanup of stale `AWSDNSRecord` nodes and their edges is already handled by the existing `GraphJob.from_node_schema(AWSDNSRecordSchema(), ...)` call in `cleanup_route53()`.

### DIFF
--- a/cartography/intel/aws/route53.py
+++ b/cartography/intel/aws/route53.py
@@ -9,6 +9,7 @@ import neo4j
 from cartography.client.core.tx import load
 from cartography.client.core.tx import load_matchlinks
 from cartography.client.core.tx import read_list_of_dicts_tx
+from cartography.client.core.tx import run_write_query
 from cartography.graph.job import GraphJob
 from cartography.intel.aws.util.botocore_config import create_boto3_client
 from cartography.models.aws.route53.dnsrecord import AWSDNSRecordSchema
@@ -235,8 +236,6 @@ def transform_all_dns_data(
 
                 if transformed_rs["type"] == "A":
                     all_a_records.append(transformed_rs)
-                    # TODO consider creating IPs as a first-class node from here.
-                    # Right now we just match on them from the A record.
                 elif transformed_rs["type"] == "AAAA":
                     all_aaaa_records.append(transformed_rs)
                 elif transformed_rs["type"] == "ALIAS":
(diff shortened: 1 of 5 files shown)


cartography/intel/aws/route53.py
```diff
@@ -270,6 +271,31 @@ def transform_all_dns_data(
...
+    ip_list = []
+    for record in records:
+        ip_list.extend(record.get("ip_addresses", []))
+    if not ip_list:
+        return
+    run_write_query(
+        neo4j_session,
+        """
+        UNWIND $IP_LIST as ip
+        MERGE (ip_node:Ip{id: ip})
```
jychp: For IPv6, this `MERGE (:Ip{id: ip})` merges on the `Ip` label, but EC2 ingests IPv6 as `EC2Ipv6Address` (extra label `Ip`) and merges on `EC2Ipv6Address`. If Route53 syncs before the EC2 IPv6 module, you get two distinct `:Ip` nodes for the same address, and the `AWSDNSRecordToIpRel` matcher then links the record to both, producing duplicate `DNS_POINTS_TO` edges. Match/reuse the existing node by id regardless of which module created it, or confirm module ordering makes this unreachable.

jychp (changes requested):

cartography/intel/aws/route53.py
Dreamstick9: Im sorry I did not look through that, though I had the direction backwards. EC2-first already worked, since MERGE on :Ip picks up the EC2 node through its extra label. It's route53 first that breaks. So route53 now just merges on :Ip, and the EC2 ipv6 loader claims any existing :Ip{id} node before its own merge.

Also rebased, the branch was way behind and #3020 renamed the labels I'd hardcoded. Tests cover both orders. Used claude code for assistance while working on this hope that does not affect the quality of this PR

PR #2942, opened 2026-06-23
fix(cloudflare): avoid infinite pagination on /accounts

thenadz:
### Type of change
- [ ] Documentation update

### Summary

The Cloudflare `GET /accounts` endpoint ignores the `page` query parameter: it
returns the **same non-empty page for every page number** while reporting
`result_info.total_pages = 1`. The `cloudflare-python` V4 paginator
(`SyncV4PagePaginationArray`) decides whether to fetch another page solely by
checking whether the current page came back empty — `next_page_info()` returns
`page + 1` unconditionally and never consults `result_info.total_pages`.

The combination means `for account in client.accounts.list()` **never
terminates**: every page is non-empty, so the SDK keeps requesting `page=2, 3,
…` forever. In practice the `cloudflare` sync stage hangs and rapidly exhausts
Cloudflare's documented global API limit of 1200 requests / 300s per user,
after which all calls return `HTTP 429`.

`cartography/intel/cloudflare/accounts.py::get()` is the only call site that
iterates this endpoint's paginator. The fix reads the single authoritative page
via `.result` instead of iterating:

```python
page = client.accounts.list()
return [account.to_dict() for account in page.result or []]
```

`total_pages` is always `1` for this endpoint, so a single page is complete and
correct. Other Cloudflare submodules (zones, members, roles, dns_records) page
correctly (their endpoints return an empty array past the last page) and are
unchanged.

### Related issues or links

- Upstream SDK issue (root cause, unresolved): cloudflare/cloudflare-python#2584
  — the same empty-page-only termination logic exists in `cloudflare-go`
  (`V4PagePagination.GetNextPage`) and `cloudflare-typescript`
  (`V4PagePaginationArray.nextPageInfo`), so the guard must live at the call
  site rather than waiting for an SDK fix.
- Cloudflare API limits: https://developers.cloudflare.com/fundamentals/api/reference/limits/

### How was this tested?

- Added a unit test (`test_get_accounts_reads_single_page_not_paginator`) that
  drives the real `get()` against a mock client whose `accounts.list()` return
  value mimics `SyncV4PagePaginationArray` — both iterable (with an **unbounded**
  `__iter__` reproducing the API's repeating page) and exposing `.result`. The
  test **passes** on the fix and **hangs** (times out) against the previous
  iterating code, confirming it guards the regression. The existing integration
  test patches `get()` out, so it cannot catch this.
- `make test_lint` passes locally (black, isort, flake8 clean on both files).
- Verified against a real Cloudflare account (1 account, 7 zones, 888 DNS
  records). Before: the `cloudflare` stage looped on `/accounts`, emitting
  hundreds of `Retrying request to /accounts` lines and never completing. After:

  ```
  INFO:cartography.sync:Starting sync stage 'cloudflare'
  INFO:cartography.client.core.tx:Loaded 120 CloudflareDNSRecord nodes
  INFO:cartography.client.core.tx:Loaded 619 CloudflareDNSRecord nodes
  INFO:cartography.client.core.tx:Loaded 67 CloudflareDNSRecord nodes
  ...
  INFO:cartography.sync:Finishing sync stage 'cloudflare'
  ```
  0 retries; stage completes in seconds.

#### General

#### Proof of functionality


cartography/intel/cloudflare/accounts.py
```diff
@@ -29,7 +29,16 @@ def sync(
...
-    return [account.to_dict() for account in client.accounts.list()]
+    # The Cloudflare /accounts endpoint ignores the `page` param: it echoes the
+    # same results for every page while reporting total_pages=1 (an API-side
+    # contract violation). The SDK's V4 paginator terminates only on an empty
+    # page and never consults result_info.total_pages, so iterating
+    # client.accounts.list() never stops -> hang + rate-limit exhaustion.
+    # Upstream SDK issue: cloudflare/cloudflare-python#2584 (unfixed; the same
+    # empty-page-only logic exists in cloudflare-go and cloudflare-typescript).
+    # Read the single authoritative page rather than iterating the paginator.
+    page = client.accounts.list()
```
jychp: `page.result` returns only the first page. `accounts.list()` honors `per_page` (default 20, max 50), and since the endpoint ignores `page` there is no way to fetch the overflow through this paginator. For a token scoped to more accounts than the page size, this silently drops the rest.

Can we find an other way to do that ?

jychp (changes requested):
Thanks for your contribution, added a comment on the usage of the list() method.

jychp:
@thenadz sorry for deleting the comment, miss clicked

thenadz:
@jychp No problem. Did you need me to restate? Basically with the limitations of upstream bug it doesn't appear to be possible to paginate at all (would be happy to be proven wrong!). Working from that understanding, best we can do is take the max page size of 50 and fail loudly if the total accounts count is greater than what we receive with that first page.

So that's what the update commit does. Much better than current mainline behavior that just spins in infinite loop, but imperfect due to upstream API limitations...

Very open to further improvements if you have thoughts?

cartography/intel/cloudflare/accounts.py
```diff
@@ -29,7 +32,23 @@ def sync(
...
+    # The SDK auto-paginator increments `page` and stops only on an empty page,
+    # which never terminates on /accounts (cloudflare/cloudflare-python#2584).
+    # /accounts caps per_page at 50 and has no cursor, so fetch a single
+    # max-size page and reconcile against total_count rather than iterating. If
+    # more accounts exist than were returned, fail loudly instead of syncing a
+    # partial set (correct whether or not `page` is honored).
+    page = client.accounts.list(per_page=MAX_ACCOUNTS_PER_PAGE)
+    accounts = [account.to_dict() for account in page.result or []]
+
+    total_count = page.result_info.total_count
```
jychp: Blocking: this can raise `AttributeError` and crash the whole cloudflare stage. `page.result_info` is typed `Optional[V4PagePaginationArrayResultInfo]`, and that model declares only `page` and `per_page` — `total_count` resolves only via pydantic `extra="allow"` *if the response includes it*. Verified on cloudflare 5.3.0: `V4PagePaginationArrayResultInfo.model_validate({'page':1,'per_page':50}).total_count` raises `AttributeError`. So this line crashes when `result_info` is None or the field is absent; the `is not None` guard below only covers present-but-None. Use `result_info = page.result_info; total_count = getattr(result_info, "total_count", None) if result_info else None`.

cartography/intel/cloudflare/accounts.py
```diff
@@ -29,7 +32,23 @@ def sync(
...
+    if total_count is not None and total_count > len(accounts):
+        raise RuntimeError(
```
jychp, line 48: Question: this hard `RuntimeError` aborts the entire cloudflare sync stage for any org scoped to >50 accounts — they can never sync any cloudflare data. Failing loudly over silent partial data is defensible (answers the earlier review comment), but confirm this is preferred over logging a warning and syncing the first page.

tests/unit/cartography/intel/cloudflare/test_cloudflare.py
```diff
@@ -1,11 +1,59 @@
...
+
+
+def _page(records, total_count):
+    # Mimic SyncV4PagePaginationArray: exposes .result and .result_info, and is
+    # iterable. Iteration is unbounded (repeats the first item) to reproduce the
+    # /accounts paginator hang - get() must NOT iterate it.
+    objs = [MagicMock(to_dict=lambda r=r: r) for r in records]
+    page = MagicMock()
+    page.result = objs
+    page.result_info = MagicMock(total_count=total_count)
```
jychp: Blocking: this mocks past the bug. A `MagicMock` fabricates any attribute, so `result_info.total_count` always resolves and the tests pass regardless of the `AttributeError` crash path flagged in accounts.py:44. Neither `result_info is None` nor missing-`total_count` is covered. Construct a real `V4PagePaginationArrayResultInfo` (or a spec'd mock) so the test reflects the SDK contract it claims to guard.

jychp (changes requested):

cartography/intel/cloudflare/accounts.py
thenadz: I'd defer to your guidance in terms of established practice within this library. From outside perspective, I think I'd much prefer it loudly fail - would hate to think I had a clear view of all accounts and I was looking a small fraction of the organization.

cartography/intel/cloudflare/accounts.py
thenadz: Confirming comment seen - we discuss in later comment with decision to fail loudly given upstream SDK dependency doesn't appear to provide functional pagination.

cartography/intel/cloudflare/accounts.py
thenadz: Good catch - confirmed on `cloudflare 5.4.0`. Fixed in `83bd11df`: `total_count = getattr(page.result_info, "total_count", None)`

tests/unit/cartography/intel/cloudflare/test_cloudflare.py
thenadz: Another good catch. Fixed in `83bd11df`. Tests now build a real `V4PagePaginationArrayResultInfo` via `model_validate` instead of a `MagicMock`, with cases for absent `total_count` and `None` `result_info`

thenadz:
Hi @jychp - Are you able to take a look at responses to your past review? Would love to get a version into mainline that is able to function when provided an account API key. Definitely an imperfect solution given upstream SDK gaps, but this seems to be a substantially better solution than what is released today. Thanks so much!

PR #2948, opened 2026-06-24
Prepare azure-mgmt-storage 25.0.0 bump

jychp:
### Type of change

- [ ] Documentation update

### Summary

Normalize Azure Storage SDK payloads that moved model properties under `properties` in `azure-mgmt-storage` 25.0.0. This keeps Storage Account, Blob Container, File Share, and Data Lake ingestion reading the same Cartography fields after the SDK major bump.

This prepares/supersedes dependabot bump #2933, similar to the compatibility work needed for #2907.

### Related issues or links

- Prepares #2933
- Related #2907

### Breaking changes

None.

### How was this tested?

- `uv run --frozen pytest tests/unit/cartography/intel/azure/test_storage.py`
- `uv run --frozen pytest tests/integration/cartography/intel/azure/test_storage.py tests/integration/cartography/intel/azure/test_data_lake.py`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

No schema changes; this only preserves existing fields when Azure SDK 25 serializes them under `properties`.


cartography/intel/azure/storage.py
```diff
@@ -33,6 +33,91 @@
 logger = logging.getLogger(__name__)
 
 
+def _copy_properties(data: Dict, mapping: Dict[str, tuple[str, ...]]) -> Dict:
```
kunaals, line 36: blocking: this normalization helper should cover tables too. with azure-mgmt-storage 25, Table.as_dict() now puts the table name under properties.tableName, but AzureStorageTableSchema reads table_name. get_tables still returns raw as_dict(), so table nodes keep their ids but lose the tablename property, and the current integration test only checks ids so it misses the regression. can we normalize tables with the same _copy_properties pattern, e.g. table_name from ("table_name", "tableName")?

kunaals (changes requested):
requesting changes for the Azure Storage table shape regression

kunaals (approved):

PR #2958, opened 2026-06-26
feat(rules): Reduce false positives in cloud security rules and add firewall relevancy signal

jychp:
### Type of change
- [ ] Documentation update

### Summary

Reduces false positives across several security rules and adds a relevancy signal to the GCP firewall rules. All changes are confined to rule definitions under `cartography/rules/data/rules/` (Cypher predicates, identity fields, and output models); no node/relationship schema changes.

**Exclude resources that are non-actionable by design**
- Full administrative privileges (CIS AWS 2.15): exclude AWS service-linked roles, `OrganizationAccountAccessRole`, and StackSets execution roles, which hold admin by design. Count query aligned to the same population.
- Roles granting pod creation (CIS K8s 5.1.4): exclude immutable EKS-managed roles (`eks:*` prefix, `aws-node`, `vpc-resource-controller-role`).
- Container image coverage: exclude Kubernetes system namespaces (`kube-system`, `calico-system`, `tigera-operator`) whose images come from vendor registries and are never tracked.
- Containers using hostPorts (CIS K8s 5.2.12): exclude containers whose pod runs with `hostNetwork=true`, where the hostPort has no independent effect (the exposure is reported by the host-network rule).

**Exclude terminated instances (no live attack surface)**
- EC2 IMDSv2 (CIS AWS 6.7): exclude `terminated`/`shutting-down` instances.
- GCP compute public IPs (CIS GCP 4.9): exclude `TERMINATED` instances that retain a stale `public_ip`.
- Internet-exposed compute instances (AWS + GCP facts): exclude terminated instances behind an open management-port rule.

**Sharper predicates**
- S3 MFA Delete (CIS AWS 3.1.2): only evaluate MFA Delete on versioning-enabled buckets. MFA Delete requires versioning, and the API returns no MFA Delete field for non-versioned buckets, so the old predicate flagged every bucket on missing data.
- AI provider API key hygiene (NIST AI RMF): scope to active keys only (exclude inactive/archived Anthropic keys and keys in non-active OpenAI projects); fact, visual, and count queries aligned.

**Stable finding identity**
- Tailscale device key-expiry and device posture findings: key on `(tailnet_id, hostname)` instead of the device id. Tailscale ephemeral nodes get a fresh device id on every reconnect, which re-created the same finding each sync.

**New relevancy signal**
- Policy administration privileges and full administrative privileges (CIS AWS 2.15): add an `is_sso_reserved` output field (AWS IAM Identity Center reserved roles, which ship with admin by design) so consumers can triage break-glass SSO roles separately.
- GCP unrestricted SSH/RDP (CIS GCP 3.6/3.7): add an `in_use` field (true when the firewall's VPC holds a non-terminated instance). The rule still emits when `false` and does not filter on it, keeping it faithful to the control while letting consumers gauge relevancy. Mirrors the existing `in_use` field on the default security group rule.

### Related issues or links

N/A

### How was this tested?

- `make test_lint` (pre-commit: ruff, black, flake8, mypy) passes on all changed files.
- Full rules unit suite passes locally: `uv run pytest tests/unit/rules/` — 1223 passed. Version-pin assertions for the touched rules were updated accordingly.
- Changes are limited to declarative rule data validated by the rules test suite (identity-field/output-model consistency, framework mappings, rule registration).

#### General

#### Proof of functionality

### Notes for reviewers

These are scoped, per-rule changes. Two design choices worth calling out:
- Terminated-instance and managed-role exclusions are correctness fixes (the resource is dead or unmanageable), so they are applied in the rule itself.
- `in_use` is a relevancy signal, not a filter: the firewall rules stay faithful to the CIS control and still emit findings when `in_use=false`, leaving severity/relevancy decisions to the consumer. This matches the prior pattern established by the default security group rule.


cartography/rules/data/rules/cis_aws_storage.py
```diff
@@ -49,16 +49,20 @@ class S3MfaDeleteOutput(Finding):
...
+        "deleting object versions or changing the versioning state."
     ),
     cypher_query="""
     MATCH (a:AWSAccount)-[:RESOURCE]->(bucket:S3Bucket)
-    WHERE bucket.versioning_status IS NULL OR bucket.versioning_status <> 'Enabled'
-       OR bucket.mfa_delete IS NULL OR bucket.mfa_delete <> 'Enabled'
+    // MFA Delete requires versioning, so only evaluate it on versioning-enabled
+    // buckets. A NULL mfa_delete on such a bucket means the API reported no MFA
+    // Delete (i.e. disabled); buckets without versioning are out of scope here.
+    WHERE bucket.versioning_status = 'Enabled'
```
kunaals: blocking: this makes unversioned buckets disappear from the CIS 3.1.2 finding. the rule is still mapped to "Ensure MFA Delete is enabled on S3 buckets", and the previous behavior flagged either missing versioning or missing MFA Delete. with this predicate, a bucket with versioning disabled becomes compliant by omission. i think we should keep the old `versioning_status IS NULL OR <> 'Enabled' OR mfa_delete ...` logic here, or split this narrower check into a separate non-CIS rule.

kunaals (changes requested):
requesting changes for the S3 MFA Delete false negative; the rest looks fine.

kunaals (approved):

PR #2959, opened 2026-06-27
feat(kubernetes): connect to EKS via boto3 auth without a kubeconfig file

thenadz:
### Type of change
- [ ] Documentation update

### Summary

Adds `--k8s-eks-cluster-names` (comma-separated) and optional `--k8s-eks-region` so
the Kubernetes module can connect to Amazon EKS using the ambient AWS credentials,
without a kubeconfig file. This avoids managing a kubeconfig whose
`aws eks get-token` credentials are short-lived, and avoids the kubeconfig exec
plugin, which requires the `aws` CLI in the runtime — useful for scheduled or
containerized syncs.

For each named cluster, Cartography:
1. resolves the API endpoint and cluster CA via `eks:DescribeCluster`,
2. mints a bearer token from the boto3 session's STS client (presigned
   `GetCallerIdentity` with the `x-k8s-aws-id` header — the EKS token scheme), and
3. builds the kubernetes client in memory (no temp files): the cluster CA is loaded
   into the SSL context via `cadata`, and the token is set as the bearer credential.

`--k8s-eks-region` is optional; when omitted, boto3 resolves the region from its
standard chain (`AWS_REGION` / `AWS_DEFAULT_REGION` / AWS config). Passing
`--k8s-eks-cluster-names` implies EKS, so `--managed-kubernetes eks` is not required
to also enrich with Access Entries / OIDC.

EKS API-server CA certificates omit the Authority Key Identifier extension, which
Python 3.13 / urllib3 reject under the default `VERIFY_X509_STRICT` flag. The client
clears only that strict subflag; certificate verification against the cluster CA
stays enabled (`CERT_REQUIRED`). This is not `InsecureSkipVerify`.

`--k8s-kubeconfig` and `--k8s-eks-cluster-names` are alternatives; provide one.

Required permissions (documented in `docs/root/modules/kubernetes/config.md`):
- AWS: `eks:DescribeCluster` plus the existing EKS enrichment actions, and
  `sts:GetCallerIdentity`.
- Kubernetes RBAC bound to the calling IAM principal via an EKS access entry
  (e.g. the AWS-managed `AmazonEKSViewPolicy`).

### How was this tested?

- New unit tests in `tests/unit/cartography/intel/kubernetes/test_eks_auth.py`:
  the bearer-token format and that the `x-k8s-aws-id` header is registered on the
  STS request; the SSL context clears `VERIFY_X509_STRICT` while keeping
  `CERT_REQUIRED`; and the `describe_cluster` → client-build wiring. A region
  resolution test is added to `test_init.py`. The
  `tests/unit/cartography/intel/kubernetes/` suite passes (31 tests).
- `make test_lint` passes locally (black, isort, flake8).
- Verified against a live EKS cluster using `--k8s-eks-cluster-names`. Sample sync
  output:

  ```
  INFO:cartography.intel.kubernetes:Syncing data for k8s cluster benzinga-production...
  INFO:cartography.client.core.tx:Loaded 193 KubernetesServiceAccount nodes
  INFO:cartography.client.core.tx:Loaded 1364 KubernetesPod nodes
  INFO:cartography.client.core.tx:Loaded 1584 KubernetesContainer nodes
  INFO:cartography.client.core.tx:Loaded 365 KubernetesService nodes
  INFO:cartography.client.core.tx:Loaded 140 KubernetesIngress nodes
  INFO:cartography.sync:Finishing sync stage 'kubernetes'
  ```

#### General

#### Proof of functionality

- [x] Included Cartography sync logs from a real environment (above). This PR adds an
  alternative connection/auth method for the existing Kubernetes entities; it does
  not add or change any node or relationship schema, so the schema docs are
  unchanged. Module configuration docs are updated with the new options and required
  permissions.

### DIFF
--- a/cartography/intel/kubernetes/__init__.py
+++ b/cartography/intel/kubernetes/__init__.py
@@ -14,6 +14,7 @@ from cartography.intel.kubernetes.pods import sync_pods
 from cartography.intel.kubernetes.rbac import sync_kubernetes_rbac
 from cartography.intel.kubernetes.secrets import sync_secrets
 from cartography.intel.kubernetes.services import sync_services
+from cartography.intel.kubernetes.util import get_eks_k8s_clients
 from cartography.intel.kubernetes.util import get_k8s_clients
 from cartography.util import run_scoped_analysis_job
 from cartography.util import timeit
@@ -38,13 +39,31 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
         logger.error("Cartography update tag not provided.")
         return
 
-    if not config.k8s_kubeconfig:
-        logger.error("Kubernetes kubeconfig not provided.")
+    if config.k8s_eks_cluster_names:
+        # region_name=None lets boto3 fall back to its default region resolution.
+        boto3_session = boto3.Session(region_name=config.k8s_eks_region)
+        if not boto3_session.region_name:
+            logger.error(
+                "Could not determine an AWS region for EKS sync. Set "
+                "--k8s-eks-region or configure a default AWS region.",
+            )
+            return
+        cluster_names = [
+            c.strip() for c in config.k8s_eks_cluster_names.split(",") if c.strip()
+        ]
+        k8s_clients = get_eks_k8s_clients(cluster_names, boto3_session)
+    elif config.k8s_kubeconfig:
+        k8s_clients = get_k8s_clients(config.k8s_kubeconfig)
+    else:
+        logger.error(
+            "Kubernetes not configured: provide --k8s-kubeconfig or "
+            "--k8s-eks-cluster-names.",
+        )
         return
 
     common_job_parameters = {"UPDATE_TAG": config.update_tag}
 
-    for client in get_k8s_clients(config.k8s_kubeconfig):
+    for client in k8s_clients:
         logger.info(f"Syncing data for k8s cluster {client.name}...")
         try:
             cluster_info = sync_kubernetes_cluster(
@@ -68,7 +87,10 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
 
             # Extract region from cluster ARN (works for EKS; None for non-EKS clusters)
             region: str | None = None
-            if config.managed_kubernetes == "eks":
+            is_eks = config.managed_kubernetes == "eks" or bool(
+                config.k8s_eks_cluster_names
+            )
+            if is_eks:
                 # EKS clusters always have a valid ARN — let ValueError propagate if not
                 region = get_region_from_arn(cluster_external_ref)
                 boto3_session = boto3.Session()
--- a/cartography/intel/kubernetes/util.py
+++ b/cartography/intel/kubernetes/util.py
@@ -1,4 +1,6 @@
+import base64
 import logging
+import ssl
 from datetime import datetime
 from typing import Any
 from typing import Callable
@@ -6,6 +8,7 @@ from typing import Callable
 from dateutil.parser import isoparse
 from kubernetes import config
 from kubernetes.client import ApiClient
+from kubernetes.client import Configuration
 from kubernetes.client import CoreV1Api
 from kubernetes.client import CustomObjectsApi
 from kubernetes.client import NetworkingV1Api
@@ -15,6 +18,9 @@ from kubernetes.client.exceptions import ApiException
 from kubernetes.config.kube_config import KubeConfigMerger
 
 logger = logging.getLogger(__name__)
+
+# EKS bearer-token lifetime is ~15 min; mint just before use.
+_EKS_TOKEN_TTL_SECONDS = 60
 
 
 class KubernetesContextNotFound(Exception):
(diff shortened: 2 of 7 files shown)


thenadz:
@kunaals  - Would love your eyes on this PR when you have a moment. Adds boto3-native EKS auth (presigned STS token, no kubeconfig / no aws CLI), which sits alongside the EKS access-entry and aws-auth work you've been reviewing. No new lib dependencies. Big win for enabling use of modern AWS IAM auth cleanly when using Cartography to review pod structure. Designed to be fully backward compatible with existing auth solutions already supported. cc: @jychp

cartography/intel/kubernetes/util.py
```diff
@@ -16,6 +19,9 @@
 
 logger = logging.getLogger(__name__)
 
+# EKS bearer-token lifetime is ~15 min; mint just before use.
+_EKS_TOKEN_TTL_SECONDS = 60
```
jychp, line 23: **blocking:** `ExpiresIn=_EKS_TOKEN_TTL_SECONDS` (60s) sets `X-Amz-Expires=60` on the presigned GetCallerIdentity URL, but the token is minted once at client build and reused for the cluster's entire sync. The PR's own sample loads thousands of resources, which takes far longer than 60s, so the API starts returning 401 mid-sync. Use 900 (the `aws eks get-token` / EKS max lifetime) — the code comment already says ~15 min.

cartography/intel/kubernetes/util.py
```diff
@@ -131,6 +138,79 @@ def get_k8s_clients(kubeconfig: str) -> list[K8sClient]:
...
+    cfg.host = endpoint
+    cfg.api_key = {"authorization": f"Bearer {token}"}
+    cfg.verify_ssl = True
+
+    api_client = ApiClient(configuration=cfg)
+    api_client.rest_client.pool_manager.connection_pool_kw["ssl_context"] = ssl_context
+    return api_client
+
+
+def get_eks_k8s_clients(
```
jychp, line 191: **blocking:** all clusters' tokens/ApiClients are minted here before the sync loop in `__init__.py` runs, so a later cluster's token can already be expired before its sync begins, and each `ApiClient` carries one static token for its whole sync. Mint the token lazily per cluster immediately before syncing, or wire `Configuration.refresh_api_key_hook` to re-mint on demand.

cartography/intel/kubernetes/util.py
```diff
@@ -131,6 +138,79 @@ def get_k8s_clients(kubeconfig: str) -> list[K8sClient]:
...
+    cluster_names: list[str], boto3_session: Any
+) -> list[K8sClient]:
+    """Build K8sClients for EKS clusters using boto3 auth instead of a kubeconfig file."""
+    eks = boto3_session.client("eks")
+    clients = []
+    for cluster_name in cluster_names:
+        described = eks.describe_cluster(name=cluster_name)["cluster"]
```
jychp, line 198: **suggestion:** no error isolation — one invalid cluster name makes `describe_cluster` raise and aborts the entire sync before any cluster is processed. The kubeconfig path syncs each cluster in a try/except; wrap the per-cluster build (or move it into the sync loop) so one bad cluster doesn't kill the rest.

cartography/intel/kubernetes/util.py
```diff
@@ -131,6 +138,79 @@ def get_k8s_clients(kubeconfig: str) -> list[K8sClient]:
     return clients
 
 
+def _get_eks_token(cluster_name: str, boto3_session: Any) -> str:
```
jychp, line 141: **nitpick:** type `boto3_session` as `boto3.session.Session` instead of `Any`, consistent with the rest of the module.

cartography/intel/kubernetes/__init__.py
```diff
@@ -38,13 +39,31 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
         logger.error("Cartography update tag not provided.")
         return
 
-    if not config.k8s_kubeconfig:
-        logger.error("Kubernetes kubeconfig not provided.")
+    if config.k8s_eks_cluster_names:
```
jychp, line 42: **suggestion:** when both `--k8s-kubeconfig` and `--k8s-eks-cluster-names` are set, EKS silently wins and the kubeconfig is ignored. The docs say to provide one — error out (or warn) when both are supplied rather than silently picking.

cartography/intel/kubernetes/__init__.py
```diff
@@ -68,7 +87,10 @@ def start_k8s_ingestion(session: Session, config: Config) -> None:
...
             # Extract region from cluster ARN (works for EKS; None for non-EKS clusters)
             region: str | None = None
-            if config.managed_kubernetes == "eks":
+            is_eks = config.managed_kubernetes == "eks" or bool(
+                config.k8s_eks_cluster_names
+            )
+            if is_eks:
                 # EKS clusters always have a valid ARN — let ValueError propagate if not
                 region = get_region_from_arn(cluster_external_ref)
                 boto3_session = boto3.Session()
```
jychp, line 96: **question:** `--k8s-eks-region` configures the session used to build clients, but this enrichment session is created region-less (region is passed to `sync_eks` separately). Confirm `sync_eks` always uses the passed `region` for client creation; otherwise `--k8s-eks-region` won't govern the EKS enrichment calls.

jychp (changes requested):

PR #2960, opened 2026-06-29
feat(gcp): ingest GCP API Keys (apikeys.googleapis.com)

jychp:
### Type of change
- [ ] Documentation update

### Summary

GCP's actual API Keys product (`apikeys.googleapis.com`, `apikeys.Key`) was not modeled in Cartography at all, so real API Keys were missing from the graph. The only GCP nodes carrying the generic `APIKey` ontology label were `GCPServiceAccountKey` (user-managed service-account keys), which conflated two distinct resource types.

This adds a net-new intel module and data model for GCP API Keys, ingested as a dedicated `GCPApiKey` node type, distinct from `GCPServiceAccountKey`:

- New `intel/gcp/apikeys.py` (GET / TRANSFORM / LOAD / CLEANUP), listing keys under `projects/{id}/locations/global` via the `apikeys` v2 API, with the standard GCP HTTP-error handling (permission denied / API disabled / billing disabled).
- New `models/gcp/apikeys.py`: `GCPApiKeySchema` with a `RESOURCE` edge from `GCPProject` and the `APIKey` extra label for ontology parity. The secret key material (`keyString`) is never ingested.
- Ontology mapping: added a `GCPApiKey` entry to the API Keys ontology mapping so `_ont_*` fields are populated (`_ont_name` from `display_name`, falling back to the resource name; `_ont_created_at`/`_ont_updated_at` from the key timestamps).
- Wired into the per-project GCP sync, the `Services` map, and the selective-sync `RESOURCE_FUNCTIONS` (new `apikeys` key).
- Timestamps stored as native datetimes.

### How was this tested?

- New integration test `tests/integration/cartography/intel/gcp/test_apikeys.py` asserts node creation, the `GCPProject`-[:RESOURCE]->`GCPApiKey` edge, the `APIKey` ontology label, and that `_ont_source` / `_ont_name` are populated.
- Existing GCP integration tests, selective-sync tests, and ontology mapping unit tests pass.

#### General

#### Proof of functionality

### Notes for reviewers

- `GCPApiKey` deliberately carries the `APIKey` ontology label alongside `GCPServiceAccountKey`; the two remain separate node labels. Consumers that previously counted `APIKey` nodes as "service-account keys" should narrow on the specific label.
- `display_name` is optional upstream, so the ontology `name` mapping coalesces to the resource name to keep `_ont_name` populated.


cartography/intel/gcp/apikeys.py
```diff
@@ -0,0 +1,154 @@
...
+    transformed = []
+    for key in keys:
+        restrictions = key.get("restrictions")
+        transformed.append(
+            {
+                "id": key["name"],
+                "uid": key.get("uid"),
+                "name": key["name"],
+                "displayName": key.get("displayName"),
+                "createTime": _parse_dt(key.get("createTime")),
```
kunaals, line 95: non-blocking: this stores API key timestamps as native datetimes, while the other APIKey ontology sources seem to store raw timestamp strings. that means shared fields like _ont_created_at may be mixed-type across providers. i'd either keep the RFC3339 strings here for consistency, or handle timestamp normalization across the ontology layer in a follow-up.

tests/integration/cartography/intel/gcp/test_apikeys.py
```diff
@@ -0,0 +1,95 @@
...
+        MERGE (project:GCPProject{id: $project_id})
+        ON CREATE SET project.firstseen = timestamp()
+        SET project.lastupdated = $update_tag
+        """,
+        project_id=TEST_PROJECT_ID,
+        update_tag=TEST_UPDATE_TAG,
+    )
+
+
+@patch.object(
```
kunaals, line 25: non-blocking: this test covers load/relationships well, but patching get_api_keys means the pagination and handled HttpError paths are still untested. a small unit test around get_api_keys plus transform_api_keys would make the risky parts of the new module easier to maintain.

kunaals (approved):
approving to unblock, left some nits

PR #2962, opened 2026-06-29
feat(aws-ec2): add Transit Gateway route table ingestion

wheel5up:
## Summary

Extends the AWS EC2 Transit Gateway module to ingest **TGW route tables and their routes, associations, and propagations**. Previously cartography modeled TGW topology (transit gateways and attachments) but not the routing layer, so you could see *which* VPCs/accounts attach to a TGW but not *how traffic routes* between them. This closes that gap.

## New nodes

- `AWSTransitGatewayRouteTable`
- `AWSTransitGatewayRoute`
- `AWSTransitGatewayRouteTableAssociation`
- `AWSTransitGatewayRouteTablePropagation`

## New relationships

Naming follows `docs/root/dev/writing-intel-modules.md` (active verbs, parent→child, no passive `*_BY` forms):

```
(AWSTransitGateway)-[:CONTAINS]->(AWSTransitGatewayRouteTable)
(AWSTransitGatewayRouteTable)-[:HAS_ROUTE]->(AWSTransitGatewayRoute)
(AWSTransitGatewayRoute)-[:ROUTES_TO_TGW]->(AWSTransitGateway)
(AWSTransitGatewayRoute)-[:ROUTES_TO_TGW_ATTACHMENT]->(AWSTransitGatewayAttachment)
(AWSTransitGatewayRouteTableAssociation)-[:ASSOCIATED_WITH]->(AWSTransitGatewayRouteTable)
(AWSTransitGatewayRouteTable)-[:PROPAGATES]->(AWSTransitGatewayRouteTablePropagation)
(AWSAccount)-[:RESOURCE]->(...)   # account ownership for all new nodes
```

## Why this is useful

Enables queries like "which CIDRs route to which attachment/VPC" and "are these two accounts actually routable through the shared TGW" — not just "are they attached." TGW attachment ≠ connectivity; routing is governed by which route table an attachment is associated with and what routes it contains.

```cypher
-- Routes through a TGW and their targets
MATCH (tgw:AWSTransitGateway)-[:CONTAINS]->(rt:AWSTransitGatewayRouteTable)
      -[:HAS_ROUTE]->(r:AWSTransitGatewayRoute)-[:ROUTES_TO_TGW_ATTACHMENT]->(att:AWSTransitGatewayAttachment)
RETURN tgw.tgw_id, r.destination_cidr_block, r.state, att.id
```

## Implementation notes

- Registered as the `ec2:tgw_route_table` sync in `RESOURCE_FUNCTIONS`.
- Uses `search_transit_gateway_routes` as a fallback when `describe_transit_gateway_route_tables` omits `Routes`.
- Uses `get_*`/`describe_*` fallback for associations/propagations to work across botocore versions (the `get_*` API names are newer).
- The association/propagation APIs don't echo the parent route table id, so it's injected and a stable id (`route_table_id|attachment_id`) is synthesized for dedup + relationship matching.
- **No new IAM required** — the `SecurityAudit` managed policy already grants the `ec2:Describe*`/`Get*`/`Search*` calls.

## Testing

- Unit tests for transform, load, cleanup, and the `search_transit_gateway_routes` fallback.
- Validated end-to-end against a live multi-account org (transit gateways spanning 26 accounts): nodes and all relationships load with correct direction, counts match, cleanup works.


wheel5up:
Good catch — this is a real first-run ordering bug. Routes were loaded before route tables, so on an empty/first-time graph the `(RouteTable)-[:HAS_ROUTE]->(Route)` matcher had no parent to attach to and no `HAS_ROUTE` edges were created (a warm graph masked it because the tables already existed from a prior sync). Fixed in 12eff1c5 by loading route tables before routes, plus a regression test asserting the load order.

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,448 @@
...
+                "TransitGatewayRouteTableId": route_table_id,
+                "Filters": [{"Name": "state", "Values": state_values}],
+            }
+            if next_token:
+                params["NextToken"] = next_token
+            resp = client.search_transit_gateway_routes(**params)
+            routes.extend(resp.get("Routes", []))
+            next_token = resp.get("NextToken")
+            if not next_token:
+                break
```
achantavy: Rather than do this, would it make more sense to use the built-in boto3 paginator? Like this: https://github.com/cartography-cncf/cartography/blob/58ce395e136744bc2747143aa30fd280ffe926f8/cartography/intel/aws/ec2/instances.py#L104-L107

Or is that not feasible because we need these filters?

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,448 @@
...
+        Region=region,
+        AWS_ID=current_aws_account_id,
+    )
+
+
+@timeit
+def cleanup_transit_gateway_route_tables(
+    neo4j_session: neo4j.Session, common_job_parameters: dict[str, Any]
+) -> None:
+    logger.debug("Running TGW route tables cleanup")
```
achantavy, line 174: Order matters. I would clean up the leaf nodes first and then the others. E.g., clean up the routes first, and then the route tables.

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,448 @@
...
+            e.response["Error"]["Message"],
+        )
+    return route_tables
+
+
+def get_transit_gateway_routes_for_table(
+    boto3_session: boto3.session.Session, region: str, route_table_id: str
+) -> list[dict[str, Any]]:
+    """Search Transit Gateway routes for a specific route table as a fallback when
+    describe_transit_gateway_route_tables does not include Routes in the response.
```
achantavy: Weird, when does this happen?

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,495 @@
...
+@timeit
+@aws_handle_regions
+def get_transit_gateway_route_tables(
+    boto3_session: boto3.session.Session, region: str
+) -> list[dict[str, Any]]:
+    client = create_boto3_client(
+        boto3_session, "ec2", region_name=region, config=get_botocore_config()
+    )
+    route_tables: list[dict[str, Any]] = []
+    try:
```
achantavy: The try+except is not necessary here. Pretty sure aws_handle_regions handles this already: https://github.com/cartography-cncf/cartography/blob/e8b75e84db061ba5cd7b8fa9296315ed532d2d13/cartography/util.py#L719

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,495 @@
...
+    boto3_session: boto3.session.Session, region: str, route_table_id: str
+) -> list[dict[str, Any]]:
+    """Search Transit Gateway routes for a specific route table as a fallback when
+    describe_transit_gateway_route_tables does not include Routes in the response.
+    """
+    client = create_boto3_client(
+        boto3_session, "ec2", region_name=region, config=get_botocore_config()
+    )
+    routes: list[dict[str, Any]] = []
+    try:
```
achantavy: Remove the try+except here

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,495 @@
...
+    except botocore.exceptions.ClientError as e:
+        logger.warning(
+            "Could not retrieve Transit Gateway Route Tables due to boto3 error %s: %s. Skipping.",
+            e.response["Error"]["Code"],
+            e.response["Error"]["Message"],
+        )
+    return route_tables
+
+
+def get_transit_gateway_routes_for_table(
```
achantavy, line 44: Decorate with @aws_handle_regions and @timeit

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,495 @@
...
+    rather than re-enumerating, avoiding per-region cartesian duplication.
+    """
+    client = create_boto3_client(
+        boto3_session, "ec2", region_name=region, config=get_botocore_config()
+    )
+    props: list[dict[str, Any]] = []
+    try:
+        # Prefer get_ API names if available (newer service models use get_* instead of describe_*)
+        api_name = None
+        if hasattr(client, "get_transit_gateway_route_table_propagations"):
```
achantavy: We should just pick a single API and use it instead of trying to grab it dynamically and handle the returns in the different situations. This avoids complexity.

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,495 @@
...
+    """Fetch propagations for the given route tables in this region.
+
+    Iterates only the route tables passed in (already fetched for this region)
+    rather than re-enumerating, avoiding per-region cartesian duplication.
+    """
+    client = create_boto3_client(
+        boto3_session, "ec2", region_name=region, config=get_botocore_config()
+    )
+    props: list[dict[str, Any]] = []
+    try:
```
achantavy: Remove try + except

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,495 @@
...
+    except botocore.exceptions.ClientError as e:
+        logger.warning(
+            "Could not retrieve Transit Gateway route table associations due to boto3 error %s: %s. Skipping.",
+            e.response.get("Error", {}).get("Code"),
+            e.response.get("Error", {}).get("Message"),
+        )
+    return associations
+
+
+def get_transit_gateway_route_table_propagations(
```
achantavy, line 302: Decorate with aws_handle_regions and timeit

cartography/models/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+@dataclass(frozen=True)
+class AWSTransitGatewayRouteToRouteTableRel(CartographyRelSchema):
+    target_node_label: str = "AWSTransitGatewayRouteTable"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"id": PropertyRef("transit_gateway_route_table_id")},
+    )
+    # INWARD so the edge reads (RouteTable)-[:HAS_ROUTE]->(Route): active verb,
+    # parent->child, replacing the passive/unclear ROUTE_OF.
```
achantavy: This looks like the output of a coding agent thinking to itself haha. ROUTE_OF isn't mentioned anywhere else -- let's remove this comment

tests/unit/cartography/intel/aws/ec2/test_tgw_route_tables_search_fallback.py
```diff
@@ -0,0 +1,90 @@
+import boto3
```
achantavy: Delete this file altogether -- unit test is not necessary. We only need integration tests to test the outcomes. Here we don't need  to make sure functions are called.

tests/unit/cartography/intel/aws/ec2/test_tgw_route_tables_load_cleanup.py
```diff
@@ -0,0 +1,133 @@
+from cartography.intel.aws.ec2 import tgw_route_tables
```
achantavy: Delete this file -- no need unit tests. Keep the integ tests

achantavy (changes requested):
Thanks for the PR! Adding tgw route coverage is super cool. Quick first pass:

- Please update the schema docs
- For the get_* functions:
  - Pick a specific API instead of picking describe_* vs get_* at runtime based on the attr available from the library; this adds a ton of complexity. By making it simple at the get, it'll become simpler at the transform.
  - Remove try+excepts bc it's mostly handled by the aws_handle_regions decorator already
  - I think we can use the standard boto3 paginator instead of handling next token ourselves. I left an example in a comment.
- Remove unit tests, keep the integ tests.

Lmk if you have any questions!

g2wheeler:
Thanks for the thorough review, @achantavy! I've pushed 5 new commits addressing every item from your pass and the cubic bot's earlier one. Responding to each point below:

---

**"For the get_* functions: Pick a specific API instead of picking describe_* vs get_* at runtime based on the attr available from the library; this adds a ton of complexity."**
→ Fixed in [`06e1d0ac`](https://github.com/cartography-cncf/cartography/commit/06e1d0ac). Both association/propagation fetchers now call `get_transit_gateway_route_table_associations`/`get_transit_gateway_route_table_propagations` directly, no `hasattr()` branching. I verified against the live botocore 1.40.x service model and the AWS CLI command list that `describe_transit_gateway_route_table_associations`/`propagations` never existed as real operations under any SDK version — the `describe_*` fallback was dead code from the start. The transform got simpler as a result: no more dual-shape handling for `State` vs `AssociationState.State`.

**"Rather than do this, would it make more sense to use the built-in boto3 paginator? ... Or is that not feasible because we need these filters?"** (on the search-routes fallback)
→ Confirmed `search_transit_gateway_routes` has no paginator support at all in botocore (`Operation cannot be paginated`), so the manual `NextToken` loop stays there. However, I *did* convert the associations/propagations fetchers to use `client.get_paginator(...)` in `06e1d0ac`, since those two operations do have real paginators.

**"Order matters. I would clean up the leaf nodes first and then the others."**
→ Fixed in [`158b9dec`](https://github.com/cartography-cncf/cartography/commit/158b9dec). `cleanup_transit_gateway_route_tables` now deletes Route/Association/Propagation before the parent RouteTable.

**"Weird, when does this happen?"** (on the `if not rtb.get(Routes)` fallback guard)
→ Fixed in [`06e1d0ac`](https://github.com/cartography-cncf/cartography/commit/06e1d0ac). Good catch — I checked the botocore output shape for `DescribeTransitGatewayRouteTables` and it never includes a `Routes` field at all, so that guard was always true. Removed the misleading conditional; routes are now fetched unconditionally per route table via the search API.

**"The try+except is not necessary here. Pretty sure aws_handle_regions handles this already"** / **"Remove the try+except here"** (x2 more instances) / **"Decorate with @aws_handle_regions and @timeit"** (x2)
→ All fixed in [`06e1d0ac`](https://github.com/cartography-cncf/cartography/commit/06e1d0ac). Removed every redundant `try/except botocore.exceptions.ClientError` block and added the missing `@aws_handle_regions`/`@timeit` decorators to `get_transit_gateway_routes_for_table`, `get_transit_gateway_route_table_associations`, and `get_transit_gateway_route_table_propagations`.

**"This looks like the output of a coding agent thinking to itself haha. ROUTE_OF isn't mentioned anywhere else -- let's remove this comment"**
→ Fixed in [`aabe5457`](https://github.com/cartography-cncf/cartography/commit/aabe5457).

**"Delete this file altogether -- unit test is not necessary. We only need integration tests to test the outcomes."** (both `test_tgw_route_tables_load_cleanup.py` and `test_tgw_route_tables_search_fallback.py`)
→ Fixed in [`d78c379c`](https://github.com/cartography-cncf/cartography/commit/d78c379c). Both files deleted, along with the now-dead `test_transform_associations_reads_nested_association_state` unit test (its code path no longer exists after the `describe_*` removal). Added `tests/integration/cartography/intel/aws/ec2/test_tgw_route_tables.py` + fixture data in its place — it mocks only the boto3 client boundary (not the `get_*` wrapper functions), so the id-injection/synthesis logic is actually exercised, and covers node creation, all 6 new relationship types, AWSAccount ownership, and a stale-data cleanup pass across two update tags.

**"Please update the schema docs"**
→ Fixed in [`e294fb66`](https://github.com/cartography-cncf/cartography/commit/e294fb66). Added `AWSTransitGatewayRouteTable`/`AWSTransitGatewayRoute`/`AWSTransitGatewayRouteTableAssociation`/`AWSTransitGatewayRouteTablePropagation` sections to `docs/root/modules/aws/schema.md`, cross-checked field enums against the live botocore service model.

---

**One pre-existing issue I found but did *not* fix, flagging for visibility:** `AWSTransitGatewayRoute.origin` reads `route.get("Origin")`, but the real `SearchTransitGatewayRoutes` API returns this data under the key `Type` (`static`/`propagated`), not `Origin` — so `origin` is always null in production today. I documented this honestly in the schema docs rather than silently "fixing" unreviewed logic, but happy to patch it in a follow-up commit if you'd like it addressed now vs. as a separate PR.

All 5 commits pass `pre-commit run --all-files` (isort/black/flake8/pyupgrade/mypy), the full unit suite (311 passed), and the full `tests/integration/.../aws/ec2/` suite (63 passed, including the 2 new integration tests).

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Confirmed search_transit_gateway_routes has no paginator support at all in botocore (Operation cannot be paginated), so the manual NextToken loop stays here — not feasible to swap in a paginator for this one. I did convert get_transit_gateway_route_table_associations and get_transit_gateway_route_table_propagations to use client.get_paginator(...) in 06e1d0ac, since those two do have real paginators.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in 158b9dec — cleanup now deletes Route/Association/Propagation before the parent RouteTable.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Good catch — checked the botocore output shape for DescribeTransitGatewayRouteTables and it never includes a Routes field at all, so this fallback guard was always true. Removed the misleading conditional in 06e1d0ac; routes are now fetched unconditionally per route table via the search API.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in 06e1d0ac — removed.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in 06e1d0ac — removed.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in 06e1d0ac — both decorators added.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in 06e1d0ac — now calls get_transit_gateway_route_table_propagations directly via client.get_paginator(...), no hasattr() branching. Confirmed the describe_* variant never existed as a real operation on any botocore version, so the fallback was dead code.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in 06e1d0ac — removed.

cartography/intel/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in 06e1d0ac — both decorators added.

cartography/models/aws/ec2/tgw_route_tables.py
wheel5up: Fixed in aabe5457 — comment removed.

tests/unit/cartography/intel/aws/ec2/test_tgw_route_tables_search_fallback.py
wheel5up: Fixed in d78c379c — file deleted; coverage now lives in the new tests/integration/cartography/intel/aws/ec2/test_tgw_route_tables.py.

tests/unit/cartography/intel/aws/ec2/test_tgw_route_tables_load_cleanup.py
wheel5up: Fixed in d78c379c — file deleted; same integration test above covers the cleanup path across two update tags.

cartography/intel/aws/ec2/tgw_route_tables.py
```diff
@@ -0,0 +1,422 @@
...
+    """Search Transit Gateway routes for a specific route table.
+
+    describe_transit_gateway_route_tables does not include Routes in its
+    response, so this is called for every route table to fetch its routes.
+    """
+    client = create_boto3_client(
+        boto3_session, "ec2", region_name=region, config=get_botocore_config()
+    )
+    routes: list[dict[str, Any]] = []
+    # search_transit_gateway_routes may not have a paginator in all botocore versions.
```
achantavy, line 56: I would remove this comment. It can mislead the reader to believe that we support multiple botocore versions, which is a non-goal, haha.

achantavy (approved):
Thanks for the updates, I re-reviewed and tested this e2e and it works, thank you!

cartography/intel/aws/ec2/tgw_route_tables.py
achantavy: Edit: actually this isn't a blocker, this is fine

PR #2963, opened 2026-06-29
feat(scaleway): add object storage buckets

jychp:
### Type of change

### Summary

Adds a Scaleway **Object Storage** intel module. Scaleway Object Storage is S3-compatible and is not exposed by the Scaleway Python SDK, so buckets are collected with `boto3` against the regional S3 endpoints (`fr-par`, `nl-ams`, `pl-waw`).

Each bucket becomes a `ScalewayObjectStorageBucket` node linked to its `ScalewayProject` (the project is derived from the S3 owner id, which Scaleway reports as `<project_id>:<project_id>`). The model captures the public-exposure signals that matter for findings, mirroring the AWS S3 and GCP bucket models:

- `acl_public` — derived from the bucket ACL (`AllUsers` / `AuthenticatedUsers` grants)
- `anonymous_access` / `anonymous_actions` — derived from the bucket policy via `policyuniverse` (same approach as the AWS S3 module)
- plus `region`, `endpoint`, `creation_date`, `tags`, `owner_id`, `versioning_status`

Cross-provider wiring:
- **Ontology**: the node carries the `ObjectStorage` semantic label and is registered in `OBJECT_STORAGE_ONTOLOGY_MAPPING`, so it populates `_ont_name`, `_ont_location`, `_ont_encrypted` (Scaleway encrypts at rest by default → static `true`, as GCP does), `_ont_versioning`, and `_ont_public` (`anonymous_access OR acl_public`, so ACL-only public buckets are not missed).
- **Rules**: a Scaleway fact is added to the `object_storage_public` rule so a public Scaleway bucket is surfaced by the "Public Object Storage Attack Surface" detection alongside AWS / Azure / GCP.

Note on the Scaleway API: a bucket policy can lock the owner out of the admin endpoints (`get_bucket_acl` / `get_bucket_versioning` / ... return `AccessDenied`), and `get_bucket_policy_status` is unreliable, so each per-bucket detail call is exception-tolerant and public exposure is derived from the ACL + policy document directly.

### How was this tested?

- `make test_lint` passes locally (isort / black / flake8 / mypy).
- New integration test `test_load_scaleway_object_storage_buckets` (public + private bucket, asserts the node, the `ObjectStorage` `_ont_*` fields, and the `RESOURCE` edge to the project). Existing scaleway storage tests still pass.
- Rules unit suite passes; the new rule fact's Cypher (query / visual / count) parses and runs against Neo4j.
- End-to-end against a real Scaleway account: created one public (anonymous bucket policy) and one private bucket, ran the real `get()` + `transform()` across all three regions → the public bucket was flagged (`anonymous_access=true`), the private one clean, project correctly resolved. Test buckets were deleted afterward.

Sample sync log:

```
INFO:cartography.intel.scaleway.storage.objectstorage:Loading 2 Scaleway ObjectStorageBuckets in project '0681c477-...' into Neo4j.
INFO:cartography.client.core.tx:Loaded 2 ScalewayObjectStorageBucket nodes
```

#### General

#### Proof of functionality

- [x] Included Cartography sync logs demonstrating successful synchronization of the new entity.

### Notes for reviewers

This is the Object Storage step of the broader Scaleway coverage effort. Public exposure is intentionally folded into bucket properties (like the GCP model) rather than modeled as separate ACL-grant / policy-statement nodes (the AWS approach); that finer granularity can be added later if needed.

### DIFF
--- a/cartography/intel/scaleway/storage/objectstorage.py
+++ b/cartography/intel/scaleway/storage/objectstorage.py
@@ -46,34 +46,48 @@ def sync(
     projects_id: list[str],
     update_tag: int,
 ) -> None:
-    buckets = get(client)
+    buckets = get(client, projects_id)
     buckets_by_project = transform_buckets(buckets)
     load_buckets(neo4j_session, buckets_by_project, update_tag)
     cleanup(neo4j_session, projects_id, common_job_parameters)
 
 
 @timeit
-def get(client: scaleway.Client) -> list[dict[str, Any]]:
+def get(client: scaleway.Client, projects_id: list[str]) -> list[dict[str, Any]]:
     """Scaleway Object Storage is S3-compatible and is not exposed by the
-    Scaleway Python SDK, so we use boto3 against the regional S3 endpoints."""
+    Scaleway Python SDK, so we use boto3 against the regional S3 endpoints.
+
+    S3 calls are scoped to the access key's preferred Project unless the key is
+    suffixed with ``@<project_id>``, so we iterate every project explicitly to
+    cover multi-project orgs (and to keep ingestion aligned with cleanup)."""
     buckets: list[dict[str, Any]] = []
-    for region in OBJECT_STORAGE_REGIONS:
-        s3 = boto3.client(
-            "s3",
-            region_name=region,
-            endpoint_url=f"https://s3.{region}.scw.cloud",
-            aws_access_key_id=client.access_key,
-            aws_secret_access_key=client.secret_key,
-        )
-        listing = s3.list_buckets()
-        owner_id = listing.get("Owner", {}).get("ID", "")
-        for bucket in listing.get("Buckets", []):
-            buckets.append(_get_bucket_details(s3, bucket, region, owner_id))
+    for project_id in projects_id:
+        for region in OBJECT_STORAGE_REGIONS:
+            s3 = boto3.client(
+                "s3",
+                region_name=region,
+                endpoint_url=f"https://s3.{region}.scw.cloud",
+                aws_access_key_id=f"{client.access_key}@{project_id}",
+                aws_secret_access_key=client.secret_key,
+            )
+            try:
+                listing = s3.list_buckets()
+            except ClientError as e:
+                # Don't let one region/project abort the whole Scaleway sync.
+                logger.warning(
+                    "Could not list Scaleway buckets in project '%s' region '%s': %s",
+                    project_id,
+                    region,
+                    e.response.get("Error", {}).get("Code"),
+                )
+                continue
+            for bucket in listing.get("Buckets", []):
+                buckets.append(_get_bucket_details(s3, bucket, region, project_id))
     return buckets
 
 
 def _get_bucket_details(
-    s3: Any, bucket: dict[str, Any], region: str, owner_id: str
+    s3: Any, bucket: dict[str, Any], region: str, project_id: str
 ) -> dict[str, Any]:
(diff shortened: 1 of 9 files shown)


cartography/intel/scaleway/storage/objectstorage.py
```diff
@@ -0,0 +1,216 @@
...
+def get(client: scaleway.Client) -> list[dict[str, Any]]:
+    """Scaleway Object Storage is S3-compatible and is not exposed by the
+    Scaleway Python SDK, so we use boto3 against the regional S3 endpoints."""
+    buckets: list[dict[str, Any]] = []
+    for region in OBJECT_STORAGE_REGIONS:
+        s3 = boto3.client(
+            "s3",
+            region_name=region,
+            endpoint_url=f"https://s3.{region}.scw.cloud",
+            aws_access_key_id=client.access_key,
```
kunaals: Scaleway scopes Object Storage S3 calls to the API key's preferred Project unless the access key is suffixed with `@<project_id>`. Since this sync receives `projects_id`, we should loop projects and use `f"{client.access_key}@{project_id}"`; otherwise multi-project orgs only ingest the preferred project's buckets while cleanup runs for every project.

cartography/intel/scaleway/storage/objectstorage.py
```diff
@@ -0,0 +1,216 @@
...
+            aws_secret_access_key=client.secret_key,
+        )
+        listing = s3.list_buckets()
```
kunaals: Worth wrapping this listing call like the detail calls below. A single region/project S3 error currently aborts the whole Scaleway sync path; logging and continuing would preserve the other regions/projects.

cartography/models/ontology/mapping/data/object_storage.py
```diff
@@ -100,8 +100,50 @@
...
+                    special_handling="equal_boolean",
+                    extra={"values": ["Enabled"]},
+                ),
+                # A bucket is public if its policy grants anonymous access OR its ACL
+                # grants AllUsers/AuthenticatedUsers (the model stores these signals
+                # separately, so ACL-only public buckets must not be missed).
+                OntologyFieldMapping(
+                    ontology_field="public",
+                    node_field="anonymous_access",
+                    special_handling="or_boolean",
```
kunaals: This collapses unknown exposure to false because `or_boolean` coalesces nulls. Since `FETCH_FAILED` intentionally means unknown, consider computing a tri-state `public` field in the transform and mapping it directly.

docs/root/modules/scaleway/schema.md
```diff
@@ -530,3 +532,30 @@ A Security Group Rule is a single firewall rule (inbound or outbound) belonging
...
+### ScalewayObjectStorageBucket
+
+An Object Storage bucket is an S3-compatible container for objects. Scaleway Object Storage is not exposed by the Scaleway Python SDK, so it is collected through the regional S3-compatible endpoints.
+
+> **Ontology Mapping**: This node has the extra label `ObjectStorage` to enable cross-platform queries for object storage across different systems (e.g., S3Bucket, GCPBucket).
+
+| Field      | Description                                  |
+|------------|----------------------------------------------|
+| id         | Bucket name (globally unique).               |
+| name       | Bucket name.                                 |
```
kunaals, line 545: Small docs mismatch: `OBJECT_STORAGE_REGIONS` also includes `it-mil`.

kunaals (approved):
approving to unblock, few nits

PR #2965, opened 2026-06-30
feat(scaleway): add VPC/IPAM network layer and Load Balancers

jychp:
### Type of change
- [x] Documentation update

### Summary
Extends the Scaleway module with the networking and load-balancing layers, which were missing entirely while AWS / GCP / Azure all cover them by default.

**Network (VPC + IPAM)**
- `ScalewayVpc` (extra label `VirtualNetwork`)
- `ScalewayPrivateNetwork`, linked to its VPC
- `ScalewaySubnet` (extra label `Subnet`), linked to its private network
- `ScalewayIP` (IPAM allocations), linked to its subnet

Subnets are read from the embedded private-network payload (no extra call). IPs attach to their subnet via `source.subnet_id` (the IPAM list API does not populate `source.private_network_id`); the private network is reachable transitively.

**Load Balancer**
- `ScalewayLoadBalancer` (extra label `LoadBalancer`), with its public IP(s) flattened onto the node
- `ScalewayLBFrontend` (inbound listener), linked to its LB and the backend it routes to (`ROUTES_TO`)
- `ScalewayLBBackend` (server pool + health check), linked to its LB

**Ontology**: `ScalewayVpc`, `ScalewaySubnet` and `ScalewayLoadBalancer` are wired into the `VirtualNetwork` / `Subnet` / `LoadBalancer` ontology mappings for cross-cloud queries, mirroring the AWS / GCP / Azure entries.

**Regions refactor**: consolidates per-region listing behind a single `DEFAULT_REGIONS` and a `list_all_regions` helper that fans a region-scoped SDK fetcher across all regions, skipping regions where the service is not deployed (Load Balancer is absent from `it-mil`). VPC, IPAM and Object Storage now share this path; `OBJECT_STORAGE_REGIONS` is removed.

### How was this tested?
- New integration tests: `tests/integration/cartography/intel/scaleway/test_network.py` and `test_loadbalancers.py`.
- Full local suite green: `make test_lint`, plus the Scaleway and ontology integration/unit tests (57 passing).
- End-to-end against a live Scaleway project: created a VPC + private network (with subnets) + IPAM IP, and a load balancer + backend + frontend via the `scw` CLI, ran the real `get`/`transform`/`load` path into Neo4j, then deleted the resources. Observed sync output:
  ```
  Loading 1 Scaleway VPCs in project '<project>' into Neo4j.
  Loading 1 Scaleway PrivateNetworks in project '<project>' into Neo4j.
  Loading 1 Scaleway IPs in project '<project>' into Neo4j.
  Loading 1 Scaleway LoadBalancers in project '<project>' into Neo4j.
  ```
  Verified the full `Project -> Vpc -> PrivateNetwork -> Subnet -> IP` and `Project -> LoadBalancer -> Frontend -> Backend` chains, the flattened LB public IP, and the populated `_ont_*` fields on the VPC / subnet / load balancer nodes.

#### General

#### Proof of functionality

- [x] Updated the schema documentation (`docs/root/modules/scaleway/schema.md`).

### Notes for reviewers
Two commits, one per layer (network, then load balancer). The load-balancer commit also carries the regions consolidation. The `ScalewayIP -> resource` (e.g. instance) edge is intentionally deferred: `resource.id` points at the attached object (an instance private NIC, not the server), so the match target varies by `resource.type_`; left as a `TODO` until those resources are modeled.

### DIFF
--- a/cartography/intel/scaleway/loadbalancers/loadbalancers.py
+++ b/cartography/intel/scaleway/loadbalancers/loadbalancers.py
@@ -77,6 +77,11 @@ def transform_loadbalancers(
     frontends_by_project: dict[str, list[dict[str, Any]]] = {}
     backends_by_project: dict[str, list[dict[str, Any]]] = {}
 
+    # Frontends / backends inherit the project of their parent load balancer.
+    # Resolve it from the parent LBs rather than the embedded child `lb`, so a
+    # partial child payload can't strand children under PROJECT_ID=None.
+    project_by_lb_id = {lb.id: lb.project_id for lb in lbs}
+
     for lb in lbs:
         formatted_lb = scaleway_obj_to_dict(lb)
         ip_addresses = [ip["ip_address"] for ip in (formatted_lb.get("ip") or [])]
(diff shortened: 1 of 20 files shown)


cartography/intel/scaleway/loadbalancers/loadbalancers.py
```diff
@@ -0,0 +1,168 @@
...
+
+    for frontend in frontends:
+        formatted_frontend = scaleway_obj_to_dict(frontend)
+        lb = formatted_frontend.get("lb") or {}
+        formatted_frontend["lb_id"] = lb.get("id")
+        formatted_frontend["backend_id"] = (
+            formatted_frontend.get("backend") or {}
+        ).get("id")
+        # Frontends inherit the project of their parent load balancer.
+        frontends_by_project.setdefault(lb.get("project_id"), []).append(
```
kunaals: nit: since we already have the parent `lbs`, can we derive `project_id` from a `lb_id -> project_id` map instead of trusting the embedded child `lb`? That avoids grouping/loading LB children under `PROJECT_ID=None` if Scaleway ever returns a partial child payload. Same for backends below.

kunaals (approved):
lgtm, left one nit

cartography/intel/scaleway/loadbalancers/loadbalancers.py
kunaals: this is a super nit though and totally fine if we just want to say this is fine as is

PR #2968, opened 2026-06-30
feat(databricks): add intel module with workspace identity coverage

jychp:
### Type of change
- [ ] Documentation update

### Summary

First slice of a new Databricks intel module, scoped to workspace-level identity. Brings:

- `DatabricksWorkspace` (extra label `Tenant`), keyed on workspace host, capturing the token management settings (`tokens_enabled`, `max_token_lifetime_days`).
- Workspace SCIM coverage: `DatabricksUser` (`UserAccount`), `DatabricksServicePrincipal` (`ServiceAccount`), `DatabricksGroup` (`UserGroup`), with `MEMBER_OF` edges from users / SPs / nested groups to their parent groups. Nested-group nesting is read from both the parent's `members` and the child's own `groups` so coverage does not depend on which side Databricks returns.
- `DatabricksToken` from the token management API, with `OWNER_OF` edges from the owning user or service principal.
- Authentication: personal access token (PAT) or OAuth M2M (workspace-level service principal). Module skips cleanly when not configured.
- CLI / `Config` / `sync.py` (`_LazyStage`) wiring, schema docs (`docs/root/modules/databricks/`), and ontology semantic mappings (`tenants`, `useraccounts`, `serviceaccounts`, `groups`) so `_ont_*` projection fields land for cross-provider queries.

Node ids for User / SP / Group / Token are workspace-scoped (`{workspace_id}/{scim_id}` and `{workspace_id}/{token_id}`); the raw SCIM/token id stays as an indexed `scim_id` / `token_id` property. This keeps multi-workspace ingestion (planned in the follow-up) from collapsing same-id principals or tokens across workspaces.

Account-level API surface (`accounts.cloud.databricks.com`) -- `DatabricksAccount` node, account SCIM, federation policies, workspace assignments, permission assignments, cross-cloud links to AWS / GCP / Azure -- is intentionally deferred to a follow-up so this PR stays reviewable and can be exercised end-to-end against a workspace-only sandbox.

### Related issues or links

N/A

### How was this tested?

- Mocked integration suite: `pytest tests/integration/cartography/intel/databricks/ -v` -- 6 passing, covering workspace, users, service principals, groups (incl. nested), tokens, and an explicit multi-workspace collision test for `DatabricksToken.id`.
- End-to-end against a real Databricks workspace (PAT auth). The sync loaded the workspace + SCIM users + service principals + groups + a freshly created PAT, and queries confirmed:
  - Composite ids on User / SP / Group / Token with the raw scim/token id on `scim_id` / `token_id`.
  - `Tenant`, `UserAccount`, `ServiceAccount`, `UserGroup` labels applied and `_ont_*` ontology projection fields populated.
  - `(:DatabricksUser)-[:MEMBER_OF]->(:DatabricksGroup)` and `(:DatabricksUser)-[:OWNER_OF]->(:DatabricksToken)` materialised correctly.
- `make test_lint` clean.

E2E sync log excerpt (run against a sandbox workspace):

```
INFO:cartography.intel.databricks.workspaces:[email protected] in 1.13 seconds
INFO:cartography.intel.databricks.groups:[email protected] in 0.94 seconds
INFO:cartography.intel.databricks.users:[email protected] in 1.21 seconds
INFO:cartography.intel.databricks.service_principals:[email protected] in 0.78 seconds
INFO:cartography.intel.databricks.tokens:[email protected] in 0.42 seconds
```

#### General

#### Proof of functionality

### Notes for reviewers

- IDs are intentionally workspace-scoped composites. The raw SCIM / token id is preserved on `scim_id` / `token_id` (indexed) so existing operator queries by raw id keep working.
- Ontology mappings are registered for `Tenant`, `UserAccount`, `ServiceAccount`, and `UserGroup`; cross-provider queries like \`MATCH (u:UserAccount) WHERE u._ont_email = '...'\` will now surface Databricks rows.
- `DatabricksGroup` nesting reads from both `members` (parent's downward list) and `groups` (child's upward list) and dedupes, since the Databricks SCIM response can populate either side.
- Account-level API + cross-cloud links are deliberately out of scope for this PR; the follow-up will add `DatabricksAccount`, account SCIM, federation policies, workspace assignments, and permission assignments.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -2479,7 +2479,11 @@ class CLI:
                 )
                 databricks_token = os.environ.get(databricks_token_env_var)
             databricks_client_secret = None
-            if databricks_client_id and databricks_client_secret_env_var:
+            if databricks_client_secret_env_var:
+                # Read the secret whenever the env-var flag is set, even if
+                # --databricks-client-id is missing, so the module entry's
+                # partial-OAuth guard sees the asymmetric configuration and
+                # fails loudly instead of silently skipping ingestion.
                 logger.debug(
                     "Reading Databricks OAuth M2M secret from environment variable %s",
                     databricks_client_secret_env_var,
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -52,6 +52,7 @@ PANEL_TAILSCALE = "Tailscale Options"
 PANEL_OPENAI = "OpenAI Options"
 PANEL_ANTHROPIC = "Anthropic Options"
 PANEL_AIRBYTE = "Airbyte Options"
+PANEL_DATABRICKS = "Databricks Options"
 PANEL_DOCKER_SCOUT = "Docker Scout Options"
 PANEL_TRIVY = "Trivy Options"
 PANEL_SYFT = "Syft Options"
@@ -107,6 +108,7 @@ MODULE_PANELS = {
     "openai": PANEL_OPENAI,
     "anthropic": PANEL_ANTHROPIC,
     "airbyte": PANEL_AIRBYTE,
+    "databricks": PANEL_DATABRICKS,
     "docker_scout": PANEL_DOCKER_SCOUT,
     "trivy": PANEL_TRIVY,
     "syft": PANEL_SYFT,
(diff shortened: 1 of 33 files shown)


cartography/cli.py
```diff
@@ -2429,6 +2470,24 @@ def run(
...
+            # Read Databricks credentials
+            databricks_token = None
+            if databricks_token_env_var:
+                logger.debug(
+                    "Reading Databricks PAT from environment variable %s",
+                    databricks_token_env_var,
+                )
+                databricks_token = os.environ.get(databricks_token_env_var)
+            databricks_client_secret = None
+            if databricks_client_id and databricks_client_secret_env_var:
```
kunaals: this drops `--databricks-client-secret-env-var` unless `--databricks-client-id` is also set, so a secret-only oauth config gets passed downstream as "not configured" and silently skips the databricks module instead of hitting the partial-config guard. can we read the secret whenever the env-var flag is provided, or validate the flag pair here, so bad oauth config fails loudly?

kunaals (approved):
approving with 1 nit

PR #2970, opened 2026-06-30
feat(scaleway): add Managed Databases coverage (RDB, Redis, MongoDB)

jychp:
### Type of change

### Summary

Closes the managed-database gap in the Scaleway intel module: AWS / GCP / Azure all ship RDS-style coverage by default, Scaleway didn't.

Adds three new intel modules:

- `ScalewayRdbInstance` — Managed PostgreSQL / MySQL (extra label: `Database`)
- `ScalewayRedisCluster` — Managed Redis (extra label: `Database`)
- `ScalewayMongoDBInstance` — Managed MongoDB (extra label: `Database`)

Endpoints are flattened onto the instance/cluster node (`is_public`, `public_endpoint_*`, `private_endpoint_*`) so the public-DB exposure question can be answered directly, mirroring how `RDSInstance.publicly_accessible` works in AWS.

Each engine type can be attached to one or more VPC private networks via `ATTACHED_TO -> ScalewayPrivateNetwork`.

Ontology mapping is wired up so the new nodes get `_ont_db_*` fields (name, type, version, endpoint, port, encrypted, location) populated and cross-cloud database queries work alongside RDS / Cloud SQL / Azure SQL.

### Related issues or links

N/A

### How was this tested?

**Integration tests:** New `tests/integration/cartography/intel/scaleway/test_databases.py` exercises all three engines with fixture data, asserts the flattened endpoint fields, the `Database` ontology label + `_ont_*` fields, project ownership, and the `ATTACHED_TO` edge.

**End-to-end against the live Scaleway sandbox:** Provisioned one instance of each engine with public endpoints, ran the three new ``sync()`` entry points against a dedicated Neo4j database, then deleted everything:

```
ScalewayRdbInstance     carto-e2e-pg     is_public=True  public=51.159.25.40  :10380  type=PostgreSQL-15
ScalewayRedisCluster    carto-e2e-redis  is_public=True  public=163.172.174.42  :6379   type=redis
ScalewayMongoDBInstance carto-e2e-mongo  is_public=True  public=<...>.mgdb.fr-par.scw.cloud  :27017  type=mongodb
```

Each node carried the `Database` label with `_ont_db_name` / `_ont_db_type` / `_ont_db_endpoint` / `_ont_db_port` / `_ont_db_location` populated.

Full Scaleway suite: 19 passed. Lint: clean (`make test_lint`).

#### General

#### Proof of functionality

### Notes for reviewers

Two engines, two SDK quirks worth flagging:

- **Redis is zone-scoped**, not region-scoped. The other two managed DB APIs accept a `region` parameter (and `list_all_regions` handles them); Redis only accepts a `zone`, so the Redis intel module derives the zone list (`fr-par-1`, `nl-ams-1`, ...) from the shared `DEFAULT_REGIONS` constant and handles the "service not available in this zone" error locally.
- **Endpoint shape differs between engines.** RDB returns a list with one entry per access method (`load_balancer` / `direct_access` / `private_network`); Redis returns `public_network` vs `private_network`; MongoDB returns a `dns_record` for both. Rather than modelling endpoints as a separate node (cf. PR 4's LB frontends/backends), I flattened the most useful scalars onto the parent so simple "is this DB public?" queries don't need a join. If someone needs per-endpoint detail later we can lift them into their own node without breaking the current shape.

### DIFF
--- a/cartography/cli.py
+++ b/cartography/cli.py
@@ -52,6 +52,7 @@ PANEL_TAILSCALE = "Tailscale Options"
 PANEL_OPENAI = "OpenAI Options"
 PANEL_ANTHROPIC = "Anthropic Options"
 PANEL_AIRBYTE = "Airbyte Options"
+PANEL_DATABRICKS = "Databricks Options"
 PANEL_DOCKER_SCOUT = "Docker Scout Options"
 PANEL_TRIVY = "Trivy Options"
 PANEL_SYFT = "Syft Options"
@@ -107,6 +108,7 @@ MODULE_PANELS = {
     "openai": PANEL_OPENAI,
     "anthropic": PANEL_ANTHROPIC,
     "airbyte": PANEL_AIRBYTE,
+    "databricks": PANEL_DATABRICKS,
     "docker_scout": PANEL_DOCKER_SCOUT,
     "trivy": PANEL_TRIVY,
     "syft": PANEL_SYFT,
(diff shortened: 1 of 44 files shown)


kunaals (approved):
lgtm

PR #2973, opened 2026-06-30
feat(tags): unify tag-shape ontology with a :Tag extra label

jychp:
### Type of change

### Summary

Bring `AzureTag`, `GCPLabel`, and `TenableAssetTag` onto the same cross-provider ontology contract that `AWSTag` already uses: they all now carry `ExtraNodeLabels(["Tag"])`. A query of the form

```cypher
MATCH (r)-[:TAGGED|LABELED|HAS_TAG]->(t:Tag {key: 'team'}) RETURN labels(r), t.value
```

now matches tags from AWS, Azure, GCP, and Tenable. Before this change the same query silently returned AWS rows only because the other tag-shaped nodes never carried the unified `:Tag` label.

`TenableAssetTag` also renames its `tag_key` / `tag_value` properties to `key` / `value` so the same `{key, value}` pattern works across all providers. The legacy `tag_key` / `tag_value` properties are kept on the node as DEPRECATED aliases that read from the same source values, and will be removed in v1.0.0.

`TailscaleTag` and `AirbyteTag` are intentionally left out: their shape is a single `name` field, not a `(key, value)` pair, so the cross-provider `:Tag` semantics don't apply to them.

### Related issues or links

- Fixes #

### Breaking changes

None. `:Tag` is additive on existing schemas, and `TenableAssetTag.tag_key` / `tag_value` continue to be written alongside the new `key` / `value` until v1.0.0.

### How was this tested?

- `make test_lint` (pre-commit on the changed files): passes.
- `uv run pytest tests/unit/cartography/intel/tenable/ tests/unit/cartography/intel/gcp/test_labels.py tests/unit/cartography/graph -q`: passes.

#### General

#### Proof of functionality

### Notes for reviewers

Follow-ups intentionally left out of this PR so it stays narrow:
- `GCPLabel` keeps its `:LABELED` relationship; unifying it with `:TAGGED` is a heavier rename and can land later.
- `TenableAssetTag` keeps its `:HAS_TAG` relationship for the same reason.
- No migration job is included: pre-existing tag nodes will pick up `:Tag` (and Tenable nodes will gain `key`/`value`) on the next sync, since the loader sets extra labels and properties via MERGE/SET.


cartography/models/tenable/tags.py
```diff
@@ -59,13 +65,34 @@ class TenableAssetTagToAssetRel(CartographyRelSchema):
...
 @dataclass(frozen=True)
 class TenableAssetTagSchema(CartographyNodeSchema):
     label: str = "TenableAssetTag"
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Tag"])
     properties: TenableAssetTagNodeProperties = TenableAssetTagNodeProperties()
     sub_resource_relationship: TenableAssetTagToTenantRel = TenableAssetTagToTenantRel()
     other_relationships: OtherRelationships = OtherRelationships(
         [
             TenableAssetTagToAssetRel(),
+            TenableAssetTagToAssetTaggedRel(),
```
kunaals, line 96: since this pr adds the new `tagged` relationship, can we also assert it in the tenable tag sync test? right now the test still only proves the deprecated `has_tag` edge, so a broken `tagged` matcher/direction would pass unnoticed. i'd add the same kind of small assertion for one gcp label path too, since those new rels are copy-pasted.

kunaals (changes requested):

kunaals (approved):

PR #2976, opened 2026-07-01
feat(databricks): add Unity Catalog coverage (metastore, catalogs, grants, ...)

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds full Unity Catalog (data plane) coverage to the Databricks intel module,
building on the existing workspace / identity / compute foundation. Unity
Catalog is where the sensitive-data access signal lives, so this ingests the
whole UC surface and, crucially, materialises principal → data access edges for
attack-path analysis.

New node types (one commit per resource type):

- `DatabricksMetastore` (+ `ASSIGNED_METASTORE` from the workspace)
- `DatabricksStorageCredential` — linked to the cloud identity it assumes
  (`ASSUMES_ROLE` → `AWSPrincipal`, `IMPERSONATES` → `GCPServiceAccount`; Azure
  identity ids carried as properties)
- `DatabricksExternalLocation` — `USES_CREDENTIAL` + `BACKED_BY` → `S3Bucket` /
  `GCPBucket` (parsed from the storage URL)
- `DatabricksCatalog` / `DatabricksSchema` / `DatabricksTable` /
  `DatabricksVolume` — the `CONTAINS` hierarchy, tables/volumes linked to their
  backing cloud bucket
- `DatabricksFunction`, `DatabricksConnection` (foreign / Lakehouse Federation)
- `DatabricksRegisteredModel` + `DatabricksModelVersion`
- `DatabricksOnlineTable` (`SOURCED_FROM` → table)
- `DatabricksVectorSearchEndpoint` + `DatabricksVectorSearchIndex`
- `DatabricksArtifactAllowlist`

UC grants are materialised as
`(:DatabricksUser|DatabricksGroup|DatabricksServicePrincipal)-[:HAS_PRIVILEGE {privileges}]->(:DatabricksSecurable)`.
All grantable UC objects share a `DatabricksSecurable` label so a single
relationship type covers every securable, and grant principals are resolved to
the workspace-scoped principal node id (never matched by bare name) so a name
shared across workspaces can't attach to the wrong principal.

Node ids are metastore-scoped (`{metastore_id}/{full_name}`) for UC securables
and workspace-scoped for workspace-API objects, so multi-workspace ingestion
never collapses distinct objects. When a workspace has no metastore assigned,
the whole UC surface is skipped and any previously-ingested UC nodes/edges are
cleaned up.

### Related issues or links

N/A

### How was this tested?

- New integration tests for every resource type (28 databricks tests, all
  green) plus a unit test for the storage-URL parser.
- `make test_lint` passes locally (isort / black / flake8 / mypy).
- End-to-end against a live Databricks sandbox: created a schema, managed
  volume, registered model and a `USE_SCHEMA` grant under the `workspace`
  catalog, ran the real sync into Neo4j, and confirmed the nodes, the
  catalog→schema `CONTAINS` edge, and the group→schema `HAS_PRIVILEGE` grant
  edge all landed; then deleted the sandbox resources.

#### General

#### Proof of functionality

### Notes for reviewers

- One commit per resource type to keep the diff reviewable.
- `DatabricksSecurable` is a shared label on the grantable UC nodes so UC grants
  target a single label; the grant sync reads securables straight from the
  graph and is therefore independent of UC sync order.
- Online tables have no list endpoint, so managed UC tables are probed via
  get-by-name (404 = skip); flagged inline.
- Azure ADLS storage linking is intentionally left out for now (containers are
  keyed by ARM resource id, which isn't derivable from an `abfss://` URL);
  documented in the URL parser.
- Ontology mapping and security rules are deliberately out of scope here and
  planned as a follow-up.


cartography/intel/databricks/grants.py
```diff
@@ -0,0 +1,212 @@
...
+from cartography.intel.databricks.util import DatabricksWorkspaceClient
+from cartography.intel.databricks.util import skip_or_raise_http
+from cartography.models.databricks.grant import DatabricksGroupGrantRel
+from cartography.models.databricks.grant import DatabricksServicePrincipalGrantRel
+from cartography.models.databricks.grant import DatabricksUserGrantRel
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+# Maps a securable node label to its Unity Catalog permissions API path segment.
```
kunaals, line 18: this leaves out several grantable unity catalog securables. `get_securables()` only discovers nodes with `DatabricksSecurable` and `full_name`, but metastores, storage credentials, external locations, and registered models do not all enter that path, so their `HAS_PRIVILEGE` edges never get fetched/materialized. can we add those securables to the shared grant discovery path and map them to the right permissions api type?

cartography/intel/databricks/grants.py
```diff
@@ -0,0 +1,212 @@
...
+    node id, so a downstream MatchLink resolves it to the right principal node.
+    """
+    grants: list[dict[str, Any]] = []
+    for s in securables:
+        uri = (
+            f"/api/2.1/unity-catalog/permissions/"
+            f"{s['securable_type']}/{s['full_name']}"
+        )
+        try:
+            response = api_session.get(uri)
```
kunaals: this fetches only the first permissions page for each securable. the permissions response can return a next-page token, so large grant sets can be silently truncated and then cleanup will remove edges that were just on later pages. can we page through the permissions endpoint before loading/cleaning grants?

cartography/intel/databricks/catalogs.py
```diff
@@ -0,0 +1,94 @@
...
+    """Sync UC catalogs and return them so schemas can iterate per catalog."""
+    catalogs = get(api_session)
+    transformed = transform(catalogs)
+    load_catalogs(
+        neo4j_session,
+        transformed,
+        workspace_id,
+        common_job_parameters["UPDATE_TAG"],
+    )
+    cleanup(neo4j_session, common_job_parameters)
```
kunaals: i think this cleanup still needs to move out of the parent sync. deleting stale catalogs before child uc resources have finished syncing/cleaning can detach hierarchy edges, and then a later child-sync failure can leave orphaned schemas/tables/volumes/functions around. can we run uc cleanup centrally in reverse dependency order after the whole uc sync succeeds?

cartography/intel/databricks/storage_credentials.py
```diff
@@ -0,0 +1,112 @@
...
+    for c in creds:
+        name = c["name"]
+        if not name:
+            raise ValueError("Databricks storage credential returned with empty name")
+        aws = c.get("aws_iam_role") or {}
+        azure_mi = c.get("azure_managed_identity") or {}
+        gcp = c.get("databricks_gcp_service_account") or {}
+        result.append(
+            {
+                "id": c.get("id") or name,
```
kunaals: falling back to bare `name` for the node id seems risky here. storage credential names are metastore-scoped, so an older/malformed payload without `id` can collapse same-named credentials across metastores into one node. can we either fail loudly when the api id is missing or use a metastore-scoped fallback?

kunaals (changes requested):

kunaals (approved):

PR #2982, opened 2026-07-02
feat(databricks): jobs + SQL, ML serving + apps + content, and Delta Sharing coverage

jychp:
### Type of change
- [ ] Documentation update

### Summary

Extends the Databricks module beyond identity / compute / Unity Catalog to cover
workspace workloads, ML serving + apps + content, and Delta Sharing. Three
self-contained commits, one per area:

**Jobs + SQL** (`DatabricksJob`, `DatabricksJobTask`, `DatabricksPipeline`,
`DatabricksSqlWarehouse`, `DatabricksQuery`, `DatabricksAlert`,
`DatabricksDataSource`, `DatabricksDashboard`)
- Jobs + tasks with run-as identity, and task edges to the pipeline / cluster /
  warehouse each task drives (`RUNS_PIPELINE`, `USES_CLUSTER`, `USES_WAREHOUSE`).
- DLT pipelines with run-as and `PUBLISHES_TO` the Unity Catalog catalog.
- SQL warehouses (incl. serverless flag), saved queries, alerts (`MONITORS` a
  query), data sources (`BACKED_BY` a warehouse), and Lakeview + legacy
  dashboards folded into one node by `dashboard_type`.

**ML serving + apps + content** (`DatabricksServingEndpoint`,
`DatabricksServedEntity`, `DatabricksGenieSpace`, `DatabricksApp`,
`DatabricksRepo`, `DatabricksGitCredential`, `DatabricksNotebook`)
- Serving endpoints + served entities (`SERVES`); served entities carry the
  third-party provider for `EXTERNAL_MODEL` egress visibility.
- Genie spaces (`USES_WAREHOUSE`), Databricks Apps (with the run-as service
  principal client id retained), repos and stored git credentials.
- Notebooks are materialised lazily from the paths job tasks reference (no
  workspace-tree walk, no content/permissions), carrying the code-to-cloud
  `(:DatabricksJobTask)-[:RUNS_NOTEBOOK]->(:DatabricksNotebook)` edge.

**Delta Sharing** (`DatabricksShare`, `DatabricksRecipient`,
`DatabricksProvider`, `DatabricksCleanRoom`)
- Shares (`CONTAINS` from metastore) with a `SHARED_WITH` edge to recipients
  built from each share's permission assignments.
- Recipients expose `authentication_type` (TOKEN = open bearer-token external
  sharing vs DATABRICKS = account-to-account), providers, and clean rooms
  (skipped gracefully when external OpenSharing is disabled on the metastore).

Repos are listed under `path_prefix=/Workspace`: without a prefix the endpoint
only returns legacy `/Repos`-rooted repos and silently omits git folders under
user home directories. Delta Sharing cleanup is centralised in
`_cleanup_unity_catalog` so the no-metastore path purges those nodes too, and a
skipped share-permission fetch carries forward the last-known recipients so
`SHARED_WITH` edges are not dropped on a transient 403/404.

Account-level surfaces (cloud account configs, OAuth custom app integrations,
service principal federation policies) live on the `accounts` API and are left
to the account-level module.

### How was this tested?

- `make test_lint` passes; 46 Databricks unit + integration tests pass locally.
- End-to-end against a live Databricks workspace: created real resources (jobs
  with notebook / pipeline / SQL tasks, a DLT pipeline, query, alert, dashboard,
  app, repo, share, notebook), ran the sync, verified the nodes and edges landed
  (e.g. `JobTask-[:RUNS_PIPELINE|USES_WAREHOUSE|RUNS_NOTEBOOK]`, `Pipeline-[:PUBLISHES_TO]->Catalog`,
  `Share-[:CONTAINS]-Metastore`, 37 serving endpoints + 37 served entities), then
  deleted the resources. Clean rooms gracefully skipped where external
  OpenSharing is disabled.

Sample sync logs from the live workspace:

```
INFO:cartography.client.core.tx:Loaded 1 DatabricksWorkspace nodes
INFO:cartography.client.core.tx:Loaded 37 DatabricksServingEndpoint nodes
INFO:cartography.client.core.tx:Loaded 37 DatabricksServedEntity nodes
```

#### General

#### Proof of functionality

- [x] Updated the schema documentation (`docs/root/modules/databricks/schema.md`).

### Notes for reviewers

Each area is its own commit and can be reviewed independently. New node types are
workspace-scoped; Delta Sharing and Unity-Catalog-adjacent nodes are
metastore-scoped and cleaned centrally so they are purged when a workspace loses
its metastore.

### DIFF
--- a/cartography/intel/databricks/clean_rooms.py
+++ b/cartography/intel/databricks/clean_rooms.py
@@ -8,8 +8,8 @@ from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.intel.databricks.util import DatabricksWorkspaceClient
 from cartography.intel.databricks.util import epoch_ms_to_datetime
-from cartography.intel.databricks.util import scoped_id
 from cartography.intel.databricks.util import skip_or_raise_http
+from cartography.intel.databricks.util import uc_id
 from cartography.models.databricks.clean_room import DatabricksCleanRoomSchema
 from cartography.util import timeit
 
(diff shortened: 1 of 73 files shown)


cartography/models/databricks/notebook.py
```diff
@@ -0,0 +1,71 @@
...
+
+@dataclass(frozen=True)
+# (:DatabricksJobTask)-[:RUNS_NOTEBOOK]->(:DatabricksNotebook)
+# Matches every job task whose notebook_path equals this notebook's path; the
+# notebook is a lightweight, path-keyed node (no content / permissions) derived
+# only from the workloads that reference it, so there is no full workspace walk.
+class DatabricksNotebookToJobTaskRel(CartographyRelSchema):
+    target_node_label: str = "DatabricksJobTask"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"notebook_path": PropertyRef("path")},
```
kunaals: this should include workspace scope in the match. notebook nodes are keyed by `scoped_id(workspace_id, path)`, but this relationship matches tasks globally by `notebook_path`, so two workspaces with the same path can create cross-workspace `RUNS_NOTEBOOK` edges.

cartography/intel/databricks/shares.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _recipient_names(
+    neo4j_session: neo4j.Session,
+    api_session: DatabricksWorkspaceClient,
+    share_name: str,
+    share_scoped_id: str,
+) -> list[str]:
+    """Return the recipients granted access to a share (its SELECT grantees)."""
+    try:
+        data = api_session.get(
```
kunaals: this permissions endpoint is paginated. reading only the first page can drop recipients from `SHARED_WITH`, and the subsequent cleanup will make that partial result authoritative. can we loop on `next_page_token` here like the other uc list paths?

cartography/intel/databricks/clean_rooms.py
```diff
@@ -0,0 +1,102 @@
...
+
+    The endpoint returns 400/403 when external OpenSharing is disabled on the
+    metastore; that is expected and skippable, so it never aborts the sync.
+    """
+    results: list[dict[str, Any]] = []
+    params: dict[str, Any] = {"page_size": 100}
+    while True:
+        try:
+            response = api_session.get("/api/2.0/clean-rooms", params=params)
+        except requests.HTTPError as e:
```
kunaals, line 55: i don’t think we should return an empty inventory for 400/403 here. open sharing being disabled is a fetch skip, not proof there are no clean rooms, and central uc cleanup will delete previously ingested nodes after this returns `[]`. use an explicit skip sentinel and gate cleanup/load for this resource.

kunaals (changes requested):

kunaals (approved):

PR #2983, opened 2026-07-02
feat(rules): improve CIS rule accuracy: reduce false positives and align finding cardinality (AWS/GCP/K8s)

jychp:
### Type of change

- [ ] Documentation update

### Summary

A batch of accuracy fixes to existing CIS rules, one commit per rule. Each targets a false positive, a detection gap, or a finding-cardinality mismatch observed on real data:

- **AWS EBS encryption** (`aws_ebs_volume_encryption`): the predicate `volume.encrypted = false` silently treated volumes with `encrypted IS NULL` as compliant, hiding unencrypted volumes. Now matches `NULL` as well.
- **Kubernetes bind/impersonate/escalate** (`kubernetes_bind_impersonate_escalate_permissions`): the default RBAC aggregation ClusterRoles (`cluster-admin`, `admin`, `edit`, `view`) ship on every conformant cluster with these verbs by design, so the `system:` prefix filter alone left them as pure noise. They are now excluded from the ClusterRole fact; user-defined namespace Roles reusing those names still trigger.
- **GCP BigQuery table CMEK** (`gcp_bigquery_tables_without_cmek`): a table with no `kms_key_name` inherits its dataset's default CMEK, so tables under a dataset with `default_kms_key_name` set are in fact encrypted. Those are now excluded to remove false positives.
- **AWS IAM users with direct policies** (`aws_users_with_direct_policy_attachments`, CIS 2.14): the fact emitted one row per `(user, policy)` attachment, so a user with several direct attachments produced multiple findings for the same per-user control. Now aggregated to one finding per user with the attachments surfaced as list fields. Inline policies (which have no ARN) are counted via `coalesce(policy.arn, policy.id)`.
- **AWS full-admin policies** (`aws_policies_with_full_administrative_privileges`, CIS 2.15): the fact emitted one row per `(policy, principal, statement)` tuple. Aggregated to one finding per `(account, policy)` with the attaching principals as evidence. Grouping keeps the account dimension because AWS-managed policies are global and shared across accounts.
- **AWS remote-admin SSH/RDP** (`aws_ipv4_remote_administration_ports_open_to_internet`, `aws_ipv6_...`): added an `in_use` relevancy flag (true when the security group has at least one non-terminated EC2 instance), mirroring the GCP unrestricted SSH/RDP rules. The finding still emits when `in_use` is false; consumers use the flag to gauge relevancy rather than suppressing the finding.

### Related issues or links

None.

### How was this tested?

- `uv run pytest tests/unit/rules/` (1247 passed). The rule unit suite validates rule construction, that every `identity_fields` / returned alias exists on the output model, and framework mappings, exercising all six modified rules.
- `pre-commit run` (isort, black, flake8, mypy) clean on the touched files.

#### General

#### Proof of functionality

### Notes for reviewers

These are pure rule-definition changes (Cypher predicates plus Finding output-model fields); no synced node or relationship schema was touched, so the schema-doc and sync-log checklists do not apply. The two cardinality changes (CIS 2.14 per-user, CIS 2.15 per-`(account, policy)`) intentionally alter finding counts to match the control boundary; `identity_fields` and the count queries were re-keyed together so `passing = total - failing` stays correct. The `in_use` flag on the AWS remote-admin rules deliberately does not filter findings, matching the existing GCP convention.

### DIFF
--- a/cartography/rules/data/rules/cis_aws_iam.py
+++ b/cartography/rules/data/rules/cis_aws_iam.py
@@ -560,8 +560,9 @@ class AdminPolicyAttachedOutput(Finding):
     policy_name: str | None = None
     policy_id: str | None = None
     policy_arn: str | None = None
-    statement_sid: str | None = None
-    principal_arn: str | None = None
+    principal_arns: list[str] | None = None
+    principal_count: int | None = None
+    statement_sids: list[str] | None = None
     account_id: str | None = None
     account: str | None = None
     # True for AWS IAM Identity Center (SSO) reserved roles, which ship with the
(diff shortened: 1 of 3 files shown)


cartography/rules/data/rules/cis_4_0_gcp.py
```diff
@@ -969,6 +969,12 @@ class BigQueryTableCmekMissingOutput(Finding):
     WHERE (table.kms_key_name IS NULL OR table.kms_key_name = '')
       AND (table.expiration_time IS NULL OR table.expiration_time = '')
       AND (table.type IS NULL OR NOT table.type IN ['VIEW', 'EXTERNAL'])
+      // A table with no kms_key_name inherits its dataset's default CMEK, so a
+      // dataset-level default_kms_key_name means the table is in fact encrypted.
+      AND NOT EXISTS {
```
kunaals: this exclusion is too broad. dataset default cmek only helps for tables created after the default was set; existing tables can still have empty `table.kms_key_name` until they're copied or rewritten with a key. since the graph does not track when `dataset.default_kms_key_name` was configured relative to `table.creation_time`, this can hide real cis 7.2 findings. i'd keep this scoped to the table's own `kms_key_name` and drop the dataset-default exclusion.

kunaals (changes requested):

kunaals (approved):

PR #2985, opened 2026-07-02
feat(scaleway): IAM permissions, ontology coverage, and container code-to-cloud

jychp:
### Type of change
- [ ] Documentation update

### Summary

Extends the Scaleway module with cross-provider IAM, ontology, and container supply-chain (code-to-cloud) coverage so Scaleway participates in the same graph queries and rules as AWS/GCP/Azure. Scaleway IAM is project/org-scoped and its registry exposes no per-action or per-resource data through the management SDK, so this work resolves the equivalent signals from the data Scaleway does expose (policy/rule graph, the OCI registry endpoint).

- **IAM permissions** — materialize `(:ScalewayUser|Application|Group)-[:HAS_ROLE]->(:ScalewayPermissionSet)` and factual `(...)-[:CAN_ACCESS]->(:ScalewayProject)` scope edges (declarative MatchLinks) resolved from the policy/rule graph. The AWS/GCP `permission_relationships` engine does not apply (no action strings / resource ARNs); `CAN_ACCESS` carries the rule's project scope (org-scoped rules fan out to all projects) and a `has_condition` flag.
- **Ontology coverage** — map KMS keys to `EncryptionKey`; label serverless functions (`Function`), containers (`ComputeService` + `Container`), and namespaces (`ComputeNamespace`), each with ontology field mappings.
- **Rules** — flag Scaleway principals granted the `IAMManager` permission set in `policy_administration_privileges` (via the new `HAS_ROLE` edge).
- **Container registry** — model image tags and digest-addressed images (`ContainerRegistry` → `REPO_IMAGE` → `ImageTag` → `IMAGE` → `Image`), matching the canonical cross-provider shape.
- **Workload → image** — resolve serverless containers' `registry_image` to a digest at ingest and draw `(:ScalewayServerlessContainer)-[:HAS_IMAGE]->(:Image)`, completing the `RESOLVED_IMAGE` chain for Scaleway.
- **Code-to-cloud** — enrich registry images from the OCI Distribution endpoint: `layer_diff_ids` + `ImageLayer` nodes (dockerfile-matching arm) and `source_uri`/`source_revision`/`source_file` (provenance arm, from OCI labels/annotations and buildx SLSA attestations). The existing registry-agnostic GitHub/GitLab supply-chain matchers then draw `(:Image)-[:PACKAGED_FROM]->(:GitHubRepository|:GitLabProject)`. A Scaleway-specific unmatched-image query ensures every image in a namespace (per repository identity) reaches dockerfile analysis rather than one representative per namespace.

### How was this tested?

- New and updated unit/integration tests for every change (IAM permissions, ontology mappings/labels, the rule, registry tags/images, `HAS_IMAGE`, image layers/provenance, and GitHub/GitLab `PACKAGED_FROM`). `make test_lint` passes locally.
- End-to-end against a real Scaleway sandbox using the `scw` CLI + Docker (all resources created for the run were deleted afterward):
  - IAM: verified `HAS_ROLE` / `CAN_ACCESS` and the `IAMManager` rule against real policies.
  - Registry/code-to-cloud: pushed images (including a buildx `--provenance` build) to a real Scaleway Container Registry and ran the actual sync — confirmed `layer_diff_ids`, `ImageLayer` + `HAS_LAYER`, `source_uri`/`source_revision`, and the full `(:ContainerRegistry)-[:REPO_IMAGE]->(:ImageTag)-[:IMAGE]->(:Image)` + `RESOLVED_IMAGE` chain. This also surfaced (and fixed) that registry blob reads 307-redirect to object storage.

#### General

#### Proof of functionality

### Notes for reviewers

Design decisions worth a look:
- **Permissions granularity**: Scaleway IAM has no per-resource ARNs, so permissions are modeled as `HAS_ROLE` (principal → permission set) + a factual `CAN_ACCESS` scope edge, rather than the AWS/GCP action×ARN engine.
- **Registry shape**: the Scaleway *namespace* is the `ContainerRegistry` (mirrors GCP Artifact Registry, which also holds many images); the named image from `list_images` is folded onto the tag (`image_name`, `uri`, `visibility`) rather than modeled as a separate node.
- **No GitHub/GitLab-side matching changes**: their supply-chain matchers are registry-agnostic; Scaleway plugs in via the canonical labels + `source_uri`. The one addition is a Scaleway-specific unmatched-image query so sibling / same-named-across-namespace images all reach dockerfile analysis.
- The OCI-1.1 referrers API is not implemented by the Scaleway registry (404); attestations are read from the image index's attestation manifest instead.

### DIFF
--- a/cartography/intel/scaleway/container_registry/namespaces.py
+++ b/cartography/intel/scaleway/container_registry/namespaces.py
@@ -36,7 +36,7 @@ def sync(
     org_id: str,
     projects_id: list[str],
     update_tag: int,
-) -> None:
+) -> dict[str, str]:
     namespaces, images, tags = get(client, org_id)
     namespaces_by_project, images_by_project, tags_by_project = transform_namespaces(
         namespaces, images, tags
@@ -49,6 +49,15 @@ def sync(
         update_tag,
     )
     cleanup(neo4j_session, projects_id, common_job_parameters)
+    # Return the tag URI -> digest map so downstream syncs (e.g. serverless
+    # containers) can resolve a `registry_image` pull URI to its digest and
+    # declare a HAS_IMAGE relationship to the Image node.
+    return {
+        tag["uri"]: tag["digest"]
+        for tags in tags_by_project.values()
+        for tag in tags
+        if tag.get("uri") and tag.get("digest")
+    }
 
 
 @timeit
--- a/cartography/intel/scaleway/container_registry/namespaces.py
+++ b/cartography/intel/scaleway/container_registry/namespaces.py
@@ -6,13 +6,19 @@ import scaleway
 from scaleway.registry.v1 import Image
 from scaleway.registry.v1 import Namespace
 from scaleway.registry.v1 import RegistryV1API
+from scaleway.registry.v1 import Tag
+from scaleway_core.api import ScalewayException
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.scaleway.utils import DEFAULT_REGIONS
 from cartography.intel.scaleway.utils import list_all_regions
 from cartography.intel.scaleway.utils import scaleway_obj_to_dict
 from cartography.models.scaleway.container_registry.image import (
     ScalewayContainerRegistryImageSchema,
+)
+from cartography.models.scaleway.container_registry.image_tag import (
+    ScalewayContainerRegistryImageTagSchema,
 )
 from cartography.models.scaleway.container_registry.namespace import (
     ScalewayContainerRegistryNamespaceSchema,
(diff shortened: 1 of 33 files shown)


cartography/intel/scaleway/container_registry/namespaces.py
```diff
@@ -76,16 +126,53 @@ def transform_namespaces(
...
+        if meta["endpoint"] and meta["name"]:
+            row["uri"] = f"{meta['endpoint']}/{meta['name']}:{tag.name}"
+        else:
+            row["uri"] = None
+        tags_by_project.setdefault(project_id, []).append(row)
+
+        if tag.digest:
+            project_digests = seen_digests.setdefault(project_id, set())
+            if tag.digest not in project_digests:
+                project_digests.add(tag.digest)
```
kunaals, line 162: this digest-only load clears the fields owned by the supply-chain enrichment. `ScalewayContainerRegistryImageSchema` sets every property, so rows shaped as just `{"digest": ...}` write `layer_diff_ids`, `source_uri`, `source_revision`, and `source_file` to null before `supply_chain.sync` runs. if an oci fetch fails afterward, the cleanup guard preserves layer nodes, but the image has already lost the fields that drive `HAS_LAYER` and `PACKAGED_FROM`. can we make the registry inventory load avoid overwriting those enrichment fields?

kunaals (changes requested):

kunaals (approved):

PR #2986, opened 2026-07-02
feat(ontology): finding->package/device edges and endpoint finding labels

jychp:
### Type of change
- [x] Documentation update

### Summary

Expands ontology coverage for findings, packages and devices so cross-tool queries work regardless of the scanner that produced the data. Four self-contained commits:

1. **Map Semgrep SCA findings to the Package ontology.** `SemgrepSCAFinding` is labeled `:SecurityIssue` (consistent with the SAST/Secrets/OSS findings). Semgrep dependency nodes (`GoLibrary`/`NpmLibrary`) get a `normalized_id`/`type` computed with the shared `make_normalized_package_id` helper and are registered in the `packages` mapping so they dedup into the abstract `:Package`. The packages linking job then promotes `(:SemgrepSCAFinding)-[:AFFECTS]->(:Package)`, mirroring the existing Trivy path.

2. **Constrain `AFFECTS` to Package.** Adds `RelConstraint(CVE -> Package, AFFECTS)` and `RelConstraint(SecurityIssue -> Package, AFFECTS)`. A package can be affected both by a CVE-backed finding (Trivy) and a rule-based security issue (Semgrep SCA).

3. **Label endpoint findings.** `S1AppFinding` is CVE-shaped (keyed on `cve_id`) so it gets `:CVE` and is registered in the `cves` mapping. `AWSInspectorFinding` is mixed, so it is labeled by `type` via conditional labels: `PACKAGE_VULNERABILITY` -> `:CVE`, `NETWORK_REACHABILITY` -> `:SecurityIssue`.

4. **`AFFECTS` edges from endpoint findings to the abstract `Device`.** The devices linking job walks the `OBSERVED_AS` bridge to MERGE `(:S1AppFinding)-[:AFFECTS]->(:Device)` (via `S1Agent`) and `(:CrowdstrikeFinding)-[:AFFECTS]->(:Device)` (via `CrowdstrikeHost` -> `SpotlightVulnerability` -> `CrowdstrikeFinding`). Cleanup-only rels on `DeviceSchema` let `GraphJob` remove stale edges. This edge is owned by the abstract `Device` node (like `User`/`Package`), so no `RelConstraint` is added.

AWS Inspector is intentionally excluded from the Device edge: its findings target a `ComputeInstance`, not a `Device`.

### Related issues or links

Part of the ongoing ontology normalization effort (see upstream discussion #2772).

### How was this tested?

- Unit: `tests/unit/cartography/graph/test_ontology_rel_constraints.py` (constraint guard), `tests/unit/cartography/intel/ontology/`, `tests/unit/cartography/intel/semgrep/`.
- Integration (Neo4j): new assertions in `tests/integration/cartography/intel/ontology/test_devices.py` (finding -> Device), `tests/integration/cartography/intel/semgrep/test_dependencies.py` (SCA -> Package), `tests/integration/cartography/intel/aws/inspector/test_inspector_sync.py` and `tests/integration/cartography/intel/sentinelone/test_finding.py` (label assertions).
- `pre-commit run` (black, isort, flake8, mypy, pyupgrade) passes on all changed files.

#### General

#### Proof of functionality

### Notes for reviewers

The conditional-label loader removes-then-sets per entry, so a single label can only be driven by one condition (two entries sharing a label would clobber each other). That is why `AWSInspectorFinding` maps one `type` per label and `CODE_VULNERABILITY` is intentionally left unlabeled for now.

### DIFF
--- a/cartography/intel/semgrep/dependencies.py
+++ b/cartography/intel/semgrep/dependencies.py
@@ -166,7 +166,11 @@ def transform_dependencies(raw_deps: List[Dict[str, Any]]) -> List[Dict[str, Any
         version = raw_dep["package"]["versionSpecifier"]
         id = f"{name}|{version}"
 
-        # Cross-tool dedup key for the Package ontology node (matches Trivy/Syft).
+        # Cross-tool dedup key for the Package ontology node.
+        # ponytail: gomod dedup is best-effort. Go versions vary by source (bare
+        # "1.2.3" here vs purl-derived "v1.2.3") and names appear as full module
+        # path or short name, so some Go packages won't merge with Trivy/Syft.
+        # Normalize the leading "v" and the name form on all sources if it matters.
         pkg_type = ECOSYSTEM_TO_PACKAGE_TYPE.get(raw_dep["ecosystem"])
         normalized_id = make_normalized_package_id(
             name=name,
--- a/cartography/intel/semgrep/dependencies.py
+++ b/cartography/intel/semgrep/dependencies.py
@@ -11,6 +11,7 @@ from requests.exceptions import ReadTimeout
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.trivy.util import make_normalized_package_id
 from cartography.models.semgrep.dependencies import SemgrepGoLibrarySchema
 from cartography.models.semgrep.dependencies import SemgrepNpmLibrarySchema
 from cartography.stats import get_stats_client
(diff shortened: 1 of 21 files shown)


cartography/intel/semgrep/dependencies.py
```diff
@@ -158,6 +166,14 @@ def transform_dependencies(raw_deps: List[Dict[str, Any]]) -> List[Dict[str, Any
         version = raw_dep["package"]["versionSpecifier"]
         id = f"{name}|{version}"
 
+        # Cross-tool dedup key for the Package ontology node (matches Trivy/Syft).
```
kunaals: nit/non-blocking: for gomod deps, this builds `golang|...|1.2.3` from semgrep's bare `versionSpecifier`, but purl-derived go packages can keep the `v` prefix (`golang|...|v1.2.3`). that means semgrep and trivy/syft go deps may miss the intended `Package` dedup. normalizing the leading `v` or adding a cross-tool test would make the contract explicit.

kunaals (approved):

PR #2996, opened 2026-07-03
feat(databricks): account API, cloud configs, permission resolution, rules + ontology

jychp:
### Type of change
- [ ] Documentation update

### Summary

Completes the Databricks module to full coverage across the account and workspace API surfaces, building on the existing workspace-level coverage.

**Account API (AWS / GCP; Azure has no Databricks account API)**
- `DatabricksAccountClient` (account OAuth M2M) sharing a base client with the workspace client, gated on new `--databricks-account-id` / `--databricks-account-client-id` / `--databricks-account-client-secret-env-var` flags (workspace-only setups are unchanged).
- `DatabricksAccount` node (`Tenant`) that parents the workspace via a soft `RESOURCE` edge only formed when account creds are configured.
- Account SCIM users / service principals / groups (carrying the same `UserAccount` / `ServiceAccount` / `UserGroup` ontology labels as the workspace-level nodes), account-to-workspace permission assignments, and federation policies (account-wide + per-service-principal).

**Account-level workspace cloud configs (AWS / GCP)**
- Credential / storage / network configurations, private access settings, VPC endpoints, customer-managed encryption keys, network connectivity configs, log delivery, budgets, and account settings, each linked to the AWS / GCP nodes already in the graph where they exist (e.g. `ASSUMES_ROLE` -> `AWSPrincipal`, `BACKED_BY` -> `S3Bucket`, `USES_VPC` -> `AWSVpc`, `REFERENCES_KEY` -> `KMSKey` / `GCPCryptoKey`).

**Permission resolution (principal -> resource)**
- Materialises `HAS_PERMISSION` edges from workspace object ACLs (clusters, cluster policies, instance pools, jobs, pipelines, SQL warehouses, serving endpoints, apps, secret scopes) via a shared `DatabricksAclObject` label, so access-path analysis works beyond inventory.

**Rules, ontology, and code-to-cloud**
- Four Databricks security detection rules: personal access tokens that never expire, workspaces without a token lifetime cap, IP access lists that allow all addresses, and token-authenticated Delta Sharing recipients.
- Ontology labels: `Database` (Unity Catalog catalog / schema / table), `ObjectStorage` (external location / volume), `NetworkAccessControl` (IP access list), plus the account identity labels above, with mappings.
- Code-to-cloud analysis job linking `DatabricksRepo` -> `GitHubRepository` by normalised URL.

`WORKLOAD_PARENT` edges from compute to the workspace are intentionally not added: the workspace is a `Tenant`, not part of the ontology workload chain, and the existing `RESOURCE` edges already express that parentage.

### How was this tested?

- `make test_lint` passes (black / isort / flake8 / mypy / pyupgrade).
- Full unit suite and all Databricks integration tests pass; ontology mapping, rel-constraint, rules, and schema-doc tests included.
- Live end-to-end run against a real workspace: created a secret scope and a cluster policy with ACLs, ran the module, and confirmed the `HAS_PERMISSION` edges resolved exactly (group/user -> object with `permission_level`); test resources were deleted afterward.

Example sync output (permission resolution against the live workspace):

```
HAS_PERMISSION: DatabricksGroup 'users'  -> DatabricksClusterPolicy 'cartography-e2e-policy' [CAN_USE]
HAS_PERMISSION: DatabricksGroup 'admins' -> DatabricksClusterPolicy 'cartography-e2e-policy' [CAN_USE]
HAS_PERMISSION: DatabricksUser 'jeremy@...' -> DatabricksSecretScope 'cartography-e2e' [MANAGE]
HAS_PERMISSION: DatabricksGroup 'users'  -> DatabricksSecretScope 'cartography-e2e' [READ]
```

The account-level API surface is covered by integration tests with recorded API responses; a live account-level run is pending an account-admin sandbox (the workspace sandbox available here cannot exercise the account OAuth flow non-interactively). Partial-read cleanup is guarded across the ACL / assignment / federation syncs so a `403`/`404` on one object never deletes still-valid edges.

#### General

#### Proof of functionality

### Notes for reviewers

- New account-level nodes are scoped to `DatabricksAccount`; the workspace-only path is unchanged when the `--databricks-account-*` flags are absent.
- The ACL / permission-assignment / federation-policy syncs skip cleanup when any per-object read returned `403`/`404`, to avoid deleting edges that were not refreshed that run (mirrors the existing artifact-allowlists completeness gate).


cartography/intel/databricks/federation_policies.py
```diff
@@ -0,0 +1,174 @@
...
+                    "sp_scim_id": sp["scim_id"],
+                }
+            )
+    return policies, complete
+
+
+def _flatten(policy: dict[str, Any], account_id: str) -> dict[str, Any]:
+    oidc = policy.get("oidc_policy") or {}
+    return {
+        "id": account_scoped_id(account_id, str(policy["id"])),
```
kunaals: This will abort on the real federation policy shape. Databricks' generated model exposes `policy_id` and `uid`, not `id`; with SDK-shaped input this raises `KeyError: 'id'`. The fixture currently invents `id`, so it misses the production path.

cartography/intel/databricks/permissions.py
```diff
@@ -0,0 +1,328 @@
...
+# straight from each node's intel transform: clusters key by cluster_id, jobs by
+# job_id, apps / serving endpoints by name, and so on.
+_ACL_OBJECT_BY_LABEL = {
+    "DatabricksCluster": ("clusters", "cluster_id"),
+    "DatabricksClusterPolicy": ("cluster-policies", "policy_id"),
+    "DatabricksInstancePool": ("instance-pools", "instance_pool_id"),
+    "DatabricksJob": ("jobs", "job_id"),
+    "DatabricksPipeline": ("pipelines", "pipeline_id"),
+    "DatabricksSqlWarehouse": ("sql/warehouses", "warehouse_id"),
+    "DatabricksServingEndpoint": ("serving-endpoints", "name"),
```
kunaals: Serving endpoint permissions need to be fetched by the endpoint's system `id`, not `name`. Databricks documents `ServingEndpoint.id` as the value used by the Permissions API, but the serving endpoint transform never stores it. As written, those ACL reads can 404, mark the workspace partial, and suppress `HAS_PERMISSION` cleanup.

cartography/intel/databricks/account_workspaces.py
```diff
@@ -0,0 +1,78 @@
...
+
+@timeit
+def transform(workspaces: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    result: list[dict[str, Any]] = []
+    for w in workspaces:
+        deployment_name = w.get("deployment_name")
+        numeric_id = w.get("workspace_id")
+        if not deployment_name or numeric_id is None:
+            continue
+        host_id = f"{deployment_name}.cloud.databricks.com".lower()
```
kunaals: This hardcodes the AWS workspace suffix. On GCP the workspace host is `<deployment>.gcp.databricks.com`, so this creates a duplicate workspace node instead of colliding with the workspace API node; the account `RESOURCE` edge and assignment edges then attach to the wrong node.

cartography/models/databricks/workspace.py
```diff
@@ -12,14 +18,45 @@ class DatabricksWorkspaceNodeProperties(CartographyNodeProperties):
     host: PropertyRef = PropertyRef("host", extra_index=True)
     tokens_enabled: PropertyRef = PropertyRef("tokens_enabled")
     max_token_lifetime_days: PropertyRef = PropertyRef("max_token_lifetime_days")
+    # Numeric account-API workspace id + deployment/name, set by the account
+    # workspaces sync when account creds are configured (None on the
+    # workspace-only path). Lets workspace permission assignments key off the
+    # numeric id the account API reports.
+    workspace_id: PropertyRef = PropertyRef("workspace_id", extra_index=True)
```
kunaals, line 25: Adding this to the shared workspace schema means every workspace load writes it. The workspace-only sync does not provide `workspace_id`, `deployment_name`, or `workspace_name`, so a normal workspace sync after the account sync will set these account-enriched properties to null.

cartography/intel/databricks/account_settings.py
```diff
@@ -0,0 +1,115 @@
...
+                    setting_name,
+                )
+                continue
+            raise
+        data["_setting_name"] = setting_name
+        results.append(data)
+    return results
+
+
+def _setting_value(data: dict[str, Any]) -> str | None:
```
kunaals, line 67: This misses the real personal-compute response shape. The SDK model has `personal_compute: {"value": "ON"|"DELEGATE"}`, not `boolean_val`; with that shape this returns `None`, so the one security signal this module loads is silently lost.

cartography/intel/databricks/budgets.py
```diff
@@ -0,0 +1,90 @@
...
+    try:
+        data = api_session.get(uri) or {}
+    except requests.HTTPError as exc:
+        if exc.response is not None and exc.response.status_code == 404:
+            logger.info(
+                "Databricks budgets endpoint not available for this account; skipping."
+            )
+            return []
+        raise
+    return data.get("budgets", []) or []
```
kunaals: The budgets list endpoint paginates with `next_page_token`. Fetching only the first page means later-page budgets are omitted and then deleted by the unconditional cleanup. Please walk the token, like the existing `uc_list` helper does for the other Databricks paginated APIs.

cartography/data/jobs/analysis/databricks_code_to_cloud.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Databricks - Code to cloud repo linking",
+  "statements": [
+    {
+      "__comment": "Link a Databricks Git-backed repo to the GitHub repository that hosts its source, normalising the trailing .git so the workspace repo and the ingested GitHubRepository match. Runs in the analysis phase so both modules have already synced.",
+      "query": "MATCH (r:DatabricksRepo) WHERE r.url IS NOT NULL AND toLower(coalesce(r.provider, '')) CONTAINS 'github' WITH r, replace(r.url, '.git', '') AS rurl MATCH (gh:GitHubRepository) WHERE replace(gh.url, '.git', '') = rurl MERGE (r)-[rel:SOURCED_FROM]->(gh) ON CREATE SET rel.firstseen = timestamp() SET rel.lastupdated = $UPDATE_TAG",
```
kunaals: Two issues here: this packaged job is never invoked by the Databricks sync, so the documented `SOURCED_FROM` edge won't exist in a default run; and `replace(url, '.git', '')` strips every `.git`, not just a trailing suffix. Use a suffix trim and wire the job into the module if this edge is meant to be produced normally.

kunaals (changes requested):

kunaals (approved):

PR #2999, opened 2026-07-07
fix(okta): retry non-JSON SDK errors across all paged requests

jychp:
### Type of change

### Summary
The legacy Okta SDK parses every error response body as JSON and raises `JSONDecodeError` (instead of `OktaError`) when Okta, a proxy, or a rate-limiter returns an empty or non-JSON error body. These edge responses are typically transient, but they crash the sync outright.

A previous fix added a bounded retry, but only around the group-members request. The group list fetch and the three application listing requests were still unprotected, so a transient non-JSON response on any of those still aborted the entire sync (the group list is fetched first, before the protected member fetch is ever reached).

This change lifts the retry into a single shared helper, `okta_paged_request_with_retry()` in `utils.py`, and routes every paged Okta request through it:

- group list (`_get_okta_groups`)
- group members (`_get_okta_group_members`)
- application list (`_get_okta_applications`)
- application-assigned users (`_get_application_assigned_users`)
- application-assigned groups (`_get_application_assigned_groups`)

Normal `OktaError` responses keep their existing behavior and are left for each caller to handle.

### Breaking changes
None.

### How was this tested?
- New unit tests for the shared helper in `tests/unit/cartography/intel/okta/test_utils.py` (retry-then-succeed and raise-after-max-attempts).
- Existing group-member retry tests repointed at the shared helper.
- `make test_lint` passes locally (flake8, black, isort, mypy).

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/okta/utils.py
+++ b/cartography/intel/okta/utils.py
@@ -1,6 +1,8 @@
 # Okta intel module - utility functions
+import json
 import logging
 import time
+from typing import Callable
 
 from okta.framework import PagedResults
 from okta.framework.ApiClient import ApiClient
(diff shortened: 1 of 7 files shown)


cartography/intel/okta/utils.py
```diff
@@ -15,6 +17,46 @@
...
 
+# The legacy Okta SDK parses every error body as JSON. When Okta / a proxy /
+# a rate-limiter returns an empty or non-JSON error body, the SDK raises
+# JSONDecodeError instead of OktaError, which crashes the sync. These edge
+# responses are usually transient, so we retry a few times before giving up.
+OKTA_REQUEST_ATTEMPTS = 3
+OKTA_RETRY_DELAY_SECONDS = 1
+
+
+def okta_paged_request_with_retry(
```
kunaals, line 31: this fixes the groups/apps callers, but users still bypass this helper via `UsersClient.get_paged_users()` in `users.py`. that sdk path also goes through `ApiClient.get/get_path`, so a non-json okta response can still crash user sync without retry. can we route the users pager through the same retry wrapper too?

kunaals (approved):
approving with small nit

PR #3006, opened 2026-07-09
fix(vercel): page access-groups with the 'next' cursor parameter

vidigoat:
### Type of change

### Summary
`paginated_get()` in the Vercel module always sends the `pagination.next` value back under the `until` query parameter. That is only correct for Vercel's timestamp-cursor endpoints (e.g. deployments, which the helper was originally written against).

The access-groups endpoints use a different pagination scheme: `/v1/access-groups` and the per-group `/members` and `/projects` sub-resources return an opaque **string** continuation cursor and expect it back under a **`next`** query parameter, not `until` (see the Vercel REST API reference for [list access groups](https://vercel.com/docs/rest-api/reference/endpoints/access-groups/list-access-groups-for-a-team-project-or-member) and [list members](https://vercel.com/docs/rest-api/reference/endpoints/access-groups/list-members-of-an-access-group)).

Because the cursor was sent under the wrong parameter name, Vercel ignored it and kept returning the first page. Since `pagination.next` stayed non-null, the `while True` loop never met its exit condition, so `get()` never terminated for:
- any team with more than one page of access groups (>100), or
- any access group with more than 100 members or more than 100 projects.

The fix adds a `pagination_param` argument to `paginated_get()` (defaulting to `until`, so deployments and every other existing caller are unchanged) and passes `pagination_param="next"` at the three access-groups call sites.

### How this maps to the graph
No node or relationship shape changes. This only fixes the API fetch/pagination layer that feeds `VercelAccessGroup` nodes and their `HAS_MEMBER` / `HAS_ACCESS_TO` relationships, so large teams' access groups (and their members/projects beyond the first 100) are now actually paginated through and ingested instead of hanging the sync.

### Breaking changes
None. The default `pagination_param="until"` preserves the behavior of every existing caller.

### How was this tested?
Added `tests/unit/cartography/intel/vercel/test_util.py` covering `paginated_get`, which previously had no unit coverage:
- follows a `next` string cursor across pages and terminates, asserting the cursor is sent under `next` (not `until`) — this test fails on the current code and passes with the fix
- default `until` cursor pagination still works (deployments-style)
- single-page, bare-array, plan-gated `403`, and missing-`result_key` cases

`make test_lint` passes locally (isort, black, flake8, pyupgrade, mypy). The full unit suite passes (`uv run pytest tests/unit`, 2934 passed).

#### General

#### Proof of functionality


cartography/intel/vercel/util.py
```diff
@@ -18,6 +18,7 @@ def paginated_get(
     team_id: str,
     params: dict[str, Any] | None = None,
     limit: int = 100,
+    pagination_param: str = "until",
```
jychp, line 21: The `until` default now silently governs the other ~15 `paginated_get` callers (projects, domains, aliases, dnsrecords, webhooks, integrations, projectdomains, team members). Any of them that pages via Vercel's opaque `next` cursor rather than the `until` timestamp cursor has the same infinite-loop bug this PR fixes. Pre-existing, not a regression here, but confirm each genuinely-paginating endpoint uses the timestamp scheme or the fix is incomplete for them.

jychp (changes requested):
Thanks for your contribution.

Can you check the following point and fix the DCO failure ?

cartography/intel/vercel/util.py
vidigoat: Fixed the DCO, the commit is amended with a Signed-off-by line and the check is green now.

On the pagination question: I went through every `paginated_get` caller. They split cleanly and the `until` default is correct for all the timestamp endpoints.

- Bare-array endpoints with no `pagination` object: `edge-config`, `edge-config/{id}/tokens`, `integrations/log-drains`. These return a raw list, so the `isinstance(data, list)` branch returns before `pagination_param` is ever read.
- Timestamp-cursor endpoints (everything else here: `projects`, `deployments`, `domains`, `aliases`, `dnsrecords`, `projectdomains`, `webhooks`, `integrations/configurations`, `teams/{id}/members`, ...). These all use Vercel's standard `Pagination` object where `next`/`prev` are timestamps. The domains reference documents `pagination.next` as "Timestamp that must be used to request the next page", and the only pagination query params on these endpoints are `since`/`until` (there is no `next` param). So passing the cursor as `until` is correct.
- Continuation-cursor endpoints, which are only the access-groups ones: `/v1/access-groups`, `.../members`, `.../projects`. The reference defines `next` there as a "Continuation cursor" string and the request param is literally `next`. Those are the three call sites this PR switches to `pagination_param="next"`.

So the only endpoints that need the non-default param are access-groups, which are exactly the ones changed here. Happy to add a one-line comment at each call site noting the scheme if you'd like it explicit.

jychp (approved):

PR #3009, opened 2026-07-09
fix(subimage): read frameworks from Page[T] envelope

jychp:
### Type of change

### Summary
The frameworks endpoint changed its response envelope from `{"frameworks": [...]}` to the standardized paginated form `{"items": [...], "total_count": N, "limit": N, "offset": 0}`. The `get()` helper still read the old `frameworks` key, so every sync failed with `KeyError: 'frameworks'`. This reads the `items` key instead. Only this module was affected; the sibling modules parse bare lists or other keys.

### How was this tested?
Added a unit test that drives `get()` against a mocked HTTP response using the paginated envelope and asserts the extracted list. The existing integration test mocked `get()` entirely, so it never exercised the JSON key extraction and did not catch the drift.

```
$ .venv/bin/python -m pytest tests/unit/cartography/intel/subimage/test_frameworks.py -q
1 passed
```

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/subimage/frameworks.py
+++ b/cartography/intel/subimage/frameworks.py
@@ -32,7 +32,7 @@ def get(api_session: requests.Session, base_url: str) -> list[dict[str, Any]]:
         timeout=_TIMEOUT,
     )
     response.raise_for_status()
-    return response.json()["frameworks"]
+    return response.json()["items"]
 
 
 def transform(raw_data: list[dict[str, Any]]) -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


cartography/intel/subimage/frameworks.py
```diff
@@ -32,7 +32,7 @@ def get(api_session: requests.Session, base_url: str) -> list[dict[str, Any]]:
         timeout=_TIMEOUT,
     )
     response.raise_for_status()
-    return response.json()["frameworks"]
+    return response.json()["items"]
```
kunaals: this now reads the page[t] envelope, but it only returns the first `items` array. since the response includes `total_count`, `limit`, and `offset`, this can silently truncate when there is more than one page; `sync()` then runs cleanup and could delete frameworks from later pages. if the endpoint is guaranteed single-page, can we document/assert that? otherwise this should paginate before cleanup.

kunaals (changes requested):

kunaals (approved):

PR #3010, opened 2026-07-09
feat(tailscale): link devices to cloud instances

jychp:
### Type of change
- [x] Documentation update

### Summary
Adds a global ontology analysis job that links current Tailscale devices to current EC2/GCP compute instances with `IS_INSTANCE` when hostname/DNS or public endpoint IP correlation is unique. This lets Tailscale endpoint nodes connect back to cloud compute resources without treating ambiguous hostnames as authoritative.

### Related issues or links
N/A

### Breaking changes
None.

### How was this tested?
- `uv run --frozen pytest -q tests/integration/cartography/intel/ontology/test_tailscale_device_instance_linking.py`
- `uv run --frozen pytest -q tests/unit/cartography/data/jobs/test_syntax.py tests/integration/cartography/data/jobs/test_syntax.py`
- `uv run --frozen pre-commit run --files cartography/data/jobs/analysis/tailscale_device_instance_linking.json cartography/intel/ontology/__init__.py tests/integration/cartography/intel/ontology/test_tailscale_device_instance_linking.py docs/root/modules/tailscale/schema.md`

#### General

#### Proof of functionality

- [ ] Updated the schema README. Not applicable: `docs/schema/README.md` is not present in this checkout.

### Notes for reviewers
The analysis deliberately requires uniqueness on both sides of the hostname/IP match to avoid linking common or reused hostnames across accounts and projects.

### Test plan
- Run the focused ontology integration test to verify EC2 hostname and GCP endpoint-IP linkage.
- Run analysis job syntax tests to verify the new Cypher parses and executes.
- Confirm stale `TailscaleDevice-[:IS_INSTANCE]->ComputeInstance` edges are removed when they are not refreshed by the current run.


cartography/data/jobs/analysis/tailscale_device_instance_linking.json
```diff
@@ -0,0 +1,16 @@
...
...AS device_key WITH device, instance WHERE device_key = instance_key RETURN DISTINCT device, instance UNION MATCH (device:TailscaleDevice) WHERE device.lastupdated = $UPDATE_TAG AND device.client_connectivity_endpoints IS NOT NULL UNWIND device.client_connectivity_endpoints AS endpoint WITH device, endpoint WHERE endpoint =~ '^\\\\d+\\\\.\\\\d+\\\\.\\\\d+\\\\.\\\\d+:\\\\d+$' WITH device, split(endpoint, ':')[0] AS ip WHERE ip <> '' MATCH (instance:ComputeInstance) WHERE instance.lastupdated = $UPDATE_TAG AND (instance:EC2Instance OR instance:GCPInstance) WITH device, ip, instance, [instance.publicipaddress, instance.public_ip] AS raw_ips UNWIND raw_ips AS raw_ip WITH device, instance, ip, raw_ip WHERE raw_ip IS NOT NULL AND trim(toString(raw_ip)) = ip RETURN DISTINCT device, instance } WITH device, collect(DISTINCT instance) AS candidate_instances WHERE size(candidate_instances) = 1 WITH device, candidate_instances[0] AS instance WITH instance, collect(DISTINCT device) AS candidate_devices WHERE size(candidate_devices) = 1 WITH candidate_devices[0] AS device, instance MERGE (device)-[r:IS_INSTANCE]->(instance) ON CREATE SET r.firstseen = timestamp() SET r.lastupdated = $UPDATE_TAG",
```
kunaals: this IP branch extracts all IPv4 Tailscale endpoints, but it only compares them against public instance fields. Can we include the EC2 private IP field here as well, i think it is instance.privateipaddress? would be good to add a regression test for 10.x.x.x:41641 -> EC2Instance.privateipaddress, and keep the one-to-one guard so overlapping private ranges do not create ambiguous links

cartography/data/jobs/analysis/tailscale_device_instance_linking.json
```diff
@@ -0,0 +1,16 @@
...
+      "iterative": false
+    },
+    {
+      "__comment": "Cleanup stale TailscaleDevice IS_INSTANCE edges after all current edges have been merged.",
+      "query": "MATCH (:TailscaleDevice)-[r:IS_INSTANCE]->(:ComputeInstance) WHERE r.lastupdated <> $UPDATE_TAG OR r.lastupdated IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r RETURN COUNT(*) AS TotalCompleted",
```
kunaals, line 11: this only works when Tailscale and AWS/GCP were loaded in the same sync invocation as ontology. if ontology runs separately, or one source module is skipped, this query will not refresh otherwise-valid matches and the cleanup could delete valid, old edges. i think we'll have to make this cleanup a bit more involved or loosen the lastupdated = $UPDATE_TAG in the creation query

kunaals (changes requested):
big question is on cleanup

kunaals (approved):

PR #3014, opened 2026-07-10
feat(kubernetes): model NetworkPolicy and capture container listening ports

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds two Kubernetes ingestion signals that let the graph express intra-namespace
network reachability. Today the Kubernetes module models neither, so any consumer
that reasons about pod-to-pod reachability has to assume a flat namespace where
every pod is reachable from every other. That is both inaccurate and, on dense
clusters, expensive.

**1. New `KubernetesNetworkPolicy` node.** No NetworkPolicy-aware label existed, so
a namespace locked down by a default-deny policy was indistinguishable from a
wide-open one. This adds a `KubernetesNetworkPolicy` node carrying the raw
`pod_selector`, `policy_types`, and ingress/egress rule sets, plus derived
`restricts_ingress` / `restricts_egress` booleans. Relationships:

```
(:KubernetesCluster)-[:RESOURCE]->(:KubernetesNetworkPolicy)
(:KubernetesNamespace)-[:CONTAINS]->(:KubernetesNetworkPolicy)
(:KubernetesNetworkPolicy)-[:APPLIES_TO]->(:KubernetesPod)
```

The `podSelector` is resolved to pods at ingest and emitted as `APPLIES_TO`,
mirroring the existing `KubernetesService`/`TARGETS` pattern. An empty selector
applies to every pod in the namespace, per Kubernetes semantics.

**2. Container listening ports on `KubernetesContainer`.** The node only carried
`host_ports` (hostPort mappings, rarely used), so there was no record of the
`containerPort`s a container actually listens on. This adds `container_ports`
(a JSON list of `{container_port, protocol, name}`) and a flat, queryable
`container_port_numbers` list of the TCP/UDP port numbers. An empty
`container_port_numbers` means the container declares no listening ports.

Design notes:
- `policy_types` follows the documented Kubernetes default when unset (Ingress
  always applies; Egress applies when egress rules exist) rather than trusting the
  field to be populated by the apiserver.
- Only equality-based `matchLabels` are resolved into `APPLIES_TO` edges. A selector
  that uses set-based `matchExpressions` emits no edges (rather than over-matching on
  its `matchLabels` subset and claiming false coverage); the full selector is still
  retained on the `pod_selector` property for later evaluation.

### Related issues or links

N/A

### How was this tested?

- New unit tests for the NetworkPolicy transform (`transform_network_policies`),
  covering empty-selector-selects-all, label-scoped matching, egress policies, and
  the raw rule/selector JSON shape.
- New unit test for container port extraction in `transform_pods`, covering TCP/UDP
  filtering and the structured/flat representations.
- New integration test loading `KubernetesNetworkPolicy` and asserting the
  `RESOURCE`, `CONTAINS`, and `APPLIES_TO` relationships and the restriction flags,
  plus an integration assertion for `container_port_numbers`.
- Commands (integration tests use the standard Neo4j test container):
  ```
  pytest tests/unit/cartography/intel/kubernetes
  pytest tests/integration/cartography/intel/kubernetes
  ```
  89 kubernetes unit + integration tests pass. `make test_lint` (pre-commit) passes.

#### General

#### Proof of functionality

### Notes for reviewers

- This is purely upstream ingestion. It adds the node/fields to the graph; querying
  them is left to downstream consumers.
- I did not attach real-environment sync logs because I do not have a live cluster
  with NetworkPolicies to sync against; the behavior is exercised by the new unit and
  integration tests instead.


jychp:
### Proof of functionality: real-environment sync

Ran the new `KubernetesNetworkPolicy` + container-port ingestion against a live EKS
cluster (Kubernetes v1.31) using the standard sync path
(`sync_kubernetes_cluster` -> `sync_namespaces` -> `sync_pods` -> `sync_network_policies`).
Environment-specific identifiers (cluster, namespace, workload pod) are redacted;
standard EKS system components are shown as-is.

**Cartography sync logs (INFO):**
```
INFO cartography.client.core.tx: Loaded 1 KubernetesNetworkPolicy nodes
INFO cartography.graph.statement: Completed KubernetesNetworkPolicy statement #1
INFO cartography.graph.statement: Completed KubernetesNetworkPolicy statement #2
INFO cartography.graph.statement: Completed KubernetesNetworkPolicy statement #3
INFO cartography.graph.statement: Completed KubernetesNetworkPolicy statement #4
INFO e2e: sync_pods returned 10 pods
```
The full run also loaded the cluster, namespaces, pods, and containers and ran each
cleanup job with no exceptions.

**Post-sync verification (Cypher):**

`KubernetesNetworkPolicy` node:
```
policy_types=['Egress']  restricts_ingress=False  restricts_egress=True
pod_selector.match_labels = {app.kubernetes.io/name: <redacted>, app.kubernetes.io/instance: <redacted>}
```

`podSelector` resolved to a live pod via matchLabels:
```
(:KubernetesNetworkPolicy)-[:APPLIES_TO]->(:KubernetesPod)   1 edge  (redacted pod)
```

Wiring present:
```
(:KubernetesCluster)-[:RESOURCE]->(:KubernetesNetworkPolicy)
(:KubernetesNamespace)-[:CONTAINS]->(:KubernetesNetworkPolicy)
```

Container declared listening ports (`container_port_numbers`), showing protocol
handling and cross-protocol de-duplication on real workloads:
```
coredns            [53, 9153]   # UDP dns + TCP dns-tcp + TCP metrics; 53 deduped across protocols
aws-node           [61678]
aws-eks-nodeagent  [8162]
<app-a>            [8000]
<app-b>            [8080]
```
The structured `container_ports` property retains `{container_port, protocol, name}` per port.

**Counts:** 1 cluster, 7 namespaces, 10 pods, 11 containers, 1 NetworkPolicy, 1 `APPLIES_TO` edge. Sync completed cleanly with no errors.

cartography/intel/kubernetes/networkpolicies.py
```diff
@@ -0,0 +1,245 @@
...
+from cartography.intel.kubernetes.util import K8sClient
+from cartography.models.kubernetes.networkpolicies import KubernetesNetworkPolicySchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def get_network_policies(client: K8sClient) -> list[V1NetworkPolicy]:
+    items = k8s_paginate(client.networking.list_network_policy_for_all_namespaces)
```
kunaals: could we use `raise_on_forbidden=True` here and skip load/cleanup on a 401 or 403, like the secrets sync does? right now a missing permission becomes an empty result and cleanup can delete previously synced policies. the documented clusterrole also needs `list networkpolicies`, otherwise the recommended setup won't ingest this data at all.

cartography/intel/kubernetes/networkpolicies.py
```diff
@@ -0,0 +1,245 @@
...
+      edges that do not hold), a selector that uses matchExpressions produces no
+      APPLIES_TO edges rather than over-matching on its matchLabels subset. The
+      full selector is retained on the node's pod_selector property regardless.
+    - Otherwise the equality-based matchLabels are matched.
+    """
+    selector = pod_selector or {}
+    match_labels: dict[str, str] | None = selector.get("match_labels")
+    match_expressions = selector.get("match_expressions")
+
+    if match_expressions:
```
kunaals: can we resolve `matchExpressions` here instead of dropping all `APPLIES_TO` edges for those policies? these are standard selectors, and returning no matches makes valid policies invisible to anything using these edges for reachability. supporting `In`, `NotIn`, `Exists`, and `DoesNotExist` should cover the kubernetes selector semantics.

kunaals (changes requested):

kunaals (approved):

PR #3025, opened 2026-07-14
fix(cve): guard _get_primary_metric against empty CVSS metric lists

vidigoat:
### Type of change

### Summary

`_get_primary_metric` in `cartography/intel/cve/feed.py` returns the first metric (`metrics[0]`) when no entry has `type == "Primary"`. When the CVSS version key maps to an **empty list**, that raises `IndexError: list index out of range`, which propagates out of `transform_cves` and aborts the entire CVE sync.

The NVD 2.0 feed reaches this path via:

```python
cvss31_metrics = cve.get("metrics", {}).get("cvssMetricV31")
cvss31 = _get_primary_metric(cvss31_metrics)
```

The equivalent helper in `cartography/intel/cve_metadata/nvd.py` already guards exactly this case:

```python
return metrics[0] if metrics else None
```

This aligns the `cve.feed` copy with that behavior (returning `None` for an empty list instead of crashing) and widens the annotation to `Optional[...]`, since callers pass the result of `dict.get()` which may be `None`.

### Related issues or links

None — found while reading the CVE module. The two copies of `_get_primary_metric` had diverged; this makes them consistent.

### How was this tested?

Added unit tests in `tests/unit/cartography/intel/cve/test_feed.py` covering the primary, fallback, empty-list, and `None` cases. The empty-list test fails before the change (`IndexError`) and passes after. Full `test_feed.py` passes (9 tests). `black`, `isort`, `flake8`, and `pyupgrade --py36-plus` are clean on both changed files.

#### General

#### Proof of functionality


jychp:
@vidigoat thanks for your contribution, will review it quickly.

You should use the new CVE Metadata module, it is more robust and more aligned with the new cartography patterns

vidigoat:
Thanks! Agreed that `cve_metadata` is the better home for CVE work going forward, I can see it's the module wired with the newer patterns (typed models, effect_tags, EPSS).

Quick context on why I targeted `cve/feed.py` here: the `cve` stage is still registered in `sync.py` (`start_cve_ingestion`), so anyone running that stage can still hit this `IndexError` and lose the whole sync. And `cve_metadata`'s own `_get_primary_metric` already guards the empty-list case (`return metrics[0] if metrics else None`) - this PR just brings the older copy to parity with it, nothing more.

That said, happy to defer to you on scope: if the `cve` stage is on its way out, I'll close this and put the effort into `cve_metadata` instead - is there anything there you'd like a hand with? Otherwise this stays a safe 8-line consistency fix for the module while it's still live.

jychp (commented):
Thx for the fix, as said before you should use CVE Metada that is more complete if possible.

jychp (approved):

PR #3031, opened 2026-07-14
fix(databricks): retry transient API failures

kunaals:
### Type of change

### Summary

Databricks enforces per-endpoint, per-workspace REST API rate limits and returns HTTP 429 when a client exceeds them. The shared Cartography Databricks client previously attempted each request only once, so a transient rate-limit response aborted the full sync.

Configure bounded retries for idempotent GET requests and OAuth client credentials token requests in the shared workspace/account client. The policy uses exponential backoff for 429 and transient 5xx responses, honors Retry-After when Databricks supplies it, and still propagates failures after retry exhaustion so partial ingestions cannot trigger unsafe cleanup.

### Related issues or links

- [Databricks REST API rate limits](https://docs.databricks.com/api/gcp/workspace/introduction#rate-limits)
- [Databricks SDK retry behavior](https://github.com/databricks/databricks-sdk-py#retries)

### How was this tested?

- Added a unit test that performs a real 429 then 200 request sequence and verifies the retry count and Retry-After sleep.
- Added policy assertions covering the retry budget, backoff, status codes, and allowed methods.

#### General
- [x] The linter passes locally (make test_lint).

#### Proof of functionality

- [ ] Updated the schema documentation.

- [ ] Used the NodeSchema data model.

### Notes for reviewers

This changes shared Databricks HTTP transport behavior only; it does not change any synced entity, graph schema, or cleanup contract.

### DIFF
--- a/cartography/intel/databricks/util.py
+++ b/cartography/intel/databricks/util.py
@@ -7,6 +7,8 @@ from typing import Any
 import neo4j
 import requests
 from dateutil import parser as dateutil_parser
+from requests.adapters import HTTPAdapter
+from urllib3 import Retry
 
 logger = logging.getLogger(__name__)
 
@@ -94,6 +96,7 @@ def skip_or_raise_http(error: requests.HTTPError, *skippable_statuses: int) -> N
 # Connect and read timeouts of 60 seconds each.
 _TIMEOUT = (60, 60)
 _SCIM_PAGE_SIZE = 100
+_RETRYABLE_STATUS_CODES = (429, 500, 502, 503, 504)
 
 
 def scoped_id(workspace_id: str, scim_id: str) -> str:
@@ -136,6 +139,14 @@ class _BaseDatabricksClient:
         self._client_secret = client_secret
         self._access_token_expiry: float | None = None
         self._session = requests.Session()
+        retry_policy = Retry(
+            total=5,
+            backoff_factor=1,
+            status_forcelist=_RETRYABLE_STATUS_CODES,
+            allowed_methods=frozenset({"GET"}),
+            respect_retry_after_header=True,
+        )
+        self._session.mount("https://", HTTPAdapter(max_retries=retry_policy))
 
     def _token_url(self) -> str:
         raise NotImplementedError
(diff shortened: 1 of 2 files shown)


cartography/intel/databricks/util.py
```diff
@@ -136,6 +139,14 @@ def __init__(
         self._client_secret = client_secret
         self._access_token_expiry: float | None = None
         self._session = requests.Session()
+        retry_policy = Retry(
+            total=5,
+            backoff_factor=1,
+            status_forcelist=_RETRYABLE_STATUS_CODES,
+            allowed_methods=frozenset({"GET"}),
```
jychp: The OAuth token fetch in `authenticate()` (line 161) is a POST, so `allowed_methods={"GET"}` excludes it from retries. That POST is the first request of every sync and Databricks rate-limits token endpoints too; a single 429 there aborts the whole sync, the exact failure this PR sets out to prevent. The `client_credentials` grant is safe to retry: add `POST` to `allowed_methods`, or retry the token request explicitly.

tests/unit/cartography/intel/databricks/test_util.py
```diff
@@ -1,9 +1,28 @@
...
+    client = DatabricksWorkspaceClient(
+        "https://example.cloud.databricks.com",
+        token="test-token",
+    )
+
+    # Act
+    retry_policy = client._session.adapters["https://"].max_retries
+
+    # Assert
+    assert retry_policy.total == 5
```
jychp, line 44: This test only asserts the constructor stored the values it was given; it never exercises a retry and would pass even if retries never fired. The PR body claims coverage of "backoff, status codes, GET-only behavior, and Retry-After handling" but none are exercised. Mount a mocked adapter / `requests_mock` that returns 429 then 200 and assert the GET is retried and `Retry-After` honored, or drop the overstated claim.

jychp (approved):
approved with nits

cartography/intel/databricks/util.py
kunaals: good catch. i added post to the retry allowlist so oauth token requests use the same bounded retry policy.

tests/unit/cartography/intel/databricks/test_util.py
kunaals: agreed. i added a real 429 then 200 request test that checks the retry count and retry-after sleep.

PR #3032, opened 2026-07-15
feat(models): make extra node labels declarative

jychp:
### Type of change
- [x] Documentation update

### Summary
- Introduce declarative, documented `ExtraNodeLabel` subclasses for unconditional, conditional, ontology, and provider-local shared labels.
- Reject raw strings and direct base-class construction while preserving generated Cypher behavior.
- Migrate all model declarations and add build-time validation for conditional fields without changing `CartographyNodeSchema`.

### Breaking changes
Cartography API Library: `ExtraNodeLabels` now accepts only `ExtraNodeLabel` subclass instances. Raw strings and the removed `ConditionalNodeLabel` type must be replaced with named declarative label classes.

### How was this tested?
- `make test_unit`
- `make test_lint`
- `uv run --frozen pytest -q tests/integration/cartography/graph/test_querybuilder_conditional_labels.py tests/integration/cartography/graph/test_querybuilder_labels_and_var_num_rels.py`

#### General

#### Proof of functionality

### Notes for reviewers
DataModel introspection, generated shared-label documentation, and ontology mapping-registry reconciliation are intentionally deferred until the introspection branch is integrated.


cartography/models/core/nodes.py
```diff
@@ -69,40 +68,27 @@ def __post_init__(self):
...
-        - All conditions must be met for the label to be applied (AND logic)
-        - The query generated is: MATCH (n:<primary_label> {field: value, ...}) SET n:<conditional_label>
+        label: The Neo4j label name.
+        conditions: Node property names and exact string values that must match.
+        ontology: Whether this is a cross-provider ontology label.
     """
 
-    label: str
-    conditions: dict[str, str]
+    label: str = field(init=False)
```
kunaals: could we keep definition metadata on the class and conditions on the instance? right now each subclass redeclares `label` as a dataclass field, so `ImageOntologyLabel({"type": "image"})` binds the dict to `label`, and callers can override a supposedly fixed label with `label="anything"`.

something like this preserves introspection while making the contract explicit:

```python
@dataclass(frozen=True)
class ExtraNodeLabel:
    conditions: dict[str, str] = field(default_factory=dict)
    label: ClassVar[str]
    kind: ClassVar[LabelKind]


class ImageOntologyLabel(ExtraNodeLabel):
    """a concrete single-platform container image."""

    label = "Image"
    kind = LabelKind.ONTOLOGY
```

`kind` also gives the introspector room for ontology, interface, canonical, and compatibility labels instead of a bool it'll probably outgrow.

cartography/models/aws/ecr/image.py
```diff
@@ -186,19 +189,10 @@ class ECRImageBaseSchema(CartographyNodeSchema):
...
-            ConditionalNodeLabel(
-                label="ImageAttestation",
-                conditions={"type": "attestation"},
-            ),
-            ConditionalNodeLabel(
-                label="ImageManifestList",
-                conditions={"type": "manifest_list"},
-            ),
+            LegacyECRImageLabel(),
+            ImageOntologyLabel(conditions={"type": "image"}),
```
kunaals: could we export canonical instances and add a small helper for conditions? we can keep the named classes and docstrings for introspection without making every call site construct them directly:

```python
IMAGE = ImageOntologyLabel()

# at the schema
IMAGE.when(type="image")
```

```python
def when(self, **conditions: str) -> Self:
    return replace(self, conditions=conditions)
```

this feels a little more unix-y to me: declarations stay plain data, and one small mechanism handles composition.

cartography/models/aws/extra_labels.py
```diff
@@ -0,0 +1,787 @@
...
+
+@dataclass(frozen=True)
+class LoadBalancerV2Label(ExtraNodeLabel):
+    """A aws node participating in the shared LoadBalancerV2 graph interface."""
+
+    label: str = "LoadBalancerV2"
+
+
+@dataclass(frozen=True)
+class LegacyACMCertificateLabel(ExtraNodeLabel):
```
kunaals: could the compatibility-only labels be values rather than one type each? there are ~100 mechanically identical `Legacy*Label` classes here, and class identity doesn't seem to add much for aliases.

```python
LEGACY_ACM_CERTIFICATE = CompatibilityLabel(
    label="ACMCertificate",
    remove_in="v1.0.0",
)
```

the semantic labels can still keep named classes and docstrings; this just moves repetitive migration data out of the type hierarchy.

cartography/models/core/nodes.py
```diff
@@ -113,28 +99,19 @@ class ExtraNodeLabels:
...
-        >>> extra_labels = ExtraNodeLabels([
-        ...     'Resource',
-        ...     'AWSResource',
-        ...     ConditionalNodeLabel(label='Critical', conditions={'severity': 'high'}),
-        ... ])
+        labels: Declarative labels to apply to the node.
     """
 
-    labels: list[Union[str, ConditionalNodeLabel]]
+    labels: list[ExtraNodeLabel]
```
kunaals: could we make these collections actually immutable while we're here? `frozen=True` prevents rebinding, but `labels.append(...)` and mutations to `conditions` still work. tuples, or a small frozen condition value, would make introspection output deterministic and leave the door open to safely caching compiled query fragments later.

.agents/skills/add-node-type/references/advanced-properties.md
```diff
@@ -22,9 +25,9 @@ class ECRImageSchema(CartographyNodeSchema):
     properties: ECRImageNodeProperties = ECRImageNodeProperties()
     sub_resource_relationship: ECRImageToAccountRel = ECRImageToAccountRel()
     extra_node_labels: ExtraNodeLabels = ExtraNodeLabels([
-        ConditionalNodeLabel(label="Image",              conditions={"type": "IMAGE"}),
-        ConditionalNodeLabel(label="ImageAttestation",   conditions={"type": "IMAGE_ATTESTATION"}),
-        ConditionalNodeLabel(label="ImageManifestList",  conditions={"type": "IMAGE_MANIFEST_LIST"}),
+        ImageOntologyLabel(conditions={"type": "IMAGE"}),
```
kunaals: these examples have drifted from the actual ecr model. the transform stores `image`, `attestation`, and `manifest_list`, and conditional matching uses exact equality, so these uppercase conditions won't match `AWSECRImage.type`. can we update this block, the table/prose below, and the duplicate example in `writing-intel-modules.md` to use the values already used by the production schemas?

```suggestion
        ImageOntologyLabel(conditions={"type": "image"}),
        ImageAttestationOntologyLabel(conditions={"type": "attestation"}),
        ImageManifestListOntologyLabel(conditions={"type": "manifest_list"}),
```

kunaals (changes requested):

cartography/models/core/nodes.py
kunaals: to make the direction clearer, i don’t think we necessarily need one subclass per label. a single frozen value type plus exported constants would keep everything available for import-time introspection while removing most of the class ceremony:

```python
class LabelKind(StrEnum):
    ONTOLOGY = "ontology"
    INTERFACE = "interface"
    COMPATIBILITY = "compatibility"
    CANONICAL = "canonical"


@dataclass(frozen=True, slots=True)
class ExtraNodeLabel:
    label: str
    description: str
    kind: LabelKind
    conditions: tuple[tuple[str, str], ...] = ()
    remove_in: str | None = None

    def when(self, **conditions: str) -> Self:
        return replace(self, conditions=tuple(sorted(conditions.items())))


IMAGE = ExtraNodeLabel(
    label="Image",
    description="a concrete single-platform container image.",
    kind=LabelKind.ONTOLOGY,
)
```

schemas would then compose IMAGE or IMAGE.when(type="image"). this keeps metadata explicit and immutable, avoids runtime discovery, and leaves the query builder as the one small mechanism consuming it.

kunaals (approved):

cartography/models/extra_labels.py
```diff
@@ -0,0 +1,54 @@
+from cartography.models.core.nodes import ExtraNodeLabel
+
```
achantavy, line 2: Nit: I feel like the linter should have caught that this only has 1 line break; I thought that we required imports and code to have 2 spaces in between.

docs/root/dev/writing-intel-modules.md
```diff
@@ -147,9 +147,38 @@ See [below](#indexescypher) for more information on indexes.
 
 #### Extra node labels
 
-You can add additional Neo4j labels to your nodes using `ExtraNodeLabels`:
+`ExtraNodeLabel` is a single immutable value type. Define labels once as
```
achantavy: Rewrite this in the form of what the developer needs. Instead of saying what it is, keep the form of "You can add additional Neo4j labels to your nodes using `ExtraNodeLabels`:" but then type in how the developer is supposed to implement it.

Explain a small bit on what this is for so the developer is not confused.

cartography/models/aws/extra_labels.py
```diff
@@ -0,0 +1,883 @@
...
+    kind=LabelKind.COMPATIBILITY,
+    remove_in="v1.0.0",
+)
+
+
+LEGACY_CLOUD_WATCH_LOG_GROUP = ExtraNodeLabel(
+    label="CloudWatchLogGroup",
+    description="Compatibility label for the deprecated `CloudWatchLogGroup` aws node label.",
+    kind=LabelKind.COMPATIBILITY,
+    remove_in="v1.0.0",
```
achantavy: The convention of the project for semver strings is `n.n.n`, not `vn.n.n`. I think we should keep this consistent.

cartography/models/aws/extra_labels.py
```diff
@@ -0,0 +1,883 @@
...
+    label="CloudTrailTrail",
+    description="Compatibility label for the deprecated `CloudTrailTrail` aws node label.",
+    kind=LabelKind.COMPATIBILITY,
+    remove_in="v1.0.0",
+)
+
+
+LEGACY_CLOUD_WATCH_LOG_GROUP = ExtraNodeLabel(
+    label="CloudWatchLogGroup",
+    description="Compatibility label for the deprecated `CloudWatchLogGroup` aws node label.",
```
achantavy, line 198: Someone skimming the code might get the incorrect impression that we no longer now support CloudWatchLogGroups. Would it be useful to store the label this got replaced with, if available?

cartography/graph/querybuilder.py
```diff
@@ -429,7 +428,8 @@ def _build_node_properties_statement(
 
     Args:
         node_property_map (Dict[str, PropertyRef]): Mapping of node attribute names as str to PropertyRef objects.
-        extra_node_labels (Optional[ExtraNodeLabels], optional): ExtraNodeLabels object to set on the node as string.
+        extra_node_labels (Optional[ExtraNodeLabels], optional): Declarative labels
+            to apply unconditionally when their conditions are empty.
```
achantavy: > to apply unconditionally when their conditions are empty.

this is a very confusing phrase. I'd try to keep it to an example to make concrete

cartography/graph/querybuilder.py
```diff
@@ -447,11 +447,8 @@ def _build_node_properties_statement(
         >>> # i.node_prop_2 = $Prop2
         >>> # (note: 'id' is excluded as it's handled by MERGE)
 
-        >>> # With extra labels
-        >>> extra_labels = ExtraNodeLabels(['Resource', 'CloudAsset'])
-        >>> set_clause = _build_node_properties_statement(node_property_map, extra_labels)
-        >>> # Returns the property assignments plus:
-        >>> # i:Resource:CloudAsset
+        With `RESOURCE` and `CLOUD_ASSET` in `extra_node_labels`,
+        the returned property assignments also contain `i:Resource:CloudAsset`.
```
achantavy: this is a confusing phrase also. This docstring has made the effort to show the REPL's journey up to this point: keep it concrete. I would show the literal value of `set_clause` instead of saying "returns the property assignments"

cartography/graph/querybuilder.py
```diff
@@ -472,14 +469,13 @@ def _build_node_properties_statement(
         ],
     )
 
-    # Set extra labels on the node if specified (excluding conditional labels)
+    # Set unconditional extra labels on the node.
     if extra_node_labels:
-        # Filter out ConditionalNodeLabel objects - only include string labels
-        string_labels = [
-            label for label in extra_node_labels.labels if isinstance(label, str)
+        unconditional_labels = [
```
achantavy: I would not name this `unconditional`. I think it just makes it more complicated.

I think if there's a variable called `string_labels` and then `conditional_labels`, I'm able to differentiate.

cartography/graph/querybuilder.py
```diff
@@ -1228,8 +1224,8 @@ def build_conditional_label_queries(
     Generate Neo4j queries to apply conditional labels to nodes.
 
     Conditional labels are labels that are only applied to nodes matching specific conditions.
-    This function generates one query per ConditionalNodeLabel defined in the node schema's
-    extra_node_labels.
+    This function generates remove and set queries for each extra label with
+    nonempty conditions.
```
achantavy: Probably could remove this sentence altogether. "Generate neo4j queries to apply conditional labels to nodes" is complete enough I think.

docs/root/dev/writing-intel-modules.md
```diff
@@ -147,9 +147,38 @@ See [below](#indexescypher) for more information on indexes.
...
 
 ```python
+from cartography.models.core.nodes import ExtraNodeLabel, LabelKind
+
+# cartography/models/aws/extra_labels.py
+AWS_RESOURCE = ExtraNodeLabel(
+    label="AWSResource",
+    description="A resource managed within an AWS account.",
+)
+LEGACY_ECR_IMAGE = ExtraNodeLabel(
```
achantavy: a compatibility label deserves its own section of explanation about why these exist and when they will go away. right now it's jumbled in here and listed all out like it's written for an agent and not a person

docs/root/dev/writing-intel-modules.md
```diff
@@ -147,9 +147,38 @@ See [below](#indexescypher) for more information on indexes.
...
+uppercase constants in the appropriate label module, export them, and reuse
+those constants in node schemas. Do not define one subclass per label or rely
+on class docstrings.
 
 ```python
+from cartography.models.core.nodes import ExtraNodeLabel, LabelKind
+
+# cartography/models/aws/extra_labels.py
+AWS_RESOURCE = ExtraNodeLabel(
+    label="AWSResource",
```
achantavy: Use a real example instead of writing in the abstract. This is currently confusing because a developer may think that it's necessary to declare this for all resources in AWS.

docs/root/dev/writing-intel-modules.md
```diff
@@ -159,24 +188,35 @@ class EMRClusterSchema(CartographyNodeSchema):
...
 #### Conditional node labels
 
 ```{warning}
 Conditional labels are a specialized feature primarily used for ontology mapping scenarios where a single data source produces records that map to different semantic types. Most intel modules do not need this feature.
 ```
 
-Sometimes you want to apply labels only when certain conditions are met. Use `ConditionalNodeLabel` for this:
+Sometimes you want to apply labels only when certain conditions are met.
+Compose a conditional value from an exported constant with
+`CONSTANT.when(field="value")`:
```
achantavy: Use a concrete example and link to a permalink in the code to describe what is happening. A developer may incorrectly think that the kwarg name is literally "field" when in fact it is actually any key that is present in the data dict.

Say that this is based on the properties set in the data dict being processed by the loader.

cartography/models/core/nodes.py
```diff
@@ -68,73 +70,88 @@ def __post_init__(self):
...
-        >>> # Simple string labels (applied to all nodes)
-        >>> extra_labels = ExtraNodeLabels(['Resource', 'AWSResource'])
-
-        >>> # Mix of simple and conditional labels
-        >>> extra_labels = ExtraNodeLabels([
-        ...     'Resource',
-        ...     'AWSResource',
-        ...     ConditionalNodeLabel(label='Critical', conditions={'severity': 'high'}),
-        ... ])
+        labels: Declarative labels to apply to the node.
```
achantavy: This docstring lost the types

cartography/models/core/nodes.py
```diff
@@ -68,73 +70,88 @@ def __post_init__(self):
...
-    A conditional label that is applied to nodes only when specific field conditions are met.
+class LabelKind(str, Enum):
+    """Semantic role of an additional Cartography node label."""
 
-    Conditional labels allow you to dynamically apply labels to nodes based on their property values.
-    During ingestion, after the main node creation/update, a separate query is run to match nodes
-    that meet the specified conditions and apply the label.
+    STANDARD = "standard"
+    ONTOLOGY = "ontology"
+    COMPATIBILITY = "compatibility"
```
achantavy, line 85: Since this gets stored at https://docs.cartography.dev/references/model.html#cartography.models.core.nodes.ConditionalNodeLabel, we should include docstrings on the meaning behind each value.

achantavy (changes requested):
I'm supportive of the change. Left comments on documentation. 


Also this is not a refactor; it is a feature that breaks existing interfaces e.g. `cartography.models.core.nodes.ConditionalNodeLabel` is getting deprecated in favor of the new interface.

achantavy (approved):

PR #3033, opened 2026-07-15
fix(gcp): route Cloud Asset Inventory calls through the standard retry wrapper

vidigoat:
### Type of change

### Summary

The `cai` (Cloud Asset Inventory) module was the only GCP intel module still calling `request.execute(num_retries=...)` directly instead of the shared `gcp_api_execute_with_retry()` helper that the other 26 GCP modules use.

Because of that, its Asset Inventory list calls missed the standard retryable-error handling every other GCP module gets:
- exponential backoff on HTTP 5xx / 429,
- retry of the legacy `403 rateLimitExceeded` / `userRateLimitExceeded` responses that older GCP APIs return, classified via `is_retryable_gcp_http_error`,
- the common give-up / backoff logging.

Asset Inventory is rate-limit-prone (100 requests/min per project), so a transient error there aborted the whole sync instead of backing off and retrying.

This routes both `cai` getters through `gcp_api_execute_with_retry` (which still applies `num_retries=GCP_API_NUM_RETRIES` internally, unchanged) and drops the now-unused local `GOOGLE_API_NUM_RETRIES` constant. No behavior change on the success path.

### Related issues or links

None — found while auditing retry handling across the GCP modules; `cai` was the sole holdout.

### How was this tested?

Added `tests/unit/cartography/intel/gcp/test_cai_http_errors.py` (mirroring the other `*_http_errors.py` suites): asserts that a transient `503` on the first call is retried and the getter still returns its data, and that a non-retryable `404` propagates without retrying. The retry tests fail before this change (the `503` aborts the call) and pass after. Full `tests/unit/cartography/intel/gcp/` suite passes (327 passed). `black`, `isort`, `flake8`, `pyupgrade --py36-plus`, and `mypy` are clean on both changed files.

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/gcp/cai.py
+++ b/cartography/intel/gcp/cai.py
@@ -9,14 +9,13 @@ from googleapiclient.discovery import Resource
 
 from cartography.client.core.tx import load
 from cartography.intel.gcp.util import determine_role_type_and_scope
+from cartography.intel.gcp.util import gcp_api_execute_with_retry
 from cartography.models.gcp.iam import GCPProjectRoleSchema
 from cartography.models.gcp.iam import GCPServiceAccountSchema
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
 
-# Maximum number of retries for Google API requests (handles transient errors and rate limiting)
-GOOGLE_API_NUM_RETRIES = 5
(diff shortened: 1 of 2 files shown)


jychp (approved):
Good catch, thx for the fix.

PR #3039, opened 2026-07-17
feat(cli): add --log-timestamps to prepend timestamps to log output

### LINKED ISSUE #1213: Feature Request: Add time to logs/output

vidigoat:
### Type of change
- [ ] Documentation update

### Summary
Fixes #1213.

Adds an opt-in `--log-timestamps` flag so terminal users can see when each log line was emitted (and therefore how long individual sync stages take):

```
$ cartography --log-timestamps ...
2026-07-17T17:12:18+0530 INFO cartography.sync: Starting sync with update tag '1784288538'
2026-07-17T17:12:18+0530 INFO cartography.sync: Starting sync stage 'create-indexes'
```

The default format is untouched, per the discussion on the issue: log aggregators attach their own timestamp field to each record, so embedding a second timestamp in the message would be redundant there. The flag simply installs a `logging.Formatter` (ISO-8601 timestamp, level, logger name) on the root logger's handlers, so it composes with `--quiet`/`--debug` and touches nothing when unset.

### Related issues or links
- #1213

### How was this tested?
- New unit tests in `tests/unit/cartography/test_cli.py` covering the formatter's output shape and the untouched-by-default behavior.
- Ran the CLI live with and without the flag against an unreachable Neo4j: with the flag every line is prefixed `2026-07-17T17:12:18+0530 INFO cartography.sync: ...`; without it the output is byte-identical to master (`INFO:cartography.sync:...`).
- Full unit suite passes; `black`/`isort`/`flake8` clean; `mypy` reports no new errors on `cartography/cli.py` (the one pre-existing `unused-ignore` on master merely shifts line numbers).

#### General

#### Proof of functionality


tests/unit/cartography/test_cli.py
```diff
@@ -0,0 +1,51 @@
+import io
+import logging
+import re
+
+from cartography.cli import _apply_timestamp_log_format
+
+
+def test_apply_timestamp_log_format_prepends_timestamp_and_level():
```
achantavy: I would delete this test. We don't need any unit tests for this feature.

achantavy (changes requested):
Hi, thanks for taking this on. 

I don't think this feature needs unit tests. (Additionally the one currently in this PR is also pretty sloppy: it's got try+excepts like it's app code :\)

Anyways, to merge, I need to see screenshots of a successful sync run where the timestamps are shown.

vidigoat:
Thanks for the review — removed the unit tests. Here's a successful sync run with the timestamps (create-indexes + analysis against a local neo4j):

successful sync run with timestamps

achantavy (approved):

PR #3046, opened 2026-07-17
fix(databricks): tolerate HTTP 400 on ungrantable securables in grants sync

jychp:
### Type of change

### Summary

The grants stage aborted the **entire** databricks sync when the Unity Catalog permissions endpoint returned **400 Bad Request** for a system-managed securable (for example a `registered_model` under the built-in `system` catalog, such as `system.ai.*`).

The module intentionally ingests the Databricks-provided `system` catalog (its `system.ai` schemas return 200, so their registered models are loaded as nodes). `grants.sync` then reads every securable back and calls the per-securable permissions endpoint; Databricks rejects grant-listing on a system-managed registered model with a 400. `grants.get` only tolerated 403/404 via `skip_or_raise_http`, so the 400 re-raised and aborted the whole sync.

**Fix:** tolerate 400 across the per-object "read ACL / grants" family that iterates a securable/object set which can contain structurally-ineligible objects:

- `grants.py` `get()` (the crash) — `/api/2.1/unity-catalog/permissions/{type}/{full_name}`
- `permissions.py` `get()` — `/api/2.0/permissions/{object_type}/{object_ref}`
- `permissions.py` `get_secret_scope_acls()` — `/api/2.0/secrets/acls/list`
- `shares.py` per-share probe — `/api/2.1/unity-catalog/shares/{name}/permissions`

This mirrors `online_tables.py` / `dashboards.py`, which already treat 400 as "this object is not eligible for this probe". The list-driven probes (`pipelines`, `workspace_assignments`, `federation_policies`, `artifact_allowlists`) stay strict: they iterate lists the API itself returned, so a 400 there is a malformed request, not an ineligible object, and must still abort.

**Cleanup safety:** the whole-workspace `HAS_PRIVILEGE` cleanup previously ran unconditionally, so a skipped securable's grants would be deleted as stale. `grants.get()` now returns `(grants, complete)` and the central grants cleanup is gated on it (new `clean_grants` flag on `_cleanup_unity_catalog`, matching the existing `clean_artifact_allowlists` / `clean_clean_rooms` pattern). The read stays **complete only** for a securable known to be structurally non-grantable (a `registered_model` in the `system` catalog), which by definition has no `HAS_PRIVILEGE` edges. Every other skip, a 403/404, or an unexpected 400 on a securable that may already hold grants, flags the read incomplete so cleanup is skipped rather than deleting still-valid edges. All three statuses remain skippable, so an unexpected 400 never crashes the sync; it just conservatively inhibits cleanup for that run.

### How was this tested?

- Added unit-style tests (mirroring `test_ip_access_lists.py`) that drive a mocked HTTP layer against the real functions:
  - `grants`: 400 on a `system` registered_model is skipped and keeps the read complete; a 400 on a real catalog and a 403 both flag the read incomplete; a 500 propagates.
  - `permissions.get` / `get_secret_scope_acls`: a 400 object is skipped, `complete` is `False`, remaining objects still load; a 500 propagates.
  - `shares._recipient_names`: a 400 skips and carries forward last-known recipients from the graph; a 500 propagates.
- `make test_lint` passes (black, isort, flake8, mypy, pyupgrade).
- The 400-skip/complete-flag logic was additionally exercised against the real functions locally (Docker/Neo4j was unavailable in this environment, so the full integration suite was not run end-to-end here).

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/databricks/permissions.py
+++ b/cartography/intel/databricks/permissions.py
@@ -203,11 +203,10 @@ def get(
         try:
             response = api_session.get(uri)
         except requests.HTTPError as e:
-            # An object the caller can't read the ACL on (403), that vanished
-            # mid-sync (404), or that is structurally ineligible for this probe
-            # (400) is skippable; any other error must abort so the permission
-            # cleanup does not drop still-valid HAS_PERMISSION edges.
-            skip_or_raise_http(e, 400, 403, 404)
+            # An object the caller can't read the ACL on (403) or that vanished
+            # mid-sync (404) is skippable; any other error must abort so the
+            # permission cleanup does not drop still-valid HAS_PERMISSION edges.
+            skip_or_raise_http(e, 403, 404)
             complete = False
             logger.warning(
                 "Skipping permissions for %s %s: %s",
@@ -268,10 +267,7 @@ def get_secret_scope_acls(
                 "/api/2.0/secrets/acls/list", params={"scope": scope["name"]}
             )
         except requests.HTTPError as e:
-            # Skip a scope whose ACL is forbidden (403), gone (404), or
-            # structurally ineligible (400); any other error must abort so the
-            # permission cleanup does not drop still-valid HAS_PERMISSION edges.
-            skip_or_raise_http(e, 400, 403, 404)
+            skip_or_raise_http(e, 403, 404)
             complete = False
             logger.warning("Skipping secret scope ACL for %s: %s", scope["name"], e)
             continue
--- a/cartography/intel/databricks/permissions.py
+++ b/cartography/intel/databricks/permissions.py
@@ -203,10 +203,11 @@ def get(
         try:
             response = api_session.get(uri)
         except requests.HTTPError as e:
-            # An object the caller can't read the ACL on (403) or that vanished
-            # mid-sync (404) is skippable; any other error must abort so the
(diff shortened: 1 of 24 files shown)


cartography/intel/databricks/permissions.py
```diff
@@ -203,10 +203,11 @@ def get(
...
         except requests.HTTPError as e:
-            # An object the caller can't read the ACL on (403) or that vanished
-            # mid-sync (404) is skippable; any other error must abort so the
-            # permission cleanup does not drop still-valid HAS_PERMISSION edges.
-            skip_or_raise_http(e, 403, 404)
+            # An object the caller can't read the ACL on (403), that vanished
+            # mid-sync (404), or that is structurally ineligible for this probe
+            # (400) is skippable; any other error must abort so the permission
+            # cleanup does not drop still-valid HAS_PERMISSION edges.
+            skip_or_raise_http(e, 400, 403, 404)
```
kunaals: do we know that 400 is always a structurally ineligible object across these other ACL probes? databricks also documents 400 as `BAD_REQUEST`, so a malformed path or API contract change could get swallowed and leave cleanup disabled. might be safer to keep 400 scoped to the known system registered-model case (or classify a concrete error code/message) unless we've observed it on these endpoints too

kunaals (approved):
approving to unblock, one small thing to consider

PR #3050, opened 2026-07-19
docs: add entra, tenable, and ubuntu to the module menu

vidigoat:
### Type of change
- [x] Documentation update

### Summary

Follow-up to #3045: three more modules are missing from the docs module menu.

- **entra** already has full documentation (config, examples, schema) but is unreachable from the navigation
- **tenable** and **ubuntu** only have schema docs; this adds minimal glob index pages (same pattern as jamf) so their existing docs are reachable

Also raised in the CNCF Slack channel alongside a companion PR documenting missing CLI flags.

### Related issues or links

- Follow-up to #3045

### How was this tested?

- Full Sphinx docs build (`uv run --group doc ./docs/build.sh`, exit 0) and verified `modules/entra/index.html`, `modules/tenable/index.html`, and `modules/ubuntu/index.html` are generated and linked from the menu
- pre-commit hooks pass on changed files

### Checklist
- [x] I have updated relevant documentation

PR #3051, opened 2026-07-19
docs: document all user-visible CLI flags and add drift test

vidigoat:
### Type of change
- [x] Documentation update

### Summary

While going through the CLI docs I found that 25 user-visible CLI flags are not documented anywhere on the docs site, and two config pages instruct flags that don't exist. This PR (announced in the CNCF Slack channel):

**Fixes incorrect docs:**
- bigfix: the config page instructed `--bigfix_password_env_var` (underscores — not a valid flag) and didn't mention `--bigfix-root-url`
- workos: the config page referenced `--workos-api-key`; the actual flag is `--workos-apikey-env-var`

**Fills gaps:**
- jamf and ubuntu had no config page at all; adds them
- GCP selective syncing (`--gcp-requested-syncs`) was undocumented while the AWS equivalent has a full section; adds a matching section
- Documents the remaining missing flags on their module config pages (aws, azure, cve, gitlab, oci, slack) and the core/Neo4j flags (`--debug`, `--version`, `--neo4j-max-connection-lifetime`, `--neo4j-liveness-check-timeout`) on the CLI usage page — each description sourced from the flag's own `help=` text in `cli.py`

**Keeps it fixed:**
- Adds `test_cli_doc` alongside the existing `test_schema_doc`: every non-hidden CLI flag must appear somewhere under `docs/root`. Flags declared `hidden=True` and `--experimental-*` flags are excluded. The failure message names the missing flags and where to document them.

### Related issues or links

- Companion to #3050 (module menu completion)

### How was this tested?

- `pytest tests/unit` — 3115 passed
- Verified the new test catches drift: removing a flag's doc line makes it fail with an actionable message naming that flag
- Full Sphinx docs build (`uv run --group doc ./docs/build.sh`, exit 0); verified `modules/jamf/config.html` and `modules/ubuntu/config.html` are generated and the GCP selective-sync section renders
- All pre-commit hooks pass on changed files (black, flake8, mypy, isort, etc.)

### Checklist
- [x] I have added tests for new functionality
- [x] I have updated relevant documentation


docs/root/modules/ubuntu/config.md
```diff
@@ -0,0 +1,6 @@
+## Ubuntu Security Configuration
```
kunaals: please add this page to ubuntu's toctree. it currently builds as an orphan and adds a new toc.not_included warning; #3050's explicit ubuntu index lists only schema.

tests/unit/cartography/test_doc.py
```diff
@@ -25,3 +26,42 @@ def test_schema_doc():
...
+
+def test_cli_doc():
+    """Test that every user-visible CLI flag is documented.
+
+    This test checks that each non-hidden CLI flag defined in cartography/cli.py
+    appears somewhere in the docs, ensuring that the documentation is up-to-date
+    with the current CLI surface. Flags declared with `hidden=True` and
+    experimental flags are excluded.
+    """
+    flag_regex = re.compile(r'"(--[a-z0-9-]+)"')
```
kunaals: please use typer's option registry instead of parsing cli.py source. this misses the generated --install-completion and --show-completion options and treats a --help mention as a declared option.

docs/root/usage/cli.md
```diff
@@ -64,11 +64,12 @@ Core, Neo4j, StatsD, and Analysis panels are always shown regardless of selected
 
 | Option | Description |
 |--------|-------------|
-| `-v, --verbose` | Enable debug logging |
+| `--debug` | Enable verbose logging (`-v, --verbose` is deprecated and will be removed in v1.0.0) |
```
kunaals: please include -d here; it is the supported short alias for --debug, while -v and --verbose are deprecated.

kunaals (changes requested):

vidigoat:
Thanks @kunaals — all three addressed:
- ubuntu/config.md now has a toctree (config + schema), so no more toc.not_included warning.
- Documented `-d` as the supported short alias for `--debug` in cli.md.
- Reworked test_cli_doc to introspect the Typer/Click command registry instead of regex-parsing cli.py, so it reflects the real CLI surface (including the generated --install-completion/--show-completion) and uses each option's actual hidden state. Test passes.

vidigoat:
Good catches, thanks — both fixed:
- Added `modules/ubuntu/index` to the root toctree (between trivy and vercel), so it's no longer an orphan.
- Fixed test_cli_doc to build the app with every panel visible (`MODULE_PANELS.values() | ALWAYS_SHOW_PANELS`). With an empty set every module flag was hidden and silently skipped — now the test actually checks all of them (it still passes, so the module flags are all documented).

vidigoat:
@kunaals I've brought this up to date with master. All three of your review points are addressed (ubuntu toctree, `-d` alias in cli.md, and `test_cli_doc` now introspects the Typer/Click registry with every panel visible). Could you take another look when you have a moment?

PR #3054, opened 2026-07-20
feat(ontology): materialize (:ComputeService)-[:RUNS_IMAGE]->(:Image) inventory

jychp:
### Type of change
- [x] Documentation update

### Summary

Materialize the runtime image inventory of each logical workload as a first-class ontology edge:

```
(:ComputeService)-[:RUNS_IMAGE]->(:Image)
```

This is a standalone runtime-composition fact ("this workload runs this image"), useful for inventory, attack-path, and "what is actually running on this service" queries. It is bounded by the number of `(workload, image)` pairs, not by the replica fan-out, so consumers can answer "which workloads run an affected image" through a bounded pattern without scanning every container replica.

A new typed analysis job (`WORKLOAD_RUNS_IMAGE`) runs in the ontology module after `RESOLVED_IMAGE_JOBS`. It collapses each running container up its `WORKLOAD_PARENT` chain to the owning `ComputeService` and `MERGE`s one edge per `(workload, image)`. The typed framework's `MERGE` plus generated mark-gone cleanup (`DELETE r WHERE r.lastupdated <> $UPDATE_TAG`) make it idempotent across incremental syncs.

Design points:
- **Zero-or-more-hop collapse (`*0..6`)** also covers serverless workloads that carry both `:ComputeService` and `:Container` on a single node (e.g. `ScalewayServerlessContainer`), which have no intermediate `WORKLOAD_PARENT` hop.
- **Active-state filter** accepts both `running` (AWS ECS `RUNNING`, Kubernetes `running`, Azure `Running`, GCP Cloud Run) and `ready` (Scaleway serverless `ContainerStatus.READY`), because `_ont_state` carries the raw per-provider status string.
- **`exposed_internet`** is denormalized onto the edge as the OR of the service-level signal (`svc.exposed_internet`, e.g. GCP Cloud Run ingress) and any running replica's signal (`rt.exposed_internet`, e.g. AWS ECS / Kubernetes), so workload-level exposure is readable without fanning back out to replicas.
- **Standalone functions** (AWS Lambda, GCP Cloud Functions, Azure Function Apps, Scaleway serverless functions) carry only `:Function` with no `ComputeService` controller and are intentionally not materialized; they remain per-instance and low cardinality.

Also registers the canonical `ComputeService -> Image` `RUNS_IMAGE` `RelConstraint` and updates the ontology schema docs.

### How was this tested?

- New integration tests in `tests/integration/cartography/intel/ontology/test_runs_image.py` cover: replica collapse + dedup, `running`/`ready` state filtering, exposure denormalization (replica-level and service-level), idempotency / mark-gone across syncs, self-service (Scaleway-style) zero-hop collapse, and the no-controller / standalone-function cases (no edge).
- Governance guards pass: `tests/unit/cartography/graph/test_analysis_cleanup_coverage.py`, `tests/unit/cartography/graph/test_ontology_rel_constraints.py`.
- Full ontology integration suite passes (`tests/integration/cartography/intel/ontology/`).

#### General

#### Proof of functionality

### Notes for reviewers

`RUNS_IMAGE` only appears in environments that have `ComputeService` controllers, which for Kubernetes require the workload-controller modeling from #3053. The edge is derived (analysis-only), so like `RESOLVED_IMAGE` it is registered in `ONTOLOGY_REL_CONSTRAINTS` for documentation but is not enforced against a declarative schema.

### DIFF
--- a/cartography/analysis/ontology/analysis.py
+++ b/cartography/analysis/ontology/analysis.py
@@ -622,6 +622,58 @@ FUNCTION_RESOLVED_IMAGE = AnalysisJob(
     ),
 )
 RESOLVED_IMAGE_JOBS = (CONTAINER_RESOLVED_IMAGE, FUNCTION_RESOLVED_IMAGE)
+WORKLOAD_RUNS_IMAGE = AnalysisJob(
+    name="Workload RUNS_IMAGE inventory analysis",
+    short_name="workload_runs_image_analysis",
+    cleanup_iterationsize=1000,
+    statements=(
+        AnalysisStatement(
+            comment=(
+                "Materialize the runtime image inventory of each ComputeService "
+                "workload, deduped per (workload, image), with workload-level internet "
+                "exposure denormalized onto the edge. Anchoring on ComputeService means "
+                "this covers workloads that run containers: ECS services, Cloud Run "
+                "services/jobs, Kubernetes controllers, and Scaleway serverless "
+                "containers. Standalone runtimes with no ComputeService controller "
+                "(bare pods, and functions such as AWS Lambda / GCP Cloud Functions / "
+                "Azure Function Apps / Scaleway serverless functions, which carry only "
+                ":Function) are intentionally NOT materialized here; they are per-instance "
+                "and low cardinality, and stay on the read-side live-collapse path. The "
+                "rt:Function branch therefore matches nothing today and exists only to "
+                "cover a future serverless function that is itself modeled as a "
+                "ComputeService (the symmetric case of Scaleway serverless containers). "
+                "The *0..6 lower bound of 0 covers serverless workloads "
+                "(e.g. ScalewayServerlessContainer) that carry both :ComputeService and "
+                ":Container on a single node with no intermediate WORKLOAD_PARENT hop. "
+                "Exposure is the OR of the service-level signal (svc.exposed_internet, "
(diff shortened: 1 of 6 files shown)


cartography/analysis/ontology/analysis.py
```diff
@@ -622,6 +622,58 @@
...
+            MATCH (svc:ComputeService)<-[:WORKLOAD_PARENT*0..6]-(rt)-[:RESOLVED_IMAGE]->(img:Image)
+            WHERE (rt:Container OR rt:Function)
+              AND (NOT rt:Container OR toLower(rt._ont_state) IN ['running', 'ready'])
+            WITH svc, img, collect(coalesce(rt.exposed_internet, false)) AS rt_flags
+            WITH svc, img, (coalesce(svc.exposed_internet, false) OR true IN rt_flags) AS exposed
+            """,
+            effects=(
+                AddRelationship(
+                    "svc",
+                    "RUNS_IMAGE",
```
kunaals: `HAS_RUNTIME_IMAGE` feels more accurate than `RUNS_IMAGE` because this edge represents runtime composition, not consistently observed execution

kunaals (changes requested):
one small change to rel label, after that will approve

kunaals (approved):

PR #3059, opened 2026-07-21
feat(circleci): code-to-cloud fallback supply-chain matcher

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds a CircleCI supply-chain matcher that produces low-confidence, fallback code-to-cloud edges for container images that the existing ladder cannot reach.

**Why:** the current `PACKAGED_FROM` match methods are dead for images built without SLSA provenance **and** without layer history (BuildKit-no-history, Kaniko, buildpacks, squashed images): `provenance` needs `Image.source_uri`, `dockerfile_analysis` needs `created_by` layer history, and the `package_owner_repo` fallback only works for GHCR. On a generic registry (ECR, GCP AR, Docker Hub) these images get no code-to-cloud edge at all, so `(Container)-[:RESOLVED_IMAGE]->(:Image)-[:PACKAGED_FROM]->(:CodeRepository)` never resolves. The CI/CD system is the only actor that still knows "build from repo R produced this image", so we mine the CircleCI `/pipeline` feed for that link.

**What:** two new rungs, appended **below** the existing ladder, each edge carrying `match_method` + `confidence` (consumers filter at their own threshold):

| match_method | Signal | Confidence |
|---|---|---|
| `circleci_tag_revision` | An `ImageTag` name equals a `/pipeline`-feed run's `vcs.revision` (git SHA) that resolves to a single repo | medium (0.5) |
| `circleci_config_binding` | A registry namespace parsed from `/pipeline/{id}/config` binds to a repo; images under that namespace inherit it | low (0.25) |

Where a rung identifies the building project it also emits `(:Image)-[:PACKAGED_BY]->(:CircleCIProject)`.

Design invariants (mirroring the GitHub/GitLab supply-chain modules):
- **MatchLink producer, not a node loader.** Pipeline runs and config YAML are consumed transiently in memory and never persisted as nodes; only edges are written.
- **Lazy / residual-only.** The `/pipeline` feed is fetched only when unmatched candidate images exist; cross-provider gating is implicit (images already matched this run by GitHub/GitLab, which sync earlier, are excluded).
- **Time-bounded feed** (lookback window, newest-first with early-stop via a new optional `stop_when` predicate on `paginated_get`) rather than an arbitrary page cap.
- **No org-wide cleanup.** The feed is a partial enumeration (CircleCI API v2 has no list-projects endpoint), so a stale-cleanup would delete valid edges once a build ages out of the window; edges are upsert-only, removed by node-level DETACH DELETE. This mirrors the `projects` module.
- **Ambiguity is dropped, not guessed.** SHAs/short-prefixes that resolve to more than one target, and namespaces pushed to by more than one observed repo, produce no edge.

`circleci_oidc_push` (a third rung in the original design) is intentionally **deferred**: it requires registry push-event / pusher-identity ingestion (ECR CloudTrail `PutImage` data events, GCP AR audit logs) that does not exist in the graph today.

### Related issues or links

N/A

### How was this tested?

- New unit tests (pure logic, no Neo4j): revision-map construction and ambiguity dropping, tag-revision matching (exact / short-SHA prefix / ambiguous / no-match), the time-window predicate (including unparseable timestamps), registry-namespace parsing, config-destination parsing, and singleton config bindings.
- New integration tests (real Neo4j, mocked CircleCI API): `circleci_tag_revision` PACKAGED_FROM + PACKAGED_BY; cross-provider gating (an image already matched this run is excluded); no ephemeral run/config nodes created; `circleci_config_binding` PACKAGED_FROM.
- Commands:
  ```
  uv run --frozen pytest tests/unit/cartography/intel/circleci tests/integration/cartography/intel/circleci
  # 33 passed
  ```
- `pre-commit` (isort, black, flake8, pyupgrade, mypy) passes on all changed files.

#### General

#### Proof of functionality

- [ ] Included Cartography sync logs from a real environment. N/A: this module creates no nodes (MatchLink/edge producer only) and adds no new synced entity; behavior is covered by the integration tests above.

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) (`docs/root/modules/circleci/schema.md`).
- [ ] Updated the schema README (no per-edge entry needed; the new edges are documented in the CircleCI module schema).

### Notes for reviewers

- The matcher is modeled on `cartography/intel/gitlab/supply_chain.py` and reuses the shared `normalize_vcs_url` helper. `PACKAGED_FROM` uses two provider-specific target schemas (GitHub keyed on `id`, GitLab on `web_url`) to match the existing provenance matchers.
- Two soundness limits are documented in code as accepted tradeoffs: `circleci_config_binding` can mis-bind a namespace shared by a project outside the lookback window (hence its low confidence), and `tag_revision`/config uniqueness is enforced to avoid arbitrary attribution.
- `paginated_get` gains an optional `stop_when` predicate (default `None`, no behavior change for existing callers) to support the time-bounded early stop.


cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,557 @@
...
+# gives a meaningful "recent builds" window (mirrors the CloudTrail lookback pattern)
+# instead of an arbitrary page count. _MAX_FEED_PAGES is only a runaway-pagination guard.
+_FEED_LOOKBACK_DAYS = 30
+_MAX_FEED_PAGES = 100
+
+_SHA_RE = re.compile(r"^[0-9a-f]{7,40}$")
+# docker push <ref> and docker build ... -t <ref>: the two shapes that name a push
+# destination in a CircleCI config's shell steps.
+_DOCKER_PUSH_RE = re.compile(r"docker\s+push\s+(\S+)")
+_DOCKER_BUILD_TAG_RE = re.compile(r"docker\s+build\b[^\n]*?-t\s+(\S+)")
```
kunaals: `docker build -t` does not mean the image was pushed. a test-only build can bind the namespace and create `packaged_from` edges for unrelated registry images. can we require an actual push signal here, or drop the config-binding rung for now?

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,557 @@
...
+              }]->())
+          )
+        WITH img,
+             [u IN collect(DISTINCT t.uri) WHERE u IS NOT NULL] AS tag_uris,
+             collect(DISTINCT t.name) AS tags
+        // Prefer the image's own uri, else a tag reference. ECR images (AWSECRImage) carry
+        // no uri, so the ImageTag uri is what lets _registry_namespace recover the registry
+        // namespace for the config-binding rung.
+        RETURN img.digest AS digest,
+               coalesce(img.uri, head(tag_uris), img.digest) AS uri,
```
kunaals: `collect(distinct t.uri)` has no ordering, so `head(tag_uris)` picks an arbitrary registry location for an image present in multiple registries. if the bound namespace is not selected, the match is silently missed. can we keep and check all of the uris?

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,557 @@
...
+    older stragglers from the final (boundary) page. These are high-volume ephemeral CI
+    telemetry, consumed transiently in memory and never loaded as graph nodes.
+    """
+    cutoff = datetime.now(timezone.utc) - timedelta(days=lookback_days)
+    runs = paginated_get(
+        api_session,
+        f"{base_url}/pipeline",
+        params={"org-slug": org_slug},
+        max_pages=_MAX_FEED_PAGES,
+        stop_when=lambda run: _run_older_than(run, cutoff),
```
kunaals: does circleci guarantee this feed is globally newest-first across pages? i could not find that guarantee in the api contract, and without it one old item can stop pagination before newer runs on later pages. can we avoid the early stop unless the ordering is documented?

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,557 @@
...
+            # One config per project is enough for a namespace binding; the feed is
+            # newest-first, so the first run per slug is the most recent.
+            runs_by_slug: dict[str, dict[str, Any]] = {}
+            for run in runs:
+                slug = run.get("project_slug")
+                if slug and slug not in runs_by_slug:
+                    runs_by_slug[slug] = run
+            namespace_binding, repo_meta = build_config_bindings(
+                list(runs_by_slug.values()),
+                lambda pipeline_id: _get_pipeline_config(
```
kunaals: once any residual image exists, this fetches one config per recent project serially, which can add hundreds of api calls per org. can we avoid this org-wide config sweep unless there is a bounded way to target candidate projects?

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,557 @@
...
+
+    if all_matches:
+        logger.info(
+            "Loading %d CircleCI fallback PACKAGED_FROM edge(s) for org %s",
+            len(all_matches),
+            org_id,
+        )
+        _load_provider_routed_matchlinks(neo4j_session, all_matches, org_id, update_tag)
+
+    # No org-wide stale cleanup: the /pipeline feed is a partial, recency-bounded
```
kunaals: this leaves the previous circleci edge behind when a later run makes the match ambiguous or points the image at a different repo. i reproduced this across two syncs: the old edge kept its original `lastupdated` after the second run dropped the match. can we clean up edges for images whose evidence was re-evaluated without doing an org-wide cleanup?

kunaals (changes requested):

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,446 @@
...
+
+    :param neo4j_session: Neo4j session
+    :param org_id: CircleCI organization id used for scoping
+    :param update_tag: The current sync update tag
+    :param limit: Optional cap on the number of images returned
+    :return: List of {"digest", "tags": [tag name, ...]} dicts
+    """
+    query = """
+        MATCH (img:Image)<-[:IMAGE]-(t:ImageTag)<-[:REPO_IMAGE]-(repo:ContainerRegistry)
+        WHERE NOT exists((img)-[:PACKAGED_FROM {lastupdated: $update_tag}]->())
```
kunaals, line 107: i might be missing something, but could an existing github or gitlab `PACKAGED_FROM` edge cause us to skip cleanup of an older circleci edge here? i was able to get both edges to survive. would it make sense to keep these images in the cleanup scope?

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,446 @@
...
+        matched: dict[str, str] | None = None
+        for raw_tag in image["tags"]:
+            tag = raw_tag.strip().lower()
+            if not _is_probable_sha(tag):
+                continue
+            if tag in revision_map:
+                matched = revision_map[tag]
+                break
+            # Short-SHA: accept only if the prefix resolves to a single unique target.
+            prefix_targets = {
```
kunaals: question: since ambiguous full revisions are removed first, could a short prefix that matches one of those revisions later appear unique? should we resolve prefixes against the full set instead?

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,446 @@
...
+    revision). Produces at most one row per image digest.
+
+    revision_map already holds only revisions with a unique target. A short prefix is only
+    accepted when every full revision it prefixes resolves to that same single target;
+    prefixes matching two or more distinct targets are ambiguous and skipped.
+    """
+    matches: list[dict[str, Any]] = []
+    for image in images:
+        matched: dict[str, str] | None = None
+        for raw_tag in image["tags"]:
```
kunaals, line 241: question: if sha tags point to different repositories, does this make the result depend on the order returned by neo4j? would it be safer to treat multiple repository matches as ambiguous?

cartography/intel/circleci/supply_chain.py
```diff
@@ -0,0 +1,446 @@
...
+    already been upserted with the current update_tag and so survives.
+
+    Bounded to re-evaluated images on purpose: an image with no fresh evidence (its build
+    aged out of the lookback window) is left untouched, so we never delete a still-valid
+    edge just because its build is no longer in the feed. This is why there is no org-wide
+    cleanup (see sync()).
+    """
+    if not digests:
+        return
+    neo4j_session.run(
```
kunaals: nit: could we use `run_write_query()` here for the usual transaction retry handling?

kunaals (commented):
the earlier comments look addressed. i had a few follow-up questions before approving.

kunaals (approved):

PR #3065, opened 2026-07-22
feat(rules): declare an affected-node anchor (asset_label) on every Fact

jychp:
### Type of change
- [x] Documentation update

### Summary

A `Fact` exposed the affected resource's id (via `asset_id_field`, and only on some facts) but never the **Neo4j label** of that node. The label lived only implicitly inside the Cypher `MATCH (x:Label)`, so it was not machine-readable. A consumer that wants to locate the affected node in the graph (for enrichment, ownership, exposure/blast-radius analysis) had to guess the label from a field name (`pod_id` -> `KubernetesPod`) — a brittle, per-consumer mapping. It failed hardest on Kubernetes facts, whose id aliases carry no label, and on the facts that exposed no `asset_id_field` at all.

This change makes every Fact declare, as first-class self-describing metadata, the label of the single node it is about:

- **Model**: add a **required** `asset_label: str` to `Fact`, reusing the existing `asset_id_field` as the id half. `Fact.__post_init__` enforces that both are set and that `asset_id_field` is returned by `cypher_query` (`... AS <name>`), so a fact that omits the anchor fails to construct.
- **Results/JSON**: mirror `asset_label` / `asset_id_field` onto `FactResult` so they reach `cartography-rules run --output json` (same precedent as `identity_fields`).
- **All facts populated**: every fact now declares a resolvable `(asset_label, asset_id_field)` anchor. `asset_id_field` was added where it was missing. Namespace-aggregated Kubernetes facts now `OPTIONAL MATCH` and return the `KubernetesNamespace` id; the ingress-nginx EoL fact anchors on its parent `KubernetesCluster`.
- **Umbrella labels**: added `EntraPrincipal` / `ScalewayPrincipal` node labels (mirroring the existing `AWSPrincipal` / `GCPPrincipal`) so the polymorphic IAM admin-capability facts anchor on a single label; the OpenAI key fact uses the already-shared `APIKey` label.
- **Docs**: updated `docs/root/usage/rules.md` so every `Fact(...)` example includes the now-required fields (copying a snippet no longer raises `ValueError`) and the explanation no longer describes `asset_id_field` as compliance-metric-only.

### Related issues or links

- Fixes #

### Breaking changes

Not an API break. Note a compliance-metric behavior change: adding `asset_id_field` to a few facts that previously omitted it shifts their "failing" count from row-count to distinct-anchor-count. Affected facts (documented inline): `aws_admin_policy_attached`, `cloud_security_product_deactivated`, `k8s_service_account_tokens_mounted`, and the ingress-nginx EoL fact.

### How was this tested?

- `make test_lint` (pre-commit: black, isort, flake8, mypy) passes.
- `tests/unit/rules/` passes (1996 tests), including the new parametrized `tests/unit/rules/test_asset_anchor.py`, which asserts for every fact that `asset_label` is set, `asset_id_field` is set, is a field on the rule's output model, and is returned by the query — plus a JSON round-trip.
- `tests/unit/cartography/test_doc.py::test_schema_doc`, the ontology mapping tests, and the Entra/Scaleway model tests pass with the new umbrella labels.
- Full registry import verified: 182 facts across 127 rules each resolve to a valid `(label, id)` anchor.

#### General

#### Proof of functionality

- [x] The `EntraPrincipal` / `ScalewayPrincipal` labels are additive `extra_node_labels`; `test_schema_doc` passes with no manual schema-doc edits required.

### Notes for reviewers

- The anchor deliberately reuses `asset_id_field` rather than introducing a second id field: `asset_label` names the node type, `asset_id_field` names the output field carrying that node's `.id`.
- For nodes that can be one of several labels (Entra/Scaleway IAM principals, OpenAI keys), the anchor points at a shared umbrella label rather than guessing a subtype.
- The umbrella labels are applied at load time, so existing graphs pick them up on the next sync.

### DIFF
--- a/cartography/rules/data/rules/cloud_security_product_deactivated.py
+++ b/cartography/rules/data/rules/cloud_security_product_deactivated.py
@@ -32,6 +32,12 @@ aws_guard_duty_detector_disabled = Fact(
     WITH DISTINCT a, r.region AS region
     RETURN COUNT(*) AS count
     """,
+    # Anchor on the AWSAccount node. NOTE: the finding unit is (account, region), so
+    # account_id driving the failing-count collapses multiple disabled regions of one
+    # account into a single failing asset; the (account_id, region) identity_fields keep
+    # each region's finding distinct.
+    asset_label="AWSAccount",
+    asset_id_field="account_id",
     identity_fields=("account_id", "region"),
     module=Module.AWS,
     maturity=Maturity.EXPERIMENTAL,
--- a/cartography/rules/data/rules/cloud_security_product_deactivated.py
+++ b/cartography/rules/data/rules/cloud_security_product_deactivated.py
@@ -29,7 +29,7 @@ aws_guard_duty_detector_disabled = Fact(
     """,
     cypher_count_query="""
     MATCH (a:AWSAccount)-[:RESOURCE]-(r:AWSEC2Instance|AWSEKSCluster|AWSLambda|AWSECSCluster|AWSRDSInstance|AWSRDSCluster)
-    WITH DISTINCT a, r.region AS region
+    WITH DISTINCT a
     RETURN COUNT(*) AS count
     """,
     # Anchor on the AWSAccount node. NOTE: the finding unit is (account, region), so
--- a/cartography/rules/data/rules/nist_ai_rmf.py
+++ b/cartography/rules/data/rules/nist_ai_rmf.py
@@ -82,6 +82,7 @@ RISKY_SCOPE_PREFIXES_CYPHER = _as_cypher_list(RISKY_SCOPE_PREFIXES)
 # =============================================================================
 class NistAiAppInventoryOutput(Finding):
     app_name: str | None = None
+    asset_node_id: str | None = None
     app_client_id: str | None = None
     app_source: str | None = None
     match_method: str | None = None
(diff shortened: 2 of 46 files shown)


cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -146,6 +146,7 @@ class NistAiAppInventoryOutput(Finding):
     MATCH (app:ThirdPartyApp)
     RETURN COUNT(app) AS count
     """,
+    asset_label="ThirdPartyApp",
```
kunaals, line 151: i don't think `app_client_id` is a valid node anchor here. for entra and keycloak apps, it resolves to the oauth client id, which can differ from the node's actual `id`. could we return `app.id as asset_node_id` and use that as the anchor instead?

cartography/rules/data/rules/cloud_security_product_deactivated.py
```diff
@@ -32,6 +32,12 @@
     WITH DISTINCT a, r.region AS region
     RETURN COUNT(*) AS count
     """,
+    # Anchor on the AWSAccount node. NOTE: the finding unit is (account, region), so
+    # account_id driving the failing-count collapses multiple disabled regions of one
+    # account into a single failing asset; the (account_id, region) identity_fields keep
+    # each region's finding distinct.
+    asset_label="AWSAccount",
+    asset_id_field="account_id",
```
kunaals, line 40: this makes failing distinct by account, but the count query still counts account-region pairs. an account failing in two regions would report `total=2, failing=1, passing=1`. could we count distinct accounts so both sides use the same unit?

cartography/rules/data/rules/subimage_coverage.py
```diff
@@ -39,13 +39,16 @@
     WHERE toLower(app._ont_name) = toLower(m.id)
     RETURN count(m) AS count
     """,
+    asset_label="SubImageModule",
+    asset_id_field="module_id",
```
kunaals, line 43: this has the same unit mismatch when multiple apps match one module: the count query counts each match, while failing is deduplicated by `module_id`. could we use `count(distinct m)` here?

kunaals (approved):
approving to unblock, some nits

PR #3077, opened 2026-07-27
fix(trivy): only set cve_id and :CVE on CVE-backed findings

jychp:
### Type of change

- [x] Documentation update

### Summary

Trivy reports several vulnerability identifier schemes, not just CVEs: GitHub advisories (`GHSA-*`), Debian advisories (`DLA-*`, `DSA-*`, `TEMP-*`), `RUSTSEC-*`, `openSUSE-SU-*`, and others.

Today the transform copies the raw `VulnerabilityID` into `cve_id` unconditionally and the schema applies `:CVE` as an unconditional label. So a `GHSA-xxxx-xxxx-xxxx` finding ends up with a bogus `cve_id`, a bogus `_ont_cve_id` (via the trivy ontology mapping), and the `:CVE` semantic label it does not deserve. `MATCH (c:CVE)` therefore returns non-CVE findings.

This PR classifies the identifiers a finding actually reports, taking the union of the primary `VulnerabilityID` and any `VendorIDs`:

| Reported identifiers | `cve_id` | `ghsa_id` | `has_cve` | `:CVE` |
|---|---|---|---|---|
| `CVE-2024-11111` | set | null | `"true"` | yes |
| `GHSA-aaaa-bbbb-cccc` | null | set | `"false"` | no |
| `TEMP-…`, `DLA-…`, `RUSTSEC-…` | null | null | `"false"` | no |
| `DSA-5895-1` + `CVE-2025-31115` in `VendorIDs` | set | null | `"true"` | yes |

`:CVE` becomes conditional (`CVE.when(has_cve="true")`), and `:Risk` stays unconditional. `id` (`TIF|<vuln-id>`) and `name` keep tracking the primary identifier for every scheme, so nothing becomes unreachable.

This aligns Trivy with the other mixed-identifier scanners in the repo (`TenableFinding`, `SemgrepSCAFinding`, `GitHubDependabotAlert`, `AWSInspectorFinding`), which all already gate `:CVE` behind a conditional label and leave `cve_id` null for non-CVE identifiers. Trivy was the last one still doing it unconditionally.

### Related issues or links

Builds on #3032 (declarative extra node labels) and #3082 (conditional labels applied per row), both merged into this branch.

### How was this tested?

- New unit tests on `transform_scan_results()`: one covering a `CVE-*`, a `GHSA-*` and a `TEMP-*` primary identifier, one covering identifiers arriving through `VendorIDs`.
- New integration test (`tests/integration/cartography/intel/trivy/test_trivy_finding_identifiers.py`) driving `sync_single_image()` with a mixed-identifier fixture, asserting `cve_id` / `ghsa_id` / `has_cve` and the presence/absence of the `:CVE` and `:Risk` labels for each scheme. A second case seeds the pre-fix state (raw id in `cve_id`, `:CVE` applied) and asserts the sync corrects it.

```
uv run pytest tests/unit/cartography/intel/trivy tests/integration/cartography/intel/trivy -q                      # 40 passed
uv run pytest tests/unit/cartography/graph tests/integration/cartography/graph \
              tests/integration/cartography/intel/ontology -q                                                      # 298 passed
make test_lint                                                                                                     # all hooks pass
```

Post-sync graph checks, both expected to return 0 rows:

```cypher
MATCH (f:TrivyImageFinding) WHERE f.cve_id IS NOT NULL AND NOT f.cve_id STARTS WITH 'CVE-' RETURN f.id
```

```cypher
MATCH (f:TrivyImageFinding:CVE) WHERE f.cve_id IS NULL RETURN f.id
```

#### General

#### Proof of functionality

### Notes for reviewers

**No migration job.** Existing bad data self-heals on the next full sync. The generated property SET clause is a direct `i.cve_id = item.cve_id`, so `None` nulls out stale values, and since #3082 the conditional label is maintained by a per-row `FOREACH` pair, whose negative branch strips a stale `:CVE` from any finding that is re-loaded. A finding that is no longer reported is deleted by cleanup instead. The integration test covers exactly this path.

**Ontology mapping unchanged.** `trivy_mapping` in `cartography/models/ontology/mapping/data/cves.py` is keyed on the primary label `TrivyImageFinding`, not on `:CVE`, so `_ont_cve_id` simply becomes null for non-CVE findings. This is the same behaviour `SemgrepSCAFinding` already relies on. `ghsa_id` gets no ontology field since none exists for GHSA today.

**On `VendorIDs`.** The union is classified rather than the primary identifier alone, so an identifier is recognised whichever slot it arrives in. In practice `VendorIDs` carries vendor advisory ids only: upstream documents the field as `// e.g. RHSA-ID and DSA-ID`, and every occurrence in our own captured fixtures is a `DSA-*` on a CVE-primary finding. So the added branch is defensive rather than load-bearing on today's Trivy output.

**Not in scope.** Non-CVE findings keep only `:Risk`; I did not add a `SecurityIssue` conditional label the way Semgrep and AWS Inspector do, since that would require adding `SecurityIssue` ontology field mappings (title/severity/status/first_seen) for Trivy. Happy to do it in a follow-up if wanted.

The redundant `WHERE c.cve_id STARTS WITH "CVE"` guard in `cartography/intel/cve_metadata/__init__.py` is left in place as defence in depth.

### DIFF
--- a/cartography/intel/trivy/scanner.py
+++ b/cartography/intel/trivy/scanner.py
@@ -73,11 +73,20 @@ def transform_scan_results(
                     data_source_id = result["DataSource"].get("ID")
                     data_source_name = result["DataSource"].get("Name")
 
+                # Trivy reports more than CVEs: GHSA-, DLA-/DSA-, TEMP-, RUSTSEC-, etc.
+                # Only CVE-prefixed ids populate cve_id (which gates the :CVE label);
+                # GHSA-prefixed ids land in ghsa_id. Other schemes populate neither and
+                # remain readable through name/id.
+                vuln_id = result["VulnerabilityID"]
+                is_cve = vuln_id.startswith("CVE-")
+
                 # Transform finding data
                 finding = {
-                    "id": f'TIF|{result["VulnerabilityID"]}',
-                    "VulnerabilityID": result["VulnerabilityID"],
-                    "cve_id": result["VulnerabilityID"],
+                    "id": f"TIF|{vuln_id}",
+                    "VulnerabilityID": vuln_id,
+                    "cve_id": vuln_id if is_cve else None,
+                    "ghsa_id": vuln_id if vuln_id.startswith("GHSA-") else None,
+                    "has_cve": "true" if is_cve else "false",
                     "Description": result.get("Description"),
                     "LastModifiedDate": result.get("LastModifiedDate"),
                     "PrimaryURL": result.get("PrimaryURL"),
--- a/cartography/intel/trivy/scanner.py
+++ b/cartography/intel/trivy/scanner.py
@@ -75,19 +75,20 @@ def transform_scan_results(
 
                 # Trivy reports more than CVEs: GHSA-, DLA-/DSA-, TEMP-, RUSTSEC-, etc.
                 # The primary identifier lives in VulnerabilityID, and vendor advisory
-                # ids (e.g. RHSA-, DSA-) in VendorIDs. Classify the union so an
-                # identifier is recognised whichever slot it appears in, rather than
-                # assuming the scheme of the primary id. cve_id gates the :CVE label;
-                # schemes we have no field for populate neither and stay readable
-                # through name/id.
+                # ids (e.g. RHSA-, DSA-) in VendorIDs. Keep the whole set on the finding
+                # and classify it, so an identifier is recognised whichever slot it
+                # appears in and no authority is dropped for lack of a dedicated field.
+                # cve_id gates the :CVE label.
                 vuln_id = result["VulnerabilityID"]
-                all_ids = [vuln_id, *(result.get("VendorIDs") or [])]
+                vulnerability_ids = list(
+                    dict.fromkeys([vuln_id, *(result.get("VendorIDs") or [])]),
+                )
                 cve_id = next(
-                    (vid for vid in all_ids if vid.startswith("CVE-")),
+                    (vid for vid in vulnerability_ids if vid.startswith("CVE-")),
                     None,
                 )
                 ghsa_id = next(
-                    (vid for vid in all_ids if vid.startswith("GHSA-")),
+                    (vid for vid in vulnerability_ids if vid.startswith("GHSA-")),
                     None,
                 )
 
@@ -95,6 +96,7 @@ def transform_scan_results(
                 finding = {
                     "id": f"TIF|{vuln_id}",
                     "VulnerabilityID": vuln_id,
+                    "vulnerability_ids": vulnerability_ids,
                     "cve_id": cve_id,
                     "ghsa_id": ghsa_id,
                     "has_cve": "true" if cve_id else "false",
--- a/cartography/intel/trivy/scanner.py
+++ b/cartography/intel/trivy/scanner.py
@@ -74,19 +74,30 @@ def transform_scan_results(
                     data_source_name = result["DataSource"].get("Name")
 
                 # Trivy reports more than CVEs: GHSA-, DLA-/DSA-, TEMP-, RUSTSEC-, etc.
-                # Only CVE-prefixed ids populate cve_id (which gates the :CVE label);
-                # GHSA-prefixed ids land in ghsa_id. Other schemes populate neither and
-                # remain readable through name/id.
+                # The primary identifier lives in VulnerabilityID, and vendor advisory
+                # ids (e.g. RHSA-, DSA-) in VendorIDs. Classify the union so an
+                # identifier is recognised whichever slot it appears in, rather than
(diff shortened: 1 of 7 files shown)


cartography/intel/trivy/scanner.py
```diff
@@ -73,11 +73,20 @@ def transform_scan_results(
                     data_source_id = result["DataSource"].get("ID")
                     data_source_name = result["DataSource"].get("Name")
 
+                # Trivy reports more than CVEs: GHSA-, DLA-/DSA-, TEMP-, RUSTSEC-, etc.
+                # Only CVE-prefixed ids populate cve_id (which gates the :CVE label);
+                # GHSA-prefixed ids land in ghsa_id. Other schemes populate neither and
+                # remain readable through name/id.
+                vuln_id = result["VulnerabilityID"]
```
kunaals, line 82: trivy also exposes alternate identifiers in `VendorIDs`. could we classify the union of `VulnerabilityID` and `VendorIDs`, preserve the full list, and derive `cve_id`, `ghsa_id`, and `has_cve` from that? as written, `has_cve` means “the primary id starts with `CVE-`”, and a cve-primary finding loses its ghsa alias.

cartography/models/trivy/findings.py
```diff
@@ -65,7 +68,14 @@ class TrivyFindingToOntologyImageRel(CartographyRelSchema):
 class TrivyImageFindingSchema(CartographyNodeSchema):
     label: str = "TrivyImageFinding"
     scoped_cleanup: bool = False
-    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Risk", "CVE"])
+    # Trivy reports CVEs alongside other identifier schemes (GHSA-, DLA-, TEMP-, ...),
+    # so :CVE is only applied to CVE-backed findings. :Risk covers all of them.
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(
+        [
+            "Risk",
+            ConditionalNodeLabel(label="CVE", conditions={"has_cve": "true"}),
```
kunaals: could we reconcile this once after all trivy reports load instead of after every non-empty image load? i agree the remove/re-add is needed to clear stale labels, but this schema is unscoped, so `load()` currently removes and re-adds every trivy `:CVE` label once per report. a temporary analysis migration plus one post-ingestion reconciliation would preserve that behavior without repeating the global pass.

docs/root/modules/trivy/schema.md
```diff
@@ -1,15 +1,22 @@
...
 |-------|-------------|
 | firstseen | Timestamp of when a sync job first discovered this node |
 | lastupdated | Timestamp of the last time the node was updated |
-| **id** | Unique identifier for the finding (format: TIF|CVE-ID) |
-| name | The vulnerability ID (e.g. CVE-2024-1234) |
-| cve_id | The CVE identifier |
+| **id** | Unique identifier for the finding (format: TIF|VULNERABILITY-ID, e.g. `TIF|CVE-2024-1234` or `TIF|GHSA-xxxx-xxxx-xxxx`) |
+| name | The vulnerability ID, whatever its scheme (e.g. CVE-2024-1234, GHSA-xxxx-xxxx-xxxx) |
+| cve_id | The CVE identifier. Set only when the vulnerability ID is a CVE, null otherwise |
+| ghsa_id | The GitHub advisory identifier. Set only when the vulnerability ID is a GHSA, null otherwise |
```
kunaals: `GHSA-*` and `CVE-*` aren’t mutually exclusive identities; the same vulnerability can have both. could we describe these fields in terms of which identifiers are present in the report rather than which scheme the primary id uses?

kunaals (changes requested):

cartography/intel/trivy/scanner.py
```diff
@@ -73,11 +73,31 @@ def transform_scan_results(
...
 
+                # Trivy reports more than CVEs: GHSA-, DLA-/DSA-, TEMP-, RUSTSEC-, etc.
+                # The primary identifier lives in VulnerabilityID, and vendor advisory
+                # ids (e.g. RHSA-, DSA-) in VendorIDs. Classify the union so an
+                # identifier is recognised whichever slot it appears in, rather than
+                # assuming the scheme of the primary id. cve_id gates the :CVE label;
+                # schemes we have no field for populate neither and stay readable
+                # through name/id.
+                vuln_id = result["VulnerabilityID"]
+                all_ids = [vuln_id, *(result.get("VendorIDs") or [])]
```
kunaals: we build the full identifier list here, but then only keep the first cve and ghsa. could we also store the full list on the finding, maybe as `vulnerability_ids`? otherwise aliases like dsa/rhsa are still lost, and we’ll need another schema change every time we want to support a new identifier authority.

kunaals (approved):
one last nit and then its good

PR #3079, opened 2026-07-27
feat(railway): add Railway intel module

jychp:
### Type of change

- [ ] Documentation update

### Summary

Adds a new intel module for [Railway](https://railway.com), a PaaS that hosts container workloads with public HTTPS domains, raw TCP proxies, persistent volumes, environment variables and long-lived API tokens. None of that surface was previously visible in the graph.

The module is modelled on the **cloud-provider** modules (Scaleway first, then AWS/GCP/Azure) rather than the SaaS ones: two-level tenancy with per-project scoped cleanup, the cloud relationship vocabulary, region-as-a-string, subpackage-per-domain layout, ontology semantic labels, and an internet-exposure `Fact` instead of a bespoke analysis job.

**16 node types** across five tiers:

| Tier | Nodes |
|---|---|
| Tenancy | `RailwayWorkspace`, `RailwayProject` (both `Tenant`) |
| Identity | `RailwayUser` (`UserAccount` + `RailwayPrincipal`), `RailwayApiToken`, `RailwayProjectToken` (both `APIKey`) |
| Compute | `RailwayEnvironment`, `RailwayService`, `RailwayServiceInstance` (`ComputeService`), `RailwayDeployment` (`Container`) |
| Exposure | `RailwayServiceDomain`, `RailwayCustomDomain`, `RailwayTCPProxy` |
| Storage / config | `RailwayVolume`, `RailwayVolumeInstance` (`BlockStorage`), `RailwayVariable` (`Secret`), `RailwayDeploymentTrigger` |

Key modelling decisions:

- **Two tenancy levels**, mirroring `ScalewayOrganization` → `ScalewayProject`. A workspace owns projects; a project owns every deployable resource. Both carry `Tenant`, and every leaf scopes its cleanup to the *nearest* tenant rather than chaining two hops.
- **`RESOURCE` is reserved for Tenant ⇒ X** and appears only as a `sub_resource_relationship`. Containment between non-tenant nodes uses `HAS`; volumes use `MOUNTS`; public entry points use `EXPOSE`; tokens use `OWNED_BY`.
- **`Service` vs `ServiceInstance`.** A Railway service is an environment-agnostic shell; source image or repo, region, replicas, health check and domains all live on the per-environment instance. That is where `ComputeService` goes, with `RailwayDeployment` as the `Container` linked by `WORKLOAD_PARENT` — the same split GCP Cloud Run uses for `Service` / `ServiceContainer`.
- **Variable names only.** Railway can return variable values; Cartography never requests or stores them. The GraphQL selection omits `value`, the schema has no `value` property, and a test asserts both.
- **Cross-module links to GitHub.** `ServiceInstance-[:DEPLOYED_FROM]->GitHubRepository` and `DeploymentTrigger-[:TRACKS]->GitHubRepository`, joined on `owner/name` against `GitHubRepository.fullname`. Best effort: no edge forms if the GitHub module has not ingested the repo.

### Related issues or links

- [Railway public API reference](https://docs.railway.com/reference/public-api)

### How was this tested?

- **Unit** (`tests/unit/cartography/intel/railway/`): GraphQL error handling, Relay pagination including the unchanged-cursor guard, 429 / `Retry-After` behaviour, the aliased TCP-proxy batch builder and its UUID validation, and bundle-pagination draining.
- **Integration** (`tests/integration/cartography/intel/railway/`): 30 tests driving `sync()` end to end against a testcontainers Neo4j, asserting nodes and relationships with `check_nodes` / `check_rels`. Includes per-project cleanup scoping, both GitHub join negative cases, and the API-token completeness guard.
- **Full suite**: `3921 passed`. `make test_lint` clean.
- **Live environment**: the module was run against a real Railway Hobby workspace provisioned for the purpose (project, two services, two environments, generated domain, custom variables, two volumes, two TCP proxies, deployments), then the workspace was torn down. Fixtures are captured from those real API responses with identifiers and PII replaced.

#### General

#### Proof of functionality

- [x] Included Cartography sync logs from a real environment.

Sync against a live Railway workspace:

```
INFO:cartography.sync:Starting sync stage 'railway'
INFO:cartography.intel.railway:Syncing Railway workspace e36b380e-285d-4784-b311-483e58caf7db
INFO:cartography.client.core.tx:Loaded 1 RailwayWorkspace nodes
INFO:cartography.client.core.tx:Loaded 1 RailwayProject nodes
INFO:cartography.client.core.tx:Loaded 1 RailwayUser nodes
INFO:cartography.client.core.tx:Loaded 2 RailwayEnvironment nodes
INFO:cartography.client.core.tx:Loaded 2 RailwayService nodes
INFO:cartography.client.core.tx:Loaded 2 RailwayServiceInstance nodes
INFO:cartography.client.core.tx:Loaded 1 RailwayServiceDomain nodes
INFO:cartography.client.core.tx:Loaded 2 RailwayTCPProxy nodes
INFO:cartography.client.core.tx:Loaded 2 RailwayDeployment nodes
INFO:cartography.client.core.tx:Loaded 2 RailwayVolume nodes
INFO:cartography.client.core.tx:Loaded 2 RailwayVolumeInstance nodes
INFO:cartography.client.core.tx:Loaded 14 RailwayVariable nodes
INFO:cartography.sync:Finishing sync stage 'railway'
```

Resulting graph (32 nodes, 73 relationships):

```
=== REL COUNTS ===          === ONTOLOGY LABELS ===
  EXPOSE            3         BlockStorage      2
  HAS              20         ComputeService    2
  MEMBER_OF         1         Container         2
  MOUNTS            2         Secret           14
  RESOURCE         31         Tenant            2
  USES_SECRET      14         UserAccount       1
  WORKLOAD_PARENT   2
```

Invariants checked against that live graph:

- Every `RESOURCE` edge originates from a `Tenant`-labelled node — zero violations.
- No `RailwayVariable` has a `value` property — count 0.
- The new exposure `Fact` fired on both service instances, including a managed Postgres reachable from the internet through its TCP proxy.

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules) — new `docs/root/modules/railway/{index,config,schema}.md`, registered in `docs/root/index.md`, `docs/root/module-list.md`, `docs/root/usage/schema.md` and the README module list.
- [ ] Updated the schema README — `docs/schema/README.md` does not exist in the repo; schema docs live under `docs/root/modules/`.

### Notes for reviewers

Three things are worth a close look.

**1. Per-project scoped cleanup is the easiest thing to get wrong here.** Because leaves scope to `RailwayProject` rather than to the workspace, each `cleanup()` loops over projects and `.copy()`s the common job parameters so a project's scope never leaks into a later job. `test_railway_project_cleanup_removes_only_stale_projects` covers the tenant level; each domain's cleanup follows the same shape.

**2. Railway returns HTTP 200 on auth failure**, reporting the problem in the `errors` array. `call_railway_api()` therefore *raises* rather than logging and continuing — treating such a response as an empty result set would let a cleanup job delete every node in the workspace. Two places deliberately soften this, and both had to be careful:

- `me` is unreadable by a workspace-scoped token. That is fine as long as `--railway-workspace-id` was given; without either, the module skips rather than syncing nothing.
- `apiTokens` returns `Not Authorized` (not an empty list) for tokens that are not account-scoped — verified against the live API. `get_api_tokens()` returns a `(tokens, readable)` pair and the sync skips **both** the load and the cleanup when unreadable, so switching from an account token to a workspace token cannot silently delete previously ingested tokens. A test also pins the converse: a genuinely empty inventory still cleans up.

**3. Bundle pagination.** Every connection in `PROJECT_BUNDLE_QUERY` is capped at `first: $first`, so a project with more than 100 of any resource comes back truncated — and the cleanup jobs treat what they are given as exhaustive. `get_project_bundle()` drains each truncated connection with a targeted follow-up before returning, including recursing into environments returned by the follow-up itself. The follow-up documents are composed from the same field-selection constants as the bundle so the two cannot drift, and all nine documents were validated against the live Railway schema. Projects that fit in one page still cost exactly one request.

**Fixture caveats.** A throwaway Hobby project cannot produce custom domains (they need a real domain plus DNS verification), non-`ADMIN` roles (a personal workspace has exactly one member and no per-project membership), or deployment triggers (they need a linked GitHub account). Those three are hand-written from the live GraphQL schema and called out in the fixture docstrings; everything else is exactly what the API returned.

**Deferred.** `sshPublicKeys`, `trustedDomains`, `workspaceIdentityProviders` and audit logs are Pro/Enterprise-only and could not be exercised, so they are left for a follow-up. `has_saml` and `has_2fa_enforcement` do ship as workspace booleans.

**Docs discrepancy worth knowing:** Railway's docs show `deploymentTriggers(projectId:)` as a project-level root field, but the live schema requires `serviceId` and `environmentId` too. It is queried under `Environment` in the bundle instead, where it costs nothing extra.


cartography/models/railway/workspace.py
```diff
@@ -0,0 +1,28 @@
...
+    has_2fa_enforcement: PropertyRef = PropertyRef("has2FAEnforcement")
+    has_saml: PropertyRef = PropertyRef("hasSAML")
+    plan: PropertyRef = PropertyRef("plan")
+
+
+@dataclass(frozen=True)
+class RailwayWorkspaceSchema(CartographyNodeSchema):
+    label: str = "RailwayWorkspace"
+    properties: RailwayWorkspaceNodeProperties = RailwayWorkspaceNodeProperties()
+    extra_node_labels: ExtraNodeLabels = ExtraNodeLabels(["Tenant"])
```
kunaals: this no longer imports on current master: `ExtraNodeLabels` now requires `ExtraNodeLabel` values, not strings. please use the canonical ontology constants and define a reusable `RAILWAY_PRINCIPAL`.

cartography/intel/railway/queries.py
```diff
@@ -0,0 +1,411 @@
...
+    createdAt
+"""
+
+# The child connections of an Environment, in the exact shape the transforms expect.
+_ENVIRONMENT_CHILDREN = f"""
+    serviceInstances(first: $first) {{
+      edges {{ node {{ {_SERVICE_INSTANCE_FIELDS} }} }}
+      {_PAGE_INFO}
+    }}
+    deployments(first: $first) {{
```
kunaals: this drains the entire deployment history on every sync, then models every failed/removed attempt as a `Container`. can we bound this to current/latest deployments and keep historical revisions out of the container ontology?

cartography/intel/railway/serviceinstances.py
```diff
@@ -0,0 +1,115 @@
...
+    by_project: dict[str, list[dict[str, Any]]] = {}
+    for project_id, bundle in bundles.items():
+        transformed = []
+        for instance in iter_service_instances(bundle):
+            source = instance.get("source") or {}
+            latest_deployment = instance.get("latestDeployment") or {}
+            tcp_proxies = tcp_proxies_by_instance.get(instance["id"], [])
+            transformed.append(
+                {
+                    **instance,
```
kunaals, line 92: `region` is normally null unless the service overrides the workspace default, so most `ComputeService` nodes lose their effective region. please fall back to `workspace.preferredRegion` and clarify how multi-region replicas are represented.

cartography/intel/railway/serviceinstances.py
```diff
@@ -0,0 +1,115 @@
...
+        instances.extend(unwrap_edges(environment["serviceInstances"]))
+    return instances
+
+
+def _is_publicly_exposed(
+    instance: dict[str, Any],
+    tcp_proxies: list[dict[str, Any]],
+) -> bool:
+    domains = instance.get("domains") or {}
+    if domains.get("serviceDomains"):
```
kunaals: this treats any domain or proxy object as live exposure, including `CREATING` and `DELETED` states. please gate both `is_publicly_exposed` and the `EXPOSE` edges on the active lifecycle states.

cartography/models/ontology/mapping/data/blockstorage.py
```diff
@@ -129,8 +129,46 @@
...
+# A Railway volume instance is the actual disk: the project-scoped RailwayVolume is only a
+# definition, so size, region and state all live here.
+railway_mapping = OntologyMapping(
+    module_name="railway",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="RailwayVolumeInstance",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name", node_field="mount_path", required=True
```
kunaals: `mount_path` isn't the volume's name or identifier, so `_ont_name` becomes values like `/var/lib/postgresql/data`. please use the instance id or denormalize the parent volume name.

tests/integration/cartography/intel/railway/test_users.py
```diff
@@ -0,0 +1,298 @@
...
+    other_workspace_id = "1a1a1a1a-1a1a-1a1a-1a1a-1a1a1a1a1a1a"
+    # Deliberately stale: written by a previous run, and this run has not reached that
+    # workspace yet.
+    neo4j_session.run(
+        """
+        MERGE (w:RailwayWorkspace {id: $other_workspace_id})
+        SET w.lastupdated = $stale_tag
+        MERGE (u:RailwayUser {id: $user_id})
+        SET u.lastupdated = $stale_tag
+        MERGE (w)-[r:RESOURCE]->(u)
```
kunaals, line 260: these fixture edges omit `_sub_resource_label` and `_sub_resource_id`, so they survive regardless of whether workspace scoping is wired correctly. please add the other workspace's scope properties so this actually tests isolation.

kunaals (changes requested):

cartography/intel/railway/deployments.py
```diff
@@ -0,0 +1,114 @@
...
+def transform(
+    bundles: dict[str, dict[str, Any]],
+) -> tuple[dict[str, list[dict[str, Any]]], dict[str, list[dict[str, Any]]]]:
+    deployments: dict[str, list[dict[str, Any]]] = {}
+    triggers: dict[str, list[dict[str, Any]]] = {}
+
+    for project_id, bundle in bundles.items():
+        # The deployment each service instance is currently serving. Everything else in the
+        # window is a superseded revision.
+        current_ids = {
```
kunaals, line 39: question: can `latestDeployment` point to a failed or crashed deployment? if so, we’d label that attempt as the current `Container`, even though an earlier successful deployment may still be serving. should we also check the deployment status here?

cartography/intel/railway/queries.py
```diff
@@ -0,0 +1,420 @@
...
+  environment(id: $environmentId) {{
+    serviceInstances(first: $first, after: $after) {{
+      edges {{ node {{ {_SERVICE_INSTANCE_FIELDS} }} }}
+      {_PAGE_INFO}
+    }}
+  }}
+}}
+"""
+
+ENVIRONMENT_DEPLOYMENTS_QUERY = f"""
```
kunaals: nit: looks like `ENVIRONMENT_DEPLOYMENTS_QUERY` isn’t used anymore now that deployments are fetched with the environment bundle. can we remove it?

kunaals (approved):
approving with a nit and a question

PR #3080, opened 2026-07-27
feat(rules): add SOC 2 compliance mappings

jychp:
### Type of change
- [x] Documentation update

### Summary
- Add a SOC 2 Trust Services Criteria framework helper with canonical criterion metadata.
- Reuse existing provider and provider-neutral detections as technical evidence for SOC 2 controls.
- Document automatable coverage gaps that require additional data models, and add mapping consistency tests.

### Related issues or links
None.

### How was this tested?
- `uv run pytest tests/unit/rules` (1,270 passed)
- Commit-time checks passed.

#### General

#### Proof of functionality

### Notes for reviewers
SOC 2 criteria that require manual organizational evidence are intentionally excluded. TODO blocks identify technical criteria that could be automated after extending Cartography's data models.


kunaals (commented):
the framework mappings look reasonable overall. i had a few questions about the underlying rule behavior before approving.

cartography/rules/data/rules/security_monitoring_gaps.py
```diff
@@ -0,0 +1,142 @@
...
+        OR resource:AWSLambda
+        OR resource:AWSECSCluster
+        OR resource:AWSRDSInstance
+        OR resource:AWSRDSCluster
+      )
+    WITH DISTINCT account, resource.region AS region
+    RETURN COUNT(*) AS count
+    """,
+    asset_label="AWSAccount",
+    asset_id_field="account_id",
```
kunaals, line 81: question: the total here counts account-region pairs, while `asset_id_field` identifies findings by account. could one account missing security hub in two regions end up showing total 2, failing 1, and passing 1? should the total and finding use the same identity?

cartography/rules/data/rules/github_change_management.py
```diff
@@ -0,0 +1,77 @@
...
+    description=(
+        "Detects active GitHub repositories that have no branch protection rule "
+        "requiring at least one approving pull request review."
+    ),
+    cypher_query="""
+    MATCH (repo:GitHubRepository)
+    WHERE coalesce(repo.archived, false) = false
+      AND coalesce(repo.disabled, false) = false
+      AND NOT EXISTS {
+          MATCH (repo)-[:HAS_RULE]->(rule:GitHubBranchProtectionRule)
```
kunaals: i might be missing some matching elsewhere, but could a repository pass this check when reviews are only required on something like `release/*` and the default branch is still unprotected? should we match the protection pattern against `repo.defaultbranch` and consider rulesets too?

cartography/rules/data/rules/device_security_posture_gaps.py
```diff
@@ -355,13 +293,336 @@ class DeviceSecurityPostureGapOutput(Finding):
...
+    id="device_malware_protection_gaps",
+    name="Devices with malware protection gaps",
+    description=(
+        "Detects tampering, malware infections, active endpoint threats, missing "
+        "security agents, reduced endpoint-protection functionality, and disabled "
+        "platform integrity protections."
+    ),
+    cypher_query="""
+    CALL {
+        MATCH (asset:Device)-[:OBSERVED_AS]->(device:DuoPhone)
```
kunaals: question: does requiring an `OBSERVED_AS` relationship here exclude provider devices that haven’t been mapped to `Device` yet? the previous provider-specific queries could still report those nodes. should we keep that coverage or confirm the mapping is exhaustive first?

cartography/rules/data/rules/cis_4_0_gcp.py
```diff
@@ -1653,9 +1702,80 @@ class BucketUniformAccessOutput(Finding):
...
+    id="gcp_kms_key_without_rotation_policy",
+    name="GCP KMS encryption keys without a rotation policy",
+    description=(
+        "Detects symmetric GCP KMS encryption keys that have no configured "
+        "automatic rotation period."
+    ),
+    cypher_query="""
+    MATCH (project:GCPProject)-[:RESOURCE]->(key:GCPCryptoKey)
+    WHERE key.purpose = 'ENCRYPT_DECRYPT'
+      AND (key.rotation_period IS NULL OR key.rotation_period = '')
```
kunaals: question: this catches keys with no rotation schedule, but would a schedule like `31536000s` still pass even though CIS 1.10 calls for rotation within 90 days? since `rotation_period` is already available, could we check the maximum period here too?

cartography/rules/data/rules/security_monitoring_gaps.py
jychp: Fixed in 0e1ec8939. Both Security Hub facts now count distinct AWS accounts, matching their AWSAccount asset anchor.

cartography/rules/data/rules/github_change_management.py
jychp: Fixed in 0e1ec8939. The rule now requires branch protection to match the default branch and also recognizes active branch rulesets with pull request review requirements that include, and do not exclude, the default branch.

cartography/rules/data/rules/device_security_posture_gaps.py
jychp: Fixed in 0e1ec8939. Malware and management checks are now provider-specific Facts that query provider nodes directly, so coverage no longer depends on OBSERVED_AS mappings.

cartography/rules/data/rules/cis_4_0_gcp.py
jychp: Fixed in 0e1ec8939. The KMS rule now flags missing schedules and rotation periods longer than 7,776,000 seconds (90 days).

cartography/rules/data/rules/github_change_management.py
```diff
@@ -0,0 +1,121 @@
...
+        "requiring at least one approving pull request review."
+    ),
+    cypher_query="""
+    MATCH (repo:GitHubRepository)
+    WHERE coalesce(repo.archived, false) = false
+      AND coalesce(repo.disabled, false) = false
+      AND NOT (
+        EXISTS {
+            MATCH (repo)-[:HAS_RULE]->(rule:GitHubBranchProtectionRule)
+            WHERE rule.pattern = repo.defaultbranch
```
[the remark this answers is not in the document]
jychp: Fixed in 66842c3f0. Branch-protection patterns are now evaluated against the default branch using GitHub-style path wildcards, including *, **, **/, ?, and character classes.

cartography/rules/data/rules/github_change_management.py
```diff
@@ -0,0 +1,121 @@
...
+                  -[:CONTAINS_RULE]->(rule:GitHubRulesetRule)
+            WHERE ruleset.target = 'BRANCH'
+              AND ruleset.enforcement = 'ACTIVE'
+              AND rule.type = 'PULL_REQUEST'
+              AND coalesce(rule.parameters_required_approving_review_count, 0) >= 1
+              AND (
+                '~ALL' IN coalesce(ruleset.conditions_ref_name_include, [])
+                OR '~DEFAULT_BRANCH' IN coalesce(ruleset.conditions_ref_name_include, [])
+                OR ('refs/heads/' + repo.defaultbranch)
+                   IN coalesce(ruleset.conditions_ref_name_include, [])
```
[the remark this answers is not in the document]
jychp: Fixed in 66842c3f0. Ruleset include and exclude patterns now use the same GitHub-style ref matching, while preserving ~ALL and ~DEFAULT_BRANCH handling. A matching exclusion takes precedence.

kunaals (approved):

PR #3082, opened 2026-07-28
perf(querybuilder): apply conditional node labels per row instead of via global remove/set passes

jychp:
### Type of change

- [ ] Documentation update

### Summary

Conditional node labels (`ExtraNodeLabel.when(...)`, added in #3032) were maintained by two extra
Cypher queries that `load()` ran *after* every batch write: a blanket `REMOVE` followed by a filtered
`SET`, per conditional label. For a schema with no `sub_resource_relationship` there is nothing to
narrow by, so the pair swept every node carrying the primary label:

```cypher
MATCH (n:GCPArtifactRegistryImage:Image) REMOVE n:Image
MATCH (n:GCPArtifactRegistryImage) WHERE n.type = "image" SET n:Image
```

Since `load()` is typically called in a loop (per project, per report), that whole-label sweep was
repeated once per iteration. Two schemas are in this unscoped bucket today: `aibom/component.py`
(6 conditional labels, so 12 sweeps per `load()`, and `load()` runs once per report) and
`gcp/artifact_registry/image.py` (3 labels, once per project).

This PR folds the conditional label into the ingestion query itself, one pair of `FOREACH` clauses
per row, mirroring how unconditional labels are already folded into the `SET` clause. `FOREACH` over
a one-or-zero element list is the standard pure-Cypher conditional-write idiom, so no APOC dependency
is introduced. The post-load queries and `build_conditional_label_queries()` are deleted;
condition-field index creation is unchanged.

Generated Cypher for `GCPArtifactRegistryImageSchema`:

```cypher
UNWIND $DictList AS item
    MERGE (i:GCPArtifactRegistryImage{id: item.digest})
    ON CREATE SET i.firstseen = timestamp()
    SET
        ...
        i._ont_digest = item.digest
    FOREACH (_ IN CASE WHEN i.type = "image" THEN [1] ELSE [] END | SET i:Image)
    FOREACH (_ IN CASE WHEN i.type = "image" THEN [] ELSE [1] END | REMOVE i:Image)
    FOREACH (_ IN CASE WHEN i.type = "attestation" THEN [1] ELSE [] END | SET i:ImageAttestation)
    FOREACH (_ IN CASE WHEN i.type = "attestation" THEN [] ELSE [1] END | REMOVE i:ImageAttestation)
    ...
```

No cleanup pass is needed. A node holding a stale conditional label is in exactly one of two states:
it is re-loaded during the sync and the negative `FOREACH` strips the label row by row, or it is no
longer reported and cleanup deletes the node outright. The graph converges without ever scanning the
whole label.

GCP Artifact Registry had built its own load helper with an `apply_labels` opt-out to run the sweep
once per sync rather than per batch. That workaround is no longer needed, so
`apply_conditional_labels()` and the flag are removed. `load_nodes_without_relationships()` itself
stays: it has a second, unrelated reason to exist (batch progress logging).

Declarations sharing a label are now ORed together instead of clobbering each other, which removes
the single-condition-per-label limitation that was documented on `AWSInspectorFindingSchema`.

### Related issues or links

Builds on #3032.

### How was this tested?

- `make test_lint` passes.
- `make test_unit` passes (3877 tests).
- `make test_integration` passes (1186 tests), against Neo4j 2025.10.

The strongest signal: the 7 pre-existing integration tests in
`tests/integration/cartography/graph/test_querybuilder_conditional_labels.py` passed **unchanged**
before any test file was touched, including
`test_conditional_labels_updated_when_conditions_change` (which flips a condition and asserts the
old label is gone).

Test changes:
- `tests/unit/.../test_querybuilder_conditional_labels.py`: rewritten to assert on the generated
  ingestion query rather than on the deleted `build_conditional_label_queries()`. Adds coverage for
  OR-ed declarations of one label, and for a label declared both unconditionally and conditionally
  (which must get no removal clause).
- `tests/integration/.../test_querybuilder_conditional_labels.py`: two new cases for the OR-ed
  declaration path (label set from either declaration; label removed once neither holds).
- `tests/unit/cartography/client/core/test_tx.py`: dropped the
  `build_conditional_label_queries` patches.

### Notes for reviewers

**Behavior difference worth knowing.** The old blanket remove-then-set also corrected nodes that were
*not* in the current batch, e.g. one whose condition property was changed by some other code path.
The per-row approach only corrects nodes that are re-loaded. These are equivalent in practice,
because node properties only ever change through `load()` — but it is a real difference and worth a
reviewer's eye rather than being buried.

**Two deliberate choices in the generated Cypher**, both worth a look:

1. **The predicate reads `i.<property>`, not `item.<key>`.** A condition names a *node property*, and
   property names do not always match the source dict key (on `TrivyImageFindingNodeProperties`,
   `name` is fed by `PropertyRef("VulnerabilityID")`). Reading `i.<property>` sidesteps resolving the
   `PropertyRef` source key and preserves the old semantics exactly, since the deleted queries also
   matched on `n.<field>`. It matters concretely: three ECR schemas share the `ECRImage` label, and a
   load through one must not mislabel based on a key another one populates. It works because the
   `SET` clause has already been applied for the current row when the `FOREACH` executes.

2. **The removal clause inverts the branches of the same positive predicate, rather than negating
   it.** `i.x <> "v"` evaluates to `null` when `i.x` is missing, which would silently skip the removal
   and leave a stale label on a node whose condition property disappeared. Swapping `THEN [1] ELSE []`
   for `THEN [] ELSE [1]` is null-safe and keeps one predicate string per label.

**Placement.** The `FOREACH` clauses could not be appended inside `_build_node_properties_statement`:
the ontology statement that follows it starts with `",\n"` and continues the same `SET` clause. They
get their own template slot between the ontology statement and the relationship attach.

**A label declared both unconditionally and conditionally** is left entirely to the unconditional
`SET` clause. `ExtraNodeLabels` does not reject duplicate label names, and a negative `FOREACH` would
otherwise strip a label the `SET` clause had just applied. Guarded and tested.

**Blast radius.** The generated Cypher changes for all 11 schema files that declare conditional
labels, not just the two unscoped ones: `aibom/component.py`, `aws/ecr/image.py`,
`aws/inspector/findings.py`, `aws/ssm/parameters.py`, `gcp/artifact_registry/image.py`,
`github/container_images.py`, `github/dependabot_alerts.py`, `github/personal_access_tokens.py`,
`gitlab/container_images.py`, `semgrep/findings.py`, `tenable/findings.py`.

Docs in `docs/root/dev/writing-intel-modules.md` and the `add-node-type` skill reference both stated
that conditional labels are "applied in a separate query after ingestion"; both are updated.

No schema documentation or schema README changes are needed: which labels each schema declares, and
under what conditions, is unchanged. Only the mechanism that applies them changed.

### DIFF
--- a/cartography/graph/querybuilder.py
+++ b/cartography/graph/querybuilder.py
@@ -483,6 +483,98 @@ def _build_node_properties_statement(
     return set_clause
 
 
+def _build_conditional_labels_statement(
+    extra_node_labels: ExtraNodeLabels | None = None,
+) -> str:
+    r"""
+    Generate Neo4j clauses that apply conditional extra labels to the node of the current row.
+
+    Conditional labels are declared with ``ExtraNodeLabel.when(...)``. For each such label we emit a
+    pair of ``FOREACH`` clauses: one that adds the label when the conditions hold, and one that
+    removes it when they do not. ``FOREACH`` over a one-or-zero element list is the standard
+    pure-Cypher conditional-write idiom, so no APOC dependency is needed.
+
+    Applying the label row by row means a sync never has to scan a whole label to keep conditional
+    labels in sync: a node is either re-loaded (and corrected here) or no longer reported (and
+    deleted by cleanup).
+
+    Args:
+        extra_node_labels (ExtraNodeLabels | None): Extra labels declared on the node schema.
+            Defaults to None.
+
+    Returns:
+        str: Neo4j ``FOREACH`` clauses, or an empty string if there are no conditional labels.
+
+    Examples:
+        >>> extra_node_labels = ExtraNodeLabels([IMAGE.when(type="image")])
+        >>> print(_build_conditional_labels_statement(extra_node_labels))
+        FOREACH (_ IN CASE WHEN i.type = "image" THEN [1] ELSE [] END | SET i:Image)
+        FOREACH (_ IN CASE WHEN i.type = "image" THEN [] ELSE [1] END | REMOVE i:Image)
+
+    Note:
+        - The conditions name *node properties*, not source dict keys, so the predicate reads
+          ``i.<property>``. The preceding SET clause has already run for the current row, so the
+          value is up to date. This also keeps the semantics identical for schemas that share a
+          primary label but populate it from different source keys.
+        - Conditions within one label declaration are ANDed; several declarations of the same label
+          are ORed together.
+        - The negative clause inverts the branches of the same positive predicate rather than
+          negating it, so a node whose condition property is missing (``null`` predicate) still has
+          the stale label removed.
+        - A label declared both unconditionally and conditionally is left to the unconditional SET
+          clause; otherwise the negative clause would strip a label that was just applied.
+    """
+    if not extra_node_labels:
+        return ""
+
+    unconditional_labels = {
+        label.label for label in extra_node_labels.labels if not label.conditions
+    }
+
+    # Group condition sets by label name, preserving declaration order.
+    conditions_by_label: dict[str, list[tuple[tuple[str, str], ...]]] = {}
+    for label in extra_node_labels.labels:
+        if not label.conditions or label.label in unconditional_labels:
+            continue
+        conditions_by_label.setdefault(label.label, []).append(label.conditions)
+
+    if not conditions_by_label:
+        return ""
+
+    set_template = Template(
+        "FOREACH (_ IN CASE WHEN $predicate THEN [1] ELSE [] END | SET i:$label)",
+    )
+    remove_template = Template(
+        "FOREACH (_ IN CASE WHEN $predicate THEN [] ELSE [1] END | REMOVE i:$label)",
+    )
+
+    clauses = []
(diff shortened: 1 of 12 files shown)


cartography/graph/querybuilder.py
```diff
@@ -483,6 +483,98 @@ def _build_node_properties_statement(
...
+    remove_template = Template(
+        "FOREACH (_ IN CASE WHEN $predicate THEN [] ELSE [1] END | REMOVE i:$label)",
+    )
+
+    clauses = []
+    for label_name, condition_sets in conditions_by_label.items():
+        predicates = []
+        for conditions in condition_sets:
+            comparisons = [
+                f'i.{field_name} = "{_escape_cypher_string(str(field_value))}"'
```
kunaals, line 556: since these predicates now run against the already-bound `i` node, the condition-field indexes in `build_create_index_queries()` no longer speed up label matching. could we remove that block and update the two index tests so fresh installs don’t keep creating indexes solely for the deleted global queries?

kunaals (approved):

PR #3084, opened 2026-07-29
feat(supabase): add Supabase intel module

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds a `supabase` intel module reading the [Supabase Management API](https://supabase.com/docs/reference/api/introduction) (`https://api.supabase.com/v1`, personal access token).

Supabase is managed Postgres plus auth, storage and edge functions, so an organisation using it holds production databases, object storage buckets, service-role keys and serverless code that were previously invisible to Cartography. This module brings that estate into the graph and into the shared ontology, so Supabase resources are queryable alongside AWS, GCP, Scaleway and Vercel assets.

Tenancy follows the two-level shape already used by Scaleway: `SupabaseOrganization` (no sub-resource, root) -> `SupabaseProject` (sub-resource = organization) -> everything else (sub-resource = project). Both carry the `Tenant` ontology label, since a Supabase project is the isolation boundary in the same way an AWS account is.

16 node types:

| Area | Nodes | Ontology label |
| --- | --- | --- |
| Tenancy | `SupabaseOrganization`, `SupabaseProject` | `Tenant` |
| Identity | `SupabaseOrganizationMember` | `UserAccount` |
| Data | `SupabaseDatabase`, `SupabasePooler` | `Database` |
| Keys | `SupabaseApiKey`, `SupabaseSigningKey` | `APIKey` |
| Compute | `SupabaseEdgeFunction`, `SupabaseSecret` | `Function`, `Secret` |
| Storage | `SupabaseStorageBucket` | `ObjectStorage` |
| Network | `SupabaseCustomHostname` | `DNSRecord` |
| Auth | `SupabaseAuthConfig`, `SupabaseSSOProvider`, `SupabaseThirdPartyAuthIntegration` | `IdentityProvider` |
| Findings | `SupabaseSecurityAdvisorFinding` | `SecurityIssue` |
| Branching | `SupabaseBranch` | none |

Security-relevant posture is captured rather than just inventory: whether TLS is enforced on the database, the allowed CIDR list (an unrestricted project shows `0.0.0.0/0`), whether PITR is on, whether the legacy JWT `anon`/`service_role` keys are still accepted, which Postgres schemas PostgREST exposes publicly, whether an edge function skips JWT verification, whether a storage bucket is public, and Supabase's own security-advisor lints (RLS disabled on a public table, security-definer views).

Ten ontology mappings are registered, so for example a public Supabase bucket is reachable through the existing cross-provider query `MATCH (o:ObjectStorage) WHERE o._ont_public = true`.

**No secret material is read or stored.** Notably, `GET /v1/projects/{ref}/api-keys` returns the full `api_key` value for every key type *even when the `reveal` parameter is omitted*, including the `service_role` secret, contrary to what the parameter implies. The transform builds a fresh dict from an explicit field list, so the value is dropped and no node property can hold it; only the id, name, type, prefix and server-side hash are kept. Likewise dropped: the function-secret values, the PostgREST `jwt_secret`, the pooler connection strings (they embed credentials), and the auth captcha/SMTP/hook secrets. The pgsodium root-key endpoint and the saved-SQL-snippets endpoint are never called at all.

**Unreadable endpoints never delete anything.** Many endpoints are plan-gated or beta. A tolerated status means "we could not read this inventory", which is treated differently from a `200` returning an empty list meaning "this inventory is empty": only the latter drives a cleanup. Without that distinction, losing an entitlement, having a token scope revoked, or a transient `403` would silently wipe real keys, branches, buckets, secrets and findings from the graph. Stale data recovers on the next successful sync; deleted data does not.

### Related issues or links

None.

### How was this tested?

**Automated**

```
make test_lint                                                  # exit 0
uv run --frozen pytest tests/unit                               # 3934 passed
uv run --frozen pytest tests/integration                        # 1226 passed
uv run --frozen pytest tests/integration/.../intel/supabase/    # 42 passed
```

42 integration tests exercise `sync()` end to end with only the `get*` boundaries mocked, plus 15 unit tests for the HTTP tolerance rules and timestamp helpers. Beyond the happy path they cover the secret-redaction guarantees (including a check that no string property anywhere in the graph contains a fixture key value), the plan-gated paths, and the three tenancy/cleanup invariants described below.

**Live sync against a real Supabase project**

Run against a real organisation and project on the free tier, with a local Neo4j. Project ref redacted below.

```
INFO:cartography.sync:Starting sync stage 'supabase'
INFO:cartography.client.core.tx:Loaded 1 SupabaseOrganization nodes
INFO:cartography.client.core.tx:Loaded 1 SupabaseOrganizationMember nodes
WARNING:cartography.intel.supabase.util:Supabase returned 400 for https://api.supabase.com/v1/projects/<project-ref>/vanity-subdomain - skipping (feature unavailable on this plan).
INFO:cartography.client.core.tx:Loaded 1 SupabaseProject nodes
INFO:cartography.client.core.tx:Loaded 1 SupabaseDatabase nodes
INFO:cartography.client.core.tx:Loaded 4 SupabaseApiKey nodes
INFO:cartography.client.core.tx:Loaded 2 SupabaseSigningKey nodes
INFO:cartography.client.core.tx:Loaded 1 SupabaseEdgeFunction nodes
INFO:cartography.client.core.tx:Loaded 8 SupabaseSecret nodes
INFO:cartography.client.core.tx:Loaded 2 SupabaseStorageBucket nodes
INFO:cartography.client.core.tx:Loaded 1 SupabaseAuthConfig nodes
INFO:cartography.client.core.tx:Loaded 1 SupabasePooler nodes
WARNING:cartography.intel.supabase.util:Supabase returned 400 for https://api.supabase.com/v1/projects/<project-ref>/custom-hostname - skipping (feature unavailable on this plan).
INFO:cartography.sync:Finishing sync stage 'supabase'
```

Verified in the graph:

- All relationships resolved, including `(:SupabasePooler)-[:CONNECTS_TO]->(:SupabaseDatabase)` and the `RESOURCE` hierarchy.
- Real posture ingested: `ssl_enforced` false, `db_allowed_cidrs` `["0.0.0.0/0"]`, `pitr_enabled` false, `legacy_api_keys_enabled` true, PostgREST exposing `public,graphql_public`, org owner without MFA.
- Ontology labels and `_ont_*` properties applied (`Tenant`, `Database`, `APIKey`, `UserAccount`, and `ObjectStorage` / `Function` / `Secret` once those resources existed).
- **Idempotent**: two consecutive syncs left the node count unchanged and deleted nothing.
- **No secret leakage**: the 13 distinct real secret values reachable through the API (4 API keys, the JWT secret, the service_role key, the database URL) were fetched and grepped against every string property in the graph. Zero matches, and none of the forbidden property names exist.
- **Load and cleanup of real resources**: created a private bucket, a public bucket, an edge function and a function secret, confirmed each appeared with correct properties and edges, then deleted all four and confirmed the next sync removed them from the graph. The public bucket was reachable via `MATCH (o:ObjectStorage) WHERE o._ont_public = true`.

### Notes for reviewers

**Two places where the published OpenAPI spec does not match live behaviour.** Both were found by running the module, not by reading docs, and both are commented in the code:

1. Plan-gated endpoints answer **HTTP 400 with `error.code: "entitlement_required"`**, not 402/403/404. On a free-tier organisation `/vanity-subdomain` and `/custom-hostname` both do this, which aborted the entire sync on the first live attempt. `get_json()` matches on the error code rather than on the bare 400, so genuine bad requests still raise.
2. `/api-keys` returns the `api_key` value **without** `reveal`, as described in the summary.

**Three tenancy and cleanup invariants worth reviewing closely**, each with a regression test that was individually confirmed to fail against the pre-fix code:

1. API key node ids are always prefixed with the project ref. The API returns `anon` and `service_role` as the *actual* ids of the legacy keys, identical in every project, so an unprefixed id made two projects share one node with mutually overwriting metadata.
2. `SupabaseOrganizationMember` is a membership, not a person: the id is `<org slug>/<user id>` because `role_name` is per-organization. Keyed on `user_id` alone, the last organisation synced overwrote the member's role everywhere else. This mirrors `AWSUser` being scoped per account; the canonical `User` ontology node is what unifies the memberships.
3. Per-resource cleanups are skipped when the corresponding read was unavailable, as described in the summary.

**Known limitation.** The live run was performed before those three fixes landed, and I no longer hold a token for that account, so the fixes are covered by the regression tests rather than by a second live run. The API-shape findings above were verified live.

**Deliberately out of scope**, flagged rather than silently added: read replicas (the org-scoped projects endpoint exposes a `databases[]` array with `PRIMARY`/`READ_REPLICA` entries and compute size), JIT database grants, service health and upgrade eligibility, and a Supabase Fact for the existing `object_storage_public` rule. `SupabaseSigningKey`, `SupabaseEdgeFunction`, `SupabaseSSOProvider`, `SupabaseThirdPartyAuthIntegration` and `SupabaseBranch` keep their bare API-assigned ids; these are UUIDs, verified live only for signing keys and inferred from shape for the others. Applying the project-ref prefix uniformly as a defensive invariant would be a small follow-up if reviewers prefer it.

#### General

#### Proof of functionality

- [x] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md) (registered in `docs/root/usage/schema.md`, plus `docs/root/index.md` and `docs/root/module-list.md`).


cartography/models/supabase/organization.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@dataclass(frozen=True)
+class SupabaseOrganizationSchema(CartographyNodeSchema):
+    label: str = "SupabaseOrganization"
+    properties: SupabaseOrganizationNodeProperties = (
+        SupabaseOrganizationNodeProperties()
+    )
+    # The organization is the root of the Supabase hierarchy, so it deliberately
+    # has no sub_resource_relationship and is never cleaned up by a scoped job.
```
kunaals, line 32: question: `/v1/organizations` only returns organizations the token currently belongs to. if access to an organization is removed, it looks like that organization and its existing subtree are never touched again. should an unrestricted sync clean up stale organization roots, while keeping allowlisted syncs scoped?

cartography/intel/supabase/projects.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+@timeit
+def get(api_session: requests.Session, base_url: str) -> list[dict[str, Any]]:
+    """
+    List every project the token can see. This endpoint (unlike the org-scoped one)
+    carries `organization_slug`, `created_at` and the primary `database` object, so
+    it is the single source of truth for both the project and database nodes.
+    """
+    return get_json(api_session, f"{base_url}/v1/projects")
```
kunaals, line 188: i might be reading the api wording too literally, but `/v1/projects` is documented as returning projects the user previously created, while `/v1/organizations/{slug}/projects` lists projects for that organization. could this miss shared organization projects and then clean them up as if they were deleted? can we confirm the global endpoint includes every visible organization project, or use the organization-scoped endpoint?

cartography/intel/supabase/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+import cartography.intel.supabase.storage
+from cartography.config import Config
+from cartography.intel.supabase.util import build_session
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+
+@timeit
+def start_supabase_ingestion(neo4j_session: neo4j.Session, config: Config) -> None:
```
kunaals, line 22: nit: could we add a small orchestration test covering an organization disappearing between syncs? the lower-level module tests don’t exercise the root cleanup and fan-out behavior here.

kunaals (changes requested):

cartography/intel/supabase/projects.py
```diff
@@ -0,0 +1,477 @@
...
+        _all_fields(DATABASE_POSTURE_FIELDS),
+    )
+    transformed = transform_database(project, posture, stored)
+    load_databases(
+        neo4j_session,
+        transformed,
+        project["ref"],
+        common_job_parameters["UPDATE_TAG"],
+    )
+    cleanup_databases(neo4j_session, common_job_parameters)
```
kunaals, line 385: could we skip database cleanup when the project has no `database` enrichment? the organization endpoint can return projects that are missing from the creator-only `/v1/projects` response, so an empty transform here may mean "database details unavailable" rather than "database deleted." in that case cleanup could remove a previously ingested database node.

kunaals (approved):
one more thing to check and then its gtg

PR #3085, opened 2026-07-29
fix(rules): validate the Fact asset anchor and reserve Finding.source

jychp:
### Type of change

- [ ] Documentation update

### Summary

The rules spec accepts two `Fact` declarations that nothing validates, and both let a rule ship in a state no consumer can resolve. This adds the missing validation and fixes every existing violation.

**1. `Finding.source` was doubly defined.** `source` is documented as module provenance and is set by `Rule.parse_results` from `fact.module`. But six facts aliased `_ont_source AS source`, and the query row silently overwrote the module value, so one field carried two unrelated meanings depending on the fact. Nothing prevented the next fact from doing the same.

`Fact.__post_init__` now rejects a `cypher_query` that aliases the reserved `Finding` fields (`source`, `extra`). The affected facts return the per-row provider as `ontology_source` instead, declared on their output models and used in `identity_fields` where the provider was the uniqueness discriminator for a shared ontology label. Two related bugs surfaced while doing this:

- In `nist_ai_rmf`, `AS source` was a plain duplicate of the adjacent `AS app_source`, so it is simply dropped.
- In `mfa_missing`, the ontology fact also filled its `status` display column with the source module name (`a._ont_source AS status`), duplicating the same value twice in one row. It now returns `'no-mfa'`, matching its AWS sibling fact. The model field stays, since that sibling uses it legitimately.

**2. A fact could declare an `asset_label` its own query never matched.** `asset_label` + `asset_id_field` form the `(label, id)` anchor consumers use to locate the offending node, but nothing checked that the query guarantees the label. `openai_nist_ai_stale_or_unowned_api_keys` declared `asset_label="APIKey"` while matching `k:OpenAIApiKey OR k:OpenAIAdminApiKey`; six facts diverged this way. The rows a fact returns and the asset it claims could drift apart with no CI signal.

`Fact.__post_init__` now requires the query to match the declared label, and the six offenders are fixed:

| Fact | Before | After |
| --- | --- | --- |
| `openai_nist_ai_stale_or_unowned_api_keys` | `MATCH (k)` | `MATCH (k:APIKey)`, provider filter kept in `WHERE` |
| `gcp_bigquery_table_cmek_missing` | id derived from `table.dataset_id` | matches the `GCPBigQueryDataset` node it anchors on |
| `azure_trust_relationship_manipulation` | `labels(principal) IN [...]` | `MATCH (principal:EntraPrincipal)` |
| `azure_account_manipulation_permissions` | `labels(principal) IN [...]` | `MATCH (principal:EntraPrincipal)` |
| `azure_policy_manipulation_capabilities` | `labels(principal) IN [...]` | `MATCH (principal:EntraPrincipal)` |
| `scaleway_policy_manipulation_capabilities` | `labels(principal) IN [...]` | `MATCH (principal:ScalewayPrincipal)` |

The umbrella-label rewrites are equivalent by construction: `ENTRA_PRINCIPAL` is carried unconditionally by exactly `EntraUser` / `EntraGroup` / `EntraServicePrincipal`, and `SCALEWAY_PRINCIPAL` by exactly `ScalewayUser` / `ScalewayApplication` / `ScalewayGroup`. They also let Neo4j use the label index instead of scanning and filtering.

**3. Supporting changes.** A new unit test checks every fact's `asset_label` against the labels the declarative model actually writes, via a `tests.utils.all_graph_labels()` helper shared with the existing ontology rel-constraint guard. The allowlist for labels written only by legacy handwritten Cypher is currently empty: all 182 facts pass. The `RETURN`-alias regex that was duplicated in three places is now `cartography.rules.spec.model.returned_aliases()`.

Rule `version` fields are bumped wherever query semantics or the output model changed.

Both new checks live in `Fact.__post_init__` rather than in a test, matching the existing `asset_id_field` precedent: the rule registry fails to import on violation, so a non-compliant fact cannot reach a consumer. The label-registry check needs `cartography.models` reflection, so that one stays a unit test.

### Related issues or links

None.

### How was this tested?

- `make test_lint` passes.
- `make test_unit` passes: 4247 tests, 75% coverage.
- New tests: `tests/unit/rules/test_asset_label.py` (per-fact sweep of the registry, both checks) plus three negative cases in `tests/unit/rules/test_model.py` proving each guard raises on a non-compliant `Fact`.
- Against a populated local Neo4j: all 36 touched queries (`cypher_query` / `cypher_visual_query` / `cypher_count_query`) `EXPLAIN` cleanly, and every touched rule runs end-to-end through `cartography-rules run --output json` with exit 0.
- The `:APIKey` addition returns byte-identical rows to the previous query on that graph, and all 6 OpenAI keys present carry `:APIKey`.
- Output before/after for `unmanaged-account`, showing the two concepts now separated in the same payload:
  ```json
  {
    "source": "Cross-Cloud",
    "extra": {},
    "email": null,
    "id": "arn:aws:iam::REDACTED:user/terraform",
    "ontology_source": "aws"
  }
  ```

#### General

#### Proof of functionality

### Notes for reviewers

Two things worth a close look:

1. **The BigQuery rewrite is the one behavior change.** Anchoring on the real `GCPBigQueryDataset` node means a table whose dataset was not ingested now drops out of the result. That is correct for a dataset-anchored finding, but it is a visible change rather than a pure refactor. The id join is exact: dataset nodes use `f"{project_id}:{dataset_id}"` (`intel/gcp/bigquery_dataset.py`) and tables store the same composite in `dataset_id` (`intel/gcp/bigquery_table.py`). `dataset_name` now comes from the node instead of `split(dataset_id, ':')[-1]` string surgery.

2. **The Azure and Scaleway rewrites are not row-count verified.** The graph I tested against has no Entra or Scaleway data, so those four facts returned 0 rows before and after. They are backed by query planning plus the model source showing the umbrella labels are carried unconditionally by exactly the schemas each filter enumerated. A reviewer with Azure or Scaleway data in a graph could confirm parity quickly.

Also deliberate, and open to a different call:

- The sibling Anthropic key fact still anchors on the provider label `AnthropicApiKey` while the OpenAI one anchors on the shared `APIKey`. Converging them would change the label consumers resolve against, so it is left alone with a comment noting the inconsistency.
- `OntologyFieldMapping.required` is read nowhere in the codebase; ontology labels and `_ont_*` properties are applied unconditionally regardless of it. That is out of scope here, but it means `required=True` in the mapping data is documentation only, which is easy to misread as an enforced constraint.


cartography/rules/spec/model.py
```diff
@@ -13,6 +13,21 @@
 
 logger = logging.getLogger(__name__)
 
+# Matches the `RETURN expr AS alias` convention every fact query uses. This
+# over-approximates: intermediate `WITH x AS y` clauses match too, so the alias
+# set is only ever used for conservative "is this name produced at all" checks.
+_RETURN_ALIAS_RE = re.compile(r"\bAS\s+(\w+)", re.IGNORECASE)
```
kunaals: nit: i might be missing a constraint on fact queries, but this appears to collect aliases from intermediate `WITH` clauses too. i was able to get `asset_id_field` validation to pass when the alias wasn’t returned, and a temporary `WITH x AS source` was rejected. can we limit this check to the final `RETURN` projection?

cartography/rules/spec/model.py
```diff
@@ -307,22 +322,40 @@ def __post_init__(self) -> None:
...
-            re.findall(r"\bAS\s+(\w+)", self.cypher_query, re.IGNORECASE)
-        )
-        if self.asset_id_field not in returned_aliases:
+        aliases = returned_aliases(self.cypher_query)
+        if self.asset_id_field not in aliases:
             raise ValueError(
                 f"Fact '{self.id}' asset_id_field '{self.asset_id_field}' is not returned "
                 f"by its cypher_query (expected a '... AS {self.asset_id_field}' alias)."
             )
+        if not re.search(rf":\s*{re.escape(self.asset_label)}\b", self.cypher_query):
```
kunaals: question: this checks that the label and asset id alias both appear, but not that they refer to the same node. i was able to match an `:AWSUser` and return an `AWSRole` id as the asset id while still passing validation. should the validation tie the id expression to the labeled variable, or narrow the guarantee in the docs?

kunaals (approved):
approving with one nit to fix and a question

PR #3086, opened 2026-07-29
refactor(ontology): split Package into Package and PackageVersion

jychp:
### Type of change

- [ ] Documentation update

### Summary

The canonical ontology `Package` node is keyed on `normalized_id` =
`{type}|{namespace/}{normalized_name}|{version}`, so every node it creates models *one version of one
package*, not the package itself. The name is misleading, and there is no node that answers "which
versions of `lodash` do we have?" without string-splitting ids.

This PR renames that node to `PackageVersion` and introduces a new version-independent `Package`
node keyed on `{type}|{namespace/}{normalized_name}`, linked to its versions by `HAS_VERSION`:

```
(:Package:Ontology {id: 'pypi|h11'})-[:HAS_VERSION]->(:PackageVersion:Ontology {id: 'pypi|h11|0.14.0'})
(:PackageVersion)-[:DETECTED_AS]->(:TrivyPackage|:SyftPackage|:GitHubDependency|...)
(:PackageVersion)-[:DEPLOYED]->(:Image)
(:CVE|:SecurityIssue)-[:AFFECTS]->(:PackageVersion)
```

`Package` is a pure aggregation node: every derived edge (`DETECTED_AS`, `DEPLOYED`, `AFFECTS`,
`SHOULD_UPDATE_TO`, `DEPENDS_ON`) stays on `PackageVersion`, where it already lives and where it
follows the renamed node for free.

Implementation notes:

- `make_normalized_package_name_id()` sits next to `make_normalized_package_id()` in
  `cartography/intel/trivy/util.py`, reusing `parse_purl()` and `normalize_package_name()`, so both
  keys share one normalization implementation (PEP 503 for Python, lowercase elsewhere).
- `Package` nodes are derived by grouping the `PackageVersion` rows the ontology sync already reads.
  No new property is emitted on provider nodes, so the intel modules are untouched.
- The ontology rel constraints for `AFFECTS` and `DEPLOYED` are retargeted to `PackageVersion`, and a
  `Package -HAS_VERSION-> PackageVersion` constraint is added.
- `Package` stays a reserved primary label through a new `ONTOLOGY_CANONICAL_SCHEMAS` registry, since
  `ONTOLOGY_MODELS["packages"]` now maps to `PackageVersionSchema`. Provider schemas still cannot
  claim either label.
- Range-only dependencies (`GitHubDependency` without an exact version, `GitHubDependabotAlert`
  package fields) remain outside the ontology, exactly as before. Attaching them to the new
  version-independent `Package` is a natural follow-up but is deliberately out of scope here.

### Related issues or links

None.

### Breaking changes

Queries that match `(:Package)` and expect a version-bearing node must be updated to
`(:PackageVersion)`. The same applies to every edge that used to hang off `Package`
(`DETECTED_AS`, `DEPLOYED`, `AFFECTS`, `SHOULD_UPDATE_TO`, `DEPENDS_ON`). After this change,
`(:Package)` is the version-independent grouping node.

Migration path: `cartography/data/jobs/analysis/ontology_package_version_rename_migration.json` runs
at the start of the ontology sync, before the packages sync, and relabels legacy nodes in place:

```cypher
MATCH (n:Package:Ontology) WHERE n.version IS NOT NULL
WITH n LIMIT $LIMIT_SIZE
REMOVE n:Package SET n:PackageVersion
```

An in-place relabel preserves node identity and every attached edge, so nothing has to be rebuilt.
It is guarded on `:Ontology` (non-ontology `:Package` nodes belong to the existing legacy Trivy
migration) and on `version IS NOT NULL` (what distinguishes a legacy node from the new
version-independent `Package`), which makes it idempotent. Running it before the packages sync means
no version-independent `Package` node exists yet when it fires, so the guard cannot misfire.

### How was this tested?

- `make test_lint` passes (including mypy).
- `tests/unit`: 3892 pass. New cases cover `make_normalized_package_name_id()` (namespaced purl,
  purl without version, PEP 503 name, component fallback, agreement with the versioned key) and
  `transform_packages()` (versions collapse into one `Package`, ecosystems do not collide, purl
  namespace is preserved, rows without a derivable key are skipped).
- Integration suites for ontology, semgrep, trivy, syft, github, gitlab and socketdev pass (172
  tests). The ontology packages test now asserts the `Package` nodes and their `HAS_VERSION` edges
  after a sync.
- `tests/integration/cartography/intel/ontology/test_package_version_migration.py` runs the
  migration job against a real Neo4j and asserts the relabelled node keeps its `elementId` and its
  `DETECTED_AS` edge, and that a second run leaves both the new version-independent `Package` and the
  `PackageVersion` untouched.

#### General

#### Proof of functionality

### Notes for reviewers

- While rewriting the `Package` section of the ontology schema docs, I fixed pre-existing drift: the
  `DETECTED_AS` list omitted `GitHubDependency`, `GitLabDependency` and `SocketDevDependency`.
- `docs/root/modules/aws/schema.md` carried a stale `### Package` section whose field table described
  the old Trivy id format (`{version}|{name}`) while the listed edges are the ontology ones. It is
  now titled `PackageVersion` with the canonical id format.
- Left untouched on purpose: the same file documents
  `(Package, Dependency)-[DEPLOYED]->(AWSEC2Instance)`. No code in the repo creates that edge; it was
  already stale before this PR, so I would rather have it removed in a separate change than renamed
  here and made to look real.


cartography/intel/ontology/packages.py
```diff
@@ -21,16 +24,79 @@ def sync(
...
+            pkg_type=row.get("type"),
+        )
+        if package_id is None:
+            logger.debug(
+                "Skipping package version '%s': no versionless key could be derived.",
+                row.get("normalized_id"),
+            )
+            continue
+        pkg_type, _, name = package_id.partition("|")
+        namespace = None
```
kunaals: question: the purl spec treats `namespace` and `name` as separate components, so `pkg:npm/%40types/node` has namespace `@types` and name `node`. here we store name as `@types/node` while also storing namespace as `@types`. is the full scoped name intentional? if so, could we clarify that in the property docs?

cartography/intel/ontology/packages.py
```diff
@@ -21,16 +24,79 @@ def sync(
...
+        parsed_purl = parse_purl(row.get("purl"))
+        if parsed_purl:
+            namespace = parsed_purl["namespace"]
+        package = packages.setdefault(
+            package_id,
```
kunaals, line 83: nit: `namespace` looks dependent on which package version is processed first. i reproduced the same `npm|@types/node` package ending up with either `namespace: null` or `namespace: "@types"` when i reversed a purl-less row and a purl-backed row. can we preserve a non-null namespace from any grouped row and add a reversed-order test?

kunaals (approved):

PR #3021, opened 2026-07-13
feat(docs): generate schema docs from data model

jychp:
### Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [x] Documentation update
- [ ] Other (please describe):

### Summary

- Add a runtime data-model introspector that discovers node schemas, relationships, MatchLinks, generated ontology properties, indexes, and provenance without maintaining a duplicate registry.
- Add human-readable descriptions to `PropertyRef` and consume schema and relationship docstrings when rendering documentation.
- Generate missing module `schema.md` pages in the Sphinx staging directory while preserving manually authored pages. LastPass is the first migrated module and no longer commits derived schema documentation.

### Related issues or links

- Follow-up integration planned for the typed analysis effects in https://github.com/cartography-cncf/cartography/pull/2922

### How was this tested?

- `uv run pytest tests/unit/cartography/models/test_introspection.py tests/unit/cartography/models/test_schema_docs.py`
- `uv run pytest tests/unit/cartography/test_doc.py`
- `make test_lint`
- `uv sync --frozen --group doc && uv run ./docs/build.sh`
- Full model introspection completed with 740 nodes, 1750 relationships, and no diagnostics.

#### General
- [ ] I have read the [contributing guidelines](https://cartography-cncf.github.io/cartography/dev/developer-guide.html).
- [x] The linter passes locally (`make test_lint`).
- [x] I have added/updated tests that prove my fix is effective or my feature works.

#### Proof of functionality

- [ ] Screenshot showing the graph before and after changes.
- [x] New or updated unit/integration tests.

#### If you are adding or modifying a synced entity
- [ ] Included Cartography sync logs from a real environment demonstrating successful synchronization of the new/modified entity. Logs should show:
  - The sync job starting and completing without errors
  - The number of nodes/relationships created or updated
  - Example:
    ```
    INFO:cartography.intel.aws.ec2:Loading 42 EC2 instances for region us-east-1
    INFO:cartography.intel.aws.ec2:Synced EC2 instances in 3.21 seconds
    ```

#### If you are changing a node or relationship
- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules).
- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md).

[...]

PR #3087, opened 2026-07-29
fix(cloudflare): scope CloudflareDNSRecord to the account, not the zone

jychp:
### Type of change

- [ ] Documentation update

### Summary

`CloudflareDNSRecordSchema.sub_resource_relationship` pointed at `CloudflareZone`, an intermediate hierarchical parent. Every other node in the module points at `CloudflareAccount`, which is the module's only `Tenant`-labelled node.

`sub_resource_relationship` is also the cleanup scope (`scoped_cleanup` defaults to `True`), so DNS records were reconciled per zone while the rest of the module was reconciled per account, and "walk `RESOURCE` once to reach the tenant" needed a special case for DNS records. It also diverged from every other DNS provider in the repo:

| Node | sub-resource | zone edge |
|---|---|---|
| `GCPRecordSet` | `GCPProject` | `(:GCPDNSZone)-[:HAS_RECORD]->` |
| `ScalewayDnsRecord` | `ScalewayProject` | `(:ScalewayDnsZone)-[:HAS_RECORD]->` |
| `VercelDNSRecord` | `VercelTeam` | `(:VercelDomain)-[:HAS_DNS_RECORD]->` |
| **`CloudflareDNSRecord`** (before) | **`CloudflareZone`** | (was the sub-resource) |

The record's sub-resource is now `CloudflareAccount`. The zone link moves to `other_relationships` as `HAS_RECORD` (matching GCP and Scaleway), and the old zone `RESOURCE` edge is kept as a deprecated rel, following the [`AzureSQLDatabase`](https://github.com/cartography-cncf/cartography/blob/master/cartography/models/azure/sql/sqldatabase.py) precedent.

Two consequences are handled here:

- **Cleanup can no longer run once per zone.** Account-scoped cleanup inside the zone loop would delete the records of zones not yet synced in the current run and recreate them moments later, resetting `firstseen` on every sync. `dnsrecords.sync()` now takes the account's zones, collects their records in one pass, then does a single load and a single cleanup, mirroring `cartography/intel/scaleway/dns/dns.py`. The zone loop in `intel/cloudflare/__init__.py` existed only to call this sync, so it collapses.

- **Records ingested before this change only carry the zone edge**, so the account-scoped cleanup would never see them and they would linger in the graph forever. `cloudflare_dnsrecord_resource_edge_migration.json` MERGEs the account edge along the `account -> zone -> record` path, scoped by `$account_id`, copying the record's own `lastupdated` so stale records stay eligible for cleanup instead of being rescued by the new edge. It runs *before* the zone sync, whose cleanup would otherwise delete a stale zone and take the only path to its records with it. Same shape as `github_dependency_manifest_resource_edge_migration.json`; remove in v1.0.0.

### Related issues or links

Raised by the automated reviewer on #3021 as a P1 against `cartography/models/cloudflare/dnsrecord.py`. It is a pre-existing data model issue that the PR's diff made visible rather than something that PR introduced, so it is fixed separately here.

### Breaking changes

None. The `(:CloudflareZone)-[:RESOURCE]->(:CloudflareDNSRecord)` edge is still produced (deprecated, removed in v1.0.0), so consumers querying it keep working. Records gain a second incoming `RESOURCE` edge from `CloudflareAccount`; a consumer assuming exactly one incoming `RESOURCE` edge would now see two.

### How was this tested?

- `make test_lint` clean; `tests/unit` 3881 passed; `tests/integration/cartography/intel/cloudflare` and `tests/integration/cartography/data` passed.
- The demo seed was run against a local Neo4j to confirm the three edges are produced:
  ```
  {'name': 'simpson.corp', 'account': '37418d7e-...', 'zone_has_record': 'be68b067-...', 'zone_resource': 'be68b067-...'}
  {'name': 'www.simpson.corp', 'account': '37418d7e-...', 'zone_has_record': 'be68b067-...', 'zone_resource': 'be68b067-...'}
  ```
- The migration was verified against a hand-built pre-migration graph (a legacy record carrying only the zone edge, with a stale `lastupdated`):
  ```
  legacy record present before: 1
  after cleanup WITHOUT backfill (expect 1, unreachable): 1
  account edge after backfill: <Record lu=111111>
  after cleanup WITH backfill (expect 0): 0
  ```
  This is now covered by `test_cloudflare_dnsrecord_resource_edge_migration_cleans_legacy_records`, which also asserts a *current* legacy record survives the same pass, locking in that the backfilled edge does not rescue records the API stopped returning.
- Generated cleanup queries were inspected to confirm the new scoping:
  ```
  MATCH (n:CloudflareDNSRecord)<-[s:RESOURCE]-(:CloudflareAccount{id: $account_id})
  MATCH (n)<-[r:HAS_RECORD]-(:CloudflareZone)
  ```

No sync logs from a real environment: I have no Cloudflare tenant to sync against, so the seed and integration tests stand in for that.

#### General

#### Proof of functionality

- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md). (no such file in the repo)

### Notes for reviewers

- **Choice of `HAS_RECORD`.** There is no canonical `DNSZone -> DNSRecord` label in `ONTOLOGY_REL_CONSTRAINTS`, and the three existing providers disagree: `HAS_RECORD` (GCP, Scaleway), `HAS_DNS_RECORD` (Vercel), `MEMBER_OF_DNS_ZONE` (AWS, outward). I picked the majority. Worth deciding separately whether one of these should become a `RelConstraint`.
- **Out of scope:** the repo-wide sweep for other `sub_resource_relationship` targets that do not carry `Tenant`. Kubernetes is already a known, whitelisted case (the cluster is deliberately modelled as the tenant), so a sweep needs to distinguish deliberate choices from oversights.
- **Adjacent fragility:** `tests/integration/cartography/data/jobs/test_syntax.py` holds one shared `parameters` dict for every analysis job, so each new job introducing a parameter must remember to add it there or CI fails with `ParameterMissing`. This PR adds `account_id`. Might be worth deriving those parameters automatically.


cartography/data/jobs/analysis/cloudflare_dnsrecord_resource_edge_migration.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Cloudflare DNS record RESOURCE edge migration",
+  "statements": [
+    {
+      "__comment__": "DEPRECATED: temporary backward-compatibility migration. Earlier versions scoped CloudflareDNSRecord cleanup via RESOURCE from CloudflareZone (per-zone); the schema now uses RESOURCE from CloudflareAccount as the sub_resource. Backfill the RESOURCE edge for every existing CloudflareDNSRecord reachable from the currently-syncing CloudflareAccount via its zones, so records that the API no longer returns fall inside the account-scoped cleanup instead of lingering forever. lastupdated is copied from the record, which keeps stale records stale and therefore eligible for cleanup. Scoped by $account_id so multi-account refreshes don't redo the full graph each time. Idempotent (MERGE). Remove in v1.0.0.",
+      "query": "MATCH (a:CloudflareAccount {id: $account_id})-[:RESOURCE]->(:CloudflareZone)-[:RESOURCE]->(r:CloudflareDNSRecord) MERGE (a)-[rel:RESOURCE]->(r) ON CREATE SET rel.lastupdated = r.lastupdated",
```
kunaals, line 6: i think this misses one legacy case. before this change, removing a zone could `DETACH DELETE` the zone before dns record cleanup ran, leaving its records orphaned. those records no longer have the account → zone → record path this migration needs, so they will still live forever. could we clean up ownership-orphaned `CloudflareDNSRecord` nodes here and add a regression test for a zone removed before upgrade?

cartography/intel/cloudflare/dnsrecords.py
```diff
@@ -19,13 +20,17 @@ def sync(
...
     common_job_parameters: Dict[str, Any],
-    zone_id: str,
+    account_id: str,
+    zones: List[Dict[str, Any]],
 ) -> None:
-    dnsrecords = get(client, zone_id)
+    # DNS records are scoped to the account (the tenant), so all of the account's
+    # zones are collected before a single load and a single cleanup pass. Cleaning
+    # up per zone would delete the records of zones not yet synced in this run.
+    dnsrecords = transform_dnsrecords(client, zones)
```
kunaals: could we avoid building one account-wide record list here? the sdk auto-paginates, so a large account can hold every record from every zone in memory before the first neo4j batch. loading each fully fetched zone as we go, then running account cleanup once after every zone succeeds, keeps the same cleanup safety with bounded memory. could we also set a sensible `per_page` to reduce the number of api calls?

cartography/models/cloudflare/dnsrecord.py
```diff
@@ -27,17 +28,50 @@ class CloudflareDNSRecordNodeProperties(CartographyNodeProperties):
...
-# (:CloudflareDNSRecord)<-[:RESOURCE]-(:CloudflareZone)
+# (:CloudflareZone)-[:HAS_RECORD]->(:CloudflareDNSRecord)
 class CloudflareDNSRecordToZoneRel(CartographyRelSchema):
     target_node_label: str = "CloudflareZone"
     target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
-        {"id": PropertyRef("zone_id", set_in_kwargs=True)},
+        {"id": PropertyRef("zone_id")},
+    )
+    direction: LinkDirection = LinkDirection.INWARD
+    rel_label: str = "HAS_RECORD"
```
kunaals, line 63: small scope question: do we need to introduce `HAS_RECORD` in this fix? keeping the existing zone `RESOURCE` edge as an `other_relationship` already preserves the zone link while the new account `RESOURCE` edge fixes cleanup. since there isn’t a canonical dns-zone relationship yet, this would avoid a duplicate edge and a new public contract.

kunaals (changes requested):

cartography/models/cloudflare/dnsrecord.py
jychp: he dupe is temporary and killing it is the whole point. If the zone link stays RESOURCE, a record ends up with two incoming RESOURCE edges meaning different things (tenant vs hierarchical parent) forever, so "walk RESOURCE to find the tenant" returns both the zone and the account and you still have to filter on the label. That's the ambiguity the issue is about. With HAS_RECORD the redundancy only lasts until v1.0.0.

And HAS_RECORD isn't a new name, it's already what GCPRecordSet and ScalewayDnsRecord use for this same edge. So we go from three vocabularies to two, not four.

kunaals (approved):

Issue #3083, opened 2026-07-28
AWS CloudTrail: Support GitHub immutable OIDC subject claims


kunaals:
### Expected behavior

CloudTrail `AssumeRoleWithWebIdentity` events using either legacy or immutable GitHub OIDC subjects should create an `ASSUMED_ROLE_WITH_WEB_IDENTITY` relationship to the matching `GitHubRepository`.

### Actual behavior

GitHub now emits immutable subjects for new repositories, for example:

`repo:octo-org@123456/octo-repo@456789:pull_request`

`_extract_github_repo_from_username()` returns `octo-org@123456/octo-repo@456789`, while `GitHubRepoAssumeRoleWithWebIdentityMatchLink` matches `GitHubRepository.fullname`, which is `octo-org/octo-repo`. The MatchLink does not resolve, so the relationship is not created.

### To reproduce

```python
_extract_github_repo_from_username(
    "repo:octo-org@123456/octo-repo@456789:pull_request"
)
# actual: octo-org@123456/octo-repo@456789
# expected: octo-org/octo-repo
```

Normalize repository names from both legacy and immutable subject formats and add fixtures for each.

### Additional context

GitHub documents the immutable subject format here: https://docs.github.com/en/actions/reference/security/oidc#immutable-subject-claims

sanskar-singh-2403:
hey @kunaals  can i pick this up?

Issue #3078, opened 2026-07-27
AWS IAM: model trust-policy Condition keys on the role trust edge (OIDC sub/aud scoping)


Dashtid:
# AWS IAM: model trust-policy `Condition` keys on the role trust edge (OIDC `sub`/`aud` scoping)

## Description

`transform_role_trust_policies` reads only the `Principal` element of each `AssumeRolePolicyDocument` statement; the `Condition` element is dropped:

https://github.com/cartography-cncf/cartography/blob/cc86566f358186e80219caf51974fb79c502917e/cartography/intel/aws/iam.py#L574-L578

`role["AssumeRolePolicyDocument"]` is not read anywhere else in `cartography/`, so no part of the trust condition reaches the graph. This is distinct from the identity/resource policy path: `_transform_policy_statements` does serialize `Condition` onto `AWSPolicyStatement` ([iam.py#L799-L800](https://github.com/cartography-cncf/cartography/blob/cc86566f358186e80219caf51974fb79c502917e/cartography/intel/aws/iam.py#L799-L800)), and #2891 built on that to annotate permission edges with `has_condition` / `condition_keys` / `conditions`. The trust edge never got the equivalent, so there is now an asymmetry: a `CAN_READ` edge carries its condition metadata, while the `TRUSTS_AWS_PRINCIPAL` edge that governs who can obtain the role in the first place carries only `lastupdated`.

The practical consequence shows up on GitHub Actions / GitLab CI OIDC roles, where `sub` is the entire access-control decision. Two roles that differ only in `sub` scoping produce isomorphic subgraphs:

| | `gha-pinned` | `gha-org-wide` |
|---|---|---|
| Principal | `oidc-provider/token.actions.githubusercontent.com` | identical |
| Operator | `StringEquals` | `StringLike` |
| `...:aud` | `sts.amazonaws.com` | identical |
| `...:sub` | `repo:octo-org/octo-repo:ref:refs/heads/main` | `repo:octo-org/*` |
| Graph | `(:AWSRole)-[:TRUSTS_AWS_PRINCIPAL]->(:AWSFederatedPrincipal)` | identical |

[...]

PR #3088, opened 2026-07-30
feat(aws): model trust-policy conditions on the role trust edge (#3078 tier 1)

Dashtid:
### Type of change
- [x] Documentation update

### Summary

Tier 1 of #3078, in the shape agreed on the issue: option 3 (MatchLink), one aggregated row per `(role, principal)`, unconditional-allow-wins as in #2891. Tier 2 (the derived `trust_subject_*` / `trust_audiences` properties) is deliberately left for a follow-up.

**The gap.** `transform_role_trust_policies` read only the `Principal` element of each `AssumeRolePolicyDocument` statement, so the trust `Condition` never reached the graph. Two roles differing only in their OIDC `sub` scoping produced isomorphic subgraphs:

| | `gha-pinned` | `gha-org-wide` |
|---|---|---|
| Principal | `oidc-provider/token.actions.githubusercontent.com` | identical |
| `...:sub` | `StringEquals repo:octo-org/octo-repo:ref:refs/heads/main` | `StringLike repo:octo-org/*` |
| Graph before | `(:AWSRole)-[:TRUSTS_AWS_PRINCIPAL]->(:AWSFederatedPrincipal)` | identical |

The second is satisfied by every repository under `octo-org` — IAM's `*` is not path-aware, so it spans both `/` and `:` — including repositories created after the role, and including the `pull_request` subject. Reachability scored them the same.

**Two commits**, deliberately separate:

**1. `feat(aws): model trust-policy conditions on the role trust edge`** — additive. The trust edge now carries the property names #2891 established for permission edges:

- `has_condition` — true only when every statement trusting the principal is gated by a `Condition`; an unconditional statement anywhere wins.
- `condition_keys` — the sorted union of referenced context keys.
- `conditions` — the raw operator maps as a JSON string.

`parse_condition_blob` and `extract_condition_context_keys` are reused as-is. AWS evaluates conditions at request time, so edges are annotated, never resolved.

The relationship moves off `AWSRoleSchema` and becomes a MatchLink. A node-schema relationship with a `one_to_many` target matcher compiles to a single `principal.arn IN $list` clause, so every trust edge from a role would share one set of property values — but conditions are per-statement. The MatchLink emits one aggregated row per `(role, principal)`, matching the shape `STSAssumeRoleAllowMatchLink` and `GitHubRepoAssumeRoleWithWebIdentityMatchLink` already use.

This commit also fixes a crash: `statement["Principal"]` was indexed directly, so a `NotPrincipal`-only statement — legal in a trust policy — raised `KeyError` and aborted the whole account sync.

**2. `fix(aws): only draw role trust edges where the statement grants assume-role`** — this addresses the review point that the transform ignored `Effect` and `Action`. Verified against `master` before writing it:

```python
# Effect ignored: an explicit Deny drew the same edge as an Allow
{"Effect": "Deny", "Principal": {"AWS": ".../attacker"}, "Action": "sts:AssumeRole"}
# -> trusted_aws_principals: ['arn:aws:iam::999999999999:role/attacker']

# Action ignored: sts:TagSession cannot obtain credentials, but was
# indistinguishable from sts:AssumeRoleWithWebIdentity
{"Effect": "Allow", "Principal": {"Federated": "...githubusercontent.com"}, "Action": "sts:TagSession"}
# -> trusted_aws_principals: ['...oidc-provider/token.actions.githubusercontent.com']
```

Unlike conditions, `Effect` is resolvable at sync time, so this filters rather than annotates, following the explicit-deny-wins precedence of `evaluate_policy_for_permissions` (not callable directly here — trust policies match on `Principal`, not `Resource`). Deliberate choices, all covered by tests:

- A `Deny` suppresses the edge only when **unconditional**. A conditional `Deny` applies at request time and cannot be resolved statically, so it is left to the `Allow` statements.
- `Deny` statements no longer materialize federated/service principals or external account stubs, since a `Deny` is not evidence the principal is trusted.
- Action matching is case-insensitive and wildcard-aware: `sts:AssumeRole*`, `sts:*` and `*` all qualify.
- `NotAction` is treated as granting. Resolving it means enumerating the complement of the action namespace, and guessing wrong would silently drop a real trust edge.

**This is the only commit in the series that removes edges**, which is a different review risk from adding properties — hence keeping it separate. Happy to split it into its own PR if you would rather land tier 1 alone. No test fixture in the repo loses an edge.

### Related issues or links

- Addresses #3078 (tier 1; tier 2 tracked there, so this does not auto-close it)
- Follows the condition-annotation precedent from #2891 / #2250
- The immutable-subject interaction raised on #3078 is #3083. Worth noting for whoever takes tier 2: `repo:octo-org/*` does **not** match `repo:octo-org@123456/octo-repo@456789:...`, since the literal prefix `repo:octo-org/` fails against `repo:octo-org@123456/`. The same wildcard is over-broad for legacy-format repos and fail-closed for immutable-format ones, which is why tier 2 should keep the raw pattern rather than reduce it to a single "is wildcard" boolean.

### Breaking changes

No schema-breaking change: the relationship label, direction, and endpoints are unchanged, and existing queries over `(:AWSRole)-[:TRUSTS_AWS_PRINCIPAL]->(:AWSPrincipal)` keep working. Commit 2 does change which edges exist — `Deny`-only and non-assume-action statements no longer produce one — which is the intended correctness fix.

### How was this tested?

Unit — 26 new tests in `tests/unit/cartography/intel/aws/iam/test_role_trust_conditions.py`: the two differently scoped OIDC trusts staying distinguishable, unconditional-allow-wins, one row per `(role, principal)`, cross-statement condition aggregation, the `SAML:aud` condition already present in the shared fixture, `"Condition": {}` not counting as a condition, unparseable blobs failing safe toward "conditional", the `NotPrincipal` skip, Deny precedence, conditional-Deny, and a parametrized action-matching matrix.

Integration — 3 new tests in `tests/integration/cartography/intel/aws/iam/test_iam.py` against Neo4j: the condition properties reaching the edge and telling the two roles apart, `has_condition=false` for the mixed role, and stale trust edges being removed by the MatchLink cleanup on a later update tag.

Suites run locally against Neo4j 5-community:
- `tests/unit` — 4316 passed. (4 failures are pre-existing on `master` on this Windows host: cp1252 decode and path handling in `test_label_migrations`, `test_object_store`, `test_doc`. Confirmed by stashing the branch and re-running.)
- `tests/integration/cartography/intel/aws/` and `tests/integration/rules/` — 327 passed, including `test_iam_role_external_account_trust`, `test_identitycenter` and `test_cloudtrail_management_events`, which are the existing consumers of this edge.
- `pre-commit run` over the changed files — black, isort, flake8, mypy, pyupgrade all pass.

One note on the existing `test_load_roles_creates_trust_relationships`: it passes in file order but fails when run in isolation, on `master` as well as on this branch, because it depends on an earlier test in the file having created the current account's root principal. Left alone here rather than folded into this PR.

#### General

#### Proof of functionality

- [x] Updated the schema documentation (`docs/root/modules/aws/schema.md`: the new edge properties, their aggregation semantics, an example Cypher query for org-wide OIDC trusts, and the Effect/Action rules).
- [ ] Updated the schema README — `docs/schema/README.md` does not exist on `master`; the AWS schema doc above appears to be the current location.

### DIFF
--- a/cartography/intel/aws/iam.py
+++ b/cartography/intel/aws/iam.py
@@ -1,4 +1,5 @@
 import enum
+import fnmatch
 import json
 import logging
 import time
@@ -560,6 +561,50 @@ def transform_access_keys(
     return access_key_data
 
 
+# The sts actions that let a principal actually obtain credentials for a role. A trust
+# statement granting only e.g. sts:TagSession does not make the role assumable.
+ASSUME_ROLE_ACTIONS = (
+    "sts:assumerole",
+    "sts:assumerolewithwebidentity",
+    "sts:assumerolewithsaml",
+)
+
+
+def _action_matches_assume_role(action: str) -> bool:
+    """Whether an Action entry covers at least one assume-role action.
+
+    IAM action names are case-insensitive and may use `*` and `?` wildcards, so
+    "sts:AssumeRole*", "sts:*" and "*" all qualify.
+    """
+    pattern = action.lower()
+    return any(
+        fnmatch.fnmatchcase(assume_action, pattern)
+        for assume_action in ASSUME_ROLE_ACTIONS
+    )
+
+
+def _statement_grants_assume_role(statement: dict[str, Any]) -> bool:
+    """Whether a trust statement's Action element covers an assume-role action.
+
+    NotAction is deliberately treated as matching. Resolving it correctly means
+    enumerating the complement of the action namespace, and guessing wrong here would
+    silently drop a real trust edge, so we keep the edge and let the operator inspect it.
+    """
+    if "NotAction" in statement:
+        logger.warning(
+            "Trust statement uses NotAction, which is not modeled. Treating it as "
+            "granting assume-role rather than dropping the trust edge."
+        )
+        return True
+
+    actions = ensure_list(statement.get("Action") or [])
+    return any(
+        _action_matches_assume_role(action)
+        for action in actions
+        if isinstance(action, str)
+    )
+
+
 def _aggregate_trust_conditions(
     statement_conditions: list[Any],
 ) -> dict[str, Any]:
@@ -631,6 +676,9 @@ def transform_role_trust_policies(
         # A role can trust the same principal from several statements under different
         # conditions, so these are aggregated per (role, principal) below.
         trusted_principal_conditions: dict[str, list[Any]] = {}
+        # Principals named by an explicit Deny. Deny beats Allow in IAM evaluation, so
+        # these draw no trust edge regardless of what the Allow statements say.
+        denied_principals: set[str] = set()
 
         for statement in role["AssumeRolePolicyDocument"]["Statement"]:
             if "Principal" not in statement:
@@ -644,8 +692,26 @@ def transform_role_trust_policies(
                 )
                 continue
 
+            if not _statement_grants_assume_role(statement):
+                # e.g. a statement granting only sts:TagSession. It names a principal but
+                # does not let it obtain credentials for the role.
+                continue
+
+            is_deny = statement.get("Effect") == "Deny"
             condition = statement.get("Condition")
             principal_entries = _parse_principal_entries(statement["Principal"])
+
+            if is_deny:
+                # An unconditional Deny settles the question: that principal cannot
+                # assume the role, so no edge. A conditional Deny only applies when its
+                # condition holds at request time, which we cannot evaluate here, so it
+                # is left to the Allow statements. Either way a Deny is not evidence
+                # that the named principal or its account exists, so no nodes are
+                # created from it.
+                if not condition:
+                    denied_principals.update(arn for _, arn in principal_entries)
+                continue
+
             for principal_type, principal_arn in principal_entries:
                 if principal_type == "Federated":
                     # Add this to list of federated nodes to create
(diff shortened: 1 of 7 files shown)


Dashtid:
Rebased onto `b49f163d` — the branch had fallen 48 commits behind and `mergeStateStatus` was `BEHIND`, so this was unmergeable regardless of review. Head is now `f7071dd4`.

Current state, in case it is useful for triage: `sphinx` and `Test Suite` are both sitting at `action_required` on attempt 1, so no CI has executed on this head. The last run to actually execute was the one approved on 2026-08-23, which surfaced a real failure in `test_relationship_docstrings_are_user_facing_descriptions` — my docstring led with a Cypher pattern and the word `one_to_many`, both on that test's forbidden list since the generated-docs work in #3021. That is fixed; the docstring now reads as a user-facing description with the modelling rationale moved to a comment above the class.

Verified locally against the rebased tree: `tests/unit -k "iam or trust or role"` 433 passed, `tests/unit/cartography/models/` 56 passed. The one remaining local failure, `test_documentation_schema_links_target_generated_headings`, is a Windows cp1252 `UnicodeDecodeError` and fails identically on clean `origin/master` — checked in a throwaway worktree rather than assumed.

Is there anything outstanding on the design side, or is this just waiting on a workflow approval?

PR #3089, opened 2026-07-30
feat(modal): add Modal intel module

jychp:
### Type of change

- [ ] Documentation update

### Summary

Adds a new intel module for [Modal](https://modal.com), a serverless compute platform, so a
Modal workspace's tenancy, identity, compute, storage and network surface lands in the graph
and joins the cross-provider ontology.

25 node types across four areas:

| Area | Nodes |
|---|---|
| Tenancy | `ModalWorkspace`, `ModalEnvironment` (both `Tenant`) |
| Identity & access | `ModalWorkspaceMember` (`UserAccount`), `ModalWorkspaceRole` / `ModalEnvironmentRole` (`PermissionRole`), `ModalServiceUser` (`ServiceAccount`), `ModalApiToken` / `ModalProxyToken` (`APIKey`) |
| Compute | `ModalApp` (`ComputeService`), `ModalFunction` (`Function`), `ModalClass`, `ModalSandbox` (`Container`), `ModalSandboxTunnel`, `ModalTask` (`ComputePod`), `ModalCluster`, `ModalImage` |
| Storage & network | `ModalSecret` (`Secret`), `ModalVolume` / `ModalNetworkFileSystem` (`FileStorage`), `ModalDict`, `ModalQueue`, `ModalDomain`, `ModalDomainDNSRecord`, `ModalProxy`, `ModalProxyIP` |

Beyond `RESOURCE`, the module contributes the canonical ontology edges `WORKLOAD_PARENT`
(Container/ComputePod/Function → ComputeService), `HAS_ROLE`
(UserAccount/ServiceAccount → PermissionRole) and `OWNED_BY` (APIKey → ServiceAccount), so
Modal participates in existing cross-provider rules. **No `LEGACY_REL_WHITELIST` entry was
needed.**

#### Why the module talks to a private protocol

Modal publishes no REST management API: everything is gRPC behind the Python client, and a
large part of the inventory (apps, app layout, sandbox detail, NFS, image tags, tasks,
clusters, service users, richer member/environment records, proxies, domains, environment role
principals) is reachable only through Modal's internal protocol. All of it is isolated in
`cartography/intel/modal/util.py`, the single import site for `modal` / `modal_proto`, so a
protocol break requires changes in one file. The dependency is pinned `modal>=1.5.0,<2` for
that reason.

Modal *does* document public listing helpers (`X.objects.list()`, `Sandbox.list`,
`get_app_objects`), but they are deliberately unused: they are wrapped by
`synchronizer.create_blocking` and **deadlock** when called from our own event loop with
`RPC request ... made outside of task context`. Reproduced against modal 1.5.3. The module
runs one `asyncio.run()` over async resource syncs, mirroring
`cartography/intel/microsoft/entra`.

### Related issues or links

- Fixes #

### How was this tested?

**Automated:** 1865 unit tests and 34 new Modal integration tests pass; `make test_lint`
(including mypy) is clean. Integration tests exercise `sync()` with only the module's own
`get_*` / `list_*` functions mocked, using fixtures whose shapes were captured from a real
Modal workspace.

**Live, end to end**, against a real Modal workspace with a throwaway `cartography-e2e`
environment (created and torn down): the CLI ingests the workspace, environments, member,
both role sets, apps, functions, class, sandbox, tunnel, task, secret, volume, Dict and Queue,
applies every ontology label with its `_ont_*` properties as native temporal types, and stores
no secret value. Sync logs below.

Specifically verified on live data:
- the web endpoint's `web_url` is populated while the plain function's is null
- `HAS_METHOD` resolves through Modal's `Worker.*` class-service-function naming
- the sandbox's derived state, `ResourceValue` unwrapping (`memory_mb: 128`,
  `milli_cpu: 125`) and encrypted-vs-unencrypted tunnel flag
- `CREATED_BY` resolves on `display_name`, and a creator with no matching member yields no edge
- deleting the environment in Modal cascades cleanup to all its scoped resources, leaving no
  orphans, while the surviving environment is untouched

Not exercised against the live API, only against fixtures: `ModalDomain`,
`ModalDomainDNSRecord`, `ModalProxy`, `ModalProxyIP` and `ModalNetworkFileSystem`. Custom
domains need a paid add-on (the API answers `UNIMPLEMENTED`, which the adapter treats as "no
domains"), and the test workspace had no proxies or legacy shares. Their fixtures come from the
protobuf definitions.

<details>
<summary>Cartography sync logs from a live Modal workspace</summary>

Compute:

```
INFO:cartography.client.core.tx:Loaded 1 ModalWorkspace nodes
INFO:cartography.intel.modal:Syncing Modal workspace jeremy-78340 (ac-DyLbE2VtEfgvSEhzMQAOcP)
INFO:cartography.client.core.tx:Loaded 2 ModalEnvironment nodes
INFO:cartography.client.core.tx:Loaded 3 ModalWorkspaceRole nodes
INFO:cartography.client.core.tx:Loaded 1 ModalWorkspaceMember nodes
INFO:cartography.intel.modal:Syncing Modal environment cartography-e2e
INFO:cartography.client.core.tx:Loaded 3 ModalEnvironmentRole nodes
INFO:cartography.client.core.tx:Loaded 1 (ModalWorkspaceMember)-[HAS_ROLE]->(ModalEnvironmentRole) relationships
INFO:cartography.client.core.tx:Loaded 2 ModalApp nodes
INFO:cartography.client.core.tx:Loaded 1 ModalSandbox nodes
INFO:cartography.client.core.tx:Loaded 1 ModalSandboxTunnel nodes
INFO:cartography.client.core.tx:Loaded 1 ModalClass nodes
INFO:cartography.client.core.tx:Loaded 3 ModalFunction nodes
INFO:cartography.client.core.tx:Loaded 1 ModalTask nodes
INFO:cartography.sync:Finishing sync stage 'modal'
```

Storage:

```
INFO:cartography.client.core.tx:Loaded 1 ModalSecret nodes
INFO:cartography.client.core.tx:Loaded 1 ModalVolume nodes
INFO:cartography.client.core.tx:Loaded 1 ModalDict nodes
INFO:cartography.client.core.tx:Loaded 1 ModalQueue nodes
```

Resulting relationships, and Modal's contribution to the cross-provider workload chain:

```
(:ModalWorkspace)-[:RESOURCE]->(:ModalEnvironment)          x2
(:ModalEnvironment)-[:RESOURCE]->(:ModalApp)                x2
(:ModalFunction)-[:WORKLOAD_PARENT]->(:ModalApp)            x3
(:ModalClass)-[:WORKLOAD_PARENT]->(:ModalApp)               x1
(:ModalSandbox)-[:WORKLOAD_PARENT]->(:ModalApp)             x1
(:ModalTask)-[:WORKLOAD_PARENT]->(:ModalApp)                x1
(:ModalClass)-[:HAS_METHOD]->(:ModalFunction)               x1
(:ModalSandbox)-[:EXPOSES]->(:ModalSandboxTunnel)           x1
(:ModalWorkspaceMember)-[:HAS_ROLE]->(:ModalWorkspaceRole)  x1
(:ModalWorkspaceMember)-[:HAS_ROLE]->(:ModalEnvironmentRole) x1

(:ComputePod)-[:WORKLOAD_PARENT]->(:ComputeService)  x1
(:Container)-[:WORKLOAD_PARENT]->(:ComputeService)   x1
(:Function)-[:WORKLOAD_PARENT]->(:ComputeService)    x3
```

Environment-deletion cascade:

```
INFO:cartography.intel.modal.environments:Modal environment en-PzdwLykk6h5U7QeiQ1Pzmv no longer exists; removing its scoped resources.
```

</details>

#### General

#### Proof of functionality

- [x] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md) — registered in `docs/root/usage/schema.md`, `docs/root/index.md` and `docs/root/module-list.md`.

### Notes for reviewers

Reviewing in commit order should be easiest: wiring and adapter, then tenancy and identity, the
cascade fix, compute, then storage and network.

**Where I would focus review:**

1. **`cartography/intel/modal/util.py`** is the load-bearing file. It converts everything to
   plain dicts so no other module or test ever sees protobuf, which is what makes the
   `get_*`-only mocking boundary possible.

2. **Two-level tenancy.** Environment-scoped objects sub-resource to `ModalEnvironment`, not to
   the workspace, because every Modal listing call is keyed by environment: workspace-scoping
   would make `--modal-environments=prod` wipe `dev`. `ModalWorkspace` itself intentionally has
   no sub-resource relationship and never runs a cleanup, since a global cleanup would delete a
   sibling workspace ingested by a second token into the same graph.

3. **The environment-deletion cascade** (`65e763f92`, found in review of the first commit).
   Environment-scoped cleanups reach their targets *through* the environment node, and the entry
   point only iterates environments still present in the API listing, so a deleted environment
   used to leave its children behind as orphans that still read as live. `environments.sync()`
   now diffs the graph against the listing and cascades **before** deleting the nodes; ordering
   matters. Extension point: every environment-scoped resource must expose
   `cleanup_for_environment()` and be registered in `_ENVIRONMENT_SCOPED_CLEANUPS`.

4. **Partial-enumeration safety.** Functions and tasks fan out per app, so if any app fails,
   cleanup is skipped for the whole environment rather than deleting workloads that were merely
   not seen. Sandboxes do the same when pagination does not complete. Both directions are
   tested, including that a *complete* run does still delete.

5. **Labels deliberately withheld**, each with a rationale in the model and a test where it
   matters: `ModalImage` is not `Image` (a Modal image id is not a digest, so it could never
   join a registry image and would pollute `RESOLVED_IMAGE`); `ModalCluster` is not
   `ComputeCluster`; `ModalEnvironment` is not `ComputeNamespace` (its constraints would clash
   with the `RESOURCE` edge); `ModalDomainDNSRecord` is not `DNSRecord` (desired config, not
   observed DNS state); `ModalDict` / `ModalQueue` stay provider-only.

**API gaps worth knowing about, documented in `config.md` and the schema doc.** These are
Modal limitations, not omissions:

- Secret **values** are never returned by any read API, so none are stored.
- `Function.secret_ids`, `volume_mounts` and `proxy_id` are **write-only**, so
  `USES_SECRET`, volume-mount and proxy-routing edges **cannot be built at all**. Secret
  consumption is only determinable from source code.
- `requires_proxy_auth` is **not readable**, so Cartography records a web endpoint's URL but
  cannot say whether it is protected. Treat any non-null `ModalFunction.web_url` as potentially
  unauthenticated.
- A deployed function's GPU, CPU, memory, timeout, region and cloud are likewise write-only and
  therefore absent. Sandboxes *do* expose resources, regions and tunnels.
- Modal has **no read-only API token scope**, so the workspace node records which credential
  performed the sync (`synced_with_principal_type`), and the docs recommend a dedicated service
  user with `viewer` environment roles.

**Follow-up, deliberately out of scope:** promoting `ModalProxyIP` to the canonical `PublicIP`
would mean editing shared ontology model files, which does not belong in a new-provider change,
so it is left as a plain provider node. Billing/cost data and app deployment history are also
out of scope.


pyproject.toml
```diff
@@ -108,6 +108,10 @@ dependencies = [
     "azure-synapse-artifacts>=0.17.0",
     "slack-sdk>=3.37.0",
     "workos>=8.0.0",
+    # Modal ships no public REST API, so the intel module talks to the gRPC service
+    # through this client. Upper-bounded because the module also uses Modal's internal
+    # protocol (see cartography/intel/modal/util.py), which is unversioned.
+    "modal>=1.5.0,<2",
```
kunaals: i think the minimum modal version needs to be 1.5.3 here. `EnvironmentGetRolesRequest` isn’t available in 1.5.0–1.5.2, so those allowed versions fail when the sync reaches environment roles.

cartography/intel/modal/util.py
```diff
@@ -0,0 +1,876 @@
...
+    """List workspace members.
+
+    Uses the private `WorkspaceMembersList` rather than `Workspace.members.list()`: the
+    public dataclass drops `identity_provider_type` and `idp_external_id`, which are the
+    fields that make non-SSO members in an SSO workspace queryable.
+    """
+    response = await retry_transient_errors(client.stub.WorkspaceMembersList, Empty())
+    return [
+        {
+            "id": item.user_id,
```
kunaals, line 291: should this use the membership id, or maybe a workspace/user composite? the same user can belong to multiple workspaces, and `user_id` collapses their workspace-specific roles and membership data into one node.

cartography/intel/modal/util.py
```diff
@@ -0,0 +1,876 @@
...
+            client.stub.ImageListTags,
+            api_pb2.ImageListTagsRequest(
+                environment_name=environment_name,
+                page_token=page_token,
+            ),
+        )
+        for item in response.items:
+            images.append(
+                {
+                    "id": item.image_id,
```
kunaals, line 814: this loses aliases when the same image is published under multiple names or tags. since every row merges on `image_id`, only the last tag survives. we should preserve all of them somehow.

cartography/intel/modal/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+            environment_job_parameters,
+        )
+
+        for label, module in (
+            ("secrets", cartography.intel.modal.secrets),
+            ("volumes", cartography.intel.modal.volumes),
+            ("network file systems", cartography.intel.modal.nfs),
+            ("dicts", cartography.intel.modal.dicts),
+            ("queues", cartography.intel.modal.queues),
+            ("proxies", cartography.intel.modal.proxies),
```
kunaals: this looks like the same workspace-wide `ProxyList` request runs once for every environment. it would be better to fetch it once before the loop and partition the results by environment.

cartography/intel/modal/util.py
```diff
@@ -0,0 +1,876 @@
...
+    before_timestamp = 0.0
+    complete = False
+    for _ in range(_SANDBOX_PAGE_LIMIT):
+        request = api_pb2.SandboxListRequest(
+            environment_name=environment_name,
+            include_finished=False,
+        )
+        if before_timestamp:
+            request.before_timestamp = before_timestamp
+        response = await retry_transient_errors(client.stub.SandboxList, request)
```
kunaals: this only covers legacy sandboxes. modal says v2 sandboxes aren’t returned by this list, so v2 workloads and their tunnels will be missing from the graph.

kunaals (changes requested):

kunaals (approved):

PR #3090, opened 2026-07-30
fix(ontology,route53): stop deleting route53-owned DNS_POINTS_TO edges and make ELB alias targets match

jychp:
### Type of change

- [x] Documentation update

### Summary

Makes `(:AWSDNSRecord)-[:DNS_POINTS_TO]->(:AWSLoadBalancerV2)` reliable. Two independent
problems were breaking it, one causing the edge to flap and the other keeping it absent
entirely. One commit per concern.

**1. The ontology DNS linking cleanup deleted route53-owned edges.**

Each generated `ontology_dnsrecords_<target>` job has two statements: one for generic
`DNSRecord` sources, one for `GCPRecordSet`, which needs `UNWIND` over its list-valued
`data`. The second declared `source_label="DNSRecord"` even though its match is
`(dns:GCPRecordSet)`.

Cleanup effects dedupe by frozen-dataclass equality and `cleanup_where` is a field of
`RelationshipEffect`, so the guarded and unguarded effects were not equal, both survived,
and the job emitted two deletes for the same edge:

```
MATCH (source:DNSRecord)-[r:DNS_POINTS_TO]->(target:AWSLoadBalancerV2)
WHERE r.lastupdated <> $UPDATE_TAG
WITH r LIMIT $LIMIT_SIZE
DELETE r
```

That one deleted every route53-owned edge whenever the ontology jobs ran with a different
update tag than the AWS sync that wrote them, which is the case when intel and analysis run
as separate cartography processes. The next AWS sync recreated them, so the edge appeared to
flap. Three targets were affected: `AWSLoadBalancerV2`, `AWSLoadBalancer`, `AWSEC2Instance`.

This is a regression of #2987, which fixed the same thing in
`cartography/data/jobs/analysis/ontology_dnsrecords_linking.json`; #2922 replaced that JSON
job with typed `AnalysisJob`s and dropped the guard on one of the two statements.

`source_label` only feeds cleanup-query construction, never the `MERGE`, so no edge creation
changes. The GCP statement now declares the same guarded effect as the generic one,
`cleanup_where` included, so the two dedupe into a single delete: `GCPRecordSet` carries the
`DNSRecord` label and is never an `AWSDNSRecord`, so that one delete already covers those
edges and a separate `GCPRecordSet` cleanup would be redundant.

**2. Route53 alias targets never matched their load balancer.**

Route53 stores alias targets pointing at an ELB with a `dualstack.` prefix, the mechanism
that lets one name resolve to both A and AAAA. The ELB APIs return the same load balancer's
`DNSName` without it, and the matchers compare for equality, so an A or AAAA alias to an ALB
never linked. The prefix was handled nowhere in the codebase and the fixtures used
un-prefixed values, so no test could surface it.

The prefix is removed only for alias targets that actually point at a load balancer, keyed on
`.elb.` in the name. Route53 also allows an alias to another record in the same hosted zone,
and an ordinary CNAME's value is zone data written by the operator; in both of those a leading
`dualstack.` belongs to a genuinely different hostname and must survive.

Separately, the ELB and ELBv2 APIs preserve the load balancer name's case in `DNSName` while
Route53 lowercases alias targets, so any load balancer with an uppercase character in its
name never linked either.

**3. `route53` had no `module_dependencies` entry.**

The route53 loader creates `DNS_POINTS_TO` by matching nodes that already exist, so selecting
`route53` without those modules produced zero such edges with no warning, and
`cleanup_route53` then deleted the ones a previous run had created.

### Related issues or links

Regression of #2987, introduced by #2922.

### Breaking changes

None, but two behavior changes are worth calling out for anyone querying these properties
directly:

- `AWSLoadBalancer.dnsname` and `AWSLoadBalancerV2.dnsname` are now lowercased. `id` is
  unchanged and still carries the raw AWS value, so for a mixed-case load balancer
  `id != dnsname` where they used to be equal. The schema docs claimed `id` was "set to the
  `dnsname`" and are corrected.
- `AWSDNSRecord.value` for CNAME and ALIAS records is now lowercased and has the trailing
  root dot removed. Alias targets pointing at a load balancer additionally lose Route53's
  `dualstack.` prefix. Non-alias A/AAAA values are untouched, since those feed `ip_addresses`
  and the `Ip` node id matcher.

No migration is needed: property values are overwritten by the next sync's `MERGE` and Neo4j
maintains the `extra_index` as they change.

### How was this tested?

`4311` unit tests and `427` integration tests across the AWS, Kubernetes and ontology
suites, all green. `make test_lint` passes repo-wide.

Every fix was confirmed to actually guard its bug by reverting it and watching the new test
fail, not just by watching the new test pass:

| Reverted | Failing test |
|---|---|
| `source_label="GCPRecordSet"` | the cleanup-guard invariant names the conflicting edge; the ontology integration test loses the edge |
| the `dualstack.` strip | `test_sync_route53_links_alias_to_mixed_case_load_balancer` |
| the lowercased `dnsname` | the mixed-case route53, Service and Ingress tests fail to link |
| the `match_filter` guard | the new match-guard unit test |

The `dualstack.` coverage lives in `GET_ZONES_MIXED_CASE_ALIAS_RESPONSE` rather than the
shared route53 fixture, whose DNS name is not a real ELB name (no `.elb.`) and so cannot
exercise an ELB-scoped strip.

#### General

#### Proof of functionality

### Notes for reviewers

**Why ingestion-time normalization rather than `ignore_case=True`.** The framework supports
`ignore_case` on target node matchers, but it renders `toLower(n.dnsname) = toLower($ref)`,
and `dnsname` carries `extra_index=True`, so the index would stop being used and the match
would become a scan of every load balancer. Normalizing both sides at ingestion keeps exact
equality and keeps the index.

**Why ELBv2 needed a new `DNSNameLower` key.** `LoadBalancerV2NodeProperties` pointed both
`id` and `dnsname` at `DNSName`. That raw value is also the `LoadBalancerId` that listeners,
target groups and every `{"id": PropertyRef("LoadBalancerId")}` matcher join against, so
lowercasing it in place would have silently orphaned them for any mixed-case load balancer.
The transform now emits a separate lowercased copy and `id` keeps the raw one; a unit test
asserts exactly that invariant.

**Why the Kubernetes side had to move too.** `AWSLoadBalancerV2.dnsname` is a join key for
two Kubernetes MatchLinks as well as route53, and `load_balancer_dns_names` comes verbatim
from `status.loadBalancer.ingress[].hostname`, which the AWS Load Balancer Controller fills
with the raw ELB `DNSName`. Normalizing only the AWS side would have fixed route53 and broken
Kubernetes, so all four producers of a `dnsname` join key are normalized together.

**`elasticsearch` runs after `route53`, and that is deliberate.** While adding the
`module_dependencies` entry I found that `AWSDNSRecord -> AWSESDomain` can never be created
on a first run, because `elasticsearch` sits after `route53` in `RESOURCE_FUNCTIONS`. I did
not reorder it and deliberately left `AWSESDomain` out of the dependency list:
`cartography/intel/aws/elasticsearch.py` has a raw-Cypher cleanup that `DETACH DELETE`s any
stale `DNSRecord` pointing at an `AWSESDomain`. If route53 ran first and attached its record,
that cleanup would delete the whole route53-owned node on every subsequent run, before
route53 re-stamps it. Fixing the ES edge properly means replacing that cleanup first, which
is separate work. Two tests in `test_init.py` now pin both bounds, mirroring
`test_kms_syncs_before_kms_dependent_resources`: `route53` after its four target modules, and
before `elasticsearch`. `module_dependencies` is only a `logger.warning`, so it could not
carry that on its own.

**On whether the guard really preserves route53's ownership.** `cleanup_where` only stops
deletion, so it is fair to ask whether a separate ontology run refreshes a route53-owned
edge's `lastupdated` with its own tag. It does not: `AWSDNSRecord` is excluded from the
`MERGE`'s match too, via the `match_filter` column of `DNS_RECORD_TARGETS`. Nothing asserted
that, though, and the two guards live in different columns of the same table, so the last
commit adds a `lastupdated` assertion plus two unit tests that derive the route53-owned
`DNS_POINTS_TO` target labels from `AWSDNSRecordSchema` and require both guards for each.
Deriving them from the schema means adding a relationship to the route53 loader extends the
requirement automatically instead of leaving an unguarded target behind.

The four ontology DNS targets without the guard (`AWSCloudFrontDistribution`, `GCPInstance`,
`AzureAppService`, `AzureFunctionApp`) and the `KubernetesIngress` job do match
`AWSDNSRecord`, which is correct: route53 owns none of those edges, so the ontology job is
their sole owner and should stamp them.


cartography/intel/aws/route53.py
```diff
@@ -42,6 +42,33 @@ def _normalize_dns_address(address: str) -> str:
...
+    """
+    Normalize an AliasTarget's DNSName, which additionally needs the `dualstack.` prefix that
+    Route53 puts on alias targets pointing at an ELB removed: the ELB APIs return the same
+    load balancer's DNSName without it, so leaving it in place means the alias never matches.
+
+    Only alias targets get this treatment. An ordinary CNAME's value is zone data written by
+    the operator, where a leading `dualstack.` belongs to a genuinely different hostname, so
+    stripping it there would rewrite the record's target.
+    """
+    return _normalize_dns_target(value).removeprefix("dualstack.")
```
kunaals: could we only strip `dualstack.` when this alias actually targets an elb? route53 also allows an alias to another record in the same hosted zone, so a valid target like `dualstack.internal.example.com.` currently becomes `internal.example.com`. that changes the stored value and can link to the wrong dns record.

tests/integration/cartography/intel/ontology/test_dnsrecords.py
```diff
@@ -0,0 +1,69 @@
...
+    neo4j_session.run(
+        """
+        MERGE (lb:AWSLoadBalancerV2 {id: $lb_dns_name})
+        SET lb.dnsname = $lb_dns_name,
+            lb.lastupdated = $aws_tag
+
+        MERGE (owned:AWSDNSRecord:DNSRecord {id: 'HOSTED_ZONE/elbv2.example.com/ALIAS'})
+        SET owned.name = 'elbv2.example.com',
+            owned.value = $lb_dns_name,
+            owned.lastupdated = $aws_tag
```
kunaals, line 31: could we set `owned._ont_value = $lb_dns_name` in this fixture? without it, the ontology match excludes this node before reaching the `AWSDNSRecord` guard, so this assertion would still pass if that guard regressed.

cartography/intel/aws/resources.py
```diff
@@ -127,6 +127,12 @@
         "lambda_function": lambda_function.sync,
         "rds": rds.sync,
         "redshift": redshift.sync,
+        # `route53` matches already-existing nodes to create DNS_POINTS_TO edges, so it must
```
kunaals, line 130: could we add an ordering test for this, similar to the existing kms dependency test? correctness depends on route53 staying after these four resources and before elasticsearch, while the runtime check is only a warning.

docs/root/modules/aws/schema.md
```diff
@@ -1800,7 +1800,7 @@ Representation of an AWS DNS [ResourceRecordSet](https://docs.aws.amazon.com/Rou
 |lastupdated| Timestamp of the last time the node was updated|
 |**id**| The zoneid for the record, the value of the record, and the type concatenated together|
 |type| The record type of the DNS record (A, AAAA, ALIAS, CNAME, NS, etc.)|
-|value| If it is an A or AAAA record, this is the IP address the DNSRecord resolves to. For CNAME or ALIAS records, this is the target hostname or AWS resource name. If it is an NS record, the `name` is used here.|
+|value| If it is an A or AAAA record, this is the IP address the DNSRecord resolves to. For CNAME or ALIAS records, this is the target hostname or AWS resource name, lowercased and with the trailing root dot removed. Alias targets additionally have Route53's `dualstack.` prefix removed, since the ELB APIs report the same load balancer without it; an ordinary CNAME's value keeps any leading `dualstack.` because there it is part of a genuinely different hostname. If it is an NS record, the `name` is used here.|
```
kunaals: small docs nit: could we update the `id` row above while we are here? the code builds the id from the zone id, record name, and type, not the record value. the current wording makes it sound like this normalization changes node ids.

cartography/analysis/ontology/analysis.py
```diff
@@ -241,7 +241,13 @@
...
                             "DNS_POINTS_TO",
                             "target",
-                            source_label="DNSRecord",
+                            # Must stay GCPRecordSet, not the broader DNSRecord: cleanup
+                            # effects dedupe by value, so a DNSRecord source label here
+                            # emits a second cleanup that lacks the cleanup_where guard
+                            # of the statement above and deletes provider-owned edges.
+                            # GCPRecordSet also carries the DNSRecord label, so the
+                            # generic cleanup still reaches GCP-sourced stale edges.
+                            source_label="GCPRecordSet",
```
kunaals: could we keep the cleanup source as `DNSRecord` and pass the same `cleanup_where` here? `GCPRecordSet` already carries the `DNSRecord` label and cannot be an `AWSDNSRecord`, so this separate effect adds another cleanup that appears to be a subset of the guarded generic one.

kunaals (changes requested):

cartography/intel/aws/route53.py
```diff
@@ -42,6 +42,37 @@ def _normalize_dns_address(address: str) -> str:
...
+    Route53 puts on alias targets pointing at an ELB: the ELB APIs return the same load
+    balancer's DNSName without it, so leaving it in place means the alias never matches.
+
+    The prefix is only an ELB artifact, hence the `.elb.` check. An alias may also point at
+    another record in the same hosted zone, and an ordinary CNAME's value is zone data written
+    by the operator; in both of those a leading `dualstack.` belongs to a genuinely different
+    hostname and stripping it would rewrite the record's target.
+    """
+    normalized = _normalize_dns_target(value)
+    if ".elb." not in normalized:
```
kunaals: one more edge case: `.elb.` can also be part of a valid same-zone hostname, so this still rewrites `dualstack.app.elb.internal.example.com`. can we match the aws-owned suffixes instead? checking `without_dualstack.endswith((".elb.amazonaws.com", ".elb.amazonaws.com.cn"))` keeps arbitrary zone names intact while covering aws, govcloud, and china.

kunaals (approved):
one more nit, otherwise looks good

cartography/intel/aws/route53.py
kunaals: codex is recommending something like
```python
_ELB_DNS_SUFFIXES = (
    ".elb.amazonaws.com",
    ".elb.amazonaws.com.cn",
)


def _normalize_alias_target(value: str) -> str:
    normalized = _normalize_dns_target(value)
    without_dualstack = normalized.removeprefix("dualstack.")

    if (
        normalized.startswith("dualstack.")
        and without_dualstack.endswith(_ELB_DNS_SUFFIXES)
    ):
        return without_dualstack

    return normalized
```

PR #3091, opened 2026-07-30
feat(netlify): add Netlify intel module

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds a Netlify intel module. Netlify is a full hosting platform, not just a static CDN: alongside sites and deploys it runs serverless functions, ephemeral cloud dev servers with public URLs, AI agent runners that push branches and open pull requests, and a managed Postgres (Netlify DB, on Neon). It also holds team membership, environment variables flagged as secrets, deploy keys, DNS zones, TLS certificates and injected JS snippets. None of that was reachable from the graph.

19 node types, scoped to one team per run (`--netlify-account-slug`), mapped into the ontology as `Tenant`, `UserAccount`, `ComputeService`, `ComputeInstance`, `Function`, `Database`, `Snapshot`, `Secret`, `ThirdPartyApp`, `DNSZone`, `DNSRecord` and `Certificate`, plus a best-effort `DEPLOYED_FROM` edge to `GitHubRepository`.

**Conventions.** Follows the AWS/Azure/GCP shape rather than the Vercel/Railway one: explicit positional tenant id and update tag on every sync, resource-qualified `get_`/`transform_`/`load_`/`cleanup_` names, a minimal `common_job_parameters` carrying only `UPDATE_TAG` plus a `NETLIFY_ACCOUNT_ID` set and popped around the fan-out, a `_sync_one_account` helper, and `merge_module_sync_metadata` at the end. Azure was the closest precedent: single-level tenant, no regions, REST, flat one-file-per-service layout.

**Cleanup shape.** Each per-site domain loops the whole site list inside its own sync and then cleans up once. Cleaning inside the loop would delete the not-yet-visited sites' nodes from the previous run and recreate them later in the same pass: a hole for concurrent readers, and permanent loss if a later request fails. It is also the only shape that correctly removes the children of a site that disappeared between runs.

**Secrets.** Several endpoints return live credentials. None are ingested; presence booleans and metadata are kept instead:

| Field | Handling |
|---|---|
| `envVar.values[].value` | dropped (Netlify masks a secret to its last 4 chars, which still leaks) |
| `databaseBranch.connection_string` | dropped. `GET /sites/{id}/database`, whose whole body is one, is never called |
| `buildHook.url` | dropped, plus the `msg` field that repeats it |
| `hook.data` | dropped (Slack webhook URL / target URL / git provider token) |
| `site.jwt_secret` | dropped, replaced by `has_jwt_secret` |
| `deploy.skew_protection_token` | dropped |
| `serviceInstance.config`, `.env`, `.auth_url` | dropped |

**Corrections to the published OpenAPI spec**, all found against the live API:

- `GET /sites/{id}/functions` returns single-letter keys (`n`, `r`, `rg`, `m`, `s`, ...), not `{id, name, sha, region}`, and answers with a bare object rather than an array when the site has one bundle.
- `GET /api/v1/sites/{site_id}/env` is double-prefixed in the spec and 404s; site-scoped variables come from the account endpoint with `?site_id=`.
- `GET /sites/{id}/ssl` answers with a JSON `null`, not a 404, when the site has no custom domain.
- `/database/branches` and `/database/snapshots` are enveloped, not bare arrays.

**Synthesised ids**, where Netlify gives none or gives an unstable one: certificates have no identifier at all (`<site_id>_ssl`, deliberately excluding `expires_at` so a renewal does not churn the node identity), function ids are per-build content hashes (`<site_id>|<branch>|<name>`), snippet ids are positional array indices (`<site_id>|<index>`), and `branch_id` is only unique within a site since the primary branch is called `production` on every Netlify DB.

**Shared identities.** `NetlifyUser` and `NetlifyDeployKey` carry no `sub_resource_relationship`; their team edges are MatchLinks. A Netlify user can belong to several teams with a different role in each, and `GET /deploy_keys` takes no team parameter so every team's sync sees every key the token can see. Anchoring either to the team being synced makes the generated node cleanup `DETACH DELETE` rows another team's sync just refreshed, and makes the `MEMBER_OF` cleanup delete memberships of every *other* team, because the account on its second `MATCH` is unconstrained. Both were reproduced against a real Neo4j before being fixed. Same shape as `RailwayUser` and `GitHubUser`; both labels join `GLOBAL_NODE_LABELS`. Consequence for callers: a person removed from their last team keeps a bare node with no membership, so ask "who is on this team" through `MEMBER_OF` rather than by node existence. Documented in the schema.

### Related issues or links

- Netlify API reference: https://open-api.netlify.com/
- Netlify API guide (auth, 500 req/min rate limit, `Link`-header pagination): https://docs.netlify.com/api-and-cli-guides/api-guides/get-started-with-api/

### How was this tested?

**Unit and integration.** 4313 unit tests and 35 Netlify integration tests pass; `make test_lint` is clean. Integration tests patch only the `get_netlify_*` functions and call the real `sync()`, asserting with `check_nodes` / `check_rels`. `tests/unit/.../netlify/test_util.py` covers pagination: the `Link`-header walk, params sent on the first request only, the empty-page break, `404` to `[]`, plan-gated `403` to partial, and the raise-on-unusable-next case.

Dedicated cleanup coverage in `test_cleanup.py` (a stale site removed with its children, a stale DNS record removed without its zone, one domain's cleanup not touching another) and in `test_users_multi_team.py` (a second team's sync preserving the first team's edges, removal from one team leaving the other intact, removal from the last team leaving the identity but no membership, per-team roles on the edge, and the deploy key case). All five multi-team tests fail on a model that anchors the identity to a team.

**Against a real Netlify team.** Fixtures under `tests/data/netlify/` are captured live and scrubbed, which is how the spec corrections above were found. A site with two functions, two env vars (one secret), a build hook, a snippet, a deploy key and a Netlify DB was provisioned with the `netlify` CLI, synced, and then torn down.

```
INFO:cartography.sync:Starting sync stage 'netlify'
INFO:cartography.client.core.tx:Loaded 1 NetlifyAccount nodes
INFO:cartography.client.core.tx:Loaded 1 NetlifyUser nodes
INFO:cartography.client.core.tx:Loaded 1 NetlifySite nodes
INFO:cartography.client.core.tx:Loaded 1 NetlifyDeploy nodes
INFO:cartography.client.core.tx:Loaded 2 NetlifyFunction nodes
INFO:cartography.client.core.tx:Loaded 1 NetlifyDatabaseBranch nodes
INFO:cartography.client.core.tx:Loaded 2 NetlifyEnvVar nodes
INFO:cartography.client.core.tx:Loaded 1 NetlifyBuildHook nodes
INFO:cartography.client.core.tx:Loaded 1 NetlifySnippet nodes
INFO:cartography.client.core.tx:Loaded 1 NetlifyDeployKey nodes
INFO:cartography.sync:Finishing sync stage 'netlify'
INFO:cartography.sync:Finishing sync with update tag '1785434415'
```

Resulting graph: 6 ontology labels applied (`Tenant`, `UserAccount`, `ComputeService`, `Function`, `Database`, `Secret`), 11 `RESOURCE` edges plus `HAS_DEPLOY`, `HAS_FUNCTION`, `HAS_ENV_VAR`, `HAS_BUILD_HOOK`, `HAS_SNIPPET`, `HAS_DATABASE_BRANCH`, `DEPLOYED_BY` and `UPDATED_BY`.

Three things checked directly against that graph:

- **No credential on any property.** A scan of every property of every node against patterns for Postgres connection strings, build hook URLs, Netlify PATs and masked secret values returns nothing.
- **The conditional `Secret` label is correct.** `CARTO_SECRET` (`is_secret=True`) carries `:Secret`; `CARTO_PUBLIC` does not.
- **Cleanup behaves.** Deleting the build hook on Netlify and re-syncing removed exactly that node. Deleting the site and re-syncing removed its deploy, functions, dev server, agent runner, database branches, env vars and snippets, leaving the account, the member and the account-level deploy key alone. Two consecutive syncs with different update tags leave the user's `RESOURCE` and `MEMBER_OF` edges intact.

#### General

#### Proof of functionality

- [ ] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md): `docs/schema/README.md` does not exist on `master`; this checklist item looks stale. `docs/root/usage/schema.md`, `docs/root/index.md`, `docs/root/module-list.md` and the `README.md` platform list are all updated.

### Notes for reviewers

Worth a closer look:

- **`models/netlify/user.py` and `models/netlify/deploykey.py`.** Each opens with a docstring quoting the two cleanup statements that make a node-schema relationship unsafe for a shared identity. This is the least obvious part of the module.
- **Deploy history is out of scope.** Only the deploy currently published on each site is ingested, read from the `published_deploy` already embedded in the site payload, so it costs no extra request and gives a bounded set cleanup can treat as exhaustive. The full history is paginated and can hold thousands of entries per site.
- **Agent runner sessions are out of scope** (prompt, step list, result diff: live execution records, not inventory). The runner itself is ingested, with its PR state and the branch it pushed.
- **No selective sync.** With 17 domain modules a `--netlify-requested-syncs` option (AWS/GCP style, via a `RESOURCE_FUNCTIONS` registry) is the natural follow-up. Azure ships without one, so it did not seem like a blocker for a first cut.
- **Not validated end to end**, and covered by hand-written fixtures only: DNS zones and records (`createDnsZone` returns HTTP 500 on a Free-plan team), notification hooks (`createHookBySiteId` returns 422 through the CLI), add-on service instances, forms (detection only runs during a Netlify build, and the test site was deployed prebuilt), TLS certificates (provisioning needs a custom domain with delegated DNS) and database snapshots. Each is called out in the docstring of `tests/data/netlify/misc.py` and `tests/data/netlify/dns.py`.
- Netlify's request budget is roughly `10S + Z + 5` for `S` sites and `Z` DNS zones, against a 500 req/min limit, so about 45 sites per minute. Documented in `docs/root/modules/netlify/config.md`.


cartography/intel/netlify/functions.py
```diff
@@ -0,0 +1,124 @@
...
+    neo4j_session: neo4j.Session,
+    data: list[dict[str, Any]],
+    account_id: str,
+    update_tag: int,
+) -> None:
+    load(
+        neo4j_session,
+        NetlifyFunctionSchema(),
+        data,
+        lastupdated=update_tag,
```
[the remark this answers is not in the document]
jychp: Not applying this one, and I want to be explicit about why rather than just resolve it.

`lastupdated` here is not a data timestamp. It is cartography's sync generation marker: `config.update_tag` is set to `int(time.time())` in [`cartography/sync.py`](https://github.com/cartography-cncf/cartography/blob/master/cartography/sync.py) and the value is then used as the cleanup predicate, `WHERE n.lastupdated <> $UPDATE_TAG` in `cartography/graph/cleanupbuilder.py`. Every node loaded in a run gets the same integer, and cleanup deletes whatever does not carry it. All 752 `lastupdated=update_tag` call sites across the intel modules do exactly this.

Passing a `datetime` or an ISO string here would not be a cosmetic change: the comparison is an equality test against the run-scoped tag, so a per-node timestamp would make cleanup delete everything on every run.

The native-temporal-types convention does apply to timestamps that are actual data, and this module follows it. `created_at`, `updated_at`, `published_at`, `expires_at`, `last_activity_date` and the rest are stored as the ISO 8601 strings the Netlify API returns, with no epoch conversion.

The other nine comments are addressed in f237640a6.

cartography/intel/netlify/util.py
```diff
@@ -0,0 +1,273 @@
...
+    return all_results
+
+
+@timeit
+def get_list(
+    api_session: requests.Session,
+    url: str,
+    params: dict[str, Any] | None = None,
+    result_key: str | None = None,
+    allow_missing: bool = False,
```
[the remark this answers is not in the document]
jychp: The premise does not hold: that endpoint does not answer 404. I checked against the live API rather than reasoning about it, on a site created with no functions and then again after deploying it without any:

```
$ netlify api searchSiteFunctions --data '{"site_id":"<site with no functions>"}'
{}
```

`200 {}` in both cases. For contrast, a real 404 on this API surfaces as an error through the same client, which is how I found that the spec's `/api/v1/sites/{site_id}/env` path is broken.

So no sync failure: `get_list` takes the `isinstance(data, dict)` branch, returns `[{}]`, and `transform_netlify_functions` reads `bundle.get("functions") or []` and emits nothing. I verified end to end against a team whose only site has zero functions: the sync completes and creates no `NetlifyFunction` nodes.

Not adding `allow_missing=True`, since that would add tolerance for a response the API does not send, and the previous review round asked for exactly this default to be tightened.

The comment did catch a real error though, just not that one. My own `get_netlify_functions` docstring claimed "A site with no functions answers 404" — an assumption I never verified, because the site I built the fixtures from had functions. Fixed in 8e12b8645 to state what was actually observed. That commit also handles `{}` explicitly instead of letting it become a phantom `[{}]` bundle, which was only harmless by accident, and adds a unit test for it.

cartography/intel/netlify/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+    account_slug: str,
+    update_tag: int,
+    common_job_parameters: dict[str, Any],
+) -> None:
+    """
+    Sync every resource of one Netlify team.
+
+    Ordering matters in two ways. Users come first because deploys, env vars and agent runners
+    all point at them. Sites come next because every remaining domain is scoped to a site, and
+    because the site list is what those domains iterate over. Deploy keys come last: the edge
```
kunaals: i think the deploy keys need to load before the sites. on a clean graph the site loader has no key node to match, so `USES_DEPLOY_KEY` doesn’t appear until the second sync. the current site test pre-creates the key and hides this case.

cartography/intel/netlify/deploykeys.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    return get_list(api_session, f"{base_url}/deploy_keys")
+
+
+def transform_netlify_deploy_keys(
+    deploy_keys: list[dict[str, Any]],
+    account_id: str,
+) -> list[dict[str, Any]]:
+    """Carry the team id for the MatchLink, which has no kwargs to read it from."""
+    return [{**key, "account_id": account_id} for key in deploy_keys]
```
kunaals: this seems to attach every token-visible key to the current team. since `/deploy_keys` has no team filter, a multi-team token will give every team a `RESOURCE` edge to every key. can we derive ownership from the sites that actually reference each key?

cartography/models/netlify/user.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Netlify reports everything on a per-team membership payload, so anything team-scoped that
+    landed here would be overwritten by whichever team's sync ran last. `pending`,
+    `managed_by_directory_sync`, `role`, `site_access` and the membership timestamps are therefore
+    all on the MEMBER_OF edge instead.
+    """
+
+    # Netlify returns two ids on a membership: `id` identifies the membership row and `user_id`
+    # identifies the person. We key on `user_id` so one human is one node across teams.
+    id: PropertyRef = PropertyRef("user_id")
```
kunaals: pending invitations may not have a `user_id` yet. using it as the node key makes the whole load fail on a null-id merge, so this case needs to be skipped or represented separately.

kunaals (changes requested):

cartography/intel/netlify/users.py
```diff
@@ -0,0 +1,188 @@
...
+def cleanup_netlify_users(
+    neo4j_session: neo4j.Session,
+    account_id: str,
+    update_tag: int,
+) -> None:
+    # Only the edges are cleaned up, and only this team's. The NetlifyUser node is never deleted
+    # here: other teams and other modules may still reference the identity. A person removed from
+    # their last team keeps a bare node with no membership, which is the same outcome Railway and
+    # GitHub produce for a shared principal.
+    for matchlink in (
```
kunaals, line 180: one invitation cleanup case is still missing here. when an invite is accepted or revoked, these jobs remove its edges but leave the NetlifyInvite node behind forever. please delete invite nodes once they have no remaining INVITED_TO edge. that still preserves an address invited to another team.

kunaals (changes requested):

kunaals (approved):

PR #3092, opened 2026-07-30
perf(container-images): reuse cached layer metadata

jychp:
### Type of change
- [ ] Documentation update
- [x] Other: Performance improvement

### Summary
- Add a shared graph shape and refresh helper for digest-addressed container image layers.
- Skip immutable manifest/config layer pulls for already-enriched digests across ECR, GHCR, GitLab, GCP Artifact Registry, and Scaleway.
- Refresh image, layer, and relationship timestamps before cleanup, and move ECR enrichment cleanup after layer processing to preserve live closures.

### Related issues or links
None.

### How was this tested?
- `.venv/bin/python -m pytest tests/unit/cartography/intel/test_container_image_layers.py tests/unit/cartography/intel/gitlab/test_container_images.py tests/unit/cartography/intel/aws/test_ecr_image_layers.py tests/unit/cartography/intel/gcp/test_artifact_registry_supply_chain.py tests/unit/cartography/intel/scaleway/test_container_registry_supply_chain.py` (103 passed)
- `.venv/bin/python -m pytest tests/integration/cartography/intel/github/test_packages.py tests/integration/cartography/intel/aws/test_ecr_layers.py tests/integration/cartography/intel/gitlab/test_container_registry.py tests/integration/cartography/intel/gcp/test_artifact_registry.py tests/integration/cartography/intel/scaleway/test_container_registry_supply_chain.py` (37 passed)
- Pre-commit hooks: isort, Black, flake8, pyupgrade, and mypy passed.

#### General
- [ ] I have read the [contributing guidelines](https://docs.cartography.dev/dev/developer-guide.html).

#### Proof of functionality

### Notes for reviewers
The common helper treats `layer_diff_ids IS NOT NULL` as a completed enrichment, including valid images with an empty layer list. Provider-specific registry clients, transforms, and provenance handling remain separate.


cartography/intel/aws/ecr_image_layers.py
```diff
@@ -1279,6 +1304,9 @@ async def _fetch_single_image_layers_with_uri(
 
 def cleanup(neo4j_session: neo4j.Session, common_job_parameters: dict) -> None:
     logger.debug("Running image layer cleanup job.")
+    GraphJob.from_node_schema(ECRImageSchema(), common_job_parameters).run(
```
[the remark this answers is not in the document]
jychp: Fixed in d3ab27ef3. ECR layer cleanup now deletes only stale HAS_LAYER and BUILT_FROM relationships for the account; AWSECRImage inventory nodes remain owned by the base ECR sync. Added an integration regression test proving stale inventory images survive layer cleanup.

cartography/intel/gitlab/container_images.py
```diff
@@ -170,6 +188,9 @@ def get_container_images(
 
             media_type = manifest.get("mediaType")
             is_manifest_list = media_type in MANIFEST_LIST_MEDIA_TYPES
+            if not is_manifest_list and digest in skip_digests:
+                observed_and_skipped.add(digest)
```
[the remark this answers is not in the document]
jychp: Fixed in d3ab27ef3. Cached regular and child manifests now remain in a separate attestation-discovery collection while image/layer transforms still skip them, so cosign .sig/.att probing continues without refetching config blobs.

cartography/intel/container_image_layers.py
```diff
@@ -0,0 +1,266 @@
...
+
+    digest_predicate = (
+        f"img.{shape.image_digest_property} IN $digests"
+        if unique_digests is not None
+        else "true"
+    )
+    query = f"""
+    MATCH (img:{shape.image_label})
+    WHERE {digest_predicate}
+      AND img.{shape.image_layer_ids_property} IS NOT NULL
```
kunaals, line 182: could we include the expected layer relationships in this completeness check too? ecr writes the nodes and each relationship set separately, so a partial run can leave all the layer nodes here while missing `HAS_LAYER`/`HEAD`/`TAIL`/`NEXT`. we would then skip the fetch, and the refresh below can only update relationships that still exist.

cartography/intel/aws/ecr_image_layers.py
```diff
@@ -1279,6 +1304,17 @@ async def _fetch_single_image_layers_with_uri(
 
 def cleanup(neo4j_session: neo4j.Session, common_job_parameters: dict) -> None:
     logger.debug("Running image layer cleanup job.")
+    run_write_query(
+        neo4j_session,
+        """
+        MATCH (:AWSAccount {id: $AWS_ID})-[:RESOURCE]->
+              (:AWSECRImage)-[relationship:HAS_LAYER|BUILT_FROM]->()
+        WHERE relationship.lastupdated <> $UPDATE_TAG
```
kunaals, line 1342: could we skip this cleanup when any ecr layer fetch was incomplete? exhausted transient failures are caught per image above, so this can delete the existing layer and `BUILT_FROM` relationships for an image that is still live.

cartography/intel/aws/ecr_image_layers.py
```diff
@@ -1327,7 +1363,8 @@ def sync(
             img.attests_digest AS attests_digest,
             img.media_type AS media_type,
             img.artifact_media_type AS artifact_media_type,
-            img.child_image_digests AS child_image_digests
+            img.child_image_digests AS child_image_digests,
+            img.layer_diff_ids AS layer_diff_ids
```
kunaals: could we return the existing source, invocation, and parent provenance fields here too? the late-attestation path copies this map into a full enrichment load, so when a new attestation only supplies some fields, the omitted existing values get written back as null.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -48,6 +51,17 @@
...
+GCP_ARTIFACT_REGISTRY_LAYER_GRAPH = ContainerImageLayerGraphShape(
+    image_label="GCPArtifactRegistryImage",
+    layer_label="GCPArtifactRegistryImageLayer",
+    scope_label="GCPProject",
+    scope_properties=("id",),
+    image_scoped=False,
+    image_layer_rel_types=(),
+    stable_image_rel_types=("BUILT_FROM",),
+    stable_image_rels_are_matchlinks=True,
+)
```
kunaals, line 64: could we set `layer_id_property="id"` here? `id` and `diff_id` contain the same value for these layers, but only `id` is indexed. the generic completeness and refresh queries currently match on `diff_id`, which causes repeated layer-label scans.

cartography/intel/gcp/artifact_registry/supply_chain.py
```diff
@@ -909,10 +924,12 @@ async def _fetch_all_image_provenance(
     token_manager = _TokenManager(resolved)
     sbom_artifacts_by_digest = _sbom_artifacts_by_subject_digest(docker_artifacts_raw)
 
+    skip_digests = skip_digests or set()
     single_images = [
         a
         for a in docker_artifacts_raw
         if a.get("mediaType", "") in SINGLE_IMAGE_MEDIA_TYPES
+        and _artifact_digest(a) not in skip_digests
```
kunaals: could we keep doing the provenance/referrer check for cached digests? a failed referrer or sbom request can still load `layer_diff_ids`, and then the next run filters the digest out here and never retries. this also prevents discovering attestations published after the image.

cartography/intel/github/container_images.py
```diff
@@ -551,10 +481,11 @@ def sync_container_images(
     # Refresh lastupdated for digests we observed in versions but whose
     # manifest/config fetch we skipped via cross-run dedup. Must happen
     # before cleanup, otherwise the still-live images get reaped.
-    _refresh_skipped_image_lastupdated(
+    refresh_layer_closures(
```
kunaals, line 526: could we also keep the package `HAS_IMAGE` edge current for these cached digests? they skip `load_container_images()`, and this helper only refreshes outbound image relationships, so cleanup deletes the existing incoming `GitHubPackage` relationship.

cartography/intel/scaleway/container_registry/supply_chain.py
```diff
@@ -67,9 +80,43 @@ def sync(
...
+        project_images = [
+            image for image in images_to_enrich if image.get("project_id") == project_id
+        ]
+        complete_digests = get_complete_layer_digests(
+            neo4j_session,
+            SCALEWAY_LAYER_GRAPH,
+            (str(image["digest"]) for image in project_images if image.get("digest")),
+            {"id": project_id},
+        )
+        project_fetches, project_skips = partition_layer_fetches(
```
kunaals: could we avoid caching the digest when its attestation fetch failed? that error is caught while the config and layers still load, so the digest becomes complete here and the attestation is never retried.

cartography/intel/container_image_layers.py
```diff
@@ -0,0 +1,266 @@
...
+def refresh_layer_closures(
+    neo4j_session: neo4j.Session,
+    shape: ContainerImageLayerGraphShape,
+    digests: Iterable[str],
+    scope_values: Mapping[str, Any],
+    update_tag: int,
+) -> None:
+    """Refresh an unchanged image and its existing layer closure before cleanup."""
+
+    unique_digests = sorted(set(digests))
```
kunaals, line 229: could we chunk these digests into bounded writes? in steady state this can pass every cached image through several relationship updates in one transaction, which could become millions of writes for a large account.

kunaals (changes requested):

kunaals (approved):
approving to unblock though i think that remaining cubic comment has merit

jychp:
> approving to unblock though i think that remaining cubic comment has merit

was fixed locally and not pushed 🤦

PR #3095, opened 2026-07-31
fix(aws): handle null groups in aws-auth mappings

AlexanderDeBattista:
### Type of change

### Summary

A `mapRoles`/`mapUsers` entry in the `aws-auth` ConfigMap can set `groups: null`. A principal mapped to a username with no additional RBAC groups is valid. `parse_aws_auth_map` preserves that as `None`, and `transform_aws_auth_mappings` read it with `mapping.get("groups", [])`, whose default only applies when the key is missing. A present-but-null value returns `None`, so `for group_name in group_names:` raised `TypeError: 'NoneType' object is not iterable`.

Because the exception is unhandled, it aborts the entire `kubernetes` sync stage when `--managed-kubernetes eks` is set, so no Kubernetes resources are ingested.

Coercing with `mapping.get("groups") or []` treats null the same as an empty list. The same guard is applied to `kubernetesGroups` on EKS access entries, which has the identical pattern.

### How was this tested?

Added a unit test (`test_transform_aws_auth_mappings_handles_null_groups`) covering `groups: null` on both a role and a user mapping; it reproduces the `TypeError` on `master` and passes with the fix. `make test_unit` and `make test_lint` pass locally.

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/kubernetes/eks.py
+++ b/cartography/intel/kubernetes/eks.py
@@ -271,7 +271,7 @@ def transform_aws_auth_mappings(
         for mapping in auth_mappings.get("roles", []):
             role_arn = mapping.get("rolearn")
             username = mapping.get("username")
-            group_names = mapping.get("groups", [])
+            group_names = mapping.get("groups") or []
 
             if not role_arn:
                 continue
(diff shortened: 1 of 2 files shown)


jychp (approved):
Good catch! Thx for the contribution.

PR #3107, opened 2026-08-03
feat(cloudflare): add R2, Workers, and WAF ruleset ingestion

jychp:
### Type of change

- [ ] Documentation update

### Summary

The Cloudflare module covered accounts, roles, members, zones and DNS records. Three security-relevant surfaces were invisible:

- **R2**: buckets can be exposed to the internet through their managed `r2.dev` domain or a custom domain, with nothing in the graph to show it.
- **Workers**: code running at the edge, and the routes that decide which public hostnames invoke it.
- **WAF**: the Rulesets engine, where all request filtering, rate limiting and bot mitigation configuration lives.

This adds five node types. Each uses `CloudflareAccount` as its sub-resource, including the zone-scoped ones, following the DNS-record precedent: the account is the tenant, and per-zone cleanup would delete rows belonging to zones not yet synced in the run.

| Node | Ontology label | Notes |
| --- | --- | --- |
| `CloudflareR2Bucket` | `ObjectStorage` | The bucket listing does not report exposure, so the managed and custom domains are read per bucket to resolve `public`. An enabled custom domain links the bucket to the zone hosting it. |
| `CloudflareWorkerScript` | `Function` | Alongside `AWSLambda`, `GCPCloudFunction`, `AzureFunctionApp`. |
| `CloudflareWorkerRoute` | none | `(:CloudflareZone)-[:HAS_ROUTE]->(:CloudflareWorkerRoute)-[:ROUTES_TO]->(:CloudflareWorkerScript)`, which is what makes a Worker's internet exposure queryable. |
| `CloudflareRuleset` | `NetworkAccessControl` (conditional) | Rulesets deploy at the account level and the zone level; both are read. |
| `CloudflareRulesetRule` | none | Matches the GCP precedent where `GCPCloudArmorPolicy`, not its rules, carries the label. |

No CLI or config change: the existing `--cloudflare-token-env-var` token covers all three, so `sync.py`, `cli.py` and `config.py` are untouched.

#### Ruleset identity is per deployment, not per API ID

A Cloudflare-provided ruleset keeps the **same API ID in every scope and account that deploys it**. Keying nodes on that ID would merge the account-level and zone-level deployments of a managed ruleset (so `scope` would flip to whichever was written last) and merge the deployments of separate accounts read with the same token (so `RESOURCE` and `HAS_RULESET` edges would span tenants).

So `CloudflareRuleset` is keyed by deployment, `<scope_id>/<ruleset_id>`, and `CloudflareRulesetRule` by `<ruleset deployment id>/<rule_id>`. Zone IDs and account IDs are both globally unique, so the two scope forms cannot collide. The raw API IDs are kept as indexed `ruleset_id` / `rule_id` properties.

For the same reason, an `execute` rule cannot name its target with the shared API ID alone: it would join every deployment of that ruleset, across accounts. The target is resolved in the scope the rule was read from and exposed as `(:CloudflareRulesetRule)-[:EXECUTES]->(:CloudflareRuleset)`, pointing at exactly one deployment. `executed_ruleset_id` still records the raw API ID, since that is what identifies *which* Cloudflare managed ruleset is involved.

#### The `NetworkAccessControl` label is conditional

Rulesets also cover caching, transforms, redirects and custom errors. Labelling a cache ruleset as access control would pollute cross-provider firewall queries, so the label is gated on a precomputed `security_ruleset` field via `ExtraNodeLabel.when()`, which only supports exact string equality (hence the flag rather than a phase list). Non-security rulesets still receive `_ont_name` without the label.

#### Deliberate exclusions

- **Contents of `kind: managed` rulesets are not expanded.** They are vendor-owned rule collections, not customer configuration, and would add hundreds of nodes per account. The managed ruleset node itself is ingested, along with the customer's `execute` rule that turns it on, which is where the operator's actual choice lives.
- **R2 bucket versioning**: no `r2.buckets.*` endpoint in the SDK exposes it, so `_ont_versioning` stays unmapped.
- **Legacy `firewall.rules` / `filters`**: deprecated by Cloudflare in favour of Rulesets.

#### Failure behaviour

The R2 stage degrades instead of failing the run. If the bucket listing is refused, R2 is skipped with a warning and the Workers and ruleset stages still sync; cleanup is skipped too, so buckets from earlier runs are not deleted on a permission error. If only the per-bucket domain lookups are refused, buckets are ingested with exposure left unresolved.

The Workers and ruleset stages still let API errors propagate, on the basis that a missing scope is a configuration error that should be loud. Happy to make them degrade the same way if reviewers prefer consistency here.

### Related issues or links

- Cloudflare ruleset deployment scopes: https://developers.cloudflare.com/ruleset-engine/managed-rulesets/deploy-managed-ruleset/
- Ruleset phase list: https://developers.cloudflare.com/ruleset-engine/reference/phases-list/
- R2 bucket listing (cursor-based, see note below): https://developers.cloudflare.com/api/resources/r2/subresources/buckets/methods/list/

### How was this tested?

```
uv run --frozen pytest tests/unit tests/integration/cartography/intel/cloudflare tests/integration/cartography/intel/ontology
# 4622 passed

make test_lint
# all hooks pass

uv sync --group doc && uv run ./docs/build.sh
# build succeeded; all five nodes appear in the generated Cloudflare schema
```

Six new integration tests. Beyond the node and relationship round-trips, they cover the cases that motivated the design:

- **`test_shared_managed_ruleset_does_not_merge_scopes_or_accounts`**: the same managed ruleset in both scopes across two accounts. Asserts four distinct deployments each keeping its own `scope`, plus Cypher tenancy invariants (no deployment owned by more than one account, none attached to another account's zone, no `execute` rule with more than one target or targeting another account's deployment).
- Only the `http_request_firewall_custom` ruleset carries `:NetworkAccessControl`; the `http_request_cache_settings` one does not.
- The `ObjectStorage` and `Function` labels plus their `_ont_*` projections.
- Worker routes without a script attached create no `ROUTES_TO` edge.
- A refused R2 listing skips cleanup and leaves previously ingested buckets in place.

#### General

#### Proof of functionality

  **Not provided.** I do not have a Cloudflare account with R2, Workers and account-level WAF to sync against, so rather than post logs I cannot stand behind, the API surface was verified against the pinned `cloudflare` SDK (5.6.0) response models and the docs linked above, and the six integration tests exercise the full get/transform/load/cleanup path against a real Neo4j. Flagging this explicitly for reviewers who can run it against a live tenant.

- [x] Updated the [schema documentation](https://github.com/cartography-cncf/cartography/tree/master/docs/root/modules): the Cloudflare schema is generated from the data model since #3021, so the docstrings and `PropertyRef(description=...)` strings are the source. Verified by building the docs.
- [ ] Updated the schema README: `docs/schema/README.md` no longer exists after #3021; this checklist item looks stale.

### Notes for reviewers

Worth a close look:

1. **The deployment-keyed IDs.** This is the least conventional part of the change. The alternative was modelling deployments as a separate node type from rulesets; I took the ID-qualification route as the smaller change. If you would rather see `CloudflareRuleset` (shared identity) and a distinct deployment node, say so and I will restructure.

2. **`client.r2.buckets.list()` is not an SDK paginator** and its response model drops the pagination cursor, unlike every other Cloudflare list endpoint used here. Buckets are ordered lexicographically, so `get()` walks pages with `start_after`. Worth confirming against an account with more than 1000 buckets.

3. **Cost of R2 exposure resolution**: two extra API calls per bucket (managed domain, custom domains). That is what `_ont_public` requires, but it is a real per-bucket cost on accounts with many buckets.

4. **`runtime` is deliberately unmapped for Workers.** Every Worker runs on `workerd`, which is not comparable to the per-language runtimes the other providers report; `compatibility_date` carries the runtime version instead.


egallis31:
https://github.com/cartography-cncf/cartography/pull/3106 - Thanks for adding this feature set, closing this PR in favor of official maintainer

jychp:
> #3106 - Thanks for adding this feature set, closing this PR in favor of official maintainer

@egallis31 sorry for that, did not see your contribution ...

egallis31:
> > #3106 - Thanks for adding this feature set, closing this PR in favor of official maintainer
> 
> @egallis31 sorry for that, did not see your contribution ...

no worries at all! yours is much more fleshed out! Thanks for the great work :)

cartography/intel/cloudflare/rulesets.py
```diff
@@ -0,0 +1,247 @@
...
+    client: Cloudflare,
+    account_id: str,
+    zone_id: str | None,
+    update_tag: int,
+) -> None:
+    """
+    Load the rulesets of a single scope, and the rules of the customer-authored
+    ones. A `zone_id` of None reads the account-level rulesets.
+    """
+    rulesets = get_rulesets(client, account_id, zone_id)
```
kunaals: this list includes rulesets that are available to a zone, not only ones it has deployed. right now every managed or account-level ruleset returned here gets a zone-qualified node and `HAS_RULESET` edge, even when the zone never executes it. i think the enabled `execute` rule in the entry-point ruleset should be the deployment signal.

cartography/intel/cloudflare/r2buckets.py
```diff
@@ -0,0 +1,227 @@
...
+    account_id: str,
+    exposure: Dict[str, Dict[str, Any]],
+) -> List[Dict[str, Any]]:
+    """
+    Qualify the bucket IDs with the account and merge in the resolved exposure.
+    """
+    result: List[Dict[str, Any]] = []
+    for bucket in buckets:
+        transformed = bucket.copy()
+        transformed["id"] = f"{account_id}/{bucket['name']}"
```
kunaals: jurisdiction needs to be part of the bucket identity here. cloudflare allows the same account to have same-named buckets in different jurisdictions, so `account/name` can collide. exposure is also keyed only by name, and the domain lookups omit jurisdiction. i’d carry jurisdiction through the id, exposure key, and managed/custom domain requests, and make sure the listing covers each jurisdiction.

kunaals (approved):
seems quite close, just two small changes that i think would make it more robust and then lets merge

PR #3110, opened 2026-08-03
feat(snowflake): add Snowflake intel module

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds a Snowflake intel module.

Snowflake holds an account-scoped RBAC model (users, roles, role hierarchy, grants), its own network access controls, and pivots into S3 / GCS / Azure storage through stages, external volumes and integrations. None of that was previously visible in the graph.

The module covers **61 node types**: account and organization, users and service users, roles and database roles, role hierarchy, grants and ownership, programmatic access tokens and credential posture, warehouses / compute pools / resource monitors, the database → schema → table hierarchy, stages, external volumes, secrets, network policies and rules, the six integration kinds, tasks, pipes, functions, procedures, SPCS services, image repositories, shares, listings, replication groups, and the policy surfaces.

**Transport is hybrid.** REST API v2 covers most of the inventory, but several security-relevant surfaces exist only as SQL, so those go through the SQL API (`POST /api/v2/statements`) on the same credential: security and storage integrations, programmatic access tokens, MFA enrollment, `SHOW GRANTS ON ACCOUNT`, shares and listings, replication groups, resource monitors, and policy attachments. **No Snowflake driver is added as a dependency** — both surfaces are plain authenticated HTTP.

**Cross-provider connections.** Stages and external volume storage locations link to `AWSS3Bucket` / `GCPBucket` / `AzureStorageAccount` and to the `AWSKMSKey` encrypting them; storage / API / catalog / notification integrations link to the `AWSRole` they assume; notification integrations and auto-ingest pipes link to their `AWSSNSTopic`.

**Ontology.** 24 Snowflake node types are mapped across 16 semantic categories (`Tenant`, `UserAccount`, `ServiceAccount`, `PermissionRole`, `Database`, `NetworkAccessControl`, `ObjectStorage`, `FileStorage`, `Secret`, `APIKey`, `IdentityProvider`, `ComputeCluster`, `ComputeService`, `Container`, `Image`, `ContainerRegistry`, `Function`).

### Related issues or links

- [Snowflake REST API](https://docs.snowflake.com/en/developer-guide/snowflake-rest-api/snowflake-rest-api)

### How was this tested?

- `make test_lint` — clean.
- **2035 unit tests** and **92 integration tests** pass. Integration tests patch only the `get*` / `run_sql` boundary and call `sync()`, asserting nodes and relationships via `check_nodes` / `check_rels`.
- `uv run ./docs/build.sh` exits 0 and generates the 61-node schema page with no Snowflake warnings.
- **Verified end to end against a live Snowflake Enterprise account** with 43 seeded test resources (databases, schemas, tables, secure views, materialized views, dynamic tables, streams, sequences, stages, file formats, secrets, tags, overloaded functions, an owner-rights procedure, a warehouse, a resource monitor, a task DAG, network rule + policy, roles with grants, a database role, a `TYPE = SERVICE` user, an alert, a masking policy, and an image repository).
- **Idempotency / cleanup safety**: two consecutive runs with different update tags produce identical node and relationship counts with zero un-refreshed nodes.

Real sync logs from that account:

```
INFO cartography.intel.snowflake.databases Loading 6 Snowflake databases for account RVTZWBN.UU10371.
INFO cartography.client.core.tx Loaded 6 SnowflakeDatabase nodes
INFO cartography.intel.snowflake.schemas Loading 9 Snowflake schemas for account RVTZWBN.UU10371.
INFO cartography.client.core.tx Loaded 9 SnowflakeSchema nodes
INFO cartography.intel.snowflake.roles Loading 9 Snowflake roles for account RVTZWBN.UU10371.
INFO cartography.client.core.tx Loaded 9 SnowflakeRole nodes
INFO cartography.intel.snowflake.users Loading 1 Snowflake human users and 1 service users for account RVTZWBN.UU10371.
INFO cartography.client.core.tx Loaded 1 SnowflakeUser nodes
INFO cartography.client.core.tx Loaded 1 SnowflakeServiceUser nodes
INFO cartography.intel.snowflake.network_policies Loading 1 Snowflake network policies for account RVTZWBN.UU10371.
INFO cartography.client.core.tx Loaded 1 SnowflakeNetworkPolicy nodes
INFO cartography.intel.snowflake.tasks Loading 2 Snowflake tasks for account RVTZWBN.UU10371.
INFO cartography.client.core.tx Loaded 2 SnowflakeTask nodes
INFO cartography.intel.snowflake.grants Loading 697 Snowflake grant edges and 9 role edges for account RVTZWBN.UU10371.
INFO cartography.client.core.tx Loaded 697 (SnowflakePrincipal)-[HAS_PRIVILEGE]->(SnowflakeSecurable) relationships
INFO cartography.client.core.tx Loaded 2 (SnowflakeUser)-[HAS_ROLE]->(SnowflakeRole) relationships
INFO cartography.client.core.tx Loaded 1 (SnowflakeServiceUser)-[HAS_ROLE]->(SnowflakeRole) relationships
INFO cartography.client.core.tx Loaded 6 (SnowflakeRole)-[INCLUDES]->(SnowflakeRole) relationships
```

Resulting graph: 332 Snowflake nodes, 701 relationships. The real role hierarchy resolves (`ACCOUNTADMIN → SECURITYADMIN → USERADMIN`, `ACCOUNTADMIN → SYSADMIN`), along with `CHILD_TASK -[:PRECEDED_BY]-> ROOT_TASK`, `CARTOGRAPHY_POLICY -[:ALLOWS]-> PLANT_EGRESS`, and `MONITORED_BY` / `USES_WAREHOUSE` / `ASSUMES` / `READS_FROM` / `OWNED_BY`.

#### General

#### Proof of functionality

- [x] Updated the schema documentation (generated from the data model; every schema class has a docstring and every displayed `PropertyRef` a `description=`).
- [x] Updated `docs/root/usage/schema.md`, `docs/root/module-list.md` and the README module list.

### Notes for reviewers

**Why this is one commit.** Two unit guards (`test_sync.py::test_available_modules_import` and `test_doc.py::test_schema_doc`) require `TOP_LEVEL_MODULES`, the packages under `cartography/intel/`, and the schema doc links to agree, so a new module cannot be split without a red intermediate commit.

Areas worth a close look:

1. **Node id scheme** (`sf_id` / `sf_fqn` in `intel/snowflake/util.py`). Ids are `<ORG>.<ACCOUNT>/<object type>/<qualified name>`. Unlike Databricks' `uc_id`, the object type has to be part of the key: a stage, a pipe, a stream and a task can all be named `FOO` in one schema, and a role and a warehouse can share a name at the account level. `sf_fqn` quotes any component that is not a plain uppercase identifier, and is used on **both** sides of every parent/child id computation so a child's recomputed `parent_schema_id` is byte-identical to the parent's own id.

2. **Two shared labels for grants.** `SnowflakePrincipal` and `SnowflakeSecurable` collapse six grantee kinds × ~40 object kinds onto a single `HAS_PRIVILEGE` MatchLink. The role edges in `models/snowflake/role_grant.py` deliberately name **concrete** labels instead: `HAS_ROLE` and `INCLUDES` are mandated by `ONTOLOGY_REL_CONSTRAINTS`, and the CI guard resolves a shared extra label only to itself, so routing them through `SnowflakePrincipal` would have silently exempted them from verification. `HAS_PRIVILEGE` / `OWNS` have no constraint entry, so the shared labels cost nothing there.

3. **Cleanup is gated on completeness.** Any surface Snowflake refuses to answer is skipped *along with its cleanup*, so a missing privilege or an edition limitation never deletes previously collected data. The subtle case: a 403 while listing one database silently truncates the schema list, and the object-level syncs that walk it report `complete=True` in good faith because they read every schema they were handed. Their cleanup is account-scoped, so that incompleteness is propagated to every schema-scoped surface in `_SCHEMA_SCOPED_SURFACES` before `_cleanup` runs. That set is derived from the module tuples rather than hand-written, so adding a surface cannot silently omit it. Both directions are tested (`test_an_unreadable_database_does_not_delete_its_tables` and `test_a_complete_run_still_deletes_stale_tables`).

4. **Account-level grants name the account by its locator**, not by the organization-qualified identifier the account node is keyed on, so `securable_id()` resolves that case specially. Grants also arrive **one row per privilege**, so they are aggregated into one edge carrying a `privileges` list.

5. **Least privilege.** `config.md` documents a genuinely read-only role: identities and grants come from `SNOWFLAKE.ACCOUNT_USAGE` (up to ~2h lag) with only `IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE`, `MONITOR`, and `USAGE`. `MANAGE GRANTS` is documented as *optional*, with the tradeoff stated plainly: it enables the real-time endpoints but is not read-only, since a role holding it can also grant and revoke account-wide.

6. **No secret material is stored.** The users payload returns a redacted `password`, the secrets endpoint returns `secret_string`, and stages return `credentials` / `encryption` objects. None are carried onto nodes; only booleans (`has_password`, `has_credentials`, `has_encryption_key`) and, for SAML certs, a SHA-256 fingerprint.

Known limitations, all documented in the code:

- Columns are not modelled as nodes (cardinality); tables carry `column_count`. This means masking-policy attachment at column granularity is recorded on the table, not the column.
- `SHOW GRANTS` renders function signatures differently from the object API, so grant edges to overloaded functions may not resolve.
- `SnowflakeTag` deliberately does **not** carry the `Tag` ontology label: the canonical concept expects key/value pairs and a Snowflake tag *definition* has no value. Tag assignments would be the correct carrier.
- Cleanup suppression is per-surface, not per-database: one unreadable database keeps stale objects in readable databases for that run too. Per-database scoping would require every schema-level model to scope cleanup to the database rather than the account.


cartography/intel/snowflake/util.py
```diff
@@ -0,0 +1,593 @@
...
+            format=serialization.PublicFormat.SubjectPublicKeyInfo,
+        )
+        digest = hashlib.sha256(public_der).digest()
+        return f"SHA256:{base64.b64encode(digest).decode()}"
+
+    def _mint_jwt(self) -> tuple[str, datetime]:
+        """Return a freshly signed ``(assertion, expiry)`` pair."""
+        private_key = self._load_private_key()
+        fingerprint = self._public_key_fingerprint(private_key)
+        qualified_user = f"{self.account_id}.{self.user}"
```
kunaals: key-pair auth won’t work with this claim. snowflake requires periods in the account identifier to be replaced with hyphens before building `iss` and `sub`; `MYORG.MYACCT.SVC` is rejected, while `MYORG-MYACCT.SVC` is valid. can we keep the dotted id for graph scoping but use the hyphenated account identifier for the jwt claims?

cartography/intel/snowflake/roles.py
```diff
@@ -0,0 +1,100 @@
...
+    neo4j_session: neo4j.Session,
+    client: SnowflakeClient,
+    common_job_parameters: dict,
+) -> list[dict[str, Any]]:
+    """Sync Snowflake account roles and return them for the grant syncs to walk.
+
+    Runs early: users, the role hierarchy and every grant edge resolve against
+    role nodes, and the grant syncs need the role list to know what to enumerate.
+    """
+    roles = transform(get(client), client.account_id)
```
kunaals: this doesn’t actually fall back to `ACCOUNT_USAGE` without `MANAGE GRANTS`. the roles endpoint follows `SHOW ROLES` visibility, so this can return a successful but incomplete list and cleanup will delete the roles it couldn’t see. can we implement the documented `ROLES` / `GRANTS_TO_ROLES` / `GRANTS_TO_USERS` path, or require `MANAGE GRANTS` and skip cleanup when completeness isn’t known?

cartography/intel/snowflake/shares.py
```diff
@@ -0,0 +1,254 @@
...
+            consumer
+            for row in consumers_by_share.get(name, [])
+            for consumer in [to_text(row.get("grantee_name"))]
+            if consumer
+        }
+        consumers.update(name_list(share.get("to")))
+
+        transformed.append(
+            {
+                "id": sf_id(account_id, "share", sf_fqn(name)),
```
kunaals: inbound share ids need the provider account too. `SHOW SHARES` returns the bare share name and a separate `owner_account`, while a database’s `origin` is `<provider_account>.<share_name>`. using only `name` here merges same-named shares from different providers and means `CREATED_FROM_SHARE` can’t match the share built from `origin`. can we key inbound shares with both fields and test the full shares-to-databases path?

cartography/models/snowflake/service.py
```diff
@@ -0,0 +1,535 @@
...
+
+
+@dataclass(frozen=True)
+# (:SnowflakeServiceContainer)-[:HAS_IMAGE]->(:SnowflakeImage)
+class SnowflakeServiceContainerToImageRel(CartographyRelSchema):
+    """A Snowflake service container runs this image from an account image repository."""
+
+    target_node_label: str = "SnowflakeImage"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"digest": PropertyRef("image_digest")},
```
kunaals: matching only on digest can attach a container to every repository copy of the same image. `SHOW SERVICE CONTAINERS` also gives us the full `image_name`, so we can use its repository path together with the digest to resolve one `SnowflakeImage`. can we qualify this matcher and add a duplicate-digest test?

kunaals (changes requested):

cartography/intel/snowflake/grants.py
```diff
@@ -0,0 +1,537 @@
...
+            if granted_to == "USER":
+                kind = (
+                    "service_user_to_role"
+                    if grantee_name in service_user_names
+                    else "user_to_role"
+                )
+                edges[kind].append(
+                    {**common, "grantee_id": sf_id(account_id, "user", grantee_name)},
+                )
+            elif granted_to == "ROLE":
```
kunaals: snowflake returns `ACCOUNT ROLE` here, but this only accepts `ROLE`. that means the new account usage path drops role hierarchy edges. please handle the documented value and update the fixture to match the real response.

cartography/intel/snowflake/access_tokens.py
```diff
@@ -0,0 +1,180 @@
...
+                "role_restriction": role_restriction,
+                # Role node ids are keyed on the bare role name, matching the role
+                # sync; null when the token is unrestricted.
+                "role_restriction_id": (
+                    sf_id(account_id, "role", role_restriction)
+                    if role_restriction
+                    else None
+                ),
+                "status": to_text(token.get("status")),
+                "mins_to_bypass_network_policy_requirement": to_int(
```
kunaals: the response field is `mins_to_bypass_required_network_policy`, not `mins_to_bypass_network_policy_requirement`. right now this property will always be null. please use the provider field name and update the fixture.

docs/root/modules/snowflake/config.md
```diff
@@ -0,0 +1,277 @@
...
+complete one, so without the `ACCOUNT_USAGE` views Cartography cannot establish
+that it saw every role. It then keeps the data it has and skips role, database
+role and grant cleanup rather than risk deleting roles it merely could not see.
+The graph stays correct, but stale roles are never removed.
+
+## Optional Permissions
+
+| Privilege | Enables | Cost of granting |
+|---|---|---|
+| `MANAGE GRANTS ON ACCOUNT` | Makes the object API and `SHOW GRANTS` account-wide, so roles and grants are current rather than up to two hours stale. Also enables enumerating other users' programmatic access tokens. | Not read-only: a role with `MANAGE GRANTS` can also grant and revoke privileges account-wide. Grant it only if the staleness matters more than the blast radius. |
```
kunaals: `MANAGE GRANTS` doesn’t give this role access to other users’ programmatic access tokens. snowflake requires a user-level privilege such as `MODIFY PROGRAMMATIC AUTHENTICATION METHODS`, so this setup won’t produce the inventory described here. please update the permission guidance and the related log message.

kunaals (changes requested):

kunaals (approved):

PR #3112, opened 2026-08-04
feat(anthropic): add Workload Identity Federation auth and expand Admin API coverage

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds Workload Identity Federation as an authentication method for the Anthropic module, and expands coverage from 4 node types to 19.

Two things changed on Anthropic's side since the module was written:

1. **The Admin API now accepts OAuth bearer tokens** with the `org:admin` scope alongside `sk-ant-admin` keys, and Workload Identity Federation (GA) mints them. A collector running in Kubernetes, EKS, GKE or GitHub Actions can now drive the API with no long-lived secret.
2. **Some endpoints refuse Admin API keys outright.** Service accounts, federation issuers and federation rules return 403 to a `sk-ant-admin` key and are only readable with an `org:admin` OAuth token, so cartography could not see them at all. Those are exactly the IAM primitives a security graph should carry.

Both auth methods are supported. Federation wins when both are configured; a partially configured federation triple raises rather than silently falling back to skipping the module.

**New node types** (15): `AnthropicServiceAccount`, `AnthropicFederationIssuer`, `AnthropicFederationRule`, `AnthropicInvite`, `AnthropicRbacGroup`, `AnthropicRbacRole`, `AnthropicRbacRolePermission`, `AnthropicRateLimit`, `AnthropicSkill`, `AnthropicSkillVersion`, `AnthropicAgent`, `AnthropicEnvironment`, `AnthropicDeployment`, `AnthropicVault`, `AnthropicMemoryStore`.

**Enriched existing nodes**: `AnthropicOrganization` gains a real `name` from `GET /organizations/me` (the id was previously scraped from a response header and defaulted to `""`), which lets it map onto the Tenant ontology. `AnthropicApiKey` gains `expires_at` and the `principal` it acts as, which is what distinguishes a machine credential from a human's. `AnthropicWorkspace` gains its customer-managed encryption key binding and data-residency posture.

The federation subgraph makes the trust chain queryable end to end: a rule declares which assertions from which issuer may mint a token, `ASSUMES` the service account that token acts as, and that service account carries the `organization_role` the token inherits. An over-broad matcher becomes a visible escalation path, such as a GitHub Actions `subject_prefix` ending in a wildcard, which also matches fork pull requests.

Six commits, one per logical change.

### Related issues or links

- [Workload Identity Federation](https://platform.claude.com/docs/en/manage-claude/workload-identity-federation)
- [WIF reference](https://platform.claude.com/docs/en/manage-claude/wif-reference)
- [Manage WIF with the Admin API](https://platform.claude.com/docs/en/manage-claude/wif-admin-api)

### How was this tested?

- **Unit** (`tests/unit/cartography/intel/anthropic/test_auth.py`, 15 new tests): the exchange request body, that the resulting header is `Authorization: Bearer` and not `X-Api-Key`, that a near-expiry token triggers exactly one refresh, that the assertion is re-read from disk on every exchange, and the configuration guards.
- **Integration** (11 test files under `tests/integration/cartography/intel/anthropic/`): one per node type, asserting nodes and every relationship.
- One regression test runs two syncs at different update tags so it exercises cleanup for real, and was confirmed to fail without its fix.
- Full suites green locally: 4572 unit, 1347 integration. `pre-commit run --all-files` clean.

**Not tested**: the live token exchange. It needs a real organization with a Console-created `org:admin` federation rule, a service account with `organization_role: admin`, and an OIDC assertion, none of which a fixture can stand in for. Flagging this explicitly rather than checking the sync-logs box below.

#### General

#### Proof of functionality

- [ ] Included Cartography sync logs from a real environment. **Not provided**: see the note above. The syncs are covered by integration tests against a real Neo4j instead.

- [x] Updated the schema documentation. Node and relationship descriptions live in the dataclasses and render into the generated schema page; `docs/root/modules/anthropic/config.md` was rewritten for both auth methods.
- [ ] Updated the schema README. Not applicable: no new module.

### Notes for reviewers

Three decisions worth a look:

1. **`requests.auth.AuthBase`** is a new idiom for this repo, used so a federated token can refresh itself mid-sync without callers knowing. The closest precedent for the problem is Tailscale, which solves it with a 401 response hook instead.

2. **Per-workspace failures are warnings, not fatal.** Skills, agents and the rest have no organization-wide listing and are out of reach of an `org:admin` token, and there is no way to downscope one: the token endpoint accepts only the `jwt-bearer` grant, with no RFC 8693 exchange and no request-side scope. So each workspace needs its own exchange, and a failed one is expected rather than exceptional. The service account may not be a member of that workspace, or replay protection may have rejected a reused assertion, and both come back as an opaque `400 invalid_grant` that gives no way to tell them apart. Workspace-plane nodes therefore hang off their workspace rather than the organization, so one workspace failing to exchange cannot delete another's resources at cleanup.

3. **A transient `null` on an RBAC group's `roles`** is carried forward from the graph rather than written as empty. Skipping the load would not have been enough, because cleanup deletes any relationship whose `lastupdated` misses the current tag. The residual tradeoff: a role legitimately removed during an outage leaves a stale edge until the next healthy sync. That direction is deliberate, since over-declaring a permission is visible and self-correcting while under-declaring it is silent.

Deliberately out of scope, all noted in the code or docs: `/v1/models` (a static public catalog), `/v1/files`, `/v1/sessions`, `/v1/messages/batches` (high-volume transient records that would dominate the graph), `/v1/tunnels` (needs a third federation rule with `workspace:manage_tunnels`, which cannot share a rule with `workspace:developer`), and the Compliance API (a fourth credential family that does not accept `org:admin` tokens). No security rules either; the data is in place to write them in a follow-up.

### DIFF
--- a/cartography/intel/anthropic/__init__.py
+++ b/cartography/intel/anthropic/__init__.py
@@ -6,6 +6,7 @@ import requests
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
+import cartography.intel.anthropic.agents
 import cartography.intel.anthropic.apikeys
 import cartography.intel.anthropic.federation
 import cartography.intel.anthropic.invites
@@ -13,16 +14,80 @@ import cartography.intel.anthropic.organization
 import cartography.intel.anthropic.ratelimits
 import cartography.intel.anthropic.rbac
 import cartography.intel.anthropic.serviceaccounts
+import cartography.intel.anthropic.skills
 import cartography.intel.anthropic.users
 import cartography.intel.anthropic.workspaces
 from cartography.config import Config
 from cartography.intel.anthropic.auth import AnthropicAuth
+from cartography.intel.anthropic.auth import AnthropicCredential
 from cartography.intel.anthropic.auth import is_federated
 from cartography.intel.anthropic.auth import make_assertion_source
 from cartography.intel.anthropic.auth import make_credential
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
+
+
+def _build_api_session(credential: AnthropicCredential) -> requests.Session:
+    api_session = requests.session()
+    retry_policy = Retry(
+        total=5,
+        backoff_factor=1,
+        status_forcelist=[429, 500, 502, 503, 504],
+        allowed_methods=["GET"],
+    )
+    api_session.mount("https://", HTTPAdapter(max_retries=retry_policy))
+    api_session.auth = AnthropicAuth(credential)
+    api_session.headers.update({"anthropic-version": "2023-06-01"})
+    return api_session
+
+
+def _sync_workspace_plane(
+    neo4j_session: neo4j.Session,
+    config: Config,
+    common_job_parameters: dict[str, Any],
+    workspaces: list[dict[str, Any]],
+) -> None:
+    """Sync the per-workspace resources, one federated token per workspace.
+
+    Skills, agents and the rest have no organization-wide listing and are out of
+    reach of an org:admin token, so each workspace needs its own exchange against a
+    workspace:developer federation rule. There is no way to downscope the org:admin
+    token already in hand: the only grant type the token endpoint accepts is
+    jwt-bearer, so every workspace re-presents the original identity token.
+    """
+    for workspace in workspaces:
+        workspace_id = workspace["id"]
+        try:
+            credential = make_credential(
+                identity_token_file=config.anthropic_identity_token_file,
+                identity_token_env_var=config.anthropic_identity_token_env_var,
+                federation_rule_id=config.anthropic_workspace_federation_rule_id,
+                organization_id=config.anthropic_organization_id,
+                service_account_id=config.anthropic_service_account_id,
+                workspace_id=workspace_id,
+            )
+            api_session = _build_api_session(credential)
+            workspace_job_parameters = {
+                **common_job_parameters,
+                "WORKSPACE_ID": workspace_id,
+            }
+            cartography.intel.anthropic.skills.sync(
+                neo4j_session, api_session, workspace_job_parameters
+            )
+            cartography.intel.anthropic.agents.sync(
+                neo4j_session, api_session, workspace_job_parameters
+            )
+        except requests.exceptions.HTTPError as exc:
+            # A failed exchange is expected and undiagnosable: the service account
+            # may not be a member of this workspace, or replay protection may have
+            # rejected a reused assertion. Both surface as an opaque 400
+            # invalid_grant. Skip the workspace rather than abandon the others.
+            logger.warning(
+                "Skipping the workspace plane for Anthropic workspace %s: %s",
+                workspace_id,
+                exc,
+            )
 
 
 def _sync_enterprise_rbac(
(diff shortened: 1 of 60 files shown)


cartography/intel/anthropic/__init__.py
```diff
@@ -1,18 +1,192 @@
...
+    common_job_parameters: dict[str, Any],
+    workspaces: list[dict[str, Any]],
+) -> None:
+    """Sync the per-workspace resources, one federated token per workspace.
+
+    Skills, agents and the rest have no organization-wide listing and are out of
+    reach of an org:admin token, so each workspace needs its own exchange against a
+    workspace:developer federation rule. There is no way to downscope the org:admin
+    token already in hand: the only grant type the token endpoint accepts is
+    jwt-bearer, so every workspace re-presents the original identity token.
```
kunaals, line 70: this assertion is exchanged once for the organization and then reused for every workspace. federation issuers check `jti` for single use by default, so the workspace exchanges can be rejected as replays. each exchange needs a fresh assertion, or we need to document and enforce a compatible issuer configuration.

cartography/intel/anthropic/auth.py
```diff
@@ -0,0 +1,328 @@
...
+        self._assertion_source = assertion_source
+        self._federation_rule_id = federation_rule_id
+        self._organization_id = organization_id
+        self._service_account_id = service_account_id
+        self._workspace_id = workspace_id
+        self._base_url = base_url
+        self._token: str | None = None
+        self._token_expires_at: float = 0
+
+    def get_headers(self) -> dict[str, str]:
```
kunaals, line 141: a fixed five-minute refresh window makes a valid 60-second token look expired immediately. please follow anthropic’s refresh windows and keep using a cached token until the mandatory refresh point if an early refresh fails.

cartography/intel/anthropic/agents.py
```diff
@@ -0,0 +1,262 @@
...
+) -> list[dict[str, Any]]:
+    return paginated_get_by_page(
+        api_session,
+        f"{base_url}/memory_stores",
+        timeout=_TIMEOUT,
+        headers=_AGENT_MEMORY_HEADERS,
+    )
+
+
+def transform_agent(agent: dict[str, Any]) -> None:
```
kunaals, line 124: the managed agents response has objects under `model.effort` and the tool configuration rather than the scalar values expected here. this can send a map to neo4j and mark allowed tools incorrectly. please transform the documented response shape and add a provider-shaped fixture.

cartography/intel/anthropic/__init__.py
```diff
@@ -1,18 +1,192 @@
...
+                **common_job_parameters,
+                "WORKSPACE_ID": workspace_id,
+            }
+            cartography.intel.anthropic.skills.sync(
+                neo4j_session, api_session, workspace_job_parameters
+            )
+            cartography.intel.anthropic.agents.sync(
+                neo4j_session, api_session, workspace_job_parameters
+            )
+        except requests.exceptions.HTTPError as exc:
```
kunaals: this only catches `HTTPError`. retry exhaustion, timeouts, and connection failures will still stop the loop and prevent later workspaces from syncing. please catch the request-level exceptions that this client can raise and skip only the affected workspace.

cartography/intel/anthropic/__init__.py
```diff
@@ -23,49 +197,126 @@ def start_anthropic_ingestion(neo4j_session: neo4j.Session, config: Config) -> N
...
+            "endpoints reject Admin API keys. Configure Workload Identity "
+            "Federation to ingest them.",
+        )
+
     cartography.intel.anthropic.apikeys.sync(
         neo4j_session,
         api_session,
         common_job_parameters,
     )
+
```
kunaals, line 351: the default workspace has no id and isn’t returned by the workspace list, so it never gets synced here. anthropic accepts `default` as its workspace identifier; please include that workspace explicitly.

cartography/intel/anthropic/ratelimits.py
```diff
@@ -0,0 +1,150 @@
...
+    The API returns limits as a nested list of objects, which Neo4j cannot store as
+    a node property, and gives no identifier for either the group or the individual
+    limit. The id is synthesised from the fields that make a limit unique: its scope,
+    group type, models and limit type.
+    """
+    results: list[dict[str, Any]] = []
+    for group in groups:
+        group_type = group.get("group_type")
+        models = group.get("models") or []
+        scope = workspace_id or "organization"
```
kunaals, line 119: this id doesn’t include the organization, so identical rate-limit records from two organizations will merge into one node. please scope the id to the organization.

kunaals (changes requested):

PR #3117, opened 2026-08-04
feat(miradore): add Miradore MDM intel module

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds a Miradore intel module. Miradore is an MDM/UEM comparable to Jamf, Kandji and Intune,
all of which already have modules; this closes the gap so a Miradore-managed fleet becomes
visible in cross-provider device and identity queries.

The module ingests eight node types: `MiradoreTenant`, `MiradoreDevice`, `MiradoreUser`,
`MiradoreOrganization`, `MiradoreLocation`, `MiradoreTag`, `MiradoreConfigProfile` and
`MiradoreConfigProfileDeployment`.

**Why API v1 and not the Swagger-documented v2.** Miradore's Swagger page documents API v2,
which is action-oriented and cannot be used for inventory ingestion: it has no list endpoint
at all, only `POST /api/v2/Device` (create), `PATCH`/`DELETE /api/v2/Device/{id}`,
`POST /api/v2/Device/{id}/{Lock,Wipe,Reboot,LostMode}`, and two marginal GETs
(`/api/v2/Device/{id}/Location`, `/api/v2/Device/SuspendedDevices`). Reading an inventory
requires API v1, an XML REST interface whose `get` operation supports `select`, `filters`,
`orderby` and paging. Responses are parsed with `xmltodict`, already a dependency and already
used by the BigFix module, so this adds no dependency.

**One device node, not one per platform.** Unlike Jamf, the Miradore API models every
platform as a single `Device` item discriminated by a `Platform` enum, so there is one
`MiradoreDevice`. It carries the platform-specific security posture Miradore reports:
encryption and passcode state, iOS supervision and jailbreak detection, Android root state
and patch level, macOS Activation Lock, and Windows Secure Boot, antivirus and firewall
status.

**Tenant-scoped identities.** Miradore numbers its items per tenant, so `Device` 1001 exists
in every site. The ingestion query merges on `id` alone, so a raw identifier would let two
sites overwrite each other's properties and cross-wire their relationships. Nodes are
identified as `<site name>/<Miradore ID>`, the raw identifier is kept as `miradore_id`, and
every relationship matcher resolves on the scoped identity. This follows the existing
convention (`f"{account_id}/{name}"` for Cloudflare Worker scripts,
`f"{org_id}/{version}"` for Sentry releases).

**Ontology.** `MiradoreDevice` feeds canonical `Device` nodes matched on serial number, with
a hostname matchlink as fallback. The serial is coalesced from `Device.SerialNo` (added in
API spec 1.20), `InvDevice.SerialNumber`, then `InvDevice.HardwareSerialNumber`.
`MiradoreUser` carries the `UserAccount` label, so canonical `User` nodes gain `OWNS` edges
to their devices by traversing `HAS_ACCOUNT`.

### Related issues or links

- [Miradore API v1 specification](https://www.miradore.com/knowledge/integrations/programmers-guide-to-api-v1/)
- [Miradore API v2](https://www.miradore.com/knowledge/integrations/miradore-api-v2/) (no inventory read endpoints, hence v1)

### How was this tested?

Unit and integration tests, plus an end-to-end run against a real Miradore tenant.

```
uv run --frozen pre-commit run --all-files   # all hooks pass (cold mypy cache)
uv run --frozen pytest tests/unit            # 4609 passed
uv run --frozen pytest tests/integration     # 1402 passed
```

Three findings came out of testing against the live tenant, each of which changed the
implementation or confirmed a design choice:

1. The spec documents `select=ChildItem.ChildAttribute`, one level only, but the security
   data lives two levels deep (`Security.iOS.*`, `Security.Android.*`). A live request with
   those paths returns HTTP 200. That is not a false positive: a control request with a bogus
   property is rejected with `HTTP 400 "Entity 'Device' does not have property
   'ThisFieldDoesNotExist'"`, so the API does validate selects.
2. The `count` attribute on `<Items>` is the **total** number of matching items, not the page
   size: with `rows=5`, `ConfigProfile` returns `count="10"` with 5 elements, then 5 more on
   page 2, then 0 on page 3 still reporting `count="10"`. Paging therefore terminates on a
   short page rather than on `count`.
3. `ConfigProfile.OSCategory` returns `WindowsDesktop` and `macOS`, which spec 1.20 does not
   document (it lists `Android, iOS, WindowsPhone`). Its property description was corrected to
   stay open-ended rather than assert a closed set.

#### General

#### Proof of functionality

```
INFO:cartography.sync:Starting sync stage 'miradore'
INFO:cartography.client.core.tx:Loaded 1 MiradoreTenant nodes
INFO:cartography.graph.statement:Completed MiradoreOrganization statement #1
INFO:cartography.graph.statement:Completed MiradoreLocation statement #1
INFO:cartography.graph.statement:Completed MiradoreTag statement #1
INFO:cartography.client.core.tx:Loaded 10 MiradoreConfigProfile nodes
INFO:cartography.client.core.tx:Loaded 1 MiradoreUser nodes
INFO:cartography.sync:Finishing sync stage 'miradore'
INFO:cartography.sync:Finishing sync with update tag '1785864100'
```

Graph state after that run, with the `ontology` stage enabled:

```
MiradoreTenant           1     id = "subimage", labels [MiradoreTenant, Tenant]
MiradoreConfigProfile    10    ids "subimage/1" .. "subimage/10", all RESOURCE-attached
MiradoreUser             1     id "subimage/1", labels [MiradoreUser, UserAccount]
                               _ont_email set, _ont_active = true, _ont_source = "miradore"
User (canonical)         1     built from MiradoreUser, HAS_ACCOUNT edge present
```

Timestamps land as native Neo4j temporals (`2026-08-04T16:36:07`), not epoch integers.

The tenant used for the run has no enrolled devices, so `MiradoreDevice` and the device-side
paths (`OBSERVED_AS`, `OWNS`, tags, profile deployments) are covered by the integration tests
rather than by live data. Worth flagging explicitly rather than implying full live coverage.

Schema pages are generated from the data model; `docs/root/usage/schema.md` was updated to
link the new module, and `docs/root/index.md` / `docs/root/module-list.md` list it.

### Notes for reviewers

Worth a close look:

- **`cartography/intel/miradore/util.py`** — the `scoped_id()` helper and the pagination loop.
  The API key travels in the query string (API v1 has no header auth), so it is only ever
  passed via `requests`' `params` and never interpolated into a URI that could reach a log
  line. A unit test asserts the key is absent from the request URI, and the captured sync log
  above contains no occurrence of it.
- **`tests/integration/cartography/intel/miradore/test_miradore.py`** —
  `test_two_tenants_with_the_same_miradore_ids_stay_isolated` syncs identical fixtures under
  two site names and asserts that no relationship crosses a tenant boundary, using a generic
  Cypher check on the id prefix so it also covers relationships added later. I verified it
  fails when `scoped_id()` is neutralised.
- Relationship labels reuse existing conventions: `RESOURCE`, `MEMBER_OF`, `OWNS`, `TAGGED`
  (as in the Azure tag models) and `HAS_DEPLOYMENT` (as in AWS API Gateway). No new label was
  invented. `Category` is folded into `MiradoreDevice` as properties rather than given its own
  node, since it carries no data beyond an id and a name.

Deliberately out of scope: `InvApplication` (per-device app lists, high volume, and the v1
`select` grows large), and the pure hardware telemetry sub-items (`InvAndroidCamera`,
`InvAndroidSensor`, `NetworkUsage`, `ReportedLocation`, `PhysicalMemory`, `LogicalDisk`,
`Battery`, `BIOS`, `CPU`, `InvStorage`).

Unrelated pre-existing issue noticed while wiring the ontology, not addressed here: the
canonical `User` builder in `cartography/models/ontology/mapping/data/users.py` reads
`_ont_inactive`, while 19 of the 20 provider mappings in `useraccounts.py` write `_ont_active`.
`User.active` is therefore null for nearly every provider, including this one. Fixing it means
touching shared ontology code for all providers, so it belongs in its own change.


cartography/intel/miradore/util.py
```diff
@@ -0,0 +1,201 @@
...
+        try:
+            response = api_session.get(uri, params=params, timeout=_TIMEOUT)
+        except requests.exceptions.Timeout:
+            logger.warning(
+                "Miradore: requests.get('%s') timed out on page %d.",
+                uri,
+                page,
+            )
+            raise
+        response.raise_for_status()
```
kunaals: `params` still ends up in the prepared url, and `raise_for_status()` includes that full url in the exception. since the sync runner logs the exception, a 400/401 response will print the miradore api key. can we raise a sanitized error that keeps the status and request path but drops the query string, with a regression test that checks the key never appears?

cartography/intel/miradore/util.py
```diff
@@ -0,0 +1,201 @@
...
+    return f"rows={page_size},page={page},dateformat={_API_DATE_FORMAT}"
+
+
+def _extract_items(payload: dict[str, Any], item: str) -> list[dict[str, Any]]:
+    """Pull the `<Item>` elements out of a parsed `<Content><Items>` document.
+
+    xmltodict collapses a single repeated element into a dict rather than a one-element
+    list, and omits the key entirely when the page is empty, so normalize both here.
+    """
+    items = (payload.get("Content") or {}).get("Items")
```
kunaals: this treats an unexpected 200 response the same as an empty page. if miradore or a proxy returns an error or login document without `Content.Items`, the sync continues into cleanup and deletes that resource’s existing nodes. can we raise unless the expected envelope is present, while still accepting a real `<Items count="0" />` response?

kunaals (changes requested):

cartography/intel/miradore/util.py
```diff
@@ -0,0 +1,228 @@
...
+    while True:
+        # The API key travels in the query string, so it must only ever be passed via
+        # `params`. Never interpolate it into a URI that could reach a log line.
+        params = {
+            "auth": api_key,
+            "select": select,
+            "options": _build_options(page, page_size),
+        }
+        try:
+            response = api_session.get(uri, params=params, timeout=_TIMEOUT)
```
kunaals: the normal http error path is sanitized now, but an exhausted retry still raises before that helper runs. `RetryError` includes the request url, so a 429 or 5xx can still put `auth=<api key>` in the logs. please catch the request exception here and raise a key-free error `from None`, then add a test that goes through the retry adapter.

kunaals (changes requested):

kunaals (approved):

PR #3127, opened 2026-08-04
perf(tests): cut unit tests 60s->29s and integration 310s->229s

jychp:
### Type of change

- [ ] Documentation update
- [x] Other (please describe): CI performance

### Summary

CI test time was 3m46s for unit tests and over 10 minutes for integration tests. Neither number came from the tests doing too much work: both were dominated by two specific pathologies.

**Unit tests: 60s -> 29s** (4584 tests, locally; CPU time 96.2s -> 25.8s)

`tests/utils.load_models()` walked every model module using `issubclass(v, _MODEL_CLASSES)`. All four bases are `abc.ABC` with very large subclass trees (`CartographyRelSchema` alone has ~1973 subclasses), so every miss walked the whole tree through `ABCMeta.__subclasscheck__`, and the ABC negative cache was invalidated continuously by the new classes being imported during the walk itself:

```
6.98M _abc_subclasscheck calls for 9523 issubclass calls
load_models() call 1: 10.39s   (5733 classes)
load_models() call 2:  8.35s   (5733 classes)   <- the cache never warms
MRO membership test:   0.018s  (5733 classes)   <- identical results, verified equal
```

One pass cost 10.4s, and it was called uncached 20+ times across the guard tests (10 times in `test_model.py` alone). Walking the MRO instead is what `cartography/models/introspection.py:548` already does in production code; only `tests/utils.py` used `issubclass`. Memoizing the walk makes every call after the first free.

The ~20 per-call-site `issubclass(element, Cartography*)` loops paid the same one-time ABC cache warm-up (2.77s on the first pass, 0.002s after), so they now use `TypeGuard` predicates that preserve the type narrowing `issubclass` gave mypy.

**Integration tests: 310s -> 229s** (1400 tests, locally, `--no-cov`)

`test_sync_collaborators_per_repo` and `test_sync_circleci_label_provenance_links_github_repository` patch `repos.get` and the collaborator helpers, but not `get_repo_privileged_details_by_url`. Their repo fixtures carry no privileged fields, so `_repos_need_privileged_details()` returns true and `sync()` goes looking for them: `handle_rate_limit_sleep()` issues a **real request to api.github.com**, the fake token gets a 401, and `fetch_all()` retries five times with exponential backoff (2+4+8+16s) before `sync()` swallows the error and the test passes.

Profiling the two tests together showed **63.03s of 73s was `time.sleep`**. They drop from 31.9s and 31.3s to 3.0s and 3.2s. This also removes a CI dependency on GitHub's API being reachable and unthrottled from the runner IP.

**Also:** the CI test job is split so unit and integration run in parallel rather than sequentially, and coverage runs under the `sys.monitoring` backend, which measures the same statements for less overhead on Python 3.12+.

### Related issues or links

N/A

### How was this tested?

Everything below was measured locally, before and after, on Python 3.13.

Unit suite, identical pass count and identical coverage:

```
before: 4584 passed in 59.95s   TOTAL 106376 stmts / 25752 missed / 75.79%
after:  4584 passed in 29.40s   TOTAL 106376 stmts / 25752 missed / 75.79%
```

The five model-walking guard test files, run in isolation, went from 51.7s combined (12.4 + 9.3 + 10.6 + 9.4 + 10.0) to 9.09s. What remains is the ~8s of `cartography.intel` import cost, not the walk.

Integration suite, against a `neo4j:5-community` testcontainer:

```
before: 1400 passed in 310.14s
after:  1400 passed in 229.43s
```

`make test_lint` passes.

### Notes for reviewers

**Branch protection needs updating.** Splitting the test job removes the `unit-and-integration-tests` required check and replaces it with `unit-tests` and `integration-tests`. Without that change, PRs will block on a check that never reports.

**Two optimizations were tried, measured, and deliberately dropped** rather than shipped on a plausible-sounding hypothesis:

1. *Deduplicating index DDL.* `load()` calls `ensure_indexes()` unconditionally, so the suite issues 43748 `CREATE INDEX IF NOT EXISTS` statements for the 3589 distinct indexes the models define. Memoizing it cut that to 3623 calls and removed 25s of measured DDL time, but wall clock did not improve (226.6s and 232.2s across two runs, against 229.4s without it) and `load()` was consistently ~10s slower. The likely mechanism is that the repeated no-op statements give indexes time to reach ONLINE before the MERGEs that use them. Worth revisiting with a real hypothesis, but not shipped here.
2. *Putting the container's `/data` on tmpfs.* Only 1.12x lower commit latency, and a trivial write transaction costs 3.35ms against ~17ms for a `load()`, so transaction durability is not the constraint.

**Where the remaining integration time goes.** Only 32s of the 229s is Python CPU; the rest is blocked on Neo4j round trips. Instrumenting `neo4j.Session` by query category:

| category | time | calls |
|---|---|---|
| `load()` ingestion | 75.2s | 4775 |
| cleanup statements | 47.5s | 9806 |
| `CREATE INDEX` (DDL) | 35.8s | 43748 |
| reads / assertions | 21.8s | 5051 |
| `DETACH DELETE` | 2.2s | 1154 |

Worth noting that the ~800 full-graph wipes cost 2.2s in total, so they are not the problem they look like. Most of what is left is useful work: ingestion and cleanup.

**One issue found but not fixed here:** `tests/integration/cartography/intel/azure/test_deferred_cleanup.py` resolves `management.azure.com`, because a `MagicMock` credential leaks through to the Azure SDK. The time cost is negligible (under 0.8s), but it is a real network call from CI. I swept the whole integration suite for non-local DNS resolution and these two tests plus the two GitHub ones fixed here were the only hits.


tests/integration/cartography/intel/github/test_repos.py
```diff
@@ -1010,6 +1010,14 @@ def collabs_side_effect(repo_raw_data, affiliation, org, api_url, token):
     assert check_nodes(neo4j_session, "GitHubRulesetRule", ["id"]) == set()
 
 
+# The test repo data carries no privileged fields, so `sync()` reaches out for them.
+# Stub the fetch: unpatched it calls the real api.github.com and burns ~30s in retry
```
achantavy, line 1036: lol wow, good catch

tests/utils.py
```diff
@@ -30,9 +32,35 @@ def unwrapper(func):
...
+_MODEL_CLASS_SET = frozenset(_MODEL_CLASSES)
 
 
-def load_models(module, module_name: str | None = None) -> Generator[
+# The model base classes are ABCs with thousands of subclasses, so `issubclass()`
+# against them goes through `ABCMeta.__subclasscheck__` and walks the whole subclass
+# tree on every miss. Walk the MRO instead, like `cartography.models.introspection`
+# does. These are the hot predicates for the guard tests that iterate over every
+# model class in the codebase.
+def is_node_schema(cls: Type) -> TypeGuard[Type[CartographyNodeSchema]]:
```
achantavy, line 43: One day I hope we won't need these to check but oh well; I get it

achantavy (approved):

PR #3130, opened 2026-08-05
feat(rules): detect ChainDrop (Aug 2026) Shai-Hulud npm compromise

EmilioDNA:
### Type of change

- [ ] Documentation update

### Summary

Adds two `Fact`s to the `malicious_npm_dependencies_shai_hulud` rule covering the August 2026 "ChainDrop" wave of the Shai-Hulud npm supply-chain attack. The wave began with a takeover of the `keyv` maintainer account and spread through the keyv/cacheable dependency family into unrelated organizations via stolen publishing tokens. Malicious releases carry valid GitHub Actions provenance, so provenance attestation cannot distinguish them from legitimate releases.

- **`malicious-npm-dependencies-shai-hulud-aug-2026-github`**: matches exactly pinned malicious versions across the confirmed package set, including the full `@keyv/*` scope republished at `6.0.0`.
- **`malicious-npm-dependencies-shai-hulud-aug-2026-at-risk-github`**: covers blast radius. Several releases in this wave are patch bumps (`flat-cache` `6.1.24`, `file-entry-cache` `11.1.6`), so an existing floating range such as `^6.1` resolves to a malicious version without any manifest change. This `Fact` is scoped to requirements containing `^`, `~`, or `>` so it stays disjoint from the pinned `Fact`, and it normalizes the operator prefixes the dependency graph emits before comparing major versions.

The rule version is bumped from `0.2.0` to `0.3.0` and new `RuleReference`s cite public advisories for the ChainDrop attack (StepSecurity, JFrog, Aikido.dev, BleepingComputer).

### How was this tested?

Added 14 new unit tests in `tests/unit/rules/test_malicious_npm_dependencies_shai_hulud.py` covering: rule registration, fact ID uniqueness, ChainDrop entry-point coverage, `@keyv/*` scoped-family coverage, parity between the pinned and at-risk fact package sets, query/visual-query agreement, the floating-range-only scoping of the at-risk fact, operator-prefix normalization, `identity_fields`/`asset_id_field` correctness, and archived/disabled repo exclusion.

```
uv run --frozen pytest -q tests/unit/rules/test_malicious_npm_dependencies_shai_hulud.py
14 passed

uv run --frozen pytest -q tests/unit/rules/
2553 passed
```

`pre-commit run -a` (isort, black, flake8, pyupgrade, mypy) passes on the changed files.

#### General

#### Proof of functionality

### Notes for reviewers

This wave is still evolving; the package list reflects confirmed IOCs as of 2026-08-04 and is not necessarily exhaustive for the full campaign. Both new facts are marked `Maturity.EXPERIMENTAL` per the existing convention for newly added detections in this rule.


cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
```diff
@@ -3079,6 +3079,248 @@
...
+        { name: '@cacheable/net', version: '2.1.1' },
+        { name: 'ecto', version: '5.0.1' },
+        { name: 'picasso.js', version: '2.11.6' },
+        { name: '@deliveroo/reevent', version: '1.0.1' },
+        { name: '@or-sdk/invitations', version: '1.4.9' },
+        { name: '@picsart/ai-sdk', version: '3.32.2' },
+        { name: '@qlik/embed-runtime', version: '1.6.4' }
+    ] AS vulnerable
+    UNWIND vulnerable AS v
+    MATCH (d:Dependency {ecosystem: 'npm', name: v.name})--(manifest:GitHubDependencyGraphManifest)--(r:GitHubRepository)
```
achantavy: Use exact rel labels whereever possible in a rule instead of `--`

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
```diff
@@ -3079,6 +3079,248 @@
...
+      { name: 'ecto', version: '5.0.1' },
+      { name: 'picasso.js', version: '2.11.6' },
+      { name: '@deliveroo/reevent', version: '1.0.1' },
+      { name: '@or-sdk/invitations', version: '1.4.9' },
+      { name: '@picsart/ai-sdk', version: '3.32.2' },
+      { name: '@qlik/embed-runtime', version: '1.6.4' }
+    ] AS vulnerable
+    UNWIND vulnerable AS v
+        MATCH path = (d:Dependency {ecosystem: 'npm', name: v.name})
+                    --(manifest:GitHubDependencyGraphManifest)--(r:GitHubRepository)
```
achantavy: rel label

tests/unit/rules/test_malicious_npm_dependencies_shai_hulud.py
```diff
@@ -0,0 +1,145 @@
+import re
```
achantavy, line 1: Can you explain why these tests are needed? I almost want to remove this file because at first glance I don't see why they're unique from testing the standard rules framework.

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
```diff
@@ -3079,6 +3079,248 @@
...
+      { name: 'ecto', major: '5', version: '5.0.1' },
+      { name: 'picasso.js', major: '2', version: '2.11.6' },
+      { name: '@deliveroo/reevent', major: '1', version: '1.0.1' },
+      { name: '@or-sdk/invitations', major: '1', version: '1.4.9' },
+      { name: '@picsart/ai-sdk', major: '3', version: '3.32.2' },
+      { name: '@qlik/embed-runtime', major: '1', version: '1.6.4' }
+    ] AS at_risk
+    UNWIND at_risk AS a
+        MATCH path = (d:Dependency {ecosystem: 'npm', name: a.name})
+                    --(manifest:GitHubDependencyGraphManifest)--(r:GitHubRepository)
```
achantavy: rel names

achantavy (changes requested):
Quick first pass, thanks for the PR!

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
```diff
@@ -3079,6 +3079,248 @@
...
+        { name: '@picsart/ai-sdk', version: '3.32.2' },
+        { name: '@qlik/embed-runtime', version: '1.6.4' }
+    ] AS vulnerable
+    UNWIND vulnerable AS v
+    MATCH (d:Dependency {ecosystem: 'npm', name: v.name})--(manifest:GitHubDependencyGraphManifest)--(r:GitHubRepository)
+    WHERE REPLACE(d.requirements, "= ", "") = v.version
+      AND coalesce(r.archived, false) = false
+      AND coalesce(r.disabled, false) = false
+    RETURN r.fullname as repo, d.name as name, d.requirements as current_version, v.version AS vulnerable_version
+    """,
```
kunaals, line 3125: should this return `r.id` through a separate asset id field? `repo` currently contains `r.fullname`, but `GitHubRepository.id` is the repository url, so i don't think consumers can resolve the declared `(GitHubRepository, repo)` anchor back to the node. maybe keep `repo` for display and add `repo_id` for `asset_id_field`?

kunaals (changes requested):

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
```diff
@@ -3079,6 +3079,248 @@
...
+        { name: '@deliveroo/reevent', major: '1', version: '1.0.1' },
+        { name: '@or-sdk/invitations', major: '1', version: '1.4.9' },
+        { name: '@picsart/ai-sdk', major: '3', version: '3.32.2' },
+        { name: '@qlik/embed-runtime', major: '1', version: '1.6.4' }
+    ] AS at_risk
+    UNWIND at_risk AS a
+    MATCH (d:Dependency {ecosystem: 'npm', name: a.name})--(manifest:GitHubDependencyGraphManifest)--(r:GitHubRepository)
+    WHERE d.requirements IS NOT NULL
+      AND (d.requirements CONTAINS '^' OR d.requirements CONTAINS '~' OR d.requirements CONTAINS '>')
+      AND split(trim(replace(replace(replace(replace(d.requirements, '^', ''), '~', ''), '>', ''), '=', '')), '.')[0] = a.major
```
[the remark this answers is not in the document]
kunaals: +1 to this

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
EmilioDNA: Updated:
I added the r.id as repo_id. 
Also had to update the class definition by adding the repo_id since for previous waves this was not included:
class MaliciousNpmDependenciesShaiHuludOutput(Finding):
    repo: str | None = None
    **repo_id: str | None = None**
    name: str | None = None
    current_version: str | None = None
    vulnerable_version: str | None = None

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
EmilioDNA: Updated:
verified GitHubRepository with direction=INWARD to:
(r:GitHubRepository)-[:HAS_MANIFEST]->(manifest:GitHubDependencyGraphManifest)-[:HAS_DEP]->(d:Dependency {ecosystem: 'npm', name: v.name})

**NOTE: this rel labels are not defined previous waves of this rules, however, since I didn't author them, I kept the same. I could create an additional PR to fix those too.**

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
EmilioDNA: Updated:
verified GitHubRepository with direction=INWARD to:
(r:GitHubRepository)-[:HAS_MANIFEST]->(manifest:GitHubDependencyGraphManifest)-[:HAS_DEP]->(d:Dependency {ecosystem: 'npm', name: v.name})

**NOTE: this rel labels are not defined previous waves of this rules, however, since I didn't author them, I kept the same. I could create an additional PR to fix those too.**

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
EmilioDNA: Updated:
verified GitHubRepository with direction=INWARD to:
(r:GitHubRepository)-[:HAS_MANIFEST]->(manifest:GitHubDependencyGraphManifest)-[:HAS_DEP]->(d:Dependency {ecosystem: 'npm', name: v.name})

**NOTE: this rel labels are not defined previous waves of this rules, however, since I didn't author them, I kept the same. I could create an additional PR to fix those too.**

tests/unit/rules/test_malicious_npm_dependencies_shai_hulud.py
EmilioDNA: I decreased from 14 unit tests to 8, removing the most generic ones and kept the "more-specific-to-rules" ones:

The curated package/version lists themselves (test_aug_2026_fact_covers_chaindrop_entry_points, _keyv_scoped_family) — for a rule that is a hardcoded IOC list, drift in that list is the actual security regression.
Cross-Fact consistency: the pinned and at-risk Facts must cover the same packages (test_aug_2026_facts_cover_the_same_packages) and cypher_query/cypher_visual_query must stay in sync (test_aug_2026_queries_and_visual_queries_agree) — nothing generic diffs either of these, for any rule.
Security-relevant filters with no generic coverage: archived/disabled exclusion, and the disjointness between the pinned and at-risk Facts.
A direct regression guard for the rel-labels fix above (test_aug_2026_facts_use_explicit_relationship_labels), so this specific defect can't silently reappear.

If you consider that these unit tests are also covered by the standard rules framework, I can simply remove them.

cartography/rules/data/rules/malicious_npm_dependencies_shai_hulud.py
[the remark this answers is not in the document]
EmilioDNA: Valid, and worth being upfront about: the predicate only compares major-version components, so it both over-flags and under-flags. ^6.2.0 and >6.1.24 match a malicious 6.1.24 even though neither range includes it, and ^5.0.0 || ^6.1.0 is missed because split('.')[0] on the first clause reads 5.

I looked for a real fix and there isn't one at this layer — I checked both Cypher and APOC, and neither has semver-aware comparison or npm-style range intersection. Handling || unions, hyphen ranges, x-ranges, and the 0.x caret case properly needs an actual parser, which means a Python-side change at sync time: a new dependency plus logic in cartography/intel/github/repos.py computing a boolean on the Dependency node at ingestion. That's out of scope for a rules-data change.

For this PR I'd propose keeping the major-only heuristic as-is and scoping down what it claims — describing matches as candidates that need manual confirmation rather than confirmed hits. Happy to pick up the semver-aware comparison separately if that's a direction you'd take.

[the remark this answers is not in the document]
achantavy: Fair points, we gotta implement a semver library parser haha, or maybe one day implement an abstraction around this that "compiles" or "resolves down" to a big big list of hardcoded entries in the cypher query.

Anyway, not a huge blocker for me if it isn't a blocker for anyone else, but lets do make sure to make this as correct and useful as possible.

achantavy (approved):

kunaals (approved):

PR #3140, opened 2026-08-10
feat(zizmor): ingest zizmor GitHub Actions audit reports

jychp:
### Type of change
- [ ] Documentation update

### Summary

Cartography's GitHub module records what a workflow is *allowed* to do: its triggers, its `permissions` block, the secrets it references, whether the actions it calls are pinned. It records nothing about what is actually *wrong* with it.

[zizmor](https://github.com/zizmorcore/zizmor) is a static analyzer for GitHub Actions that reports CI supply-chain weaknesses: template injection, unpinned action references, excessive permissions, impostor commits, and 37 other audits. This PR adds a module that ingests its reports so those findings can be joined to the workflows, actions, and repositories already in the graph.

Like `trivy`, `syft`, and `aibom`, the module **does not run the binary**. It reads pre-generated reports through the shared `ReportSource` abstraction, so local paths, `s3://`, `gs://`, and `azblob://` all work.

**Graph model.** One new node type, `ZizmorFinding`, carrying the ontology label `SecurityIssue`:

```
(:ZizmorFinding)-[:AFFECTS]->(:GitHubWorkflow)     matched on {repo_url, path}
(:ZizmorFinding)-[:AFFECTS]->(:GitHubAction)       only for findings located on a `uses` key
(:ZizmorFinding)-[:FOUND_IN]->(:GitHubRepository)  matched on {id: repository URL}
```

This is what makes the data worth having; the two sides compose:

```cypher
MATCH (f:ZizmorFinding)-[:AFFECTS]->(w:GitHubWorkflow)<-[:HAS_WORKFLOW]-(r:GitHubRepository)
WHERE f.severity = 'high' AND w.permissions_contents = 'write'
RETURN r.name, w.path, f.audit_id
```

The action edge is resolved **structurally**, from a location whose YAML route ends at a `uses` key, rather than from a hardcoded list of audit identifiers. That covers `unpinned-uses`, `impostor-commit`, `known-vulnerable-actions`, `stale-action-refs`, `typosquat-uses` and the rest without tracking zizmor's audit registry as it evolves.

**Why a mapping file.** zizmor's JSON carries no repository identity: for a local run, the only path information is the literal command-line argument (`{"Local": {"verbatim_path": "./.github/workflows/ci.yml"}}`). `--zizmor-source` therefore points at a YAML file declaring which repository each report belongs to, following the precedent set by `--semgrep-oss-source`.

```yaml
repositories:
  - owner: "cartography-cncf"
    repo: "cartography"
    url: "https://github.com/cartography-cncf/cartography"
    branch: "master"
    reports:
      - "s3://security-artifacts/zizmor/cartography/master.json"
```

### Related issues or links

- [zizmor](https://github.com/zizmorcore/zizmor) and its [audit reference](https://docs.zizmor.sh/audits/)

### How was this tested?

**48 unit tests + 8 integration tests**, plus an end-to-end run against this repository's own workflows.

Generated a real report and ingested it:

```
zizmor --format=json-v1 --no-exit-codes --offline --persona=auditor . > zizmor-report.json
cartography --selected-modules zizmor --zizmor-source zizmor-mapping.yaml
```

```
INFO:cartography.sync:Starting sync stage 'zizmor'
INFO:cartography.intel.zizmor:Processing Zizmor repository mapping for cartography-cncf/cartography.
INFO:cartography.intel.zizmor:Zizmor repository cartography-cncf/cartography processed 1/1 report sources successfully.
INFO:cartography.intel.zizmor:Transformed 27 zizmor findings for repository cartography-cncf/cartography.
INFO:cartography.intel.zizmor:Loading 27 ZizmorFinding objects into the graph.
INFO:cartography.client.core.tx:Loaded 27 ZizmorFinding nodes
INFO:cartography.intel.zizmor:Running ZizmorFinding cleanup job for repository https://github.com/cartography-cncf/cartography.
INFO:cartography.sync:Finishing sync stage 'zizmor'
```

Resulting graph, against `GitHubWorkflow`/`GitHubAction` nodes built from this repo's five real workflow files:

| audit | severity | persona | nodes |
|---|---|---|---|
| undocumented-permissions | low | pedantic | 9 |
| concurrency-limits | low | pedantic | 5 |
| anonymous-definition | informational | pedantic | 4 |
| excessive-permissions | high / medium | pedantic | 2 / 2 |
| cache-poisoning | high | regular | 2 |
| secrets-outside-env | medium | auditor | 1 |
| template-injection | low | pedantic | 1 |
| use-trusted-publishing | informational | regular | 1 |

27 `AFFECTS`→`GitHubWorkflow` across all five files with **zero orphans**, 27 `FOUND_IN`→`GitHubRepository`, and 2 `AFFECTS`→`GitHubAction` on `astral-sh/setup-uv`. The only two findings positioned on a `uses:` key are the only two carrying an action edge, which is what the structural rule is supposed to produce.

Three behaviours were then exercised against that live graph rather than only against fixtures:

- **Cleanup.** Re-ingested a report truncated to 2 findings; the 25 stale ones were removed with their edges, the 2 survivors kept.
- **Stale relationships.** Re-ingested with `astral-sh/setup-uv`'s pinned SHA bumped at the same YAML route. The finding node survives by design, and its `AFFECTS` edge moved to the new action with **0 edges left on the old one**.
- **Corrupt input.** Appended a malformed entry to the report. It was rejected outright, cleanup was skipped, and all 27 existing findings survived.

Every non-obvious behaviour has a test that was **verified to fail when its fix is neutralised** (multi-primary handling, relationship cleanup, skip-blocks-cleanup, duplicate-URL rejection), so the tests are pinning behaviour rather than just passing.

Full suites: `make test_lint` clean, `tests/unit` 4631 passed, `tests/integration` 1410 passed. Sphinx builds and generates `docs/root/modules/zizmor/schema.md` from the data model.

#### General

#### Proof of functionality

- [x] Updated the [schema README](https://github.com/cartography-cncf/cartography/blob/master/docs/schema/README.md) (`docs/root/usage/schema.md` and `docs/root/module-list.md`).

### Notes for reviewers

Three decisions worth your attention, and one limitation I chose rather than hid.

**1. Cleanup uses hand-written Cypher instead of `GraphJob.from_node_schema()`.** `ZizmorFinding` has no tenant-like sub-resource, so a generated cleanup job would be unscoped and delete every stale finding globally. That would force skipping cleanup entirely whenever a single report in the mapping file is unreadable, leaving stale findings behind for every other repository. Scoping on `repository_url` keeps a failure contained to the repository it affects, following `cleanup_oss_semgrep_sast_findings`. This is a deliberate deviation from the declarative-cleanup guidance and is justified in the function's docstring. Happy to switch to the all-or-nothing declarative form if you'd rather not carry the exception.

**2. Relationships are cleaned separately from nodes.** A finding's id hashes its audit, repository, file path, and YAML route, but *not* its `uses` reference. Bumping the action a step calls keeps all four, so the node survives the sync while its `AFFECTS` edge must move to the new action. Deleting only stale nodes would leave the edge to the old action attached.

**3. Cleanup is withheld whenever anything was skipped.** A finding that is well-formed but unjoinable (read from stdin, so it has no path) is still an open finding. Cleaning up after skipping it would delete it as though it had been fixed. The trade-off is visible staleness instead of silent deletion: someone who habitually pipes stdin to zizmor will see that repository never cleaned, with only a `WARNING` to say so. I took that over silent data loss, but it is a judgement call.

**Limitation: one branch per repository.** The mapping file rejects duplicate repository URLs, because `branch` is not part of a finding's id, so two entries for one repository collide on the same nodes and let a successful entry authorize cleanup that a failed entry should have withheld. Supporting a repository audited on several branches means threading `branch` through the id and the cleanup scope; that is a larger change and I left it out deliberately. The error message says so explicitly rather than corrupting data quietly.

**Not in this PR, in case you want them tracked:** SARIF input support, and security rules under `cartography/rules/` built on `ZizmorFinding`.


cartography/intel/zizmor/transform.py
```diff
@@ -0,0 +1,383 @@
...
+            continue
+
+        concrete = location.get("concrete")
+        if not isinstance(concrete, dict):
+            continue
+        feature = concrete.get("feature")
+        if not isinstance(feature, str):
+            continue
+
+        candidate = feature.strip()
```
kunaals, line 261: for `ref-version-mismatch`, the first `uses` location can contain only a value like `v3`, while a related location contains the complete `actions/checkout@...` reference. this currently looks for an action named `owner:v3` and loses the `AFFECTS` edge. please prefer a complete `uses` value and add a real-output fixture.

cartography/intel/zizmor/transform.py
```diff
@@ -0,0 +1,383 @@
...
+
+def _visible_locations(finding: dict[str, Any]) -> list[dict[str, Any]]:
+    """
+    Return the finding's locations, dropping zizmor's internal `Hidden` ones.
+    """
+    locations = finding.get("locations")
+    if not isinstance(locations, list):
+        return []
+
+    visible: list[dict[str, Any]] = []
```
kunaals: malformed entries inside `locations` are silently dropped, but the report still authorizes cleanup. a partially corrupt report can therefore delete findings from the previous sync. please reject the report or block cleanup whenever any location is malformed.

docs/root/modules/zizmor/config.md
```diff
@@ -0,0 +1,112 @@
...
+| Field | Required | Description |
+|-------|----------|-------------|
+| `repositories` | Yes | Non-empty list of repository entries. |
+| `owner` | Yes | GitHub organization or user that owns the repository. |
+| `repo` | Yes | Repository name. |
+| `url` | Yes | Repository URL. Must match `GitHubRepository.id`. |
+| `branch` | Yes | Branch the scanned workflow files were read from. |
+| `reports` | Yes | Non-empty list of report locators. Each must resolve to exactly one JSON artifact. |
+
+A repository may only appear once. List all of its reports under that single
```
kunaals: this says reports from different branches can share one mapping, but the mapping has only one branch and finding ids don’t include it. findings from the same location on `main` and `release` will collide. please require every report in an entry to describe the same branch.

docs/root/modules/zizmor/config.md
```diff
@@ -0,0 +1,112 @@
...
+## Input Artifacts
+
+Cartography ingests pre-generated zizmor JSON reports. It does not run the
+zizmor binary.
+
+### Generate Input Artifacts
+
+Produce reports with the versioned JSON format:
+
+```bash
```
kunaals, line 43: this command needs `--no-ignores` if the report is meant to include suppressed findings. without it, zizmor omits most ignored findings entirely. the description of `ignored` should also be limited to the ignore types that zizmor actually marks in its json output.

cartography/intel/zizmor/transform.py
```diff
@@ -0,0 +1,383 @@
...
+    local = key.get("Local")
+    if not isinstance(local, dict):
+        # Stdin, or a future key variant we do not understand.
+        return None
+
+    verbatim_path = local.get("verbatim_path")
+    if not isinstance(verbatim_path, str) or not verbatim_path.strip():
+        return None
+
+    path = verbatim_path.strip()
```
kunaals: this path normalization still leaves standalone absolute paths machine-specific, and it chooses the wrong `.github/` segment when the repository itself is named `.github`. please use the repository root when it’s available, or reject absolute paths that can’t be normalized safely.

kunaals (changes requested):

kunaals (approved):

PR #3143, opened 2026-08-11
feat: compute internet exposure for Scaleway, Railway, Modal, Netlify, Supabase, Cloudflare and Vercel

jychp:
### Type of change

- [x] Documentation update

### Summary

Cartography computes internet exposure consistently for AWS, GCP, Azure and Kubernetes: a
node carries `exposed_internet` / `exposed_internet_type`, and `EXPOSE` edges run from an
internet-facing frontend to the asset it puts at risk. None of the providers added in 2026
followed that contract. Each one either exposed raw provider flags and left every rule to
re-derive reachability in Cypher, or exposed nothing at all.

The concrete consequence was silent under-reporting. The ontology
`WORKLOAD_HAS_RUNTIME_IMAGE` job ORs `exposed_internet` across a workload and its replicas
to denormalize the verdict onto the `HAS_RUNTIME_IMAGE` edge, and its own comment claims
Scaleway serverless containers are in scope, but nothing ever wrote the property. Any
"internet-exposed workload running a vulnerable image" query therefore missed these
providers without saying so.

This PR writes down the contract, fixes three correctness problems in it, and implements
exposure for Scaleway, Railway, Modal, Netlify, Supabase, Cloudflare and Vercel.

**Fixing the contract first.** `EXPOSE` carried two contradictory directions:
AWS/GCP/Azure/Kubernetes point from the load balancer to the workload, while Railway pointed
from the workload to its domain. No cross-provider traversal of the edge was meaningful. The
majority direction wins, Railway and Modal are flipped, Modal's `EXPOSES` is renamed, and
`EXPOSE` becomes a `RelConstraint` so the existing CI guard prevents future drift. The
contract itself is now written in `docs/root/usage/tutorial.md` instead of being folklore
spread across four analysis modules.

**Preferring the standard mechanisms.** The contract states that computing exposure in
`transform()` is preferred when the signal is local to one API response, as
`cartography.intel.aws.apigateway` and `cartography.intel.aws.elasticsearch` already do. An
early draft of this work reached for an analysis job everywhere; one commit
(`refactor(scaleway)`) deletes most of it, because those jobs were re-deriving in Cypher what
the transforms had already computed in Python. Exactly two analysis jobs are added, and only
because they genuinely span node types:

- `SCALEWAY_INSTANCE_EXPOSURE` reads security-group rules, Public Gateway PAT rules and load
  balancer backends, which are three separate syncs.
- `SCALEWAY_LB_EXPOSE_EDGES` resolves `ScalewayLBBackend.pool`, which holds IP addresses
  rather than server ids, against instances. A MatchLink was considered first per
  relationship > MatchLink > analysis job, but its matcher cannot be scoped to the project,
  and private IPs are reusable across projects, so it would over-match.

**Erring towards reporting exposure.** Where the API is ambiguous, the verdict leans towards
"exposed" rather than hiding an attack surface. Scaleway object storage keeps no verdict at
all when neither the bucket ACL nor its policy could be read, so "could not tell" stays
distinct from "not exposed". A Netlify access gate only counts when its context is `all`,
since a password on deploy previews leaves production open. An unrecognised Vercel
`deploymentType` counts as not covering the deployment.

**Closing the contract's own gaps.** With every provider on the same convention, two
inconsistencies in it were worth fixing rather than documenting:

- Thirteen labels received `exposed_internet` and/or `exposed_internet_type` from an analysis
  job without declaring them on the node schema, so they were absent from the generated
  schema docs and unindexed.
- Five resources ended up with a verdict no rule reported. Cloudflare R2 and Supabase Storage
  buckets were already mapped to the ObjectStorage ontology's `_ont_public`, so the data was
  in the graph and simply unqueried; Scaleway Data Warehouse, Serverless SQL and Search
  gained their verdict earlier in this branch. All five now have a `Fact`.

**Three bugs found and fixed along the way**, all in the dangerous direction:

- Vercel `trustedSources` was read as an IP allowlist. It is not: it lets already-protected
  deployments accept short-lived OIDC tokens instead of long-lived secrets. Treating it as a
  gate marked genuinely public projects `exposed_internet = false` and dropped them from the
  inventory. The real IP allowlist is the separate `trustedIps` field.
- Railway's exposure was correct but stored under a provider-specific name, so nothing
  cross-provider, including the ontology roll-up, could see it.
- `EXPOSE` was untraversable across providers, as above.

### Related issues or links

- Vercel Trusted Sources: https://vercel.com/changelog/trusted-sources-for-deployment-protection

### Breaking changes

**Two `EXPOSE` edges reverse direction, and one is renamed.** Both endpoints shipped recently
(Railway 2026-07-29, Modal 2026-08-01), so the blast radius is small, and the next sync
recreates the edge while cleanup removes the old one. Any query traversing
`(:RailwayServiceInstance)-[:EXPOSE]->` or `(:ModalSandbox)-[:EXPOSES]->` must be inverted.

`RailwayServiceInstance.is_publicly_exposed` is **not** removed. It is kept, marked
deprecated for v1.0.0, and stays in step with the new `exposed_internet`.

Nothing else changes for a provider that already computed exposure. Adding the explicit
`False` pass to AWS and Kubernetes was part of an earlier revision of this branch and has been
dropped: it altered stored data on the two most-used providers for a convenience this PR does
not need, and it deserves its own PR where that is the subject rather than a side effect.

### How was this tested?

- `make test_lint` passes.
- Full suite green locally: **6064 tests** (`pytest tests/unit tests/integration`).
- Verdicts are asserted in the per-resource sync tests, where `transform()` actually runs,
  rather than against a hand-seeded graph.
- New `tests/integration/cartography/intel/scaleway/test_exposure.py` covers the two analysis
  jobs: seven instances spanning direct, PAT, outbound-only, `drop`, scoped-CIDR, no-public-IP
  and stopped; a load balancer that is public, one that is private and one with no frontend;
  and a backend pool that no frontend routes to, which must yield no `EXPOSE` edge.
- An end-to-end test asserts a Scaleway serverless container's verdict reaches
  `HAS_RUNTIME_IMAGE.exposed_internet`, with nothing hand-set. This is the regression the PR
  exists to close, and it complements `test_has_runtime_image_self_service_workload`, which
  injects the property directly.
- New unit tests pin the ambiguous cases: Netlify context-scoped gates, and Vercel's
  `trustedSources` not hiding a public deployment.
- The Cypher of `object_storage_public` and `database_instance_exposed` had never been executed
  by a test, for any provider. New tests under `tests/integration/rules/` seed a graph and run
  each of the five new facts, including an r2.dev-only bucket and a bucket with neither flag
  set so the `coalesce` in the R2 query is covered.
- Verified via `build_data_model` that every new property is introspected with **0
  diagnostics**, so they all render in the generated schema docs.

No real-environment sync logs: I have no Scaleway, Railway, Modal, Netlify, Supabase or
Vercel tenant to run against. Every verdict is derived from data these modules already
ingest, so no new API call or permission is involved.

#### General

#### Proof of functionality

- [ ] Included Cartography sync logs from a real environment. **Not provided**, see the
  testing note above: no tenant available for these providers, and no new entity or API call
  is introduced.

- [x] Updated the schema documentation. The per-module `schema.md` files are build artifacts
  generated from the node schemas by `inspect_data_model()`, and every new property carries a
  description, so they regenerate automatically. The hand-written exposure contract lives in
  `docs/root/usage/tutorial.md`.
- [ ] Updated the schema README. Not applicable: no new node label or module is added.

- [x] Used the NodeSchema data model. No new intel module here; the changes extend existing
  ones.

### Notes for reviewers

Worth a close look:

- **The `EXPOSE` direction decision** (`f29cbeea7`, `3c9c14720`, `9389bb5e6`). Flipping
  Railway and Modal rather than AWS is the call I would most want challenged.
- **`refactor(scaleway)`** (`9a441c729`) deletes 313 lines of analysis-job code added earlier
  in the same branch. Reviewing that commit alone shows which mechanism each signal ended up
  using and why.
- **`SCALEWAY_LB_EXPOSE_EDGES`** uses an `EXISTS { }` subquery and a project join. The join is
  load-bearing, not decorative: without it, reusable private IPs match instances in other
  projects.

Deliberately left out, each with a reason in the code:

- `exposed_internet_type` is still undeclared on `AWSLoadBalancer` and `AWSLoadBalancerV2`.
  Their jobs call `SetProperty("elb", "exposed_internet_type", None)`, so the value is always
  null. That looks like a remnant keeping the generated cleanup removing a stale type from
  older graphs, so it is untouched, and declaring an always-null property would mislead.
- **Decoupling the Kubernetes exposure job from `AWSLoadBalancerV2` is blocked by missing
  data, not by its Cypher.** `KubernetesNode` ingests no address at all (only `provider_id`
  and `instance_id`), so NodePort and hostNetwork exposure are not computable;
  `USES_LOAD_BALANCER` targets only `AWSLoadBalancerV2` for both services and ingresses, so
  no non-AWS load balancer is ever linked and switching to the `LoadBalancer` ontology label
  would be cosmetic; and keying off `_ont_scheme = 'internet_facing'` instead would report
  internet-facing ALBs that no security group actually opens. There is an avenue via
  `(:KubernetesNode)-[:IS_INSTANCE]->(:AWSEC2Instance)`, but `instance_id` is null outside
  AWS, so it trades one AWS coupling for another and would need port-level reasoning to avoid
  false positives.
- `ScalewayKapsuleCluster` gets no verdict. `clusters.py:236` already explains that the
  apiserver is always internet-reachable and that the IP allowlist config is not ingested, so
  a verdict would be misleading until those ACLs are collected.
- `NetlifyFunction` is untouched. Its endpoint is public, but Netlify's site-level gates cover
  functions too, and that is a join to the parent site rather than a local signal.
- Vercel `trustedIps` is not consulted. It is a real gate, but it is absent from the
  documented response schema of the projects listing this module reads, so a project relying
  on it alone is reported as exposed. Confirming what `/v9/projects` actually returns needs a
  real account.
- No `Module` enum entries for Vercel, Netlify or Modal: nothing here defines a rule for them,
  and an enum value no rule uses would be an orphan. `Module.SUPABASE` is added because the
  Supabase Storage rule lands with it.


cartography/analysis/scaleway/analysis.py
```diff
@@ -0,0 +1,159 @@
...
+    statements=(
+        AnalysisStatement(
+            comment=(
+                "A public IP plus an inbound accept for 0.0.0.0/0. Both are required, as in "
+                "AWS_EC2_ASSET_EXPOSURE_INSTANCE: an open group on an instance with no public "
+                "address is not reachable. No port filter, since exposed_internet means "
+                "reachable at all. TODO: ::/0 is not matched, only 0.0.0.0/0."
+            ),
+            match="""
+            MATCH (instance:ScalewayInstance)-[:MEMBER_OF_SCALEWAY_SECURITY_GROUP]->(:ScalewaySecurityGroup)<-[:MEMBER_OF_SCALEWAY_SECURITY_GROUP]-(rule:ScalewaySecurityGroupRule)
```
kunaals, line 74: could this also account for `sg.inbound_default_policy = 'accept'`? scaleway treats that as allowing unmatched inbound traffic, and we already ingest the field. right now a public-ip instance with an accept default and no explicit rule falls through to the final false pass, so we'd persist a reachable instance as not exposed. could we cover that case and add a regression test with no explicit rule?

kunaals (approved):

PR #3144, opened 2026-08-11
fix(cve_metadata): bound Neo4j transaction memory for the ENRICHES fan-out

jychp:
### Type of change

- [ ] Documentation update

### Summary

The `cve_metadata` sync fails reproducibly on large graphs with
`Neo.TransientError.General.MemoryPoolOutOfMemoryError` ("the allocation of an extra 2.0 MiB would
use more than the limit 3.5 GiB ... `dbms.memory.transaction.total.max` threshold reached"), and then
burns roughly an hour retrying before it gives up. On the observed run, 22144 CVE nodes were enriched
year by year: the 3016-row bucket took about six minutes to commit, and the following 2400-row bucket
raised after a full hour.

There are three causes.

**1. No per-transaction row bound.** `load_cve_metadata` called `load()` with no `batch_size`, so it
used the default of 10000. The only upstream batching is `_build_yearly_cve_batches`, which buckets
CVE ids by publication year, and every bucket lands below 10000. The chunker in `load_graph_data`
therefore never split anything: **one CVE year was one Neo4j transaction**. The existing "keep memory
bounded" comment in the sync loop is about the Python heap, not Neo4j transaction memory.

**2. The `ENRICHES` edge fans out per row.** `CVEMetadataToCVERel` matches `target_node_label="CVE"`
on `{"cve_id": PropertyRef("id")}`, deliberately loose so it reaches every node carrying the `:CVE`
semantic label (the deprecated CVE node plus `AWSInspectorFinding`, `SemgrepSCAFinding`,
`TrivyImageFinding`, and so on). Each input row expands to one relationship MERGE per match, and that
count is unbounded and graph-dependent. Rows times fan-out is what exhausts the pool. The fan-out is
intentional and covered by a regression test, so batch size is the only lever available here.

**3. The oversized payload was replayed, not shrunk.** `MemoryPoolOutOfMemoryError` maps to
`neo4j.exceptions.TransientError`, which carries `_retryable = True`. It was therefore in
`_NETWORK_EXCEPTIONS` and `_run_with_retry` replayed it up to five times with exponential backoff,
re-sending the same `DictList` each time. The neo4j driver's own managed-transaction retry
(`Session._run_transaction`) adds another layer: it catches `(DriverError, Neo4jError)` and retries
anything whose `is_retryable()` is true, and its retry timer starts only *after* the first attempt, so
a slow transaction is always replayed in full at least once. Replaying a batch that is too large for
the pool can never succeed. That is the wasted hour.

#### What this PR changes

**`cartography/intel/cve_metadata/__init__.py`**: adds `CVE_METADATA_BATCH_SIZE = 500`, passed to
`load()`, following the existing convention of a named per-module constant with a why-comment
(`ECR_LAYER_BATCH_SIZE`, `CONTAINER_IMAGE_BATCH_SIZE`). Sizing: about 2400 rows exhausted a 3.5 GiB
pool, roughly 1.5 MB of transaction memory per row on that graph, so 500 leaves about 3x headroom.

**`cartography/client/core/tx.py`**: treats transaction-memory exhaustion as a signal to shrink
rather than to back off. This is deliberately at the shared layer: the retry classification is
generic, not specific to this module.

- `_is_transaction_memory_error` classifies the two Neo4j memory error codes.
- A new `Neo4jTransactionMemoryError`, deliberately **not** a `neo4j.exceptions.Neo4jError` subclass.
  `write_list_of_dicts_tx` converts at the transaction-function boundary, so the error is invisible to
  the driver's managed-transaction retry and an oversized batch is no longer replayed in full before
  our code sees it.
- `_run_with_retry` re-raises immediately instead of consuming five network retries. This is the
  backstop for commit-time errors, which are raised from `tx._commit()` outside the transaction
  function.
- `load_graph_data` halves the batch and retries the same rows. The reduced size then persists for the
  rest of the load, so the cost of discovering it is paid once rather than per chunk. Below a floor of
  10 rows it fails with an actionable message, because at single-digit batch sizes the per-row fan-out
  dominates and shrinking further just converts one failure into thousands of slow transactions.

Shrinking is safe because a failed transaction commits nothing and the generated queries are
idempotent MERGEs. `load_matchlinks` inherits the behavior for free, since it delegates to
`load_graph_data`.

#### Note on blast radius

Cause 3 is a shared-layer fix, so it changes behavior for every module that writes through `load()`,
`load_matchlinks()`, or `load_graph_data()`. The one realistic downside is a genuinely transient pool
exhaustion caused by a concurrent sync, which today sometimes succeeds on a plain retry and would now
shrink instead. Shrinking still completes the load, so this should be strictly better, but it is worth
a reviewer's attention.

`load_matchlinks_cartesian_product` keeps its own two-dimensional batching and is not converted to
adaptive shrinking. It still benefits from the classification change: an out-of-memory error there now
fails in seconds instead of after five replays.

This bounds rows, not fan-out. If a single `cve_id` matches a very large number of `:CVE`-labelled
findings, even 500 rows can exhaust the pool; the adaptive shrink absorbs that. A durable follow-up
would be to move `CVEMetadataToCVERel` out of the node ingestion query into its own `load_matchlinks`
with an independent batch size. That is out of scope here.

### How was this tested?

- `make test_lint` passes.
- `uv run --frozen pytest tests/unit`: 4594 passed.
- `uv run --frozen pytest tests/integration`: 1402 passed, against a local Neo4j. The shared write
  path is on every module's load path, so the full suites were run rather than just the touched
  modules.
- **End-to-end against a real Neo4j, using the driver's real error rather than a synthetic one.**
  `db.memory.transaction.max` was temporarily capped at 2 MiB and a 2000-row load was driven through
  `load_graph_data`. This confirmed the error code is
  `Neo.TransientError.General.MemoryPoolOutOfMemoryError`, that it surfaces at execution time (so the
  driver never replays it), that the batch size walked 2000 → 1000 → 500 → 250 → 125 → 62 → 31, and
  that **all 2000 rows were written** with no loss. The setting was restored afterward.

  ```
  WARNING cartography.client.core.tx Neo4j transaction memory exceeded. Halving batch size to 1000
  and retrying from row 0 of 2000. Consider setting an explicit batch_size constant for this module.
  Error: Neo4j ran out of transaction memory. This is not retryable: replaying the same payload will
  fail again. Lower the batch_size passed to load()/load_matchlinks() for this module, or raise the
  transaction memory limit named in the original error below. Original error:
  {neo4j_code: Neo.TransientError.General.MemoryPoolOutOfMemoryError} {message: The allocation of an
  extra 4.9 MiB would use more than the limit 2.0 MiB. Currently using 2.0 MiB.
  db.memory.transaction.max threshold reached} {gql_status: 51N72}
  ...
  >>> rows written: 2000 (expected 2000)
  ```

New unit tests:

- `tests/unit/cartography/client/core/test_tx.py`: error classification, no retry on memory errors,
  other `TransientError` codes still retried, conversion in `write_list_of_dicts_tx` (asserting the
  raised type is outside the `Neo4jError` hierarchy, which is the regression guard for the
  driver-retry bypass), batch halving, sticky reduced size, the floor, and no shrinking for unrelated
  errors.
- `tests/unit/cartography/intel/cve_metadata/test_init.py`: `load_cve_metadata` passes a bounded
  `batch_size`.

#### General

#### Proof of functionality

### Notes for reviewers

The two parts are separable if you would prefer them split: the `cve_metadata` batch size alone
resolves the reported failure, and the `tx.py` change is the generic hardening that keeps any module
with heavy per-row fan-out from burning an hour on a payload that can never succeed. Happy to split
them into two PRs if you would rather review them independently.

Two design choices worth a second opinion:

- **The reduced batch size is sticky for the rest of the load** rather than being re-derived per
  chunk. A stack-of-halves design rediscovers the failure once per chunk, which with the 10000-row
  default over a large list means many wasted attempts. It never re-grows, which is the conservative
  choice.
- **The floor is 10, not 1.** Below that the per-row fan-out dominates, so failing with a message that
  points at the fan-out is more actionable than crawling through thousands of single-row transactions.

### DIFF
--- a/cartography/client/core/tx.py
+++ b/cartography/client/core/tx.py
@@ -40,6 +40,47 @@ _NETWORK_EXCEPTIONS: tuple[type[BaseException], ...] = (
     neo4j.exceptions.TransientError,
 )
 
+# Neo4j reports transaction-memory exhaustion as a retryable TransientError, but replaying an
+# identical payload can never succeed: the batch itself is too large for the pool. Handle these
+# by shrinking the batch (see load_graph_data) rather than by backing off.
+_TRANSACTION_MEMORY_ERROR_CODES = frozenset(
+    {
+        "Neo.TransientError.General.MemoryPoolOutOfMemoryError",
+        "Neo.TransientError.General.OutOfMemoryError",
+    }
+)
+# Below this, halving no longer buys meaningful memory relief: at single-digit batch sizes the
+# per-row fan-out dominates, so shrinking further just converts one failure into thousands of
+# slow transactions. Fail with an actionable message instead.
+_MIN_ADAPTIVE_BATCH_SIZE = 10
+
+
+class Neo4jTransactionMemoryError(Exception):
+    """
+    A write transaction exceeded Neo4j's transaction memory.
+
+    Deliberately NOT a ``neo4j.exceptions.Neo4jError`` subclass. The driver's own managed
+    transaction retry (``Session._run_transaction``) catches ``(DriverError, Neo4jError)`` and
+    replays anything whose ``is_retryable()`` is True, and ``MemoryPoolOutOfMemoryError``
+    reports True. Raising a type the driver does not recognise takes this error out of that
+    retry loop as well as out of ours.
+    """
+
+
+def _is_transaction_memory_error(exc: BaseException) -> bool:
+    """
+    Determine if an exception is Neo4j running out of transaction memory.
+
+    The driver marks every TransientError retryable, so ``_NETWORK_EXCEPTIONS`` would
+    otherwise catch this wholesale and replay the same oversized batch.
+
+    :param exc: The exception to check
+    :return: True if Neo4j exhausted its transaction memory pool or heap, False otherwise
+    """
+    if not isinstance(exc, neo4j.exceptions.Neo4jError):
+        return False
+    return exc.code in _TRANSACTION_MEMORY_ERROR_CODES
+
 
 def _is_retryable_client_error(exc: Exception) -> bool:
     """
@@ -194,6 +235,12 @@ def _run_with_retry(operation: Callable[[], T], target: str) -> T:
                 )
             return result
         except _NETWORK_EXCEPTIONS as exc:
+            if _is_transaction_memory_error(exc):
+                # Same payload, same result. Let the caller shrink the batch instead.
+                # write_list_of_dicts_tx converts execution-time memory errors before the
+                # driver ever sees them; this branch is the backstop for commit-time ones,
+                # which are raised from tx._commit() outside the transaction function.
+                raise Neo4jTransactionMemoryError(str(exc)) from exc
             if network_attempts >= _MAX_NETWORK_RETRIES - 1:
                 raise
             network_attempts += 1
@@ -612,8 +659,18 @@ def write_list_of_dicts_tx(
     Note:
         This function is typically used internally by higher-level functions like
         ``load_graph_data()`` rather than being called directly by user code.
+
+        Transaction-memory exhaustion is re-raised as ``Neo4jTransactionMemoryError``. That
+        type is invisible to the driver's managed-transaction retry, so an oversized batch
+        fails immediately instead of being replayed in full; ``load_graph_data()`` then
+        shrinks the batch and retries.
     """
-    tx.run(query, kwargs).consume()
+    try:
+        tx.run(query, kwargs).consume()
+    except neo4j.exceptions.Neo4jError as exc:
+        if _is_transaction_memory_error(exc):
+            raise Neo4jTransactionMemoryError(str(exc)) from exc
+        raise
 
 
 def write_matchlink_cartesian_product_tx(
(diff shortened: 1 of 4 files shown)


cartography/client/core/tx.py
```diff
@@ -40,6 +40,47 @@
     neo4j.exceptions.TransientError,
 )
 
+# Neo4j reports transaction-memory exhaustion as a retryable TransientError, but replaying an
+# identical payload can never succeed: the batch itself is too large for the pool. Handle these
+# by shrinking the batch (see load_graph_data) rather than by backing off.
+_TRANSACTION_MEMORY_ERROR_CODES = frozenset(
+    {
+        "Neo.TransientError.General.MemoryPoolOutOfMemoryError",
```
kunaals, line 48: i think this needs to stay scoped to the cve loader. the production limit here, `dbms.memory.transaction.total.max`, is shared by all running transactions, so the same batch can succeed once concurrent work drains. making this a global no-retry rule turns a valid transient recovery into a new failure mode for every loader.

cartography/client/core/tx.py
```diff
@@ -40,6 +40,47 @@
...
+        "Neo.TransientError.General.OutOfMemoryError",
```
kunaals, line 49: this is a different failure mode from the transaction pool error. neo4j documents `General.OutOfMemoryError` as jvm heap exhaustion, so shrinking every loader batch here can mask a capacity problem rather than fix it. can we leave this on the existing error path?

cartography/client/core/tx.py
```diff
@@ -194,6 +235,12 @@ def _run_with_retry(operation: Callable[[], T], target: str) -> T:
                 )
             return result
         except _NETWORK_EXCEPTIONS as exc:
+            if _is_transaction_memory_error(exc):
```
kunaals, line 238: this only bypasses driver retries for errors raised inside `write_list_of_dicts_tx`. a retryable error from `tx._commit()` is retried by `Session.execute_write()` before this wrapper sees it, so the commit-time backstop described here does not actually fail fast. could we either cover that path with a managed-transaction test or narrow this claim?

kunaals (changes requested):

cartography/client/core/tx.py
```diff
@@ -679,23 +746,58 @@ def load_graph_data(
...
+        try:
+            execute_write_with_retry(
+                neo4j_session,
+                write_list_of_dicts_tx,
+                query,
+                DictList=data_batch,
+                **kwargs,
+            )
+        except Neo4jTransactionMemoryError as exc:
+            if effective_batch_size <= _MIN_ADAPTIVE_BATCH_SIZE:
```
kunaals, line 779: why stop at 10 rows? 10 can still exceed the limit while 5 or 1 would succeed. with this floor, an adaptive load can fail even though a smaller batch would complete. i think we should shrink to 1 before giving up, or make the floor caller-owned.

cartography/client/core/tx.py
```diff
@@ -679,23 +746,58 @@ def load_graph_data(
...
+                raise Neo4jTransactionMemoryError(
+                    f"Neo4j ran out of transaction memory even at batch_size="
+                    f"{effective_batch_size}. Each input row likely fans out to many matched "
+                    f"nodes; reduce the fan-out of this query or raise the transaction "
+                    f"memory limit named in the causing error."
+                ) from exc
+            effective_batch_size = max(
```
kunaals, line 786: this halves the configured batch size rather than the slice that actually failed. for example, a 2,400-row input with the 10,000 default is retried unchanged at 5,000 and 2,500 before it really gets smaller. can we derive the next size from `len(data_batch)` instead?

PR #3146, opened 2026-08-11
feat(gcp): add SSL policy and target proxy support

ashton-suire:
```md
   ### Type of change
   - [x] Documentation update


   ### Summary

   Adds GCP Compute SSL policy support so Cartography can analyze TLS posture for GCP HTTPS/SSL load balancers.

   This PR adds ingestion for:

   - `GCPSslPolicy`
   - `GCPTargetHttpsProxy`
   - `GCPTargetSslProxy`

   It also adds relationships that allow users to traverse from forwarding rules to target proxies and then to attached SSL policies:

   - `(:GCPForwardingRule)-[:ROUTES_TO]->(:GCPTargetHttpsProxy)`
   - `(:GCPForwardingRule)-[:ROUTES_TO]->(:GCPTargetSslProxy)`
   - `(:GCPTargetHttpsProxy)-[:USES]->(:GCPSslPolicy)`
   - `(:GCPTargetSslProxy)-[:USES]->(:GCPSslPolicy)`
   - `(:GCPTargetSslProxy)-[:ROUTES_TO]->(:GCPBackendService)`

   The implementation supports global SSL policies and target HTTPS proxies, regional SSL policies and regional target HTTPS proxies, and global target SSL proxies.

   This enables users to identify load balancers using GCP's default SSL policy, weak TLS versions, or weak/custom cipher configurations.

   This also hardens GCP org-wide sync behavior by skipping projects/resources blocked by `compute.projects.get` permission errors or VPC Service Controls, matching existing
 permission-denied skip behavior.


   ### Related issues or links

   - N/A


   ### Breaking changes

   None.


   ### How was this tested?

   Local validation:

   ```bash
   make test_lint
   uv run --frozen pytest -q tests/unit/cartography/intel/gcp/test_ssl_policy_exposure.py tests/unit/cartography/intel/gcp/test_compute_http_errors.py
 tests/unit/cartography/intel/gcp/test_storage.py tests/unit/cartography/intel/gcp/test_util.py::TestIsPermissionDeniedError
   uv run --frozen pytest -q tests/integration/cartography/intel/gcp/test_ssl_policy_exposure.py
   uv run --frozen --group doc ./docs/build.sh
 ```

 Results:

 ```text
   make test_lint: passed
   GCP targeted unit tests: 37 passed
   SSL policy integration test: 1 passed
   docs build: succeeded
 ```

 Real-environment validation:

 Ran a full GCP sync against a real GCP environment with this branch enabled. Sanitized validation queries confirmed:

 - 28 GCPTargetHttpsProxy nodes had attached GCPSslPolicy nodes.
 - 0 target HTTPS proxies were using GCP's default SSL policy.
 - Forwarding rule → target proxy → SSL policy traversal resolved correctly.
 - All observed attached SSL policies were RESTRICTED / TLS_1_2.
 - Regional proxy query ran cleanly; this environment had only global target HTTPS proxies.
 - 0 dangling forwarding-rule target references were found.

 Sanitized sync evidence:

 ```text
   INFO:cartography.intel.gcp.ssl_policy:Syncing GCP SSL policies for project <redacted>
   INFO:cartography.intel.gcp.target_https_proxy:Syncing GCP target HTTPS proxies for project <redacted>
   INFO:cartography.intel.gcp.target_ssl_proxy:Syncing GCP target SSL proxies for project <redacted>
   INFO:cartography.graph.job:Finished job GCPSslPolicy
   INFO:cartography.graph.job:Finished job GCPTargetHttpsProxy
   INFO:cartography.graph.job:Finished job GCPTargetSslProxy

   Validation query counts:
   GCPTargetHttpsProxy nodes with attached GCPSslPolicy: 28
   GCPTargetHttpsProxy nodes using default SSL policy: 0
   Dangling forwarding-rule target references: 0
 ```

 ### Checklist

 #### General

 - [x] The linter passes locally (make test_lint).

 #### Proof of functionality

 - [x] Included sanitized Cartography sync logs demonstrating successful synchronization of the new/modified connector behavior. Logs should show:

 - [x] Updated model docstrings and PropertyRef.description values used to generate schema documentation.
 - [x] Built the documentation with uv run ./docs/build.sh.

 ### Notes for reviewers

 Review focus areas:

 - Relationship direction for forwarding rules to target proxies uses traffic-flow semantics: (:GCPForwardingRule)-[:ROUTES_TO]->(:GCPTargetHttpsProxy|GCPTargetSslProxy).
 - SSL policy resources are loaded before target proxies, and target proxies are loaded before forwarding rules, so relationships can resolve in the same sync cycle.
 - Regional SSL policy and regional target HTTPS proxy support is included; the real validation environment only had global target HTTPS proxies, while the integration test covers
   regional fixtures.
 ```

 ```

### CODE REVIEW

**Comment by ashton-suire** [comment]
```md
   Addressed in latest push, feedback from cubic:

   - Added expected access/API-denial handling for SSL policy, target HTTPS proxy, and target SSL proxy list calls.
   - These collectors now skip without aborting the broader Compute sync and avoid cleanup when the collection was not verifiably listed.
   - Moved forwarding-rule cleanup into a `finally` so cleanup still runs after global forwarding rules load if a regional listing raises.
   - Added unit coverage for denied collector cleanup behavior and forwarding-rule cleanup-on-regional-error behavior.

   Validation:
   - pre-commit on changed files passed
   - targeted GCP unit/integration tests passed: 42 passed
 ```

File: cartography/intel/gcp/compute.py [hunk 1]
```diff
@@ -1394,17 +1397,19 @@ def sync_gcp_forwarding_rules(
...
-    for r in regions:
-        fwd_response = get_gcp_regional_forwarding_rules(project_id, r, compute)
-        if fwd_response is None:
-            # Invalid region, skip this one
-            continue
-        forwarding_rules = transform_gcp_forwarding_rules(fwd_response)
-        load_gcp_forwarding_rules(
-            neo4j_session, forwarding_rules, gcp_update_tag, project_id
-        )
+    try:
```
kunaals: this cleanup still runs when a regional forwarding-rule request fails. because cleanup is project-scoped, it deletes previously ingested rules from the failed region even though we never got a complete replacement set. can we only clean up after the regional loop completes successfully?

cartography/intel/gcp/ssl_policy.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    logger.info("Syncing GCP SSL policies for project %s", project_id)
+    response = get_gcp_global_ssl_policies(project_id, compute)
+    if response is None:
+        return
+    policies = transform_gcp_ssl_policies(response, project_id)
+    load_gcp_ssl_policies(neo4j_session, policies, gcp_update_tag, project_id)
+
+    cleanup_safe = True
+    for r in regions:
```
kunaals: this lists every region derived from zones, and target https proxies do the same below. that adds two compute api calls per region per project. both resources support aggregatedlist, which returns global and regional results together. can we use that here and preserve the no-cleanup behavior when a scope is unreachable?

cartography/models/gcp/compute/target_ssl_proxy.py
```diff
@@ -0,0 +1,125 @@
...
+        "description", description="An optional description of this target proxy."
+    )
+    service: PropertyRef = PropertyRef(
+        "service_partial_uri",
+        description="A partial resource URI of the backend service this target proxy forwards to.",
+    )
+    ssl_policy: PropertyRef = PropertyRef(
+        "ssl_policy_partial_uri",
+        description="A partial resource URI of the SSL policy attached to this target proxy. "
+        "Absent means the client is using GCP's insecure default SSL policy.",
```
kunaals: minor wording point: an absent sslpolicy means the proxy has no ssl policy configured, rather than that it is using a specific gcp insecure default policy. the ingestion behavior is already right; could we make the description and matching test say that instead?

kunaals (changes requested):

cartography/intel/gcp/compute.py
ashton-suire: Yep, good point. I changed this back.

Forwarding rule cleanup now only runs after the regional forwarding-rule loop finishes successfully. If a regional request fails, we leave the existing data alone instead of cleaning up a region we could not refresh.

cartography/models/gcp/compute/target_ssl_proxy.py
ashton-suire: I updated the model descriptions and the test comment to say an absent `sslPolicy` means no SSL policy is configured on the proxy.

cartography/intel/gcp/ssl_policy.py
ashton-suire: I switched SSL policies and target HTTPS proxies over to aggregated list calls:

> `sslPolicies().aggregatedList(returnPartialSuccess=True)`
> `targetHttpsProxies().aggregatedList(returnPartialSuccess=True)`

That removes the per-region calls. The transforms now handle the aggregated global and regional response shape. Cleanup is skipped if the aggregated response reports an unreachable scope, so we do not delete data when GCP only gave us a partial result.

I updated the fixtures and tests too.

ashton-suire:
```md
   Thanks again for the review. All three requested updates are in the latest push.

   Summary:
   - Forwarding-rule cleanup now only runs after the regional forwarding-rule loop completes successfully.
   - SSL policies and target HTTPS proxies now use aggregated list calls with `returnPartialSuccess=True`.
   - Cleanup is skipped if the aggregated response reports an unreachable scope.
   - Updated the absent `sslPolicy` wording to say no SSL policy is configured on the proxy.
   - Updated fixtures/tests for the aggregated global and regional response shape.

   Validation:
   - pre-commit on changed files passed
   - targeted GCP unit/integration tests passed: 42 passed
   - docs build succeeded

   Ready for another look when you have a chance.
 ```

File: cartography/intel/gcp/ssl_policy.py [hunk 4]
```diff
@@ -0,0 +1,161 @@
...
+        except HttpError as e:
+            if classify_gcp_http_error(e) in _EXPECTED_SKIP_CATEGORIES:
+                logger.warning(
+                    "GCP: Unable to list SSL policies for project %s; skipping this collector. %s",
+                    project_id,
+                    summarize_gcp_http_error(e),
+                )
+                return None
+            raise
+        items.update(res.get("items", {}))
```
kunaals: this overwrites an earlier page when aggregatedlist returns the same scope again. a scope with more than one page would keep only the final page, then cleanup could delete the omitted policies. can we merge each scope’s resource list across pages instead, and add a two-page same-scope regression test? the target https proxy collector has the same pattern.

kunaals (changes requested):

cartography/intel/gcp/ssl_policy.py
ashton-suire: Good catch. Fixed in the latest push.

I changed the aggregated response handling so repeated scopes across pages are merged instead of overwritten. So if `aggregatedList` returns `global` or the same region on more than one page, we now append the `sslPolicies` / `targetHttpsProxies` from each page instead of keeping only the last page.

I added regression tests for the two-page same-scope case for both collectors.

I also preserved unreachable-scope warnings across pages, so cleanup stays disabled if any page tells us a scope was not fully readable.

kunaals (approved):

PR #3151, opened 2026-08-13
fix(semgrep): handle null reachability in SCA vuln transform

amdadulbari:
### Type of change

### Summary

`transform_sca_vulns` in `cartography/intel/semgrep/findings.py` crashed with `AttributeError: 'NoneType' object has no attribute 'upper'` whenever a Semgrep Supply Chain (SCA) vulnerability had `reachability: null`.

Semgrep only runs reachability analysis for supported ecosystems/rules; for everything else the API returns `reachability: null`. The transform called `.upper()` on that value unconditionally, so any SCA sync that returned such a finding failed.

The fix mirrors the null handling already used one line below for the sibling `reachable_condition` field, and in `_determine_exposure`, which both already treat `reachability` as nullable:

```python
sca_vuln["reachability"] = (
    vuln["reachability"].upper() if vuln["reachability"] else None
)
```

### Related issues or links

Unreported bug found while reading the module. Not tied to an existing issue.

### How was this tested?

- Added `tests/unit/cartography/intel/semgrep/test_findings.py` with a regression test (`test_transform_sca_vulns_handles_null_reachability`) plus a happy-path assertion. The regression test **fails before the fix** (`AttributeError`) and **passes after**.
- `make test_unit` for the Semgrep module: 22 passed.
- `black`, `isort`, and `flake8` are clean on the changed files.

#### General

#### Proof of functionality
- [x] New unit tests.


tests/unit/cartography/intel/semgrep/test_findings.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+
+import copy
+
+from cartography.intel.semgrep.findings import transform_sca_vulns
+from tests.data.semgrep.sca import SCA_RESPONSE
+
+
+def test_transform_sca_vulns_uppercases_reachability():
+    raw_vulns = copy.deepcopy(SCA_RESPONSE["findings"])
```
jychp: Add `# Arrange`, `# Act`, and `# Assert` phase comments to both new tests.

jychp (approved):

PR #3153, opened 2026-08-13
feat(huntress): add Huntress EDR intel module

jychp:
### Type of change
- [ ] Documentation update

### Summary

Adds an intel module for [Huntress](https://www.huntress.com/), a managed EDR/MDR platform. It ingests the account, the customer organizations under it, the agents deployed on their endpoints, the incident reports the Huntress SOC raises, and the users who have access to the console.

This gives the graph the same two things the existing EDR connectors provide: per-endpoint EDR coverage (which devices are protected, and by extension which are not), and security findings attached to the canonical ontology `Device`.

Graph shape:

```
(:HuntressAccount:Tenant)
  -[:RESOURCE]-> (:HuntressOrganization:Tenant)
  -[:RESOURCE]-> (:HuntressAgent)                        -[:MEMBER_OF]-> (:HuntressOrganization)
  -[:RESOURCE]-> (:HuntressIncidentReport:SecurityIssue) -[:AFFECTS]->   (:HuntressAgent)
                                                         -[:MEMBER_OF]-> (:HuntressOrganization)
  -[:RESOURCE]-> (:HuntressUser:UserAccount)             -[:HAS_ROLE]->  (:HuntressRole:PermissionRole)
                                                         -[:MEMBER_OF]-> (:HuntressOrganization)

(:Device:Ontology)        -[:OBSERVED_AS]-> (:HuntressAgent)   # serial_number, hostname matchlink fallback
(:User)                   -[:HAS_ACCOUNT]-> (:HuntressUser)
(:HuntressIncidentReport) -[:AFFECTS]->     (:Device)          # analysis job walking OBSERVED_AS
```

Ontology mappings added for `tenants`, `devices`, `security_issues`, `useraccounts` and `roles`.

Scope is deliberately EDR plus console access. Billing/reseller/invoices, identities (ITDR), SAT and signals are not ingested.

#### Design decisions worth a reviewer's attention

**The account is the sub-resource owner, not the organization.** An API credential resolves to exactly one Huntress account and reads every organization under it, so the account is the correct cleanup boundary. Agents and incident reports additionally point at their organization with `MEMBER_OF`. This mirrors `MiradoreDevice`, which is a sub-resource of the tenant and a `MEMBER_OF` its organization.

**Roles are synthesised.** Huntress exposes no role object, only a bare `permissions` string on each membership. Following the `ModalWorkspaceRole` precedent, a role node is materialised with `id = "<account or organization ID>/<permission label>"`. Scoping the id keeps "Admin on the account" and "Admin on org X" as two distinct grants, and materialising it as a node (rather than a property) lets Huntress console access participate in the cross-provider `(:UserAccount)-[:HAS_ROLE]->(:PermissionRole)` rules that `ONTOLOGY_REL_CONSTRAINTS` already blesses.

Trade-off: per-membership metadata (`membership.id`, `created_at`) is not retained. Users carry their grants through `one_to_many` `HAS_ROLE` / `MEMBER_OF` edges instead of a reified membership node. Happy to reify it if reviewers would rather keep those fields.

**No `load_tenant()` helper.** Unlike Miradore, whose tenant node is synthesised from a config value and carries only an `id`, `HuntressAccount` has real properties (`name`, `subdomain`, `status`). A minimal `load_tenant({"id": ...})` would therefore null them out on every call. Instead `account.sync()` runs first and returns the `account_id` that scopes everything downstream.

**A 403 on memberships must not trigger cleanup.** Listing memberships needs a permission Huntress does not grant to every API credential. `memberships.get()` returns `None` rather than an empty list on 403, so `sync()` skips both the load and the cleanup; an empty list would look like a successful empty sync and delete every user and role a previously-authorized run had ingested. This follows `cartography/intel/gitlab/ci_variables.py`. There is an integration test for exactly this.

### Related issues or links

- [Huntress API reference](https://api.huntress.io/docs) (the OpenAPI document itself is at `https://api.huntress.io/v1/swagger_doc.json`; the `/docs` page is a Scalar SPA)

### How was this tested?

**Automated**

- `make test_lint` passes, and `pre-commit run --all-files` exits 0 without modifying the tree on a second run.
- `make test_unit`: 4727 passed. 24 of them are new Huntress unit tests covering the config guard, session reuse and teardown, cursor pagination (including the single-page response shape, a repeated page token, and a body missing the collection key), the 403-to-`None` path, and the transforms.
- `make test_integration`: 1524 passed. 3 new Huntress integration tests exercise the real `sync()` functions end to end: the full graph, tenant-scoped cleanup (a stale agent is dropped while another account's agent survives), and the memberships-403 case. 2 new ontology device tests cover `OBSERVED_AS` by serial and the hostname fallback, plus 1 for `HuntressIncidentReport -[:AFFECTS]-> Device`.
- `uv run --group doc ./docs/build.sh` succeeds and generates `modules/huntress/schema.html`.

**Against the real Huntress API**

Run against a live Huntress tenant with `--selected-modules create-indexes,huntress,ontology --ontology-devices-source huntress`. Sanitized logs:

```
INFO:cartography.sync:Starting sync stage 'huntress'
INFO:cartography.client.core.tx:Loaded 1 HuntressAccount nodes
INFO:cartography.client.core.tx:Loaded 1 HuntressOrganization nodes
INFO:cartography.graph.statement:Completed HuntressOrganization statement #1
INFO:cartography.graph.statement:Completed HuntressOrganization statement #2
INFO:cartography.graph.statement:Completed HuntressAgent statement #1
INFO:cartography.graph.statement:Completed HuntressAgent statement #2
INFO:cartography.graph.statement:Completed HuntressAgent statement #3
INFO:cartography.graph.statement:Completed HuntressIncidentReport statement #1
INFO:cartography.graph.statement:Completed HuntressIncidentReport statement #2
INFO:cartography.graph.statement:Completed HuntressIncidentReport statement #3
INFO:cartography.graph.statement:Completed HuntressIncidentReport statement #4
WARNING:cartography.intel.huntress.memberships:Huntress API credentials are not authorized to list memberships. Skipping console users and roles.
INFO:cartography.sync:Finishing sync stage 'huntress'
```

Resulting graph (identifiers redacted):

```
['HuntressAccount', 'Tenant']       1
['HuntressOrganization', 'Tenant']  1

account: {name: '<redacted>', subdomain: '<redacted>', status: 'enabled',
          _ont_name: '<redacted>', _ont_status: 'active', _ont_source: 'huntress'}
rels:    (:HuntressAccount)-[:RESOURCE]->(:HuntressOrganization)  x1
```

This confirms basic auth, cursor pagination, the account and organization syncs, the ontology tenant mapping (`enabled` normalising to `active`), and graceful degradation on the memberships 403.

**Please read this before approving:** the tenant available for testing is empty (0 agents, 0 incident reports) and its credential is not authorized for `/v1/memberships`. So the real-environment run above does **not** exercise the agent, incident report, or memberships-happy-path loads. Those are covered only by integration tests against fixtures derived from the vendor's OpenAPI examples. If a reviewer has access to a populated Huntress tenant, a confirmation run on the agent and incident report syncs would be genuinely valuable.

#### General

#### Proof of functionality

- [x] Tested the connector against a real cloud, SaaS, or provider environment. (Partially: see the caveat above. Account and organization syncs verified live; agents, incident reports and memberships are fixture-tested only.)

### Notes for reviewers

- The whole module lands in one commit on purpose: the `test_available_modules_import` and `test_schema_doc` guards both fail on any split that adds `cartography/intel/huntress/` without simultaneously registering it in `TOP_LEVEL_MODULES` and `docs/root/usage/schema.md`.
- `HuntressIncidentReport.body` and `.summary` are unbounded analyst prose and are deliberately left out of every index, since Neo4j rejects indexed values above roughly 8 KB.
- `indicator_counts` is dropped in the transform: it is a map, which Neo4j cannot store as a node property. Only `indicator_types` is kept.
- The `_HUNTRESS_INCIDENT_STATUS` ontology map enumerates all six values from the endpoint's `status` filter enum, not just the three the field description lists. The generated `CASE` has no `ELSE`, so an unenumerated value would silently normalise to null.
- Huntress documents a 60 req/min rate limit while the live `x-huntress-api-call-limit` header reports 120. The session retries 429 with exponential backoff either way.
- Not ingested, and easy follow-ups if wanted: `/v1/signals` (raw detection feed, but weakly typed: `entity` and `details` are free-form objects) and per-incident `/v1/incident_reports/{id}/remediations` (one call per incident, which is expensive against the rate limit; the first ten remediations are already inlined on the incident and summarised on the node).


cartography/intel/huntress/memberships.py
```diff
@@ -0,0 +1,166 @@
...
+            },
+        )
+        if organization_id is not None:
+            entry["organization_ids"].add(organization_id)
+
+        permissions = membership.get("permissions")
+        if permissions is None:
+            continue
+        scope_id = organization_id if organization_id is not None else account_id
+        role_id = f"{scope_id}/{permissions}"
```
kunaals: account and organization ids share this role id namespace, so an account and organization with the same numeric id collapse into one role and attach users to the wrong grant. can we include the scope type in the id and add an overlapping-id regression test?

kunaals (changes requested):

cartography/intel/huntress/memberships.py
```diff
@@ -0,0 +1,171 @@
...
+        # Every membership documents a user. A malformed one is not skipped: skipping it
+        # would leave that user out of the load while the cleanup below still runs,
+        # silently deleting a console user a previous sync had ingested.
+        if not isinstance(user, dict):
+            raise ValueError(
+                f"Huntress returned membership {membership.get('id')!r} with no user object",
+            )
+        user_id = required_id(user, "membership user")
+
+        organization = membership.get("organization")
```
kunaals: this treats any membership without a usable organization id as account-scoped, even when the account object is absent or the organization object is malformed. that can turn an organization grant into an account-wide role. can we validate exactly one scope object and its id before synthesizing the role?

kunaals (approved):

PR #3157, opened 2026-08-14
feat(orca): add Orca Security module

kunaals:
### Type of change

### Summary

Adds an organization-wide Orca Security intel module focused on Orca's unique security findings instead of duplicating cloud inventory that Cartography already ingests from AWS, Azure, and GCP.

The module:

- authenticates with an Orca service token and discovers the organization through `/api/user/action`;
- streams `Alert` and `VulnerabilityV2` results from the Serving Layer API with bounded retries and fail-closed pagination;
- models `OrcaOrganization:Tenant`, `OrcaAlert:SecurityIssue`, and per-target `OrcaVulnerabilityFinding:CVE` occurrences;
- owns findings directly through organization-scoped `RESOURCE` relationships and lets CVE metadata enrich vulnerability occurrences directly;
- retains exact provider and related-Inventory identifiers as finding context for future native-cloud correlations; and
- defers cleanup until both finding feeds complete, preserving the last-known-good graph after partial failures.

V1 deliberately does not enumerate standalone Orca Inventory, create duplicate `OrcaAsset` nodes, or add speculative `AFFECTS` relationships. It also avoids generic CVE nodes, `Risk`, `LINKED_TO`, MatchLinks, and fuzzy correlations.

### Related issues or links

- [Orca Security API client](https://github.com/orcasecurity/terraform-provider-orcasecurity/blob/master/orcasecurity/api_client/api_client.go)
- [Orca Security organization API contract](https://github.com/orcasecurity/terraform-provider-orcasecurity/blob/master/orcasecurity/api_client/organizations.go)
- [Orca Security Serving Layer query builder](https://github.com/orcasecurity/content-hub/blob/main/content/response_integrations/third_party/partner/orca_security/core/query_builder.py)

### Breaking changes

None. The connector has not been released.

### How was this tested?

- `uv run pytest -q tests/unit/cartography/intel/orca tests/unit/cartography/test_cli_orca.py tests/unit/cartography/test_config.py tests/unit/cartography/test_sync.py` (73 passed)
- `uv run pytest -q tests/integration/cartography/intel/orca/test_orca.py` (7 passed)
- `make test_lint`
- `uv run ./docs/build.sh`

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

No Orca endpoint or token was available in the implementation environment. Before production use, a read-only organization-wide tenant should still validate the current regional endpoint, least-privilege permissions, page-size/rate-limit behavior, and sanitized synchronization counts.

The Alert query requests related Inventory only as optional target context. Public Orca examples do not establish the exact full-graph response shape, so authenticated Swagger/Request Builder output must confirm it. Alert ingestion and cleanup do not depend on that context, and the module never falls back to names, IP addresses, or guessed joins.

Authenticated validation must also confirm that offset pagination remains stable while a sync runs through either a unique deterministic order or snapshot semantics. Orca public examples do not establish that contract, so this remains part of the real-environment validation gap.


cartography/intel/orca/vulnerabilities.py
```diff
@@ -0,0 +1,269 @@
...
+                        "type": "object",
+                        "operator": "has",
+                    },
+                ],
+            },
+        },
+        "additional_models[]": ["InstalledPackage", "Inventory"],
+        "flat_json": True,
+        "full_graph_fetch": {"enabled": True},
+        "max_tier": 2,
```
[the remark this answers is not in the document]
kunaals: Valid risk, but there is no publicly established unique VulnerabilityV2 sort key to add safely. Orca official query builders leave ordering optional, so inventing one could make the query invalid. The PR notes now call out deterministic ordering or snapshot semantics as an authenticated tenant-validation requirement.

cartography/intel/orca/api.py
```diff
@@ -0,0 +1,263 @@
...
+logger = logging.getLogger(__name__)
+
+CONNECT_TIMEOUT_SECONDS = 10
+READ_TIMEOUT_SECONDS = 120
+REQUEST_TIMEOUT = (CONNECT_TIMEOUT_SECONDS, READ_TIMEOUT_SECONDS)
+SERVING_LAYER_PATH = "/api/serving-layer/query"
+ORGANIZATION_PATH = "/api/user/action"
+MAX_PAGES = 10_000
+_PROGRESS_PAGE_INTERVAL = 10
+_RETRY_STATUS_CODES = (408, 429, 502, 503, 504)
```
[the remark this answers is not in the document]
kunaals: Leaving this unchanged. The requested retry contract intentionally limits status retries to 408, 429, 502, 503, and 504; a 500 is not assumed transient. Any such failure aborts the sync before cleanup, preserving the last-known-good graph.

cartography/intel/orca/response.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+
+
+def parse_datetime(value: Any, field: str) -> datetime | None:
+    """Parse an optional Orca RFC 3339 timestamp into a Neo4j-safe datetime."""
+    if value is None:
+        return None
+    normalized = require_nonempty_string(value, field)
+    try:
+        timestamp = datetime.fromisoformat(
```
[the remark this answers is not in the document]
kunaals: Fixed in 7334fd5e1 by using dateutil.parser.isoparse.

cartography/intel/orca/api.py
```diff
@@ -0,0 +1,263 @@
...
+            raise RuntimeError(
+                f"Orca {result_name} returned more rows than total_items",
+            )
+
+        page_count += 1
+        yield rows
+        start_index = next_index
+
+        if page_count % _PROGRESS_PAGE_INTERVAL == 0:
+            logger.info(
```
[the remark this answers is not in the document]
kunaals: Fixed in 7334fd5e1; page-level progress now logs at DEBUG.

cartography/intel/orca/vulnerabilities.py
```diff
@@ -0,0 +1,251 @@
...
+        if not cve_ids:
+            raise ValueError("Orca vulnerability row omitted a canonical CveId")
+        source_link = optional_nonempty_string(
+            vulnerability.get("SourceLink"),
+            "Orca vulnerability.SourceLink",
+        )
+        common_fields = {
+            "orca_id": orca_id,
+            "description": vulnerability.get("Description"),
+            "references": [source_link] if source_link else [],
```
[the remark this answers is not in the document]
kunaals: Leaving this intentionally narrow. V1 maps the SourceLink field established by the public contract; the earlier References handling was speculative and had no representative public fixture. We can add References after authenticated contract capture proves its shape.

cartography/intel/orca/vulnerabilities.py
```diff
@@ -0,0 +1,251 @@
...
+        return value or [{}]
+    raise ValueError(
+        "Orca vulnerability.InstalledPackage must be an object or object list",
+    )
+
+
+def _package_key(package: dict[str, Any]) -> tuple[str, ...]:
+    for key in ("id", "PURL", "CPE"):
+        value = optional_nonempty_string(
+            package.get(key),
```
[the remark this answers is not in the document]
kunaals: The VulnerabilityV2 query explicitly requests flat_json=true, and Orca public GetVulnerabilityDetails output contains flat InstalledPackage scalar fields. Envelope handling applies to the non-flat Alert response. Keeping direct field access here avoids supporting a contradictory, unobserved shape; authenticated tenant capture can supersede the public contract if needed.

cartography/intel/orca/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    if not api_endpoint and not api_token:
+        logger.info(
+            "Orca import is not configured - skipping this module. "
+            "Set orca_api_endpoint and orca_api_token to enable.",
+        )
+        return
+    if not api_endpoint or not api_token:
+        raise ValueError(
+            "Orca requires both orca_api_endpoint and orca_api_token",
+        )
```
jychp: blocking: Log and return here instead of raising. All 42 other intel modules skip on incomplete config; a raise from a top-level stage propagates out of `Sync.run` and kills the entire cartography run. Since `--orca-api-token-env-var` defaults to `ORCASECURITY_API_TOKEN`, an operator who merely has that variable exported but no `--orca-api-endpoint` loses their whole multi-module sync. The `normalize_api_endpoint()` ValueError on line 34 has the same blast radius: catch it and skip.

cartography/intel/orca/vulnerabilities.py
```diff
@@ -0,0 +1,251 @@
...
+        build_query(),
+        page_size=PAGE_SIZE,
+        result_name="vulnerabilities",
+    ):
+        vulnerabilities = transform(page, organization_id)
+        page_ids = {vulnerability["id"] for vulnerability in vulnerabilities}
+        if len(page_ids) != len(vulnerabilities) or page_ids & seen_ids:
+            raise RuntimeError(
+                "Orca vulnerabilities response contained duplicate identities",
+            )
```
jychp, line 285: blocking: A cross-page identity collision aborts the whole Orca sync. Finding identity is `(organization, asset_unique_id, cve_id, package_key)`, and nothing in the query guarantees Orca emits at most one `VulnerabilityV2` row per that tuple: every row without an `InstalledPackage` collapses to the `("asset-wide",)` key, and two advisory records for the same package collide as well. Merge duplicates or skip them with a warning instead of raising.

cartography/intel/orca/vulnerabilities.py
```diff
@@ -0,0 +1,251 @@
...
+                    "package_version": package.get("Version"),
+                    "purl": package.get("PURL"),
+                    "cpe": package.get("CPE"),
+                    "source_package": package.get("SourcePackage"),
+                }
+                existing = transformed.get(finding_id)
+                if existing is not None and existing != finding:
+                    raise ValueError(
+                        "Orca vulnerability rows contained conflicting identities",
+                    )
```
jychp: Same collision on the intra-page path, and it is stricter than it looks: `common_fields` carries `orca_id`, which is the row id, so two rows sharing an identity can never compare equal and always raise. Restrict the equality check to identity-bearing fields, or merge deterministically.

cartography/intel/orca/alerts.py
```diff
@@ -0,0 +1,186 @@
...
+    transformed: list[dict[str, Any]] = []
+    unresolved_targets = 0
+    for raw_alert in raw_alerts:
+        data = require_object(
+            raw_alert.get("data"),
+            "Orca Alert.data",
+        )
+        alert_id = require_nonempty_string(
+            unwrap_value(data["AlertId"]),
+            "Orca AlertId",
```
jychp, line 81: Use `field_value(data, "AlertId")` here. Direct subscripting raises a bare `KeyError` on a missing field while every other field in this module raises a `ValueError` naming the Orca field, so the one mandatory field has the worst error message.

cartography/intel/orca/alerts.py
```diff
@@ -0,0 +1,186 @@
...
+        transformed.append(
+            {
+                "id": f"orca:{organization_id}:{alert_id}",
+                "orca_id": alert_id,
+                "title": title,
+                "details": field_value(data, "Details"),
+                "severity": field_value(data, "Severity"),
+                "category": field_value(data, "Category"),
+                "alert_type": alert_type,
+                "orca_score": field_value(data, "OrcaScore"),
```
jychp: Validate these scalars the way the target context is validated. `severity`, `category`, `status`, `alert_type`, `details`, `console_url`, and `orca_score` go straight into `load()` unchecked, so a nested object or list in any of them surfaces as a Neo4j driver error mid-write instead of a named Orca contract violation.

cartography/models/ontology/mapping/data/cves.py
```diff
@@ -30,6 +30,21 @@
     "INFORMATIONAL": "info",
 }
 
+# Orca's VulnerabilityV2 query emits uppercase CVSS severity names. Include
+# lowercase variants so normalized results remain stable across API casing.
+_ORCA_CVSS_SEVERITY = {
```
jychp, line 35: Build this from the existing constant: `_ORCA_CVSS_SEVERITY = {**_CVSS_SEVERITY, "none": "info", "low": "low", ...}`. The five uppercase entries are a verbatim copy of `_CVSS_SEVERITY` and will drift from it.

docs/root/modules/orca/config.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Configure Cartography
+
+Set `--orca-api-endpoint` to your regional HTTPS API origin without `/api` or a
+route. Cartography reads the token from `ORCASECURITY_API_TOKEN` by default.
+
+| Option | Default | Required | Description |
+|--------|---------|----------|-------------|
+| `--orca-api-endpoint` |  | Yes | Regional Orca API origin. |
+| `--orca-api-token-env-var` | `ORCASECURITY_API_TOKEN` | Yes | Environment variable that contains the Orca API token. |
```
jychp: Change Required to No for `--orca-api-token-env-var`; it has a default, so it is only required when the token lives under a different variable name.

README.md
```diff
@@ -119,6 +119,7 @@ password options in [the rules docs](https://docs.cartography.dev/usage/rules.ht
 - [Netlify](https://docs.cartography.dev/modules/netlify/index.html) - Accounts, Users, Invites, Sites, Deploys, Functions, Dev Servers, Agent Runners, Database Branches, Database Snapshots, Environment Variables, Build Hooks, Notification Hooks, Deploy Keys, Snippets, Service Instances, DNS Zones, DNS Records, Certificates, Forms
 - [Okta](https://docs.cartography.dev/modules/okta/index.html) - users, groups, organizations, roles, applications, factors, trusted origins, reply URIs, federation to AWS roles, federation to AWS Identity Center
 - [OpenAI](https://docs.cartography.dev/modules/openai/index.html) - Organization, AdminApiKey, User, Project, ServiceAccount, ApiKey
+- [Orca Security](https://docs.cartography.dev/modules/orca/index.html) - Organizations, security alerts, per-target CVE findings
```
jychp: Move this line below Oracle Cloud Infrastructure to keep the list alphabetical.

jychp (dismissed):

cartography/intel/orca/__init__.py
kunaals: fixed incomplete or invalid config now logs and skips without aborting the overall sync

cartography/intel/orca/vulnerabilities.py
kunaals: duplicate finding identities are now skipped with a counted warning across pages

cartography/intel/orca/vulnerabilities.py
kunaals: intra-page duplicates now keep the first occurrence and emit a counted warning

cartography/intel/orca/alerts.py
kunaals: switched this to field_value() so missing ids produce a named contract error

cartography/intel/orca/alerts.py
kunaals: alert strings and orca_score are now validated before load with field-specific errors

cartography/models/ontology/mapping/data/cves.py
kunaals: updated the orca map to compose _CVSS_SEVERITY and only add lowercase variants

docs/root/modules/orca/config.md
kunaals: fixed the table: the option is not required because it has a default

cartography/intel/orca/alerts.py
```diff
@@ -0,0 +1,213 @@
...
+from cartography.intel.orca.response import unwrap_value
+from cartography.models.orca import OrcaAlertSchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+PAGE_SIZE = 100
+
+
+def build_query() -> dict[str, Any]:
```
achantavy, line 29: Make a private func?

cartography/intel/orca/api.py
```diff
@@ -0,0 +1,262 @@
...
+    payload = _request_json(session, "GET", api_endpoint, ORGANIZATION_PATH)
+    data = payload.get("data")
+    if not isinstance(data, dict):
+        raise RuntimeError("Orca organization response data must be an object")
+    organization_id = data.get("organization_id")
+    organization_name = data.get("organization_name")
+    if not isinstance(organization_id, str) or not organization_id.strip():
+        raise RuntimeError("Orca organization response omitted organization_id")
+    if not isinstance(organization_name, str) or not organization_name.strip():
+        raise RuntimeError("Orca organization response omitted organization_name")
```
achantavy, line 125: Nits: are these isinstances actually needed or is there a way to do it without? 

Either way, non-blocking

achantavy (approved):
Just small nits but looks good

PR #3160, opened 2026-08-17
Ttsujimoto/digitalocean sdk upgrade

damixen:
### Type of change

- [ ] Documentation update

### Summary
- Migrated the DigitalOcean integration from the deprecated python-digitalocean SDK to the official DigitalOcean pydo SDK.
- Updated DigitalOcean resource handling to work with the dictionary-based responses returned by pydo.
- Added reusable pagination handling for DigitalOcean APIs that return paginated results, including support for a configurable maximum number of pages.
- Updated DigitalOcean integration tests to work with the new SDK and response format.
- Added unit tests covering pagination behavior.
- Updated test mocks to use mocked pydo clients/API responses rather than the old DigitalOcean SDK.
- Removed the old python-digitalocean dependency.
- Updated the dependency lock file.

### Related issues or links
N/A

### Breaking changes
No breaking changes

### How was this tested?
- make test_lint
- make test_unit (1 failed, as the same as the baseline master branch (32d094c5))
- make test_integration (2 failed, as same as the baseline master branch (32d094c5))
- Ran against a real DigitalOcean account with multiple projects and multiple droplets.

#### General

#### Proof of functionality

- Baseline master 32d094c5
2026-08-17_12-16-07-baseline

- New
2026-08-17_12-20-53-new

  - Example:
    ```
    ```


cartography/intel/digitalocean/util/pagination.py
```diff
@@ -0,0 +1,40 @@
...
+            "prev": "https://api.digitalocean.com/v2/tags?page=1",
+            "first": "https://api.digitalocean.com/v2/tags?page=1",
+            "last": "https://api.digitalocean.com/v2/tags?page=3"
+            }
+        }
+    }
+    """
+    data: list[dict[str, Any]] = []
+
+    for page_num in range(1, max_pages + 1):
```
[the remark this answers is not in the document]
damixen: I agree that returning a partial inventory could cause previously ingested resources to be cleaned up incorrectly.

I'm planning to follow the existing pagination pattern from **aws_paginate** in [cartography/util.py](https://github.com/cartography-cncf/cartography/blob/master/cartography/util.py#L519) and use a larger default maximum page count, e.g. **10000**. 

If the maximum is reached while additional pages are still available, I'll raise a **RuntimeError** rather than return a partial result. None will allow unlimited pagination.

cartography/intel/digitalocean/util/pagination.py
```diff
@@ -0,0 +1,40 @@
...
+        result = list_function(page=page_num, **kwargs)
+        data.extend(result.get(target_key, []))
```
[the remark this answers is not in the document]
damixen: I will treat the missing target object with a loud failure. I think this's good from a developer's point of view, especially if this function is used for other resources in the future.

cartography/intel/digitalocean/platform.py
```diff
@@ -38,18 +39,26 @@ def sync(
 
 
 @timeit
-def get_account(manager: Manager) -> Account:
-    return manager.get_account()
+def get_account(client: Client) -> Dict[str, Any]:
+    result = client.account.get()
+    if result is None:
+        return {}
```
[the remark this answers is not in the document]
damixen: I agree. I checked the pydo API reference, and it says account is the top-level response key for a successful client.account.get() call. The SDK raises HttpResponseError for API failures rather than returning an empty response. I'll use result["account"] so an unexpected response shape fails loudly.

cartography/intel/digitalocean/compute.py
```diff
@@ -34,8 +35,33 @@ def sync(
 
 
 @timeit
-def get_droplets(manager: Manager) -> list:
-    return manager.get_all_droplets()
+def get_droplets(client: Client) -> list:
+    return get_paginated_list(client.droplets.list, "droplets")
```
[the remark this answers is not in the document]
damixen: I will increase the limit and add loud failure as in:
https://github.com/cartography-cncf/cartography/pull/3160#discussion_r3799310270

.claude/skills
```diff
@@ -1 +1 @@
-../.agents/skills
\ No newline at end of file
+../.agents/skills
```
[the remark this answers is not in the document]
damixen: @jychp Should we exclude this file from linting?

pyproject.toml
```diff
@@ -60,7 +60,8 @@ dependencies = [
     "statsd>=4.0.0",
     "packaging>=26.0.0",
     "packageurl-python>=0.17.0",
-    "python-digitalocean>=1.16.0",
+    "adal>=1.2.4",
```
kunaals: this merge brings back `adal` and `azure-cli-core`, which were just removed from master, and the lockfile rolls cryptography, msal, pyopenssl, and workos back with them. can we drop both unrelated dependencies and regenerate the lockfile against current master?

cartography/intel/digitalocean/util/pagination.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    data: list[dict[str, Any]] = []
+
+    if max_pages is not None and max_pages <= 0:
+        raise ValueError("max_pages must be positive or None.")
+
+    page_num = 1
+
+    while True:
+        result = list_function(page=page_num, **kwargs)
```
kunaals: this leaves pydo at its 20-item page default, while the old client fetched 200. because project resources are already fetched once per project, this can turn a sync into 5-10x more requests and push larger accounts into rate limiting. can we default `per_page` to 200 here while still allowing callers to override it, and cover that default in the test?

cartography/intel/digitalocean/compute.py
```diff
@@ -34,8 +35,33 @@ def sync(
...
+
+
+@timeit
```
kunaals: this helper runs once per droplet, so `@timeit` emits one statsd timer per asset when metrics are enabled. can we remove the decorator and keep timing at the fetch and sync boundaries?

kunaals (changes requested):

pyproject.toml
damixen: I will do that. Thanks for pointing that out.

cartography/intel/digitalocean/util/pagination.py
damixen: Addressed in the latest commit.

cartography/intel/digitalocean/compute.py
damixen: Addressed in the latest commit.

pyproject.toml
damixen: Addressed in the latest commit.

kunaals (approved):
LGTM

Issue #3118, opened 2026-08-04
Scan image-backed Railway deployments through the Image ontology


kunaals:
# Scan image-backed Railway deployments through the Image ontology

## Description

Railway services deployed from an explicit Docker image run as containers, and Cartography correctly gives the current `RailwayDeployment` the `Container` label. The image reference is stored only on the parent `RailwayServiceInstance.source_image`, though, so the deployment has no `_ont_image`, `HAS_IMAGE`, or `RESOLVED_IMAGE`. Consumers that discover runtime images through the ontology cannot scan these Railway workloads.

This issue applies only when Railway reports `serviceInstance.source.image`. Git-backed services have `source.image = null` and are tracked separately as exact deployed source revisions; they must not be assigned an invented image identity.

Suggested implementation:

- When transforming deployments, match the current deployment to its service instance and copy `source_image` onto the deployment as `image_uri`. Parse `image_digest` when the source is digest-pinned.
- Map those properties to the Container ontology image fields. Keep `SLEEPING` as `suspended`, but include both `running` and `suspended` deployments when discovering images to scan; sleeping services wake when traffic arrives.
- Let the Trivy path scan pullable `_ont_image` references that do not yet have `RESOLVED_IMAGE`.
- When Trivy learns a concrete digest and no provider-owned `Image` exists, load a digest-keyed fallback node carrying both `TrivyImage` and the canonical `Image` ontology label. Link the workload with `HAS_IMAGE`, then run the existing resolved-image analysis to produce `RESOLVED_IMAGE`.
- Reuse a provider-owned `Image` when the digest already exists rather than creating a parallel node.

Add an integration test with an image-backed Railway service showing:

1. the current deployment receives `_ont_image`;
2. a sleeping deployment remains eligible for scanning;
3. Trivy creates or reuses a concrete `Image`;
4. the deployment ends with `RESOLVED_IMAGE` and the finding affects that image.

Digest-pinned sources can be modeled exactly. A mutable tag can move after Railway deploys it, so it should not be presented as the exact running digest unless Railway exposes that digest.

## Out of scope

Git-backed Railway services require exact source-revision evidence rather than an invented image identity. That source-scanning design is separate from this image-specific issue.

## Motivation

Railway services are real container workloads, including sleeping services that restart on incoming traffic. Leaving image-backed services outside the image ontology means vulnerability and SBOM consumers miss a deployed workload that can become reachable again immediately.

## Alternatives Considered

[...]

PR #3161, opened 2026-08-18
Add Railway deployments to the image ontology

ChrisJr404:
### Type of change
- [ ] Documentation update

### Summary

Wire `RailwayDeployment` into the container image ontology so image consumers can actually reach Railway workloads.

A Railway deployment already gets the `Container` label on its current revision, but the image it runs only lives on the parent `RailwayServiceInstance.source_image`. So the deployment had no image reference of its own, no `HAS_IMAGE`, and never picked up `RESOLVED_IMAGE` — vuln/SBOM consumers that walk the image ontology couldn't see it.

This copies `source_image` onto the current deployment as `image_uri` (parsing out the digest when the reference is digest-pinned via the shared `parse_image_uri` helper) and attaches the same `HAS_IMAGE` edges to the ECR / GitLab / Artifact Registry / GHCR image nodes that the other `Container` modules (Cloud Run, ECS, Kubernetes) already use. Matching is by digest, so an existing provider-owned `Image` gets reused rather than duplicated. Once `HAS_IMAGE` is present the existing `CONTAINER_RESOLVED_IMAGE` analysis derives `RESOLVED_IMAGE` with no extra work.

Only the current revision gets the image copied onto it. Railway keeps a row for every past deploy attempt, and a superseded revision may well have run a different image than the instance's current source — same reason the `Container` label is already conditional on `lifecycle="current"`.

Nothing new is pulled from the Railway API here — `source.image` is already ingested onto the service instance. This is purely a graph-modeling change over data the connector already syncs.

Scope note: this is the deployment-side ontology wiring from #3118. The generic Trivy path in that issue (scanning pullable references that don't yet resolve to a provider image, and loading a digest-keyed fallback `Image` when Trivy learns a concrete digest) is a separate, larger change and isn't part of this PR.

### Related issues or links

- Refs #3118

### How was this tested?

- `pytest tests/unit/cartography/intel/railway/` — added `test_deployments.py` covering the transform: a registry-backed current deployment gets `image_uri`, a git-backed one (`source.image = null`) stays empty, a digest-pinned reference splits into `image_uri` + `image_digest`, and a superseded revision gets nothing. 24 passed.
- `pytest tests/unit/cartography/graph/test_ontology_rel_constraints.py tests/unit/cartography/graph/test_ontology_fields.py` — the four new `HAS_IMAGE` rels are whitelisted alongside the existing container modules. 29 passed.
- Extended `tests/integration/.../test_deployments.py` to assert the current deployment carries `image_uri` from its instance (the existing fixture already has an `nginxdemos/hello` service and a git-backed one). This needs Neo4j to run and I don't have a live instance here, so I haven't executed it — flagging that for a reviewer.
- `flake8` clean, `mypy` clean on the changed files, `black` + `isort` applied.

#### General
- [x] The linter passes locally (flake8 / black / isort / mypy on the changed files).

#### Proof of functionality

- [ ] Built the documentation with `uv run ./docs/build.sh` (didn't run the docs build locally).

### Notes for reviewers

The change lives entirely in the transform + model — no new Railway API surface — so it shouldn't need real-environment sync evidence the way a new connector would. The one thing worth a look is the `(serviceId, environmentId)` join I use to find each deployment's parent instance; it's the same key the existing `WORKLOAD_PARENT` edge matches on.

### DIFF
--- a/cartography/models/railway/deployment.py
+++ b/cartography/models/railway/deployment.py
@@ -53,6 +53,18 @@ class RailwayDeploymentNodeProperties(CartographyNodeProperties):
     created_at: PropertyRef = PropertyRef(
         "createdAt", description="Time when the deployment was created."
     )
+    # Copied from the parent RailwayServiceInstance.source_image onto the current revision so
+    # the image ontology can reach it. Railway only exposes an image reference for the current
+    # deployment, so historical revisions leave these null.
+    image_uri: PropertyRef = PropertyRef(
+        "image_uri",
+        extra_index=True,
+        description="Container image reference this deployment runs, when deployed from a registry.",
+    )
+    image_digest: PropertyRef = PropertyRef(
+        "image_digest",
+        description="Digest (e.g. `sha256:...`) when `image_uri` is pinned by digest; `None` for tag-based references.",
+    )
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


cartography/models/railway/deployment.py
```diff
@@ -53,6 +53,18 @@ class RailwayDeploymentNodeProperties(CartographyNodeProperties):
     created_at: PropertyRef = PropertyRef(
         "createdAt", description="Time when the deployment was created."
     )
+    # Copied from the parent RailwayServiceInstance.source_image onto the current revision so
+    # the image ontology can reach it. Railway only exposes an image reference for the current
+    # deployment, so historical revisions leave these null.
+    image_uri: PropertyRef = PropertyRef(
```
kunaals, line 59: these fields don’t reach the container ontology yet. `railway_mapping` still maps only state, so `_ont_image` and `_ont_image_digest` stay empty and tag-based image discovery won’t see railway. please map `image` to `image_uri` and `image_digest` to `image_digest`, then assert both ontology fields in the integration test.

cartography/models/railway/deployment.py
```diff
@@ -101,6 +113,92 @@ class RailwayDeploymentToServiceInstanceRel(CartographyRelSchema):
...
+
+
+@dataclass(frozen=True)
+# (:RailwayDeployment)-[:HAS_IMAGE]->(:AWSECRImage), joined on digest.
+class RailwayDeploymentToECRImageRel(CartographyRelSchema):
+    """Links a deployment to the image it runs, hosted in Amazon ECR."""
+
+    target_node_label: str = "AWSECRImage"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"digest": PropertyRef("image_digest")},
```
kunaals, line 128: this digest can point to an `ImageManifestList`. without `architecture_normalized`, the resolved-image analysis can’t select its runnable child and leaves `RESOLVED_IMAGE` empty. railway’s docs specify `linux/amd64`, so please carry `amd64` onto current deployments and add a manifest-list resolution test.

kunaals (changes requested):

PR #3174, opened 2026-08-22
fix(gcp): add HTTP timeout to Vertex AI paginated API requests

### LINKED ISSUE #3173: GCP Vertex AI API requests missing timeout parameter -  sync process can hang indefinitely
### Expected behavior

All outbound HTTP requests in Cartography intel modules should include a `timeout` parameter to prevent the sync process from hanging indefinitely when an external API becomes unresponsive. This is consistent with other modules in the codebase (e.g., `bigfix/computers.py`, `github/util.py`, `semgrep/deployment.py`, `salesforce/util.py`) that all define and use a `_TIMEOUT` constant.

The GCP Vertex AI paginated resource fetcher (`list_all_resources_paginated`) should include a timeout on both the `session.get()` and `requests.get()` calls to ensure bounded execution time per API call.

### Actual behavior

The function `list_all_resources_paginated()` in `cartography/intel/gcp/vertex/utils.py` (lines 259–262) makes HTTP requests **without any timeout**:

```python
while True:
    params: dict[str, str] = {}
    if page_token:
        params["pageToken"] = page_token

    if session is not None:
        response = session.get(url, headers=request_headers, params=params)  # ← No timeout
    else:
        response = requests.get(url, headers=request_headers, params=params)  # ← No timeout
```

If the GCP Vertex AI API becomes unresponsive (network partition, API outage, DNS failure after initial connection), the HTTP request blocks **indefinitely**. Because this code runs inside a `while True` pagination loop, there is no recovery path — the process hangs permanently.

This causes:

- **Full pipeline stall:** Cartography runs sync modules sequentially. A hung Vertex AI sync blocks all subsequent modules (AWS, Azure, GitHub, etc.) from executing.
- **Silent failure:** The process appears alive (not crashed) but makes no progress. Health checks that only monitor process liveness won't detect this condition.
- **Resource leak:** The hung connection holds a TCP socket and Python thread indefinitely, potentially exhausting connection pools in long-running deployments.

Additionally, `import requests` is placed inside the function body (line 248) rather than at the module level, which is a minor code quality issue (PEP 8 violation).

### To Reproduce

1. Configure Cartography with GCP credentials and enable the Vertex AI sync module.
2. Start a Cartography sync.
3. Simulate an unresponsive Vertex AI endpoint using one of the following methods:
   - Block outbound traffic to `us-central1-aiplatform.googleapis.com` after the initial connection using a firewall rule.
   - Use a network proxy to introduce indefinite latency on the Vertex AI API responses.
   - Set up a mock server that accepts the TCP connection but never sends a response.
4. Observe that the Cartography sync process hangs at the Vertex AI resource fetch step and never recovers or times out.

### Logs

No error logs are produced — the process simply hangs silently on the `requests.get()` call. The last visible log entry would be the debug message from the previous successful page fetch. No timeout error or exception is raised.

```
INFO - Starting sync stage 'gcp'
INFO - Syncing GCP Vertex AI models for project my-project
DEBUG - Fetching Vertex AI models from us-central1...
<process hangs here indefinitely — no further output>
```

### Screenshots

N/A — This is a backend hang, not a UI issue.

- **Cartography release version or commit hash:** `f0b8e89eedb516faf3ce691c278dad2fe7354693`
- **Python version:** 3.13
- **OS:** Windows (reproducible on any OS)

### Additional context

**Suggested fix — Add timeout parameter:**

```python
_TIMEOUT = (60, 60)  # (connect_timeout, read_timeout) — consistent with other modules

def list_all_resources_paginated(...):
    # ... existing code ...
    while True:
        params: dict[str, str] = {}
        if page_token:
            params["pageToken"] = page_token

        if session is not None:
            response = session.get(url, headers=request_headers, params=params, timeout=_TIMEOUT)
        else:
            response = requests.get(url, headers=request_headers, params=params, timeout=_TIMEOUT)
```

**Suggested fix — Add pagination safety limit:**

The `while True` loop also lacks a maximum page counter. Consider adding a safety limit to prevent infinite pagination if the API misbehaves:

```python
_MAX_PAGES = 1000

for page_num in range(_MAX_PAGES):
    # ... existing pagination logic ...
    page_token = data.get("nextPageToken")
    if not page_token:
        break
else:
    logger.warning(
        "Reached max page limit (%d) for %s in %s/%s — possible API issue",
        _MAX_PAGES, resource_type, project_id, location,
    )
```

**Other modules for reference** — The following modules correctly implement `_TIMEOUT` and can serve as patterns:

| Module | File | Timeout |
|--------|------|---------|
| BigFix | `cartography/intel/bigfix/computers.py` | `_TIMEOUT = 30` |
| GitHub | `cartography/intel/github/util.py` | `_TIMEOUT = (60, 60)` |
| Semgrep | `cartography/intel/semgrep/deployment.py` | `_TIMEOUT = (60, 60)` |
| Salesforce | `cartography/intel/salesforce/util.py` | `_TIMEOUT = 30` |
| Spacelift | `cartography/intel/spacelift/util.py` | `_TIMEOUT = (10, 30)` |

The `requests` library documentation explicitly recommends: *"You should nearly always set a timeout on outbound requests."* — [Python Requests: Timeouts](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts)

ajinkyap9:
### Type of change

### Summary

Add timeout parameter to `session.get()` and `requests.get()` calls in `paginate_vertex_api()` to prevent indefinite hangs when the GCP Vertex AI API becomes unresponsive.

**Problem:**
The `paginate_vertex_api()` function in `cartography/intel/gcp/vertex/utils.py` makes HTTP requests without a `timeout` parameter. If the GCP Vertex AI API becomes unresponsive (network partition, API outage, DNS failure), the request blocks forever. Since this runs inside a `while True` pagination loop, there is no recovery path — the entire Cartography sync pipeline hangs permanently.

**Changes:**
- Add `_TIMEOUT = (60, 60)` constant for `(connect_timeout, read_timeout)`, consistent with other modules (`github/util.py`, `semgrep/deployment.py`, `salesforce/util.py`, `bigfix/computers.py`).
- Apply `timeout=_TIMEOUT` to both `session.get()` and `requests.get()` calls.
- Replace unbounded `while True` loop with a `for range(max_pages)` safety limit (`max_pages=1000`) to prevent infinite pagination if the API returns `nextPageToken` indefinitely.
- Move `import requests` from function body to module level (PEP 8 compliance).

Without a timeout, an unresponsive API call blocks the Python thread forever inside the while-True pagination loop. Since Cartography runs sync modules sequentially, this hangs the entire pipeline with no recovery path.

### Related issues or links

- Fixes #3173

### Breaking changes

None. This is a non-breaking bug fix that adds timeout protection to existing HTTP calls. The 60-second timeout is generous enough to handle normal API latency.

### How was this tested?

- Code review of the change against the existing patterns used by other modules in the codebase.
- Verified that other modules (`bigfix/computers.py`, `github/util.py`, `semgrep/deployment.py`, `salesforce/util.py`, `spacelift/util.py`) use the same `_TIMEOUT` pattern successfully.

#### General

#### Proof of functionality

### Notes for reviewers

This is a minimal, targeted fix for issue #3173. The `_TIMEOUT = (60, 60)` value matches the convention used by `github/util.py` and `semgrep/deployment.py`. The `max_pages=1000` safety limit is defensive — in practice, Vertex AI responses are well-bounded, but this protects against edge cases where the API misbehaves.

The `requests` import was moved from inside the function body to the module level for PEP 8 compliance and minor performance improvement (avoids re-importing on every function call).


achantavy (changes requested):
Hi, thanks for thje PR. Please remove the max pages param because it changes the behavior of the code.

Do keep the timeout bc I think that is the useful addition of this code: timing out and raising an exc is better than hanging indefinitely.

So yeah, the only change should be adding a timeout.

ajinkyap9:
Sure, I'll remove the max page param

tests/unit/cartography/intel/gcp/vertex/test_utils.py
```diff
@@ -88,3 +88,45 @@ def test_list_vertex_ai_resources_for_location_reraises_google_api_call_errors()
             location="us-central1",
             project_id="test-project",
         )
+
+
+@pytest.mark.parametrize("use_session", [True, False])
+def test_paginate_vertex_api_uses_timeout(use_session, monkeypatch):
```
achantavy: Remove this unit test, it's not needed

achantavy (changes requested):
Please remove the unnecessary unit test

tests/unit/cartography/intel/gcp/vertex/test_utils.py
ajinkyap9: okay, Sure I'll do it

PR #3176, opened 2026-08-23
feat(GCP): add GCP VPN tunnels and VPC peerings

bhargav2427:
### Type of change

- [ ] Documentation update

### Summary

Adds three new GCP Compute resources to the graph: **VPC peerings** (`GCPVpcPeering`), **VPN gateways** (`GCPVpnGateway`), and **VPN tunnels (`GCPVpnTunnel`). These are needed to model cross-network and cross-project connectivity: which VPCs peer with each other, and which VPN tunnels connect which gateways.

- VPC peerings are extracted from each network's existing `peerings` field (no extra API call). Each peering links to its local VPC via `LOCAL_NETWORK` and to the peer VPC via `PEER_NETWORK`. When the peer VPC lives in a project that hasn't been synced, a `GCPVpc` stub node is created so the relationship still resolves.
- VPN gateways and tunnels are synced per region. Gateways sync before tunnels so `USES_GATEWAY` relationships resolve in the same sync cycle; peer gateways outside the synced project are represented as `GCPVpnGateway` stub nodes linked via `CONNECTS_TO_GATEWAY`.
- Also hardens subnet sync: if any region returns incomplete data (timeout/invalid region), subnet cleanup is skipped for that project so existing `GCPSubnet` nodes are preserved instead of being deleted based on partial data.

### Related issues or links

- Fixes #

### How was this tested?
- Unit tests: `tests/unit/cartography/intel/gcp/test_compute.py` (transform logic for peerings, gateways, tunnels, stub extraction) and `test_compute_http_errors.py` (subnet timeout now returns None so cleanup is skipped). `uv run pytest tests/unit/cartography/intel/gcp/ -q` → all pass.
- Integration tests: `tests/integration/cartography/intel/gcp/test_compute.py` covers node creation, all new relationships (RESOURCE, LOCAL_NETWORK, PEER_NETWORK, PART_OF_VPC, USES_GATEWAY, CONNECTS_TO_GATEWAY), stub creation, and scoped cleanup against Neo4j.
- `pre-commit run` (isort/black/flake8/pyupgrade/mypy) passes on all changed files.

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### DIFF
--- a/cartography/intel/gcp/compute.py
+++ b/cartography/intel/gcp/compute.py
@@ -11,6 +11,7 @@ from googleapiclient.discovery import Resource
 from googleapiclient.errors import HttpError
 
 from cartography.client.core.tx import load
+from cartography.client.core.tx import run_write_query
 from cartography.graph.job import GraphJob
 from cartography.intel.gcp.backendservice import sync_gcp_backend_services
 from cartography.intel.gcp.cloud_armor import sync_gcp_cloud_armor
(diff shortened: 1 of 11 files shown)


cartography/intel/gcp/compute.py
```diff
@@ -1351,9 +1812,135 @@ def sync_gcp_vpcs(
...
+    :param compute: The GCP Compute resource object
+    :param project_id: The project ID to sync
+    :param regions: List of regions
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    :param common_job_parameters: dict of other job parameters to pass to Neo4j
+    :return: Nothing
+    """
+    incomplete = False
+    for r in regions:
+        vpn_gateway_res = get_gcp_vpn_gateways(project_id, r, compute)
```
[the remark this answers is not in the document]
bhargav2427: This is a valid comment. I have fixed in newer commit.

cartography/intel/gcp/compute.py
```diff
@@ -207,6 +213,136 @@ def get_gcp_vpcs(projectid: str, compute: Resource) -> Resource:
...
+    compute.vpnGateways.list permission), so that other Compute resources still
+    sync and cleanup is safely suppressed for the project.
+    :param projectid: The project ID
+    :param region: The region to pull VPN gateways from
+    :param compute: The compute resource object created by googleapiclient.discovery.build()
+    :return: Response object containing data on all GCP VPN gateways for a given project,
+        or None if the region must be skipped
+    """
+    try:
+        req = compute.vpnGateways().list(project=projectid, region=region)
```
kunaals: this makes a separate request chain for every region, and the tunnel sync repeats the same pattern. that adds dozens of compute api calls per project even though google has aggregated list endpoints for both resources. let's use those with partial success enabled and keep cleanup disabled whenever any returned scope is incomplete.

cartography/intel/gcp/compute.py
```diff
@@ -1351,9 +1836,135 @@ def sync_gcp_vpcs(
...
+    :param vpc_res: The return data from get_gcp_vpcs()
+    :param project_id: The project ID to sync
+    :param gcp_update_tag: The timestamp value to set our new Neo4j nodes with
+    :param common_job_parameters: dict of other job parameters to pass to Neo4j
+    :return: Nothing
+    """
+    peerings = transform_gcp_vpc_peerings(vpc_res)
+    # Create peer VPC stubs first so that PEER_NETWORK relationships resolve
+    # even when the peer project is not synced.
+    _create_vpc_stubs(
```
kunaals, line 1933: these peer stubs have no cleanup path when the other project isn’t synced. once the peering disappears, its vpc stub stays in the graph forever; the vpn gateway stubs below have the same problem. let's remove unowned stubs once nothing references them and add a two-sync test for it.

kunaals (changes requested):

cartography/intel/gcp/compute.py
bhargav2427: Ack! I’ve added the cleanup in the latest commit.

cartography/intel/gcp/compute.py
bhargav2427: I have fixed this.

cartography/intel/gcp/compute.py
```diff
@@ -207,6 +214,134 @@ def get_gcp_vpcs(projectid: str, compute: Resource) -> Resource:
...
+                    summarize_gcp_http_error(e),
+                )
+                incomplete = True
+                break
+            raise
+        # Aggregated responses key items by scope ("regions/{region}"). A scope
+        # is incomplete if it carries an `error` entry, or a `warning` with any
+        # code other than the benign NO_RESULTS_ON_PAGE (e.g. UNREACHABLE means
+        # the region could not be queried, so its data is missing).
+        for scope, scope_data in res.get("items", {}).items():
```
kunaals, line 327: the aggregated response can also report missing data through the top-level `unreachables` and `warning` fields, outside `items`. those currently leave `incomplete` false, so cleanup can still delete gateways or tunnels from an unreadable region. please mark the page incomplete when `unreachables` is non-empty or its top-level warning indicates partial results, and add one shared regression test.

kunaals (changes requested):
one more change, then this should be good to merge. thanks for the contribution!

cartography/intel/gcp/compute.py
bhargav2427: Yes! I missed that. Thank You so much!!

kunaals (approved):

PR #3180, opened 2026-08-24
feat(railway): model filesystem snapshots

kunaals:
### Type of change
- [ ] Documentation update

### Summary

Adds a Railway-specific filesystem snapshot node with the cross-provider `FilesystemSnapshot` ontology label. It represents the exact repository revision backing a current Git-based Railway deployment. The snapshot uses a namespaced Cartography ID and retains the Railway deployment ID separately for the `SCANNED_AS` relationship.

Current deployments link to snapshots with `SCANNED_AS`, snapshots link to matching GitHub repositories with `SNAPSHOT_OF`, and snapshots remain scoped to their Railway project. Snapshots are only created when Railway supplies both a repository and a full commit SHA; scoped cleanup removes stale snapshots.

### Related issues or links

- Depends on #3179

### How was this tested?

- `uv run pytest tests/unit/cartography/intel/ontology/test_ontology_mapping.py tests/unit/cartography/graph/test_ontology_rel_constraints.py tests/unit/cartography/models/test_schema_docs.py tests/integration/cartography/intel/railway -q` (72 passed)
- `make test_lint`
- `uv run --group doc ./docs/build.sh`

#### General

#### Proof of functionality

### Notes for reviewers

This PR is intentionally stacked on #3179, which adds the exact Railway source revision used here. Trivy result ingestion is outside this PR.

### DIFF
--- a/cartography/intel/railway/deployments.py
+++ b/cartography/intel/railway/deployments.py
@@ -10,6 +10,9 @@ from cartography.intel.railway.serviceinstances import iter_service_instances
 from cartography.intel.railway.utils import unwrap_edges
 from cartography.models.railway.deployment import RailwayDeploymentSchema
 from cartography.models.railway.deploymenttrigger import RailwayDeploymentTriggerSchema
+from cartography.models.railway.filesystem_snapshot import (
+    RailwayFilesystemSnapshotSchema,
+)
 from cartography.util import timeit
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


cartography/intel/railway/deployments.py
```diff
@@ -22,28 +25,36 @@ def sync(
...
         # window is a superseded revision.
-        current_ids = {
-            (instance.get("latestDeployment") or {}).get("id")
-            for instance in iter_service_instances(bundle)
-        }
-        current_ids.discard(None)
+        current_instances: dict[str, dict[str, Any]] = {}
+        for instance in iter_service_instances(bundle):
+            latest_deployment = instance.get("latestDeployment") or {}
+            if latest_deployment.get("id"):
```
jychp: P1: Do not treat `latestDeployment` as the currently serving revision without validating its status. Railway keeps the previous healthy deployment serving when the latest attempt fails, so this creates `SCANNED_AS` for inactive code and omits the live revision.

cartography/intel/railway/deployments.py
```diff
@@ -66,11 +77,27 @@ def transform(
...
+                current_instance = current_instances.get(deployment["id"])
+                source = (current_instance or {}).get("source") or {}
+                if source_revision and source.get("repo"):
+                    project_snapshots.append(
+                        {
+                            "id": f"railway:filesystem-snapshot:{deployment['id']}",
+                            "deployment_id": deployment["id"],
+                            "kind": "source",
+                            "source_revision": source_revision,
+                            "source_repo": source["repo"],
```
jychp: P1: Source the repository and root directory from immutable deployment metadata instead of the mutable service-instance configuration. Otherwise the same deployment snapshot can be rewritten with a new repository or root while retaining the old commit.

jychp (changes requested):

cartography/intel/railway/deployments.py
kunaals: good catch. this now falls back to the newest successful or sleeping deployment for the service when the latest attempt is not serving.

cartography/intel/railway/deployments.py
kunaals: good catch. snapshots now use the repository and root directory from immutable deployment metadata, so later service changes cannot rewrite their provenance.

cartography/intel/railway/deployments.py
```diff
@@ -22,30 +27,60 @@ def sync(
...
+            for deployment in unwrap_edges(environment["deployments"])
+        ]
+        serving_by_instance: dict[tuple[str, str], dict[str, Any]] = {}
+        for deployment in project_deployment_nodes:
+            if deployment.get("status") not in _SERVING_DEPLOYMENT_STATUSES:
+                continue
+            key = (deployment["environmentId"], deployment["serviceId"])
+            previous = serving_by_instance.get(key)
+            if previous is None or deployment["createdAt"] > previous["createdAt"]:
+                serving_by_instance[key] = deployment
```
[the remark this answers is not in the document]
kunaals: this now uses railway’s active deployment records directly, including when they fall outside the history window

jychp (approved):

PR #3181, opened 2026-08-24
feat(trivy): ingest filesystem scan reports

kunaals:
### Type of change
- [x] Documentation update

### Summary

Extends Trivy ingestion to accept source filesystem and repository reports in addition to container image reports.

Cartography matches a report to `FilesystemSnapshot` nodes using the repository URL and exact 40-character commit from Trivy metadata. Findings attach with `AFFECTS`, packages attach with `DEPLOYED`, and canonical `PackageVersion` relationships are propagated by the existing ontology package sync. One report can attach to multiple snapshots representing the same repository revision.

Existing image report matching and ingestion remain unchanged.

### Related issues or links

- Depends on #3180

### How was this tested?

- `uv run pytest tests/unit/cartography/intel/trivy tests/integration/cartography/intel/trivy tests/integration/cartography/intel/ontology/test_packages.py tests/unit/cartography/graph/test_ontology_rel_constraints.py tests/unit/cartography/models/test_schema_docs.py -q` (69 passed)
- `make test_lint`
- `uv run --group doc ./docs/build.sh`
- Trivy 0.69.3 against a normal clone with a public remote produced `ArtifactType: repository`, the public repository URL in `Metadata.RepoURL`, and the exact commit in `Metadata.Commit`.

#### General

#### Proof of functionality

### Notes for reviewers

Trivy emits Git metadata only when it can identify the scan target as a repository. The documentation therefore requires scanning the checked-out repository root while retaining its Git metadata. Reports without an exact graph match are skipped, and cleanup is skipped when no reports match.

### DIFF
--- a/cartography/intel/trivy/__init__.py
+++ b/cartography/intel/trivy/__init__.py
@@ -173,8 +173,8 @@ def _normalize_repository(value: str) -> str:
 
 def _get_filesystem_scan_targets(
     neo4j_session: Session,
-) -> dict[tuple[str, str], list[str]]:
-    """Return source repository and revision pairs mapped to snapshot IDs."""
+) -> dict[tuple[str, str, str | None], list[str]]:
+    """Return source repository, revision, and root mapped to snapshot IDs."""
     result = neo4j_session.run(
         """
         MATCH (snapshot:FilesystemSnapshot)
@@ -182,14 +182,18 @@ def _get_filesystem_scan_targets(
           AND snapshot.source_revision IS NOT NULL
         RETURN snapshot.id AS id,
                snapshot.source_repo AS repository,
-               snapshot.source_revision AS revision
+               snapshot.source_revision AS revision,
+               snapshot.root_directory AS root_directory
         """,
     )
-    targets: dict[tuple[str, str], list[str]] = {}
+    targets: dict[tuple[str, str, str | None], list[str]] = {}
     for record in result:
         repository = _normalize_repository(record["repository"])
         revision = record["revision"].lower()
-        targets.setdefault((repository, revision), []).append(record["id"])
+        root_directory = (record["root_directory"] or "").strip("/") or None
+        targets.setdefault((repository, revision, root_directory), []).append(
+            record["id"]
+        )
     return targets
 
 
@@ -209,7 +213,7 @@ def _prepare_trivy_data(
     trivy_data: dict[str, Any],
     image_uris: set[str],
     digest_aliases: dict[str, str],
-    filesystem_targets: dict[tuple[str, str], list[str]],
+    filesystem_targets: dict[tuple[str, str, str | None], list[str]],
     source: str,
 ) -> tuple[dict[str, Any], str | None, list[str]] | None:
     """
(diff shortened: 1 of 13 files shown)


cartography/intel/trivy/scanner.py
```diff
@@ -322,37 +328,94 @@ def sync_single_image(
     """
     try:
         _, results, image_digest = _parse_trivy_data(trivy_data, source)
-
-        # Transform all data in one pass
-        findings_list, packages_list, fixes_list = transform_scan_results(
+        num_findings = _sync_scan_results(
```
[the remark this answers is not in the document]
kunaals: good catch. fixed in 3e87f22d4.

cartography/intel/trivy/scanner.py
```diff
@@ -322,37 +328,94 @@ def sync_single_image(
...
+        if artifact_type not in {"filesystem", "repository"}:
+            raise ValueError(
+                f"Expected a filesystem or repository Trivy report for {source}",
+            )
+        if not filesystem_snapshot_ids:
+            raise ValueError(f"No filesystem snapshot targets for {source}")
+
+        results = trivy_data.get("Results", [])
+        if not results:
+            logger.info("No vulnerabilities found for %s", source)
```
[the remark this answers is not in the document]
kunaals: good catch. fixed in 3e87f22d4.

cartography/intel/trivy/__init__.py
```diff
@@ -205,14 +248,44 @@ def _prepare_trivy_data(
...
+    ):
+        repository_key = _normalize_repository(repo_url)
+        revision_key = revision.lower()
+        if isinstance(root_directory, str):
+            root_key = root_directory.strip("/") or None
+            snapshot_ids = filesystem_targets.get(
+                (repository_key, revision_key, root_key),
+                [],
+            )
+        else:
```
jychp: P1: Do not attach a repository-wide report to every root-specific snapshot. Trivy 0.69 does not emit `Metadata.RootDirectory`, so services using different monorepo roots at the same commit receive each other's packages and vulnerabilities.

cartography/intel/trivy/__init__.py
```diff
@@ -295,6 +381,14 @@ def sync_trivy_from_report_reader(
         )
         return
 
+    if image_uris and processed_image_reports == 0:
```
jychp, line 387: P1: Scope cleanup independently for image and filesystem targets in both directions. This guard protects image data during a filesystem-only batch, but an image-only batch still runs global cleanup and deletes valid filesystem scan data.

cartography/intel/trivy/scanner.py
```diff
@@ -424,6 +488,33 @@ def cleanup(neo4j_session: Session, common_job_parameters: dict[str, Any]) -> No
     )
 
 
+@timeit
+def cleanup_filesystem_snapshot_relationships(
```
jychp: P2: Remove Trivy entities that become unreferenced after filesystem relationship cleanup. A clean rescan currently deletes only snapshot edges, leaving stale findings, packages, fixes, and internal relationships indefinitely.

jychp (changes requested):

cartography/intel/trivy/__init__.py
kunaals: good catch. cleanup is now symmetric: image-only batches clean only image relationships, while filesystem-only batches clean only filesystem relationships.

jychp (approved):

PR #3187, opened 2026-08-24
feat(kubernetes): ingest GPU and persistent storage

kunaals:
### Type of change

- [x] Documentation update

### Summary

Expose Kubernetes GPU capacity, allocatable GPU count, product labels, and container GPU requests and limits without adding vendor-specific node types. Preserve complete resource request, limit, capacity, and allocatable maps as sorted JSON so other extended resources remain queryable.

Ingest StorageClasses, PersistentVolumes, and PersistentVolumeClaims. Connect pods to claims with `MOUNTS`, claims to volumes with `BOUND_TO`, and volumes and claims to storage classes with `USES_STORAGE_CLASS`. During the pre-1.0 permission migration, a 401 or 403 skips storage ingestion and cleanup so an upgraded deployment does not erase previously synced storage data.

### How was this tested?

- Added relevant tests 
- Ran the modified connector against a live Kubernetes 1.34 API and loaded the result into a new isolated Neo4j database. The redacted result was:

  ```text
  live Kubernetes sync completed
  {"claims": 0, "nodes": 2, "pods": 18, "storage_classes": 1, "volumes": 0}
  ```

- Loaded anonymized Kubernetes SDK response shapes into the isolated database to exercise the complete GPU pod to PVC to PV to StorageClass path:

  ```text
  {"claims": 1, "gpu_pods": 1, "max_gpu_limit": 8, "max_gpu_request": 8, "storage_classes": 1, "volumes": 1}
  ```

#### General

#### Proof of functionality

### Notes for reviewers

The fixtures use synthetic UIDs, example domains, and generic scheduler labels. They contain no customer names, cluster identifiers, storage handles, or credentials.

### DIFF
--- a/cartography/intel/kubernetes/pods.py
+++ b/cartography/intel/kubernetes/pods.py
@@ -8,8 +8,10 @@ from kubernetes.client.models import V1Pod
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.intel.kubernetes.util import format_resource_quantities
 from cartography.intel.kubernetes.util import get_controller_owner_reference
 from cartography.intel.kubernetes.util import get_epoch
+from cartography.intel.kubernetes.util import get_gpu_quantity
 from cartography.intel.kubernetes.util import k8s_paginate
 from cartography.intel.kubernetes.util import K8sClient
 from cartography.models.kubernetes.containers import KubernetesContainerSchema
@@ -103,6 +105,16 @@ def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[st
 
         # Extract resource requests and limits
         if container.resources:
+            requests = container.resources.requests or {}
+            limits = container.resources.limits or {}
+            containers[container.name]["resource_requests"] = (
+                format_resource_quantities(requests)
+            )
+            containers[container.name]["resource_limits"] = format_resource_quantities(
+                limits
+            )
+            containers[container.name]["gpu_request"] = get_gpu_quantity(requests)
+            containers[container.name]["gpu_limit"] = get_gpu_quantity(limits)
             if container.resources.requests:
                 containers[container.name]["memory_request"] = (
                     container.resources.requests.get("memory")
@@ -125,6 +137,10 @@ def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[st
                 containers[container.name]["memory_limit"] = None
                 containers[container.name]["cpu_limit"] = None
         else:
+            containers[container.name]["resource_requests"] = "{}"
+            containers[container.name]["resource_limits"] = "{}"
+            containers[container.name]["gpu_request"] = None
+            containers[container.name]["gpu_limit"] = None
             containers[container.name]["memory_request"] = None
             containers[container.name]["cpu_request"] = None
             containers[container.name]["memory_limit"] = None
@@ -287,6 +303,14 @@ def transform_pods(
         workload_parent = _resolve_pod_workload_parent(
             pod, rs_owner_map, workloads_available=workloads_available
         )
+        persistent_volume_claim_names = sorted(
+            {
+                volume.persistent_volume_claim.claim_name
+                for volume in (pod.spec.volumes or [])
+                if getattr(volume, "persistent_volume_claim", None)
+                and getattr(volume.persistent_volume_claim, "claim_name", None)
+            }
+        )
         transformed_pods.append(
             {
                 **workload_parent,
@@ -317,6 +341,11 @@ def transform_pods(
                         and getattr(volume.host_path, "path", None)
                     ]
                 ),
+                "persistent_volume_claim_names": persistent_volume_claim_names,
+                "persistent_volume_claim_ids": [
+                    f"{cluster_name}/{pod.metadata.namespace}/{claim_name}"
+                    for claim_name in persistent_volume_claim_names
+                ],
                 "service_account_id": (
                     f"{cluster_name}/{pod.metadata.namespace}/{service_account_name}"
                 ),
--- a/cartography/intel/kubernetes/pods.py
+++ b/cartography/intel/kubernetes/pods.py
@@ -303,14 +303,20 @@ def transform_pods(
         workload_parent = _resolve_pod_workload_parent(
             pod, rs_owner_map, workloads_available=workloads_available
         )
-        persistent_volume_claim_names = sorted(
-            {
-                volume.persistent_volume_claim.claim_name
-                for volume in (pod.spec.volumes or [])
-                if volume.persistent_volume_claim
+        persistent_volume_claim_name_set = set()
+        for volume in pod.spec.volumes or []:
+            if (
+                volume.persistent_volume_claim
                 and volume.persistent_volume_claim.claim_name
-            }
-        )
+            ):
+                persistent_volume_claim_name_set.add(
+                    volume.persistent_volume_claim.claim_name
+                )
+            elif volume.ephemeral:
+                persistent_volume_claim_name_set.add(
+                    f"{pod.metadata.name}-{volume.name}"
+                )
+        persistent_volume_claim_names = sorted(persistent_volume_claim_name_set)
         transformed_pods.append(
             {
                 **workload_parent,
--- a/cartography/intel/kubernetes/pods.py
+++ b/cartography/intel/kubernetes/pods.py
@@ -115,27 +115,10 @@ def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[st
             )
             containers[container.name]["gpu_request"] = get_gpu_quantity(requests)
             containers[container.name]["gpu_limit"] = get_gpu_quantity(limits)
-            if container.resources.requests:
-                containers[container.name]["memory_request"] = (
-                    container.resources.requests.get("memory")
-                )
-                containers[container.name]["cpu_request"] = (
-                    container.resources.requests.get("cpu")
-                )
-            else:
-                containers[container.name]["memory_request"] = None
-                containers[container.name]["cpu_request"] = None
-
-            if container.resources.limits:
-                containers[container.name]["memory_limit"] = (
-                    container.resources.limits.get("memory")
-                )
-                containers[container.name]["cpu_limit"] = (
-                    container.resources.limits.get("cpu")
-                )
-            else:
-                containers[container.name]["memory_limit"] = None
-                containers[container.name]["cpu_limit"] = None
+            containers[container.name]["memory_request"] = requests.get("memory")
+            containers[container.name]["cpu_request"] = requests.get("cpu")
+            containers[container.name]["memory_limit"] = limits.get("memory")
+            containers[container.name]["cpu_limit"] = limits.get("cpu")
         else:
             containers[container.name]["resource_requests"] = "{}"
             containers[container.name]["resource_limits"] = "{}"
--- a/cartography/intel/kubernetes/pods.py
+++ b/cartography/intel/kubernetes/pods.py
@@ -307,8 +307,8 @@ def transform_pods(
             {
                 volume.persistent_volume_claim.claim_name
                 for volume in (pod.spec.volumes or [])
-                if getattr(volume, "persistent_volume_claim", None)
-                and getattr(volume.persistent_volume_claim, "claim_name", None)
+                if volume.persistent_volume_claim
+                and volume.persistent_volume_claim.claim_name
             }
         )
         transformed_pods.append(
--- a/cartography/intel/kubernetes/storage.py
+++ b/cartography/intel/kubernetes/storage.py
@@ -66,9 +66,9 @@ def transform_persistent_volumes(
     for volume in volumes:
         spec = volume.spec
         status = volume.status
-        claim_ref = getattr(spec, "claim_ref", None)
-        csi = getattr(spec, "csi", None)
-        storage_class_name = getattr(spec, "storage_class_name", None)
+        claim_ref = spec.claim_ref
+        csi = spec.csi
+        storage_class_name = spec.storage_class_name
         transformed.append(
             {
                 "id": f"{cluster_name}/{volume.metadata.name}",
@@ -86,11 +86,11 @@ def transform_persistent_volumes(
                     else None
                 ),
                 "volume_mode": spec.volume_mode,
-                "phase": getattr(status, "phase", None),
-                "claim_namespace": getattr(claim_ref, "namespace", None),
-                "claim_name": getattr(claim_ref, "name", None),
-                "csi_driver": getattr(csi, "driver", None),
-                "csi_volume_handle": getattr(csi, "volume_handle", None),
+                "phase": status.phase if status else None,
+                "claim_namespace": claim_ref.namespace if claim_ref else None,
+                "claim_name": claim_ref.name if claim_ref else None,
+                "csi_driver": csi.driver if csi else None,
+                "csi_volume_handle": csi.volume_handle if csi else None,
                 "labels": json.dumps(volume.metadata.labels or {}, sort_keys=True),
             }
         )
@@ -107,7 +107,7 @@ def transform_persistent_volume_claims(
         namespace = claim.metadata.namespace
         storage_class_name = spec.storage_class_name
         volume_name = spec.volume_name
-        requests = getattr(spec.resources, "requests", None) or {}
+        requests = spec.resources.requests if spec.resources else {}
         transformed.append(
             {
                 "id": f"{cluster_name}/{namespace}/{claim.metadata.name}",
@@ -127,7 +127,7 @@ def transform_persistent_volume_claims(
                 "access_modes": sorted(spec.access_modes or []),
                 "requested_storage": requests.get("storage"),
                 "volume_mode": spec.volume_mode,
-                "phase": getattr(status, "phase", None),
+                "phase": status.phase if status else None,
                 "labels": json.dumps(claim.metadata.labels or {}, sort_keys=True),
             }
         )
--- a/cartography/intel/kubernetes/storage.py
+++ b/cartography/intel/kubernetes/storage.py
@@ -107,7 +107,7 @@ def transform_persistent_volume_claims(
         namespace = claim.metadata.namespace
         storage_class_name = spec.storage_class_name
         volume_name = spec.volume_name
-        requests = spec.resources.requests if spec.resources else {}
+        requests = (spec.resources.requests if spec.resources else None) or {}
         transformed.append(
             {
                 "id": f"{cluster_name}/{namespace}/{claim.metadata.name}",
--- a/cartography/intel/kubernetes/storage.py
+++ b/cartography/intel/kubernetes/storage.py
@@ -45,6 +45,7 @@ def transform_storage_classes(
     return [
         {
             "id": f"{cluster_name}/{storage_class.metadata.name}",
+            "uid": storage_class.metadata.uid,
             "name": storage_class.metadata.name,
             "creation_timestamp": get_epoch(storage_class.metadata.creation_timestamp),
             "deletion_timestamp": get_epoch(storage_class.metadata.deletion_timestamp),
@@ -68,6 +69,8 @@ def transform_persistent_volumes(
         status = volume.status
         claim_ref = spec.claim_ref
         csi = spec.csi
+        csi_driver = csi.driver if csi else None
+        csi_volume_handle = csi.volume_handle if csi else None
         storage_class_name = spec.storage_class_name
         transformed.append(
             {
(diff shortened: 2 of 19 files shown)


cartography/models/kubernetes/storage.py
```diff
@@ -0,0 +1,354 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class KubernetesStorageClassNodeProperties(CartographyNodeProperties):
```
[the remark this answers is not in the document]
kunaals: fixed- StorageClass, PersistentVolume, and PersistentVolumeClaim nodes now ingest creation and deletion timestamps as epoch values, with integration coverage for both fields

cartography/intel/kubernetes/storage.py
```diff
@@ -0,0 +1,227 @@
...
+                "cluster %s (status %s). Skipping storage sync and preserving "
+                "previously synced StorageClass/PersistentVolume/"
+                "PersistentVolumeClaim nodes. Grant list on storageclasses, "
+                "persistentvolumes, and persistentvolumeclaims; these permissions "
+                "will be required in v1.0.0.",
+                client.name,
+                error.status,
+            )
+            return
+        raise
```
jychp, line 257: Replace this `raise` with a log + `return`.

`sync_storage` runs before `sync_pods` in `start_k8s_ingestion`, and the loop there re-raises whatever it catches. So a single transient 5xx or timeout on a storage list call skips pods, secrets, services, network policies, ingress and the exposure jobs for this cluster, then stops every cluster left in the loop.

`sync_workloads` faces the same hazard and handles it: log an error and `return`, so load and cleanup are skipped and previously synced nodes are preserved. Do the same here.

cartography/intel/kubernetes/util.py
```diff
@@ -12,13 +15,36 @@
...
+    return json.dumps(
+        {str(key): str(value) for key, value in (resources or {}).items()},
+        sort_keys=True,
+    )
+
+
+def get_gpu_quantity(resources: dict[str, Any] | None) -> int | None:
+    quantities = []
+    for key, value in (resources or {}).items():
+        if not str(key).endswith("/gpu"):
```
jychp: **question**: `endswith("/gpu")` misses MIG (`nvidia.com/mig-1g.10gb`) and Intel (`gpu.intel.com/i915`), and blindly sums heterogeneous vendor keys into one integer. Confirm that is intended and state the limitation in the `gpu_capacity`/`gpu_request` descriptions.

cartography/intel/kubernetes/pods.py
```diff
@@ -287,6 +303,14 @@ def transform_pods(
         workload_parent = _resolve_pod_workload_parent(
             pod, rs_owner_map, workloads_available=workloads_available
         )
+        persistent_volume_claim_names = sorted(
```
jychp: Generic ephemeral volumes (`volume.ephemeral`) also produce a PVC, named `<pod>-<volume>`. Those pods get no `MOUNTS` edge today. Is it on purpose ?

cartography/models/kubernetes/storage.py
```diff
@@ -0,0 +1,378 @@
...
+    )
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "USES_STORAGE_CLASS"
+    properties: KubernetesPersistentVolumeToStorageClassRelProperties = (
+        KubernetesPersistentVolumeToStorageClassRelProperties()
+    )
+
+
+@dataclass(frozen=True)
+class KubernetesPersistentVolumeSchema(CartographyNodeSchema):
```
jychp, line 257: Can we give the `KubernetesPersistentVolume` the ontology storage labels: `ExtraNodeLabels` with `BLOCK_STORAGE` / `FILE_STORAGE` conditioned on `volume_mode`. Every other provider's disk/volume node carries them (AWS EBS, Azure disk, Scaleway, Railway), and without them PVs stay invisible to cross-provider storage queries.

cartography/models/kubernetes/storage.py
```diff
@@ -0,0 +1,378 @@
...
+    claim_name: PropertyRef = PropertyRef(
+        "claim_name",
+        description="Name of the bound PersistentVolumeClaim, when present.",
+    )
+    csi_driver: PropertyRef = PropertyRef(
+        "csi_driver",
+        extra_index=True,
+        description="CSI driver that manages the volume, when the volume uses CSI.",
+    )
+    csi_volume_handle: PropertyRef = PropertyRef(
```
jychp, line 166: Can we add a relationship from `KubernetesPersistentVolume` to the backing cloud disk. `csi_volume_handle` is the EBS volume id (`vol-...`) for the AWS EBS CSI driver and the disk resource id on Azure/GCP; resolving it at ingest is the code-to-cloud edge this data is worth having.

cartography/models/kubernetes/storage.py
```diff
@@ -0,0 +1,378 @@
...
+from cartography.models.core.relationships import CartographyRelProperties
+from cartography.models.core.relationships import CartographyRelSchema
+from cartography.models.core.relationships import LinkDirection
+from cartography.models.core.relationships import make_target_node_matcher
+from cartography.models.core.relationships import OtherRelationships
+from cartography.models.core.relationships import TargetNodeMatcher
+
+
+@dataclass(frozen=True)
+class KubernetesStorageClassNodeProperties(CartographyNodeProperties):
```
jychp, line 15: **nitpick**: Add `uid` to `KubernetesStorageClass`. `KubernetesPersistentVolume` and `KubernetesPersistentVolumeClaim` both carry it; the omission is unexplained.

cartography/models/kubernetes/containers.py
```diff
@@ -83,6 +83,23 @@ class KubernetesContainerNodeProperties(CartographyNodeProperties):
...
+    resource_limits: PropertyRef = PropertyRef(
+        "resource_limits",
+        description="All container resource limits, including extended resources, stored as a JSON-encoded object.",
+    )
+    gpu_request: PropertyRef = PropertyRef(
+        "gpu_request",
+        extra_index=True,
+        description="Total integer request across extended resource keys ending in `/gpu`.",
+    )
+    gpu_limit: PropertyRef = PropertyRef(
```
jychp, line 99: Add `extra_index=True` to `gpu_limit`, and to `gpu_allocatable` on the node schema. The query you ship in `queries.md` filters on `container.gpu_limit`, so indexing only the `request`/`capacity` half is arbitrary.

docs/root/modules/kubernetes/queries.md
```diff
@@ -17,3 +17,22 @@ RETURN k.name, k.api_server_url, k.kubeconfig_tls_configuration_status,
...
 ORDER BY k.name;
 ```
+
+## Map GPU workloads to persistent storage
+
+Find GPU-requesting containers, their scheduled nodes, and the persistent storage
+mounted by their pods:
+
+```cypher
+MATCH (container:KubernetesContainer)<-[:CONTAINS]-(pod:KubernetesPod)
```
jychp: Use `MATCH (container:KubernetesContainer)-[:WORKLOAD_PARENT]->(pod:KubernetesPod)`. The pod-to-container `CONTAINS` edge is marked deprecated and slated for removal in v1.0.0; do not ship a new doc example built on it.

jychp (changes requested):

cartography/intel/kubernetes/storage.py
kunaals: good catch. storage api failures now skip load and cleanup, preserve previously synced storage nodes, and allow the remaining kubernetes modules to continue.

cartography/intel/kubernetes/util.py
kunaals: good catch. gpu quantities now include full-gpu, nvidia mig, and intel scheduling resources. the descriptions clarify that heterogeneous scheduling units are summed and are not normalized to physical gpus.

cartography/intel/kubernetes/pods.py
kunaals: good catch. generic ephemeral volumes now map to their generated `<pod>-<volume>` pvc name, so the pod gets the expected `MOUNTS` relationship.

cartography/models/kubernetes/storage.py
kunaals: keeping this as-is because `volumeMode` describes how kubernetes presents the volume, not the backing storage type. an ebs volume commonly uses `Filesystem`, so applying `FILE_STORAGE` from that value would misclassify it.

cartography/models/kubernetes/storage.py
kunaals: added `BACKED_BY` relationships for aws ebs and azure disk csi handles. i left gcp out because cartography does not currently have a gcp persistent-disk node to target.

cartography/models/kubernetes/storage.py
kunaals: good catch. added the storageclass uid to the transform and schema with integration coverage.

cartography/models/kubernetes/containers.py
kunaals: added the `gpu_limit` index because the shipped query filters on it. i left `gpu_capacity` and `gpu_allocatable` unindexed because that query only returns them, so those indexes would not support its access path.

docs/root/modules/kubernetes/queries.md
kunaals: good catch. updated the example to use `KubernetesContainer-[:WORKLOAD_PARENT]->KubernetesPod`.

cartography/intel/kubernetes/storage.py
```diff
@@ -0,0 +1,251 @@
...
+        storage_classes = transform_storage_classes(
+            get_storage_classes(client), client.name
+        )
+        volumes = transform_persistent_volumes(
+            get_persistent_volumes(client), client.name
+        )
+        claims = transform_persistent_volume_claims(
+            get_persistent_volume_claims(client), client.name
+        )
+    except ApiException as error:
```
jychp, line 232: Only half of the failure mode I flagged is covered here. `k8s_paginate` and this `except` both catch `ApiException` only, but the kubernetes client converts just `urllib3.exceptions.SSLError` into `ApiException(status=0)` (`kubernetes/client/rest.py`). A connect refused, a DNS failure or a read timeout on any of the three list calls raises `urllib3.exceptions.MaxRetryError` / `ReadTimeoutError`, which walks straight through this handler, up to `start_k8s_ingestion`, which re-raises. We then lose pods, secrets, services, network policies, ingress and the exposure jobs for this cluster, and every remaining cluster in the loop, which is exactly the outcome this block is meant to prevent.

cartography/intel/kubernetes/storage.py
```diff
@@ -0,0 +1,251 @@
...
+                "cluster %s (status %s). Skipping storage sync and preserving "
+                "previously synced StorageClass/PersistentVolume/"
+                "PersistentVolumeClaim nodes. Grant list on storageclasses, "
+                "persistentvolumes, and persistentvolumeclaims; these permissions "
+                "will be required in v1.0.0.",
+                client.name,
+                error.status,
+            )
+            return
+        logger.error(
```
jychp: This makes `sync_storage` the only Kubernetes sync that swallows every status. `sync_secrets` (`secrets.py:123`) and `sync_network_policies` (`networkpolicies.py:287`) re-raise anything that is not 401/403.

My earlier request was about a transient failure not taking the whole cluster loop down with it; generalising to all statuses goes further than that. The cost is that an apiserver returning 500 in a loop yields a sync that reports success while the storage subgraph silently goes stale, run after run, with only an `error` log line to show for it.

cartography/intel/kubernetes/storage.py
```diff
@@ -0,0 +1,251 @@
...
+                    if csi_driver == "ebs.csi.aws.com"
+                    and csi_volume_handle
+                    and csi_volume_handle.startswith("vol-")
+                    else None
+                ),
+                "azure_disk_id": (
+                    csi_volume_handle
+                    if csi_driver == "disk.csi.azure.com"
+                    and csi_volume_handle
+                    and csi_volume_handle.lower().startswith("/subscriptions/")
```
jychp, line 121: The guard is case-insensitive but the join it feeds is not. `azure_disk_id` ends up matched verbatim against `AzureDisk.id` in `KubernetesPersistentVolumeToAzureDiskRel`, and `AzureDisk.id` always carries the ARM canonical casing (`intel/azure/compute.py:222` does `x.index("resourceGroups")`, so it would raise otherwise).

Statically provisioned PVs routinely carry a hand-written handle such as `/subscriptions/.../resourcegroups/...`. Those pass this `.lower()` check, get written to the node, and then never form a `BACKED_BY` edge. The failure is silent.

Either normalize both sides of the match, or drop the `.lower()` here so the property is only set for handles that can actually resolve, and say so in the `azure_disk_id` description.

cartography/models/kubernetes/storage.py
```diff
@@ -0,0 +1,428 @@
...
+        "deletion_timestamp",
+        description="Timestamp when the PersistentVolume was marked for deletion.",
+    )
+    cluster_name: PropertyRef = PropertyRef(
+        "CLUSTER_NAME",
+        set_in_kwargs=True,
+        extra_index=True,
+        description="Name of the Kubernetes cluster containing this PersistentVolume.",
+    )
+    capacity_storage: PropertyRef = PropertyRef(
```
jychp, line 125: `capacity_storage` and `requested_storage` keep the raw Kubernetes quantity string (`"2Pi"`), so no size comparison is possible in Cypher. Every other block-storage node in the graph exposes a numeric size (`AWSEBSVolume.size` in GiB, `AzureDisk.disksizegb`), and "find oversized or undersized volumes" is one of the main reasons to ingest this at all.

Suggest keeping the raw string and adding a parsed `capacity_storage_bytes` / `requested_storage_bytes` int alongside it.

cartography/models/kubernetes/storage.py
```diff
@@ -0,0 +1,428 @@
...
+        "id",
+        description="Identifier derived from cluster name and StorageClass name.",
+    )
+    uid: PropertyRef = PropertyRef(
+        "uid", description="Kubernetes UID of the StorageClass."
+    )
+    name: PropertyRef = PropertyRef(
+        "name", extra_index=True, description="Name of the StorageClass."
+    )
+    creation_timestamp: PropertyRef = PropertyRef(
```
jychp, line 26: Epoch ints match the rest of the Kubernetes module, so this is consistent locally, but the repo has been moving to native Neo4j temporal types: databricks, circleci, bbot, snowflake and `gcp/apikeys` all parse to `datetime`, and `docs/root/dev/writing-intel-modules.md:138` points the same way. Native datetimes are what make temporal operators usable in queries.

Three brand-new node types are a decent place to start doing it right, at the cost of a mixed convention inside `intel/kubernetes` until the existing nodes follow. Not blocking, but let's make the call deliberately rather than inherit the epoch by default.

cartography/intel/kubernetes/pods.py
```diff
@@ -103,6 +105,16 @@ def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[st
...
+            limits = container.resources.limits or {}
+            containers[container.name]["resource_requests"] = (
+                format_resource_quantities(requests)
+            )
+            containers[container.name]["resource_limits"] = format_resource_quantities(
+                limits
+            )
+            containers[container.name]["gpu_request"] = get_gpu_quantity(requests)
+            containers[container.name]["gpu_limit"] = get_gpu_quantity(limits)
             if container.resources.requests:
```
jychp: nit: The `requests` and `limits` locals computed just above already hold these values; the branches below re-read `container.resources.requests` / `.limits` and re-test them. Using the locals collapses this into `memory_request = requests.get("memory")` and friends, with no separate `else` branch needed.

docs/root/modules/kubernetes/config.md
```diff
@@ -61,6 +64,15 @@ rules:
     - pods
     - services
     - serviceaccounts
+    - persistentvolumes
```
jychp, line 73: nit: The pre-v1.0.0 tolerance note only sits above the `storage.k8s.io` block, but the handler in `sync_storage` covers all three resources at once: a missing `list persistentvolumes` behaves exactly like a missing `list storageclasses`. As written, an operator reading the "Core resources - list only" block has no way to know these two verbs are also tolerated during the migration window. Worth moving the note, or repeating it here.

tests/integration/cartography/intel/kubernetes/test_storage.py
```diff
@@ -0,0 +1,348 @@
...
+        (
+            "worker",
+            8,
+            8,
+            '{"cpu": "32", "memory": "600Gi", "nvidia.com/gpu": "8"}',
+        )
+    }
+
+
+@pytest.mark.parametrize(
```
jychp: Two coverage gaps worth closing:

1. This parametrization only covers `ApiException`. The transport exceptions raised by urllib3 (see my comment on `storage.py:218`) are the ones that actually escape the handler, and nothing here would catch that regression.
2. `test_transform_persistent_volume_claim_without_requests` only sets `spec.resources.requests = None`. The transform also guards `spec.resources` being None outright, which no test exercises.

jychp (changes requested):

cartography/intel/kubernetes/storage.py
```diff
@@ -0,0 +1,271 @@
...
+
+def transform_storage_classes(
+    storage_classes: list[V1StorageClass], cluster_name: str
+) -> list[dict[str, Any]]:
+    return [
+        {
+            "id": f"{cluster_name}/{storage_class.metadata.name}",
+            "uid": storage_class.metadata.uid,
+            "name": storage_class.metadata.name,
+            "creation_timestamp": storage_class.metadata.creation_timestamp,
```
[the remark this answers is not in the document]
kunaals: keeping native datetimes here. this was a deliberate change. these are new node types with no compatibility migration, and native neo4j temporal values support temporal operators. the existing kubernetes epoch fields can be migrated separately rather than extending that legacy representation

tests/integration/cartography/intel/kubernetes/test_storage.py
```diff
@@ -0,0 +1,410 @@
...
+    for label, name in (
+        ("KubernetesStorageClass", STORAGE_CLASS_NAME),
+        ("KubernetesPersistentVolume", VOLUME_NAME),
+        ("KubernetesPersistentVolumeClaim", CLAIM_NAME),
+    ):
+        assert check_nodes(
+            neo4j_session,
+            label,
+            ["name", "creation_timestamp", "deletion_timestamp"],
+        ) == {(name, CREATION_TIMESTAMP, DELETION_TIMESTAMP)}
```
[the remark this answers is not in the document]
kunaals: this divergence is deliberate for the three new schemas. we have documented that we now want native datetimes, and these nodes dont need backwards compat - they're new. the remaining kubernetes schemas can be migrated from epoch values separately

cartography/intel/kubernetes/storage.py
kunaals: good catch. storage sync now catches urllib3 transport failures, including connection, dns, and read-timeout failures, skips load and cleanup, and preserves prior nodes so the remaining modules and clusters can continue. added integration coverage with a concrete `MaxRetryError`.

cartography/intel/kubernetes/storage.py
kunaals: agreed on narrowing this. 401/403, status 0, 429, 5xx, and transport failures remain isolated because they represent permission or transient failures that must not stop the cluster loop. other api statuses now re-raise instead of being silently swallowed.

cartography/intel/kubernetes/storage.py
kunaals: good catch. the `BACKED_BY` target matcher now compares azure resource ids case-insensitively. the integration test covers a csi handle whose casing differs from the canonical `AzureDisk.id`.

cartography/models/kubernetes/storage.py
kunaals: added `capacity_storage_bytes` and `requested_storage_bytes` using the kubernetes quantity parser, while retaining the original quantity strings. the integration test covers `2Pi`.

cartography/models/kubernetes/storage.py
kunaals: agreed. these new schemas now store the typed kubernetes datetimes directly, so temporal comparisons work without requiring a compatibility migration for existing nodes.

cartography/intel/kubernetes/pods.py
kunaals: done. the memory and cpu fields now use the normalized `requests` and `limits` mappings directly.

docs/root/modules/kubernetes/config.md
kunaals: good catch. the persistent volume, persistent volume claim, and storage class permissions are now grouped under the same migration-tolerance note.

jychp (approved):

PR #3188, opened 2026-08-25
feat(kubernetes): map container PVC access

kunaals:
### Type of change

- [x] Documentation update

### Summary

Map Kubernetes PVC use at the level where each API field is declared.

- Pod `REFERENCES` relationships identify claims referenced by pod volumes.
- Container `MOUNTS` relationships and `persistent_volume_claim_mounts` preserve filesystem mount configuration.
- Container `USES_BLOCK_DEVICE` relationships and `persistent_volume_claim_devices` preserve raw block device mappings.

All relationships are part of the typed Pod and Container schemas, so normal ingestion and cleanup manage them without MatchLinks or a separate relationship-loading path.

### Related issues or links

- Depends on #3187

### How was this tested?

- `make test_lint`
- `uv run --frozen pytest -q tests/unit/cartography/intel/kubernetes tests/integration/cartography/intel/kubernetes` (130 passed)
- `uv run ./docs/build.sh`

The integration tests load sanitized Kubernetes API model objects through the normal storage and pod sync paths. They verify Pod references, container filesystem mounts, raw block device properties and relationships, and stale-edge cleanup after access is removed.

#### General

#### Proof of functionality

### Notes for reviewers

This PR is intentionally stacked on #3187 because it connects Pods and containers to the PersistentVolumeClaim nodes introduced there.

### DIFF
--- a/cartography/intel/kubernetes/pods.py
+++ b/cartography/intel/kubernetes/pods.py
@@ -21,6 +21,10 @@ from cartography.util import timeit
 logger = logging.getLogger(__name__)
 
 
+def _claim_id(cluster_name: str, namespace: str, claim_name: str) -> str:
+    return f"{cluster_name}/{namespace}/{claim_name}"
+
+
 def _get_claim_name_by_volume_name(pod: V1Pod) -> dict[str, str]:
     claims = {}
     for volume in pod.spec.volumes or []:
@@ -58,19 +62,23 @@ def _extract_pod_containers(
             "container_ports": json.dumps([]),
             "container_port_numbers": [],
             "persistent_volume_claim_ids": [],
+            "persistent_volume_claim_read_write_ids": [],
             "persistent_volume_claim_mounts": "[]",
             "persistent_volume_claim_device_ids": [],
             "persistent_volume_claim_devices": "[]",
         }
 
         claim_ids = set()
+        read_write_claim_ids = set()
         mount_details = []
         for mount in container.volume_mounts or []:
             claim_name = claim_by_volume_name.get(mount.name)
             if not claim_name:
                 continue
-            claim_id = f"{cluster_name}/{pod.metadata.namespace}/{claim_name}"
+            claim_id = _claim_id(cluster_name, pod.metadata.namespace, claim_name)
             claim_ids.add(claim_id)
+            if not mount.read_only:
+                read_write_claim_ids.add(claim_id)
             mount_details.append(
                 {
                     "claim_id": claim_id,
@@ -94,6 +102,9 @@ def _extract_pod_containers(
             )
         )
         containers[container.name]["persistent_volume_claim_ids"] = sorted(claim_ids)
+        containers[container.name]["persistent_volume_claim_read_write_ids"] = sorted(
+            read_write_claim_ids
+        )
         containers[container.name]["persistent_volume_claim_mounts"] = json.dumps(
             mount_details,
             sort_keys=True,
@@ -105,7 +116,7 @@ def _extract_pod_containers(
             claim_name = claim_by_volume_name.get(device.name)
             if not claim_name:
                 continue
-            claim_id = f"{cluster_name}/{pod.metadata.namespace}/{claim_name}"
+            claim_id = _claim_id(cluster_name, pod.metadata.namespace, claim_name)
             device_claim_ids.add(claim_id)
             device_details.append(
                 {
@@ -411,7 +422,7 @@ def transform_pods(
                 ),
                 "persistent_volume_claim_names": persistent_volume_claim_names,
                 "persistent_volume_claim_ids": [
-                    f"{cluster_name}/{pod.metadata.namespace}/{claim_name}"
+                    _claim_id(cluster_name, pod.metadata.namespace, claim_name)
                     for claim_name in persistent_volume_claim_names
                 ],
                 "service_account_id": (
--- a/cartography/intel/kubernetes/pods.py
+++ b/cartography/intel/kubernetes/pods.py
@@ -59,6 +59,8 @@ def _extract_pod_containers(
             "container_port_numbers": [],
             "persistent_volume_claim_ids": [],
             "persistent_volume_claim_mounts": "[]",
+            "persistent_volume_claim_device_ids": [],
+            "persistent_volume_claim_devices": "[]",
         }
 
         claim_ids = set()
@@ -94,6 +96,37 @@ def _extract_pod_containers(
         containers[container.name]["persistent_volume_claim_ids"] = sorted(claim_ids)
         containers[container.name]["persistent_volume_claim_mounts"] = json.dumps(
             mount_details,
+            sort_keys=True,
+        )
+
+        device_claim_ids = set()
+        device_details = []
+        for device in getattr(container, "volume_devices", None) or []:
+            claim_name = claim_by_volume_name.get(device.name)
+            if not claim_name:
+                continue
+            claim_id = f"{cluster_name}/{pod.metadata.namespace}/{claim_name}"
+            device_claim_ids.add(claim_id)
+            device_details.append(
+                {
+                    "claim_id": claim_id,
+                    "claim_name": claim_name,
+                    "volume_name": device.name,
+                    "device_path": device.device_path,
+                }
+            )
+        device_details.sort(
+            key=lambda detail: (
+                detail["claim_id"],
+                detail["device_path"],
+                detail["volume_name"],
+            )
+        )
+        containers[container.name]["persistent_volume_claim_device_ids"] = sorted(
+            device_claim_ids
+        )
+        containers[container.name]["persistent_volume_claim_devices"] = json.dumps(
+            device_details,
             sort_keys=True,
         )
 
--- a/cartography/intel/kubernetes/pods.py
+++ b/cartography/intel/kubernetes/pods.py
@@ -21,7 +21,22 @@ from cartography.util import timeit
 logger = logging.getLogger(__name__)
 
 
-def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[str, Any]:
+def _get_claim_name_by_volume_name(pod: V1Pod) -> dict[str, str]:
+    claims = {}
+    for volume in pod.spec.volumes or []:
+        if volume.persistent_volume_claim and volume.persistent_volume_claim.claim_name:
+            claims[volume.name] = volume.persistent_volume_claim.claim_name
+        elif volume.ephemeral:
+            claims[volume.name] = f"{pod.metadata.name}-{volume.name}"
+    return claims
+
+
+def _extract_pod_containers(
+    pod: V1Pod,
+    cluster_name: str,
+    claim_by_volume_name: dict[str, str],
+    node_arch: str | None = None,
+) -> dict[str, Any]:
     pod_containers: list[V1Container] = pod.spec.containers
     containers = dict()
     for container in pod_containers:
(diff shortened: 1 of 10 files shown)


cartography/models/kubernetes/containers.py
```diff
@@ -335,6 +351,44 @@ class KubernetesContainerToGitHubContainerImageRel(CartographyRelSchema):
     )
 
 
+@dataclass(frozen=True)
+class KubernetesContainerToPersistentVolumeClaimRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesContainerToPersistentVolumeClaimRel(CartographyRelSchema):
```
jychp, line 356: blocking: add a user-facing docstring. Every other relationship schema in this file has one and the generated schema docs render it, so this edge ships with an empty description.

cartography/models/kubernetes/containers.py
```diff
@@ -335,6 +351,44 @@ class KubernetesContainerToGitHubContainerImageRel(CartographyRelSchema):
...
+
+@dataclass(frozen=True)
+class KubernetesContainerToPersistentVolumeClaimDeviceRelProperties(
+    CartographyRelProperties
+):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class KubernetesContainerToPersistentVolumeClaimDeviceRel(CartographyRelSchema):
```
jychp, line 378: blocking: add a user-facing docstring here too, describing raw block device access.

cartography/models/kubernetes/containers.py
```diff
@@ -101,6 +101,22 @@ class KubernetesContainerNodeProperties(CartographyNodeProperties):
         extra_index=True,
         description="Total GPU scheduling-unit limit across full-GPU, NVIDIA MIG, and Intel GPU resource keys; heterogeneous units are not normalized to physical GPUs.",
     )
+    persistent_volume_claim_ids: PropertyRef = PropertyRef(
```
jychp: suggestion: drop `persistent_volume_claim_ids` and `persistent_volume_claim_device_ids` from the node properties. They duplicate the `MOUNTS` / `USES_BLOCK_DEVICE` edges and the `claim_id` already present in each JSON entry; `KubernetesPodSchema` deliberately keeps its matcher list out of the node properties and only stores `persistent_volume_claim_names`. Keep the keys in the data dict for matching only.

cartography/models/kubernetes/containers.py
```diff
@@ -101,6 +101,22 @@ class KubernetesContainerNodeProperties(CartographyNodeProperties):
...
+        "persistent_volume_claim_ids",
+        description="Identifiers of PersistentVolumeClaims mounted by the container.",
```
jychp: nitpick: cite the source API fields in the descriptions (`container.volumeMounts[]`, `container.volumeDevices[]`) the way `allow_privilege_escalation` and `host_path_volume_paths` do.

cartography/intel/kubernetes/pods.py
```diff
@@ -42,8 +57,79 @@ def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[st
...
+        )
+        containers[container.name]["persistent_volume_claim_ids"] = sorted(claim_ids)
+        containers[container.name]["persistent_volume_claim_mounts"] = json.dumps(
+            mount_details,
+            sort_keys=True,
+        )
+
+        device_claim_ids = set()
+        device_details = []
+        for device in getattr(container, "volume_devices", None) or []:
```
jychp: suggestion: make attribute access consistent. `container.volume_mounts` is read directly (which forced adding `volume_mounts=None` to four test fixtures) while `volume_devices` goes through `getattr`. Pick one: `getattr` for both, or add both fields to the fixtures and drop the `getattr`.

cartography/intel/kubernetes/pods.py
```diff
@@ -42,8 +57,79 @@ def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[st
...
+                continue
+            claim_id = f"{cluster_name}/{pod.metadata.namespace}/{claim_name}"
+            claim_ids.add(claim_id)
+            mount_details.append(
+                {
+                    "claim_id": claim_id,
+                    "claim_name": claim_name,
+                    "volume_name": mount.name,
+                    "mount_path": mount.mount_path,
+                    "read_only": bool(mount.read_only),
```
jychp, line 88: suggestion: `read_only` is the security-relevant field and it is only reachable by parsing JSON client-side, so no Cypher rule can distinguish read-only from read-write PVC access. Emit a queryable list too (e.g. `persistent_volume_claim_read_write_ids`), or carry `read_only` on the `MOUNTS` edge.

cartography/intel/kubernetes/pods.py
```diff
@@ -42,8 +57,79 @@ def _extract_pod_containers(pod: V1Pod, node_arch: str | None = None) -> dict[st
...
+            "persistent_volume_claim_devices": "[]",
         }
 
+        claim_ids = set()
+        mount_details = []
+        for mount in container.volume_mounts or []:
+            claim_name = claim_by_volume_name.get(mount.name)
+            if not claim_name:
+                continue
+            claim_id = f"{cluster_name}/{pod.metadata.namespace}/{claim_name}"
```
jychp: nitpick: `f"{cluster_name}/{namespace}/{claim_name}"` is now built in three places (twice here, once in `transform_pods`). Extract a `_claim_id()` helper so the id format has a single definition.

cartography/intel/kubernetes/pods.py
```diff
@@ -21,7 +21,22 @@
...
+    return claims
+
+
+def _extract_pod_containers(
+    pod: V1Pod,
+    cluster_name: str,
+    claim_by_volume_name: dict[str, str],
+    node_arch: str | None = None,
+) -> dict[str, Any]:
     pod_containers: list[V1Container] = pod.spec.containers
```
jychp, line 44: question: PVC mounts declared by `initContainers` and `ephemeralContainers` produce no edges, yet `_extract_pod_secrets` in this same file scans both. An init container doing read-write data prep on a PVC is invisible to any container-level query. State the gap in the module docs, or extract mounts for those containers too.

tests/integration/cartography/intel/kubernetes/test_storage.py
```diff
@@ -260,6 +296,204 @@ def test_pod_mounts_persistent_volume_claim(
...
+        "name",
+        "KubernetesPersistentVolumeClaim",
+        "name",
+        "USES_BLOCK_DEVICE",
+    )
+    neo4j_session.run(
+        """
+        MATCH (node)
+        WHERE node.uid IN $uids
+        DETACH DELETE node
```
jychp: suggestion: move this teardown into a fixture finalizer. `neo4j_session` is module-scoped and only wipes at module teardown, so a failing assertion above skips this delete and leaks `block-*` nodes into `test_sync_pods_cleans_up_removed_container_mount`, which asserts on exact relationship sets and would then fail for an unrelated reason.

tests/integration/cartography/intel/kubernetes/test_storage.py
```diff
@@ -260,6 +296,204 @@ def test_pod_mounts_persistent_volume_claim(
...
+    # Assert
+    assert ("block-training-job", "block-data") in check_rels(
+        neo4j_session,
+        "KubernetesPod",
+        "name",
+        "KubernetesPersistentVolumeClaim",
+        "name",
+        "REFERENCES",
+    )
+    assert ("block-worker", "block-data") in check_rels(
```
jychp: suggestion: assert on the full set with `==` instead of `in` / `not in`. Membership checks pass even if the sync produced extra unexpected edges.

docs/root/modules/kubernetes/index.md
```diff
@@ -29,9 +39,9 @@ policies use `APPLIES_TO` edges to identify selected pods.
 Until v1.0.0, if the identity cannot list persistent volumes, persistent volume
 claims, or storage classes, Cartography skips persistent storage ingestion and
 cleanup and preserves existing storage nodes. Pods continue to load. On a first
-sync, they have no `MOUNTS` relationships. After an earlier successful storage
-sync, pods can link to the preserved storage snapshot, which may be stale until
-permissions are restored.
+sync, they have no `REFERENCES` relationships. After an earlier successful
```
jychp: suggestion: this degraded-permission paragraph now only mentions pod `REFERENCES`. Say that container `MOUNTS` and `USES_BLOCK_DEVICE` edges are equally absent on a first sync.

docs/root/modules/kubernetes/queries.md
```diff
@@ -21,20 +21,25 @@ ORDER BY k.name;
...
+  (claim:KubernetesPersistentVolumeClaim)
 OPTIONAL MATCH (claim)-[:BOUND_TO]->(volume:KubernetesPersistentVolume)
 OPTIONAL MATCH (volume)-[:BACKED_BY]->(cloud_disk)
 OPTIONAL MATCH (claim)-[:USES_STORAGE_CLASS]->(storage_class:KubernetesStorageClass)
 RETURN pod.namespace, pod.name, container.name,
        container.gpu_request, container.gpu_limit,
        node.name, node.gpu_product, node.gpu_capacity,
-       claim.name, volume.name, volume.csi_driver,
+       type(storage_access), claim.name,
+       container.persistent_volume_claim_mounts,
```
jychp: nitpick: these two columns return the container's full JSON blob covering every claim, so on a per-claim row the JSON does not correspond to `claim.name`. Either drop them or note that they cover all of the container's claims.

jychp (approved):
approved with some comments

cartography/models/kubernetes/containers.py
kunaals: added a user-facing relationship docstring describing filesystem pvc mounts.

cartography/models/kubernetes/containers.py
kunaals: added a user-facing relationship docstring describing raw block device access.

cartography/models/kubernetes/containers.py
kunaals: done. the mount and device claim-id lists remain in the transform data only for direct relationship matching and are no longer persisted as node properties.

cartography/models/kubernetes/containers.py
kunaals: done. the property descriptions now identify `container.volumeMounts[]` and `container.volumeDevices[]` as their source fields.

cartography/intel/kubernetes/pods.py
kunaals: added `persistent_volume_claim_read_write_ids` as a queryable property. a claim is included when at least one of its mounts is read-write, while the full per-mount details remain in the json property.

cartography/intel/kubernetes/pods.py
kunaals: done. pod references, filesystem mounts, and block devices now use a shared `_claim_id()` helper.

cartography/intel/kubernetes/pods.py
kunaals: documented that container storage relationships currently cover regular application containers in `spec.containers`. init and ephemeral containers are not yet modeled as `KubernetesContainer` nodes.

tests/integration/cartography/intel/kubernetes/test_storage.py
kunaals: moved the block-storage cleanup into a fixture finalizer, so it runs even when an assertion fails.

tests/integration/cartography/intel/kubernetes/test_storage.py
kunaals: changed these to exact equality checks scoped to the block test’s pod and container. unrelated module fixtures remain allowed, while any extra edge from the test resources now fails the assertion.

docs/root/modules/kubernetes/index.md
kunaals: updated the degraded-permission section to cover missing pod `REFERENCES` and container `MOUNTS` and `USES_BLOCK_DEVICE` relationships.

docs/root/modules/kubernetes/queries.md
kunaals: dropped the container-wide json blobs from the per-claim result. the query now returns a claim-specific `read_write` value for `MOUNTS` and `null` for raw block-device access.

PR #3193, opened 2026-08-26
feat(notion): add identity and public page inventory

kunaals:
### Type of change
- [x] Documentation update

### Summary

Adds Notion as a first-class SaaS connector using one standard Notion connection token per workspace. Cartography discovers each workspace from `/v1/users/me`, inventories people and bot connections from `/v1/users`, and maps them to the `Tenant`, `UserAccount`, and `ThirdPartyApp` ontologies.

The optional `sync_public_pages` mode adds metadata-only `NotionPage` nodes for connection-visible pages whose `public_url` shows they are published to the web. It stores identifiers, title, URLs, timestamps, parent metadata, and creator relationships, but never page bodies, blocks, comments, database rows, or property values.

Notion Search is explicitly non-authoritative. The page sync processes Search responses in bounded batches and does not issue a request per missing page. Search omissions never trigger cleanup; a previously observed page is deleted only when a complete, valid pagination pass explicitly reports `public_url: null`. Explicit unpublish deletions are staged on disk until pagination completes, keeping both memory use and destructive cleanup safe for high-cardinality workspaces.

Configuration is validated before graph writes, tokens are never logged, retries are bounded and respect `Retry-After`, and identity cleanup runs only after pagination completes successfully. Guests, roles, account state, groups, and workspace-wide content enumeration are not available from this API surface and are not inferred.

### Related issues or links

- [Notion list users API](https://developers.notion.com/reference/get-users)
- [Notion retrieve token bot API](https://developers.notion.com/reference/get-self)
- [Notion Search limitations](https://developers.notion.com/reference/search-optimizations-and-limitations)
- [Notion page object](https://developers.notion.com/reference/page)
- [Notion request limits](https://developers.notion.com/reference/request-limits)

### How was this tested?

- `make test_lint` — passed
- `make test_unit` — 5,493 passed
- `make test_integration` — 1,655 passed
- Focused Notion tests — 64 unit and 7 integration tests passed
- `uv run ./docs/build.sh` — succeeded

Tests cover configuration and workspace discovery, strict GET/POST response-envelope validation, cursor pagination, malformed and non-progressing responses, person/bot transforms, workspace-scoped identity, ontology labels, relationships, cleanup isolation, bounded public-page ingestion, explicit unpublish cleanup, Search omissions, and pagination-failure cleanup suppression.

Read-only provider validation used an isolated temporary Neo4j database and a temporary Notion connection token. Workspace discovery succeeded. A two-page Search pass completed with 143 connection-visible unpublished pages and 0 public pages; the graph contained one workspace and no page nodes, as expected. Sanitized logs:

```text
INFO:cartography.intel.notion.pages:Starting Notion public page sync
INFO:cartography.intel.notion.pages:Loaded 0 public Notion pages and observed 143 unpublished pages
INFO:cartography.intel.notion.pages:Completed Notion public page sync
```

The full CLI run failed closed at `/v1/users` with `403 restricted_resource`: the temporary connection did not have Notion's Read user information capability. No identity cleanup ran. The `needs-tests` label should remain until a connection with that capability completes the identity inventory. No response bodies, identities, workspace identifiers, content, or credentials are included in this evidence.

#### General

#### Proof of functionality

- [ ] Tested the complete connector against a real cloud, SaaS, or provider environment.
- [ ] Included sanitized Cartography logs demonstrating a complete successful synchronization.

### Notes for reviewers

This is intentionally a standard-token, read-only connector. Public-page discovery is opt-in and metadata-only because the regular API does not provide an authoritative enterprise content export. The branch includes a signed merge of current `origin/master` and was validated after that merge.


cartography/intel/notion/pages.py
```diff
@@ -0,0 +1,216 @@
...
+        parent_notion_id = (
+            None
+            if parent_type == "workspace"
+            else require_nonempty_string(
+                parent.get(parent_type),
+                "Notion page parent id",
+            )
+        )
+
+        if public_url is None:
```
jychp: **blocking** - Move this unpublished check above the metadata validation on lines 76-109. Only the ID is used for unpublished pages, yet a missing or renamed `created_time`, `url`, `in_trash`, `is_locked`, `created_by`, `parent` or `properties` on any one of them raises and aborts the entire page sync, including every valid public page staged so far. The provider run in the PR description saw 143 unpublished pages and 0 public ones, so this is the dominant code path.

cartography/intel/notion/pages.py
```diff
@@ -0,0 +1,216 @@
...
+            page.get("created_time"),
+            "Notion page created_time",
+        )
+        last_edited_time = require_nonempty_string(
+            page.get("last_edited_time"),
+            "Notion page last_edited_time",
+        )
+        url = require_nonempty_string(page.get("url"), "Notion page url")
+        in_trash = require_boolean(page.get("in_trash"), "Notion page in_trash")
+        is_locked = require_boolean(page.get("is_locked"), "Notion page is_locked")
```
jychp: **suggestion** - Accept `in_trash` and `is_locked` as optional booleans. Both are informational fields that no relationship or cleanup decision depends on, so requiring them turns a Notion field rename into a total page-sync failure.

cartography/intel/notion/pages.py
```diff
@@ -0,0 +1,216 @@
...
+            continue
+
+        public_pages.append(
+            {
+                "id": scoped_id(workspace_id, notion_page_id),
+                "notion_page_id": notion_page_id,
+                "title": _get_title(properties),
+                "url": url,
+                "public_url": public_url,
+                "is_public": True,
```
jychp: **suggestion** - Drop `is_public`, or derive it from `public_url`. Only pages with a non-null `public_url` reach this branch, so the property is a hardcoded `True` that can never answer the question its name poses. Same problem with `is_workspace_member` in `users.py:62`.

cartography/intel/notion/pages.py
```diff
@@ -0,0 +1,216 @@
...
+            load_pages(neo4j_session, json.loads(line), workspace_id, update_tag)
+
+        staged_deletes.seek(0)
+        delete_batch: list[str] = []
+        for line in staged_deletes:
+            delete_batch.append(json.loads(line))
+            if len(delete_batch) == 10_000:
+                delete_confirmed_unpublished_pages(neo4j_session, delete_batch)
+                delete_batch = []
+        delete_confirmed_unpublished_pages(neo4j_session, delete_batch)
```
jychp: **question** - `NotionPage` has no cleanup job, and a page hard-deleted in Notion never comes back in Search with `public_url: null`, so its node survives every future sync with `is_public: True`. Confirm that unbounded retention is intended and state the retention bound in `docs/root/modules/notion/config.md`; consumers otherwise have to infer staleness from `lastupdated` on their own.

cartography/intel/notion/util.py
```diff
@@ -0,0 +1,237 @@
...
+        response = api_session.get(
+            f"{NOTION_API_BASE_URL}/{endpoint}",
+            params=params,
+            timeout=REQUEST_TIMEOUT,
+        )
+        response.raise_for_status()
+        page_results, has_more, next_cursor_value = _validate_paginated_payload(
+            response.json(),
+            expected_type,
+        )
```
jychp: **blocking** - Call `_warn_if_incomplete` here as well. `get_paginated` feeds `users.sync`, which runs a destructive `GraphJob` cleanup immediately after loading, so an incomplete `/v1/users` response deletes live identities with no warning. The completeness signal is only wired into `post_paginated`, which is the path that does not run cleanup.

cartography/intel/notion/util.py
```diff
@@ -0,0 +1,237 @@
...
+        "Notion paginated response request_status",
+    )
+    status_type = require_nonempty_string(
+        request_status.get("type"),
+        "Notion paginated response request_status type",
+    )
+    if status_type == "complete":
+        return
+    if status_type != "incomplete":
+        raise ValueError("Notion paginated response has an unknown request status")
```
jychp: **suggestion** - Align the two branches. An unrecognized `request_status.type` hard-fails the sync while the known `incomplete` value only logs a warning, so a future Notion status value is treated as stricter than a status Notion already documents as partial. Pick one policy for both.

cartography/intel/notion/util.py
```diff
@@ -0,0 +1,237 @@
...
+def require_object(value: Any, field: str) -> dict[str, Any]:
+    if not isinstance(value, dict):
+        raise ValueError(f"{field} must be an object")
+    return value
+
+
+def require_nonempty_string(value: Any, field: str) -> str:
+    if not isinstance(value, str) or not value.strip():
+        raise ValueError(f"{field} must be a non-empty string")
+    return value
```
jychp: **nitpick** - Return `value.strip()`. The guard strips before checking but the return value does not, so a padded Notion ID passes validation and produces a padded `scoped_id`.

cartography/models/ontology/mapping/data/thirdpartyapps.py
```diff
@@ -165,6 +165,27 @@
...
 
+notion_mapping = OntologyMapping(
+    module_name="notion",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="NotionBot",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="client_id",
+                    node_field="id",
```
jychp: **suggestion** - Map `client_id` to `notion_user_id`, not `id`. `id` is the workspace-scoped composite (`workspace-1/bot-1`), so `_ont_client_id` would carry a Cartography-internal key rather than Notion's bot UUID, unlike `SlackBot` above which maps the provider's own identifier.

cartography/models/ontology/mapping/data/tenants.py
```diff
@@ -449,6 +449,23 @@
...
+    module_name="notion",
+    nodes=[
+        OntologyNodeMapping(
+            node_label="NotionWorkspace",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="name",
+                    node_field="name",
+                    required=True,
+                ),
```
jychp: **nitpick** - Add the `# status: Not available` and `# domain: Not available` comments that every neighboring mapping carries. Do the same for the unmapped `enabled` / `native_app` / `protocol` in `thirdpartyapps.py` and `firstname` / `lastname` / `has_mfa` / `active` / `lastactivity` in `useraccounts.py`.

cartography/models/ontology/mapping/data/useraccounts.py
```diff
@@ -427,6 +427,23 @@
...
+    nodes=[
+        OntologyNodeMapping(
+            node_label="NotionUser",
+            fields=[
+                OntologyFieldMapping(
+                    ontology_field="email",
+                    node_field="email",
+                    required=True,
+                ),
+                OntologyFieldMapping(ontology_field="username", node_field="email"),
```
jychp: **question** - Explain why `username` is mapped to `email` rather than left unmapped. Notion has no username concept, and duplicating the email into both `_ont_email` and `_ont_username` makes the two ontology fields indistinguishable for Notion accounts.

PR #3198, opened 2026-08-27
feat(slack): ingest accessible private channels

kunaals:
### Type of change
- [x] Documentation update

### Summary
Cartography currently relies on the `conversations.list` default, which only requests public channels. This change explicitly requests public and private channels so private channels accessible to the configured token are ingested. It also documents that bot tokens can only discover private channels the app has joined and that organization-wide discovery requires the Enterprise Admin Conversations API.

### Related issues or links
- [Slack `conversations.list` documentation](https://docs.slack.dev/reference/methods/conversations.list/)
- [Slack `groups:read` scope](https://docs.slack.dev/reference/scopes/groups.read/)

#### General

#### Proof of functionality

#### Intel connector validation
- [ ] Tested the connector against a real Slack environment.
- [ ] Included sanitized Cartography sync logs demonstrating successful synchronization.

### Notes for reviewers
Live-provider validation is not included yet. The API request contract is covered by a unit test, and the existing Slack ingestion path passes against a disposable Neo4j integration environment.


cartography/intel/slack/channels.py
```diff
@@ -34,6 +34,7 @@ def get(
         "channels",
         team_id=team_id,
         exclude_archived=True,
+        types="public_channel,private_channel",
```
jychp: Adding groups:read without a fallback breaks existing OAuth installs because Slack scopes are granted at install time and existing tokens are not upgraded automatically. For those workspaces, auth_conversations_list() will fail with invalid_types/missing_scope, and because the error is uncaught, the entire Slack ingestion aborts — including public channels, usergroups, and users. Catch SlackApiError, log a warning, and retry with types="public_channel" so existing tenants keep working without requiring immediate re-consent.

jychp: **blocking** — This breaks every already-installed workspace of a distributed OAuth app. Slack grants scopes at install time and they are not retroactive: adding `groups:read` to the app manifest does not upgrade tokens issued to existing tenants, each one has to be sent back through the OAuth flow (and re-approved by a workspace admin where app approval is enforced). `teams.get_teams()` calling `auth_teams_list()` shows this module is already used with org-wide installs, so the multi-tenant path is a supported one. Without the fallback in the sibling comment, this ships as a fleet-wide outage that only re-consent fixes.

jychp: **question** — Justify making this unconditional instead of gating it behind a config flag like the existing `--slack-channels-memberships`. Every existing deployment starts ingesting private channel names, topics, purposes and membership edges on upgrade with no opt-in.

docs/root/modules/slack/config.md
```diff
@@ -20,6 +20,12 @@ Add these bot scopes:
 - `users:read`
 - `users:read.email`
 
+Cartography requests both public and private channels. Slack only returns private
```
jychp, line 23: **suggestion** — State that `groups:read` must be granted at install time and is not retroactive: any app already installed without it has to go back through the OAuth flow before this sync works. For distributed apps that means re-consent for every tenant, so it belongs here rather than being discovered at runtime.

tests/unit/cartography/intel/slack/test_channels.py
```diff
@@ -0,0 +1,24 @@
+from unittest.mock import Mock
```
jychp, line 2: **nitpick** — Add `tests/unit/cartography/intel/slack/__init__.py`. Without it pytest imports this file as a top-level `test_channels` module, which collides with any future unit test of the same basename in another non-package directory.

tests/unit/cartography/intel/slack/test_channels.py
```diff
@@ -0,0 +1,24 @@
...
+@patch("cartography.intel.slack.channels.slack_paginate", return_value=[])
+def test_get_requests_public_and_private_channels(mock_slack_paginate):
+    # Arrange
+    slack_client = Mock()
+
+    # Act
+    channels = get(slack_client, "T123", False)
+
+    # Assert
+    assert channels == []
```
jychp, line 20: **suggestion** — Cover the actual behavior: add a private channel to `tests/data/slack/channels.py` and assert it loads with `is_private = true` in `tests/integration/cartography/intel/slack/test_channels.py`, plus a case where `conversations_list` raises `SlackApiError("invalid_types")`. This test only mirrors the kwargs literal and `assert channels == []` is tautological given the mock, so it proves nothing about private channels reaching the graph and nothing about the scope-less path.

jychp (dismissed):

cartography/intel/slack/channels.py
kunaals: good catch. added a narrow fallback for missing_scope and invalid_types so unrelated slack errors still surface.

docs/root/modules/slack/config.md
kunaals: documented install-time scope grants, reinstallation, and the public-only fallback

tests/unit/cartography/intel/slack/test_channels.py
kunaals: added the package marker

tests/unit/cartography/intel/slack/test_channels.py
kunaals: added private-channel graph coverage plus scope and type fallback tests

cartography/intel/slack/channels.py
```diff
@@ -28,13 +32,33 @@ def get(
...
+    try:
+        channel_data = slack_paginate(
+            slack_client,
+            "conversations_list",
+            "channels",
+            team_id=team_id,
+            exclude_archived=True,
+            types="public_channel,private_channel",
+        )
+    except SlackApiError as error:
```
jychp: **blocking** — The fallback returns a partial dataset, but `sync()` still runs the unconditional `cleanup()`, so every private `SlackChannel` ingested by a previous run is deleted by the scoped `GraphJob`. Return a completeness flag from `get()` (`-> tuple[list[dict[str, Any]], bool]`) and skip `cleanup()` when the private-channel fetch was degraded, matching the existing pattern in `cartography/intel/databricks/permissions.py` and `cartography/intel/snowflake/artifact_repositories.py`.

cartography/intel/slack/channels.py
```diff
@@ -28,13 +32,33 @@ def get(
...
+        error_code = error.response.get("error")
+        if error_code not in {"invalid_types", "missing_scope"}:
+            raise
+        logger.warning(
```
jychp: **suggestion** — A token missing `channels:read` also returns `missing_scope`, so this logs "syncing public channels only" and then dies on the retry with a chained `SlackApiError`. Move the warning after the fallback call succeeds and re-raise with `from error` so the original failure is not buried.

cartography/intel/slack/channels.py
```diff
@@ -28,13 +32,33 @@ def get(
...
+            "Slack token cannot list private channels (%s); syncing public channels only.",
+            error_code,
+        )
+        channel_data = slack_paginate(
+            slack_client,
+            "conversations_list",
+            "channels",
+            team_id=team_id,
+            exclude_archived=True,
+            types="public_channel",
```
jychp: **suggestion** — Collapse the two identical `slack_paginate` blocks into a loop over `("public_channel,private_channel", "public_channel")` so the kwargs cannot drift apart.

tests/integration/cartography/intel/slack/test_channels.py
```diff
@@ -44,15 +44,20 @@ def test_load_slack_channels(neo4j_session):
...
     expected_nodes = {
-        ("SLACKCHANNEL2", "random"),
-        ("SLACKCHANNEL1", "concern-marketing-comm"),
+        ("SLACKCHANNEL1", "concern-marketing-comm", False),
+        ("SLACKCHANNEL2", "random", False),
+        ("SLACKCHANNEL3", "private-channel", True),
     }
-    assert check_nodes(neo4j_session, "SlackChannel", ["id", "name"]) == expected_nodes
+    assert (
+        check_nodes(neo4j_session, "SlackChannel", ["id", "name", "is_private"])
```
jychp: **suggestion** — Add an integration test for the degraded path: sync once with private channels, then sync again with `conversations_list` raising `missing_scope`, and assert what the contract is for `SLACKCHANNEL3` after cleanup. `tests/AGENTS.md` requires partial-failure paths to prove the cleanup outcome across two update tags.

tests/data/slack/channels.py
```diff
@@ -86,6 +86,20 @@
             ],
             "num_members": 2,
         },
+        {
+            "id": "SLACKCHANNEL3",
```
jychp: **nitpick** — Shape this fixture like a real `conversations.list` private-channel payload: add `is_channel`, `is_group: True`, `is_member: True`, `name_normalized`, and `updated`. `is_member: True` in particular documents that Slack only returns private channels the app has joined.

tests/unit/cartography/intel/slack/test_channels.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@pytest.mark.parametrize("error_code", ["invalid_types", "missing_scope"])
+@patch("cartography.intel.slack.channels.slack_paginate")
+def test_get_falls_back_to_public_channels(mock_slack_paginate, error_code):
+    # Arrange
+    slack_client = Mock()
+    public_channel = {"id": "C123", "name": "public-channel"}
+    mock_slack_paginate.side_effect = [
+        SlackApiError("Failed to list channels", {"error": error_code}),
```
jychp: **nitpick** — Pass a real `slack_sdk.web.SlackResponse` instead of a bare dict so the test proves `error.response.get("error")` works against the type the SDK actually raises with.

docs/root/modules/slack/config.md
```diff
@@ -20,6 +20,15 @@ Add these bot scopes:
...
+channels that the app has been added to, so invite the app to each private
+channel you want to ingest. Slack grants `groups:read` when the app is installed;
```
jychp: **nitpick** — Reword "Slack grants `groups:read` when the app is installed": it reads as if Slack grants the scope automatically. State that scopes take effect only at install time and are not applied retroactively to existing tokens. Also state here what happens to already-ingested private channels when the scope is lost.

jychp (approved):

PR #3202, opened 2026-08-27
feat(introspection): expose inherited materialized ontology relationships

jychp:
### Type of change
- [ ] Documentation update

### Summary
`inspect_data_model()` already exposed nodes, relationships, semantic labels, and validation-only `RelConstraint`s. Downstream consumers still had to reinvent inheritance of materialized ontology edges onto concrete carriers.

This adds:
- `OntologyExpectedEdge` for relationships actually produced under ontology labels
- `relationships_for_node()` / `NodeRelationship` so concretes like `AWSECSContainer` inherit `RESOLVED_IMAGE → Image` with the far endpoint kept semantic
- Schema-docs rendering that uses that node-centric view
- Clear docs that validation-only constraints are not inherited onto concretes

### Related issues or links

- Follows typed analysis jobs / ontology constraint work; unblocks SubImage agent schema views that previously missed `RESOLVED_IMAGE` on container concretes.

### How was this tested?

- `uv run pytest tests/unit/cartography/models/test_introspection.py tests/unit/cartography/models/test_schema_docs.py -q`
- Verified `relationships_for_node("AWSECSContainer")` yields outward `RESOLVED_IMAGE → Image`
- Verified validation-only pairs (`ServiceAccount`/`HAS_ROLE`, `SCANNED_AS`) are not inherited

#### General

#### Proof of functionality

### Notes for reviewers
Key contract: `ontology_relationship_constraints` remain validation-only; only `ontology_expected_edges` / `model.relationships` feed inheritance. Consumers should use `relationships_for_node(label)` instead of reimplementing semantic-label projection.

### DIFF
--- a/cartography/models/schema_docs.py
+++ b/cartography/models/schema_docs.py
@@ -10,6 +10,7 @@ from cartography.models.core.nodes import LabelKind
 from cartography.models.core.relationships import LinkDirection
 from cartography.models.introspection import DataModel
 from cartography.models.introspection import Node
+from cartography.models.introspection import NodeRelationship
 from cartography.models.introspection import Property
 from cartography.models.introspection import Relationship
 from cartography.models.introspection import relationship_touches_node
@@ -51,7 +52,6 @@ def render_module_schema(model: DataModel, module: str) -> str:
         raise ValueError(f'No nodes found for module "{module}".')
 
     module_relationships = module_model.relationships
-    assigned_relationships = _assign_relationships(module_relationships, module_nodes)
     module_node_labels = {node.label for node in module_nodes}
     diagram_relationships = tuple(
         relationship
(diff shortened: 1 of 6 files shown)


cartography/models/schema_docs.py
```diff
@@ -74,7 +74,9 @@ def render_module_schema(model: DataModel, module: str) -> str:
     lines.extend(["```", ""])
 
     for node in module_nodes:
-        lines.extend(_render_node(node, assigned_relationships.get(node.label, ())))
+        lines.extend(
+            _render_node(node, module_model.relationships_for_node(node.label))
```
kunaals, line 78: i might be missing something, but `module_model` appears to have already filtered out materialized ontology edges unless the module carries both semantic endpoints. aws carries both `Container` and `Image`, so this test passes, while `KubernetesContainer` and `RailwayDeployment` still omit `RESOLVED_IMAGE → Image` from their generated schemas. should we preserve these expected edges when either endpoint is carried and add a test?

kunaals (approved):
approving with one thing to note

PR #3203, opened 2026-08-27
fix(aws-ssm): make public parameter catalogs opt-in

kunaals:
### Type of change

### Summary

Make AWS-managed public SSM parameter ingestion opt-in instead of crawling the full Bottlerocket and EKS optimized AMI namespaces during every default AWS sync.

These shared namespaces can contain tens of thousands of parameters per region, while `GetParametersByPath` returns at most 10 parameters per request. Their cost therefore grows with namespace history and enabled regions even when a deployment does not use the resulting catalog. Explicit prefix selection through the existing CLI option or environment variable continues to work unchanged.

### Related issues or links

- [AWS `GetParametersByPath` API](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParametersByPath.html)

### Breaking changes

Public SSM parameter ingestion is now disabled when neither `--aws-ssm-public-parameter-prefix-allowlist` nor `AWS_SSM_PUBLIC_PARAMETER_PREFIX_ALLOWLIST` is set. Existing deployments that rely on the previous Bottlerocket and EKS optimized AMI defaults must configure those prefixes explicitly before upgrading. Without an explicit allowlist, the next AWS sync removes previously ingested `AWSPublicSSMParameter` nodes through the existing cleanup path.

### How was this tested?

- Updated existing tests

#### General

#### Proof of functionality

### Notes for reviewers

This is a draft to confirm that opt-in is the right compatibility tradeoff. A sanitized live-provider run is still required before marking it ready.


tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -882,10 +882,7 @@ def test_start_aws_ingestion_defaults_public_ssm_allowlist_when_unset(
     cartography.intel.aws.start_aws_ingestion(neo4j_session, test_config)
 
     common_job_parameters = mock_perform_analysis.call_args.args[2]
-    assert (
-        common_job_parameters["aws_ssm_public_parameter_prefix_allowlist"]
-        == cartography.intel.aws.ssm_intel.DEFAULT_PUBLIC_PARAMETER_PREFIX_ALLOWLIST
-    )
+    assert common_job_parameters["aws_ssm_public_parameter_prefix_allowlist"] == ""
```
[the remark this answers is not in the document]
kunaals: good catch. fixed this so the unset case is isolated from the runner environment.

cartography/intel/aws/ssm.py
```diff
@@ -26,9 +26,7 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_PUBLIC_PARAMETER_PREFIX_ALLOWLIST = (
-    "/aws/service/bottlerocket/,/aws/service/eks/optimized-ami/"
-)
+DEFAULT_PUBLIC_PARAMETER_PREFIX_ALLOWLIST = ""
```
jychp: `_sync_shared_public_ssm_parameters` (cartography/intel/aws/__init__.py:295) still builds a boto3 session per profile and calls `_autodiscover_account_regions` (one `ec2:DescribeRegions` per account) before `sync_public_parameters` bails on the empty allowlist, so the new default adds N wasted API calls to every sync and defeats part of the stated goal. Resolve the allowlist first and skip the profile/region loop when it is empty, while still calling `sync_public_parameters` so the legacy label migration and cleanup run.

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -852,7 +852,7 @@ def test_start_aws_ingestion(
     assert shared_call_args[6] == test_config.aws_best_effort_mode
 
 
-@mock.patch.dict(os.environ, {}, clear=False)
+@mock.patch.dict(os.environ, {}, clear=True)
```
jychp: `clear=True` wipes the entire process environment, not just the allowlist variable. Scope the isolation to the one variable with `monkeypatch.delenv("AWS_SSM_PUBLIC_PARAMETER_PREFIX_ALLOWLIST", raising=False)`; clearing everything also drops `HOME`, `AWS_*`, and `CARTOGRAPHY_*` and will diverge from real runs as soon as `start_aws_ingestion` reads another variable.

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -852,7 +852,7 @@ def test_start_aws_ingestion(
...
 @mock.patch("cartography.intel.aws.aioboto3.Session")
 @mock.patch("cartography.intel.aws.boto3.Session")
 @mock.patch("cartography.intel.aws.organizations")
```
jychp, line 898: `test_start_aws_ingestion` above has the same env leak this test was just fixed for: it passes no allowlist to `Config`, so the assertions at lines 840 and 849 resolve `AWS_SSM_PUBLIC_PARAMETER_PREFIX_ALLOWLIST` from the runner and fail spuriously when it is set. It also compares the resolved value to `DEFAULT_PUBLIC_PARAMETER_PREFIX_ALLOWLIST`, which passes for any constant value. Isolate the variable there too and assert the literal `""`.

docs/root/modules/aws/config.md
```diff
@@ -153,11 +153,13 @@ The configuration priority is:
 
 1. `--aws-ssm-public-parameter-prefix-allowlist`
 2. `AWS_SSM_PUBLIC_PARAMETER_PREFIX_ALLOWLIST`
-3. The built-in defaults: `/aws/service/bottlerocket/` and
-   `/aws/service/eks/optimized-ami/`
+3. An empty allowlist
```
jychp: `3. An empty allowlist` is not a configuration source. Drop the third list item and state that ingestion stays disabled unless 1 or 2 is set.

jychp (changes requested):

tests/integration/cartography/intel/aws/test_init.py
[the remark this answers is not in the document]
kunaals: good catch. fixed this by isolating the sibling test from the allowlist environment variable and asserting the literal default.

cartography/intel/aws/ssm.py
kunaals: good catch. the disabled path now normalizes the allowlist before profile and region discovery, then calls `sync_public_parameters` with no region sessions so migration and cleanup still run.

tests/integration/cartography/intel/aws/test_init.py
kunaals: good catch. replaced the whole-environment clear with `monkeypatch.delenv` for just this variable.

tests/integration/cartography/intel/aws/test_init.py
kunaals: good catch. isolated the sibling test too and changed its assertions to the literal empty default.

docs/root/modules/aws/config.md
kunaals: agreed. removed the third list item and stated that ingestion is disabled when neither source is set.

cartography/intel/aws/__init__.py
```diff
@@ -304,6 +304,17 @@ def _sync_shared_public_ssm_parameters(
     if "ssm" not in requested_syncs:
         return
 
+    if not ssm_intel._normalize_allowlisted_prefixes(
```
jychp: Dont call the private `ssm_intel._normalize_allowlisted_prefixes` from this module. `sync_public_parameters` re-derives the same emptiness test with `_minimize_allowlisted_prefixes(_normalize_allowlisted_prefixes(...))`, so the enabled/disabled rule now has two owners in two modules and will drift. Expose one public predicate in `ssm.py` (e.g. `public_parameter_sync_enabled(common_job_parameters)`) and call it from both sites.

cartography/intel/aws/__init__.py
```diff
@@ -304,6 +304,17 @@ def _sync_shared_public_ssm_parameters(
...
+        common_job_parameters.get("aws_ssm_public_parameter_prefix_allowlist"),
+    ):
+        ssm_intel.sync_public_parameters(
```
jychp: suggestion: Log a warning before this call. Every existing deployment lands on this path on upgrade and loses all `AWSPublicSSMParameter` nodes via `cleanup_public_ssm_parameters`, and the only signal today is a `logger.info` inside `sync_public_parameters`. Name `--aws-ssm-public-parameter-prefix-allowlist` in the message so operators know how to restore them.

cartography/intel/aws/ssm.py
jychp: Delete `DEFAULT_PUBLIC_PARAMETER_PREFIX_ALLOWLIST` and return `""` directly from `_resolve_aws_ssm_public_parameter_prefix_allowlist`. Its single remaining caller returns it verbatim, and a constant named `DEFAULT_..._ALLOWLIST` bound to the empty string reads as a leftover rather than a default.

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -152,14 +155,51 @@ def test_sync_shared_public_ssm_parameters_does_not_mask_unexpected_errors(
...
             aws_best_effort_mode=True,
         )
 
     mock_sync_public_parameters.assert_called_once()
 
 
+@mock.patch.object(cartography.intel.aws.ssm_intel, "sync_public_parameters")
+@mock.patch.object(cartography.intel.aws, "_autodiscover_account_regions")
+@mock.patch.object(cartography.intel.aws, "_get_boto3_session_for_profile")
+def test_sync_shared_public_ssm_parameters_skips_aws_calls_when_disabled(
```
jychp, line 171: Add coverage for the destructive half of this change. Nothing tests that an unset allowlist deletes previously ingested `AWSPublicSSMParameter` nodes; this test only asserts a mock call, and the `if not allowlist_prefixes` branch in `sync_public_parameters` has no test at all. Put a real test in `tests/integration/cartography/intel/aws/test_ssm.py`: load public parameters, re-sync with an empty allowlist, assert the nodes are gone.

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -152,14 +155,51 @@ def test_sync_shared_public_ssm_parameters_does_not_mask_unexpected_errors(
...
+    mock_get_session,
+    mock_autodiscover_regions,
+    mock_sync_public_parameters,
+    neo4j_session,
+):
+    # Arrange
+    common_job_parameters = {
+        "UPDATE_TAG": TEST_UPDATE_TAG,
+        "aws_ssm_public_parameter_prefix_allowlist": ",",
```
jychp: Parametrize this over `None`, `""`, and `","`. `","` only exercises prefix normalization; the default this PR creates is an absent or empty allowlist, and that is the value that must skip the AWS calls.

tests/integration/cartography/intel/aws/test_init.py
```diff
@@ -26,6 +25,10 @@
 TEST_REGIONS = ["us-east-1", "us-west-2"]
 TEST_UPDATE_TAG = 123456789
 GRAPH_JOB_PARAMETERS = {"UPDATE_TAG": TEST_UPDATE_TAG}
+PUBLIC_SSM_JOB_PARAMETERS = {
```
jychp, line 29: Turn `PUBLIC_SSM_JOB_PARAMETERS` into a fixture or copy it at each call site. Four tests now alias the same module-level dict as `common_job_parameters`, and sync code writes into that dict elsewhere in the codebase, so a future change makes these tests order-dependent.

docs/root/modules/aws/config.md
```diff
@@ -153,11 +153,12 @@ The configuration priority is:
...
 2. `AWS_SSM_PUBLIC_PARAMETER_PREFIX_ALLOWLIST`
-3. The built-in defaults: `/aws/service/bottlerocket/` and
-   `/aws/service/eks/optimized-ami/`
 
-Set the CLI option or environment variable to an empty string to disable public
-parameter ingestion.
+If neither is set, public parameter ingestion is disabled because AWS-managed
+namespaces can contain tens of thousands of parameters in every region. Set the
+CLI option or environment variable to opt in to the prefixes your deployment
+needs. Existing deployments that rely on Bottlerocket or EKS optimized AMI
```
jychp, line 160: Spell out the previous default value here so operators can copy it: `/aws/service/bottlerocket/,/aws/service/eks/optimized-ami/`. Telling them to configure those prefixes explicitly without giving the exact string forces them to dig through the diff.

cartography/cli.py
```diff
@@ -658,7 +658,7 @@ def run(
                     "--aws-ssm-public-parameter-prefix-allowlist",
                     help=(
                         "Comma-separated AWS-managed public SSM parameter prefixes to ingest. "
-                        "Set to an empty string to disable public parameter ingestion."
+                        "Public parameter ingestion is disabled by default."
```
jychp, line 664: Add an example value to the help text, e.g. `/aws/service/eks/optimized-ami/`. This option is now the only way to enable ingestion and the expected prefix format is not obvious from the description alone.

jychp (approved):
LGTM, added some nitpick and suggestions

cartography/intel/aws/__init__.py
kunaals: i exposed one canonical prefix resolver and now use it for both orchestration and ingestion, so normalization and minimization have one owner.

cartography/intel/aws/ssm.py
kunaals: i removed the empty default constant and return the default directly.

tests/integration/cartography/intel/aws/test_init.py
kunaals: i added the load-then-disable integration test. it exposed that schema cleanup skips relationship-free nodes, so i also fixed public parameter cleanup with Cartography's existing batched statement primitive.

tests/integration/cartography/intel/aws/test_init.py
kunaals: the disabled path now covers `None`, `""`, and `","`.

tests/integration/cartography/intel/aws/test_init.py
kunaals: each test now passes its own copy of the job parameters.

docs/root/modules/aws/config.md
kunaals: i documented the exact previous default value.

cartography/cli.py
kunaals: i added an example prefix to the CLI help.

jychp (approved):

PR #3218, opened 2026-08-29
feat(rules): inventory running AI agents

kunaals:
### Type of change
- [ ] Documentation update

### Summary

Adds `aibom_agent_runtime_inventory`, which reports AIBOM agents whose scanned image is currently running on a workload.

Rows are grouped by `AIBOMComponent.logical_id` and workload so the same logical agent found in repeated scans or equivalent images does not create duplicate findings. The representative component id remains attached as the graph asset anchor.

The existing `HAS_RUNTIME_IMAGE` analysis now preserves three-state exposure evidence. Any explicit public signal produces `true`, explicit non-public evidence produces `false`, and absent evidence remains unset instead of being treated as non-public.

The finding reports the workload rather than copying its context. Tenant, cluster, and current internet reachability remain available from the graph around that workload.

### Related issues or links

None.

### How was this tested?

- `make test_lint`
- Existing Cartography test suite from the original change
- Integration coverage for repeated logical agents across scans
- Integration coverage for public, explicitly internal, and unknown workload exposure

#### General

#### Proof of functionality

### Notes for reviewers

The runtime join uses the existing materialized `(:ComputeService)-[:HAS_RUNTIME_IMAGE]->(:Image)` edge. Its cost scales with the findings it must emit: logical agents per image multiplied by workloads running that image. Grouping before projection bounds duplicate scan occurrences.

`cypher_count_query` remains the compliance denominator and now counts distinct logical agents rather than physical per-scan component nodes.

### DIFF
--- a/cartography/analysis/ontology/analysis.py
+++ b/cartography/analysis/ontology/analysis.py
@@ -723,7 +723,9 @@ WORKLOAD_HAS_RUNTIME_IMAGE = AnalysisJob(
                 ":Container on a single node with no intermediate WORKLOAD_PARENT hop. "
                 "Exposure is the OR of the service-level signal (svc.exposed_internet, "
                 "where GCP writes Cloud Run ingress exposure) and any running replica's "
-                "signal (rt.exposed_internet, where AWS ECS / Kubernetes write it). "
+                "signal (rt.exposed_internet, where AWS ECS / Kubernetes write it). If "
+                "neither level has evaluated exposure, the edge keeps it unknown rather "
+                "than treating missing evidence as proof that the workload is internal. "
                 "The active-state filter accepts both 'running' (AWS ECS RUNNING, "
                 "Kubernetes running, Azure Running, GCP Cloud Run) and 'ready' (Scaleway "
                 "serverless ContainerStatus.READY) because _ont_state carries the raw "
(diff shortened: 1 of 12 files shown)


kunaals:
Thanks — went through all three. The PR has since been narrowed to just the new rule (the `exposure_known` edge property and the `aibom_agent_inventory` changes are gone), which resolves the first two by deletion. Detail on each:

**1. Stale `svc.exposed_internet` (analysis.py)** — moot now, but the premise didn't hold: the analysis framework auto-generates node-property cleanup from `SetProperty` effects (`cartography/graph/analysisbuilder.py:469` emits `REMOVE {props}`), and `PropertyEffect.cleanup_before_statements` defaults to `True`. So each exposure job wipes `exposed_internet` from its labels before re-deriving, and a stale value can't survive a run of the job that owns it.

**2. `count(hri)` counting edges rather than workloads** — this one was fair on its face; `count(DISTINCT svc)` would have been the correct form for a field named `running_workload_count`. The subquery it referred to has been removed from the PR, so it no longer applies.

**3. Count query should match the fact's predicate** — this one I'd push back on. `cypher_count_query` is defined as the denominator, not the finding count. From `docs/root/usage/rules.md`:

> Count **all** assets of the relevant type, regardless of whether they match the Fact criteria
> Enable calculation of compliance ratios (e.g., "10 public buckets out of 100 total")

Adding the exposure predicate would make count equal findings, so every rule would report 100% affected and the ratio would carry no information. The existing rules follow the same shape — `serverless_workload_exposed`'s facts count all `GCPCloudRunService`s while filtering on ingress, and `aibom_coverage_gaps` counts all `AIBOMSource`s while filtering on `gap_reason`. Leaving as-is.

PR #3223, opened 2026-09-01
feat(railway): resolve deployed images by digest

kunaals:
### Type of change

### Summary

Railway reports the immutable image digest for a deployment in `Deployment.meta.imageDigest` and a deployment-scoped image reference in `Deployment.meta.image`. The typed `ServiceSource.image` field describes current configuration and may be a mutable tag.

This stores the deployment-scoped image reference and validated SHA-256 digest on `RailwayDeployment`, maps both into the canonical Container ontology, and creates Cartography standard `HAS_IMAGE` relationships to existing registry image nodes. A configured digest is used only for the latest deployment when deployment metadata supplies no image reference; older active deployments and tagged metadata fail closed instead of inheriting a potentially newer configured digest.

Single-platform digests allow the ontology to derive `RESOLVED_IMAGE` and `HAS_RUNTIME_IMAGE`. A multi-architecture manifest digest intentionally remains linked to the manifest list unless Railway also provides runtime architecture; Cartography does not guess a platform image.

### Related issues or links

- Stacked on #3222.

### How was this tested?

- Full unit suite on the top of the stack: 5,316 passed
- Railway and image-ontology integration suite: 57 passed
- Focused deployment integration suite: 13 passed
- Repository-wide pre-commit hooks
- Read-only validation against a redacted live Railway response confirmed `meta.imageDigest` uses the `sha256:<64 hex>` shape for image-backed and source-built deployments

### Checklist

- [x] I have added or updated tests that prove the feature works.
- [x] Updated model property descriptions used to generate schema documentation.
- [x] Built the documentation.

### Notes for reviewers

Mutable tag configuration is never treated as runtime image identity. SHA-256 is deliberate because the registry image schemas currently use SHA-256 identities.


cartography/models/railway/deployment.py
```diff
@@ -32,6 +32,16 @@ class RailwayDeploymentNodeProperties(CartographyNodeProperties):
         extra_index=True,
         description="Full source commit SHA reported for this deployment.",
     )
+    source_image: PropertyRef = PropertyRef(
```
jychp: These two new properties are not carried into the Container ontology, and the mapping still claims they cannot be. In `cartography/models/ontology/mapping/data/containers.py`, `railway_mapping` maps only `state` and ends with:

```python
# name / image / image_digest / cpu / memory / region / health_status:
# Not available on Railway's Deployment type.
```

That comment is false as of this PR, and `image` / `image_digest` are established Container ontology fields: AWS ECS, Kubernetes, Azure Container Instance and Scaleway all map them.

Failure mode: a query like

```cypher
MATCH (c:Container) WHERE c._ont_image_digest = 'sha256:...' RETURN c
```

returns the ECS / Kubernetes / Azure containers running that digest but silently omits the Railway deployment running the very same one, even though the value now sits on the node. Adding the two `OntologyFieldMapping` entries (`image` -> `source_image`, `image_digest` -> `image_digest`) and dropping the stale comment would close it.

cartography/intel/railway/deployments.py
```diff
@@ -66,7 +87,34 @@ def transform(
...
+            source = (current_instance or {}).get("source") or {}
             meta = deployment.get("meta")
+            meta_image = meta.get("image") if isinstance(meta, dict) else None
+            configured_image, source_pinned_digest = parse_image_uri(
+                source.get("image"),
+            )
+            source_image = configured_image
+            meta_pinned_digest = None
+            source_fallback_digest = source_pinned_digest
+            if isinstance(meta_image, str) and meta_image.strip():
```
jychp: The source-configuration fallback can attribute a newer image to an older deployment, which is the false attribution the PR description says the design avoids.

`source` comes from the service instance's *current* configuration, and the guard here only compares repositories -- never the digest or the tag. So:

1. Service instance source is `reg.example.com/team/app@sha256:NEW`.
2. A deployment created before that re-pin reports `meta.image = "reg.example.com/team/app:v1"` and no `meta.imageDigest`.
3. `_image_repository` is `reg.example.com/team/app` on both sides, so `source_fallback_digest` survives and `image_digest` becomes `sha256:NEW`.
4. That deployment -- which actually ran v1 -- gets `HAS_IMAGE` to the NEW image, and `RESOLVED_IMAGE` / `HAS_RUNTIME_IMAGE` too if it is the current revision. Vulnerabilities of the new image are now attributed to a revision that never ran it.

The same holds, with no repository check at all, when `meta.image` is absent entirely: `source_fallback_digest` is then used unconditionally.

Two options that would both hold the stated invariant: only accept the configured digest when `meta.image` is absent *and* the deployment is `current`, or require the tag in `meta.image` to also appear in the configured reference before trusting its digest. Either way it would be worth a comment saying why the fallback is safe, since the repository comparison reads like it is doing more work than it is.

cartography/models/railway/deployment.py
```diff
@@ -106,6 +116,71 @@ class RailwayDeploymentToServiceInstanceRel(CartographyRelSchema):
     )
 
 
+@dataclass(frozen=True)
+class RailwayDeploymentToImageRelProperties(CartographyRelProperties):
+    lastupdated: PropertyRef = PropertyRef("lastupdated", set_in_kwargs=True)
+
+
+@dataclass(frozen=True)
+class RailwayDeploymentToECRImageRel(CartographyRelSchema):
```
jychp: Digests that resolve to a manifest list will stop at `HAS_IMAGE` and never produce the `RESOLVED_IMAGE` / `HAS_RUNTIME_IMAGE` the PR summary promises -- and multi-arch is the common shape for registry images, not the edge case.

`CONTAINER_RESOLVED_IMAGE` has two statements: the first excludes `i:ImageManifestList`, and the second (the one that walks `CONTAINS_IMAGE` into the per-arch child) requires `c.architecture_normalized IS NOT NULL`. `RailwayDeployment` defines no such property, unlike `AzureContainerInstance`, which sets `architecture_normalized` precisely for this branch.

So a service pinned to a multi-arch image gets `HAS_IMAGE` -> `:Image:ImageManifestList` and nothing downstream: no `RESOLVED_IMAGE`, hence no `HAS_RUNTIME_IMAGE` from `WORKLOAD_HAS_RUNTIME_IMAGE`.

Railway does not expose an architecture on the deployment as far as I can tell, so this may not be fixable here -- but if that is the case it is worth stating in the model comment and in the PR description, rather than claiming the ontology edges are derived.

cartography/intel/railway/deployments.py
```diff
@@ -18,6 +19,26 @@
 logger = logging.getLogger(__name__)
 
 
+def _normalize_image_digest(value: Any) -> str | None:
+    if not isinstance(value, str):
+        return None
+    digest = value.strip()
+    if not re.fullmatch(r"sha256:[0-9a-fA-F]{64}", digest):
```
jychp: Nit: `sha256:` only. OCI allows other algorithms (`sha512:`), and a registry using one would have its digest silently dropped rather than logged. Given the PR notes say a live Railway response only ever showed `sha256:<64 hex>`, restricting it is defensible -- just worth saying so in a comment so it reads as a decision rather than an oversight.

cartography/models/railway/deployment.py
jychp, line 35: `railway_mapping` in `containers.py` still says `image / image_digest: Not available on Railway's Deployment type` and maps neither. As of this PR they are available.

Result: `_ont_image` / `_ont_image_digest` stay null, so `MATCH (c:Container) WHERE c._ont_image_digest = $d` skips Railway while ECS, Kubernetes, Azure and Scaleway all answer. Add the two `OntologyFieldMapping` entries and drop the stale comment.

cartography/intel/railway/deployments.py
jychp, line 101: `source` is the instance's *current* config and this guard compares repositories only, never the digest.

Service re-pinned to `app@sha256:NEW`; an older deployment reports `app:v1` with no `meta.imageDigest` -> same repository -> it gets `HAS_IMAGE` to NEW, i.e. the false attribution the description says this avoids. Same risk with no check at all when `meta.image` is absent.

Either restrict the fallback to `current` deployments, or require the tag in `meta.image` to appear in the configured reference.

cartography/intel/railway/deployments.py
```diff
@@ -18,6 +19,26 @@
...
+        return None
+    return digest.lower()
+
+
+def _image_repository(image_uri: str | None) -> str | None:
+    if image_uri is None:
+        return None
+    repository = image_uri.partition("@")[0]
```
jychp: Nit: `nginx` and `docker.io/library/nginx` compare unequal, so a legitimately pinned digest is dropped. Fails closed, which is fine, but worth a comment saying the comparison is deliberately literal.

cartography/intel/railway/deployments.py
jychp, line 28: Nit: `sha256:` only, silently dropping other OCI algorithms (`sha512:`). Worth one line saying that is deliberate.

jychp (changes requested):

cartography/models/railway/deployment.py
kunaals: good catch. railway now maps `source_image` and `image_digest` to the canonical container fields.

tests/integration/cartography/intel/railway/test_deployments.py
```diff
@@ -93,6 +137,212 @@ def test_load_railway_deployments(neo4j_session):
...
+        TEST_UPDATE_TAG,
+    )
+    _run_resolved_image_analysis(neo4j_session)
+
+    # Assert
+    assert check_nodes(
+        neo4j_session,
+        "RailwayDeployment",
+        ["id", "source_image", "image_digest"],
+    ) >= {
```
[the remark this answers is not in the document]
kunaals: good catch. the deployment node assertions now cover the complete expected set.

cartography/intel/railway/deployments.py
kunaals: this repository comparison is gone; the fallback now depends on latest-deployment provenance instead.

cartography/intel/railway/deployments.py
kunaals: yep. added the rationale: the registry image schemas currently use sha-256 identities, so other algorithms fail closed.

cartography/models/railway/deployment.py
```diff
@@ -106,6 +116,71 @@ class RailwayDeploymentToServiceInstanceRel(CartographyRelSchema):
...
+    """Links a Railway deployment to an Amazon ECR image by immutable digest."""
+
+    target_node_label: str = "AWSECRImage"
+    target_node_matcher: TargetNodeMatcher = make_target_node_matcher(
+        {"digest": PropertyRef("image_digest")},
```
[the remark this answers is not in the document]
kunaals: keeping this conservative: railway does not report runtime architecture, so a manifest-list digest stays attached via `HAS_IMAGE` and we do not guess a platform image. the pr description now calls out that boundary.

cartography/intel/railway/deployments.py
kunaals: good catch. configured digests are now limited to the latest deployment with no deployment-scoped image, and tagged or older active deployments fail closed.

PR #3224, opened 2026-09-01
feat(ontology): link DNS records to service endpoints

kunaals:
### Type of change

### Summary

Cross-provider DNS analysis links records to several infrastructure hostnames, but misses public service endpoints that already carry stable hostname or URL properties.

This adds `DNS_POINTS_TO` coverage for Modal sandbox tunnels and web functions, GCP Cloud Run services, Netlify default and custom domains, and Scaleway serverless invocation domains.

Each connector preserves its raw provider value and derives an indexed canonical hostname during ingestion. URL hostnames are parsed in Python rather than in Cypher, and the analysis uses direct indexed equality against the canonical DNS fields from #3231. Netlify aliases and GCP DNS targets retain explicit canonical lists where the provider data is list-valued.

### Related issues or links

- Stacked on #3223.

### How was this tested?

- Full unit suite: 5,315 passed
- Full integration suite: 1,608 passed
- Focused cross-provider suite: 110 passed
- Repository-wide pre-commit hooks
- `uv run ./docs/build.sh`

#### General

- [x] I have added or updated tests that prove the feature works.

#### Proof of functionality

- [x] New or updated unit and integration tests.

- [x] Updated model property descriptions used to generate schema documentation.
- [x] Built the documentation.

### Notes for reviewers

Only provider models that already identify a public serving address are included. Provider-specific DNS ownership relationships remain protected by the existing cleanup guards.


cartography/analysis/ontology/analysis.py
```diff
@@ -293,7 +349,7 @@
             cleanup_iterationsize=1000,
             statements=(
                 AnalysisStatement(
-                    match=f"MATCH (dns:DNSRecord) WHERE dns._ont_value IS NOT NULL {match_filter} WITH dns MATCH (target:{target_label}) WHERE toLower(rtrim(toString(dns._ont_value), '.')) = toLower(rtrim(target.{target_property}, '.'))",
+                    match=f"MATCH (dns:DNSRecord) WHERE dns._ont_value IS NOT NULL {match_filter} WITH dns MATCH (target:{target_label}) UNWIND {target_values_expression} AS target_value WITH dns, target, target_value WHERE target_value IS NOT NULL AND trim(toString(target_value)) <> '' AND toLower(rtrim(toString(dns._ont_value), '.')) = toLower(rtrim(toString(target_value), '.'))",
```
jychp: Cosmetic: with an empty `match_filter` the generated query now contains a double space (`... WHERE dns._ont_value IS NOT NULL  WITH dns MATCH ...`). Harmless for Cypher, but it shows up in every logged query for the six new jobs -- `{match_filter}` could be joined with a leading-space-only-if-non-empty helper.

cartography/analysis/ontology/analysis.py
```diff
@@ -259,27 +259,83 @@
...
+    ),
+    (
+        "GCPCloudRunService",
+        "[split(replace(replace(target.uri, 'https://', ''), 'http://', ''), '/')[0]]",
+        "",
+        "",
+    ),
+    (
+        "NetlifySite",
+        "[target.default_domain, target.custom_domain] + coalesce(target.domain_aliases, [])",
```
jychp: `custom_domain` and `domain_aliases` are the customer's own hostnames, but candidates are only ever compared to `dns._ont_value`.

In the standard setup (`www.example.com CNAME <site>.netlify.app`) the custom domain is the record *name*, so only `default_domain` matches here; an apex A record whose `_ont_name` equals `custom_domain` gets no edge. #3222 already handles this class via `_ont_name` (`DNS_RECORD_TO_RAILWAY_CUSTOM_DOMAIN`).

Keep `[target.default_domain]` here and match the customer-owned domains on `_ont_name`.

cartography/analysis/ontology/analysis.py
```diff
@@ -259,27 +259,83 @@
...
+    (
+        "RailwayServiceDomain",
+        "[target.domain]",
+        "AND NOT dns:GCPRecordSet",
+        "",
+    ),
+    (
+        "ModalSandboxTunnel",
+        "[target.host, target.unencrypted_host]",
+        "",
```
jychp: All six new entries pass an empty `match_filter`; every pre-existing one carries `AND NOT dns:GCPRecordSet`, which is what stops the generic statement from overlapping the dedicated `GCPRecordSet` one.

Inert today only because `dnsrecords.py` leaves `value` unmapped for `GCPRecordSet` -- a decision that file flags as revisitable. Cheap to make uniform.

cartography/analysis/ontology/analysis.py
jychp: Cosmetic: an empty `match_filter` leaves a double space in every generated query for the six new jobs.

PR #3227, opened 2026-09-01
refactor(okta): model AWS role mappings

kunaals:
### Type of change

- [x] Documentation update

### Summary

Model Okta group-to-AWS role mappings as scoped MatchLinks instead of handwritten Cypher. This makes the canonical `(:OktaGroup)-[:HAS_ROLE]->(:AWSRole)` relationship available to schema introspection and generated Okta and AWS documentation.

The sync continues to create the deprecated `ALLOWED_BY` relationship for backward compatibility. It will be removed in Cartography v1. The MatchLinks are scoped to their Okta organization, stale relationships are cleaned up, and a migration cleanup removes unscoped relationships created by earlier versions.

### Related issues or links

- Follow-up to https://github.com/cartography-cncf/cartography/pull/1243#issuecomment-5351995038

### How was this tested?

- Synced Okta from a live tenant in read-only provider mode into an isolated local Neo4j database.
- Used a sanitized, database-only mapping to exercise creation, stale cleanup, and recreation against the live-synced Okta graph and local AWS Identity Center nodes. The tenant's current group names did not match the configured account-and-role naming pattern, so the validation did not rely on naturally occurring cross-module edges.

Sanitized validation output:

```text
INFO:cartography.intel.okta.awssaml:Syncing Okta SAML Integration
created=ALLOWED_BY:1,HAS_ROLE:1
stale_after_unmap=0
recreated=ALLOWED_BY:1,HAS_ROLE:1
invalid_scope=0
```

#### General

#### Proof of functionality

### Notes for reviewers

The deprecated `ALLOWED_BY` compatibility schema is intentionally excluded from generated documentation. The canonical `HAS_ROLE` relationship appears in both the Okta and AWS generated schema pages.

### DIFF
--- a/cartography/data/jobs/cleanup/okta_import_cleanup.json
+++ b/cartography/data/jobs/cleanup/okta_import_cleanup.json
@@ -31,7 +31,7 @@
       "__comment__": "Migration: drop deprecated Okta-to-Kubernetes group mapping edges, replaced by MEMBER_OF"
     },
     {
-      "query": "MATCH (:OktaOrganization{id: $OKTA_ORG_ID})-[:RESOURCE]->(:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "query": "MATCH (:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
       "iterative": true,
       "iterationsize": 100,
       "__comment__": "Migration: drop stale Okta-to-AWS role mappings created before ALLOWED_BY used scoped MatchLink metadata"
--- a/cartography/data/jobs/cleanup/okta_import_cleanup.json
+++ b/cartography/data/jobs/cleanup/okta_import_cleanup.json
@@ -31,10 +31,10 @@
       "__comment__": "Migration: drop deprecated Okta-to-Kubernetes group mapping edges, replaced by MEMBER_OF"
     },
     {
-      "query": "MATCH (:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "query": "MATCH (:OktaOrganization {id: $OKTA_ORG_ID})-[:RESOURCE]->(:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
       "iterative": true,
       "iterationsize": 100,
-      "__comment__": "Migration: drop stale Okta-to-AWS role mappings created before ALLOWED_BY used scoped MatchLink metadata"
+      "__comment__": "Migration: drop stale Okta-to-AWS role mappings for the current organization that were created before ALLOWED_BY used scoped MatchLink metadata"
     },
     {
       "query": "MATCH (n:OktaAdministrationRole) WITH n LIMIT $LIMIT_SIZE DETACH DELETE n",
--- a/cartography/data/jobs/cleanup/okta_import_cleanup.json
+++ b/cartography/data/jobs/cleanup/okta_import_cleanup.json
@@ -31,10 +31,10 @@
       "__comment__": "Migration: drop deprecated Okta-to-Kubernetes group mapping edges, replaced by MEMBER_OF"
     },
     {
-      "query": "MATCH (:OktaOrganization {id: $OKTA_ORG_ID})-[:RESOURCE]->(:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "query": "MATCH (:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
       "iterative": true,
       "iterationsize": 100,
-      "__comment__": "Migration: drop stale Okta-to-AWS role mappings for the current organization that were created before ALLOWED_BY used scoped MatchLink metadata"
+      "__comment__": "Migration: drop stale Okta-to-AWS role mappings created before ALLOWED_BY used scoped MatchLink metadata"
     },
     {
       "query": "MATCH (n:OktaAdministrationRole) WITH n LIMIT $LIMIT_SIZE DETACH DELETE n",
--- a/cartography/data/jobs/cleanup/okta_import_cleanup.json
+++ b/cartography/data/jobs/cleanup/okta_import_cleanup.json
@@ -31,6 +31,12 @@
       "__comment__": "Migration: drop deprecated Okta-to-Kubernetes group mapping edges, replaced by MEMBER_OF"
     },
     {
+      "query": "MATCH (:OktaOrganization{id: $OKTA_ORG_ID})-[:RESOURCE]->(:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
+      "iterative": true,
+      "iterationsize": 100,
+      "__comment__": "Migration: drop stale Okta-to-AWS role mappings created before ALLOWED_BY used scoped MatchLink metadata"
+    },
+    {
       "query": "MATCH (n:OktaAdministrationRole) WITH n LIMIT $LIMIT_SIZE DETACH DELETE n",
       "iterative": true,
       "iterationsize": 100,
--- a/cartography/intel/okta/awssaml.py
+++ b/cartography/intel/okta/awssaml.py
@@ -7,9 +7,12 @@ from collections import namedtuple
 
 import neo4j
 
+from cartography.client.core.tx import load_matchlinks
 from cartography.client.core.tx import read_list_of_dicts_tx
 from cartography.client.core.tx import read_single_value_tx
-from cartography.client.core.tx import run_write_query
+from cartography.graph.job import GraphJob
+from cartography.models.okta.awssaml import OktaGroupToAWSRoleAllowedByMatchLink
+from cartography.models.okta.awssaml import OktaGroupToAWSRoleHasRoleMatchLink
 from cartography.util import timeit
 
 AccountRole = namedtuple("AccountRole", ["account_id", "role_name"])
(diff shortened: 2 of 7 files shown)


cartography/data/jobs/cleanup/okta_import_cleanup.json
```diff
@@ -30,6 +30,12 @@
       "iterationsize": 100,
       "__comment__": "Migration: drop deprecated Okta-to-Kubernetes group mapping edges, replaced by MEMBER_OF"
     },
+    {
+      "query": "MATCH (:OktaOrganization{id: $OKTA_ORG_ID})-[:RESOURCE]->(:OktaGroup)-[r:ALLOWED_BY]->(:AWSRole) WHERE r._sub_resource_label IS NULL WITH r LIMIT $LIMIT_SIZE DELETE r",
```
jychp: drop the `$OKTA_ORG_ID` scoping from this migration statement. The legacy loader matched `(app:OktaApplication{name:'amazon_aws'})--(group:OktaGroup)` with no org filter, so unscoped `ALLOWED_BY` edges exist on groups whose OktaOrganization is no longer synced, and this statement can never reach them.

cartography/models/okta/awssaml.py
```diff
@@ -0,0 +1,85 @@
...
+    direction: LinkDirection = LinkDirection.OUTWARD
+    rel_label: str = "HAS_ROLE"
+    properties: OktaGroupToAWSRoleRelProperties = OktaGroupToAWSRoleRelProperties()
+
+
+@dataclass(frozen=True)
+# DEPRECATED: replaced by the canonical (:UserGroup)-[:HAS_ROLE]->(:PermissionRole)
+# edge (OktaGroupToAWSRoleHasRoleMatchLink). Kept for backward compatibility and
+# will be removed in v1.0.0.
+class OktaGroupToAWSRoleAllowedByMatchLink(CartographyRelSchema):
```
jychp: Subclass `OktaGroupToAWSRoleHasRoleMatchLink` and override only `rel_label`. The two schemas duplicate every matcher, sub-resource and properties field today and will drift.

cartography/intel/okta/awssaml.py
```diff
@@ -97,32 +108,41 @@ def query_for_okta_to_aws_role_mapping(
...
+    for schema in schemas:
+        load_matchlinks(
+            neo4j_session,
+            schema,
+            mappings,
+            lastupdated=okta_update_tag,
+            _sub_resource_label="OktaOrganization",
+            _sub_resource_id=okta_org_id,
+        )
+        GraphJob.from_matchlink(
```
jychp, line 152: Move the `GraphJob.from_matchlink(...).run()` calls out of `_load_okta_group_to_aws_roles` into a `cleanup()` function called from `sync_okta_aws_saml`. A loader must not delete.

cartography/intel/okta/awssaml.py
```diff
@@ -97,32 +108,41 @@ def query_for_okta_to_aws_role_mapping(
...
-    ON CREATE SET r.firstseen = timestamp()
-    SET r.lastupdated = $okta_update_tag
-    """
-
-    run_write_query(
-        neo4j_session,
-        ingest_statement,
-        GROUP_TO_ROLE=group_to_role,
-        okta_update_tag=okta_update_tag,
+    mappings = [mapping._asdict() for mapping in group_to_role]
```
jychp: Dedupe `group_to_role` before building `mappings`. A group assigned to both the `amazon_aws` and `amazon_aws_sso` applications resolves to the same `(okta_group_id, aws_role_arn)` pair twice and is merged twice per schema.

cartography/intel/okta/awssaml.py
```diff
@@ -259,17 +252,26 @@ def sync_okta_aws_saml(
     group_to_role_mapping = query_for_okta_to_aws_role_mapping(
         neo4j_session,
         mapping_regex,
+        okta_org_id,
     )
-    _load_okta_group_to_aws_roles(neo4j_session, group_to_role_mapping, okta_update_tag)
+    combined_group_to_role_mapping = [
```
jychp: Return `GroupRole` from `query_for_okta_to_aws_role_mapping` and delete this remapping block. Two shapes for the same concept (`{"groupid", "role"}` vs `GroupRole`) only exist because the loader signature changed.

tests/integration/cartography/intel/okta/test_awssaml.py
```diff
@@ -202,6 +234,175 @@ def test_sync_okta_aws_saml_no_matching_roles(neo4j_session):
...
+    other_org_id = "other-okta-org-id"
+    other_group_id = "other-org-group"
+    other_sso_group_id = "other-org-sso-group"
+    other_role_arn = "arn:aws:iam::4321:role/other-org-role"
+    other_sso_role_arn = "arn:aws:iam::4321:role/AWSReservedSSO_other-sso-role_abcdef"
+    neo4j_session.run(
+        """
+        MERGE (current_org:OktaOrganization {id: $CURRENT_ORG_ID})
+        MERGE (org:OktaOrganization {id: $ORG_ID})
+        MERGE (app:OktaApplication {name: "amazon_aws"})
```
jychp: This `MERGE` keys on `name` only, so both orgs share a single `OktaApplication` node and the new `(org)-[:RESOURCE]->(app)` clause is never exercised. Give each org its own app node with a distinct `id`.

tests/integration/cartography/intel/okta/test_awssaml.py
```diff
@@ -202,6 +234,175 @@ def test_sync_okta_aws_saml_no_matching_roles(neo4j_session):
...
+    relationship_count = neo4j_session.run(
+        """
+        MATCH (group:OktaGroup)-[r]->(:AWSRole)
+        WHERE group.id IN $GROUP_IDS
+              AND type(r) IN ["ALLOWED_BY", "HAS_ROLE"]
+        RETURN count(r) AS count
+        """,
+        GROUP_IDS=[other_group_id, other_sso_group_id],
+    ).single(strict=True)["count"]
+    assert relationship_count == 0
```
jychp: Set up current-org groups and roles too, and assert their `HAS_ROLE`/`ALLOWED_BY` edges are created. As written this test passes even if `sync_okta_aws_saml` is a no-op.

tests/integration/cartography/intel/okta/test_awssaml.py
```diff
@@ -202,6 +234,175 @@ def test_sync_okta_aws_saml_no_matching_roles(neo4j_session):
...
+
+    # Assert
+    relationship_count = neo4j_session.run(
+        """
+        MATCH (:OktaGroup {id: $GROUP_ID})-[r:ALLOWED_BY]->(:AWSRole)
+        RETURN count(r) AS count
+        """,
+        GROUP_ID=group_id,
+    ).single(strict=True)["count"]
+    assert relationship_count == 0
```
jychp, line 445: Cover the upgrade path that actually matters: a pre-existing unscoped `ALLOWED_BY` for a mapping that is still valid must be adopted by the sync (gaining `_sub_resource_label`) and survive this migration statement. Only the orphaned-edge case is tested.

jychp (changes requested):

cartography/data/jobs/cleanup/okta_import_cleanup.json
kunaals: yep, this needs to be global. updated it to catch orphaned groups too.

cartography/intel/okta/awssaml.py
kunaals: moved this into a separate cleanup step after loading.

cartography/intel/okta/awssaml.py
kunaals: deduped these before loading either relationship.

cartography/intel/okta/awssaml.py
kunaals: cleaned this up so the direct saml path uses `GroupRole` throughout.

tests/integration/cartography/intel/okta/test_awssaml.py
kunaals: fixed the fixture so each org has distinct application nodes.

tests/integration/cartography/intel/okta/test_awssaml.py
kunaals: added positive assertions for both current-org paths.

tests/integration/cartography/intel/okta/test_awssaml.py
kunaals: added this upgrade case and confirmed the adopted edge survives cleanup.

cartography/data/jobs/cleanup/okta_import_cleanup.json
kunaals: i accidentally reintroduced the org scope while addressing a later review. reverted that- the migration stays global so it can clean up orphaned legacy edges, while the recurring matchlink cleanup remains scoped per okta org.

tests/integration/cartography/intel/okta/test_awssaml.py
```diff
@@ -202,6 +234,302 @@ def test_sync_okta_aws_saml_no_matching_roles(neo4j_session):
...
+    role_arn = "arn:aws:iam::1234:role/myrole1"
+    _setup_okta_test_data(neo4j_session)
+    _setup_aws_test_data(neo4j_session)
+    neo4j_session.run(
+        """
+        MATCH (org:OktaOrganization {id: $ORG_ID})
+        MATCH (app:OktaApplication {name: "amazon_aws"})
+        MATCH (role:AWSRole {arn: $ROLE_ARN})
+        MERGE (group:OktaGroup {id: $GROUP_ID})
+        SET group.name = "aws#test#myrole1#1234"
```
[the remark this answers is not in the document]
kunaals: fixed. the fixture now sets the current update tag, matching the real group sync before migration cleanup.

cartography/models/okta/awssaml.py
```diff
@@ -0,0 +1,83 @@
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
kunaals: keeping this visible because it is a real compatibility edge that users can query until v1, matching how `MEMBER_OF_OKTA_GROUP` is documented. introspection exclusion is for schemas that don’t accurately describe a concrete edge.

tests/integration/cartography/intel/okta/test_awssaml.py
```diff
@@ -202,6 +234,283 @@ def test_sync_okta_aws_saml_no_matching_roles(neo4j_session):
...
+
+    # Assert
+    relationship_count = neo4j_session.run(
+        """
+        MATCH (:OktaGroup {id: $GROUP_ID})-[r:ALLOWED_BY]->(:AWSRole)
+        RETURN count(r) AS count
+        """,
+        GROUP_ID=group_id,
+    ).single(strict=True)["count"]
+    assert relationship_count == 0
```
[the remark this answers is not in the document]
kunaals: keeping the smaller orphaned-edge case. because the group has no `OktaOrganization` `RESOURCE` edge, this test already fails if the migration is scoped to `$OKTA_ORG_ID`, so it directly pins the global behavior. a second org would exercise the same predicate without adding coverage.

PR #3228, opened 2026-09-02
feat(rules): expose Fact anchor check as public API

jychp:
### Type of change

- [ ] Documentation update

### Summary

Add a public `validate_anchor(cypher_query, asset_label, asset_id_field)` helper so callers can check that a fact query's `(label, id)` pair describes one node without constructing a `Fact`.

`Fact.__post_init__` now uses that helper, so construction and the standalone check cannot diverge. The Cypher parsing helpers stay private.

### Related issues or links

None.

### How was this tested?

- `uv run pytest tests/unit/rules/test_model.py tests/unit/rules/test_asset_label.py tests/unit/rules/test_asset_anchor.py`
- `pre-commit run --files` on the three touched files

### Test plan
- Construct a valid `Fact`; it still succeeds.
- Call `validate_anchor` on a query that projects the id off the labeled variable; it returns `None`.
- Call `validate_anchor` when the id is only bound in an intermediate `WITH`; it returns `missing_id_alias`.
- Call `validate_anchor` when `asset_label` is never bound; it returns `unbound_label`.
- Call `validate_anchor` when the id is read off a different node; it returns `id_not_on_labeled_var`.
- Confirm existing `Fact` `ValueError` messages for those cases are unchanged.

#### General

#### Proof of functionality

  - Example:
    ```
    ```

### Notes for reviewers

`validate_anchor` returns a structured failure (`AnchorValidationError` with a reason code) instead of raising, so callers can validate a candidate query without building a `Fact`. Identity-field and reserved-alias checks stay on `Fact.__post_init__` only.

### DIFF
--- a/cartography/rules/spec/model.py
+++ b/cartography/rules/spec/model.py
@@ -122,6 +122,111 @@ def _expression_for_alias(cypher_query: str, alias: str) -> str | None:
     return None
 
 
+class AnchorValidationCode(str, Enum):
+    """Why ``validate_anchor`` rejected a candidate ``(asset_label, asset_id_field)``."""
+
+    EMPTY_LABEL = "empty_label"
+    EMPTY_ID_FIELD = "empty_id_field"
+    MISSING_ID_ALIAS = "missing_id_alias"
+    UNBOUND_LABEL = "unbound_label"
+    ID_NOT_ON_LABELED_VAR = "id_not_on_labeled_var"
+
+
+@dataclass(frozen=True)
+class AnchorValidationError:
+    """Structured failure from ``validate_anchor``. ``None`` from that function means valid."""
+
+    code: AnchorValidationCode
+    asset_label: str
+    asset_id_field: str
+    id_expression: str | None = None
+    asset_vars: frozenset[str] = frozenset()
+
+
+def validate_anchor(
+    cypher_query: str,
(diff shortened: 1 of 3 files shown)


cartography/rules/spec/model.py
```diff
@@ -122,6 +122,111 @@ def _expression_for_alias(cypher_query: str, alias: str) -> str | None:
...
+    asset_vars = frozenset(_variables_bound_to_label(cypher_query, asset_label))
+    if not asset_vars:
+        return AnchorValidationError(
+            code=AnchorValidationCode.UNBOUND_LABEL,
+            asset_label=asset_label,
+            asset_id_field=asset_id_field,
+        )
+    id_expression = _expression_for_alias(cypher_query, asset_id_field)
+    owners = set(_PROPERTY_OWNER_RE.findall(id_expression or ""))
+    if not owners & asset_vars:
```
kunaals: this passes when the id expression also reads from a differently labeled node. for example, `coalesce(u.arn, r.arn)` passes for `AWSUser` because `u` overlaps `asset_vars`, even though the result can be the role id. should we reject the expression when any property owner falls outside `asset_vars` and add that case to the public-api tests?

kunaals (approved):
nearly good, just fix this one issue and i think we're good to go

PR #3235, opened 2026-09-03
fix(aws): stop one regional 500 from aborting the whole account sync

jychp:
### Type of change

- [ ] Documentation update

### Summary

A transient AWS-side 500 in one optional service currently discards a multi-hour AWS sync of everything else in the account. Two independent causes, one per commit.

**1. `aws_handle_regions` did not recognize the error code.**

The decorator is supposed to degrade a regional failure into a regional skip, but its allowlist only carried `InternalServerErrorException`. `bedrock-agent` returns the Smithy-standard `InternalServerException`, so `is_aws_region_skippable_client_error` returned `False`, the decorator re-raised, and the exception propagated out of `_sync_one_account`. This is not a case of calling an endpoint that does not exist: `bedrock/__init__.py` already filters regions through `filter_regions_to_supported_service_regions(...)`, the endpoint is in botocore's endpoint data and answers, it just answers 500.

```
botocore.errorfactory.InternalServerException: An error occurred (InternalServerException)
when calling the ListKnowledgeBases operation (reached max retries: 3)
```

The server-side codes move out of `AWS_REGION_ACCESS_DENIED_ERROR_CODES`, whose name never described them, into a new `AWS_REGION_SERVER_ERROR_CODES` covering the spellings AWS actually uses. None of them is actionable per-region, so a regional 500 now logs a warning and returns `[]`, exactly what the decorator already does for the other codes. `ServerException` and `ServiceException` are in the list for the same reason: `apigatewayv2.py` already classifies that exact set as server-side retryable, and leaving a synonym out just queues up the identical one-word bug from the next service.

**2. One intel module could take down the account.**

Even with the code above recognized, any unexpected exception from any single resource function still discarded the whole account. `_sync_one_account` iterated `RESOURCE_FUNCTIONS[func_name](**sync_args)` with no per-function isolation, so the blast radius of a bug or an AWS outage in one optional service was every other service in that account, and the work was lost late, after most of the account had already been enumerated.

Each resource function now runs through `_run_resource_function`, which records the failure and continues. The aggregate is raised once the account is done, after the analysis jobs and `merge_module_sync_metadata`, so the data that did land survives while the sync still reports the failure. Credential failures (`ExpiredToken`, `RequestExpired`, `NoCredentialsError`, SSO/token retrieval) and Neo4j `ServiceUnavailable` / `SessionExpired` still propagate immediately: isolating a module only helps when the next one can succeed, and for those it cannot, so without that escape hatch every remaining module would fail the same way, each after up to ten minutes of backoff in `aws_handle_regions`.

### Related issues or links

None filed; reported directly.

### Breaking changes

None, but two behavior changes are worth calling out in review:

- A regional 500 now returns `[]`, so that module's cleanup job deletes the region's nodes for that run. This is already what happens for `AccessDenied` and every other code in the skip set, and losing one region's nodes for one cycle beats losing the account, but it is data loss rather than a no-op.
- `merge_module_sync_metadata` now runs for an account where a module failed, so such an account gets sync metadata for a partial sync. Previously the exception escaped before that write.

### How was this tested?

New unit and integration tests, plus the existing suites:

- `tests/unit/cartography/test_util.py::test_aws_handle_regions_skips_server_errors`, parametrized over all nine server-side codes, asserts the decorator returns `[]` instead of re-raising.
- `tests/integration/cartography/intel/aws/test_init.py::test_sync_one_account_isolates_resource_function_failures` asserts the modules ordered after a failing one still run, including the tags pass that must always be last, and that the aggregate exception names the failing module.
- `tests/integration/cartography/intel/aws/test_init.py::test_sync_one_account_aborts_on_expired_credentials` asserts the fatal path short-circuits and the next module is never called.

```
uv run pytest tests/unit                                       5316 passed
uv run pytest tests/integration/cartography/intel/aws/test_init.py   35 passed
make test_lint                                                 passed
```

#### General

#### Proof of functionality

### Notes for reviewers

The interesting design question is where the aggregate exception is raised. It is at the very end of `_sync_one_account`, after the analysis jobs run on whatever data did land, on the grounds that a partial graph plus a loud failure beats no graph at all. Raising before the analysis jobs instead would be a one-line move if reviewers prefer that.

### DIFF
--- a/cartography/intel/aws/__init__.py
+++ b/cartography/intel/aws/__init__.py
@@ -15,6 +15,8 @@ import aioboto3
 import boto3
 import botocore.exceptions
 import neo4j
+from neo4j.exceptions import ServiceUnavailable as Neo4jServiceUnavailable
+from neo4j.exceptions import SessionExpired as Neo4jSessionExpired
 
 from cartography.analysis.aws.analysis import AWS_EC2_ASSET_EXPOSURE_AUTO_SCALING_GROUP
 from cartography.analysis.aws.analysis import AWS_EC2_ASSET_EXPOSURE_JOBS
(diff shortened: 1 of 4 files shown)


cartography/util.py
```diff
@@ -645,6 +664,7 @@ def is_aws_region_skippable_client_error(
             error_message,
         )
         or error_code in AWS_REGION_ACCESS_DENIED_ERROR_CODES
+        or error_code in AWS_REGION_SERVER_ERROR_CODES
```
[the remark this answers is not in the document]
jychp: Keeping these out of the retry wrapper is deliberate: by the time the error reaches the decorator botocore has already exhausted its own retries (the traceback in the report shows `reached max retries: 3`), so another backoff cycle would just add up to 10 minutes per region waiting on a 500 that will never succeed, which is the exact cost this PR is removing.

cartography/util.py
```diff
@@ -645,6 +664,7 @@ def is_aws_region_skippable_client_error(
...
     )
```
kunaals, line 668: i agree that we shouldn’t add another retry cycle after botocore has already exhausted its retries. my question is whether we should propagate the final error into the new per-module isolation instead of returning an empty inventory. that would let unrelated modules continue without allowing normal cleanup to delete this module’s existing data. can we add a regression test for that path?

cartography/intel/aws/__init__.py
```diff
@@ -251,6 +349,16 @@ def _sync_one_account(
         stat_handler=stat_handler,
     )
 
+    # Raise only now that everything else in the account has been written, so that the data
+    # collected by the resource functions that did succeed survives the failure of one module.
```
kunaals, line 353: should this metadata advance when one of the requested modules failed? my concern is that the account may appear fully synced and fresh even though the run was partial. would it be safer to write this only when there are no collected failures?

cartography/intel/aws/__init__.py
```diff
@@ -197,45 +272,68 @@ def _sync_one_account(
...
+                    aioboto3_session_factory=aioboto3_session_factory,
+                ),
+                resource_function_failures,
             )
         elif func_name in ["permission_relationships", "resourcegroupstaggingapi"]:
             continue
         else:
-            RESOURCE_FUNCTIONS[func_name](**sync_args)
+            _run_resource_function(
+                func_name,
```
kunaals, line 293: the analysis part of the earlier cubic comment looks fixed. i’m still wondering about dependent resource modules, though. if a prerequisite fails, should we skip only the modules listed as depending on it while continuing with everything else?

kunaals (changes requested):

kunaals:
superseded by https://github.com/cartography-cncf/cartography/pull/3238

PR #3247, opened 2026-09-08
fix(rules): exclude inactive delegated admins

kunaals:
### Type of change

### Summary
Exclude explicitly inactive Google Workspace administrator accounts from delegated-admin findings. This applies the existing UserAccount ontology activity signal consistently to result, visual, and count queries for:

- Admins without enforced 2-Step Verification
- Super Admin accounts also marked as delegated admins
- Admin authorization of AI apps

Unknown activity remains eligible, matching existing ontology-backed identity rules and avoiding false negatives when ontology properties have not yet been projected.

#### General

#### Proof of functionality

### Notes for reviewers

This reuses the existing boolean `_ont_active` property; no new enum or schema field is needed.


cartography/rules/data/rules/cis_google_workspace.py
```diff
@@ -128,14 +128,15 @@ class AdminWithout2SVOutput(Finding):
...
-        "Detects Google Workspace admin accounts that do not have 2-Step Verification enforced. "
+        "Detects active Google Workspace admin accounts that do not have 2-Step Verification enforced. "
         "Admin accounts have elevated privileges and are high-value targets for attackers. "
         "Requires admin.directory.user.security scope."
     ),
     cypher_query="""
     MATCH (t:GoogleWorkspaceTenant)-[:RESOURCE]->(u:GoogleWorkspaceUser)
     WHERE
         (coalesce(u.is_admin, false) = true OR coalesce(u.is_delegated_admin, false) = true)
+        AND coalesce(u._ont_active, true) = true
```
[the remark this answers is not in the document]
kunaals: this is dumb

cartography/rules/data/rules/cis_google_workspace.py
```diff
@@ -354,12 +357,14 @@ class SuperAdminDualRoleOutput(Finding):
...
-        "Finds accounts that are simultaneously Super Admins and delegated admins, "
+        "Finds active accounts that are simultaneously Super Admins and delegated admins, "
         "which violates the benchmark's dedicated-admin-account guidance."
     ),
     cypher_query="""
     MATCH (t:GoogleWorkspaceTenant)-[:RESOURCE]->(u:GoogleWorkspaceUser)
-    WHERE coalesce(u.is_admin, false) = true AND coalesce(u.is_delegated_admin, false) = true
+    WHERE coalesce(u.is_admin, false) = true
+      AND coalesce(u.is_delegated_admin, false) = true
+      AND coalesce(u._ont_active, true) = true
```
jychp, line 367: Apply the same activity filter to the two super-admin threshold facts in this file. `_gw_super_admin_count_too_low` (line 233) and `_gw_super_admin_count_too_high` (line 294) still count suspended and archived Super Admins, so a tenant with 5 Super Admins of which 2 are suspended keeps failing CIS 1.1.2, and a tenant with 1 active plus 1 suspended Super Admin keeps passing CIS 1.1.1. The PR summary claims the signal is applied consistently; today it is not.

cartography/rules/data/rules/cis_google_workspace.py
```diff
@@ -151,12 +152,14 @@ class AdminWithout2SVOutput(Finding):
...
         (coalesce(u.is_admin, false) = true OR coalesce(u.is_delegated_admin, false) = true)
+        AND coalesce(u._ont_active, true) = true
         AND coalesce(u.is_enforced_in_2_sv, false) = false
     RETURN *
     """,
     cypher_count_query="""
     MATCH (u:GoogleWorkspaceUser)
-    WHERE coalesce(u.is_admin, false) = true OR coalesce(u.is_delegated_admin, false) = true
+    WHERE (coalesce(u.is_admin, false) = true OR coalesce(u.is_delegated_admin, false) = true)
+      AND coalesce(u._ont_active, true) = true
```
jychp, line 162: Filter `_gw_user_2sv_not_enforced` (line 59) the same way, or state in the PR description why suspended and archived users must stay in the all-users 2SV rule. Right now the same account is a false positive there and excluded here.

cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -386,6 +388,7 @@ class NistAiAdminAuthorizationsOutput(Finding):
         toLower(coalesce(app._ont_client_id, app.client_id, app.app_id, app.id, '')) AS normalized_client_id
     WHERE
         (coalesce(u.is_admin, false) = true OR coalesce(u.is_delegated_admin, false) = true)
+        AND coalesce(u._ont_active, true) = true
```
jychp, line 386: This count query repeats every predicate of the finding query, including the AI-term match, so `total_assets` always equals the number of failing apps and the compliance ratio is pinned at 100%. Drop the AI-term predicate from the denominator and keep only the population scope (admin-authorized `ThirdPartyApp` nodes). Pre-existing, but this PR adds a fourth predicate to both sides.

tests/unit/rules/test_cis_google_workspace.py
```diff
@@ -102,6 +106,19 @@ def test_admin_2sv_rule_includes_delegated_admins():
     assert "u.is_delegated_admin" in fact.cypher_count_query
 
 
+def test_delegated_admin_rules_exclude_inactive_accounts():
```
jychp, line 109: These assertions only prove the substring is present, not that an inactive admin is excluded; they pass even if the predicate lands in the wrong clause. Add an integration test under `tests/integration/rules/` that seeds an active and a suspended admin and asserts only the active one is returned, as `tests/integration/rules/test_mfa_missing.py` does for `_ont_active`.

jychp (approved):
approved with some comments

cartography/rules/data/rules/cis_google_workspace.py
jychp, line 139: `_ont_active` on `GoogleWorkspaceUser` is `NOT suspended AND NOT archived`, so this drops suspended admins too, and a suspended account keeps its admin role and regains access the moment another admin unsuspends it, with 2SV still unenforced. Restrict the exclusion to `archived`, or state the accepted risk in the fact description.

cartography/rules/data/rules/cis_google_workspace.py
```diff
@@ -368,12 +373,15 @@ class SuperAdminDualRoleOutput(Finding):
...
-    WHERE coalesce(u.is_admin, false) = true AND coalesce(u.is_delegated_admin, false) = true
+    WHERE coalesce(u.is_admin, false) = true
+      AND coalesce(u.is_delegated_admin, false) = true
+      AND coalesce(u._ont_active, true) = true
     RETURN *
     """,
     cypher_count_query="""
     MATCH (u:GoogleWorkspaceUser)
     WHERE coalesce(u.is_admin, false) = true
+      AND coalesce(u._ont_active, true) = true
```
jychp, line 384: Pair every `coalesce(u._ont_active, true)` in this PR with `AND NOT coalesce(u._ont_inactive, false)`, the idiom already used by `mfa_missing.py` and `unmanaged_accounts.py`. `GoogleWorkspaceUser` does not map `inactive` today, so it changes no results and keeps the activity-aware rules textually identical.

cartography/rules/data/rules/nist_ai_rmf.py
```diff
@@ -331,7 +331,7 @@ class NistAiAdminAuthorizationsOutput(Finding):
     id="gw_nist_ai_admin_app_authorizations",
     name="Google Workspace admins authorizing AI-related apps",
     description=(
-        "Finds Google Workspace administrator accounts that have authorized "
+        "Finds active Google Workspace administrator accounts that have authorized "
```
jychp, line 334: Update `ai_admin_app_authorizations.description` (line 414) as well. It still says "Identifies privileged Google Workspace identities that have authorized AI-related third-party applications", which now overstates the scope shipped by this fact.

tests/unit/rules/test_cis_google_workspace.py
```diff
@@ -26,9 +26,13 @@ def test_rules_registered_and_metadata():
...
         googleworkspace_admins_without_enforced_2sv,
     ]
+    expected_versions = {
+        "googleworkspace_admins_without_enforced_2sv": "1.0.1",
+        "googleworkspace_super_admin_accounts_used_for_daily_admin": "1.0.1",
+    }
 
     for rule in rules:
-        assert rule.version == "1.0.0"
+        assert rule.version == expected_versions.get(rule.id, "1.0.0")
```
jychp, line 35: Index `expected_versions` directly and list all five rule ids. `.get(rule.id, "1.0.0")` falls back silently on a typo'd key or a renamed rule, so a stale entry can never fail this assertion.

jychp (changes requested):

cartography/rules/data/rules/cis_google_workspace.py
kunaals: keeping these as-is because cis 1.1.1 and 1.1.2 evaluate super admin account assignments, not current sign-in eligibility. the activity filter here is limited to findings whose risk requires a currently usable privileged identity.

tests/unit/rules/test_cis_google_workspace.py
kunaals: good catch. added synthetic neo4j coverage for active, suspended, and archived admins across finding, visual, and count queries.

cartography/rules/data/rules/cis_google_workspace.py
kunaals: keeping this as-is because this control explicitly evaluates all provisioned users, while this change is scoped to privileged identity findings. changing that population should be a separate policy decision.

cartography/rules/data/rules/nist_ai_rmf.py
kunaals: good catch. the count now covers all apps authorized by active privileged users, while the ai matcher only defines the failing subset.

cartography/rules/data/rules/cis_google_workspace.py
kunaals: not adding this because `GoogleWorkspaceUser` does not map `_ont_inactive`; `_ont_active` is its canonical boolean. checking an absent duplicate property would not improve correctness.

tests/unit/rules/test_cis_google_workspace.py
kunaals: keeping the default because only version exceptions need listing. a typo in either bumped rule id falls back to `1.0.0` and fails against its actual `1.0.1`; enumerating unchanged defaults would duplicate data without strengthening this change.

cartography/rules/data/rules/cis_google_workspace.py
kunaals: keeping the canonical `_ont_active` predicate here. this finding evaluates accounts that can currently authenticate; reactivation without 2sv is a separate lifecycle-policy concern.

cartography/rules/data/rules/nist_ai_rmf.py
kunaals: good catch. updated the rule description to make the active-account scope explicit.

tests/integration/rules/test_google_workspace_active_admins.py
```diff
@@ -0,0 +1,186 @@
...
+
+    admin_2sv_fact = googleworkspace_admins_without_enforced_2sv.facts[0]
+    dual_role_fact = googleworkspace_super_admin_accounts_used_for_daily_admin.facts[0]
+
+    # Act
+    activity = _run_query(
+        neo4j_session,
+        """
+        MATCH (u:GoogleWorkspaceUser)
+        RETURN u.id AS id, u._ont_active AS active
```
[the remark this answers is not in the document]
kunaals: no change needed here: `load_googleworkspace_users()` calls the schema loader, and its generated ingestion query writes ontology properties inline from the google workspace mapping. the clean-graph integration test passes and asserts `_ont_active` before exercising the rules.

jychp (approved):

PR #3248, opened 2026-09-08
fix(salesforce): skip inaccessible connected apps

kunaals:
### Type of change

### Summary
Salesforce can hide `ConnectedApplication`, `OAuthToken`, or setup-only connected-app fields from an integration user. These optional queries currently abort the module after core organization, user, role, permission-set, and group data has loaded.

This change warns and skips only the connected-app stage for the access-denial shapes Salesforce returns (`INSUFFICIENT_ACCESS`, `INSUFFICIENT_ACCESS_OR_READONLY`, object-specific `INVALID_TYPE`, and `INVALID_FIELD` for the known setup-only fields). Classification uses structured Salesforce error codes and messages, so request URLs and SOQL cannot affect control flow. It preserves existing connected-app data by skipping cleanup and continues to raise unrelated query, transport, and response errors. The handled warning omits the org URL and SOQL.

### Related issues or links

None.

### How was this tested?

- 24 Salesforce unit and integration tests, including realistic Salesforce JSON errors wrapped in URL-bearing `requests` exceptions
- two-update-tag Neo4j integration coverage proving an OAuth-token access failure preserves the last complete connected-app snapshot
- full `make test_lint`
- Salesforce sandbox sync into an isolated local Neo4j database on the preceding commit; this follow-up changes classification and warning wording only

Sanitized live outcome (warning text normalized because the original included environment-specific exception details):
```text
INFO  Starting sync stage salesforce
INFO  Loaded organization, profiles, roles, users, permission sets, and groups
WARNING  Connected-app stage skipped because a required setup field was unavailable
INFO  Finishing sync stage salesforce
INFO  Finishing sync
```

#### General

#### Proof of functionality

- [x] Included sanitized Cartography sync logs demonstrating successful synchronization of the modified connector behavior.

### Notes for reviewers

The skipped stage deliberately does not run cleanup because its fetched scope is incomplete. Connected apps and OAuth tokens remain one snapshot because cleanup after a partial token fetch could delete valid `AUTHORIZED` relationships from the last complete sync.

### DIFF
--- a/cartography/intel/salesforce/connectedapps.py
+++ b/cartography/intel/salesforce/connectedapps.py
@@ -59,8 +59,8 @@ def sync(
             raise
         logger.warning(
             "Skipping Salesforce connected apps because required objects or setup "
-            "fields are unavailable to the integration user: %s",
-            exc,
+            "fields are unavailable to the integration user; preserving previously "
+            "synced connected app data",
         )
         return
     apps = transform(apps, tokens)
--- a/cartography/intel/salesforce/connectedapps.py
+++ b/cartography/intel/salesforce/connectedapps.py
@@ -8,6 +8,7 @@ import requests
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.intel.salesforce.util import get_salesforce_error_codes
+from cartography.intel.salesforce.util import get_salesforce_error_messages
 from cartography.intel.salesforce.util import parse_sf_datetime
 from cartography.intel.salesforce.util import SalesforceClient
 from cartography.models.salesforce.connectedapp import SalesforceConnectedAppSchema
@@ -28,14 +29,17 @@ def _is_access_error(exc: requests.HTTPError) -> bool:
     error_codes = get_salesforce_error_codes(exc)
     if error_codes & {"INSUFFICIENT_ACCESS", "INSUFFICIENT_ACCESS_OR_READONLY"}:
         return True
-    error_message = str(exc)
+    error_messages = get_salesforce_error_messages(exc)
     if "INVALID_TYPE" in error_codes:
         return any(
-            object_name in error_message
+            f"sObject type '{object_name}' is not supported." in message
+            for message in error_messages
             for object_name in ("ConnectedApplication", "OAuthToken")
         )
     return "INVALID_FIELD" in error_codes and any(
-        field_name in error_message for field_name in _SETUP_ONLY_APP_FIELDS
+        f"No such column '{field_name}' on entity 'ConnectedApplication'." in message
+        for message in error_messages
+        for field_name in _SETUP_ONLY_APP_FIELDS
     )
 
 
@@ -45,6 +49,8 @@ def sync(
     client: SalesforceClient,
     common_job_parameters: dict[str, Any],
 ) -> None:
+    # Keep apps and tokens as one snapshot: cleanup after a partial token fetch would
+    # delete valid AUTHORIZED relationships from the last complete sync.
     try:
         apps = get_apps(client)
         tokens = get_oauth_tokens(client)
@@ -52,8 +58,8 @@ def sync(
         if not _is_access_error(exc):
             raise
         logger.warning(
-            "Skipping Salesforce connected apps because the integration user "
-            "cannot access ConnectedApplication or OAuthToken: %s",
+            "Skipping Salesforce connected apps because required objects or setup "
+            "fields are unavailable to the integration user: %s",
             exc,
         )
         return
--- a/cartography/intel/salesforce/connectedapps.py
+++ b/cartography/intel/salesforce/connectedapps.py
@@ -16,16 +16,26 @@ from cartography.util import timeit
 _APP_FIELDS = (
     "Id, Name, OptionsAllowAdminApprovedUsersOnly, CreatedDate, LastModifiedDate"
 )
+_SETUP_ONLY_APP_FIELDS = (
+    "OptionsAllowAdminApprovedUsersOnly",
+    "CreatedDate",
+    "LastModifiedDate",
+)
 logger = logging.getLogger(__name__)
 
 
 def _is_access_error(exc: requests.HTTPError) -> bool:
     error_codes = get_salesforce_error_codes(exc)
-    if "INSUFFICIENT_ACCESS_OR_READONLY" in error_codes:
+    if error_codes & {"INSUFFICIENT_ACCESS", "INSUFFICIENT_ACCESS_OR_READONLY"}:
         return True
-    return "INVALID_TYPE" in error_codes and any(
-        object_name in str(exc)
-        for object_name in ("ConnectedApplication", "OAuthToken")
+    error_message = str(exc)
+    if "INVALID_TYPE" in error_codes:
+        return any(
+            object_name in error_message
+            for object_name in ("ConnectedApplication", "OAuthToken")
+        )
+    return "INVALID_FIELD" in error_codes and any(
+        field_name in error_message for field_name in _SETUP_ONLY_APP_FIELDS
     )
 
 
--- a/cartography/intel/salesforce/connectedapps.py
+++ b/cartography/intel/salesforce/connectedapps.py
@@ -1,3 +1,4 @@
+import logging
 from collections import defaultdict
 from typing import Any
 
@@ -7,12 +8,15 @@ from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
 from cartography.intel.salesforce.util import parse_sf_datetime
 from cartography.intel.salesforce.util import SalesforceClient
+from cartography.intel.salesforce.util import SalesforceQueryError
 from cartography.models.salesforce.connectedapp import SalesforceConnectedAppSchema
 from cartography.util import timeit
 
 _APP_FIELDS = (
     "Id, Name, OptionsAllowAdminApprovedUsersOnly, CreatedDate, LastModifiedDate"
 )
+_ACCESS_ERROR_CODES = frozenset({"INSUFFICIENT_ACCESS_OR_READONLY", "INVALID_TYPE"})
+logger = logging.getLogger(__name__)
 
 
 @timeit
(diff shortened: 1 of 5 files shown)


cartography/intel/salesforce/connectedapps.py
```diff
@@ -13,6 +16,27 @@
...
+        return True
+    error_message = str(exc)
+    if "INVALID_TYPE" in error_codes:
+        return any(
+            object_name in error_message
+            for object_name in ("ConnectedApplication", "OAuthToken")
+        )
+    return "INVALID_FIELD" in error_codes and any(
+        field_name in error_message for field_name in _SETUP_ONLY_APP_FIELDS
+    )
```
jychp: `str(exc)` contains the request URL, which contains the whole SOQL, so both substring guards are vacuous and misclassify real errors.

`requests` builds the message as `400 Client Error: ... for url: .../query?q=SELECT+Id%2C+Name%2C+OptionsAllowAdminApprovedUsersOnly%2C+CreatedDate%2C+LastModifiedDate+FROM+ConnectedApplication`. Verified locally:

```
MSG: 400 Client Error: Bad Request for url: https://example.my.salesforce.com/services/data/v60.0/query?q=SELECT+Id%2C+Name%2C+OptionsAllowAdminApprovedUsersOnly%2C+CreatedDate%2C+LastModifiedDate+FROM+ConnectedApplication
CreatedDate in msg? True
```

So every `INVALID_FIELD` raised from `get_apps()` matches `CreatedDate`, and every `INVALID_TYPE` matches `ConnectedApplication`, regardless of which field or object Salesforce actually complained about. A genuine schema break (renamed field after an `API_VERSION` bump, typo in `_APP_FIELDS`) gets silently downgraded to a warning and the stage is skipped forever.

Match on the structured Salesforce error `message` values, not on the formatted exception string.

cartography/intel/salesforce/util.py
```diff
@@ -15,6 +15,38 @@
...
+
+
+def get_salesforce_error_codes(error: requests.HTTPError) -> frozenset[str]:
+    if error.response is None:
+        return frozenset()
+    return frozenset(
+        item["errorCode"]
+        for item in _salesforce_errors(error.response)
+        if isinstance(item.get("errorCode"), str)
+    )
```
jychp, line 41: Add a `get_salesforce_error_messages(error) -> tuple[str, ...]` sibling that returns the structured `message` fields, and have `_is_access_error` consume it instead of `str(exc)`. Classification must never read a string that embeds the request URL or is truncated at `_MAX_ERROR_DETAIL_LENGTH`.

cartography/intel/salesforce/connectedapps.py
```diff
@@ -13,6 +16,27 @@
 _APP_FIELDS = (
     "Id, Name, OptionsAllowAdminApprovedUsersOnly, CreatedDate, LastModifiedDate"
 )
+_SETUP_ONLY_APP_FIELDS = (
+    "OptionsAllowAdminApprovedUsersOnly",
+    "CreatedDate",
+    "LastModifiedDate",
+)
```
jychp: **nitpick** - `_SETUP_ONLY_APP_FIELDS` restates three of the five names already in `_APP_FIELDS`. Build `_APP_FIELDS` from a tuple of field names so the two cannot drift.

cartography/intel/salesforce/connectedapps.py
```diff
@@ -21,8 +45,18 @@ def sync(
     client: SalesforceClient,
     common_job_parameters: dict[str, Any],
 ) -> None:
-    apps = get_apps(client)
-    tokens = get_oauth_tokens(client)
+    try:
+        apps = get_apps(client)
+        tokens = get_oauth_tokens(client)
```
jychp, line 56: Why are `get_apps()` and `get_oauth_tokens()` under one `try`? An org that exposes `ConnectedApplication` but hides `OAuthToken` loses all connected-app nodes, not just the `AUTHORIZED_USER` edges. Split the two calls and load apps with an empty token join when only `OAuthToken` is denied, or state in a comment why all-or-nothing is the intended contract.

cartography/intel/salesforce/connectedapps.py
```diff
@@ -21,8 +45,18 @@ def sync(
...
+    except requests.HTTPError as exc:
+        if not _is_access_error(exc):
+            raise
+        logger.warning(
+            "Skipping Salesforce connected apps because the integration user "
+            "cannot access ConnectedApplication or OAuthToken: %s",
+            exc,
+        )
```
jychp, line 64: The message asserts the user "cannot access ConnectedApplication or OAuthToken", which is wrong for the `INVALID_FIELD` branch where the objects are readable and only setup fields are hidden. Also, this string does not match the sanitized log pasted in the PR description ("cannot access the required setup fields"), so that evidence appears to come from an earlier commit. Reword to cover both causes and re-attach a log from the current code.

tests/unit/cartography/intel/salesforce/test_connectedapps.py
```diff
@@ -0,0 +1,120 @@
...
+def _query_error(error_code: str, object_name: str) -> requests.HTTPError:
+    message = f"sObject type '{object_name}' is not supported."
+    response = MagicMock(status_code=400, text=message)
+    response.json.return_value = [
+        {"message": message, "errorCode": error_code},
+    ]
+    return requests.HTTPError(
+        f"Salesforce query failed with HTTP 400: {error_code}: {message}",
+        response=response,
+    )
```
jychp: `_query_error` fabricates a message that production never produces: the real exception text from `query_all` starts with the `requests` `... for url: <full URL including the SOQL>` prefix before the appended Salesforce detail. Because of that omission, `test_sync_raises_unexpected_query_errors[INVALID_FIELD/UnexpectedField]` passes for a reason that does not hold at runtime (see the `_is_access_error` comment). Build the fixture from a real URL-bearing `raise_for_status()` message so the negative case actually exercises the classifier.

tests/unit/cartography/intel/salesforce/test_connectedapps.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Act
+    connectedapps.sync(
+        MagicMock(),
+        client,
+        {"ORG_ID": "00D000000000001", "UPDATE_TAG": 123456789},
+    )
+
+    # Assert
+    assert client.query_all.call_count == 2
```
jychp: Drop the `call_count` assertion; it pins call ordering rather than behavior. `mock_load`/`mock_cleanup` not being called already proves the skip.

tests/unit/cartography/intel/salesforce/test_connectedapps.py
```diff
@@ -0,0 +1,120 @@
...
+    mock_load.assert_not_called()
+    mock_cleanup.assert_not_called()
+
+
+@pytest.mark.parametrize(
+    "error",
+    [
+        _query_error("MALFORMED_QUERY", "ConnectedApplication"),
+        _query_error("INVALID_FIELD", "UnexpectedField"),
```
jychp: `sObject type 'UnexpectedField' is not supported.` is an `INVALID_TYPE` message, not an `INVALID_FIELD` one. Use the real shape: `No such column 'UnexpectedField' on entity 'ConnectedApplication'.`

tests/unit/cartography/intel/salesforce/test_connectedapps.py
```diff
@@ -0,0 +1,120 @@
...
+
+@patch.object(connectedapps, "cleanup")
+@patch.object(connectedapps, "load_connected_apps")
+def test_sync_skips_when_connected_app_setup_fields_are_hidden(mock_load, mock_cleanup):
+    # Arrange
+    client = MagicMock()
+    client.query_all.side_effect = _query_error(
+        "INVALID_FIELD",
+        "OptionsAllowAdminApprovedUsersOnly",
+    )
```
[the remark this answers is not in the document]
kunaals: eh i feel like this is not needed

jychp (changes requested):

cartography/intel/salesforce/connectedapps.py
kunaals: fixed this to classify only the structured salesforce response messages, so the request url and SOQL can’t affect the decision

cartography/intel/salesforce/util.py
kunaals: added the structured message accessor- formatted and truncated exception text is now diagnostic-only

cartography/intel/salesforce/connectedapps.py
kunaals: keeping these as one snapshot because cleanup after a partial OAuth token fetch could delete valid authorized relationships from the last complete sync. added a comment explaining that

cartography/intel/salesforce/connectedapps.py
kunaals: good catch, reworded this to cover both unavailable objects and hidden setup fields, clarified the validation evidence in the PR desc

tests/unit/cartography/intel/salesforce/test_connectedapps.py
kunaals: fixture now constructs a real requests error, and the unrelated invalid-field case remains fatal

tests/unit/cartography/intel/salesforce/test_connectedapps.py
kunaals: removed the redundant call-count assertion

tests/unit/cartography/intel/salesforce/test_connectedapps.py
kunaals: updated this to the real response shape

tests/unit/cartography/intel/salesforce/test_connectedapps.py
```diff
@@ -0,0 +1,165 @@
...
+    soql: str = _CONNECTED_APP_QUERY,
+) -> requests.HTTPError:
+    request = requests.Request(
+        "GET",
+        "https://example.my.salesforce.com/services/data/v60.0/query",
+        params={"q": soql},
+    ).prepare()
+    response = requests.Response()
+    response.status_code = 400
+    response._content = json.dumps(  # noqa: SLF001
```
jychp: `SLF001` is a ruff/flake8-self code, and this repo lints with flake8 (`setup.cfg [flake8]`, `.pre-commit-config.yaml`), which silently ignores unknown noqa codes, so this suppresses nothing and appears nowhere else in the tree. Worth dropping.

jychp (approved):
approved with one nit

tests/unit/cartography/intel/salesforce/test_connectedapps.py
kunaals: good catch. removed the ineffective suppression.

PR #3249, opened 2026-09-08
fix(socketdev): scope fix queries to alert vulnerabilities

kunaals:
### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### Summary

Socket.dev fix ingestion queried every alert-bearing repository with `vulnerability_ids=*`. That asks the provider to calculate fixes for every vulnerability even though the preceding alerts sync already returns the CVE and GHSA identifiers Cartography can correlate.

Scope each repository request to its deduplicated CVE/GHSA identifiers and send them in bounded batches. Repositories with only non-vulnerability alerts no longer make unnecessary fixes requests, while all alerts continue to be ingested by the unchanged alerts sync. Cleanup still runs when there are no fix-eligible vulnerabilities.

Fix relationships are resolved by repository, vulnerable package, and version so the same advisory can safely affect multiple packages in one repository.

The fixes endpoint can also intermittently exceed the request timeout. A single shared session retries transient read, rate-limit, and server failures with bounded local backoff; provider-supplied `Retry-After` values cannot stall the sync indefinitely.

This also makes urllib3 2.x an explicit dependency and raises the requests floor to 2.30. urllib3 2.x requires OpenSSL 1.1.1 or newer.

### Related issues or links

- [Socket.dev Fetch fixes API](https://docs.socket.dev/reference/fetch-fixes)

### Breaking changes

None.

### How was this tested?

- Added behavioral unit tests for batching, transient retries, bounded rate-limit handling, incomplete-scan cleanup safety, and duplicate-advisory package matching
- Ran the Socket.dev integration suite
- Ran the full lint and type-check suite
- Completed a real-provider sync against an isolated Neo4j database, including fix-to-alert and fix-to-dependency relationships

#### General

- [x] I have run the linter and formatter.
- [x] I have added or updated tests that prove the fix is effective.
- [x] My commits include the Developer Certificate of Origin sign-off.

#### Proof

- [ ] I have included screenshots for user-facing changes.
- [x] I have included automated test coverage.

#### Connector changes

- [x] I tested against the real provider.
- [ ] I included sanitized successful sync logs.
- [x] I removed credentials, personal data, account identifiers, and other sensitive values from the contribution.

#### Node or relationship changes

- [x] I updated the data model documentation.
- [x] I updated the schema.

#### New module

- [ ] I added module documentation.

### Notes for reviewers

Socket documents `vulnerability_ids` as a comma-separated list of GHSA or CVE identifiers or `*`. This change uses bounded explicit-ID batches because Cartography only creates fix relationships for identifiers already present on synced alerts. Retries are limited to idempotent GET requests and cap provider-supplied `Retry-After` delays.


cartography/intel/socketdev/fixes.py
```diff
@@ -207,17 +228,17 @@ def sync_fixes(
...
             repo_slug_val,
         )
-        # Use "*" to fetch fixes for all vulnerabilities in the repo
+        # Wildcard fix queries become expensive for repositories with many alerts.
         raw_response = get(
             api_token,
             org_slug,
             repo_slug_val,
-            "*",
+            ",".join(sorted(vulnerability_ids)),
```
jychp: Chunk the identifier list instead of sending every CVE/GHSA of a repo in one query string. Alerts page at 1000 per request with no cap on the total, so a repo with a few hundred vulnerability alerts already builds a multi-KB URL and a large one gets a 414/431 that `_RETRY_STATUS_CODES` does not cover. That makes this fail harder than the `*` query it replaces, on exactly the repos the PR targets.

cartography/intel/socketdev/fixes.py
```diff
@@ -25,7 +45,7 @@ def get(
     Fetch fixes for the given vulnerabilities in a repository.
     Returns the raw API response dict containing fixDetails.
     """
-    response = requests.get(
+    response = _create_session().get(
```
jychp: Build the session once in `sync_fixes` and pass it into `get()`, as `intel/huntress/util.py` and `intel/railway/__init__.py` do. As written every call constructs and abandons a `Session` plus its connection pool without `close()`, so the per-repo loop gets no keep-alive reuse and leaks sockets.

cartography/intel/socketdev/fixes.py
```diff
@@ -12,6 +14,24 @@
 logger = logging.getLogger(__name__)
 _TIMEOUT = (60, 60)
 _BASE_URL = "https://api.socket.dev/v0"
+_RETRY_STATUS_CODES = (408, 429, 500, 502, 503, 504)
```
jychp, line 19: Drop 429 from the forcelist or bound the wait. urllib3 sleeps the full `Retry-After` inside the adapter with no cap; `intel/railway/__init__.py` excludes 429 for exactly this reason. Combined with `_TIMEOUT = (60, 60)` and three retries, one repo can stall a sync for an unbounded stretch.

cartography/intel/socketdev/fixes.py
```diff
@@ -12,6 +14,24 @@
...
+    retry_policy = Retry(
+        total=3,
+        connect=3,
+        read=3,
+        status=3,
+        other=0,
+        allowed_methods=["GET"],
+        status_forcelist=_RETRY_STATUS_CODES,
+        backoff_factor=1,
+        respect_retry_after_header=True,
```
jychp: Drop `respect_retry_after_header=True`; it is the urllib3 default.

cartography/intel/socketdev/fixes.py
```diff
@@ -174,27 +194,28 @@ def sync_fixes(
...
         if cve_id:
             alerts_by_vuln[(cve_id, repo_slug_val)] = alert_id
+            vulnerability_ids_by_repo.setdefault(repo_slug_val, set()).add(cve_id)
         ghsa_id = alert.get("ghsa_id")
         if ghsa_id:
             alerts_by_vuln[(ghsa_id, repo_slug_val)] = alert_id
+            vulnerability_ids_by_repo.setdefault(repo_slug_val, set()).add(ghsa_id)
         key = alert.get("key")
         if key:
             alerts_by_vuln[(key, repo_slug_val)] = alert_id
```
jychp: Is this branch still reachable? The request now asks only for CVE/GHSA identifiers, so `fixDetails` can no longer come back keyed by an alert `key`

cartography/intel/socketdev/fixes.py
```diff
@@ -174,27 +194,28 @@ def sync_fixes(
...
 
-    if not repo_slugs:
-        logger.info("No alerts with repository info found, skipping fixes sync")
+    if not vulnerability_ids_by_repo:
+        logger.info("No repository vulnerabilities found, skipping fixes sync")
```
jychp: Fix the message: the sync no longer skips, it prunes. Say that no repository carries a CVE/GHSA identifier and stale fixes are being cleaned up.

cartography/intel/socketdev/fixes.py
```diff
@@ -207,17 +228,17 @@ def sync_fixes(
...
 
     all_fixes: list[dict[str, Any]] = []
-    for repo_slug_val in repo_slugs:
+    for repo_slug_val, vulnerability_ids in sorted(vulnerability_ids_by_repo.items()):
         logger.debug(
             "Fetching fixes for repo '%s'",
             repo_slug_val,
         )
-        # Use "*" to fetch fixes for all vulnerabilities in the repo
+        # Wildcard fix queries become expensive for repositories with many alerts.
```
jychp: Rewrite this comment to describe what the code does now. It currently explains the approach that was removed, not the scoped query below it.

tests/integration/cartography/intel/socketdev/test_fixes.py
```diff
@@ -22,6 +23,17 @@
 TEST_ORG_SLUG = "acme-corp"
 
 
+def test_fixes_session_retries_transient_failures():
```
jychp: Move this to `tests/unit/cartography/intel/socketdev/` and assert behavior instead of construction. It needs no Neo4j, reaches into the private `_create_session`, and only restates the constructor arguments written a few lines away in the source, so it cannot fail for any reason worth catching.

jychp (changes requested):
some blocking comments on batching and potential perf issues and some nitpicking comments

cartography/intel/socketdev/fixes.py
kunaals: agreed. removed the redundant explicit setting.

cartography/intel/socketdev/fixes.py
kunaals: good catch. fixed this by batching vulnerability identifiers in groups of 100, with coverage for the final partial batch.

cartography/intel/socketdev/fixes.py
kunaals: good catch. sync_fixes now creates one session, reuses it for every repository and batch, and closes it with the context manager.

cartography/intel/socketdev/fixes.py
kunaals: fixed. the message now says that no cve or ghsa identifiers were found and stale fixes are being cleaned.

cartography/intel/socketdev/fixes.py
kunaals: good catch. retries are bounded, and retry-after values are now capped at 8 seconds.

cartography/intel/socketdev/fixes.py
kunaals: yes, it was dead after scoping the request to cve and ghsa identifiers. removed it.

tests/integration/cartography/intel/socketdev/test_fixes.py
kunaals: good catch. moved this to the unit suite and replaced the constructor assertion with behavioral coverage for retries, capped delays, batching, and session reuse.

cartography/intel/socketdev/fixes.py
kunaals: fixed. the comment now describes the scoped identifier batches.

cartography/intel/socketdev/fixes.py
```diff
@@ -3,6 +3,9 @@
 
 import neo4j
 import requests
+from requests.adapters import HTTPAdapter
+from urllib3.response import BaseHTTPResponse
```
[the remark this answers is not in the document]
kunaals: made urllib3 2.x an explicit dependency and raised the requests floor to 2.30 so the declared versions stay compatible.

cartography/intel/socketdev/fixes.py
```diff
@@ -174,27 +207,27 @@ def sync_fixes(
...
         alert_id = alert["id"]
         repo_slug_val = alert.get("repo_slug")
         if not repo_slug_val:
             continue
-        repo_slugs.add(repo_slug_val)
         # Index by (vuln_id, repo_slug) for each known identifier
         cve_id = alert.get("cve_id")
         if cve_id:
             alerts_by_vuln[(cve_id, repo_slug_val)] = alert_id
+            vulnerability_ids_by_repo.setdefault(repo_slug_val, set()).add(cve_id)
```
jychp, line 245: `alerts_by_vuln` is last-write-wins on `(vuln_id, repo_slug)`, but Socket emits one alert per (package, version, vuln), so the same GHSA on two packages in one repo collapses to a single alert id that every fix entry returned for that GHSA then inherits. The resulting `(:SocketDevFix)-[:APPLIES_TO]->(:SocketDevAlert)` points at the wrong package's alert; keying on `(vuln_id, repo_slug, artifact_name, artifact_version)` and resolving from the purl at transform time would close the half of the gap the comment above claims to close. Pre-existing, so your call whether it belongs here or in a separate issue.

cartography/intel/socketdev/fixes.py
```diff
@@ -207,21 +240,28 @@ def sync_fixes(
...
-        )
-        # Use "*" to fetch fixes for all vulnerabilities in the repo
-        raw_response = get(
-            api_token,
-            org_slug,
-            repo_slug_val,
-            "*",
-        )
+    with _create_session(api_token) as api_session:
+        for repo_slug_val, vulnerability_ids in sorted(
```
jychp: This sorts `(str, set[str])` tuples and only works because the repo slugs are unique. `for repo_slug_val in sorted(vulnerability_ids_by_repo)` says the same thing without ever risking a set comparison.

cartography/intel/socketdev/fixes.py
```diff
@@ -12,11 +15,45 @@
 logger = logging.getLogger(__name__)
 _TIMEOUT = (60, 60)
 _BASE_URL = "https://api.socket.dev/v0"
+_RETRY_STATUS_CODES = (408, 429, 500, 502, 503, 504)
+_MAX_RETRY_AFTER_SECONDS = 8
+_VULNERABILITY_BATCH_SIZE = 100
```
jychp, line 22: Add a one-line comment saying where 100 comes from: a documented Socket.dev limit, or a margin on URL length. Otherwise the next person bumps it and lands back on the 414 this PR exists to avoid.

cartography/intel/socketdev/fixes.py
```diff
@@ -12,11 +15,45 @@
...
+    retry_policy = _CappedRetry(
+        total=3,
+        connect=3,
+        read=3,
+        status=3,
+        other=0,
+        allowed_methods=["GET"],
+        status_forcelist=_RETRY_STATUS_CODES,
+        backoff_factor=1,
+        backoff_max=_MAX_RETRY_AFTER_SECONDS,
```
jychp: With `backoff_factor=1` and `total=3` the backoff sequence is 0/2/4, so `backoff_max` never binds and this reuse is inert. Sharing the constant with the `Retry-After` cap reads as if the two were one knob.

cartography/intel/socketdev/fixes.py
```diff
@@ -12,11 +15,45 @@
 logger = logging.getLogger(__name__)
 _TIMEOUT = (60, 60)
 _BASE_URL = "https://api.socket.dev/v0"
+_RETRY_STATUS_CODES = (408, 429, 500, 502, 503, 504)
+_MAX_RETRY_AFTER_SECONDS = 8
+_VULNERABILITY_BATCH_SIZE = 100
+
+
+class _CappedRetry(Retry):
```
jychp, line 25: Railway solved the same unbounded-`Retry-After` problem by dropping 429 from the forcelist plus a hand-rolled loop, and gitlab has a third variant, so this makes three patterns for one problem. Worth considering a shared helper so the next module does not invent a fourth; not blocking.

pyproject.toml
```diff
@@ -55,7 +55,8 @@ dependencies = [
     "PyJWT[crypto]>=2.0.0",
     "cryptography>=45.0.0",
     "pyyaml>=5.3.1",
-    "requests>=2.22.0",
+    "requests>=2.30.0",
+    "urllib3>=2.0.0",
```
jychp, line 59: This is correct and overdue: roughly 25 modules already import urllib3 without declaring it, and `cartography/intel/orca/api.py:57` already passes the 2.x-only `backoff_max=`. Worth calling out in the description though, since it raises the floor for every install and urllib3 2.x drops OpenSSL < 1.1.1 while the commit is scoped as socketdev.

jychp (approved):
approved with some comments and potential followups

cartography/intel/socketdev/fixes.py
kunaals: good catch. fixed the lookup to include repository, package, and vulnerable version, resolving the package and version from each fix purl. added a duplicate-advisory regression test.

cartography/intel/socketdev/fixes.py
kunaals: fixed. iteration now sorts repository slugs and indexes the mapping separately.

cartography/intel/socketdev/fixes.py
kunaals: agreed. removed the inert backoff_max setting so the retry-after cap is the only use of that constant.

cartography/intel/socketdev/fixes.py
kunaals: added a comment documenting that 100 is a conservative url-length bound to avoid 414 responses, not a provider limit.

cartography/intel/socketdev/fixes.py
kunaals: agreed that the retry patterns should converge, but keeping that non-blocking refactor out of this focused fix.

pyproject.toml
kunaals: updated the pr description to call out the urllib3 2.x floor and its openssl 1.1.1 requirement.

cartography/models/socketdev/alert.py
```diff
@@ -127,6 +127,10 @@ class SocketDevAlertNodeProperties(CartographyNodeProperties):
         "artifact_type",
         description="Affected package ecosystem.",
     )
+    artifact_namespace: PropertyRef = PropertyRef(
+        "artifact_namespace",
+        description="Affected package namespace or scope.",
```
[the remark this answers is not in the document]
kunaals: good catch. updated the schema description to name the source `artifact.namespace` field.

cartography/intel/socketdev/fixes.py
```diff
@@ -170,31 +216,53 @@ def sync_fixes(
...
         alert_id = alert["id"]
         repo_slug_val = alert.get("repo_slug")
         if not repo_slug_val:
             continue
-        repo_slugs.add(repo_slug_val)
-        # Index by (vuln_id, repo_slug) for each known identifier
+        artifact_name = alert.get("artifact_name")
+        artifact_version = alert.get("artifact_version")
+        artifact_type = alert.get("artifact_type")
+        artifact_namespace = alert.get("artifact_namespace") or ""
```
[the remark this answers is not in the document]
kunaals: good catch. missing namespaces now remain `None` through ingestion and both lookup-key paths. i also added coverage for the empty provider value.

cartography/intel/socketdev/alerts.py
```diff
@@ -119,6 +119,7 @@ def transform(raw_alerts: list[dict[str, Any]]) -> list[dict[str, Any]]:
                 "artifact_name": _flatten_field(artifact.get("name")),
                 "artifact_version": _flatten_field(artifact.get("version")),
                 "artifact_type": _flatten_field(artifact.get("type")),
+                "artifact_namespace": _flatten_field(artifact.get("namespace")),
```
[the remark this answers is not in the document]
kunaals: this is generated rather than checked in: the contributor guide explicitly says not to edit module `schema.md` files manually. the local sphinx build generated the socket.dev schema with `artifact_namespace` and its description successfully.

PR #3250, opened 2026-09-09
fix(socketdev): link fixes to scoped dependencies

kunaals:
### Type of change

### Summary

Socket.dev fix ingestion manually parsed package URLs and matched only the package name and version. Percent-encoded scoped package URLs therefore did not match dependencies whose namespace and name are stored separately.

Reuse Cartography's existing package URL normalization for both sides of the lookup, while continuing to scope matches by repository. This restores `SHOULD_UPDATE_TO` relationships for scoped packages without allowing cross-repository matches.

### Related issues or links

- Follow-up to #3249

### Breaking changes

None.

### How was this tested?

- Added a regression test for an encoded scoped npm package URL
- Ran the Socket.dev integration suite
- Ran the full lint and type-check suite
- Validated against an isolated Neo4j database populated from a real provider environment; additional scoped-package relationships were created without removing existing relationships

#### General

#### Proof of functionality

### Notes for reviewers

This changes relationship matching only. It does not change node identity or schema. The shared normalizer already handles package type, namespace decoding, ecosystem-specific name normalization, and version.


tests/integration/cartography/intel/socketdev/test_fixes.py
```diff
@@ -22,6 +22,21 @@
 TEST_ORG_SLUG = "acme-corp"
 
 
+def test_build_dependency_id_matches_encoded_scoped_package():
```
jychp: Move this to `tests/unit/cartography/intel/socketdev/test_fixes.py`. It exercises a pure helper with no Neo4j session, and that file already exists. Add the `# Arrange` / `# Act and assert` phase comments while moving it.

tests/integration/cartography/intel/socketdev/test_fixes.py
```diff
@@ -22,6 +22,21 @@
...
+    dep_lookup = {
+        ("npm|@example/package|1.2.3", "frontend-app"): "dep-scoped",
```
jychp: Hardcoding the key proves only that `dict.get` works; nothing checks that `dependencies.transform()` actually emits `npm|@example/package|1.2.3`, which is the whole contract this PR relies on. Add a scoped npm dependency to `tests/data/socketdev/dependencies.py` and a matching encoded-purl fix entry to `tests/data/socketdev/fixes.py`, then assert the new `SHOULD_UPDATE_TO` rel in `test_sync_fixes`.

jychp (approved):
LGTM but tests are not really testing the change

tests/integration/cartography/intel/socketdev/test_fixes.py
kunaals: good point. moved this helper-only check to the unit suite and added the requested arrange/act phases.

tests/integration/cartography/intel/socketdev/test_fixes.py
kunaals: good catch. the fixture now runs a scoped dependency through dependencies.transform() and verifies the resulting should_update_to relationship in neo4j.

PR #3259, opened 2026-09-09
fix(socketdev): match repositories by provider identity

kunaals:
### Type of change

### Summary

Socket.dev repository responses can omit the workspace and GitHub integration metadata. When this happens, Cartography stores only the repository slug and cannot create the `SocketDevRepository -[:MONITORS]-> CodeRepository` relationship.

This change retrieves missing integration metadata from Socket.dev and matches repositories by their canonical GitHub URL. It keeps the existing Socket.dev `fullname` for alert and dependency relationships. Repositories without an authoritative GitHub identity are not guessed or linked. Standard relationship cleanup removes stale links when the integration metadata disappears.

The integration data covers both slug-only and canonical Socket.dev identities, duplicate repository names across GitHub owners, authoritative matching, and stale relationship cleanup.

### Related issues or links

None.

### How was this tested?

tested on real env and updated unit and int tests

#### General

#### Proof of functionality

### Notes for reviewers

Socket.dev documents GitHub as the supported repository integration type. The list endpoint can omit its integration metadata, so Cartography retrieves the repository detail only when that key is absent. These requests use the connector's retry policy and preserve workspace-qualified repository selection.


cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,22 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
...
+                    f"{_BASE_URL}/orgs/{org_slug}/repos/{repo['slug']}",
+                    params=(
+                        {"workspace": repo["workspace"]}
+                        if "workspace" in repo
+                        else None
+                    ),
+                    timeout=_TIMEOUT,
+                )
+                detail_response.raise_for_status()
+                repo = {**repo, **detail_response.json()}
```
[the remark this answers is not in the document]
kunaals: good catch. fixed this and added a regression test.

cartography/intel/socketdev/repositories.py
```diff
@@ -5,6 +5,7 @@
 import requests
```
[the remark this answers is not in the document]
kunaals: keeping this as-is because a Socket organization is not necessarily the GitHub owner. the current code uses Socket’s authoritative GitHub integration metadata instead.

tests/integration/cartography/intel/socketdev/test_dependencies.py
```diff
@@ -82,9 +82,9 @@ def test_sync_dependencies(mock_api, neo4j_session):
```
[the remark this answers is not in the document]
kunaals: stale from the earlier branch state. this pr no longer touches this file.

cartography/intel/socketdev/repositories.py
```diff
@@ -23,14 +31,18 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
     """
     all_repos: list[dict[str, Any]] = []
     page = 1
+    api_session = requests.Session()
```
jychp: Reuse `_create_session()` from `fixes.py` (move it to a shared `socketdev/util.py`) and wrap it in `with`. This session is never closed, and `_RETRY_POLICY` drops the `_CappedRetry` `Retry-After` cap, so a `429` carrying `Retry-After: 3600` stalls the whole sync.

cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,28 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
         data = response.json()
 
         results = data.get("results", [])
-        all_repos.extend(results)
+        for repo in results:
```
jychp: Bound this enrichment loop. On a deployment that omits `integration_meta` it fires one sequential request per repository, each with up to ~30s of retry backoff; parallelize with a bounded thread pool and log the enrichment count up front.

cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,28 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
...
+            # Some Socket deployments omit this key from list responses. An explicit
```
jychp: What confirms that an explicit null `integration_meta` always means no VCS integration? If Socket.dev ever returns null for an integrated repository, this silently drops a `MONITORS` edge that the old `_ont_fullname` match produced.

cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,28 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
...
+            # null means the repository has no GitHub integration.
+            if "integration_meta" not in repo and repo.get("slug"):
```
jychp: Log at debug when a repository is skipped because `slug` is missing. Right now the enrichment is dropped with no trace.

cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,28 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
...
+                detail_response = api_session.get(
+                    f"{_BASE_URL}/orgs/{org_slug}/repos/{repo['slug']}",
+                    params=(
+                        {"workspace": repo["workspace"]}
+                        if "workspace" in repo
```
jychp: Use `if repo.get("workspace")` instead of `"workspace" in repo`. Key presence sends an empty `workspace=` param for empty-string workspaces, which the new `repo-003` fixture has.

cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,28 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
...
+                        else None
+                    ),
+                    timeout=_TIMEOUT,
+                )
+                if detail_response.status_code == 404:
+                    logger.warning(
```
jychp: Add `org_slug` and `repo['slug']` as `%s` args. As written an operator cannot tell which repository was skipped.

cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,28 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
...
+                        "Skipping Socket.dev repository identity enrichment because "
+                        "the repository detail was not found",
+                    )
+                else:
+                    detail_response.raise_for_status()
```
jychp: Tolerate every 4xx here, not just 404. A single `403` on one repository the token cannot read raises out of `get()` and kills the entire repositories sync, including the repositories that fetched fine.

cartography/intel/socketdev/repositories.py
```diff
@@ -41,7 +53,28 @@ def get(api_token: str, org_slug: str) -> list[dict[str, Any]]:
...
+                    repo = {**repo, **detail_response.json()}
```
jychp: Merge only the key you came for: `repo = {**repo, "integration_meta": detail_response.json().get("integration_meta")}`. A detail response with a null or different `workspace` overwrites the list value, degrades `fullname` to the bare slug, and breaks the alert and dependency matches on `SocketDevRepository.fullname`.

cartography/intel/socketdev/repositories.py
```diff
@@ -63,17 +96,28 @@ def transform(raw_repos: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
 
-        # Build fullname from workspace/slug for ontology matching
+        # Keep Socket's repository identity for alert and dependency relationships.
         workspace = repo.get("workspace")
         slug = repo.get("slug")
         fullname = f"{workspace}/{slug}" if workspace and slug else slug
 
+        integration_meta = repo.get("integration_meta") or {}
+        integration_value = integration_meta.get("value") or {}
+        github_owner = integration_value.get("installation_login")
```
jychp, line 147: Does Socket.dev normalize `installation_login` and `repo_name` to lowercase? `_ont_url` comes from GitHub's html URL, which preserves the reported casing, so an exact match misses any repository with uppercase in its owner or name.

cartography/intel/socketdev/repositories.py
```diff
@@ -63,17 +96,28 @@ def transform(raw_repos: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
+        github_repo = integration_value.get("repo_name")
+        repository_url = (
+            f"https://github.com/{github_owner}/{github_repo}"
```
jychp, line 150: Hardcoding `https://github.com` breaks GitHub Enterprise Server, where `GitHubRepository.url` carries the GHES host, so `_ont_url` never matches and the `MONITORS` edge silently disappears for those users. Derive the host from `integration_meta` if Socket.dev exposes it, otherwise state the limitation in the `repository_url` description.

tests/data/socketdev/repositories.py
```diff
@@ -26,6 +26,29 @@
...
+            "integration_meta": {
+                "type": "github",
+                "value": {
+                    "installation_id": "installation-001",
+                    "installation_login": "example",
+                    "repo_id": "github-repo-003",
+                    "repo_name": "worker",
+                },
+            },
+            "workspace": "",
```
jychp: Give `repo-003` a real workspace and assert the `workspace/slug` fullname; `""` is not a shape Socket.dev returns. Add an explicit `"integration_meta": None` to `repo-001` and `repo-002` so the fixture matches what `get()` actually produces.

jychp (dismissed):

cartography/intel/socketdev/repositories.py
kunaals: socket’s schema allows this field to be missing or explicitly null. we enrich missing fields and treat null as no reported integration. the fixtures now cover both cases.

tests/data/socketdev/repositories.py
kunaals: updated `repo-003` with a real workspace and added explicit null metadata to `repo-001` and `repo-002`. `repo-004` still covers the workspace-omitted response shape.

cartography/intel/socketdev/repositories.py
kunaals: fixed this for all 4xx responses. cleanup is also skipped after a partial fetch so we don’t remove valid existing relationships.

cartography/intel/socketdev/repositories.py
kunaals: fixed this so we only copy `integration_meta`. the test confirms the detail response can’t overwrite the workspace.

cartography/intel/socketdev/repositories.py
kunaals: socket doesn’t expose a repository host and currently only documents the `github` integration type. i updated the property description to state that this is a github.com URL.

cartography/intel/socketdev/repositories.py
kunaals: fixed this so we only send `workspace` when it has a value.

cartography/intel/socketdev/repositories.py
kunaals: added the organization, repository slug, and status code to the warning.

cartography/intel/socketdev/repositories.py
kunaals: added a bounded pool of eight workers and log the enrichment count before starting.

cartography/intel/socketdev/repositories.py
kunaals: socket describes these as the github login and repository name, with no documented lowercase normalization. the transform preserves their casing, and the integration test now covers mixed-case values.

cartography/intel/socketdev/repositories.py
kunaals: moved the shared retry session into `socketdev/util.py` and made sure both sessions are closed properly.

cartography/intel/socketdev/repositories.py
```diff
@@ -1,55 +1,129 @@
...
+    api_session: requests.Session,
+    org_slug: str,
+    repo: dict[str, Any],
+) -> tuple[bool, dict[str, Any] | None]:
+    response = api_session.get(
+        f"{_BASE_URL}/orgs/{org_slug}/repos/{repo['slug']}",
+        params={"workspace": repo["workspace"]} if repo.get("workspace") else None,
+        timeout=_TIMEOUT,
+    )
+    if 400 <= response.status_code < 500:
```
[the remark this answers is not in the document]
kunaals: the maintainer explicitly requested tolerating every detail-endpoint 4xx, including 403. these failures are logged, the affected repository is not overwritten, and cleanup is skipped so existing graph state is preserved.

cartography/intel/socketdev/repositories.py
```diff
@@ -1,55 +1,129 @@
...
+        incomplete_repository_ids = set()
+        with _create_session(api_token) as api_session:
+            for repo in batch:
+                fetched, integration_meta = _fetch_integration_meta(
+                    api_session,
+                    org_slug,
+                    repo,
+                )
+                if fetched:
+                    repo["integration_meta"] = integration_meta
```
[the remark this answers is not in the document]
kunaals: socket documents these values as the github login and repository name and does not document different case normalization. the transform preserves the returned casing. changing this to a case-folded matcher would add a scan across `CodeRepository`; normalizing both provider models would be a broader ontology change without provider evidence.

cartography/intel/socketdev/repositories.py
kunaals: moved the shared retry session into `socketdev/util.py` and made sure both sessions are closed properly.

cartography/intel/socketdev/repositories.py
kunaals: fixed this for all 4xx responses. cleanup is also skipped after a partial fetch so we don’t remove valid existing relationships.

cartography/intel/socketdev/repositories.py
kunaals: fixed this so we only send `workspace` when it has a value.

cartography/intel/socketdev/repositories.py
kunaals: added a bounded pool of eight workers and log the enrichment count before starting.

cartography/intel/socketdev/repositories.py
kunaals: socket’s schema allows this field to be missing or explicitly null. we enrich missing fields and treat null as no reported integration. the fixtures now cover both cases.

cartography/intel/socketdev/repositories.py
kunaals: added the organization, repository slug, and status code to the warning.

cartography/intel/socketdev/repositories.py
kunaals: socket describes these as the github login and repository name, with no documented lowercase normalization. the transform preserves their casing, and the integration test now covers mixed-case values.

cartography/intel/socketdev/repositories.py
kunaals: socket doesn’t expose a repository host and currently only documents the `github` integration type. i updated the property description to state that this is a github.com URL.

cartography/intel/socketdev/repositories.py
kunaals: fixed this so we only copy `integration_meta`. the test confirms the detail response can’t overwrite the workspace.

cartography/intel/socketdev/repositories.py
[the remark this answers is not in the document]
kunaals: socket documents these values as the github login and repository name and does not document different case normalization. the transform preserves the returned casing. changing this to a case-folded matcher would add a scan across `CodeRepository`; normalizing both provider models would be a broader ontology change without provider evidence.

tests/data/socketdev/repositories.py
kunaals: updated `repo-003` with a real workspace and added explicit null metadata to `repo-001` and `repo-002`. `repo-004` still covers the workspace-omitted response shape.

cartography/intel/socketdev/repositories.py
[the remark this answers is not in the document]
kunaals: the maintainer explicitly requested tolerating every detail-endpoint 4xx, including 403. these failures are logged, the affected repository is not overwritten, and cleanup is skipped so existing graph state is preserved.

jychp (approved):

PR #3260, opened 2026-09-10
feat(microsoft): add delegated Entra authentication

kunaals:
### Type of change
- [x] Documentation update

### Summary
Adds an explicit experimental Microsoft delegated-authentication mode for environments where application credentials are unavailable.

The mode uses the current Azure CLI user, is pinned to the configured tenant, and attempts Entra datasets only. Microsoft Graph `403 Forbidden` responses are reported per dataset while later datasets continue. If any dataset is denied, the run exits nonzero after preserving the data it collected. Cleanup and derived federation analysis are disabled because delegated visibility can be incomplete even when Graph returns success. Application authentication remains the default and recommended path.

Intune and O365 are intentionally not attempted in delegated mode. Mixing delegated authentication with a client ID or client secret is rejected. Azure CLI tokens are cached until shortly before expiry to avoid a subprocess per Graph request.

Pagination errors now propagate in application-authenticated runs too. This prevents cleanup from treating truncated inventory as authoritative and deleting records that were not returned.

### Related issues or links

- None.

### How was this tested?

- `uv run pytest -q tests/unit/cartography/test_config.py tests/unit/cartography/intel/microsoft tests/unit/cartography/intel/microsoft/entra`: 65 passed.
- `uv run pytest -q tests/integration/cartography/intel/microsoft/entra/test_applications.py tests/integration/cartography/intel/microsoft/entra/test_users.py tests/integration/cartography/intel/microsoft/entra/test_groups.py`: 8 passed.
- `uv run pre-commit run --files ...` for all files changed by the review fixes.
- `uv run ./docs/build.sh` completed successfully.
- Ran an attended delegated-authentication sync against a real Microsoft tenant and a fresh disposable Neo4j database. The sync completed, Entra data loaded, and Intune and O365 were not attempted.

The real-provider smoke test used an account with a privileged directory role. A standard-user or deliberately restricted-user run is still required to validate partial visibility against a real tenant. Tests verify that `403 Forbidden` responses preserve collected records, produce an incomplete result, allow later datasets to continue, and prevent cleanup. They also verify that tenant ingestion remains a hard prerequisite and `401 Unauthorized` remains fatal.

#### General
- [x] I have added/updated tests that prove the feature works.

#### Proof of functionality

- [ ] Included sanitized Cartography sync logs demonstrating successful synchronization.

### Notes for reviewers

This is intentionally a degraded, one-time assessment mode rather than an unattended inventory mode. A successful request only proves what the signed-in user can see; it does not prove tenant completeness. The documentation therefore recommends application authentication and a fresh disposable database.


cartography/intel/microsoft/entra/app_role_assignments.py
```diff
@@ -27,25 +27,30 @@
...
+    :param tenant_id: Entra tenant that owns the application and service principal
     :param app_id: Application ID
     :return: Generator of app role assignment data as dicts
     """
     logger.debug("Fetching role assignments for application: %s", app_id)
 
     # Query the graph to get the service principal ID for this application
     query = """
-    MATCH (sp:EntraServicePrincipal {app_id: $app_id})
+    MATCH (tenant:AzureTenant {id: $tenant_id})-[:RESOURCE]->
```
achantavy: [nonblocking question] Is this necessary? Entra spn IDs should be indexed and are GUIDs which are global unique, which should save us the need to tie things to a tenant, and therefore save one extra parameter needing to be passed to the function.

docs/root/modules/microsoft/config.md
```diff
@@ -50,6 +53,63 @@ cartography \
...
+2. Run Cartography.
+
+   ```bash
+   cartography \
+     --selected-modules microsoft \
+     --microsoft-tenant-id '<TENANT_ID>' \
+     --microsoft-delegated-auth
+   ```
+
+Do not pass `--microsoft-client-id` or
```
achantavy: Since a key scenario for delegated auth is red teaming, we should plan to allow specifying arbitrary app IDs. Right now with the prior `az login` step we're restricting this to just Azure CLI which is a limitation. A red teamer may want to perform enumeration of the environment using a different app that they compromised other than the Azure CLI.

This comment is non-blocking only if there's a viable way to follow-up on this PR to allow for other client IDs.

achantavy (approved):
main comment is on allowing for arbitrary client ids on delegated auth

docs/root/modules/microsoft/config.md
kunaals: agreed that is a valuable scenario, we can do in follow up PR

cartography/intel/microsoft/entra/app_role_assignments.py
kunaals: microsoft documents that a multitenant application can have a separate service principal in each tenant, all referencing that same application. i tihnk this is good to have to reduce ambiguity - i feel like there is an edge case here im not thinking of that would benefit from this

kunaals: goes against YAGNI a bit but given that a multitenant app can have SP in each tenant it seems justified

PR #3267, opened 2026-09-11
feat(github): map SAML identities to canonical users

kunaals:
### Type of change

### Summary

GitHub accounts can remain disconnected from canonical users when their public profile email and organization verified-domain emails are empty. Ingest the organization's SAML external identities and use email-shaped NameIDs to link existing canonical users, without changing the meaning of GitHub's email fields.

External identities retain organization scope and provider node IDs. The existing GitHub ontology job adds unambiguous member matches unless normalized public or verified-domain email evidence identifies a different canonical user, so both sources share cleanup. GitHub SAML NameIDs, public/verified-domain email evidence, and canonical user ingestion share one Python normalization helper, so Unicode whitespace and case handling agree. The SAML conflict guard checks stored normalized values directly, including when case differences prevent an email-based account link from materializing. The analysis compares the stored lookup values using the canonical email index without changing raw SAML NameIDs, original emails, or user IDs. Fetching uses one paginated connection per organization, with no per-user requests. Each page checks the configured instance's GraphQL budget and retries transient failures up to five attempts, honoring GitHub's reset and Retry-After delays with additive jitter. The configured API URL is required for rate-limit checks. Only complete snapshots are written. Unavailable, denied, or exhausted API snapshots warn and preserve prior data while the remaining GitHub resources and organizations continue. Graph load and cleanup errors still propagate. Successful empty snapshots remove stale identities within that organization.

### Related issues or links

- [GitHub OrganizationIdentityProvider API](https://docs.github.com/en/graphql/reference/objects#organizationidentityprovider)
- [GitHub ExternalIdentity API](https://docs.github.com/en/graphql/reference/objects#externalidentity)
- [GitHub GraphQL rate-limit recovery](https://docs.github.com/en/graphql/overview/rate-limits-and-query-limits-for-the-graphql-api#exceeding-the-rate-limit)

### How was this tested?

- Current revision: 5,418 unit tests and 378 focused GitHub/ontology tests (including 141 integration tests) pass; `make test_lint` and `uv run ./docs/build.sh` pass.
- Synthetic provider responses exercise 47-minute primary resets, Retry-After waits beyond five minutes, secondary-limit backoff, same-cursor recovery, and permanent failures. A two-organization orchestration test proves continuation after identity collection failures. Real Neo4j tests prove that exhausted later-page HTTP/GraphQL failures preserve identities and canonical-user links across update tags; a separate test confirms graph-write failures still propagate.
- Regression tests cover mixed-case and Unicode email conflicts without a pre-existing account link, personal and alias emails without a competing canonical owner, cleanup of a previous SAML link, and backfill after GitHub resync. Shared jitter never shortens the provider-requested wait.
- Repeated limited live validation on `7538e2d4b` against fresh GitHub members and SAML identities plus current identity-provider prerequisites. Every independently calculated, unambiguous source-supported pair appeared, with no missing or extra pairs; a second provider sync preserved the exact set. Live alias evidence confirmed that differing emails should block SAML only when they identify another canonical user. Neo4j `EXPLAIN` confirms index seeks on `User.normalized_email` for both the SAML match and conflict check. Rate-limit exhaustion remains covered synthetically; it was not induced live.

- Limited live GitHub App validation fetched users and SAML identities into a separate temporary local database, using minimal identity-provider prerequisites. Every unambiguous source-supported pair appeared in the graph. Repeating the provider sync and ontology run preserved the exact mapping set. Provider access was read-only.

Live log excerpt below; organization names and population counts are redacted to keep customer information private. No customer records or credentials are included in the fixtures, code, or public artifacts.

```text
INFO:cartography.intel.github.external_identities:Syncing GitHub external identities for <organization>
INFO:cartography.client.core.tx:Loaded <identity-count> GitHubExternalIdentity nodes
INFO:cartography.graph.statement:Completed GitHubExternalIdentity statement #1
INFO:cartography.graph.statement:Completed GitHubExternalIdentity statement #2
INFO:cartography.graph.statement:Completed GitHubExternalIdentity statement #3
INFO:cartography.intel.github.external_identities:Synced <identity-count> GitHub external identities for <organization>
INFO:cartography.graph.statement:Completed ontology_users_github statement #1
INFO:cartography.graph.statement:Completed ontology_users_github statement #2
INFO:cartography.graph.statement:Completed ontology_users_github statement #3
```

#### General

#### Proof of functionality

- [x] Included sanitized Cartography sync logs demonstrating successful synchronization.

### Notes for reviewers

A null SAML provider is treated as unavailable, since visibility depends on credentials. Only a successful identity listing is authoritative for deletion. Non-email NameIDs remain available as source data but do not create canonical user links. Existing GitHub users must be resynced to populate the normalized email evidence before this SAML guard permits linking. Personal or alias emails without a competing canonical owner do not block an otherwise unambiguous SAML link. A conflicting canonical owner blocks the link even when case differences prevented an exact-email relationship from being created. Live validation used an installation token; classic PAT and GitHub Enterprise authentication were not exercised live.

### DIFF
--- a/cartography/analysis/ontology/analysis.py
+++ b/cartography/analysis/ontology/analysis.py
@@ -586,7 +586,7 @@ USER_HAS_GITHUB_ACCOUNT = AnalysisJob(
             match=(
                 "MATCH (g:GitHubUser)-[:MEMBER_OF]->(org:GitHubOrganization)"
                 "-[:RESOURCE]->(identity:GitHubExternalIdentity)<-[:HAS_IDENTITY]-(g) "
-                "MATCH (u:User {normalized_email: toLower(trim(identity.saml_name_id))}) "
+                "MATCH (u:User {normalized_email: identity.saml_name_id_normalized}) "
                 "WHERE u.normalized_email CONTAINS '@' "
                 "WITH g, collect(DISTINCT u) AS candidates "
                 "WHERE size(candidates) = 1 "
--- a/cartography/analysis/ontology/analysis.py
+++ b/cartography/analysis/ontology/analysis.py
@@ -582,6 +582,30 @@ USER_HAS_GITHUB_ACCOUNT = AnalysisJob(
                 ),
             ),
         ),
+        AnalysisStatement(
+            match=(
+                "MATCH (g:GitHubUser)-[:MEMBER_OF]->(org:GitHubOrganization)"
+                "-[:RESOURCE]->(identity:GitHubExternalIdentity)<-[:HAS_IDENTITY]-(g) "
+                "MATCH (u:User) "
+                "WHERE u.email CONTAINS '@' "
+                "AND toLower(trim(u.email)) = toLower(trim(identity.saml_name_id)) "
+                "WITH g, collect(DISTINCT u) AS candidates "
+                "WHERE size(candidates) = 1 "
+                "WITH g, candidates[0] AS u "
+                "WHERE NOT EXISTS { "
+                "MATCH (other:User)-[existing:HAS_ACCOUNT]->(g) "
+                "WHERE other <> u AND existing.lastupdated = $UPDATE_TAG }"
+            ),
+            effects=(
+                AddRelationship(
+                    "u",
+                    "HAS_ACCOUNT",
+                    "g",
+                    source_label="User",
+                    target_label="GitHubUser",
+                ),
+            ),
+        ),
     ),
 )
 USER_OWNS_API_KEY = AnalysisJob(
--- a/cartography/analysis/ontology/analysis.py
+++ b/cartography/analysis/ontology/analysis.py
@@ -586,15 +586,15 @@ USER_HAS_GITHUB_ACCOUNT = AnalysisJob(
             match=(
                 "MATCH (g:GitHubUser)-[:MEMBER_OF]->(org:GitHubOrganization)"
                 "-[:RESOURCE]->(identity:GitHubExternalIdentity)<-[:HAS_IDENTITY]-(g) "
-                "MATCH (u:User) "
-                "WHERE u.email CONTAINS '@' "
-                "AND toLower(trim(u.email)) = toLower(trim(identity.saml_name_id)) "
+                "MATCH (u:User {normalized_email: toLower(trim(identity.saml_name_id))}) "
+                "WHERE u.normalized_email CONTAINS '@' "
                 "WITH g, collect(DISTINCT u) AS candidates "
                 "WHERE size(candidates) = 1 "
                 "WITH g, candidates[0] AS u "
                 "WHERE NOT EXISTS { "
-                "MATCH (other:User)-[existing:HAS_ACCOUNT]->(g) "
-                "WHERE other <> u AND existing.lastupdated = $UPDATE_TAG }"
+                "MATCH (other:User)-[:HAS_ACCOUNT]->(g) "
+                "WHERE other <> u AND (other.email = g.email "
+                "OR other.email IN g.organization_verified_domain_emails) }"
             ),
             effects=(
                 AddRelationship(
--- a/cartography/helpers.py
+++ b/cartography/helpers.py
@@ -10,6 +10,11 @@ logger = logging.getLogger(__name__)
 DEFAULT_BATCH_SIZE = 1000
 
 
+def normalize_email_for_matching(value: str | None) -> str | None:
+    """Use the same Unicode whitespace and case policy on both sides of identity joins."""
+    return (value or "").strip().lower() or None
+
+
 def backoff_handler(details: Dict) -> None:
     """
     Log backoff retry attempts for monitoring and debugging.
--- a/cartography/intel/github/__init__.py
+++ b/cartography/intel/github/__init__.py
@@ -13,6 +13,7 @@ import cartography.intel.github.container_image_attestations
 import cartography.intel.github.container_image_tags
 import cartography.intel.github.container_images
 import cartography.intel.github.dependabot_alerts
+import cartography.intel.github.external_identities
 import cartography.intel.github.packages
 import cartography.intel.github.personal_access_tokens
 import cartography.intel.github.repos
@@ -115,6 +116,13 @@ def start_github_ingestion(
         token: Any = credential
 
         github_users = cartography.intel.github.users.sync(
+            neo4j_session,
+            common_job_parameters,
+            token,
+            api_url,
+            org_name,
+        )
+        cartography.intel.github.external_identities.sync(
             neo4j_session,
             common_job_parameters,
             token,
--- a/cartography/intel/github/external_identities.py
+++ b/cartography/intel/github/external_identities.py
@@ -1,0 +1,121 @@
+import logging
+from typing import Any
+
+import neo4j
+
+from cartography.client.core.tx import load
+from cartography.graph.job import GraphJob
+from cartography.intel.github.util import fetch_page
+from cartography.models.github.external_identities import GitHubExternalIdentitySchema
+from cartography.util import timeit
+
+logger = logging.getLogger(__name__)
+
+EXTERNAL_IDENTITIES_QUERY = """
+query($login: String!, $cursor: String) {
+    organization(login: $login) {
+        url
+        samlIdentityProvider {
+            externalIdentities(first: 100, after: $cursor) {
+                nodes {
+                    id
+                    samlIdentity { nameId }
+                    user { url }
+                }
+                pageInfo { endCursor hasNextPage }
+            }
+        }
+    }
+}
+"""
+
+
+def get_external_identities(
+    token: Any, api_url: str, organization: str
+) -> tuple[list[dict[str, Any]], str] | None:
+    # Complete the snapshot before writing: a denied later page must not remove
+    # identities or their account links from a previous successful sync.
+    identities: list[dict[str, Any]] = []
+    cursor = None
+    while True:
+        response = fetch_page(
+            token, api_url, organization, EXTERNAL_IDENTITIES_QUERY, cursor
+        )
+        errors = response.get("errors")
+        if errors:
+            if all(
+                error.get("type") in {"FORBIDDEN", "INSUFFICIENT_SCOPES"}
+                for error in errors
+            ):
+                logger.warning(
+                    "Skipping GitHub external identities for %s: access denied; "
+                    "preserving previously synced identities.",
+                    organization,
+                )
+                return None
+            raise RuntimeError(
+                f"GitHub external identity query failed for {organization}; "
+                "preserving previously synced identities."
+            )
+        org = response["data"]["organization"]
+        if org is None:
+            raise RuntimeError(f"GitHub organization {organization} was not returned")
+        provider = org["samlIdentityProvider"]
+        if provider is None:
+            # This nullable field is visible only to authorized credentials;
+            # absence alone cannot establish that SAML was disabled.
+            logger.info(
+                "GitHub SAML provider unavailable for %s; preserving previously synced identities.",
+                organization,
+            )
+            return None
+        connection = provider["externalIdentities"]
+        identities.extend(connection["nodes"])
+        page_info = connection["pageInfo"]
+        if not page_info["hasNextPage"]:
+            return identities, org["url"]
+        next_cursor = page_info["endCursor"]
+        if not next_cursor or next_cursor == cursor:
+            raise RuntimeError("GitHub external identity pagination did not advance")
+        cursor = next_cursor
+
+
+def transform_external_identities(
+    identities: list[dict[str, Any]], org_url: str
+) -> list[dict[str, Any]]:
+    return [
+        {
+            "id": f"{org_url}|{identity['id']}",
+            "saml_name_id": (identity.get("samlIdentity") or {}).get("nameId"),
+            "user_url": (identity.get("user") or {}).get("url"),
+        }
+        for identity in identities
+    ]
+
+
+@timeit
+def sync(
+    neo4j_session: neo4j.Session,
+    common_job_parameters: dict[str, Any],
+    token: Any,
+    api_url: str,
+    organization: str,
+) -> None:
+    logger.info("Syncing GitHub external identities for %s", organization)
+    snapshot = get_external_identities(token, api_url, organization)
+    if snapshot is None:
+        return
+    identities, org_url = snapshot
+    data = transform_external_identities(identities, org_url)
+    load(
+        neo4j_session,
+        GitHubExternalIdentitySchema(),
+        data,
+        lastupdated=common_job_parameters["UPDATE_TAG"],
+        org_url=org_url,
+    )
+    GraphJob.from_node_schema(
+        GitHubExternalIdentitySchema(),
+        {**common_job_parameters, "org_url": org_url},
+    ).run(neo4j_session)
+    logger.info("Synced %d GitHub external identities for %s", len(data), organization)
--- a/cartography/intel/github/external_identities.py
+++ b/cartography/intel/github/external_identities.py
@@ -7,6 +7,7 @@ import requests
 
 from cartography.client.core.tx import load
 from cartography.graph.job import GraphJob
+from cartography.helpers import normalize_email_for_matching
 from cartography.intel.github.util import _extract_error_message
 from cartography.intel.github.util import _get_retry_sleep_seconds_for_http_error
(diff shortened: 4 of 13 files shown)


cartography/analysis/ontology/analysis.py
```diff
@@ -582,6 +582,30 @@
                 ),
             ),
         ),
+        AnalysisStatement(
+            match=(
+                "MATCH (g:GitHubUser)-[:MEMBER_OF]->(org:GitHubOrganization)"
+                "-[:RESOURCE]->(identity:GitHubExternalIdentity)<-[:HAS_IDENTITY]-(g) "
+                "MATCH (u:User) "
```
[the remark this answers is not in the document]
kunaals: added an indexed `User.normalized_email` while preserving original emails and IDs. the original query used a hash join in my PROFILE run, but the revised conflict guard exposed a Cartesian plan without the index. the indexed query takes 17k / 170k database hits for 1k / 10k synthetic identities, with identical matches.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,121 @@
...
+
+def get_external_identities(
+    token: Any, api_url: str, organization: str
+) -> tuple[list[dict[str, Any]], str] | None:
+    # Complete the snapshot before writing: a denied later page must not remove
+    # identities or their account links from a previous successful sync.
+    identities: list[dict[str, Any]] = []
+    cursor = None
+    while True:
+        response = fetch_page(
```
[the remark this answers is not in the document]
kunaals: added bounded retries on the same page for transient HTTP failures, GraphQL timeouts, and rate limits. HTTP retries honor retry/reset headers and stop if the required wait exceeds five minutes. tests cover recovery, exhaustion, and preserving the previous snapshot after a later-page failure.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,189 @@
...
+                or attempt == 4
+                or not all(
+                    error.get("type") == "RATE_LIMITED"
+                    or error.get("message") == "timedout"
+                    for error in errors
+                )
+            ):
+                return response
+            if any(error.get("type") == "RATE_LIMITED" for error in errors):
+                delay = 60
```
jychp, line 89: A GraphQL `RATE_LIMITED` error means the hourly GraphQL budget is exhausted, so four flat 60s waits cannot clear it and this ends in the `RuntimeError` in `get_external_identities`. Call `handle_rate_limit_sleep(token)` before each page the way `fetch_all()` does, and derive this wait from the rate-limit reset instead of hardcoding 60.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,189 @@
...
+            )
+            if (
+                http_response.status_code not in _TRANSIENT_STATUS_CODES
+                and not rate_limited
+            ):
+                raise
+            delay = _get_retry_sleep_seconds_for_http_error(error, attempt + 1)
+            # Do not retry earlier than GitHub permits or let a server-provided
+            # reset time hold this optional stage indefinitely.
+            if attempt == 4 or delay > 300:
```
jychp, line 67: A primary rate-limit 403 carries an `x-ratelimit-reset` up to an hour out, so `delay > 300` raises on the ordinary recovery path, not an abusive one. Sleep until reset, or call `handle_rate_limit_sleep()` before each page so this branch is only reached for secondary limits.

cartography/intel/github/__init__.py
```diff
@@ -121,6 +122,13 @@ def start_github_ingestion(
             api_url,
             org_name,
         )
+        cartography.intel.github.external_identities.sync(
```
jychp, line 125: Any raise from this sync propagates out of `start_github_ingestion`, so repos, PATs, dependabot alerts, teams, codeowners and actions never run for this org or any later one. Wrap the call in a try/except that logs and continues: an unavailable identity snapshot is already designed to be non-fatal, and a rate limit or GraphQL error must not be fatal either.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,189 @@
...
+            )
+        except requests.HTTPError as error:
+            http_response = error.response
+            if http_response is None:
+                raise
+            rate_limited = http_response.status_code == 403 and (
+                http_response.headers.get("x-ratelimit-remaining") == "0"
+                or "retry-after" in http_response.headers
+                or "secondary rate limit"
+                in _extract_error_message(http_response).lower()
```
jychp, line 56: Hoist `_extract_error_message(http_response)` into a local; it runs twice here and re-parses the response body each time. This block also duplicates the 403 classification already inside `_get_retry_sleep_seconds_for_http_error`; move the retryability decision into `util.py` and call it from both places.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+}
+"""
+
+
+def _fetch_identity_page(
+    token: Any, api_url: str, organization: str, cursor: str | None
+) -> dict[str, Any]:
+    for attempt in range(5):
```
jychp, line 42: Extract the attempt budget into a module constant. `5` is spelled out in `range(5)`, in two `attempt == 4` checks and in the log string, and `300` on line 67 is an undocumented magic number; changing one silently breaks the others.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,189 @@
...
+        cursor = next_cursor
+
+
+def transform_external_identities(
+    identities: list[dict[str, Any]], org_url: str
+) -> list[dict[str, Any]]:
+    return [
+        {
+            "id": f"{org_url}|{identity['id']}",
+            "saml_name_id": (identity.get("samlIdentity") or {}).get("nameId"),
```
jychp, line 154: nit: Bind `(identity.get("samlIdentity") or {}).get("nameId")` to a local and reuse it for both keys instead of evaluating it twice.

cartography/analysis/ontology/analysis.py
```diff
@@ -582,6 +582,30 @@
                 ),
             ),
         ),
+        AnalysisStatement(
+            match=(
+                "MATCH (g:GitHubUser)-[:MEMBER_OF]->(org:GitHubOrganization)"
+                "-[:RESOURCE]->(identity:GitHubExternalIdentity)<-[:HAS_IDENTITY]-(g) "
+                "MATCH (u:User {normalized_email: identity.saml_name_id_normalized}) "
```
jychp, line 589: The CrowdStrike and Jamf statements in this same file still match canonical users with `toLower(trim(u.email))`, which strips a different whitespace set than `normalize_email_for_matching` and leaves the new index unused. Point them at `u.normalized_email`, or leave a TODO: two normalization policies for the same field will drift.

cartography/helpers.py
```diff
@@ -10,6 +10,11 @@
 DEFAULT_BATCH_SIZE = 1000
 
 
+def normalize_email_for_matching(value: str | None) -> str | None:
```
jychp: **suggestion** — Move this to `cartography/intel/ontology/utils.py`; `helpers.py` holds batching and backoff utilities, and this is identity-matching policy. Rename it too: it also normalizes opaque SAML NameIDs, not just emails.

jychp (changes requested):

cartography/intel/github/external_identities.py
kunaals: added the rate-limit check before every page attempt, using the configured GitHub instance. primary limits now wait for the reported reset; secondary limits use exponential backoff starting at one minute. added a regression test with a reset 47 minutes away.

cartography/intel/github/external_identities.py
kunaals: removed the five-minute cap. we now honor reset and Retry-After delays beyond five minutes while keeping the five-attempt limit. covered both cases in tests.

cartography/intel/github/__init__.py
kunaals: caught provider collection failures inside the identity sync, before any writes. failures now warn and preserve the previous snapshot while later resources and organizations continue. load and cleanup errors still propagate. added coverage for two-organization continuation and preservation of existing identities and account links.

cartography/intel/github/external_identities.py
kunaals: moved retryability into the existing util helper: it returns a delay or None for permanent failures. both the identity collector and fetch_all use it now, removing the duplicate 403 classification and repeated parsing from this block.

cartography/intel/github/external_identities.py
kunaals: added _MAX_PAGE_ATTEMPTS and used it for the loop, exhaustion checks, and logging. the 300-second cap is removed as discussed above.

cartography/intel/github/external_identities.py
kunaals: bound name_id once and reused it for the raw and normalized fields.

cartography/analysis/ontology/analysis.py
kunaals: added a TODO beside the device ownership job. moving these joins safely also requires normalizing the CrowdStrike/Jamf source emails with the same helper; changing only the User side would leave inconsistent whitespace handling. the TODO calls out source normalization and Unicode/cleanup coverage.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,196 @@
...
+                    "Skipping GitHub external identities for %s: access denied; "
+                    "preserving previously synced identities.",
+                    organization,
+                )
+                return None
+            raise ExternalIdentitySnapshotError(
+                f"GitHub external identity query failed for {organization}; "
+                "preserving previously synced identities."
+            )
+        org = response["data"]["organization"]
```
jychp: **blocking** - Guard the payload shape here instead of indexing it blind. `response["data"]`, `provider["externalIdentities"]`, `connection["nodes"]` and `page_info["hasNextPage"]` (plus `response_json["resources"]["graphql"]` inside `handle_rate_limit_sleep`) raise `KeyError`/`TypeError` on a malformed payload, and that escapes the `except (requests.RequestException, ExternalIdentitySnapshotError)` in `sync()`, aborting repos, PATs, dependabot, teams, codeowners and actions for this org and every later one. Raise `ExternalIdentitySnapshotError` on an unexpected shape so the non-fatal contract actually holds.

cartography/analysis/ontology/analysis.py
```diff
@@ -582,6 +585,30 @@
...
+                "WHERE u.normalized_email CONTAINS '@' "
+                "WITH g, collect(DISTINCT u) AS candidates "
+                "WHERE size(candidates) = 1 "
+                "WITH g, candidates[0] AS u "
+                "WHERE NOT EXISTS { "
+                "MATCH (other:User)-[:HAS_ACCOUNT]->(g) "
+                "WHERE other <> u AND (other.email = g.email "
```
jychp: This guard fails open on case. The `other.email = g.email` edge it looks for is created by `UserToUserAccountRel`, whose matcher is an exact `email` equality, so a canonical user with `Bob@Example.com` and a `GitHubUser.email` of `bob@example.com` produce no link, the guard passes, and the SAML NameID links a different `User` to that account. Compare the values directly (`g.email` and `g.organization_verified_domain_emails` normalized against `u.normalized_email`) so the conflict check does not depend on an exact-case edge having materialized first.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,196 @@
...
+                # for secondary limits, which /rate_limit cannot report.
+                delay = 60 * 2**attempt
+        logger.warning(
+            "Retrying GitHub external identity page for %s in %s seconds (attempt %d/%d).",
+            organization,
+            delay,
+            attempt + 2,
+            _MAX_PAGE_ATTEMPTS,
+        )
+        time.sleep(delay)
```
jychp: Add jitter. `repos.py` already uses `_sleep_with_jitter()` for exactly this reason; a bare `time.sleep()` makes every org and every concurrent run retry in lockstep against the same secondary limit. Move `_sleep_with_jitter` from `repos.py:287` into `util.py` and call it from all three retry loops.

cartography/intel/github/util.py
```diff
@@ -107,21 +109,28 @@ def _get_retry_sleep_seconds_for_http_error(
...
+    ):
+        # Without a reset or Retry-After, GitHub requires at least a minute,
+        # increasing exponentially if the secondary limit persists.
+        return int(60 * 2 ** (retry - 1))
+
+    return None
+
+
+def handle_rate_limit_sleep(
+    token: Any, api_url: str = "https://api.github.com/graphql"
```
jychp: Drop the default and make `api_url` required. All three call sites now pass it, and a `https://api.github.com/graphql` default means a future caller that forgets sends a GitHub Enterprise Server token to github.com and reads the wrong budget.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,196 @@
...
+"""
+
+
+def _fetch_identity_page(
+    token: Any, api_url: str, organization: str, cursor: str | None
+) -> dict[str, Any]:
+    for attempt in range(_MAX_PAGE_ATTEMPTS):
+        delay = 2 ** (attempt + 1)
+        try:
+            handle_rate_limit_sleep(token, api_url)
```
jychp: **question** - What happens on a GitHub Enterprise Server instance with rate limiting disabled? `/rate_limit` returns 404 there, `_get_retry_sleep_seconds_for_http_error` classifies 404 as permanent, so the first attempt raises and `sync()` swallows it: external identities never sync on those instances and only a warning says so. Treat a 404 from the rate-limit probe as "no limit configured" and continue.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,196 @@
...
+    common_job_parameters: dict[str, Any],
+    token: Any,
+    api_url: str,
+    organization: str,
+) -> None:
+    logger.info("Syncing GitHub external identities for %s", organization)
+    try:
+        snapshot = get_external_identities(token, api_url, organization)
+    except (requests.RequestException, ExternalIdentitySnapshotError) as error:
+        logger.warning(
```
jychp: **suggestion** - Log the error, not just `type(error).__name__`. This warning is the only signal on a path that silently keeps stale identity data, and `HTTPError` alone does not tell an operator whether it was a 403, a 502 or a reset connection. Pass `error` as a `%s` arg or add `exc_info=True`.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,196 @@
...
+import time
+from typing import Any
+
+import neo4j
+import requests
+
+from cartography.client.core.tx import load
+from cartography.graph.job import GraphJob
+from cartography.helpers import normalize_email_for_matching
+from cartography.intel.github.util import _get_retry_sleep_seconds_for_http_error
```
jychp: Rename `_get_retry_sleep_seconds_for_http_error` to a public name in `util.py`. It is now a shared contract between two modules, and a leading underscore imported across module boundaries signals the opposite.

tests/unit/cartography/intel/github/test_github.py
```diff
@@ -24,6 +28,11 @@
 from tests.data.github.rate_limit import RATE_LIMIT_RESPONSE_JSON
 
 
+@pytest.mark.parametrize(
```
jychp: Split the identity-failure cases into their own test. The `None` case never assigns `mock_external_identities_sync.side_effect`, so it re-runs the original cleanup-ordering assertions unchanged while patching a function nothing calls, and the two failure cases attach an unrelated concern to a test about deferred global cleanup. Then annotate `identity_failure: Exception | None` and drop the `@typing.no_type_check`.

cartography/intel/github/external_identities.py
```diff
@@ -0,0 +1,196 @@
...
+                delay = 60 * 2**attempt
+        logger.warning(
+            "Retrying GitHub external identity page for %s in %s seconds (attempt %d/%d).",
+            organization,
+            delay,
+            attempt + 2,
+            _MAX_PAGE_ATTEMPTS,
+        )
+        time.sleep(delay)
+    raise AssertionError("Unreachable: final attempt returns or raises")
```
jychp: Raise `RuntimeError`, not `AssertionError`. `AssertionError` reads as a stripped-under-`-O` assertion to anyone skimming, and this is a real control-flow invariant.

jychp (approved):
Some none blocking comments

PR #3269, opened 2026-09-12
fix(container-images): batch layer-cache reads

kunaals:
### Type of change

### Summary
Large registry syncs can exhaust Neo4j's heap while checking cached image layers, even when the cache is empty. Batch explicit digest lookups in groups of 500, preserving completeness checks, scope filters, and the existing whole-scope lookup used by GitHub and GitLab.

### Related issues or links
- Follow-up to #3092, prompted by a production GCP sync failure.

### How was this tested?
- 136 unit tests and 48 Neo4j integration tests across the shared helper and all five registry callers; `make test_lint`.
- Synthetic Neo4j 2025.10 Enterprise, 512 MB heap, 100,000 images: the original helper OOMs on 500,000 candidate digests; this patch returns the expected empty result in 7.9s / 4.0s. The new batch-boundary regression also fails against the original helper.

#### General

#### Proof of functionality

### DIFF
--- a/cartography/intel/container_image_layers.py
+++ b/cartography/intel/container_image_layers.py
@@ -79,9 +79,13 @@ def get_complete_layer_digests(
     shape: ContainerImageLayerGraphShape,
     digests: Iterable[str] | None,
     scope_values: Mapping[str, Any],
+    *,
+    batch_size: int = 500,
 ) -> set[str]:
     """Return digests whose layer closure is complete for the current scope."""
 
+    if batch_size <= 0:
+        raise ValueError("batch_size must be greater than zero")
     unique_digests = sorted(set(digests)) if digests is not None else None
     if unique_digests == []:
         return set()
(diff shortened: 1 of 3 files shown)


cartography/intel/container_image_layers.py
```diff
@@ -185,13 +189,22 @@ def get_complete_layer_digests(
...
-        read_list_of_values_tx,
-        query,
-        **({"digests": unique_digests} if unique_digests is not None else {}),
-        **parameters,
-    )
-    return {str(value) for value in values if value}
+    # Large IN lists can exhaust Neo4j's heap even when no layers are cached.
+    digest_batches: Iterable[list[str] | None]
+    if unique_digests is None:
+        digest_batches = [None]
```
achantavy, line 195: Style/readability nit, not blocking: it feels a little weird to put None in a list just to force one iteration.

I’d prefer keeping the cases explicit:

```
if unique_digests is None:
    values = neo4j_session.execute_read(
        read_list_of_values_tx,
        query,
        **parameters,
    )
    return {str(value) for value in values if value}

if not unique_digests:
    return set()

complete_digests: set[str] = set()
for digest_batch in batch(unique_digests, size=batch_size):
    values = neo4j_session.execute_read(
        read_list_of_values_tx,
        query,
        digests=digest_batch,
        **parameters,
    )
    complete_digests.update(str(value) for value in values if value)

return complete_digests
```

achantavy (approved):

PR #3271, opened 2026-09-14
fix(snowflake): support inherited grants and correct collection setup

kunaals:
### Type of change

- [x] Documentation update

### Summary

Fix Snowflake collection for case-sensitive object names and asynchronous SQL responses. Quote REST identifiers before URL encoding and support `statementStatusUrl` responses. Validate polling, pagination, and every HTTP redirect against the configured account origin before sending another request or accepting its data.

Inherited grant rows can omit an individual object name, so the existing object-grant transform did not represent them in the graph. Preserve these grants as `SnowflakeInheritedGrant` records linked to their principal and account/database/schema container. Store the object type separately: inherited SELECT on TABLEs inside a database is not ordinary SELECT on the database or VIEWs. Complete ACCOUNT_USAGE reads can now clean up revoked direct and inherited grants. Incomplete reads still preserve previous graph state; direct, inherited, and role-assignment cleanup have independent completeness flags. Malformed inherited rows warn and are skipped without aborting valid collection or retaining revoked direct privileges.

Keep an explicit grant projection without requiring SHOW COLUMNS metadata access. Query the inherited-grant columns first. Handle only structured missing-column errors: omit the optional fields when `IS_INHERITED` is absent, or derive scope from `INHERITED_FROM_ACCOUNT`/database/schema when an account exposes source fields instead of the documented `INHERITED_FROM` enum. Unexpected failures propagate rather than silently discarding inherited-grant information or triggering a per-role API walk.

Replace unsupported account-scoped ordinary setup grants with per-database grants for existing and future schemas, tables, event tables, views, materialized views, external tables, Iceberg tables, and dynamic tables. Dynamic tables use MONITOR; the other listed table/view types use REFERENCES. Document schema-level future-grant precedence, inherited-grant semantics, and Native App role limitations.

### Related issues or links

- [SQL API asynchronous responses](https://docs.snowflake.com/en/developer-guide/sql-api/handling-responses)
- [REST identifier contract](https://github.com/snowflakedb/snowflake-rest-api-specs/blob/main/specifications/common.yaml)
- [GRANT syntax and future-grant precedence](https://docs.snowflake.com/en/sql-reference/sql/grant-privilege)
- [Dynamic table metadata privileges](https://docs.snowflake.com/en/user-guide/dynamic-tables/privileges)
- [Inherited grants and container fields](https://docs.snowflake.com/en/user-guide/inherited-grants-using)
- [Inherited grants general availability and account opt-in](https://docs.snowflake.com/en/release-notes/2026/other/2026-09-10-inherited-grants-ga)
- [Native App role grants](https://docs.snowflake.com/en/sql-reference/sql/grant-application-role)

### How was this tested?

- Full unit suite plus all Snowflake integration tests: 5,631 passed, using disposable Neo4j. Shared model/schema checks are included in the unit suite.
- Multi-tag graph tests cover direct privilege revocation while inherited grants remain, inherited grant revocation, distinct TABLE/VIEW grants, account/database/schema scope, account isolation in both directions, independent role-assignment cleanup, and malformed inherited rows preserving only inherited state while valid rows load and direct grants revoke.
- Local HTTP tests cover off-origin redirects from polling and pagination, same-origin redirects, redirect-loop bounds, and 307/308 SQL partition redirects with path, query-only, and fragment-only Location references. The fragment cases failed before resolving against the actual response URL; all six redirect variants now preserve the correct partition without duplicating parameters. Rejected origins receive no request.
- Projection tests cover accounts with and without inherited columns and realistic structured SQL errors. Only missing `IS_INHERITED` retries without inheritance fields; the source-account layout preserves them and any other missing scope field still fails.
- `make test_lint`: passed.
- `uv run --frozen --group doc ./docs/build.sh`: passed, including generated schema documentation; 56 existing warnings elsewhere, none for Snowflake.
- Official Snowflake documentation checked. Setup grants were not applied to a fresh account, and provider privileges were not changed to exercise revocation; those transitions use synthetic provider responses with real Neo4j.

At `96e2111dc`, read-only live grant collection succeeded in an isolated local Neo4j database using account, principal, database/schema and available object prerequisites collected by the same commit. The grant collector, account-level grant collector, and their scoped cleanup completed without warnings; all object, inherited, role-assignment and account-grant completeness flags were true. An independent provider query matched all three inherited grants to graph properties and verified each exact principal and container link. No provider privileges, production configuration, or production graph were modified.

The full configured entry-point attempt at this commit reached function listing, then stopped on a TLS `WRONG_VERSION_NUMBER` handshake error before grants or top-level cleanup. Retrying the exact failed GET returned HTTP 200. Grant validation then continued explicitly against the prerequisites already collected in that isolated database; this is targeted live validation, not a successful full-module rerun. The earlier full-module run succeeded at `a33e7df33`.

Reviewed, tenant-neutral excerpt from the current grant sync:

```text
INFO:cartography.client.core.tx:Loaded 3 SnowflakeInheritedGrant nodes
```

Independent validation results:

```json
{"source_grants": 3, "inherited_grants": 3, "source_graph_parity": true, "principal_container_links_exact": true}
```

Malformed-row, redirect and revocation regressions use synthetic provider inputs with real Neo4j/local HTTP servers. The documented inherited-scope enum layout is covered by unit tests. Live revocation and fresh-account setup were not attempted.

#### General
- [x] Read the contributing guidelines.
- [x] Linter passes locally (`make test_lint`).
- [x] Added regression unit/integration tests.

#### Proof of functionality

#### Connector validation
- [x] Tested the changed connector against a real provider environment.
- [x] Included reviewed, tenant-neutral sync evidence.
- [x] Public fixtures and evidence contain no credentials, personal data, account identifiers, or customer resource data.

#### Graph schema
- [x] Updated model docstrings and PropertyRef descriptions.
- [x] Built generated schema documentation.

### Notes for reviewers

Inherited grants describe privileges on matching current and future objects inside their container. They are not expanded into effective-access edges: required container USAGE and policy restrictions still apply. Unsupported principal kinds retain their provider identity on the grant record without an invented principal link. Native App role modeling remains outside this change. ACCOUNT_USAGE latency still applies to revocation visibility.


cartography/intel/snowflake/grants.py
```diff
@@ -496,7 +497,17 @@ def sync(
         grants_by_role, grants_of_by_role = account_usage.split_grants(
             grants_to_roles, grants_to_users
         )
-        complete = True
+        inherited_count = sum(
+            to_bool(row.get("is_inherited")) is True for row in grants_to_roles
+        )
+        complete = inherited_count == 0
```
jychp: Scope this flag to grant edges only. `complete` also gates the role-hierarchy and user-role-assignment MatchLink cleanups in `cleanup()` (grants.py:387-393), which inherited grants cannot affect: they never produce `granted_on IN (ROLE, DATABASE_ROLE)` rows and never appear in `GRANTS_TO_USERS`. As written, one inherited grant row means a revoked role assignment is never deleted from the graph, permanently. Return a separate completeness flag for object grants and skip only `SnowflakeGrantMatchLink`.

cartography/intel/snowflake/grants.py
```diff
@@ -496,7 +497,17 @@ def sync(
         grants_by_role, grants_of_by_role = account_usage.split_grants(
             grants_to_roles, grants_to_users
         )
-        complete = True
+        inherited_count = sum(
```
jychp: Have `split_grants` return the skipped-row count and read it here. The skip predicate now lives in two places spelled differently (`if to_bool(...)` vs `to_bool(...) is True`), and this adds a second full pass over a row set that reaches millions on a large account.

cartography/intel/snowflake/account_usage.py
```diff
@@ -58,9 +58,10 @@
 WHERE deleted_on IS NULL
 """
 
+# Preview accounts expose IS_INHERITED; selecting it explicitly would fail on
+# accounts without that column. Keep optional grant metadata when it is available.
 _GRANTS_TO_ROLES_QUERY = """
-SELECT privilege, granted_on, name, table_catalog, table_schema, granted_to,
-       grantee_name, grant_option, granted_by, created_on
+SELECT *
```
jychp: Do not `SELECT *` on the largest query in this module. `GRANTS_TO_ROLES` is the biggest ACCOUNT_USAGE view the sync reads, and `*` ships `modified_on`, `deleted_on`, `object_id`, `granted_by_role_type` and every future column Snowflake adds, on every row, to fetch one flag. Resolve the column list once (`SHOW COLUMNS IN VIEW snowflake.account_usage.grants_to_roles`) and build the projection from it, or keep the explicit list and retry without `IS_INHERITED` when it fails to resolve.

cartography/intel/snowflake/util.py
```diff
@@ -219,19 +219,13 @@ def sf_fqn(*parts: str) -> str:
...
-
-    Deliberately not ``sf_fqn``: the REST path wants the *raw* Snowflake name,
-    not the dotted quoted form. ``sf_fqn`` would embed literal double quotes,
-    which Snowflake answers with a 404 for any database that needs quoting.
+    REST path parameters resolve SQL identifiers, so case-sensitive names need
+    identifier quoting before URL encoding. Escaping slashes keeps the entire
+    identifier in one path segment.
     """
-    return quote(name, safe="")
+    return quote(sf_fqn(name), safe="")
```
jychp, line 232: This reverses the docstring you deleted, which stated as fact that `sf_fqn`'s embedded double quotes make Snowflake answer 404 for any database that needs quoting. State in the docstring which endpoints were retested and against what, or the next person reverts this on the strength of the old comment.

cartography/intel/snowflake/util.py
```diff
@@ -471,18 +465,26 @@ def _request(self, method: str, url: str, **kwargs: Any) -> requests.Response:
...
         hanging it forever.
         """
         location = response.headers.get("Location")
+        if not location:
+            try:
+                body = response.json()
+            except ValueError:
+                body = {}
+            location = (
+                body.get("statementStatusUrl") if isinstance(body, dict) else None
```
jychp, line 517: Reject a polling URL that does not resolve under `self.host`. `_absolute()` returns any absolute URL verbatim and the poll loop calls `_apply_auth()` before each GET, so a `statementStatusUrl` pointing off-host sends the bearer token there. The `Location` header had the same gap; this widens it to the response body.

docs/root/modules/snowflake/config.md
```diff
@@ -150,16 +150,38 @@ GRANT USAGE ON WAREHOUSE CARTOGRAPHY_WH TO ROLE CARTOGRAPHY_RO;
...
+GRANT USAGE ON FUTURE SCHEMAS IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON ALL TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON FUTURE TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON ALL VIEWS IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON FUTURE VIEWS IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+```
+
+`REFERENCES` allows inspecting table and view metadata without granting `SELECT`
+on their contents. These grants are a baseline for databases, schemas, tables,
+and views; other object types can require additional privileges. A successful
```
jychp: Add `REFERENCES ON ALL/FUTURE MATERIALIZED VIEWS`, `EXTERNAL TABLES`, `DYNAMIC TABLES` and `ICEBERG TABLES IN DATABASE EXAMPLE_DB` to the block above. The module ingests all four (`materialized_views.py`, `external_tables.py`, `dynamic_tables.py`, `iceberg_tables.py`), so this recipe under-grants and the resulting gap is invisible.

tests/unit/cartography/intel/snowflake/test_account_usage.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+
+from cartography.intel.snowflake.account_usage import split_grants
+
+
+@pytest.mark.parametrize("inherited_from", ["ACCOUNT", "DATABASE", "SCHEMA"])
+@pytest.mark.parametrize("name", [None, "", "EXAMPLE_TABLE"])
+def test_inherited_grants_are_not_loaded_as_object_grants(inherited_from, name):
```
jychp: Add a case where `is_inherited` is absent from the row entirely. That is the shape every non-preview account returns and it is the one path this change must leave untouched; right now nothing asserts it.

jychp (changes requested):

cartography/intel/snowflake/grants.py
kunaals: good catch. object grants and role assignments now have separate completeness flags. inherited grants only suppress privilege-edge cleanup, so revoked role assignments and hierarchy edges can still be removed. added a two-tag regression covering both complete and incomplete account-grant coverage.

cartography/intel/snowflake/grants.py
kunaals: done. split_grants now counts inherited rows while skipping them and returns that count. this removes the duplicate predicate and the second pass.

cartography/intel/snowflake/account_usage.py
kunaals: replaced select * with an explicit projection. one show columns query checks whether is_inherited is available before including it. unavailable view metadata keeps coverage incomplete, and unexpected errors still propagate. tested both column layouts and verified the query against snowflake.

cartography/intel/snowflake/util.py
kunaals: added the snowflake rest identifier contract and noted the schema and network-rule listings we retested in the docstring. the regression tests also cover quoting and url encoding, including embedded quotes and slashes.

cartography/intel/snowflake/util.py
kunaals: fixed in the shared url resolver, covering location, statementstatusurl, and pagination links. urls must match the configured scheme, host, and port. added local http tests confirming that another origin receives no request or bearer token.

docs/root/modules/snowflake/config.md
kunaals: added the existing and future grants for all four object types. one adjustment: dynamic tables use monitor, since snowflake does not support references on them. linked the privilege docs and extended the schema-level future-grant guidance too.

tests/unit/cartography/intel/snowflake/test_account_usage.py
kunaals: added the missing-field case to the unit test alongside false and true values. it now explicitly verifies that ordinary grants survive when is_inherited is absent, while inherited rows are skipped and counted.

cartography/intel/snowflake/account_usage.py
```diff
@@ -100,7 +100,26 @@ def get_roles(client: SnowflakeClient) -> list[dict[str, Any]] | None:
...
+    )
+    if columns is None:
+        return None
+    if not columns:
+        warn_unavailable(
+            "grants from ACCOUNT_USAGE", "grant view columns are not visible"
+        )
+        return None
+    has_inherited = any(
+        (to_text(column.get("column_name")) or "").upper() == "IS_INHERITED"
```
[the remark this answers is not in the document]
kunaals: nah

cartography/intel/snowflake/account_usage.py
```diff
@@ -175,7 +194,13 @@ def split_grants(
     grants_by_role: dict[str, list[dict[str, Any]]] = {}
     grants_of_by_role: dict[str, list[dict[str, Any]]] = {}
 
+    inherited_count = 0
     for row in grants_to_roles:
+        # Inherited grants target a container's object type, not a named object.
+        # The caller suppresses object-grant cleanup but can still clean role assignments.
+        if to_bool(row.get("is_inherited")):
```
jychp: Restrict this skip to non-role securables. It drops every inherited row, including `granted_on IN (ROLE, DATABASE_ROLE)` rows that feed `grants_of_by_role`, yet `sync` still returns `role_assignments_complete=True` (grants.py:512). One such row would silently drop a hierarchy edge and the now-enabled hierarchy cleanup would then delete it: the exact failure the split was introduced to prevent. Guard with `to_text(row.get("granted_on")) not in _ROLE_SECURABLE_TYPES` so the completeness flag matches the data actually filtered.

cartography/intel/snowflake/account_usage.py
```diff
@@ -100,7 +100,26 @@ def get_roles(client: SnowflakeClient) -> list[dict[str, Any]] | None:
 @timeit
 def get_grants_to_roles(client: SnowflakeClient) -> list[dict[str, Any]] | None:
     """Every live grant to a role, or None when ACCOUNT_USAGE is unreadable."""
-    return _run(client, _GRANTS_TO_ROLES_QUERY, "grants from ACCOUNT_USAGE")
+    columns = _run(
+        client,
+        "SHOW COLUMNS IN VIEW snowflake.account_usage.grants_to_roles",
+        "grant columns from ACCOUNT_USAGE",
+    )
+    if columns is None:
```
jychp: Do not return `None` when only the column probe fails. `_run` maps every `is_sql_unavailable` error to `None`, so a `SHOW COLUMNS` refusal downgrades the whole grant sync to the per-role REST walk (thousands of calls, cleanup permanently off) even when `GRANTS_TO_ROLES` itself is readable, and `warn_unavailable` then blames a missing `IMPORTED PRIVILEGES` grant that is present. Fall back to the fixed projection without `is_inherited` instead.

cartography/intel/snowflake/grants.py
```diff
@@ -493,10 +505,18 @@ def sync(
     )
 
     if grants_to_roles is not None and grants_to_users is not None:
-        grants_by_role, grants_of_by_role = account_usage.split_grants(
+        grants_by_role, grants_of_by_role, inherited_count = account_usage.split_grants(
             grants_to_roles, grants_to_users
         )
-        complete = True
+        object_grants_complete = inherited_count == 0
```
jychp: Bound this. Any account with the inherited-grants preview enabled reports inherited rows on every sync, so `object_grants_complete` is false forever and `HAS_PRIVILEGE` edges are never cleaned up again: genuinely revoked privileges persist indefinitely behind a warning log. The rows carry `inherited_from`, `inherited_from_database` and `inherited_from_schema`, and the graph already holds account, database and schema `HAS_PRIVILEGE` edges, so map them onto the container node and keep coverage complete instead of disabling cleanup.

cartography/intel/snowflake/grants.py
```diff
@@ -493,10 +505,18 @@ def sync(
...
+        role_assignments_complete = True
+        if inherited_count:
+            logger.warning(
+                "Snowflake ACCOUNT_USAGE reported %d inherited grants. "
+                "Container-scoped inherited privileges are not yet modeled; "
+                "object grant coverage is incomplete and object grant cleanup will be skipped.",
+                inherited_count,
+            )
         logger.info(
             "Read %d Snowflake grant rows and %d role assignment rows from "
```
jychp: Subtract `inherited_count` from this count. `len(grants_to_roles)` now includes the rows that were skipped, so the log overstates what was loaded.

cartography/intel/snowflake/grants.py
```diff
@@ -505,9 +525,7 @@ def sync(
             account_id,
         )
     else:
-        grants_by_role, grants_of_by_role, complete = _walk_rest(
-            client, roles, database_roles
-        )
+        grants_by_role, grants_of_by_role, _ = _walk_rest(client, roles, database_roles)
```
jychp: Drop the third return value from `_walk_rest`. Nothing reads it now, and its docstring (grants.py:424-428) still claims it is "reported only so the caller can log the difference" between refused reads and non-exhaustive ones. Either log it here or delete it and the paragraph.

cartography/intel/snowflake/util.py
```diff
@@ -500,9 +506,34 @@ def _await_async(self, response: requests.Response) -> requests.Response:
 
     def _absolute(self, url: str) -> str:
         """Resolve a possibly-relative API URL against the account host."""
-        if url.startswith("http://") or url.startswith("https://"):
-            return url
-        return f"{self.host}{url}"
+        resolved = urljoin(f"{self.host}/", url)
```
jychp, line 539: Pass `allow_redirects=False` on the poll and pagination requests and re-validate any redirect target through here. `_absolute` checks the advertised URL, then `requests` follows redirects by default, so a same-origin `Location` that 302s to another host still reaches it and its body is parsed as Snowflake data. The token is stripped by `rebuild_auth`, the response substitution is not.

docs/root/modules/snowflake/config.md
```diff
@@ -150,16 +150,50 @@ GRANT USAGE ON WAREHOUSE CARTOGRAPHY_WH TO ROLE CARTOGRAPHY_RO;
...
+GRANT REFERENCES ON ALL VIEWS IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON FUTURE VIEWS IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON ALL MATERIALIZED VIEWS IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON FUTURE MATERIALIZED VIEWS IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON ALL EXTERNAL TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON FUTURE EXTERNAL TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON ALL ICEBERG TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT REFERENCES ON FUTURE ICEBERG TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT MONITOR ON ALL DYNAMIC TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
+GRANT MONITOR ON FUTURE DYNAMIC TABLES IN DATABASE EXAMPLE_DB TO ROLE CARTOGRAPHY_RO;
```
jychp, line 178: Add event tables to this block. `_SCHEMA_LEVEL_MODULES` ingests them via `event_tables.py` (`/schemas/{schema}/event-tables`), so this recipe under-grants exactly the way the account-scoped version did, and the "other object types can require additional privileges" hedge below makes the gap invisible. If event tables need no extra grant, say so explicitly.

jychp (changes requested):

cartography/intel/snowflake/account_usage.py
kunaals: added the guard so role and database-role rows still reach the hierarchy transform even if flagged as inherited. added regression coverage for both.

cartography/intel/snowflake/account_usage.py
kunaals: removed the metadata probe. the explicit query now handles missing inherited-grant support and the source-account preview layout through narrowly matched missing-column errors. unexpected failures still propagate.

cartography/intel/snowflake/grants.py
kunaals: agreed, the previous safeguard could retain revoked privileges indefinitely. inherited grants now have separate records linked to their principal and container, preserving the object type rather than treating table select as a privilege on the database itself. added revocation tests and verified the new records against snowflake.

cartography/intel/snowflake/grants.py
kunaals: the old count described rows read, but the distinction wasn’t clear enough. split the log into direct grants, inherited grants, and user-role assignment rows.

cartography/intel/snowflake/grants.py
kunaals: removed the unused return value and its outdated docstring.

cartography/intel/snowflake/util.py
kunaals: fixed in the shared request path. automatic redirects are disabled, each redirect target is checked against the configured origin, and redirects are bounded. added polling and pagination tests confirming that off-origin targets receive no request.

docs/root/modules/snowflake/config.md
kunaals: added references grants for existing and future event tables in both setup guides, including the schema-level future-grant precedence guidance.

docs/root/modules/snowflake/config.md
```diff
@@ -169,7 +205,73 @@ collector's own role can see, and a partial answer is indistinguishable from a
...
+example, `GRANT INHERITED REFERENCES ON ALL TABLES IN ACCOUNT TO ROLE
+CARTOGRAPHY_RO` is different from the unsupported ordinary account-wide grant.
+
+They are not required for this setup. Enabling them requires the
+[public-preview opt-in](https://docs.snowflake.com/en/user-guide/inherited-grants-intro);
+`SYSTEM$ENABLE_PREVIEW_ACCESS()` enables preview access for the whole account,
+not just the collector. Inherited grants apply to each specified object type:
+a grant on tables does not also cover views or dynamic tables.
+
+Cartography records inherited grants as `SnowflakeInheritedGrant` nodes, preserving
```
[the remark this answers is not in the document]
kunaals: the snowflake schema page is generated from the model during the docs build, so it should not be added as a handwritten file. added the missing property descriptions to the model and rebuilt the docs. the generated page includes the node properties and relationships, and the schema documentation tests pass.

tests/integration/cartography/intel/snowflake/test_grants.py
```diff
@@ -563,3 +571,273 @@ def test_quoted_database_role_keeps_its_grant_edges(
...
+            check_rels(
+                neo4j_session,
+                "SnowflakeRole",
+                "name",
+                "SnowflakeInheritedGrant",
+                "id",
+                "HAS_INHERITED_GRANT",
+            )
+        )
+        == 2
```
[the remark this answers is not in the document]
kunaals: fixed. the assertion now checks the exact source role and both inherited object types instead of only counting the relationships. the integration tests pass.

tests/integration/cartography/intel/snowflake/test_grants.py
```diff
@@ -563,3 +571,273 @@ def test_quoted_database_role_keeps_its_grant_edges(
...
+    )
+    direct = [
+        row
+        for row in SNOWFLAKE_ACCOUNT_USAGE_GRANTS_TO_ROLES
+        if row["granted_on"] == "TABLE"
+    ]
+    inherited = {
+        "granted_to": "ROLE",
+        "grantee_name": "SAFETY_INSPECTOR",
+        "granted_on": "TABLE",
```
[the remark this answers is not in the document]
kunaals: this guard is already covered by test_inherited_flag_does_not_drop_role_hierarchy, parameterized for both role and database-role rows with the inherited flag set. removing the guard makes those tests fail. keeping this integration case focused on inherited object-grant refresh and revocation.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,116 @@
...
+            )
+        if scope == "ACCOUNT":
+            container_id = account_id
+        elif scope == "DATABASE" and database:
+            container_id = sf_id(account_id, "database", sf_fqn(database))
+        elif scope == "SCHEMA" and database and schema:
+            container_id = sf_id(account_id, "schema", sf_fqn(database, schema))
+        else:
+            raise ValueError(
+                "Snowflake inherited grant has an invalid or incomplete container scope"
```
jychp: Collect and warn on a malformed row instead of raising. `load_grants` runs inside `grants.sync` before `transform_grants` and every load, so one row with a NULL scope aborts the whole Snowflake sync: no direct grants, no role-hierarchy edges, no cleanup for any surface. The compatibility projection in `get_grants_to_roles` makes this reachable on real data, because its `CASE` has no `ELSE` and yields NULL when all three `inherited_from_*` columns are NULL. Skip the row, log it with the grantee and privilege, and report object grants incomplete.

cartography/intel/snowflake/util.py
```diff
@@ -461,28 +459,67 @@ def _apply_auth(self) -> None:
...
+            target = self._absolute(urljoin(url, response.headers["Location"]))
+            response.close()
+            if response.status_code in (302, 303) and method != "HEAD":
+                method = "GET"
+            elif response.status_code == 301 and method == "POST":
+                method = "GET"
+            if response.status_code not in (307, 308):
+                kwargs.pop("json", None)
+                kwargs.pop("data", None)
+            kwargs.pop("params", None)
```
jychp, line 490: Keep `params` on 307/308. `run_sql` fetches each result partition with `params={"partition": n}`; on a method-preserving redirect this drops the parameter and silently re-reads partition 0, so a large `GRANTS_TO_ROLES` result comes back with duplicated rows instead of the remaining pages. Pop `params` only in the same branch that drops `json`/`data`.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,116 @@
...
+def transform(rows: list[dict[str, Any]], account_id: str) -> list[dict[str, Any]]:
+    grants = []
+    for row in rows:
+        scope = to_text(row.get("inherited_from"))
+        database = to_text(row.get("inherited_from_database"))
+        schema = to_text(row.get("inherited_from_schema"))
+        privilege = to_text(row.get("privilege"))
+        object_type = to_text(row.get("granted_on"))
+        grantee = to_text(row.get("grantee_name"))
+        grantee_type = (to_text(row.get("granted_to")) or "").replace("_", " ")
```
jychp: **suggestion** — Store `granted_to` verbatim on the node. This rewrite turns `APPLICATION_ROLE` into `APPLICATION ROLE`, so `grantee_type`, whose description promises that unsupported principal kinds "retain their provider identity", does not hold the value Snowflake reported. Normalize into a local variable for the principal-kind lookup only.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,116 @@
...
+from cartography.intel.snowflake.sql_values import to_text
+from cartography.intel.snowflake.util import iso_to_datetime
+from cartography.intel.snowflake.util import sf_fqn
+from cartography.intel.snowflake.util import sf_id
+from cartography.models.snowflake.inherited_grant import SnowflakeInheritedGrantSchema
+
+logger = logging.getLogger(__name__)
+
+
+def transform(rows: list[dict[str, Any]], account_id: str) -> list[dict[str, Any]]:
```
jychp: **suggestion** — Add a module docstring and docstrings on `transform`, `load_grants` and `cleanup`. Every sibling module under `cartography/intel/snowflake/` opens by naming its source view and its modeling trade-offs; this file introduces the only new node type in the PR and explains none of it in code.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def load_grants(
+    neo4j_session: neo4j.Session,
+    rows: list[dict[str, Any]],
+    account_id: str,
+    update_tag: int,
+) -> None:
+    grants = transform(rows, account_id)
+    logger.info("Loading %d Snowflake inherited grant records.", len(grants))
```
jychp: The load function has an auto log, you should not duplicate the log here.

tests/integration/cartography/intel/snowflake/test_grants.py
```diff
@@ -563,3 +571,268 @@ def test_quoted_database_role_keeps_its_grant_edges(
...
+
+    # Assert
+    assert check_nodes(
+        neo4j_session, "SnowflakeInheritedGrant", ["object_type", "lastupdated"]
+    ) == {("VIEW", TEST_UPDATE_TAG + 2)}
+
+
+def test_inherited_grant_cleanup_is_account_scoped(neo4j_session):
+    # Arrange
+    from cartography.intel.snowflake.inherited_grants import cleanup
```
jychp: Move these two imports to the top of the file with the rest.

tests/integration/cartography/intel/snowflake/test_grants.py
```diff
@@ -563,3 +571,268 @@ def test_quoted_database_role_keeps_its_grant_edges(
...
+    row = {
+        "granted_to": "ROLE",
+        "grantee_name": "READER",
+        "granted_on": "TABLE",
+        "privilege": "SELECT",
+        "inherited_from": "ACCOUNT",
+    }
+    for account in accounts:
+        neo4j_session.run(
+            "MERGE (:SnowflakeAccount:SnowflakeSecurable {id: $id})", id=account
```
jychp: **suggestion** — Delete these accounts and their grants when the test finishes. `neo4j_session` is module-scoped, so `EXAMPLE.FIRST` and `EXAMPLE.SECOND` plus their `SnowflakeInheritedGrant` nodes survive into every later test in this module; the `_clear_grant_edges` docstring already records how leftover state produced a false pass here once.

docs/root/modules/snowflake/config.md
```diff
@@ -169,7 +205,73 @@ collector's own role can see, and a partial answer is indistinguishable from a
...
+on views. Unsupported principal kinds retain their provider identity on the grant
+record without a principal link.
+
+```cypher
+MATCH (p:SnowflakePrincipal)-[:HAS_INHERITED_GRANT]->(g:SnowflakeInheritedGrant)
+      -[:APPLIES_IN]->(container:SnowflakeSecurable)
+RETURN p.id, g.privilege, g.object_type, g.container_type, container.id
+```
+
+Direct privileges remain `HAS_PRIVILEGE` edges. Both direct and inherited grants
```
jychp: **suggestion** — Name the read that gates inherited-grant cleanup. The code runs it under role-assignment completeness rather than object-grant completeness, so "cleaned up after complete reads" leaves an operator unable to predict what happens when only one of the two `ACCOUNT_USAGE` reads succeeds.

jychp (changes requested):

cartography/intel/snowflake/inherited_grants.py
kunaals: fixed: malformed inherited rows now warn and are skipped, while valid grants and role assignments continue loading. inherited cleanup has its own completeness flag, so we preserve uncertain inherited records without retaining revoked direct privileges. added a regression covering that mixed result.

cartography/intel/snowflake/util.py
kunaals: i checked this against the http semantics and requests behavior. 307/308 preserve the method and body, but location supplies the destination query. retaining params would append them again when location already contains partition=1. kept the existing behavior and added a regression for both statuses that verifies distinct partition rows without duplicate parameters.

cartography/intel/snowflake/inherited_grants.py
kunaals: removed the extra log; the shared loader already reports the count.

tests/integration/cartography/intel/snowflake/test_grants.py
kunaals: moved both imports to the top of the file.

cartography/intel/snowflake/account_usage.py
```diff
@@ -175,7 +223,15 @@ def split_grants(
...
     grants_of_by_role: dict[str, list[dict[str, Any]]] = {}
 
+    inherited_grants: list[dict[str, Any]] = []
     for row in grants_to_roles:
+        # Role hierarchy rows must reach the assignment transform even if flagged.
+        if (
+            to_bool(row.get("is_inherited"))
+            and to_text(row.get("granted_on")) not in _ROLE_SECURABLE_TYPES
+        ):
+            inherited_grants.append(row)
```
jychp: **blocking** — Keep building an object grant when the row names a concrete object. Any `is_inherited` row that carries `name` / `table_catalog` / `table_schema` is diverted here and its `HAS_PRIVILEGE` edge to that table or view is discarded, so a role that demonstrably reads `DB.SCHEMA.T` shows no edge to it at all; `test_inherited_grants_are_not_loaded_as_object_grants` parameterizes `name` over `"EXAMPLE_TABLE"` and locks the loss in. Route only name-less container rows to `inherited_grants` and let named rows fall through to `grants_by_role`.

cartography/intel/snowflake/account_usage.py
```diff
@@ -100,7 +103,48 @@ def get_roles(client: SnowflakeClient) -> list[dict[str, Any]] | None:
...
+            raise error
+        if not isinstance(detail, dict) or str(detail.get("code")).zfill(6) != "000904":
+            raise
+        missing = re.search(
+            r"invalid identifier '(IS_INHERITED|INHERITED_FROM)'",
+            str(detail.get("message", "")),
+            re.IGNORECASE,
+        )
+        if not missing:
+            raise
```
jychp: **blocking** — Retry with `optional = ""` and a warning instead of re-raising. `SnowflakeSqlError` is caught nowhere between here and `sync.run`, so a `000904` naming `INHERITED_FROM_DATABASE`, `INHERITED_FROM_SCHEMA` or `INHERITED_FROM_ACCOUNT` aborts the whole Snowflake stage before `_cleanup` ever runs, leaving every surface's stale data in place. This query previously projected only documented columns, so adding four preview columns makes that failure mode newly reachable.

cartography/intel/snowflake/grants.py
```diff
@@ -493,21 +497,26 @@ def sync(
     )
 
     if grants_to_roles is not None and grants_to_users is not None:
-        grants_by_role, grants_of_by_role = account_usage.split_grants(
+        grants_by_role, grants_of_by_role, inherited_rows = account_usage.split_grants(
             grants_to_roles, grants_to_users
         )
-        complete = True
+        inherited_grants_complete = inherited_grants.load_grants(
```
jychp: **blocking** — Have `get_grants_to_roles` report whether the inheritance columns were actually projected and AND that into this flag. When `IS_INHERITED` is unavailable the retry drops all four columns, `split_grants` finds no inherited rows, `load_grants` returns `len([]) == len([])` → `True`, and `cleanup` then deletes every `SnowflakeInheritedGrant` collected on an earlier run. Completeness currently only reflects malformed-row drops, not a degraded read.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,124 @@
...
+                    account_id,
+                    "inherited_grant",
+                    hashlib.sha256(identity.encode()).hexdigest(),
+                ),
+                "principal_id": principal_id,
+                "container_id": container_id,
+                "privilege": privilege,
+                "object_type": object_type,
+                "container_type": scope,
+                "database_name": database,
```
jychp: **suggestion** — Null `database_name` and `schema_name` outside their own scope. An `ACCOUNT` row that still carries `inherited_from_database` / `inherited_from_schema` yields a node whose `container_type` is `ACCOUNT` but whose properties claim database and schema provenance, contradicting their own PropertyRef descriptions; `test_inherited_grants_refresh_and_revoke_without_preserving_direct_grants` feeds exactly that shape and asserts nothing about it.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,124 @@
...
+def transform(rows: list[dict[str, Any]], account_id: str) -> list[dict[str, Any]]:
+    grants = []
+    for row in rows:
+        scope = to_text(row.get("inherited_from"))
+        database = to_text(row.get("inherited_from_database"))
+        schema = to_text(row.get("inherited_from_schema"))
+        privilege = to_text(row.get("privilege"))
+        object_type = to_text(row.get("granted_on"))
+        grantee = to_text(row.get("grantee_name"))
+        grantee_type = (to_text(row.get("granted_to")) or "").replace("_", " ")
```
jychp: **suggestion** — Store the raw `granted_to` value in `grantee_type` and normalize only for the `principal_kind` lookup. `.replace("_", " ")` rewrites `APPLICATION_ROLE` to `APPLICATION ROLE`, so the property documented as "unsupported principal kinds retain their identity here" holds a string Snowflake never reported.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,124 @@
...
+            )
+            continue
+        if scope == "ACCOUNT":
+            container_id = account_id
+        elif scope == "DATABASE" and database:
+            container_id = sf_id(account_id, "database", sf_fqn(database))
+        elif scope == "SCHEMA" and database and schema:
+            container_id = sf_id(account_id, "schema", sf_fqn(database, schema))
+        else:
+            logger.warning(
```
jychp: **suggestion** — Aggregate these skips into one warning per sync with a count and a few examples. `GRANTS_TO_ROLES` reaches millions of rows on a large account, and the `CASE` fallback in `get_grants_to_roles` has no `ELSE`, so a whole class of rows lands here and logs one line each.

cartography/intel/snowflake/inherited_grants.py
```diff
@@ -0,0 +1,124 @@
...
+                privilege,
+                grantor,
+            ]
+        )
+        grants.append(
+            {
+                "id": sf_id(
+                    account_id,
+                    "inherited_grant",
+                    hashlib.sha256(identity.encode()).hexdigest(),
```
jychp: **suggestion** — Build a readable qualified name instead of a SHA-256 digest. `docs/root/modules/snowflake/index.md` documents every Snowflake id as `<ORG>.<ACCOUNT>/<object type>/<qualified name>`, and `sf_id`'s own docstring rests on that readability; an opaque digest cannot be traced back to its grant from the graph, and any future change to the field order silently re-keys every node.

jychp (approved):

PR #3293, opened 2026-09-18
fix(aws): prevent false ELBV2 exposure from overlapping private IPs

kunaals:
### Type of change

### Summary

The ECS transient-failure prerequisite landed in #3294. This PR now targets `master` and contains only the ELBV2 target-resolution fix.

ELBV2 IP targets currently match every private-IP node sharing the registered address. When unrelated VPCs reuse an address, a load balancer can acquire `EXPOSE` relationships to the wrong ENIs and mark their ECS containers as internet-exposed. The MatchLink's account metadata scopes cleanup, but does not constrain endpoint matching.

Use a typed analysis job to resolve each registration to one ENI-specific private-IP identity and create the relationship. Local targets use the target-group VPC and load-balancer region; cross-VPC ECS targets use the service's explicit target-group registration and task ENI. This preserves shared-VPC targets owned by another account without treating an IP address as a globally unique identifier. IP resolution runs after ECS ingestion. `AddRelationship` generates the write and cleanup; no IP-target MatchLink, separate read-back query, or additional AWS API call is needed.

A successful sync enables generated account-scoped cleanup of old incorrect IP relationships, then the existing exposure analyses remove obsolete container relationships and flags. ELBV2 regional failures preserve previous relationships. The ECS inventory preservation needed after transport failures and handled server errors is covered by #3294; existing AWS access-denial cleanup semantics remain unchanged.

### Related issues or links

- Merged prerequisite: #3294.
- [AWS: ALB IP target types and remote targets](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-type)
- [AWS: target registration fields](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_TargetDescription.html)
- [AWS: ECS cross-VPC IP targets](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/alb.html)

### How was this tested?

107 focused tests passed. The new regressions exercise real ELBV2 collection with mocked AWS clients, identity resolution, and cleanup. Full repository lint and the documentation build also passed.

#### General

#### Proof of functionality

- [ ] Included public real-provider validation evidence and sanitized sync logs (`needs-tests`).
- [x] Removed credentials, personal data, account identifiers, and other sensitive values from the evidence. All new fixtures are synthetic.

- [x] Updated model docstrings and `PropertyRef.description` values used to generate schema documentation. The IP-target relationship is declared by the typed analysis job; the private-IP lookup property is explicitly indexed.
- [x] Built the documentation with `uv run --frozen --group doc ./docs/build.sh`.

### Notes for reviewers

Registrations with missing or ambiguous identity evidence remain unlinked. In particular, generic remote targets without explicit ECS identity evidence need additional routing or provider evidence; a single matching address in the current graph is insufficient.

ECS access-denied responses retain their existing cleanup behavior; remote targets may remain unresolved when that removes their identity evidence. The prerequisite protects transport failures and the server error that the shared decorator otherwise treats as an empty inventory.

Cross-account ingestion ordering is unchanged: a newly discovered shared-VPC target may resolve on the following sync if its owning account is ingested after the load-balancer account.


cartography/intel/aws/resources.py
```diff
@@ -168,10 +165,12 @@
...
         "ec2:autoscalinggroup": sync_ec2_auto_scaling_groups,
         "ec2:keypair": sync_ec2_key_pairs,
-        # Keep ECS last among regular resources. Its tasks and containers are
+        # Keep ECS close to analysis. Its tasks and containers are
         # short-lived, so syncing them close to AWS analysis minimizes the period
         # in which replacements lack analysis-derived relationships. EC2 instances
         # and load balancers must already exist for ECS relationships.
         "ecs": ecs.sync,
+        # Resolve IP targets after ENIs and ECS service target-group registrations.
+        "ec2:load_balancer_v2:expose": sync_load_balancer_v2_expose,
```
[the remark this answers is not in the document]
kunaals: updated after narrowing the prerequisite: #3294 preserves inventory after transport failures and InternalServerErrorException while retaining the shared decorator's retries and access-denial behavior. the regression refreshes tasks before service enumeration times out, then checks ip/container exposure and later cleanup. access-denied responses retain existing cleanup semantics.

tests/integration/cartography/intel/aws/ec2/test_load_balancer_v2s.py
```diff
@@ -256,7 +256,17 @@ def test_sync_load_balancer_v2_expose(mock_get_loadbalancer_v2_data, neo4j_sessi
     # Create AWSEC2PrivateIp nodes (normally created by ec2:network_interface)
     for ip in ["10.0.0.50", "10.0.0.51"]:
         neo4j_session.run(
-            "MERGE (ip:AWSEC2PrivateIp{id: $ip}) SET ip.lastupdated = $tag, ip.private_ip_address = $ip",
+            """
+            MERGE (ip:AWSEC2PrivateIp{id: $ip})
+            SET ip.lastupdated = $tag, ip.private_ip_address = $ip
+            MERGE (eni:AWSNetworkInterface {id: 'eni-' + $ip})
```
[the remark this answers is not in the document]
kunaals: the collision regressions are already in test_lb_ip_target_resolution.py. they seed the old address-only matches, then assert the complete ip/container exposure sets and flags across reused ips in other accounts, vpcs, and regions, plus ambiguous local and remote targets. those assertions catch both removal of the scope filter and removal of the uniqueness guard, so keeping this existing happy-path test focused.

tests/integration/cartography/intel/aws/ec2/test_lb_ip_target_resolution.py
```diff
@@ -0,0 +1,560 @@
...
+        )
+    _analyze(neo4j_session, 2)
+
+    # Assert: tasks were refreshed even though their service could not be read.
+    assert check_nodes(neo4j_session, "AWSECSTask", ["id", "lastupdated"]) == {
+        ("task-remote", 2),
+        ("task-other", 1),
+    }
+    # Continue other regions without erasing prior identity or exposure.
+    healthy_client.get_paginator.assert_called_with("list_clusters")
```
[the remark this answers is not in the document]
kunaals: replaced the mock-call assertion with graph assertions in both prs. the healthy region now returns a synthetic cluster, and the tests check its region, current update tag, and account relationship through real ecs sync. the existing assertions did not prove that region was processed, so simply deleting the mock assertion would have weakened coverage. temporarily stopping sync after a transient failure makes all five affected cases fail at the new graph assertion; normal collection passes.

cartography/analysis/aws/analysis.py
```diff
@@ -49,6 +49,60 @@
...
+                    properties={
+                        "port": Var("target.Port"),
+                        "protocol": Var("target.Protocol"),
+                        "target_group_arn": Var("target.TargetGroupArn"),
+                        "_sub_resource_label": "AWSAccount",
+                        "_sub_resource_id": Param("AWS_ID"),
+                    },
+                    # Load batches without cleanup. Only a complete account inventory
+                    # enables cleanup, including edges written by the legacy MatchLink.
+                    cleanup_where="$CLEANUP_SAFE AND r._sub_resource_label = 'AWSAccount' AND r._sub_resource_id = $AWS_ID",
```
achantavy, line 104: Nit: should we be using `_` fields for something like this?

cartography/analysis/aws/analysis.py
```diff
@@ -49,6 +49,60 @@
...
+                WITH target, lb
+                MATCH (:AWSELBV2TargetGroup {id: target.TargetGroupArn})
+                      -[:TARGETS]->(:AWSECSService)
+                      <-[:WORKLOAD_PARENT]-(:AWSECSTask)
+                      -[:NETWORK_INTERFACE]->(:AWSNetworkInterface)
+                      -[:PRIVATE_IP_ADDRESS]->(ip:AWSEC2PrivateIp {private_ip_address: target.TargetId})
+                RETURN ip
+            }
+            WITH lb, target, collect(DISTINCT ip) AS candidates
+            WHERE size(candidates) = 1
```
achantavy, line 83: Why exactly 1?

cartography/analysis/aws/analysis.py
```diff
@@ -49,6 +49,60 @@
...
+AWS_LB_IP_TARGET_EXPOSURE = AnalysisJob(
+    name="Resolve ELBV2 IP targets to private IP identities",
+    short_name="aws_lb_ip_target_exposure",
+    scope=ScopeById("AWSAccount", "AWS_ID", scope_on="lb"),
+    statements=(
+        AnalysisStatement(
+            comment="Resolve local/shared-VPC targets or explicit ECS service targets; skip ambiguous identities.",
+            # Filter the account's resources before expanding target registrations.
+            match="""
+            MATCH (lb:AWSLoadBalancerV2)
```
achantavy, line 63: Non-block: I ran this through claude a few times and it's telling me there's a cartesian product that the underlying analysis job abstraction framework introduces, but tbh I've stared at it a few times and I don't quite understand it. Definitely not a blocker but we should at some point take a close look and get the perf good if this is actually valid.

achantavy (approved):

cartography/analysis/aws/analysis.py
kunaals: these are the existing matchlink ownership fields. keeping them lets the new cleanup recognize and remove edges written by previous versions. added a comment explaining the compatibility requirement.

cartography/analysis/aws/analysis.py
kunaals: one candidate per registered target, not one target per load balancer. multiple distinct eni/ip identities mean the address is still ambiguous; linking all of them would recreate the false exposure. added a comment clarifying this.

cartography/analysis/aws/analysis.py
kunaals: confirmed the row multiplication on synthetic data. grouped each batch by load balancer so it expands only that balancer’s registrations. on neo4j 5.26.21, 500 balancers and 1,000 registrations now produce 1,000 unwind rows instead of 500,000, with identical matches. added coverage for multiple balancers across a batch boundary; no framework changes needed.

